PR #695, opened 2014-05-17
Signature Speed Up

antonio-fr:
Slightly modify ECDSA lib, so that signature have a parity bit for yR.
This is as described in ECDSA-SEC1 v2.
Bitcoin signature are 2 times faster. No need to try and test anymore.


ecdsa:
closing, out of scope

PR #824, opened 2014-09-06
Show op_return in tx dialogue

dabura667:
Very basic as of now... needs more polish. But I am open to more suggestions as to how to sanitize etc.

It seems as though these two places are the only three places necessary to show op_returns.

Please let me know if you have any other info / suggestions.

P.S.: If you load `d29c9c0e8e4d2a9790922af73f0b8d51f0bd4bb19940d9cf910ead8fbe85bc9b` "From the blockchain" with this commit, you will not be disappointed.
Hint: here's a screen cap. (I made the text box larger to accommodate.)
http://i.imgur.com/QOKyRO1.png

### DIFF
--- a/gui/qt/transaction_dialog.py
+++ b/gui/qt/transaction_dialog.py
@@ -212,7 +212,7 @@ class TxDialog(QDialog):
         vbox.addWidget(i_text)
 
         vbox.addWidget(QLabel(_("Outputs")))
-        lines = map(lambda x: x[0] + u'\t\t' + self.parent.format_amount(x[1]), self.tx.get_outputs())
+        lines = map(lambda x: x[0] + u'\t\t' + self.parent.format_amount(x[1]) if x[1] else x[0], self.tx.get_outputs())
         o_text = QTextEdit()
         o_text.setText('\n'.join(lines))
         o_text.setReadOnly(True)
--- a/lib/transaction.py
+++ b/lib/transaction.py
@@ -429,6 +429,11 @@ def get_address_from_output_script(bytes):
     if match_decoded(decoded, match):
         return 'address', hash_160_to_bc_address(decoded[1][1],5)
 
+    # OP_RETURN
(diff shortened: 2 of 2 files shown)


harding:
Hi,

Would you consider reverting this? @luke-jr and @gmaxwell have identified several problems with this feature. Paraphrasing some of their concerns:

1. **Increased risk of Bitcoin censorship:** Both Bitcoin Core nodes and Electrum servers that relay copyrighted, classified, or otherwise protected content are at increased risk of being targeted by the government for shutdown or by copyright owners for litigation.

1. **Increased risk of spam:** Spammers can now send full text messages to Electrum 2.0 users for a relatively low cost to the spammer (but a relatively high cost to full node operators who need to now store all those spam-containing transactions).

1. **Less privacy and fungibility:** Data placed on the block chain is public forever. Most people don't realize the implication of that, and some will undoubtedly post things they want to take back someday.  In addition, some people may not want coins that have an undesirable message in their history (e.g. something racist), reducing the fungibility of coins.

1. **Does not scale:** Every byte in a confirmed op_return output is a byte that wasn't used for another transaction. If there's significant competition between transaction data and message data, steps will likely be taken to prohibit message data. Users and businesses should not come to depend on the ability to leave text messages in the block chain.

I know there are other implementations available for adding and viewing text in the block chain, but adding this feature to a widely-used wallet like Electrum makes it significantly more accessible to less-technical users, likely leading to the problems described above at an accelerated rate.

For Luke's and Greg's concerns in their own words, please see [Bitcoin.org pull #704](https://github.com/bitcoin/bitcoin.org/pull/704). (I apologize for any aggressive phrasing used there; most of us would like to keep this conversation amicable.)

Thank you for considering this request to revert.  If there's anything I can do to help, please let me know.  Also please let me know if I should move this request to a new issue or pull request.

ecdsa:
note: I answered there.

ecdsa:
@harding done here: 9a6d98f899c112ceb88e5543163fcf552e82df09

PR #880, opened 2014-09-30
Add GreenAddress 2of3 recovery plugin

greenaddress:
This plugin adds a new wallet type which allows recovering/accessing GreenAddress 2of3 subaccounts (which were just deployed to GreenAddress production).

The new 2of3 model (similarly to BitGo's wallet) allows users to keep complete control of their funds using an additional backup key which in conjunction with wallet mnemonic can be used to access funds without GreenAddress signature.

This new Electrum wallet type allows setting up an Electrum wallet given (1) wallet's mnemonic, (2) subaccount backup mnemonic, (3) GreenAddress extended public key. These 3 values are provided to user on creation of "2of3 subaccount". Such Electrum wallet allows unlimited access to user's funds. It can be used in the case GreenAddress goes down or if you for some other reasons prefer to access your funds without our signatures.

Implementation notes:
 1. Trezor's `mnemonic` (https://github.com/trezor/python-mnemonic) module is required for the plugin to work.
 2. A 'multisig_recovery' wallet 'supertype' is added in `installwizard.py` to allow specifying wallet types which should be available only for recovery. It can be later used for other similar wallet providers like BitGo.
 3. Some code is duplicated where the original code was using `sorted()` pubkeys order, because GreenAddress uses always the same order of keys in multisig (2,0,1 - GA key, wallet key, subaccount key).
 4. Change addresses are not used because GreenAddress uses the same branch for change as for spending. The first level of BIP32 sequence is instead used to derive our 'REGULAR' (=1) branch which is usually used for payments. ('external social' transactions are done through 'EXTERNAL' (=2) priv-derived branch - not supported at the moment in this plugin)
 5. BIP32 child numbers are 1-based, to make sure address lists are the same as shown in GreenAddress.
 6. It was tested using regtest (with quickly hacked Electrum client to support testnet address versions) and production. It allows accessing funds in GA accounts and also limited wallet usage, like monitoring addresses and spending, but as spending doesn't use change addresses, it's recommended to avoid using it for day-to-day use. The main intention of this plugin is to allow accessing GreenAddress funds at any time, and second intention is to allow proof-of-concept verifying that users have full control of their funds.

### DIFF
--- a/gui/qt/installwizard.py
+++ b/gui/qt/installwizard.py
@@ -328,7 +328,10 @@ class InstallWizard(QDialog):
         if action == 'new':
             action, wallet_type = self.restore_or_create()
             if wallet_type == 'multisig':
-                wallet_type = self.choice(_("Multi Signature Wallet"), 'Select wallet type', [('2of2', _("2 of 2")),('2of3',_("2 of 3"))])
+                subtypes = [('2of2', _("2 of 2")),('2of3',_("2 of 3"))]
+                if action == 'restore':
+                    subtypes += map(lambda x:(x[1],x[2]), filter(lambda x:x[0]=='multisig_recovery', electrum.wallet.wallet_types))
+                wallet_type = self.choice(_("Multi Signature Wallet"), 'Select wallet type', subtypes)
                 if not wallet_type:
                     return
             elif wallet_type == 'hardware':
--- a/plugins/greenaddress_2of3_restore.py
+++ b/plugins/greenaddress_2of3_restore.py
@@ -1,0 +1,269 @@
+from PyQt4.Qt import QVBoxLayout
+import PyQt4.QtCore as QtCore
+
+import electrum
+from electrum_gui.qt.util import ok_cancel_buttons2
+from electrum_gui.qt import seed_dialog
+from electrum.account import BIP32_Account
+from electrum.bitcoin import hash_160_to_bc_address, hash_160, deserialize_xkey, CKD_pub, bip32_root, \
+    bip32_private_derivation, int_to_hex, DecodeBase58Check
+from electrum.i18n import _
+from electrum.plugins import BasePlugin, hook
+from electrum.transaction import Transaction
+from electrum.wallet import Wallet_2of3, Wallet
+from electrum.account import BIP32_Account_2of3
+
+from electrum.util import print_msg
+
+try:
+    import mnemonic
+    loaded_mnemonic = True
+except ImportError:
+    loaded_mnemonic = False
+
+
+MSG_ENTER_GAIT_MNEMONIC = _("Please enter your GreenAddress wallet mnemonic")
+MSG_ENTER_GAIT_2OF3_MNEMONIC = _("Please enter your GreenAddress 2of3 subaccount mnemonic")
+MSG_ENTER_GAIT_XPUB = _("Please enter your GreenAddress xpub key")
+
+
+class Plugin(BasePlugin):
+
+    def fullname(self):
+        return 'GreenAddress 2of3 recovery'
+
+    def description(self):
+        return 'Provides support for GreenAddress 2of3 subaccounts recovery'
+
+    def __init__(self, gui, name):
+        BasePlugin.__init__(self, gui, name)
+        self._is_available = loaded_mnemonic
+        self.wallet = None
+        electrum.wallet.wallet_types.append(('multisig_recovery', 'ga_2of3_recovery', _("GreenAddress 2of3 recovery"), GA2of3RecoveryWallet))
+
+    def is_available(self):
+        if self.wallet is None:
+            return True
+        if self.wallet.storage.get('wallet_type') == 'ga_2of3_recovery':
+            return True
+        return False
+
+    def set_enabled(self, enabled):
+        self.wallet.storage.put('use_' + self.name, enabled)
+
+    def is_enabled(self):
+        if not self.is_available():
+            return False
+
+        if not self.wallet or self.wallet.storage.get('wallet_type') == 'ga_2of3_recovery':
+            return True
+
+        return self.wallet.storage.get('use_' + self.name) is True
+
+    def enable(self):
+        return BasePlugin.enable(self)
+
+    @hook
+    def load_wallet(self, wallet):
+        self.wallet = wallet
+
+    def is_gait_mnemonic(self, text):
+        english = mnemonic.Mnemonic('english')
+        return len(text.split(' ')) == 24 and english.check(text)
+
+    def is_correct(self, mnemonic_gait, mnemonic_2of3, xpub_gait):
+        return Wallet.is_xpub(xpub_gait) and self.is_gait_mnemonic(mnemonic_gait) and self.is_gait_mnemonic(mnemonic_2of3)
+
+    def multi_seed_dialog(self, wizard):
+        vbox = QVBoxLayout()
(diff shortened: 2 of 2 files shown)


YuriRamos2:
Were this actually tested? I've been trying for the last two hours to get this to run, on ubuntu, mint and Windows. When it does, which is rare, the next button is greyed out when I fill the necessary info. This was the closest I got it to run, on Windows, it doesn't compile doing the same stuff I'm doing to sucessfully run the last electrum version, the error I get is "ImportError: No module named electrum".
I think the least you guys could do is to release a compiled windows version of this electrum version, there is absolutely no other tool to use 2of3 greenaddress wallet.

dabura667:
@YuriRamos2 Since this PR there has been a lot of changes on Master...

Perhaps try checking out the branch from greenaddress's fork?

YuriRamos2:
@dabura667 
The 2-of-3 plugin is not included in any of the forks as far as I know (I did look)

dabura667:
What I mean is to `cd` into a new directory that doesn't have an "electrum" folder, and use the following command:

`git clone -b greenaddress-2of3-recovery --single-branch https://github.com/greenaddress/electrum.git`

Then running electrum out of there.

They have similar files, but Thomas has changed a lot of things since September, and that may be affecting the performance if you are merely copying over just the plugin file or just cherrypicking the commits individually.

YuriRamos2:
Thanks for trying to help, but even after that it didn't doesn't work. Here is a print (the addresses are not holding any coins, so it's safe). It's the same error I got before.

oi

greenaddress:
@YuriRamos2 indeed this was tested (Linux and MacOsX) and by cloning the branch of our fork as per @dabura667 suggestion (thanks!)

We are working on improving the process and to provide instructions.

By the way, you need to make sure you don't use some preexisting electrum profile as it may break the specific electrum version as per issue encountered by one user here http://www.reddit.com/r/greenaddress/comments/2pp2p3/mnemonic_passphrase_for_greenaddress/cnc7zhc

greenaddress:
@YuriRamos2 the error you have seems to be due to Electrum's dependencies not being installed (in this case the mnemonic library)

Have you followed the Electrum instructions?

YuriRamos2:
At this point I think i followed every tutorial I could found about compiling electrum lol. 
Maybe I missed something, but how could electrum be installed with setup.py and now install the mnemonic library?

greenaddress:
@YuriRamos2 Don't use ubuntu or Mint but I presume you can follow the same instructions as Electrum 2.0 for support with Trezor https://www.reddit.com/r/TREZOR/comments/2jp9uk/tutorial_install_electrum_20_beta_with_trezor/

Specifically something like:

sudo apt-get install python-pip python-slowaes python-socksipy pyqt4-dev-tools
sudo pip install pyasn1 pyasn1-modules pbkdf2 tlslite qrcode
git clone git clone -b greenaddress-2of3-recovery --single-branch https://github.com/greenaddress/electrum.git
cd electrum
pyrcc4 icons.qrc -o gui/qt/icons_rc.py
python mki18n.py
sudo python setup.py install

YuriRamos2:
Followed to the letter the steps, adding before "sudo python setup.py install"
cd lib
sudo python mnemonic.py install
~/electrum

And I still get the same error as before.
I must also add that if I run "electrum" directly, to run the installed copy, the 2of3 option doesn't appear, I MUST run ./electrum to run from the folder. And if I run the installed copy the copy from the folder messes up (I get a seed without giving me any options and it closes with an error after the step that it asks for a password)

greenaddress:
presumably that's because you have two versions of electrum and that can lead to issues or confusion at the very least.

I'd try without having two versions and just use the one from the branch, otherwise you can always try to setup the pythonpath manually and run the electrum straight from the directory (but I assume that currently if you do that Electrum complains about mnemonic being missing)

YuriRamos2:
@greenaddress 
I really don't understand by what you mean I have two versions of electrum. Does running setup.py install a different version than running electrum directly?

Anyway, I just now managed to run electrum directly, on a fresh live session, without running setup.py. It gave the same error about mnemonic. I then installed mnemonic.py and I got the same stuff I get when I run setup.py.

I'm now going to try to run it in fedora, I doubt it will work, but I will give it a shot

YuriRamos2:
Same problem in fedora.

YuriRamos2:
I don't think this is relevant, but every time I used the "python mki18n.py" command I got this error, but I don't think getting the translation files are essential for electrum to run.

mint

greenaddress:
Apologies for the confusion and missing the obvious, the mnemonic module isn't an Electrum dependency. It's required by our plugin only. You need to manually `pip install mnemonic` first before trying it. We've just verified that it indeed works with newly created 2of3 account. Anyway we're going to rebase it to latest Electrum upstream sources and test again to make sure it's still appropriate for merging.

YuriRamos2:
Hey there.
This helped to get past that error, but now, there is a new one

new

greenaddress:
Seems like a bug in python-mnemonic's version newer than we used. Please try `pip install https://github.com/trezor/python-mnemonic/archive/d8328653f8247ee6b080870845e8efaadcd4984e.zip`

YuriRamos2:
Oh man, finally :) Thanks for all the help, I got it working now, I didn't use it for any transactions but it looks ok to me.

I think you guys should include a full guide on the github page of the project.

greenaddress:
Rebased and retested. Only changes requried were the addition of `window` argument in `enter_seed_box` in 256a467dd71d95efaf454bb00a5355f83159ed5d and change from 'electrum' seed introduced in da968b60e474a39bcea99c498010a478a7d75748 to standard BIP39 seed. Also cleaned up unused imports.

ecdsa:
I will not merge this plugin in the official repo, because it introduces support for non-electrum wallet types that I do not want to maintain. When we distribute software, we are commited to maintaining it in the long run, because that's what users expect us to do. That plugin would be very difficult to remove once merged, because some users might count on its continued presence. I will not merge a plugin if I feel that it cannot be removed without threatening the users bitcoins.

Feel free to distribute it on your own, though.

DarkCross71:
I don't know if I am in the right spot, but forgive me, I have trying for the last six hours trying to figure out my Mnemonic Passphrase. Is there any way to try and recover it? I wrote it down word for word I went trough the list of words and they are all there...am I screwed or is there a way to login so I can send btc? I can login, I have all the tabs ie:receive, transaction etc. I am missing the send one, and I have about 150.00 in there...

dabura667:
> I have trying for the last six hours trying to figure out my Mnemonic Passphrase. Is there any way to try and recover it? I wrote it down word for word I went trough the list of words and they are all there...

You are trying to figure out something you have written down? That makes no sense unless you can't read.

> am I screwed or is there a way to login so I can send btc? I can login...

1. Electrum does not have a "login" process. I think you are talking about GreenAddress, this is not GreenAddress's repository. Please bring up any issues with GreenAddress at their github: https://github.com/greenaddress/WalletCordova/issues
2. You are asking if there is a way to login, then say you can login... which makes no sense.

Please ask this question at GreenAddress, unless you are talking about an Electrum Mnemonic phrase, in which case none of your questions make sense, so please clarify.

DarkCross71:
Is sign in better for you? What's the difference?

dabura667:
Electrum does not have a sign in either. There is no difference between sign in and log in, but Electrum has neither.

GreenAddress, however, has a login process. Ask them.

PR #901, opened 2014-10-26
pass config.path to daemon subprocess

extempore:


ecdsa:
a related issue is that the daemon writes config properties that sometimes conflict with the gui settings

ecdsa:
I think the correct way to fix both issues is to send the config from the caller process to the daemon

ecdsa:
I merged it manually. note that the --portable option works.
be aware that you can have several daemons running on the same machine, if they do not use the same path

PR #928, opened 2014-11-19
Sort address, contact and invoice list

jennychang3:
Hello~
I add a feature for sorting address, contact and invoice list. Also, the sorting formate will remain same after restart electrum.
I think it would be much easier for users search their record.
Thanks.


ecdsa:
done in d94aee83b99ffe405a4875ccb215ab6c774c12e2

PR #933, opened 2014-11-22
On linux, give more accurate error messages so the correct packages can ...

jimmysong:
...be installed.

These are just some things I ran into while trying to install electrum on ubuntu.


Tafelpoot:
Pip is sufficient to get all the packages, except zbar (bindings only afaik) and it works cross-platform.
It would be nice to return the correct Pip commands to the user though... 
Note that slowaes requires the version number.

ecdsa:
superseeded by 16c72d286c119562a591dc06303dfeefa3ccfa01

Issue #935, opened 2014-11-24
dumpprivkeys() KeyError: '1'


xilanma:
Steps to reproduce:

1. Create new wallet
2. Go to the console tab and type: `dumpprivkeys()`

```python
>> 
Welcome to Electrum!
>> dumpprivkeys()
Traceback (most recent call last):
  File "/Users/xilanma/electrum/gui/qt/main_window.py", line 1693, in <lambda>
    return lambda *args: apply( f, (method, args, self.password_dialog ))
  File "/Users/xilanma/electrum/lib/commands.py", line 129, in _run
    result = f(*args)
  File "/Users/xilanma/electrum/lib/commands.py", line 205, in dumpprivkeys
    return [self.wallet.get_private_key(address, self.password) for address in addresses]
  File "/Users/xilanma/electrum/lib/wallet.py", line 355, in get_private_key
    return self.accounts[account_id].get_private_key(sequence, self, password)
KeyError: '1'
>> 
``` 

For some reason, this bug doesn't happen with my old wallet file, only if I create a new one.

ecdsa:
fixed in da968b60e474a39bcea99c498010a478a7d75748

Issue #884, opened 2014-10-13
Export Private keys bug


Tafelpoot:
I got this error when exporting my private keys in the Qt electrum version.
It is an xpub (seed v10) wallet with no password, containing 27 keys. It has received 1 single transaction.
UI gets stuck on the 26th key.

    Exception in thread Thread-22:
    Traceback (most recent call last):
      File "C:\WinPython-32bit-2.7.6.3\python-2.7.6\lib\threading.py", line 810, in
    __bootstrap_inner
        self.run()
      File "C:\WinPython-32bit-2.7.6.3\python-2.7.6\lib\threading.py", line 763, in
    run
        self.__target(*self.__args, **self.__kwargs)
      File "C:\Users\xxxxxxxxxxxxxx\Documents\GIT\electrum\gui\qt\main_window.py",
    line 2306, in privkeys_thread
        private_keys[addr] = "\n".join(self.wallet.get_private_key(addr, password))
      File "C:\Users\xxxxxxxxxxxxx\Documents\GIT\electrum\lib\wallet.py", line 359
    , in get_private_key
        return self.accounts[account_id].get_private_key(sequence, self, password)
    KeyError: '1'

ecdsa:
fixed in da968b60e474a39bcea99c498010a478a7d75748

PR #936, opened 2014-11-24
New dumpaddresses method to fix dumpprivkeys

xilanma:
See bug #935 and #884

Temporary fix. There is a probably a better way to do this but at least, it doesn't throw an exception anymore.


ecdsa:
obsoleted by da968b60e474a39bcea99c498010a478a7d75748

PR #937, opened 2014-11-25
Update util.py data directory for OpenBSD

bitcoindaddy:
The path "/Library/Application Support/Electrum" doesn't exist on OpenBSD. Searching for the wordlist failed when generating a wallet. Updated path for OpenBSD so it can find the file.


ecdsa:
not used anymore

PR #938, opened 2014-11-25
fix dumpprivkeys, wallet.get_private_key and export_privkeys_dialog

Tafelpoot:
Don't show private key for the next_account.
Wallet does not return the private key for PendingAccount, but an empty list instead.
It could be implemented, but I don't see the need.

See https://github.com/spesmilo/electrum/issues/884 and https://github.com/spesmilo/electrum/issues/935.
Overrides https://github.com/spesmilo/electrum/pull/936

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -2335,6 +2335,9 @@ class ElectrumWindow(QMainWindow):
 
         private_keys = {}
         addresses = self.wallet.addresses(True)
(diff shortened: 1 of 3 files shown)


Tafelpoot:
Better solution is not to include pending address by default. 
It is only required in the sync process... just add it manually in there.

ecdsa:
obsoleted by da968b60e474a39bcea99c498010a478a7d75748

PR #939, opened 2014-11-28
Fix Daemon on Windows

Tafelpoot:
- added localhost
- sys.executable to get around python2 / python3 issues
- close_fds cannot be combined with stderr and stdout on windows. see python doc.


ecdsa:
I merged it, but I don't know how it will behave in windows without close_fds.
is this understood?

Tafelpoot:
I saw this syntax is used in other projects as well. I'll look out for issues.

ecdsa:
on linux, without close_fds, the daemon fails to detect closed sessions, and as a result it does not shut itself down

PR #943, opened 2014-11-29
Add a transaction reminder plugin

changyc:
Hello, I add a plugin for scheduling a transaction. With this plugins, user can select when to execute the transaction. Currently, when the time is up, it will automatically remind you to complete the transaction. If the scheduling transaction has expired, it will remind you when you open Electrum next time.
Give me some feedback if you think this function will be useful with some adjustments. Thanks.


ecdsa:
this is a bit redundant with the payment protocol (bip70).
I would prefer these reminders to be linked to the payment protocol.

PR #945, opened 2014-11-30
Plugin providing optional support for key file 2FA

neocogent:
Adds functionality similar to Keepassx.

If plugin enabled, adds a "key icon" button to password dialogs allowing for user to choose a "key" file as 2nd factor. Tested quite a bit with new wallet, change password, show seed and export private keys dialogs. Not sure if there is another usage to test. 

It simply sha256 hashes the selected file and appends to password as hex string. So the key file is needed in addition to password for protected functions. If user ignores key button then no change should result (tested).

Hope someone else will do more testing.

Also, I am thinking of adapting Keepassx code for Yubikey HMAC challenge-response support into an Electrum plugin. Would appreciate comments on if this would be useful to the community. Should I open an issue as feature request to gauage demand?

### DIFF
--- a/gui/qt/password_dialog.py
+++ b/gui/qt/password_dialog.py
@@ -23,7 +23,7 @@ from util import *
 import re
 import math
 
-
+from electrum.plugins import run_hook
 
 def make_password_dialog(self, wallet, msg, new_pass=True):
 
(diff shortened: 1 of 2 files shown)


bobberb:
@tkhaew I would give any reason to use my YubiKey.

PR #975, opened 2015-01-09
Added ismine for validateaddress

shanew:
Output ismine field in validateaddress for our own addresses. Bitcoin core does this and is useful for identifying change addresses in a decoded transaction


ecdsa:
I'm not sure if it's good to make this command require a wallet.
maybe we could have a "ismine" command instead?

PR #976, opened 2015-01-11
Move root_name to parent class

clehner:
I was having trouble creating or using a Trezor wallet, with this error:

```
  File "/usr/local/src/electrum/gui/qt/main_window.py", line 309, in new_wallet
    wallet = wizard.run('new')
  File "/usr/local/src/electrum/gui/qt/installwizard.py", line 392, in run
    wallet.create_main_account(password)
  File "/usr/local/src/electrum/plugins/trezor.py", line 222, in create_main_account
    self.create_account('Main account', None) #name, empty password
  File "/usr/local/src/electrum/lib/wallet.py", line 1420, in create_account
    account_id, xpub, addr = self.get_next_account(password)
  File "/usr/local/src/electrum/lib/wallet.py", line 1403, in get_next_account
    derivation = self.root_name + "%d'"%int(account_id)
AttributeError: 'TrezorWallet' object has no attribute 'root_name'
```

When I gave the TrezorWallet class the attribute 'root_name' as the other wallets have, it worked. So this patch moves the definition `root_name = 'x/'` to the BIP32_Wallet class, from which the other classes that use root_name inherit, to make sure that all subclasses have the attribute defined.

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -1266,6 +1266,7 @@ class Deterministic_Wallet(Abstract_Wallet):
 
 class BIP32_Wallet(Deterministic_Wallet):
     # abstract class, bip32 logic
+    root_name = 'x/'
     gap_limit = 20
 
     def __init__(self, storage):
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks.

PR #982, opened 2015-01-18
Add arbitrary non-standard outputs

dabura667:
Seeing as 0.10 will loosen restrictions on non-standard transactions, this might be a good step towards allowing more Scripting functionality.

Todo list:
1. A method for redeeming outputs that are non-standard.
2. Perhaps the transaction dialogue should attempt to decode the hex into op codes and display them in the transaction summary?

This may be out of the focus for Electrum though... so "don't bother" also seems like a valid choice... (which is why I just left it at creating arbitrary outputs for now.)

### DIFF
--- a/gui/qt/paytoedit.py
+++ b/gui/qt/paytoedit.py
@@ -73,9 +73,15 @@ class PayToEdit(ScanQRTextEdit):
             amount = 0
         else:
             x, y = line.split(',')
-            _type = 'address'
-            address = self.parse_address(x)
-            amount = self.parse_amount(y)
+            n = re.match('^CUSTOM_OUT\s+([0-9a-fA-F]+)$', x.strip())
+            if n:
+                _type = 'custom'
+                address = n.group(1).decode('hex')
+                amount = self.parse_amount(y)
+            else:
+                _type = 'address'
+                address = self.parse_address(x)
+                amount = self.parse_amount(y)
         return _type, address, amount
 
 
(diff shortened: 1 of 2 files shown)


ecdsa:
also, luke-jr suggested to replace OP_RETURN completely

dabura667:
Yeah, I guess you could just do '6a...' and have the same effect.

Maybe we should remove OP_RETURN and just go with this instead.

What do you think of Todo: 1 and 2?

ecdsa:
yes, I think we should replace the current OP_RETURN with a 'script' type.
todo 2 seems feasible, but we need to write a real script parser
todo 1 seems more difficult in the short term, because it requires changes in Electrum servers 
(we need to know the coins available on arbitrary outputs)

dabura667:
Todo 1: I think it would also be difficult as we can know that if the output has our pubkey hash that we can spend it... but there is the possibility that the user might think they know the way to release the funds, but there is no way to check without revealing the "private key" (or whatever is needed)

Todo 1 might be better to create a full featured ScriptSig builder module for the client that could, for instance, insert sigs if they are needed, check the validity of the script locally (without leaking the information to the server.... thinking maybe something like this:

http://webbtc.com/script/a531df9b125ed8594c3d033e8560685ef5fde86a3fd5be2f1681b7ce7eb85c4a:0

)

But like I said... the scope on how full featured the Script component should be is, I think, an issue of focus. It seems like Electrum is gaining wider usage from more beginner users... so the question might be "who are we making this feature for? How will they use it? and What benefit will they gain from being able to do it within the Electrum UI?"

ecdsa:
follow up: 8863b6f294e352ada2298cf2d7a5ee3012c43b7d

ecdsa:
We are targetting all types of users. 
The only constraint is that advanced features should not be visible to users who do not need them.

Issue #1011, opened 2015-02-06
Enhancement: Restore last open wallet on startup


themighty1:
It is annoying to have to go to File->Open every time electrum starts in order to select my last open wallet.

MuhammedZakir:
I like to get this feature too. It would be a good option if we have multiple wallets.

@ecdsa Anything you have to say about this?

ecdsa:
this sounds reasonable

PR #1018, opened 2015-02-09
solution for #1011 - open last used wallet on startup

kevcooper:
A potential solution to #1011 - set the key 'last_wallet' in the config file when opening a wallet from the File > Open menu. When opening up electrum after being closed, try to read 'last_wallet' from config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ locale/
 .devlocaltmp/
 *_trial_temp
 packages
+env/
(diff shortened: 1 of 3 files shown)


ecdsa:
it looks like it will affect the command line, so "electrum getbalance" will return different results depending on which file you just opened.

kevcooper:
took out the changes to wallet.py and used the __init__.py file in qt to open the last used wallet... would this leave the command line functions working properly?

PR #1020, opened 2015-02-10
Uses PyCrypto when it is available, otherwise fall back on PBKDF2

gyger:
Uses PBKDF2 only if PyCrypto is not available. A similiar selection is already done in the PBKDF2 library.

### DIFF
--- a/plugins/trezor.py
+++ b/plugins/trezor.py
@@ -230,10 +230,22 @@ class TrezorWallet(BIP32_HD_Wallet):
 
     def mnemonic_to_seed(self, mnemonic, passphrase):
         # trezor uses bip39
-        import pbkdf2, hashlib, hmac
+        import hashlib, hmac
         PBKDF2_ROUNDS = 2048
         mnemonic = ' '.join(mnemonic.split())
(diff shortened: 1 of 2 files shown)


plugins/trezor.py
```diff
@@ -230,10 +230,22 @@ def create_main_account(self, password):
...
     def mnemonic_to_seed(self, mnemonic, passphrase):
         # trezor uses bip39
-        import pbkdf2, hashlib, hmac
+        import hashlib, hmac
         PBKDF2_ROUNDS = 2048
         mnemonic = ' '.join(mnemonic.split())
-        return pbkdf2.PBKDF2(mnemonic, 'mnemonic' + passphrase, iterations = PBKDF2_ROUNDS, macmodule = hmac, digestmodule = hashlib.sha512).read(64)
+
+        try:
+            from Crypto.Protocol import PBKDF2
```
dabura667: Shouldn't this be `from Crypto.Protocol.KDF import PBKDF2`?

gyger: Blame on me, it should be. But I don't have a Trezor to test it, so It got unnoticed by me.

ecdsa:
I don't really see the point of doing that. If you want to get rid of the pbkdf2 dependency, it seems better to rewrite the derivation code in Electrum, instead of just anning yet another dependency..

PR #1028, opened 2015-02-14
show error message alert if on UI thread

m0mchil:
Checks if current thread is the UI thread and shows a message dialog with the message. Fixes missing notification for various conditions since 0ce463c and 410ac82

### DIFF
--- a/plugins/trezor.py
+++ b/plugins/trezor.py
@@ -33,7 +33,9 @@ def log(msg):
     stderr.flush()
 
 def give_error(message):
-    if QThread.currentThread() == QtCore.QCoreApplication.instance().thread():
+    current_thread = QThread.currentThread()
+    app_instance = QtCore.QCoreApplication.instance()
+    if app_instance and (current_thread == app_instance.thread()):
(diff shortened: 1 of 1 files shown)


ecdsa:
this looks dangerous.. what are you trying to fix?

m0mchil:
In 0ce463c there was introduced a give_error() function showing a message box with any error within Trezor plugin. Then, in 410ac82 the call to QMessageBox was, correctly, removed to avoid segmentation fault if he function is called on a non-GUI thread.

With this commit a test is introduced if current thread is indeed the GUI thread and, if so, a QMessageBox is shown.

I tested it on both GUI and non-GUI threads and it seems fine, perhaps more tests are needed. Ideas are welcome.

m0mchil:
In current code, without the proposed change, there are messages from the plugin which never reach the user, for example on line 220, 'Outdated Trezor firmware...'

ecdsa:
I think it's a bad design; it would be better to call Qt message boxes explicitly in Qt methods.

PR #1037, opened 2015-02-18
Restore showing of message box error messages

m0mchil:
Now give_error() is method of TrezorWallet. Same approach as in the BTChipWallet plugin.


ecdsa:
note: BTChipWallet plugin is poorly written and should not be taken as example

PR #1038, opened 2015-02-18
Do not exit when the New Wallet dialog is canceled

pooler:
This patch fixes an issue where the Qt interface would shut down if the user cancels the creation of a new wallet.

Steps to reproduce issue (tested at 5f42573):
* Open a wallet in the Qt interface;
* Select New/Restore from the File menu;
* Enter a file name and hit OK;
* Cancel the wallet type selection dialog.


ecdsa:
fixed in 2a60828

PR #1044, opened 2015-02-23
New firmware integration

ubinity:
Prepare upcoming firmware integration


btchip:
Done from my wrong github account, but you get the point :)

PR #1062, opened 2015-03-05
Option to tip the server in a transaction

henrikssn:
This adds a new field in preferences which tips the server on every transaction.

Default server fee is 0. Server address is parsed from the welcome banner. If parsing fails then no tip is added to the transaction.

[See discussion on bitcointalk](https://bitcointalk.org/index.php?topic=975796.0)


ecdsa:
1. you don't need to parse the banner, there is already a 'server.donation_address' request in the protocol. 
2. this should be a plugin
3. you should check how it interacts with other plugins that add a fee to the transaction (trustedcoin.py)
4. it should not tip silently, tipping should be visible to the user.
5. in order to save miner fees and avoid dust, better donate a larger amount every 10 transaction

PR #1087, opened 2015-03-16
Let wallet recovery use 64 character hex strings and legacy 24 word seeds.

railgun74:
1. Allow wallet recovery from 64 character hex strings.
2. Allow wallet recovery from 24 word legacy seeds.

In Electrum 1.9.x, one could input 64 character hex strings and have using the method described in https://bitcointalk.org/index.php?topic=153990.0 . Specifically any procedure e.g the section titled "Custom seeds" at https://bitcoinspakistan.com/blog/electrum-seed-explained/ .

This results in a wallet that has an associated 24 word seed. This seed can be re entered into 1.9.x and the wallet successfully recovered. However in 2.x this no longer works and breaks compatibility with such wallets.

I propose this simple fix that restores backward compatibility with legacy 24 word seeds and 64 character hex strings.

### DIFF
--- a/contrib/make_download
+++ b/contrib/make_download
@@ -28,13 +28,21 @@ for k, n in files.items():
     link = "https://download.electrum.org/%s"%n
     if not os.path.exists(path):
         os.system("wget -q %s -O %s" % (link, path))
-    if os.path.getsize(path):
-        md5 = hashlib.md5(file(path,'r').read()).digest().encode('hex')
-        string = string.replace("##link_%s##"%k, link)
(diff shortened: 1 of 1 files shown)


ecdsa:
can you do that in 1 commit instead of 4?

railgun74:
Sure.

railgun74:
Fixed.

I wasn't able to reopen this pull request after fixing the commits. Created a new one at https://github.com/spesmilo/electrum/pull/1095

Thanks.

PR #1097, opened 2015-03-20
Support for restoring multisig wallets from BIP32 private keys

andyichr:
I have added support for restoring BIP32 private keys (xprv) from the multisig 2of2 and 2of3 wallet restore screens. This enables an important use case where one wants to create a multisig wallet capable of signing transactions using xprv keys that they already have, which may have been generated using external programs. Further, it makes the multisig wallet restore screen consistent with the regular wallet restore, which already does accept xprv keys.


ecdsa:
note: that part of the code should be made more generic, in order to allow arbitrary m-of-n wallets

Issue #955, opened 2014-12-10
merchant.py list index out of range


mauritso1:
```python
electrum/scripts/merchant$ python merchant.py request 0.1 2 120
Traceback (most recent call last):
  File "merchant.py", line 255, in <module>
    ret = send_command(cmd, params)
  File "merchant.py", line 165, in send_command
    out = f(*params)
  File "/usr/local/lib/python2.7/dist-packages/jsonrpclib/jsonrpc.py", line 274, in __call__
    return self.__send(self.__name, args)
  File "/usr/local/lib/python2.7/dist-packages/jsonrpclib/jsonrpc.py", line 225, in _request
    check_for_errors(response)
  File "/usr/local/lib/python2.7/dist-packages/jsonrpclib/jsonrpc.py", line 529, in check_for_errors
    raise ProtocolError((code, message))
jsonrpclib.jsonrpc.ProtocolError: (-32603, u'Server error:   File "/usr/local/lib/python2.7/dist-packages/Electrum-2.0-py2.7.egg/electrum/account.py", line 43, in get_address | IndexError: list index out of range')
```
```
starting daemon (PID 1509)
num rows 0
127.0.0.1 - - [10/Dec/2014 22:24:36] "POST / HTTP/1.1" 200 -
127.0.0.1 - - [10/Dec/2014 22:24:47] "POST / HTTP/1.1" 200 -
127.0.0.1 - - [10/Dec/2014 22:25:40] "POST / HTTP/1.1" 200 -
```

I tried it with 1.9.8 and got to the `KeyError: \'0\'')`, I tried it with 2.0 (github master branch), generating wallets from 2.0 GUI and putting them in the right folder etc. 

```
out = f(*params)
*** ProtocolError: (-32603, u'Server error:   File "/usr/local/lib/python2.7/dist-packages/Electrum-2.0-py2.7.egg/electrum/account.py", line 43, in get_address | IndexError: list index out of range')
params
['0.1', '2', '120', 'password']
(Pdb) num
0
```

It seems like it returns an empty list, because `num` was `0` at the time (PDB)

Is this a bug/did I miss something/did I do something stupid? If I don't know the answer to this question should I be adding an issue here at all? 

```python
def process_request(amount, confirmations, expires_in, password):
    global num
    if password != my_password:
        return "wrong password"

    try:
        amount = float(amount)
        confirmations = int(confirmations)
        expires_in = float(expires_in)
    except Exception:
        return "incorrect parameters"

[...]

PR #1123, opened 2015-03-31
patch for issue #955

Minthos:
Don't know why the line is there. Removing it seems to solve the problem.

### DIFF
--- a/scripts/merchant/merchant.py
+++ b/scripts/merchant/merchant.py
@@ -277,7 +277,6 @@ if __name__ == '__main__':
     else:
         wallet = electrum.wallet.Wallet(storage)
 
-    wallet.synchronize = lambda: None # prevent address creation by the wallet
     wallet.start_threads(network)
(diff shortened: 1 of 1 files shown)


Minthos:
Testing shows that this fix only delays the problem.

ecdsa:
that line is there because we don't want to create a wallet with watching thousands of addresses.
we only want to watch the addresses where we expect a payment.
to fix issue #955 correctly, we need to derive the address instead of calling wallet.get_address
I guess the semantics of wallet.get_address changed, and that caused the issue.

Minthos:
Is there a function other than account.create_new_address(0) that will do what we want? If not, that seems to be the easiest solution.

ecdsa:
I fixed #955 here: fb699e5294a48f12d32a5a9fb5eb5e847b10bbeb

Minthos:
Ah, sweet.

PR #1135, opened 2015-04-08
Add wallet function get_addr_received()

Kausheel:
get_addr_received() allows the user to see the total amount of coins an address has ever received.

Use case: a merchant wants to view the total amount of coins a customer has ever sent to an address. If the merchant regularly sweeps coins from the address, the standard balance query using UTXO doesn't help.

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -487,6 +487,16 @@ class Abstract_Wallet(object):
                 coins.pop(txi)
         return coins.items()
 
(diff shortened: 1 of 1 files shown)


ecdsa:
this will report incorrect amounts if the electrum servers have pruned the history

Kausheel:
Fair enough. I imagine the code will still be useful for someone connecting to their own Electrum server and disabling pruning.

PR #1138, opened 2015-04-10
New LabelSync

maran:
This LabelSync is much faster because it will only request labels that changed since the last sync. It is also using a new back-end and no longer requires any registration.

Server code is here: https://github.com/maran/electrum-sync-server

### DIFF
--- a/plugins/electrum_sync.py
+++ b/plugins/electrum_sync.py
@@ -1,0 +1,204 @@
+from electrum.util import print_error
+
+import socket
+import requests
+import threading
+import hashlib
+import json
+
+try:
+    import PyQt4
+except Exception:
+    sys.exit("Error: Could not import PyQt4 on Linux systems, you may try 'sudo apt-get install python-qt4'")
+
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks! why is it a new plugin? 
it will make things complicated if users have to choose between two plugins.
also, the commit history of the old plugin will be lost if we delete it.

maran:
It's a new plugin because there is zero compatibility with the old system. I was going to put a message in the old plugin telling people to switch with a dead line and then remove the old. 

Perhaps there is a way I could code up a migration feature that would detect the old format somehow and automate the upload to the new system?

ecdsa:
can't you import the existing encrypted labels in the new database?

maran:
That might be easier.... I did not really consider that. I will have to figure out how to convert postsql to sqlite but it should be doable. I will open a new pull request where I change the old plugin instead of adding a new one once I figured everything out.

ecdsa:
ok, thanks!

PR #1141, opened 2015-04-11
Dust threshold should be 546

kaykurokawa:
Dust threshold should be 546, as is standard in Bitcoin Core  

https://github.com/bitcoin/bitcoin/blob/df5c246ba38c5b5bdd6192b0201d2fe8245451ca/src/primitives/transaction.h#L138

### DIFF
--- a/lib/bitcoin.py
+++ b/lib/bitcoin.py
@@ -31,7 +31,7 @@ import aes
 
 ################################## transactions
 
-DUST_THRESHOLD = 5430
+DUST_THRESHOLD = 546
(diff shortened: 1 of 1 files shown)


ecdsa:
ok, I merged it, but it will cause bugs with the servers that run an old bitcoind

kaykurokawa:
I see, the servers not going to relay it right because its non standard? Is there some mechanism in place to detect version numbers on servers? 

Also what about RECOMMENDED_FEE, should we consider changing it?

Issue #1006, opened 2015-02-04
OSX .dmg - new wallet wizard issues


usezfsyo:
Selecting File > New/Restore displays a window asking for the new wallet name, but then doesn't show the new wallet creation wizard.  Main window loses focus, but no new window pops up.  If I select File > New/Restore a second time, both the new wallet name window and the new wallet creation wizard pop up at the same time.
![screen shot 2015-02-03 at 11 24 20 pm](https://cloud.githubusercontent.com/assets/6395016/6035162/10cc43a6-abfc-11e4-80f5-1cf2faf122a4.png)

ecdsa:
might be related to set_layout in the wizard

ecdsa:
does it occur with an install from source on osx?

maran:
I haven't been able to replicate this on Yosemite 10.10.1 (14B25) neither from the DMG or from source.

Would it possible you give me some more information on your system? The information can be found at "About this Mac"

usezfsyo:
I'm using Yosemite 10.10.2 (14C109). Python 2.7.9.

When I run from source, everything works as expected.  I have to be in the electrum source directory, though.  I have other issues if I try to run electrum from anywhere else.

maran:
I will try to update to 10.10.2 and see if that makes me replicate it.

ecdsa:
any news on this?

ot4me:
my electrum (installed from dmg) "hangs pretty much every time" I use the new wallet wizard. right after it asks for the new name it stops responding. dirty workaround is to kill the app, delete the wallet directory and start over.

ecdsa:
@maran did you update to his version?

maran:
@ecdsa Yes I did, and I'm able to replicate the behaviour now. The first time you bring up the wallet wizard you need to click the Electrum icon in the taskbar in order to show it. Tries after that work fine. I'm not sure how to continue debugging this but I will try.

ecdsa:
ok, maybe we just need to add a show() somewhere

ecdsa:
try last commit: c56fe45ad7b93cdb63b6dbb27f72d97c9e789258

maran:
Still happening, the window is showing, as in it is visible. It just 'spawns' behind the main window and won't come into focus.

ecdsa:
added this: 6ececd500d78375862391115f4ecb69b6a53e688

PR #1142, opened 2015-04-11
delete -psn arg on osx, fixes #1080

### LINKED ISSUE #1080: Proxy setup is greyed out in 2.02/Mac 10.10.2
I tried to run latest Electrum in my mac, but unable to connect to server, previous version worked fine.

in network panel, there's a proxy setup but greyed out, I already removed all my proxy settings in system.

here's the screenshot
https://www.amazon.com/clouddrive/share/I_ZWR9U9X0AyeL4ldlJ16A7LVXh-_-QqGnByIrpBFTc

ptrcarta:


maran:
Good find! I will verify it when I'm on my macbook tomorrow morning.

ecdsa:
why is emulate_argv=False needed?
I understand it fixes #1006, but do we understand why it fixes it?

maran:
No we don't. I tried applying this patch but binary building is currently broken again with again the freaking CA_BUNDLE problem. `assert os.path.exists(requests.utils.DEFAULT_CA_BUNDLE_PATH)`

I don't have time to go into it right now I will check it again when I'm back from my holiday.

ecdsa:
merging for now, but it would be better to find a proper fix

Issue #1157, opened 2015-04-26
View Address on Block Explorer


ThePiGuy:
Currently, you can right-click on a transaction to view it in your default block explorer.  However, can we implement a right-click feature on the "Addresses" tab so that you can view the address on your default block explorer (to see facts like total received, and a more comprehensive breakdown of transactions [I know that the details was just implemented, but it doesn't give a "total received" amount]).  Thanks!

kyuupichan:
https://github.com/spesmilo/electrum/pull/1159

PR #1159, opened 2015-04-26
Clean up block explorer handling.  Add menu item to go to block explorer...

kyuupichan:
... for an address.

Block explorer code is data-driven now.
Put block explorer defaulting in one place.
Fix URLs for insight.is and blockr.io.
Add tradeblock.com explorer.
Add menu item to view address on block explorer provided only one is selected.

This patch also resolves https://github.com/spesmilo/electrum/issues/1157


ecdsa:
thanks

PR #1168, opened 2015-04-28
Show wallet basename in tray tooltip

kyuupichan:
For those of us using multiple electrum instances simultaneously it's nice to know which electrum icon is for which wallet.


ecdsa:
that will crash with watching only. 
it would be better not to call i18n._ from wallet.py

ecdsa:
note: _() is called in restore() but that's not an example to follow..
I rather move restore away from this class

Issue #1048, opened 2015-02-26
Code reformatting/optimziation


andscho:
In the past few days I spent quite a bit of time going through Electrum's code and noticed semi-often minor non-critical formatting issues.
I know that might seem nit-picky but Python has this nice property that it basically forces everyone to use the same coding style (as defined in numerous PEPs).

E.g. there are many instances of "( something )" instead of "(something)" or too many or too few empty lines.
IMO, this should be fixed before releasing Electrum 2.0 this week because it gives a much better impression to the trained eye of a programmer.

I am using PyCharm to look at the code. It shows you these formatting issues and even offers to reformat everything autmagically (Keyboard shortcut Ctrl-Alt-L).

What do you think?

ecdsa:
ok, but not before 2.0. 
Several people have provided reformatting PRs that ended up breaking things..

andscho:
OK, cool, I understand that code formatting is not an ultra high priority :)

Have you thought about giving Pycharm and its refactoring/reformatting capabilities a try yourself? This way you could see the result immediately and can confirm that no things are broken.

andscho:
Screenshot for electrum2b3

![image](https://cloud.githubusercontent.com/assets/3459903/6423581/955d9bca-bee8-11e4-894f-c30e36de0718.png)

Tafelpoot:
heh.. I'm also in favor of growing towards PEP8... just be careful not to break anything.

ecdsa:
closing, the OP did not follow up

PR #1169, opened 2015-04-28
Fix PEP8 warnings at lib/tests

romanz:
I would be happy to help making Electrum source code PEP8-compatible, starting from the tests.
I am trying to do the minimal amount of changes to make it pass PEP8 checker (#1048).


romanz:
If it's OK, I would be happy to help to convert the rest of the code to PEP8.

ecdsa:
I agree with most of the code formatting in PEP8, but breaking xpubs just to have shorter lines, seriously?

ecdsa:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

romanz:
100%, will fix.

romanz:
I've added `# nopep8` remark at the end of lines containing xpub/xprv data.

romanz:
BTW, is this PR still relevant?
If not, we can close it...

PR #1170, opened 2015-04-28
Tray tip wallet

kyuupichan:
Well spotted Thomas.  How about this, keeping the localization bit in main_window.py?

Tested on watch-only and not watch-only wallet, after confirming your predicted failure with a watch-only wallet.


ecdsa:
why define a title() function, if it is called in only once? also, why define two basename functions ?
it looks like you took your previous PR and made minimal changes to it, instead of making minimal changes to the code itself.

kyuupichan:
ecdsa - my goal with all my patches isn't minimal changes.  It's logical clarity, which often amounts to the same thing.  But if you prefer that I'm happy to oblige.

Issue #302, opened 2013-09-18
Thousand Separators in Electrum


shawnpringle:
It is nice to have thousand separators in Electrum.  Please see my branch (fork) for a working version with this feature.  It works for the QT version which is what i use.

PR #1171, opened 2015-04-29
Modify format_satoshis to display amounts according to locale.

kyuupichan:
In particular, thousand separators ('.' or ',' or nothing) and decimal point separators are taken from the locale.

This also resolves issue 302: https://github.com/spesmilo/electrum/issues/302.

### DIFF
--- a/gui/qt/history_widget_lite.py
+++ b/gui/qt/history_widget_lite.py
@@ -20,6 +20,6 @@ class HistoryWidget(QTreeWidget):
         if date is None:
             date = _("Unknown")
         item = QTreeWidgetItem([amount, address, date])
-        if float(amount) < 0:
+        if amount.find('-') != -1:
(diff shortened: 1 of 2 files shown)


shawnpringle:
This doesn't solve issue 302:   After the decimal place no thousand separators are placed.  When you type in an amount the commas are not put in for you either.  This is not a complete implementation.
That is 0.12345678 should be displayed as 0.123,456,78 but it isn't.  When typing into the send tab, the commas are not added.

shawnpringle:
That is to say, apart from the display being only half right.  When typing in the send amount in the send tab, commas are not added and adjusted when you type your number in.

shawnpringle:
Please see : http://canada.host-ed.me/.

kyuupichan:
This was attempting to affect formatted output.  I disagree that input should be formatted, it would annoy the heck out of me having the computer put random commas or dots in my input fields.

PR #1181, opened 2015-05-02
Unconfirmed balance change fix

kyuupichan:
Currently if you have a balance of 3 BTC, and spend a 1 BTC output to send 0.3 BTC with change of 0.7 BTC, electrum shows your balance as 2 BTC with 0.7 unconfirmed.  This patch is to revert to the prior behaviour of showing a 3 BTC balance with -0.3 unconfirmed.

I originally tried to do this by modifying get_addr_utxo to take an extra argument indicating if this new behaviour or the prior behaviour is desired by the caller.  The problem is there is no good txn:index key to insert the negative amounts in the coins dictionary (we don't know the index), and I'd rather keep the logic of the returned of get_addr_utxo clean.

So instead I put similar code inline in get_addr_balance, which doesn't care about txn keys, only about the (c,u) tuple result.

Remove get_addr_balance2 which is dead code.

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -462,7 +462,11 @@ class Abstract_Wallet(object):
                 fee = v_out - v_in
         return is_relevant, is_send, v, fee
 
-    def get_addr_utxo(self, address):
+    # Returns all the unspent TXOs of an address. Unconfirmed receipts are always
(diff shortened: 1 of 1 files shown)


ecdsa:
can you try this: http://pastebin.com/h8c8QWS2

Issue #1163, opened 2015-04-27
Newly mined coins should be shown as immature


roy7:
The checkmark goes green after 6 confirmations, but that's only accurate for normal transactions and not for newly mined coins which require 100. When a deposit to the wallet is made via newly mined coins, could the status symbol update based on 100 instead of 6, and the green checkmark not appear until the 100 confirmations are up?

Took me a while to figure out why I couldn't send coins this evening since I was mistakenly thinking the checkbox meant the funds were ready for spending.

PR #1182, opened 2015-05-02
Clean up get_history().

kyuupichan:
Make it 3-step instead of four.  Do less work in general.  Avoid
quadratic insertions.

This is preparation to hopefully fix https://github.com/spesmilo/electrum/issues/1163


ecdsa:
nice!
two issues: 
 - use int not float
 - get_tx_delta might return None

PR #1183, opened 2015-05-02
Make it 3-step instead of four. Avoid quadratic insertions.

### LINKED ISSUE #1163: Newly mined coins should be shown as immature

kyuupichan:
This is preparation to hopefully fix #1163

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -712,42 +712,38 @@ class Abstract_Wallet(object):
         if domain is None:
             domain = self.get_account_addresses(None)
 
-        hh = []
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks!

PR #1184, opened 2015-05-02
Show coinbases transactions maturing over the full 100 confirmations.

kyuupichan:
The first confirmation shows the single-confirmation timer.
It switches to the 2-confirmation timer at 20 confirms.
...
5-confirmation timer at 80 confirms and finally confirms at 100.

Tested on 1FadDQWq27kyoWkwWYmNBaZvN9qFz4Dsuj.

How is this Thomas?  Note I had to use self.transactions() as the is_cb field in the tuples is always false, even for coinbases.  Try on the address above.  I suspect this is a bug.

### DIFF
--- a/gui/qt/history_widget.py
+++ b/gui/qt/history_widget.py
@@ -23,7 +23,7 @@ from util import *
 from electrum.i18n import _
 from electrum.util import block_explorer_URL, format_satoshis, format_time
 from electrum.plugins import run_hook
-
+from electrum.bitcoin import COINBASE_MATURITY
 
 class HistoryWidget(MyTreeWidget):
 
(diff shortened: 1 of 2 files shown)


ecdsa:
not good... giving two different meanings to these icons is not a good idea imo.

kyuupichan:
What would you prefer then?

ecdsa:
Just like we currently display unconfirmed coins, in the status bar. (+ xx unmatured)

kyuupichan:
Ah I see.  I thought we'd want icons different too.

ecdsa:
fixing this, I just noticed that is_coinbase is not properly set in self.txo

ecdsa:
see 687cc77

PR #1187, opened 2015-05-03
Move away from requiring network and blockchain objects to be able to re...

kyuupichan:
...quest local height.

We store it in the config object instead of in the blockchain object.
The blockchain object now refers to its config, and calls refresh_height() to update it.
The network objects also refer to the config rather than the blockchain.

This is the first of many small steps to untangle the verifier from stored state and so
permit the history tab to work in offline mode.  The refactoring will simultaneously clean
up a lot of accumulated cruft.

### DIFF
--- a/lib/blockchain.py
+++ b/lib/blockchain.py
@@ -30,18 +30,12 @@ class Blockchain(util.DaemonThread):
         self.config = config
         self.network = network
         self.lock = threading.Lock()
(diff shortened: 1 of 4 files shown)


ecdsa:
I agree with the general idea, but config should not be used for that.

ecdsa:
I merged it anyway. will try to find a better solution later

PR #1192, opened 2015-05-05
Pass the response_queue to the constructor, not start().

kyuupichan:
Removes an unnecessary Thread base-class override.  The python
documentation also strongly discourages overriding anything other
than run().


tamina-tam88:
kyuupichan:if-cleanup3 from 6 hours ago

PR #1194, opened 2015-05-05
Send requests only from the interface thread.

kyuupichan:
Currently requests are sent from the requestor's thread.  The lock is
not properly held where necessary so this is not thread-safe.  For example
it can race with the thread stopping and closing the socket the
requestor is trying to use to send with.

Resolve such races by having send_request() simply queue the requests,
which are asynchronously sent from the interface thread itself.


ecdsa:
nice, thanks

PR #1197, opened 2015-05-06
Move response getting to its own function.

kyuupichan:
This makes the operation of run() more clear.
Also, store the request time with each sent request to more simply
catch request timeouts.
No essential change in functionality.


ecdsa:
that for loop might become expensive.. 
there might be thousands of unanswered requests

PR #1199, opened 2015-05-06
Fix two races in interface.py.

kyuupichan:
First, close the socket from the thread itself rather than from
the stop() function.  This prevents another thread closing the
socket that the interface thread is simultaneously using.

Second, it was possible for the parent thread such as network.py
to start() an interface, and ether immediately or after a timeout
on a send_request() to test is_connected() and assume the connection
has failed when the thread hasn't even been scheduled or gotten
around to completing the socket connection.

The flow of control in interface.py is now reasonably clear.


ecdsa:
this needs to be rebased

PR #1208, opened 2015-05-08
Attempt to solve the config issue after opening a new wallet.

kyuupichan:


ecdsa:
this seems overly complex; I rather move height computation back into blockchain

ecdsa:
fixed in 233fd8e
the last known height is stored in the wallet file for offline mode

PR #1210, opened 2015-05-10
gui/text.py curses GUI bug fixes

venzen:
The current master branch version of the curses GUI exits on an exception when the user selects *Network* or *Settings* dialogs via `Ctrl+N` and `Ctrl+S`.

The reason is because the `self.network` object does not have the attribute `default_server` still being referred to by the **gui/text.py** code. Similarly, the code tries to access the `self.wallet.fee` attribute which, in the current master, is not assigned. `fee_per_kb` is the correct attr name.

I've updated these attrs in **gui/text.py** and tested. The curses GUI works as expected and the dialogs correctly display config settings.

### DIFF
--- a/gui/text.py
+++ b/gui/text.py
@@ -353,7 +353,7 @@ class ElectrumGui:
     def network_dialog(self):
         if not self.network: return
         auto_connect = self.network.config.get('auto_cycle')
-        host, port, protocol = self.network.default_server.split(':')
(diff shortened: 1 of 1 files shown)


venzen:
I think the Travis build failed on the indentation error in that first commit (fixed in second commit).

ecdsa:
travis failed because tests are broken

PR #1211, opened 2015-05-13
Fix createrawtransaction and add missing deserialize() calls

mikeland86:
Found these issues while trying to create, sign, and broadcast a raw transaction. 

* createrawtransaction was using old signature for Transaction constructor
* Signwithwallet and decoderawtransaction need to call deserialize on tx before they can access inputs and outputs. (Maybe adding getInputs() and getOutputs() which deserializes if needed might be nicer)


ecdsa:
thanks

PR #1217, opened 2015-05-19
Attempt clean shutdown of threads

kyuupichan:
At present when electrum shuts down, I almost always get python exceptions from the exchange rate plugin, and occasionally from the interface threads.  This seems to eliminate those.

Network object to wait for interface threads to finish
Clean shutdown for exchanger plugin

### DIFF
--- a/electrum
+++ b/electrum
@@ -75,7 +75,7 @@ from electrum import util
 from electrum import SimpleConfig, Network, Wallet, WalletStorage, NetworkProxy, Commands, known_commands, pick_random_server
 from electrum.util import print_msg, print_error, print_stderr, print_json, set_verbosity, InvalidPassword
 from electrum.daemon import get_daemon
-from electrum.plugins import init_plugins
+from electrum.plugins import init_plugins, shutdown_plugins
 
 
 # get password routine
@@ -248,7 +248,7 @@ if __name__ == '__main__':
 
         gui = gui.ElectrumGui(config, network)
         gui.main(url)
-
+        shutdown_plugins()
         if network:
             network.stop()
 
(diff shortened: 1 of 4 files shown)


ecdsa:
No, it is the opposite.
interface.py currently uses daemon threads, that's why you see this error.
util.DaemonThread are threads that do not have the python daemon flag.

kyuupichan:
So what precisely is your suggestion that I can test locally?

ecdsa:
well, to replace daemon threads with that class

kyuupichan:
OK, I have done so.  I will submit a separate pull request.  But what does this achieve?  I see nothing magic in util.DaemonThread that prevents races.  Testing is_running() is no such guarantee as its return value is only good for the instant it was tested, and not the next line of python code.  Can you explain your idea / intent, as I seem to be misunderstanding it.

With the separate pull request using util.DaemonThread, for example, here is the shutdown error I got the first time I ran it:

Exception in thread Thread-16 (most likely raised during interpreter shutdown):
  File "/usr/local/lib/python2.7/threading.py", line 810, in __bootstrap_inner
  File "/home/neil/src/electrum/plugins/exchange_rate.py", line 102, in run
  File "/home/neil/src/electrum/plugins/exchange_rate.py", line 72, in update_rate
  File "/home/neil/src/electrum/lib/simple_config.py", line 109, in get
  File "/usr/local/lib/python2.7/threading.py", line 168, in acquire
<type 'exceptions.TypeError'>: 'NoneType' object is not callable

I don't understand your reluctance to properly join() the threads that have been shut down (and providing plugins with a finalization function is arguably an improved interface anyway).

kyuupichan:
Withdrawn for now

PR #1218, opened 2015-05-20
FX plugin and interfaces to inherit from DaemonThread

kyuupichan:
If I understood your suggestion correctly.  However this doesn't solve the problem I want to fix:

Exception in thread Thread-16 (most likely raised during interpreter shutdown):
File "/usr/local/lib/python2.7/threading.py", line 810, in __bootstrap_inner
File "/home/neil/src/electrum/plugins/exchange_rate.py", line 102, in run
File "/home/neil/src/electrum/plugins/exchange_rate.py", line 72, in update_rate
File "/home/neil/src/electrum/lib/simple_config.py", line 109, in get
File "/usr/local/lib/python2.7/threading.py", line 168, in acquire
: 'NoneType' object is not callable

### DIFF
--- a/lib/interface.py
+++ b/lib/interface.py
@@ -48,11 +48,10 @@ def Interface(server, response_queue, config = None):
     else:
         raise Exception('Unknown protocol: %s'%protocol)
 
-class TcpInterface(threading.Thread):
+class TcpInterface(util.DaemonThread):
 
     def __init__(self, server, response_queue, config = None):
-        threading.Thread.__init__(self)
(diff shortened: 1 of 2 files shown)


kyuupichan:
Apologies, I missed a setting of the daemon member variable.

That seems to help solve the issue.  However starting up and shutting down now causes electrum to not exit, so I cannot really recommend you merge this in its current form.

kyuupichan:
Withdrawn for now

PR #1232, opened 2015-05-22
switch_to_interface() to start interface if not connected

kyuupichan:
Two places currently have switch-to-or-start-interface logic.
Put that common logic into switch_to_interface().


kyuupichan:
Hold on, I missed one thing here...

kyuupichan:
Thomas - I've convinced myself this doesn't change semantics of disconnected_servers.  If it's OK with you please feel free to merge it.

PR #1237, opened 2015-05-23
Remove pending_servers.

kyuupichan:
Existing code has the concept of pending servers, where a connection
thread is started but has not sent a connection notification, and
and interfaces which have received the notification.

This separation caused a couple of minor bugs, and given the cleaner
semantics of unifying the two I don't think the separation is beneficial.
The bugs:

1) When stopping the network, we only stopped the connected interface
threads, not the pending ones.  This would leave Python hanging
on exit if we don't make them daemon threads.

2) start_interface() did not check pending servers before starting
a new thread.  Some of its callers did, but not all, so it was
possible to initiate two threads to one server and "lose" one thread.

Apart form fixing the above two issues, unification causes one more
change in semantics: we are now willing to switch to a connection
that is pending (we don't switch to failed interfaces). I don't
think that is a problem: if it times out we'll just switch
again when we receive the disconnect notification, and previously the
fact that an interface was in the interaces dictionary wasn't a
guarantee the connection was good anyway: we might not have processed
a pending disconnection notification.

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -153,7 +153,6 @@ class Network(util.DaemonThread):
 
         self.irc_servers = {} # returned by interface (list from irc)
         self.recent_servers = self.read_recent_servers()
-        self.pending_servers = set()
 
         self.banner = ''
         self.heights = {}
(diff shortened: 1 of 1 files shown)


kyuupichan:
This introduces a bug.  I prefer to resolve it by giving interface.py a tri-state connection_state flag.  I can work that in without changing semantics and it will make the transition to a unified interfaces dictionary more transparent.  I will follow up with a sequence of commits to get us there that way; it will be better for long-term clarity too.

PR #1249, opened 2015-05-25
Common stale interface testing and switching

kyuupichan:
Thomas - how about this?  It separates lagging and connected tests
which seemed to be your concern.  I'm not sure I prefer it because it
is more code and requires more care about where exactly lag can be
introduced.  Anyway I wanted your feedback.  Still testing locally.

A new function switch_stale_interface() switches servers, optionally
taking a suggestion to switch to, the previous strategy of
new_blockchain_height().  It checks the protocol of the suggestion
before switching.  The protocol might not match if the network has
just been restarted and the prior interfaces had queued a notification.

This is called from various places that check for staleness.  Socket
disconnection is exclusively checked by check_interfaces() in the
main loop.

A lagging server is checked where lag can be introduced, and
in set_parameters() where auto_connect being True may have changed
our ability to switch away.

The common switching logic uses the slightly more sophisticated logic
of check_interfaces() that also considers disconnected servers.

Lagging diagnotics moved to server_is_lagging().

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -114,7 +114,7 @@ def deserialize_proxy(s):
 def deserialize_server(server_str):
     host, port, protocol = str(server_str).split(':')
     assert protocol in 'st'
-    int(port)
+    int(port)    # Throw if cannot be converted to int
     return host, port, protocol
 
 def serialize_server(host, port, protocol):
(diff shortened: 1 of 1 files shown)


kyuupichan:
I'm starting to think I may prefer this one after all.  

I'm also starting to believe that the is_connected() test and stale switch in check_interfaces() is in the wrong place - all disconnected interfaces are guaranteed to send an interface notification.  We should be switching there instead.

ecdsa:
Sorry but this does not address my point; I guess I was not clear enough.
The part of the code that is about SERVER_RETRY_INTERVAL and disconnected_servers.remove() is socket maintenance and should not be called from the blockchain thread.

PR #1253, opened 2015-05-27
Speed up painting of qr codes.

kyuupichan:
Probably speeds it up by about a factor of two.
Unfortunately it needs to be another 5x faster
for sluggishness to disappear in the GUI when
typing a description in the receive tab.

Note the old code was off-by-one.


ecdsa:
thanks

Issue #1251, opened 2015-05-26
Something wrong with payment requests


nomnombtc:
Hello,

a few days ago I set firefox to handle bitcoin: requests with electrum,
so today I ordered a Pizza and click the link on bitpay checkout page.

Now electrum wants to send 90913 mBTC instead of 90,913 mBTC
I also tried with electrum set to BTC instead mBTC and it wanted to send 90913 BTC.

here is a screenshot: http://i.imgur.com/evvE7xw.jpg

Not sure what, but something seems to be wrong there.

Greetings

kyuupichan:
This will be fixed in the next release.  In the meantime you could set your locale to one that doesn't use a comma for a decimal point.

kyuupichan:
Actually this seems to be something different to what I thought.

kyuupichan:
Can you attach the payment request text?

ecdsa:
I can reproduce that bug. 
It occurs if the amount includes a decimal point.

kyuupichan:
See PR #1254 .  Does that fix it?

PR #1254, opened 2015-05-27
Fix nasty bug in amount editors

kyuupichan:


kyuupichan:
This might be a fix for #1251, I'm unsure.

To see the issue it's fixing, type this in the console: gui.fee_e.setAmount(96)

kyuupichan:
I'll resubmit with Decimal after testing.

PR #1268, opened 2015-05-31
Make the blockchain class not a thread

kyuupichan:
Remove interface communication out of blockchain.py
into network.py.  network.py handles protocol requests
for headers and chunks.  blockchain.py continues to
handle their analysis and verification.

If an interface provides a header chain that doesn't
connect, it is dismissed, as per a previous TODO comment.

This removes a thread and another source of timeouts.
I see no performance issues with this when truncating the
blockchain.

Rename 'result' to 'header' for clarity.

### DIFF
--- a/lib/blockchain.py
+++ b/lib/blockchain.py
@@ -17,70 +17,31 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 
-import threading, time, Queue, os, sys, shutil
-from util import user_dir, print_error
+import os
 import util
 from bitcoin import *
 
 
-class Blockchain(util.DaemonThread):
-
+class Blockchain():
+    '''Manages blockchain headers and their verification'''
     def __init__(self, config, network):
-        util.DaemonThread.__init__(self)
         self.config = config
         self.network = network
-        self.lock = threading.Lock()
         self.headers_url = 'http://headers.electrum.org/blockchain_headers'
-        self.queue = Queue.Queue()
         self.local_height = 0
         self.set_local_height()
+
+    def print_error(self, *msg):
+        util.print_error("[blockchain]", *msg)
 
     def height(self):
         return self.local_height
 
-    def run(self):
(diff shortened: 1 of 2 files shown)


ecdsa:
init_headers_file should not be in a constructor, because it might block for 30 seconds.
I'm afraid we need a thread for that.

kyuupichan:
It's called from when the network thread starts, so it would block the network thread.  Is that so bad given that this typically would happen when running the install wizard?

ecdsa:
yes, the Blockchain constructor is called in Network.__init__, which is called in NetworkProxy.__init__
so it would block the gui thread.
if you dig in the history, you'll see that init_headers used to be in the constructor, and I moved it to run() because it was actually blocking the client, so I'm not inventing this...

ecdsa:
Note: we can still get rif of the blockchain thread, if this http request is performed by Network

kyuupichan:
It is not called from Network.__init__(), but from Network.run(), i.e. it is part of the network thread not the object owner's thread.

ecdsa:
merged, but it has a bug (missing notification).
testing this patch now:
```
diff --git a/lib/network.py b/lib/network.py
index 8fa264d..326bd68 100644
--- a/lib/network.py
+++ b/lib/network.py
@@ -548,7 +548,9 @@ class Network(util.DaemonThread):
                 # If not finished, get the next header
                 if next_height in [True, False]:
                     self.bc_requests.popleft()
-                    if not next_height:
+                    if next_height:
+                        self.notify('updated')
+                    else:
                         interface.print_error("header didn't connect, dismissing interface")
                         interface.stop()
                 else:

```

ecdsa:
fix added in fdd43bd5ad7f0240bfbeafc395cc7eaf673adfbc

ecdsa:
I have a question about connect_chunk: 
if it fails because the server is publishing a bad chunk, will it loop indefinitely?

PR #1271, opened 2015-06-02
Small improvements to transaction_dialog

kyuupichan:
First, merge the sign and broadcast buttons into one.  This is
essentially the action button, and signing and broadcasting
cannot be done simultaneously.

Next, if signing or broadcasting, have cancel_button show
"Cancel" as its text, otherwise "Close".  I think this is more
logical.

Next, once we've broadcast a transaction, hide the broadcast
button.

Finally, have the dialog return 1 if closed (success) and 0
if cancelled (failure).  This will be useful when hooking
into the main window the ability to see a transaction before
sending: at present, if you press Close (this patch changes
the text to Cancel) the Send tab is unconditionally cleared,
which is annoying.  We'll be able to have it clear only if
the tx is broadcast.

I've taken a careful look at the plugins that might be affected
by this (cosigner_pool and greenaddress_instant) and it does
not seem they will be affected.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -568,9 +568,10 @@ class ElectrumWindow(QMainWindow):
         d.exec_()
 
     def show_transaction(self, tx):
+        '''Show transaction dialog.  Return 1 if Close was pressed, 0 if Cancel'''
         import transaction_dialog
         d = transaction_dialog.TxDialog(tx, self)
-        d.exec_()
+        return d.exec_()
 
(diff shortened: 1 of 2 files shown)


ecdsa:
concerning the return value, I plan to make this window non-modal; 
there are use cases for keeping a tx window open while doing other things

kyuupichan:
I've got what I think is a workable solution to the issues we discussed yesterday.  I'll submit that in parts....

PR #1278, opened 2015-06-04
SocketPipe shouldn't close the socket.

kyuupichan:
This is a layering violation - the SocketPipe doesn't own
the socket and provides no other way to close the socket, leading
to unnecessary complexity like that in interface.py.

I looked at deamon.py and NetworkProxy - the two other users,
and they don't close the sockets explicitly, just let them be
garbage collected.


ecdsa:
note: SocketPipe should also not have an infinite loop in _send

PR #1281, opened 2015-06-08
Make the synchronizer not a thread.

kyuupichan:
The synchronizer's work is done from the network proxy's main loop.

A minor problem with the old synchronizer was that it considered itself
out of date if the network was out of date.  This was too generic: the
network can have pending requests unrelated to the synchronizer.  This
resulted in the synchronizer often unnecessarily flipping the wallet
between up-to-date and not-up-to-date, and causing unnecessary calls
to wallet.save_transactions().  This was observable when opening the
network dialog box: frequently just opening it would cause a wallet
status change and transaction flush, simply because the network dialog
sends a get_parameters() request.  This rework of the synchronizer does
not have that issue.

### DIFF
--- a/lib/network_proxy.py
+++ b/lib/network_proxy.py
@@ -61,10 +61,13 @@ class NetworkProxy(util.DaemonThread):
         self.blockchain_height = 0
         self.server_height = 0
         self.interfaces = []
+        self.jobs = []
 
 
     def run(self):
         while self.is_running():
+            for job in self.jobs:
+                job()
             try:
                 response = self.pipe.get()
             except util.timeout:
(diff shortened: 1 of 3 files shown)


kyuupichan:
Oh, FWIW, with this synchronizer.py is PEP0008-clean

ecdsa:
I am thinking about runing the wallet synchronization in the daemon (if there is a daemon)
this would probably impct things

kyuupichan:
OK, I'm happy to leave this for a while, perhaps things will get clearer.  For now I'm happy to finish the TX dialog changes, and finish the network.py/interface.py side of the daemon breakline.

ecdsa:
merged manually; I think the jobs thing is generic enough for future changes

ecdsa:
I fixed the synchronization issue I mentioned

PR #1284, opened 2015-06-10
Fix SSL certificate verification errors

kyuupichan:
This fixes SSL validation on (at least) DragonFlyBSD.


kyuupichan:
We should put this in util.py as it seems there are 2 other places making the same mistake.

Also, this explains why version_getter was always failing for me

PR #1285, opened 2015-06-10
Fix SSL certificate verification errors

kyuupichan:
This fixes SSL validation on (at least) DragonFlyBSD.
The version getter would always complain it failed; this
fixes that too it appears.


kyuupichan:
I just noticed that the Python docs mention for HTTPSConnection:

   Changed in version 2.7.9: context was added.

I think 2.7.9 switched verification on by default and this was added as a consequence.

FWIW my MacOSX has Python 2.7.9.

ecdsa:
replaced with requests

PR #1288, opened 2015-06-10
Clear address cache on switching interface or losing connectivity.

kyuupichan:


ecdsa:
that seems to clears the cache if any interface gets disconnected..

PR #1314, opened 2015-06-25
Have show-before-broadcast defer password request

kyuupichan:
Show-before-send currently asks for the password
and creates a signed tx before showing anything.

This is a little unnerving as you can't see what
you're being asked to sign.

This patch does a few things:

- Adds a description label to the TX dialog
- You see the tx before being asked for password;
  that is only requested on pressing Sign
- in show-before-broadcast mode, the Send button
  text is instead "Show...".  Hook up this button
  text change to the prefs dialog too.
- Remove many redundant imports.  In particular
  PyQ4 is checked in main_window.py so no need
  in tx_dialog too.

Note I had to remove disabling of the Sign button,
because if the user presses "cancel" nothing will
re-enable it.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -16,13 +16,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-import sys, time, re, threading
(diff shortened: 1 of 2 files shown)


ecdsa:
thanks. I guess the disabling is a minor issue

PR #1320, opened 2015-06-27
Split tx dialog buttons

kyuupichan:
The "actions" (sign, broadcast, send to cosigner, verify GA instant)
go on the RHS next to the "Cancel" button.

Copy, QR code and save go on the left.

### DIFF
--- a/gui/qt/transaction_dialog.py
+++ b/gui/qt/transaction_dialog.py
@@ -97,10 +97,15 @@ class TxDialog(QWidget):
 
         self.copy_button = CopyButton(lambda: str(self.tx), self.parent.app)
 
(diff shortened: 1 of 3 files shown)


kyuupichan:
Thomas - I think this is a slight improvement, as the action that the Cancel button refers to would now appear grouped with the Cancel button.   If there are no actions, the button just says Close.

The other 3 are left-aligned.

I recognize this is debatable so wanted your opinion / review.  Thanks.  This is the last of the TX dialog changes I planned.

kyuupichan:
If you don't like them flush-left, we could put another spacer to the left, with a weight of e.g. 8

ecdsa:
did you test this with  the cosigner pool plugin?

kyuupichan:
I did.

PR #1323, opened 2015-06-27
Add seed scroll

wozz:
Add scroll area for multisig wallets.  Without a scroll area, the seed text areas are too small with high N wallets.


ecdsa:
you forgot a line with vbox0 (fixed)

PR #1332, opened 2015-07-01
fix small typos at lib/interface.py and lib/commands.py

romanz:
Running `python lib/interface.py` failed with:

    AttributeError: 'X509' object has no attribute 'slow_parse'

After removing the lines, it runs successfully.


ecdsa:
thanks!

PR #1340, opened 2015-07-03
Remove apparently dead code.

kyuupichan:
I noticed this whilst debugging the daemon issue.


ecdsa:
right, this was deprecated by the wallet.txi/txo data structures

PR #1364, opened 2015-07-16
Wildcard SSL Support

neocogent:
I've tested this small change with my wildcard cert as I needed it for internal use on an electrum server.

A good example of what SubjectAltName looks like for wildcards can be seen in the wikipedia.org certificate. 

I try to make sure it can't be abused by matching sum of wildcard part and base name to full name string.


ecdsa:
thanks!

PR #1389, opened 2015-08-09
remove unused code

romanz:


ecdsa:
might need create_csr in the future, but I know where to find it

romanz:
@ecdsa , sorry for breaking the existing code.
I would like to add more unit tests, so this won't happen again.

PR #1405, opened 2015-08-19
Should fix #1403 (new Unicode issue)

### LINKED ISSUE #1403: hw1 and unicode
after upgrading to any revision past  98930575546d9a18df646e72198deb22a750992a  (unicode conversion)  I can no longer sign transactions with the HW1.
When I click the sign button, I get the following :

  File "/home/richi/sourcecode/electrum/gui/qt/util.py", line 43, in run
    self.result = self.run_task()
  File "/home/richi/sourcecode/electrum/gui/qt/main_window.py", line 1292, in sign_thread
    self.wallet.sign_transaction(tx, password)
  File "/home/richi/sourcecode/electrum/plugins/btchipwallet.py", line 397, in sign_transaction
    self.get_client() # prompt for the PIN before displaying the dialog if necessary
  File "/home/richi/sourcecode/electrum/plugins/btchipwallet.py", line 231, in get_client
    raise Exception("Could not connect to your Ledger wallet. Please verify access permissions, PIN, or unplug the dongle and plug it again")
Exception: Could not connect to your Ledger wallet. Please verify access permissions, PIN, or unplug the dongle and plug it again 

Looks like some string needs to be converted back to ascii before sending it to the ledger hardware wallet...

btchip:


ecdsa:
this should not be fixed at the plugin level.
see if df206719b74a2109be3144ba5dc62c039207df2b works

btchip:
yes, it does

PR #1410, opened 2015-08-20
Added blockonomics as blockexplorer

shivaenigma:
- blockonomics supports taking user addresses as parameter in tx url and then calculates correct net amount and highlights user address (similar to show details in electrum). 
- If you require any changes in url format in blockonomics to support this , let me know we can get that done

### DIFF
--- a/gui/qt/history_widget.py
+++ b/gui/qt/history_widget.py
@@ -94,7 +94,19 @@ class HistoryWidget(MyTreeWidget):
         tx_hash = str(item.data(0, Qt.UserRole).toString())
         if not tx_hash:
             return
-        tx_URL = block_explorer_URL(self.config, 'tx', tx_hash)
+          
+        my_tx_addresses = [] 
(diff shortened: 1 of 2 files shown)


ecdsa:
this increases the complexity of the code without obvious benefot

shivaenigma:
You don't see any benefits  of https://www.blockonomics.co/api/tx?txid=3b14f2b256a5fad043311e4b36674d455d4ff22db404e8e205af29fafcd0870c&addr=1B9Hbrcd9sSMg4acHBttjvYE9kucmNEG1k vs something like https://blockchain.info/tx/3b14f2b256a5fad043311e4b36674d455d4ff22db404e8e205af29fafcd0870c ? 

It is easier for user to see of amount of transaction instead of manually comparing his addresses.

As I said if there are any changes required in url format to decrease complexity ... I can make those in blockonomics.

ecdsa:
sorry, not interested. I think this is redundant with what the wallet already provides.

PR #1432, opened 2015-08-31
transaction: remove unused mmap-related code

romanz:
I have used the following command to check that no one uses the removed code:

```
git grep -E "map_file|seek_file|close_file|mmap"
```


romanz:
Unit tests should be fixed by #1433

PR #1436, opened 2015-09-01
fixed attempt to access account derivation before it is created

dstanchfield:
Made the change to Trezor too.


aussiehash:
Confirmed, working now with both hardware wallets.

PR #1472, opened 2015-10-07
use `pip install -t` in make_packages to aggregate packages.

akshayaurora:
use `pip install -t` in make_packages to aggregate packages & comment android.branch=master(allows it to run with latest bulldozer)


ecdsa:
I merged this to fast; I see now that this crates a lot of junk directories in 'packages'
this is not good, because we want packages to be minimal.

PR #1481, opened 2015-10-15
Exchange rate fixes

nomoon:
One fix and one addition.


nomoon:
I'm not sure if BitcoinAverage is the only API publishing historical data as CSV. They sure do have data for a lot of currencies, though!

kyuupichan:
Thanks

PR #1492, opened 2015-10-18
Make seed uncopyable

ctrlcctrlv:
[Inspired by this Reddit conversation](https://www.reddit.com/r/Bitcoin/comments/3p4bq1/electrum_v25/cw380kg)

Please merge so that no more users lose coins. Humans like to take the path of least resistance, which copying and pasting is. (I should know, just look at my username ... )


ecdsa:
I've been pondering for a while about doing this. still not sure...

ctrlcctrlv:
What are your reservations?

Advanced users can still generate tons of seeds (I have a script that does this), while amateurs are less likely to get burned.

It's too easy to just copy and paste the seed and then never look at it again right now. I feel like people are often surprised that their money is gone when they lose the seed, no matter how many warnings we put up.

ecdsa:
My reservation is that user who is completely new to this process will not copy/paste the seed, because they do not know that the next screen is going to ask for it. 
If someone chooses to copy/paste the seed, it must be because they know what the next step is.

ctrlcctrlv:
All I have to go on is personal experience; but the first time I did this I had to restart the process since I didn't even copy it, and then I copied it :wink: 

It was only after doing some reading that I realized the seed was a recovery seed, and then I got the seed from Wallet>Seed...

The New Wallet dialog clearly says to write it down, and failing that, on the second go around, not allowing copying helps assure that it will be saved.

chris-belcher:
I've had the same experience as @ctrlcctrlv 

We as software developers have some responsibility to make our applications as safe to use as possible. The idea that losing information about private keys will irreversibly destroy your money is a new concept to most humans. I think the inconvenience of experienced users should not override the actual real monetary loss of the newb.

ecdsa:
I agree; will do

ctrlcctrlv:
I made it only affect the wizard and not the menu item after an IRC discussion

ecdsa:
I reverted this, and replaced it by a clibboard.clear(); that should offer more flexibility.

PR #1499, opened 2015-10-18
Missing sys import

SmileyChris:
Also put the error message in sys.exit rather than printing. This will output it to stderr and make the exit code non-zero like it should be.


kyuupichan:
Thanks

PR #1509, opened 2015-10-21
could not view history of single address

cryptapus:


ecdsa:
the issue seems to be caused by 0371a3dc32350c1923348c216ee2c2f8eee2d4d1

kyuupichan:
Wasn't aware of this feature!  Taking a look.

PR #1527, opened 2015-10-30
Include a conditional shebang, looks for `python2` if not available runs with `python`

akshayaurora:


ecdsa:
closing: we moved to python3

PR #1530, opened 2015-10-30
amount dialog: use show_max bool to decide wheather max_button is dis…

akshayaurora:
…played


gui/kivy/main_window.py
```diff
@@ -733,6 +733,16 @@ def tx_dialog(self, tx_hash):
 
     def amount_dialog(self, label, callback):
         popup = Builder.load_file('gui/kivy/uix/ui_screens/amount.kv')
+        # Hide max_button
+        show_max = False
```
akshayaurora: Should be True by default,  this is just to show how to achieve hiding of a widget.

PR #1544, opened 2015-11-09
Fix error when loading multisig transaction

09128319047:
When loading a multisig transaction on different environments this error was found:

```
IndexError: list index out of range
```

This small patch fixes the issue and these transactions are able to load successfully.

### DIFF
--- a/lib/transaction.py
+++ b/lib/transaction.py
@@ -723,7 +723,7 @@ class Transaction:
                 # input is complete
                 continue
             for k, x_pubkey in enumerate(txin['x_pubkeys']):
-                if x_signatures[k] is not None:
+                if k < len(x_signatures) and x_signatures[k] is not None:
                     # this pubkey already signed
                     continue
                 out.add(x_pubkey)


ecdsa:
how was the initial transaction created?

09128319047:
Created with Bitcoin Core. 

How it is created should not matter, every other client tried can load the transaction and Electrum can load it with this minor fix. It does not properly error either, it just silently closes. This fixes it and allows the user to sign and broadcast the transaction.

ecdsa:
is the initial transaction already partially signed, or completely unsigned?

ecdsa:
I guess it was partially signed by bitcoind, because you did sign and broadcast.

The fact that you could sign and broadcast with that pach is pure luck, because signatures have to be added in the correct order. Your patch will fail if the order of pubkeys is different. To really fix this I guess we need to check each signature against the provided pubkeys.

09128319047:
Yes, it was partially signed.

When you say "signatures have to be added in the correct order", do you mean in general or that is how Electrum expects it?

Yes, I have noticed that this patch does work for loading but signing some will get the following error:

    txin['signatures'][ii] = sig.encode('hex')
    IndexError: list assignment index out of range

How long do you think a fix will be available? Would it be possible to donate to Electrum developers to get this patched faster?

09128319047:
I would be willing to pay a fair sum of Bitcoin to have this issue fixed.

09128319047:
esdsa: Can you please email me: electrumbounties@tutanota.com

I would like to put up a bounty to get this issue fixed as soon as possible. I can explain in more detail through email. I would also like to donate to the Electrum project in general and it is not easy for me to get on freenode IRC.

ecdsa:
email sent, no answer

rdymac:
For the record, this user `09128319047` has been accused of introducing a malicious code into https://github.com/WalletWasabi/WalletWasabi back in February 2024 . https://github.com/WalletWasabi/WalletWasabi/issues/14410 

Source: https://x.com/noosphere888x2/status/2088290231250210827

PR #1551, opened 2015-11-15
Update text.py

psb777:
Inverted QR-codes do not work with many QR scanners including Android's "Barcode Scanner" and the one used by Circle.com.


ecdsa:
my terminal uses an inverted color scheme (dark background).
pease write a pull request that can handle both (if possible, detect color scheme)

psb777:
Of course! There is no solution to this problem as the colours used are locally defined. Perhaps the easiest solution is to display both inversions?

ecdsa:
fixed in 47e5bd6

PR #1558, opened 2015-11-19
Improve get_target

fanatid:
Need attentive review.

### DIFF
--- a/lib/blockchain.py
+++ b/lib/blockchain.py
@@ -50,7 +50,11 @@ class Blockchain(util.PrintError):
                 self.print_error("prev hash mismatch: %s vs %s"
                                  % (prev_hash, header.get('prev_block_hash')))
                 return False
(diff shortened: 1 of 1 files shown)


ecdsa:
good catch. 
there is redundant code between verify_chain and verify_chunk, we should avoid that

ecdsa:
@fanatid I sent you an email, can you answer?

ecdsa:
@fanatid if you have time, please check 2a1fbf01fe63dee9f950f7ba7b4d68d456799262

fanatid:
@ecdsa I checked your commit. Nicely done!
I found two things that make code not clear:
1. function name `verify_headers` -- verify exactly **one** header, not many header**s**
2. in `verify_chunk` you save chunk, although should only verify headers.. `verify_chain` do this in right way

If you want I can make PR with changes.

ecdsa:
please do

PR #1636, opened 2016-01-17
Follow XDG base directory standard

rnhmjoj:
This makes Electrum store its files as specified in the [XDG standard](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) by default and, if any is found, will move the files from the current location to the new one.

### DIFF
--- a/lib/util.py
+++ b/lib/util.py
@@ -246,7 +246,17 @@ def user_dir():
     if 'ANDROID_DATA' in os.environ:
         return android_check_data_dir()
     elif os.name == 'posix':
-        return os.path.join(os.environ["HOME"], ".electrum")
+        xdg_default = os.path.join(os.environ["HOME"], ".local", "share")
+        xdg_dir = os.path.join(os.getenv("XDG_DATA_HOME", xdg_default), "electrum")
+
+        # handle migration to XDG compliant location
+        old_default = os.path.join(os.environ["HOME"], ".electrum")
+        if os.path.isdir(old_default):
+            print_msg("Migrating Electrum config...")
+            shutil.move(old_default, xdg_dir)
+            print_msg("Completed, moved to", xdg_dir)
+
+        return xdg_dir
     elif "APPDATA" in os.environ:
         return os.path.join(os.environ["APPDATA"], "Electrum")
     elif "LOCALAPPDATA" in os.environ:


ecdsa:
migrating the directory might break existing scripts..

ecdsa:
not completely unrelated: 60b595356b8c049adc296deb0ffe65b357500ab4

rnhmjoj:
I fixed a conflict

ysangkok:
If we change this, we should actually put the socket files somewhere else. I got a request from @flokli to use XDG dirs, but the reasoning seems to be that it allows better backups because the backup script will not backup socket files.

And should the Bitcoin headers be backed up or not? Since they are usually easily reconstructable, do they belong in cache?

flokli:
@ysangkok When following [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), it does not only only make backups more efficient (as you can skip the whole XDG_CACHE_HOME), it also simplifies dotfile management, as files in .config don't change all the time without user interaction and therefore need to be manually excluded.

From my understanding, socket files should go to `XDG_RUNTIME_DIR`:
> $XDG_RUNTIME_DIR defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

`blockchain_headers` should be something for `XDG_CACHE_HOME`.

I'm not sure about all those `recent*` keys in `config`. In case those change randomly without the user explicitly setting things, I'd also suggest to move those parts outside config into `XDG_CACHE_HOME`, too.

@rnhmjoj: the specification talks about multiple directories being possible canddates for config files. Could you use [python-xdg](https://github.com/jleclanche/python-xdg/blob/master/xdg/basedir.py) and iterate over `XDG_CONFIG_DIRS` when loading configuration? (moving current configuration to `XDG_CONFIG_HOME` should still be fine)

felixsanz:
is this going to be merged? please follow specs

lib/util.py
```diff
@@ -246,7 +246,17 @@ def user_dir():
     if 'ANDROID_DATA' in os.environ:
         return android_check_data_dir()
     elif os.name == 'posix':
```
SomberNight, line 248: note that `os.name == 'posix'` even on MacOS

oxalica:
Is this PR still active?

> I'm not sure about all those recent* keys in config.

XDG spec introduced a new `XDG_STATE_DIR` this year which covers history, recent files, view, layout and etc. These keys fit here.
See: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables

I think there should be no unresolved issues now.
- Socket files should go to `XDG_RUNTIME_DIR`
- Recent* keys should go `XDG_STATE_DIR`
- Other config keys expecting manual editing should go `XDG_CONFIG_DIR`
- Important data files which users need to backup, like wallets, go `XDG_DATA_DIR`
- Caching data files like blockchain data go `XDG_CACHE_DIR`

rnhmjoj:
> Is this PR still active?

No, sorry: I stopped using electrum and fighting the XDG crusade long ago[1]. If you still care about this you should open another PR and rebase this one.

[1]: https://github.com/sloonz/rewritefs

PR #1655, opened 2016-01-29
Align Ledger plugin with the others, somewhat.

btchip:
Still some work to be done, but that should make @kyuupichan life easier 

- Creating a new wallet, signing works 
- Reopening the wallet fails (device is not paired) 
- Signing a message fails (device is not paired, probably the same issue)
- No multisignature yet


kyuupichan:
Hi, thanks for this. I'll comment more fully in a few hours.

There is a large amount of copy and paste which should be cleaned up, I can do some of that.

You've changed the wallet type. Won't this break existing wallets?

btchip:
Maybe, depends how it's stored, I didn't check that. You can revert if it
does.

On 29 January 2016 at 07:49, Neil <notifications@github.com> wrote:

> Hi, thanks for this. I'll comment more fully in a few hours.
>
> There is a large amount of copy and paste which should be cleaned up, I
> can do some of that.
>
> You've changed the wallet type. Won't this break existing wallets?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/spesmilo/electrum/pull/1655#issuecomment-176604481>.
>



-- 
Nicolas Bacca
CEO <http://www.simulity.com/>
Mobile +33 6 30 79 86 54 - Skype nicolas_bacca.pro

kyuupichan:
It shouldn't go in broken, sorry, I'm not going to be running around fixing up stuff that's committed broken; you should be willing for the code to ship with the next Electrum release as-is.

I'm not particularly pleased with the huge amount of copy and paste either.

You changed can_create_accounts from False to True.  Why was it False before and True now?

btchip:
ok let's close this one then, I'll keep working on yours.

kyuupichan:
Not sure what you mean.  But let's do this properly.

I've created a hw_wallet directory under plugins/.  I've started with hw_wallet.py.  This is an ABC for the 3 hardware wallets we have.  I've pulled what looks like the common code with the ledger wallet up into this, and had the trezor and keepkey code inherit it.

As a first step can you have your wallet inherit this?  Please comment on my point about can_create_accounts - if you don't support account creation please comment in your code why so others can understand.  If more of your wallet code can be shared with Trezor, such as message signing, please pull the common code up yourself; that would be very helpful, even if it is not 100% the same the common bits should be in one place.

btchip:
I meant I'll keep working on the current code you have, not my modifications (so I'm dropping that pull request and branch)

We support BIP 44 account creation. Not sure why it wasn't enabled previously.

I'll check which parts can be shared. Probably not much.

kyuupichan:
Great.  I've also separated out qt handler common stuff.

Can you derive from that?

There is some stuff that you don't currently use but that's not a big issue.

I've kept char requests and pinmatrix stuff in trezor code as they're clearly irrelevant to you.  I've left the passphrase stuff as although you don't currently use it, it is plausible you might in the future.

Why do you not use a pinmatrixwidget?  Is it because you've not written one?  If so we can pull the widget code into electrum; I've wanted to change it a bit anyway.  Trezor pins are 1-9, what characters can Ledger use?

btchip:
> Can you derive from that?

Sure I'll have a look

> Why do you not use a pinmatrixwidget? Is it because you've not written one? 

Yes

> If so we can pull the widget code into electrum; I've wanted to change it a bit anyway. Trezor pins are 1-9, what characters can Ledger use?

We use 0-9, so 0 can be on the side. Then if you make the pinmatrixwidget display either numbers or ?, then it can be shared between all implementations.

PR #1662, opened 2016-01-30
Ledger plugin redesign

btchip:
Redesign Ledger plugin around the generic components


kyuupichan:
Great, this is much nicer thanks.  The restore wallet bit is wrong, I'll fix that.

If you do that client wrappers like trezor you should be able to avoid all the try..catch stuff, it *should* become unnecessary.

btchip:
yes I did that on the previous commit, will see if I can add that tomorrow
for additional clean up

PR #1663, opened 2016-01-31
Installer fix missing hw_wallet plugin

nomnombtc:
Hello,

the new hw_wallet plugin needs to be added to setup.py installer.


kyuupichan:
Thanks!

PR #1711, opened 2016-03-09
Remove penalty_func from CoinChooserPrivacy that is later redefined

JustinTArthur:
Probably a copy pasta from the development process.


kyuupichan:
Thanks

PR #1725, opened 2016-03-17
fixed requirements for kivy buildozer spec file.

bugelectrum124:


ecdsa:
chardet and urllib3 are already included in packages.
you have to build it with contrib/make_packages (see README)

PR #1738, opened 2016-03-25
Label colour - History Window

### LINKED ISSUE #1393: Feature request: Label colour - History Window
The transaction amount of outgoing transactions, when viewed in the history window, is coloured red. 

It would be a nice if the label of these transactions could be coloured the same red colour so that they stand out more against the other incoming transactions. 

Not a big issue but would be a 'nice to have'.

thephez:
Set colour of label for outgoing transactions to match the colour of the
value field (red) - closes #1393


ecdsa:
not sure if it's a god idea, but let's try

PR #1785, opened 2016-05-04
Store the daemon's network on the stdio GUI object

twopir:
I tried using the stdio interface to send a transaction, and it failed because the object didn't have a network attached. This should fix that.


ecdsa:
was that even tested? there are other occurences of network nearby

twopir:
Nope, you caught me. I fixed the other references.

PR #1851, opened 2016-06-29
Change labelsync port

maran:
I've had to move the server around again. With the latest changes to the server code it's easy to keep moving it around without any more changes to the Electrum code.

### DIFF
--- a/plugins/labels/labels.py
+++ b/plugins/labels/labels.py
@@ -19,7 +19,7 @@ class LabelsPlugin(BasePlugin):
 
     def __init__(self, parent, config, name):
         BasePlugin.__init__(self, parent, config, name)
-        self.target_host = 'sync.bytesized-hosting.com:9090'
+        self.target_host = 'sync.bytesized-hosting.com'
         self.wallets = {}
 
     def encode(self, wallet, msg):


ecdsa:
ok, can we keep using the old port number for 6 months or a year? 
there should be a grace period for users who habe not upgraded their client

maran:
Good point. I will find out if thats possible. It might not be because of
some server side ssl requirements but I will spend some time on it.

Op do 30 jun. 2016 11:27 schreef ThomasV <notifications@github.com>:

> ok, can we keep using the old port number for 6 months or a year?
> there should be a grace period for users who habe not upgraded their client
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/spesmilo/electrum/pull/1851#issuecomment-229608715>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AABRdyGJHv80n2kr4Dh8n3oNigahOtY3ks5qQ4v1gaJpZM4JBLF6>
> .
>
-- 

Sent from a mobile device excuse the brevity.

maran:
I have removed. The need for port change however I am now getting an SSL cert error from Python. Chrome considers the cert valid. Any idea how to get more info about why Python did not agree with it?

ecdsa:
Firefox is unable to connect.

maran:
Closing this PR and keeping the current running. It should be working fine on the original port again.

PR #1882, opened 2016-08-03
Typo

mcccs:
Isn't this a typo? There are 2 spaces, but there should be one.
// Mine doesn't compile because the master branch isn't compiling. Not my fault.
Note: if that's a typo, gui/qt/main_window.py:2540 has the same typo


ecdsa:
deprecated

PR #1888, opened 2016-08-08
Fail better and sooner when sweeping an amount worth less than fee

peter-conalgo:
When sweeping a private key, if the balance available was less than the 1kb fee, a negative amount was rendered as hex and that didn't go well when signing. Added some assertions to catch at a higher level.

This is just to the contain the bug, it doesn't provide user feedback, or better, offer to sweep up the UTXO as a miner donation.

Also a trivial typo-bug fix.

### DIFF
--- a/lib/bitcoin.py
+++ b/lib/bitcoin.py
@@ -115,6 +115,7 @@ def rev_hex(s):
 
 
 def int_to_hex(i, length=1):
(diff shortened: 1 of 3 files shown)


ecdsa:
hi, can you resubmit this, it has conflicts with the new code

peter-conalgo:
These changes are now included in PR #1898

PR #1909, opened 2016-08-28
Ledger pack : Nano S support, rewrite/cleanup, P2SH

btchip:
To be tested with btchip-python v0.1.18 from https://github.com/ledgerhq/btchip-python


ecdsa:
thanks!!

PR #1981, opened 2016-10-15
Fixing 2 tests in test_mnemonic.py

jonathancross:
Fixes both broken tests in test_mnemonic.py

Notes:
* prepare_seed method was removed long ago.
* custom_entropy was removed 2 weeks ago.


jonathancross:
You can see that all tests in `test_mnemonic.py` are now passing.

PR #2021, opened 2016-11-06
fix plot history plugin

yate:
Line 54 had a 'too many values to unpack' error. 

There was also an issue with matplotlib where it was using Qt5Agg as a backend which didn't seem to work with PyQt4.  I manually set the backend in the imports.


ecdsa:
fixd manually in b8328da99ffa4e0638c22ce9f86ce924a202b819

PR #2071, opened 2016-12-08
Add fee rate to transaction details dialog.

neocogent:
As per my issue 2051 feature request.

Add fee rate info as base_unit/kB and sat/byte in brackets after fee in details. 

Allow users to see rate used without having to manually calculate.

Tested with new tx and also in history list popup. Only shows when fee is available.

### DIFF
--- a/gui/qt/transaction_dialog.py
+++ b/gui/qt/transaction_dialog.py
@@ -204,6 +204,9 @@ class TxDialog(QDialog, MessageBoxMixin):
         else:
             amount_str = _("Amount sent:") + ' %s'% format_amount(-amount) + ' ' + base_unit
         fee_str = _("Transaction fee") + ': %s'% (format_amount(fee) + ' ' + base_unit if fee is not None else _('unknown'))
+        if fee is not None:
+            size = self.tx.estimated_size()
+            fee_str += '   ( %d bytes @ %s ' % (size, format_amount(fee * 1000 / size)) + base_unit + '/kB, %.0d sat/byte )' % (fee/size)
         self.amount_label.setText(amount_str)
         self.fee_label.setText(fee_str)
         run_hook('transaction_dialog_update', self)
--- a/lib/transaction.py
+++ b/lib/transaction.py
@@ -670,7 +670,7 @@ class Transaction:
     @profiler
     def estimated_size(self):
         '''Return an estimated tx size in bytes.'''
-        return len(self.serialize(-1)) / 2  # ASCII hex string
+        return len(self.serialize(-1)) / 2 if not self.is_complete() or self.raw is None else len(self.raw) / 2 # ASCII hex string
 
     @classmethod
     def estimated_input_size(self, txin):


neocogent:
Here is an example screen clip of added fee details.

http://imgur.com/a/flAJ0

ecdsa:
thank you. two comments:
1. estimated_size() should only be used when the actual size is unknown. if the tx is signed then we should display its real size
2. your pull request has a lot of merge commits, maybe you could clean it up (create a new local branch)

neocogent:
Sorry about the messy commit history. I rebased my local repo and fixed it up. Should be clean now.

Regarding estimated_size() - I was wondering about that but when I looked this function is simply takes the length of the serialized tx. So it's not really an estimate as name implies. I went looking for an alternative that maybe calculates it differently and couldn't find a better way. 

This function is a one liner as below,

    return len(self.serialize(-1)) / 2  # ASCII hex string

At the time of showing a tx preview I'd presume it's not signed yet and so the serialized size may change after signing. Is there a way to predict that better during preview?

ecdsa:
no, self.serialize(-1) is an estimate.
if the tx is signed, its actual size is ```len(self.raw)/2 ```

neocogent:
Ok. Probably better to add a tx function that makes this distinction and can be used by the tx dialog but doesn't expose the calculation (which should be internal to the tx object). It would use len(self.raw)/2 when "complete" and estimated_size when not complete. Or, even better update estimated_size to use correct length when tx is "complete", ie. estimate is actual when complete is known.

neocogent:
I've updated my commit now with modified estimated_size. While testing I noticed that this function is used to calculate the fee when "not is_complete()" and "self.raw is None" - so I had to test both conditions in order to make sure it chooses correct calc method. Note I didn't test this with a multisig wallet as currently I don't have one available (in case maybe it behaves differently).

One thing I noticed when testing this is that when a new tx is created and the preview is clicked, without touching the slider, it will show an incorrect lower fee (it doesn't use suggested rate, currently 1.17 mbtc/kB). But closing preview and moving the slider back/forth to see current rate and then clicking preview again now shows a fee based on correct rate. I think this may be a bug in fee rate usage and need to investigate separately from my changes. If that's the case I will add it as a new issue.

neocogent:
Thanks.

I've been working the last couple days on a revision to the ledger plugin - improving the user interface and adding mobile pairing/auth support. I just got an HW.1 and really want to use it with Electrum but find the text or security card methods insufficient. I have it pairing ok and I'm just trying to fix some bugs related to getting validation on the phone. Seems maybe related to my very old android version. Anyway, I hope to tidy it up and maybe submit next week.

ecdsa:
@neocogent I don't understand; are you trying to add HW1 support to the Android version?

neocogent:
No. Adding mobile phone 2FA validation to desktop Electrum. 

Currently the ledger plugin only supports text pin and security card auth. IMO both of those methods are problematic (text pin requires a second system, and "secure card" is not very secure against malware). I want to use the LedgerWallet mobile app for 2FA auth because it uses a dongle to phone encrypted channel for validation. But right now it requires using the Ledger Chrome app. I want to use Electrum with Nano/HW.1 and have mobile auth workable. IMO that would be far better than current Nano/HW.1 usage on Electrum.

I now have Electrum pairing fine with phone app but there is some problem with actual tx validation. I send the encrypted "secureScreenData" to the mobile (via websocket) but the mobile cannot find the pairing key (which it does have in its keystore). I believe this is due to some mishandling of keystores on my older version of Android (4.2.2 - pre Android keystore). Today I am building a debug version of the LedgerWallet app for debugging this (mostly because I do not have access to a newer Android phone to see if this is only the problem).

TheButterZone:
I'm a noob at reading Github but did the sat/byte sizer get removed in spesmilo's master Electrum 2.8.2? I just added those 2 lines back(?) into my transaction_dialog.py, relaunched Electrum & it seems to work fine.

neocogent:
I think it was. I also have a local tweaks branch with sat/byte restored. I haven't sync'd recently and I'm still back on 2.7.12 myself. 

Here's a link to the commit in my repo my-tweaks branch where I re-added it. It may not be exactly right for newer releases but should be a starting point.

https://github.com/neocogent/electrum/commit/b7e2a975b6a143c61c17d4168dee8ff2b00f1eac

TheButterZone:
Thanks!

PR #2079, opened 2016-12-17
Ledger new ui and mobile 2fa validation

neocogent:
Adds new mobile 2FA support and improves look and feel of user interface for both text and key card modes. I've started the PR but expect there will be revisions and much testing before merging.

If you are at all interested in mobile 2FA for your Ledger device (you should be as it's more secure than other methods) then **PLEASE TEST**, and comment here. I can only test a limited set of conditions.

I've tested this extensively on my system. I used an HW.1, Linux and an older model Android phone. I had to build Android LedgerWallet myself because it didn't work properly on my phone. But I didn't change any LW code, just ran a debug release and it worked fine (production release did not respond to requests for validation). So this needs to be tested on newer phones, and with more Ledger devices.

The pairing info is stored in the wallet not global config. I believe each dongle only stores one pairing public key so it only makes sense to track one pairing id per device and one wallet. Let me know if there is reason to do otherwise.

Here are some screen shots for what to expect:

Sign TX with key card mode: http://imgur.com/lTsZ3PZ
Sign TX with mobile: http://imgur.com/khiWLek
Pair mobile (on the fly during TX sign): http://imgur.com/RPiRh7C
Sign message (text PIN only): http://imgur.com/aPOGiUb

If you have a local fork of Electrum you can get this pull request as a branch for local testing with this command:

`git fetch origin pull/2079/head:ledger2fa`

Then switch to the new branch:  

`git checkout ledger2fa  `

and run ./electrum (well, on linux anyway).

### DIFF
--- a/RELEASE-NOTES
+++ b/RELEASE-NOTES
@@ -1,3 +1,6 @@
+# Release 2.7.12
+  various bugfixes
+
 # Release 2.7.11
   * fix offline signing (issue #195)
   * fix android crashes caused by threads
--- a/lib/version.py
+++ b/lib/version.py
@@ -1,4 +1,4 @@
-ELECTRUM_VERSION = '2.7.11' # version of the client package
+ELECTRUM_VERSION = '2.7.12' # version of the client package
 PROTOCOL_VERSION = '0.10'   # protocol version requested
 
 # The hash of the mnemonic seed must begin with this


ecdsa:
note: this needs to be tested with Nano S too, to make sure it does not break anything.

neocogent:
Yes. I don't have one so I trust someone will give it some testing. 

I have had some segfaults upon closing so forward to any other experiences. I'm still testing that and tweaking some bits. Especially those with newer Android 4.3+ to see if they respond to auth requests. I will see if I can dig up another phone to test with as well. I did think it better to post the PR as it's mostly done and can get some exposure.

btchip:
Nano S support should be pretty much transparent as it's ignoring all the 2FA logic. Thanks for the pull reques and the great addition, will be reviewing this week.

neocogent:
My HW.1 wiped itself today. Not sure why as it was working and then I tried starting Electrum and it suddenly didn't ask for PIN but instead tried to do the init/setup process. That process fails due to some threading issue. I reverted back to master branch and it was the same. It tried to start a dialog for btchipPersoWizard in the btchip module and always fails pretty much immediately. I don't think this is related to my changes. 

Anyway, I had to re-init the device and then it worked ok, but currently cannot be setup in Electrum.

neocogent:
Tried to merge in fix for thread safety on setup process but clobbered the whole PR. Now re-opened as #2085. Very sorry!

PR #2085, opened 2016-12-21
Ledger new ui and mobile 2fa validation, fix setup thread safety

neocogent:
This the same PR as #2079 but when trying to update that one I somehow reverted back before this and deleted the PR. My apologies! I will copy that PR info here but note there are a few comments that are in the other PR. And this new PR now includes a fix for thread safety of the setup process by adding a signal to safely invoke the btchipPersoDialog (which is what I was trying to merge in but it got messed up).

--------
Adds new mobile 2FA support and improves look and feel of user interface for both text and key card modes. I've started the PR but expect there will be revisions and much testing before merging.

If you are at all interested in mobile 2FA for your Ledger device (you should be as it's more secure than other methods) then **PLEASE TEST**, and comment here. I can only test a limited set of conditions.

I've tested this extensively on my system. I used an HW.1, Linux and an older model Android phone. I had to build Android LedgerWallet myself because it didn't work properly on my phone. But I didn't change any LW code, just ran a debug release and it worked fine (production release did not respond to requests for validation). So this needs to be tested on newer phones, and with more Ledger devices.

The pairing info is stored in the wallet not global config. I believe each dongle only stores one pairing public key so it only makes sense to track one pairing id per device and one wallet. Let me know if there is reason to do otherwise.

Here are some screen shots for what to expect:

Sign TX with key card mode: http://imgur.com/lTsZ3PZ
Sign TX with mobile: http://imgur.com/khiWLek
Pair mobile (on the fly during TX sign): http://imgur.com/RPiRh7C
Sign message (text PIN only): http://imgur.com/aPOGiUb

If you have a local fork of Electrum you can get this pull request as a branch for local testing with this command:

`git fetch origin pull/2085/head:ledger2fa`

Then switch to the new branch:  

`git checkout ledger2fa  `

and run ./electrum (well, on linux anyway).

### DIFF
--- a/plugins/ledger/auth2fa.py
+++ b/plugins/ledger/auth2fa.py
@@ -1,0 +1,347 @@
+import threading
+
+from PyQt4.Qt import (QDialog, QInputDialog, QLineEdit, QTextEdit, QVBoxLayout, QLabel, SIGNAL)
+import PyQt4.QtCore as QtCore
+
+from electrum.i18n import _
+from electrum_gui.qt.util import *
+from electrum.util import print_msg
+
+import os, hashlib, websocket, threading, logging, json, copy
+from electrum_gui.qt.qrcodewidget import QRCodeWidget, QRDialog
+from btchip.btchip import *
+
+DEBUG = False
+
+helpTxt = [_("Your Ledger Wallet wants tell you a one-time PIN code.<br><br>" \
+            "For best security you should unplug your device, open a text editor on another computer, " \
+            "put your cursor into it, and plug your device into that computer. " \
+            "It will output a summary of the transaction being signed and a one-time PIN.<br><br>" \
+            "Verify the transaction summary and type the PIN code here.<br><br>" \
+            "Before pressing enter, plug the device back into this computer.<br>" ),
+        _("Verify the address below.<br>Type the character from your security card corresponding to the <u><b>BOLD</b></u> character."),
+        _("Waiting for authentication on your mobile phone"),
+        _("Transaction accepted by mobile phone. Waiting for confirmation."),
+        _("Click Pair button to begin pairing a mobile phone."),
+        _("Scan this QR code with your LedgerWallet phone app to pair it with this Ledger device.<br>" 
+            "To complete pairing you will need your security card to answer a challenge." )
+        ]
+
+class LedgerAuthDialog(QDialog):
+    def __init__(self, handler, data):
+        '''Ask user for 2nd factor authentication. Support text, security card and paired mobile methods.
+        Use last method from settings, but support new pairing and downgrade.
+        '''
+        QDialog.__init__(self, handler.top_level_window())
+        self.handler = handler
+        self.txdata = data
+        self.idxs = self.txdata['keycardData'] if self.txdata['confirmationType'] > 1 else ''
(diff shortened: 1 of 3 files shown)


btchip:
> And this new PR now includes a fix for thread safety of the setup process by adding a signal to safely invoke the btchipPersoDialog

Great, I noticed that one before but was never able to fix it properly so that's quite appreciated. Also the other "random reset" bug is fixed in the upcoming firmware, so nothing bad on your side.

neocogent:
"Random reset" - is that the one that indicates XResource temporarily unavailable and XInitThread not called or sometimes shows as Segmentation Fault? I've had that randomly and spent hours trying to figure where I'm messing up with the threading or garbage collection maybe in wrong order. If that's fixed in firmware, then super! I was getting pretty frustrated with that and figured it was me. Putting in debug prints almost always made it go away, and running in gdb left no stack to trace.

I wiped my dongle on purpose to test the PersoDialog thing. I didn't try every path through the dialogs but it didn't crash, and I was able to create a new random seed, output it in a text editor, and then use the wallet.

btchip:
No, the "random reset" thing is the dongle clearing itself sometimes after too many mobile authentications. The other one is likely related to the HID stack.

ecdsa:
@btchip let me know if you have tested this pull request

ecdsa:
This ads a new dependency: websocket
Dependencies must be listed in __init__.py
It would be even better not to add new dependencies

neocogent:
I thought I checked that before beginning and Electrum already used websocket somewhere. I'm just checking again now.

ecdsa:
well, I think it would be better to not list websocket as a dependency, and have a fallback instead of crashing.
that way the plugin can stil be used if websocket is not installed.

btchip:
I've successfully tested it with different HW.1 configurations and it's quite a large improvement to my previous ugly UI - also note that there is also a dependency is on 'websocket-client', not just 'websocket' (otherwise the script fails to find websocket.WebSocketApp) 

So it's good to merge for me after this is solved (either by marking it a mandatory dependency or disabling the web pairing menu with a text prompting the user to install the module, as you wish)

ecdsa:
I prefer not to make it mandatory, because it would disable existing installations

ecdsa:
ok, merged manually with an error fallback

neocogent:
I'll have a look at this again shortly. What does "merged manually" mean?
I wanted to look at this in more detail to see if I could make it so it didn't need websocket or websocket-client. I haven't had time this week.

ecdsa:
it means I added a commit to handle the import error

neocogent:
I noticed that Electrum has a websockets module in lib already. It uses this line,

    from SimpleWebSocketServer import WebSocket, SimpleSSLWebSocketServer

and seems to extend WebSocket to provide a client class,

    class ElectrumWebSocket(WebSocket):

Is this usually installed by users or rarely used? If it's usually already installed then I could probably modify the ledger/auth2fa code to make use of that instead of requiring yet another package. Just wondering what can be done to make it easier for users and not need more dependencies.

ecdsa:
no, electrum does not expect this module to be installed.
it is only required by the daemon if you use it on a web server.

PR #2086, opened 2016-12-21
Properly handle invalid payment request URLs

alexander255:
Otherwise the client would simply hang when receiving an invalid URL dumping a hard-to-read exception to the terminal.

While this is an improvement over the previous situation (and therefor should be merged), it doesn't actually address the underlying that caused this situation in the first place: Coinbase uses payment request URLs for their processing, but also Cloudflare for their webservers. Considering how easy was to enable TOR in my Electrum instance I'd definitely call this a series of unfortunate circumstances. I'm not sure what to do about that through…

### DIFF
--- a/lib/paymentrequest.py
+++ b/lib/paymentrequest.py
@@ -67,31 +67,48 @@ PR_PAID    = 3     # send and propagated
 def get_payment_request(url):
     u = urlparse.urlparse(url)
     if u.scheme in ['http', 'https']:
-        response = requests.request('GET', url, headers=REQUEST_HEADERS)
-        data = response.content
-        print_error('fetched payment request', url, len(data))
+        try:
+            response = requests.request('GET', url, headers=REQUEST_HEADERS)
+            response.raise_for_status()
+            # Guard against `bitcoin:`-URIs with invalid payment request URLs
+            if "Content-Type" not in response.headers \
+            or response.headers["Content-Type"] != "application/bitcoin-paymentrequest":
+                data = None
+                error = "payment URL not pointing to a payment request handling server"
+            else:
+                data = response.content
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks. why is the if...else statement instide the try...except block?

alexander255:
That's the original implementation (the `try`…`catch` got added later), when I realized that it's not the only way the request can fail.
Basically it checks whether the server actually understood the request or just blurted out some default response (as Anti-DDoS-sites, captive portals, badly configured webservers in general tend to do).
Unless I'm mistaken this behaviour is also required by [`bip-0071`](https://github.com/bitcoin/bips/blob/d2a8fb089d7010d839530f879e07edc4e48b84fa/bip-0071.mediawiki):

> A **web server generating a PaymentRequest** message to initiate the payment protocol **would precede the binary message data with the following headers**:
> 
>     Content-Type: application/bitcoin-paymentrequest
>     Content-Transfer-Encoding: binary

ecdsa:
thank you, I merged it.
I think you did not understand my question, though.

alexander255:
Now that you mention it 😛 
Anyway I just though about it again and putting the `if`…`else` inside the `try`…`catch` makes sense since it makes sure that `response` is always available and represents a successful HTTP response when the `if`…`else` is reached. If the `if`…`else` where to be put after the `try`…`catch` and `url` would fail to resolve entirely (because it was bogus mainly), then `response` would never be set and `electrum` would end up in that "looks like its hanging"-state again due to the second crash when `response` is "used before assignment".

alexander255:
Semi-related: Should I submit a patch that converts the current `raise BaseException("unknown scheme")` to an UI displayable error as well?

Issue #2091, opened 2017-01-03
Windows was Identified as linux!!


7sean68:
I have git bash installed on windows ... when I started electrum it used the bash "~/.electrum" instead of "%appdata%\Electrum" which is of course not a safe place for wallets

the bug is in "/lib/util.py#L246" should check `os.name` instead
because windows may have "HOME" environmental variable

7sean68:
the bug is here [https://github.com/spesmilo/electrum/blob/a3b4409e6e8d1c510f53494ca068633ce03f0cc7/lib/util.py#L246](url)

PR #2092, opened 2017-01-03
changed false Identification of windows as linux

7sean68:
I fixed this issue [link](https://github.com/spesmilo/electrum/issues/2091)


7sean68:
ok I'll solve it

Issue #2128, opened 2017-01-21
Validating checksum when restoring from BIP39 seed


kacperzuk:
Currently when "BIP39 seed" option is selected in seed dialog, any seed will be accepted and "Seed type:" info will be hidden.

I would like to change the messages shown to user:

* leave them as they're now when "BIP39 seed" is not selected (so "Seed type: old/standard/2fa")
* show "Seed type: BIP39 (set in options)" when BIP39 checksum is valid and BIP39 was selected in options
* show "Seed type: BIP39 (set in options; unknown wordlist!)" when there are words outside of bundled wordlist and BIP39 was selected in options
* show "Seed type: BIP39 (set in options; invalid checksum!)" when words are from wordlist, but BIP39 checksum is invalid and BIP39 was selected in options

I understand problems with BIP39 (no versioning, checksum based on numbers from worldlist, multiple wordlists for different languages), but UX really suffer if Electrum doesn't warn about invalid checksum. I almost got heart attack when I kept entering seed with a typo and Electrum didn't say anything to me :).

However it will require bundling BIP39 wordlist with Electrum for as long as restoring from BIP39 will be supported and as I understand we can't rely that current wordlist used by Electrum will always be the same as BIP39.

So my question is: would you merge a PR that implements this or is the wordlist issue a deal breaker (I'll totally understand that)?

ecdsa:
We already use the English wordlist, so it's possible to checksum of English seeds without wordlist additions; I did not do it because I lacked time. I would merge it, but I do not guarantee that BIP39 will forever be supported in Electrum; it will be messy if they extend it to derive segwit addresses.

westonal:
> I almost got heart attack when I kept entering seed with a typo and Electrum didn't say anything to me

:+1:

>  I do not guarantee that BIP39 will forever be supported in Electrum

But you support it now, why can't the support be better? And why does that imply supporting it forever?

dabura667:
> why can't the support be better?

Supporting one checksum check (English only) is unfair. Every wordlist in the future will come up and say "well, you did it for English! Do it for us!" And then eventually, Thomas is stuck making sure checksum checks work for all the wordlists.

Electrum's BIP39 full support was never guaranteed.

westonal:
@dabura667 does #2727 actually cover this, and for all languages.

dabura667:
@westonal not in its current state. Though the author seems to understand what to do.

[...]

PR #2130, opened 2017-01-22
Provide warnings about invalid BIP39 checksum in seed dialog

kacperzuk:
This PR implements what I described in issue #2128 . I've added a small test case to make sure this feature is removed/updated if wordlist bundled with Electrum changes.

Please let me know if you think something should be moved/renamed. I'm not sure keystore.py is a good place for `bip39_is_checksum_valid`, but that's the only place I found with other bip39 related functions.


dabura667:
This will break non-English phrases.

It assumes all wordlists are equivalent to BIP39 counterparts, which is not the case for all.

kacperzuk:
Actually it only assumes that English wordlist is equivalent to it's BIP39 counterpart. If one of the words in phrase is not from English wordlist, it will show this warning: https://github.com/spesmilo/electrum/pull/2130/files#diff-a3125ab4a998d214198f55dfc03353d0R156 .

It only checks with English wordlist, not all of them. Also it doesn't prevent using phrase with invalid checksum/wordlist, it only shows warnings.

dabura667:
> Electrum only recognizes English wordlist for BIP39.

This is a lie though. I have recovered my Japanese BIP39 seed before with no problem.

ecdsa:
merged manually, after reworking

PR #2132, opened 2017-01-22
[WIP] Trezor: Matrix recovery support

jhoenicke:
**This is work in progress so don't merge it yet**
I want to get feedback and maybe some tips to get this finished.  So comments are welcome.

New Trezor firmware has matrix recovery support, which is a new recovery method that doesn't leak the entered words.  Requires self-compiled firmware and latest version of python-trezor from github.

TODO:
- Is there an easy way to make the dialog less wide?
- Close the dialog when finished (must be done by caller of recovery_device)
- Cancel sometimes crashes Electrum with a segfault!!
- Cancel doesn't work as expected, since electrum will still want to create the wallet
- There is this additional warning dialog, that should only really be displayed for old devices with scrambled word order recovery.  Also shouldn't the user be required to acknowledge it first, before starting recovery?
- If seed is entered wrong, Electrum crashes (I think this problem existed before).
- Make sure this doesn't break old python-trezor libraries and/or keepkey

### DIFF
--- a/plugins/trezor/clientbase.py
+++ b/plugins/trezor/clientbase.py
@@ -69,6 +69,15 @@ class GuiMixin(object):
         return self.proto.PassphraseAck(passphrase=passphrase)
 
     def callback_WordRequest(self, msg):
+        if (msg.type is not None
+            and msg.type in (self.types.WordRequestType_Matrix9,
+                             self.types.WordRequestType_Matrix6)):
+            num = 9 if msg.type == self.types.WordRequestType_Matrix9 else 6
+            char = self.handler.get_matrix(num)
+            if (char == 'x'):
+                return self.proto.Cancel()
+            return self.proto.WordAck(word=char)
+
         self.step += 1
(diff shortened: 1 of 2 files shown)


ecdsa:
@prusnak @slush0 can you have a look please?

slush0:
Hi @jhoenicke, @ecdsa,

I completely missed this pull request before, my bad! I set up testing environment with this branch and old python-trezor, but I didn't find a way how to invoke the matrix recovery workflow. Can you give me any hint?

jhoenicke:
@slush0  You need a wiped Trezor and then setup a new wallet (File->New Wallet->pick file name -> Standard Wallet->Hardware Wallet).  When you choose the wiped Trezor it asks how to set it up.

Since this path is not very well tested, it's possible that it doesn't run out of the box, though.

SomberNight:
I have decided to finish this :) see https://github.com/spesmilo/electrum/pull/4329

PR #2135, opened 2017-01-25
Ledger : Testnet support, APDU verbosity aligned with Electrum, Nano S / Blue output streaming

btchip:
Output stream requires btchip-python version 0.1.19 and BTC application 1.1.4+ (currently experimental version available at https://github.com/ledgerhq/blue-app-btc on the stream-output branch)

### DIFF
--- a/plugins/ledger/ledger.py
+++ b/plugins/ledger/ledger.py
@@ -6,12 +6,12 @@ import sys
 import traceback
 
 import electrum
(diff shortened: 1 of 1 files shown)


btchip:
@ecdsa related to that, I noticed that if several outputs have the same amount on the CSV, they end up grouped, which might be a bit confusing for the user confirming them on the hardware wallet. Is this something you'd be interested in fixing or getting a PR for ?

Issue #2160, opened 2017-02-10
Proposal: option to auto-select only from Tor nodes


leafcutterant:
I'd like to propose a setting through which the auto-selection nodes would be limited to those on the Tor network (.onion nodes). It would help users who use Electrum through Tor and want to utilize the full anonymity potential (hopping through 6 Tor nodes instead of just 3), but don't want to be bothered by .onion nodes that are down, or want higher anonymity via auto-selection from .onion nodes every time the wallet starts (because if one server is chosen manually, it will always try to connect to that.)

What do you think?

bauerj:
Sounds like a good idea.

But what should happen if there were only e.g. 7 .onion-nodes? Electrum connects to 8 servers by default to get block headers. Should this be a "prefer .onion-nodes" option?

leafcutterant:
In my opinion, Electrum should go on with fewer nodes in that case. I think those who value anonymity so much that they want to use only .onion nodes would be willing to cope with network lag.

On the the other hand, connecting to very few nodes makes it easier to serve fake data to the client. Next to the setting, there could be a fall-back setting, e.g. "Include at most ▯ non-Tor nodes", where ▯ could be changed

*  from 0 (don't connect to any nodes if no .onion nodes are available, and go on with .onion nodes even if fewer than 8 are reachable)
* to 8 (this would be the "prefer .onion-nodes" equivalent, where it would connect only to clearnet nodes if no .onion-nodes are reachable).

ahmarlen:
I'd like to express my support for this option. Whenever I connect to a .onion server, I'm doing it because I'd prefer if Electrum only used .onion servers for privacy and I'm sure other users like me feel the same way.

ecdsa:
for the moment I fail to see how this option would improve privacy.
could someone please provide more details?

btw, it would be useful to remove .onion servers from the list when the user is not using tor.

leafcutterant:
It doesn't improve privacy _directly_--it makes using only Tor nodes more convenient.

When you use Tor, you can achieve a higher degree of certainty in your anonymity if you connect to a .onion domain (a.k.a. "hidden service"), which entails going through 6 nodes in the Tor network, as opposed to connecting to a clearnet domain, which entails going through 3 nodes.

[...]

PR #2191, opened 2017-02-23
When connected to a Tor proxy, only use onion-servers

bauerj:
This commit:
- adds .onion servers to the default servers
- moves `is_tor_port` to the `Network` class
- Changes `get_servers` to only return `.onion-servers` when connected to a Tor proxy
- Makes the network thread disconnect all clearnet servers when the proxy is switched to a Tor proxy

I'm not if it would be better to use clearnet servers if there a not enough onion-servers available. Any thoughts on that?

Closes: #2160


bauerj:
This doesn't seem to add any security or privacy.

PR #2201, opened 2017-02-24
[Python3] Make proxies work

bauerj:
- `socket` doesn't have `_socketobject` any more, so we have to create that

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -40,7 +40,7 @@ from threading import Lock
 import socket
 import json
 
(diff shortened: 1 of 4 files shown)


lib/network.py
```diff
@@ -40,7 +40,7 @@
 import socket
 import json
 
-from . import socks
+import socks
```
bauerj: This will only work when #2207 is merged in python3.

PR #2206, opened 2017-02-25
[Python3] Various fixes in util.py

bauerj:


lib/util.py
```diff
@@ -590,7 +591,7 @@ def create_URI(addr, amount, message):
     if amount:
         query.append('amount=%s'%format_satoshis_plain(amount))
     if message:
-        if type(message) == unicode:
+        if six.PY2 and type(message) == unicode:
```
bauerj: `unicode` is only defined in Python 2, so make sure it doesn't get evaluated in Python 3.

lib/util.py
```diff
@@ -715,14 +716,14 @@ def _send(self, out):
 class QueuePipe:
 
     def __init__(self, send_queue=None, get_queue=None):
-        self.send_queue = send_queue if send_queue else Queue.Queue()
-        self.get_queue = get_queue if get_queue else Queue.Queue()
+        self.send_queue = send_queue if send_queue else queue.Queue()
```
bauerj: Since `queue` is imported as `queue`.

lib/util.py
```diff
@@ -796,4 +797,4 @@ def check_www_dir(rdir):
         path = os.path.join(rdir, filename)
         if not os.path.exists(path):
             print_error("downloading ", URL)
-            urllib.urlretrieve(URL, path)
+            urllib.request.urlretrieve(URL, path)
```
bauerj: https://docs.python.org/3.0/library/urllib.request.html#urllib.request.urlretrieve

lib/util.py
```diff
@@ -27,14 +27,15 @@
...
-import platform
-import shutil
 from collections import defaultdict
 from datetime import datetime
 from decimal import Decimal
 import traceback
 import urllib
 import threading
+
+import decimal
```
bauerj: Used in line 203.

PR #2236, opened 2017-03-04
Add support for BU's nolnet

kyuupichan:
Also update testnet servers
Add missing self. in print_error call

### DIFF
--- a/electrum
+++ b/electrum
@@ -361,6 +361,10 @@ if __name__ == '__main__':
         bitcoin.set_testnet()
         network.set_testnet()
 
+    if config.get('nolnet'):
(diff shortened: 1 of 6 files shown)


HostFat:
It can be cool to have this before the next upcoming release :)

ecdsa:
Why is this relevant to Electrum? Last time I looked, BU was only about the block size limit, which Electrum is not aware of. Does BU introduce other features, that would require modifications of the wallet software?

kyuupichan:
If it required other changes I'd have included them. This is all that's needed. It is useful for our testing on nolnet.

PR #2239, opened 2017-03-04
fix #2204

### LINKED ISSUE #2204: Display problem: Stuck in Synchronizing after changing network socks settings and connecting to another server
2.7.18 on windows appears to have a display problem, regardless it is fully synced it will print Synchronizing forever. The console command network.unanswered_requests will return nothing.

To reproduce, just connect to a server (on TCP or SSL, doesn't matter) directly, and wait until fully synced so that it shows green ball. Then, change at network settings to use a socks5 address (in my case I entered socks5:localhost:9150 and tried to connect to the same server, but on the onion Tor hostname). It will only show Synchronizing in this case. The same happens if you try to connect to the regular server name (not .onion) but after changing the socks5 network settings. Closing and reopening the wallet fixes it, and it will show green while connected via socks5 proxy.

Another thing is that if you open the network settings after it is in Synchronizing mode and change the server after the 2nd change it won't connect any more and show red (didn't wait for more than 5 minutes to see if it actually will remain like this but even 5 minutes seams very much for this). Does it make sense to add a button to connect/disconnect on demand, or call connect every time the network settings wizard closes?

shsmith:


lib/network.py
```diff
@@ -669,7 +671,8 @@ def maintain_sockets(self):
         # Responses to connection attempts?
         while not self.socket_queue.empty():
             server, socket = self.socket_queue.get()
-            self.connecting.remove(server)
+            if server in self.connecting:
+                self.connecting.remove(server)
             if socket:
```
kyuupichan: Just FYI - if you're not sure it's in the set, this can be done more simply with discard:

self.connecting.discard(server)

shsmith: I didn't know that. So discard is like TryRemove in some other languages. Thanks.

Issue #2009, opened 2016-10-27
FEATURE REQUEST: Console error logging in Electrum for Windows


cajuntechie:
Currently, there is no way to obtain debug information when running Electrum on Microsoft Windows. This presents a unique problem for Windows users who are experiencing problems with Electrum as they are unable to provide the development team with any useful information.

Implementing console error logging would easily resolve this.

PR #2265, opened 2017-03-09
Print exceptions to the built-in console

bauerj:
Closes: #2009

### DIFF
--- a/gui/qt/console.py
+++ b/gui/qt/console.py
@@ -32,6 +32,7 @@ class Console(QtGui.QPlainTextEdit):
 
         self.updateNamespace({'run':self.run_script})
         self.set_json(False)
+        self.hook_exceptions()
 
     def set_json(self, b):
         self.is_json = b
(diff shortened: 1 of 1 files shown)


ecdsa:
that's not going to prevent the gui from crashing.
I'll try to find another solution

bauerj:
I don't understand. How should it prevent the GUI from crashing?

ecdsa:
if there is a crash that crashes the gui, you will not see it in the console with that hook. you will just keep crashing. I think maybe the whole gui execution should be in a try..except block, and SimpleConfig instanciation should be deferred to the GUI.

bauerj:
Indeed, it also doesn't log any exceptions that occur before the GUI is opened.

However, I don't see how catching the exception outside of the GUI would solve that.

ecdsa:
well, that means we would not print the exception in the console, but we would open a popup window after the gui has crashed. maybe it can be done by sys.excepthook, not sure.

bauerj:
I'm not sure if e.g. exceptions in background threads can be caught when initializing the GUI and those should probably be logged too.

PR #2275, opened 2017-03-10
Digital Bitbox cannot send coins bug (due to a later upstream change)

douglasbakkum:
Looks like a later commit changed something and caused a bug in the Digital Bitbox plugin. This PR fixes it.

I would classify the problem as urgent as users will not be able to send coins with a Digital Bitbox when using the apps available for download on the website.

For information, the error message is `KeyError: 'pubkeys'`.

Cheers.

### DIFF
--- a/plugins/digitalbitbox/digitalbitbox.py
+++ b/plugins/digitalbitbox/digitalbitbox.py
@@ -367,7 +367,7 @@ class DigitalBitbox_KeyStore(Hardware_KeyStore):
                 if txin.get('is_coinbase'):
                     self.give_error("Coinbase not supported") # should never happen
                 
-                if len(txin['pubkeys']) > 1:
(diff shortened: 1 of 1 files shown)


ecdsa:
sorry about that.
I will create a new release.

douglasbakkum:
No worries. Thanks.

jonasschnelli:
A little glitch:
In the release, the Digital Bitbox icon does not show up.

I think it's a packaging problem. The icon is in the qrc file and it works on local builds.
https://github.com/spesmilo/electrum/blob/944c54c0301568e7b1cb341d026dc36c529568f1/icons.qrc

I guess the .qrc needs to get compiled again.

bildschirmfoto 2017-03-10 um 13 13 54

ecdsa:
oh ok, that's OSX

PR #2320, opened 2017-03-21
Tighten directory permissions

da2x:
Not using `mode` on `mkdir()` as `chmod()` works better under Windows.


ecdsa:
why do you add a f.close() inside with?
also, you added a blank line

da2x:
> why do you add a f.close() inside with?
> also, you added a blank line

Resolved. Sorry about that.

PR #2339, opened 2017-03-28
Semi-automated crash reporting

bauerj:
This adds an exception hook to the Qt GUI that shows an error window like this:

python_2017-11-01_20-45-25
python_2017-11-01_20-45-40
python_2017-11-01_20-46-00

* Default exception handling will be delayed until the window is closed (so the application doesn't terminate for exceptions in the main window)
* Includes a button that sends a crash report to a server that eventually creates a Github issue. There will only be one Github issue per `(file, method, exception_type)` combination

Closes: #2009

### DIFF
--- a/gui/qt/exception_window.py
+++ b/gui/qt/exception_window.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python
+#
+# Electrum - lightweight Bitcoin client
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+# BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+import json
+import locale
+import platform
+import traceback
+
+import requests
+from PyQt5.QtCore import QObject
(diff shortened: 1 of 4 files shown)


gui/qt/exception_window.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
+#
+# Electrum - lightweight Bitcoin client
```
bauerj: I copied this from `qrwindow.py`. Not sure about the copyright notice that was there. Should it include your name or my name?

bauerj:
Is there some other information besides the Electrum version and the traceback that would be useful?

shsmith:
Nice! This will help a lot with Windows and OS X users who cannot easily get the verbose console output.

ecdsa:
I only briefly tested it, but it does not seem to catch exceptions other than in the gui thread...

bauerj:
Indeed, good catch.

Apparently it's a bug: http://bugs.python.org/issue1230540

It's easy to patch threading to use the excepthook but I need to find out how to delegate that to the GUI thread in order to show the window.

bauerj:
Okay, so I added some more code:
- `threading.Thread` gets patched to use `excepthook` for all threads
- A signal is used to delegate the window creation to the GUI thread

bauerj:
And the prefilled Github issue now looks like this:

----------------
Traceback
=========
```
  File "C:/Users/bauerj/Documents/GitHub/electrum\lib\util.py", line 666, in run_with_except_hook
    run_original(*args2, **kwargs2)
  File "C:\Python27\lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "C:/Users/bauerj/Documents/GitHub/electrum\plugins\labels\labels.py", line 98, in pull_thread
    1/0
ZeroDivisionError: integer division or modulo by zero
```
Additional information
======================
- Electrum version: 2.8.3
- Operating system: Windows-10-10.0.14393
- Wallet type: standard

ecdsa:
this is deferred, I need to test how it behaves with binaries

lzsaver:
It looks useful. Any progress here?

bauerj:
I just rebased and update this to PyQt5.

ecdsa:
@bauerj I am worried that this is going to make github unusable, because the same traceback can be pushed multiple times. Foe example, Google's error tracking (which we use for Android) categorizes crashes, and does not open one report per user.

bauerj:
Yes, that could be an issue. Would you prefer some kind of middleware that would group tracebacks of the same kind?

ecdsa:
@bauerj how would that work?

bauerj:
It could be a web service that would accept crash reports from clients.

For all crash reports it would check if there already is a combination of exception type, file and method of the last item in the traceback. If there is no such combination, it would open an issue on Github with all the information from the client. If there already is such a combination it would update the existing issue on Github.

AFAIK there is no such service avaiable but it should be easy enough to set up.

bauerj:
I'm working on such a service on [bauerj/crashhub](https://github.com/bauerj/crashhub).

It currently creates issues like bauerj/beam#6

bauerj:
This should be ready now. It would be nice if someone could test it and express their thoughts about the process.

bauerj:
I tested the following:
 - [x] Windows binaries
 - [x] exceptions in PyQt callbacks
 - [x] exceptions in the main thread
 - [x] exceptions in a background thread (e.g. network thread)
 - [x] exceptions in Qt thread (e.g. `TorDetector`)

PR #2358, opened 2017-04-10
wine pip install win_inet_pton

romanornr:
I was working on Viacoin Electrum (which I forked from the Litecoin repo)
I noticed I couldn't build the Windows binaries correctly & started a bit of debugging and noticed this problem.

screenshot from 2017-04-09 21-22-03

h_1491764812_1306184_8be550b594

I figured it was missing wine win_inet_pton
Which should be added to the prepare-wine.sh script.

**I discovered this at Litecoin port (see link https://github.com/pooler/electrum-ltc/pull/44 )
And thought this may effected Electrum as well**

### DIFF
--- a/contrib/build-wine/build-electrum-git.sh
+++ b/contrib/build-wine/build-electrum-git.sh
@@ -74,4 +74,6 @@ patch < portable.patch
 popd
 $PYTHON "C:/pyinstaller/pyinstaller.py" --noconfirm --ascii --name $NAME_ROOT-$VERSION-portable.exe -w deterministic.spec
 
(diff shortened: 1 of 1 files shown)


bauerj:
Thanks! I can reproduce this. I wonder why it didn't show up when I first tested PySocks 😕

ecdsa:
thank you. I remember I had to install it manually. this command should not go in build-electrum-git.sh, because that script is run everytime you build. it should probably go in prepare-wine.sh instead

ecdsa:
done in c5d3adcf32c5148c9890dbcb68bfab663e554b7c

PR #2359, opened 2017-04-12
Major merchant UI reshuffle

ser:
* massive redesign of user interface, including easy to understand progress bar
* jquery-ui swapped with bootstrap
* added font awesome icons, mostly for for bitcoin currency

### DIFF
--- a/lib/util.py
+++ b/lib/util.py
@@ -632,10 +632,19 @@ def check_www_dir(rdir):
         src = os.path.join(os.path.dirname(__file__), 'www', 'index.html')
         shutil.copy(src, index)
     files = [
-        "https://code.jquery.com/jquery-1.9.1.min.js",
-        "https://raw.githubusercontent.com/davidshimjs/qrcodejs/master/qrcode.js",
-        "https://code.jquery.com/ui/1.10.3/jquery-ui.js",
-        "https://code.jquery.com/ui/1.10.3/themes/smoothness/jquery-ui.css"
+        "https://code.jquery.com/jquery-3.2.1.min.js"
+        "https://raw.githubusercontent.com/neocotic/qrious/master/dist/umd/qrious.min.js",
+        "https://raw.githubusercontent.com/minddust/bootstrap-progressbar/master/bootstrap-progressbar.min.js"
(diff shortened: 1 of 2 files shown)


Plasmmer:
Can you build it for download?

ecdsa:
I don't like to add so many new javascript dependency links.
it increases the probability of any of them being broken.
do we really need them?

ser:
Hi, unfortunately it's not possible to eat a cake and have a cake. These libraries make payment page so pretty :)

lib/util.py
```diff
@@ -632,10 +632,19 @@ def check_www_dir(rdir):
         src = os.path.join(os.path.dirname(__file__), 'www', 'index.html')
         shutil.copy(src, index)
     files = [
-        "https://code.jquery.com/jquery-1.9.1.min.js",
-        "https://raw.githubusercontent.com/davidshimjs/qrcodejs/master/qrcode.js",
-        "https://code.jquery.com/ui/1.10.3/jquery-ui.js",
-        "https://code.jquery.com/ui/1.10.3/themes/smoothness/jquery-ui.css"
+        "https://code.jquery.com/jquery-3.2.1.min.js"
+        "https://raw.githubusercontent.com/neocotic/qrious/master/dist/umd/qrious.min.js",
```
flack: This returns a 404 error. Also, AFAIK, using `raw.githubusercontent.com` in production is discouraged, because it's not very efficient. Also, they send the files with `text/plain` mime type, which will cause some browsers to not execute them as JS

ser: It's not needed to execute these files, obviously, they are downloaded just once. 

IMHO it is needed to include copies of these files into Electrum repository to be sure they are correctly distributed. 

What do you think about it, guys?

ecdsa: I would prefer this to be hosted in a separate project. will look into how to do this.

ser: Good idea, I definitely can co-work on this project, as I use it every day. Let me know what have you decided.

ecdsa:
done in 4a40b3a0b2522e6adad58631eba4f57341907d44
please see https://github.com/spesmilo/electrum-merchant

PR #2383, opened 2017-05-03
Update README.rst

flexiondotorg:
Electrum (my favourite Bitcoin wallet) is available in the snap store! 

This pull request adds installation instructions for users of [snap enabled Linux distributions](https://snapcraft.io/docs/core/install). Snaps enable ISVs and projects to make their software available to millions of Linux users via the snap store and directly control delivery of software updates to their users. Snaps of stable desktop applications will also be available via the desktop Software Center, improving the discoverability of your software. Your software can be installed with `snap install electrum` simplifying the installation instructions for your users.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,8 +20,22 @@ Electrum - Lightweight Bitcoin client
 Getting started
 ===============
 
-Electrum is a pure python application. If you want to use the
-Qt interface, install the Qt dependencies::
+Electrum is a pure python application. Electrum is available for
+`Ubuntu and other snap supported Linux distributions
+<https://snapcraft.io/docs/core/install>`_ in the snap store and
+can be simply installed with:
+
+    snap install electrum
+
+Installing a snap is very quick. Snaps are secure. They are
+isolated with all of their dependencies. Snaps also auto update
+when a new version is released.
+
+Tarball
+-------
+
+If you want to use the Qt interface, install the Qt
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -20,8 +20,22 @@ Electrum - Lightweight Bitcoin client
 Getting started
 ===============
 
-Electrum is a pure python application. If you want to use the
-Qt interface, install the Qt dependencies::
+Electrum is a pure python application. Electrum is available for
+`Ubuntu and other snap supported Linux distributions
+<https://snapcraft.io/docs/core/install>`_ in the snap store and
+can be simply installed with:
```
bauerj: I think you need to end the line with :: in order to make the next line a code block.

ecdsa:
this PR looks like a shameless plug for snap.
it does not add any useful info about electrum itself.
why should I merge it?

bauerj:
Maybe it makes more sense to add it here? https://github.com/ecdsa/electrum-web/blob/gh-pages/download.html.template

black-puppydog:
I would have actually made a similar PR were it not for this one. I would agree with @flexiondotorg that the possibility of installing via snap should be mentioned *somewhere* where potential users can find it. The website might indeed be a good idea, but personally, I stumbled over the github repo first, so for me it would have been useful there.

PR #2390, opened 2017-05-06
Allow comma separated address lists for from_addr

### LINKED ISSUE #2389: electrum transaction from many address to many address
With paytomany, I can use only one input address. I want to use more input address. Is there any way into Electrum?

I think, serialize can make this, but I don't understand to use serialize.

Help please..

dabura667:
closes #2389


ecdsa:
I merged it manually, see d28f603
I do not like it, though, because that format is not json.

Issue #1685, opened 2016-02-22
Adjust nLockTime (custom locktime)


ulrichard:
As an advanced feature (hidden from regular use), it would be cool if the nLockTime could be adjusted.
The broadcast button should probably stay disabled for such transactions.

Rspigler:
I was just about to write my own Issue under this same principle in order to enable Electrum users to activate what is often referred to as (and often requested for) a Dead Man's Switch.  I could be wrong, but all that would really be required is a GUI for nLockTime (possibly only enable this under advance settings, as suggested by @ulrichard)

As far as I can tell, Electrum would be the first wallet to enable this feature.

Rspigler:
See (https://github.com/spesmilo/electrum/issues/2401) for updated issue/feature request with bounty attached.

ysangkok:
@Rspigler Does the bounty apply to this issue too?

Rspigler:
@ysangkok, my apologies for the late response.

This issue should probably be closed; @bauerj wrote a great implementation in PR https://github.com/spesmilo/electrum/pull/2405. While this wasn't merged due to @ecdsa believing that the use case wasn't common enough to confuse users with, the work has already been done, and the payment has been sent.  

This can be confirmed here:  https://btc.blockr.io/tx/info/2f9e763141efaa405dccfa4d5b9ea8ec5c9e5eae02d081917e3f20df67a2e05b.

botweb:
I would love to have this feature. @ecdsa Wouldn't it be an option to allow such advanced features through a new tab in settings? That way power users and newbies get served.

Issue #2401, opened 2017-05-11
[$200 bounty in bitcoin] - Dead Man's Switch


Rspigler:
AFAIK there is currently no bitcoin wallet that allows easy adjustment of the nLockTime field.  This would enable a so called 'Dead Man's Switch', where transactions could be created and signed, but not broadcasted.  This transaction could then be given to loved ones, and the broadcasted transaction would then only be valid and accepted into a block at some absolute time in the future (block x).  Before that point in time, you moving your bitcoin to another address in your control permanently invalidates the transaction, and a new nLockTime signed transaction can be passed along in case something happens to you in whatever period of time you choose (1 month, 6 months, etc).

This is a greatly requested feature by the community (just google "Bitcoin Dead Man's Switch" to find many posts), but so far has no easy GUI available for this in any wallet.

This is a copy of (https://github.com/spesmilo/electrum/issues/1685), but with an extended explanation and bounty attached.

I may not be very well known in this community, but this GitHub account is associated with my main/personal PGP key, email address, real name, and photo.  However, I of course would be willing to set up an escrowed payment for those more comfortable with that.

My preferences are that the nLockTime is easily adjustable through the GUI, and that this would work with all current types of addresses/wallets already supported by Electrum (standard, multisignature, standard hardware wallet, and multisignature wallet with hardware wallet co-signers).

ecdsa:
please don't post duplicates; you're just adding to the noise

PR #2405, opened 2017-05-14
Allow users to specify lock_time for transactions

bauerj:
To do:
====
* [X] Add date-time picker to Send tab
* [X] Add timestamp to the transaction
* [x] Set sequence to `MAX-1` in order to enable the lock
* [x] Clarify the "lock time" entry in the preview window (Add "Block" for blocks and readable time for time)
* [x] Disable broadcasting if the lock time is in the future

Anything else?

Screenshots:
====
python_2017-05-14_20-23-40
python_2017-05-14_20-23-44
python_2017-05-14_22-15-19
python_2017-05-14_22-15-26

Closes: #1685
See: https://en.bitcoin.it/wiki/Protocol_documentation#tx

cc: @ulrichard @Rspigler

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -1051,6 +1051,18 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         grid.addWidget(self.fee_e, 5, 2)
         grid.addWidget(self.rbf_checkbox, 5, 3)
 
+        msg = _('The transaction will only be valid after this time (in your local timezone).')
+        locktime_label = HelpLabel(_('Valid from'), msg)
+        grid.addWidget(locktime_label, 6, 0)
+        self.locktime_disable = QCheckBox(_("Immediately"))
+        self.locktime_disable.setChecked(True)
+        self.locktime_disable.stateChanged.connect(self.toggle_locktime)
+        grid.addWidget(self.locktime_disable, 6, 1)
+        self.locktime_date = QDateTimeEdit()
+        self.locktime_date.setDateTime(QDateTime.currentDateTime())
+        self.locktime_date.setVisible(not self.locktime_disable.isChecked())
+        grid.addWidget(self.locktime_date, 6, 2)
+
         self.preview_button = EnterButton(_("Preview"), self.do_preview)
         self.preview_button.setToolTip(_('Display the details of your transactions before signing it.'))
         self.send_button = EnterButton(_("Send"), self.do_send)
@@ -1060,7 +1072,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         buttons.addWidget(self.clear_button)
         buttons.addWidget(self.preview_button)
         buttons.addWidget(self.send_button)
-        grid.addLayout(buttons, 6, 1, 1, 3)
+        grid.addLayout(buttons, 7, 1, 1, 3)
 
         self.amount_e.shortcut.connect(self.spend_max)
         self.payto_e.textChanged.connect(self.update_fee)
@@ -1120,6 +1132,9 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
 
     def update_fee(self):
         self.require_fee_update = True
+
+    def toggle_locktime(self, off):
+        self.locktime_date.setVisible(not off)
 
     def get_payto_or_dummy(self):
         r = self.payto_e.get_recipient()
@@ -1281,7 +1296,13 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         freeze_fee = self.fee_e.isVisible() and self.fee_e.isModified() and (self.fee_e.text() or self.fee_e.hasFocus())
         fee = self.fee_e.get_amount() if freeze_fee else None
         coins = self.get_coins()
-        return outputs, fee, label, coins
+
+        locktime = None
+        if not self.locktime_disable.isChecked():
+            locktime = self.locktime_date.dateTime().toPyDateTime()
+            locktime = int(time.mktime(locktime.timetuple()))
+
+        return outputs, fee, label, coins, locktime
 
     def do_preview(self):
         self.do_send(preview = True)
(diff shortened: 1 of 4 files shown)


bauerj:
Regarding the need to enable RBF: I was able to broadcast a [timelocked transaction](http://tbtc.blockr.io/tx/info/e6e9367f45240b273df7c361290a3e83fe43311efbd29771591537674f57bcd9) with RBF disabled but unable to do so with RBF enabled. So it looks like the documentation is right on this one.

ecdsa:
I don't understand; do you need to enable or to disable RBF?

bauerj:
For a timelocked transaction that's not valid yet:

- with RBF disabled: accepted by the network
- with RBF enabled: rejected by the network

So RBF needs to be enabled for the lock to work.

ecdsa:
I used your locktime fix in d4e8f4beb630d46cdc196df11dcf57540f0bec71

I am not merging the rest of this PR, because I think this should not be part of the wallet, as long as the wallet does not have unbroadcasted transaction in its history. As it is now, the wallet will create transactions that are immediately double spent the next time you use the wallet. That is too confusing.

I plan to add support for unbroadcasted transactions, though.

dabura667:
You don't need to activate RBF.

The nSequence needs to be < 0xffffffff and RBF flag is < 0xfeffffff (little endian) so just make the nSequence be (2**32)-2 ( == 0xfeffffff)

bauerj:
Yes, that's what I did in bauerj/electrum@4157fc2976a5a84d981c602143c19c650314d947

Rspigler:
Hello,

I apologize for my late response.  @bauerj , as you likely saw in (https://github.com/spesmilo/electrum/issues/2401), I had placed a $200 bounty on this feature.  Although @ecdsa did not pull the code, it seems you did do all the work necessary on a very nice implementation, and that there are plans to possibly allow this feature in the future.

Can you either:

A)  Post and sign a bitcoin address you would like me to send the funds to (so that the community may verify that I have kept my word), or

B)  If you would like to keep this address private, please send the address to my email, and encrypted with my public key.  Once you have received a satisfactory number of confirmations, please post a signed message here stating that the transaction has been completed, so once again the community may confirm.

Thank you,

Robert

bauerj:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

I don't think this will be merged. The reasoning was that this use-case isn't common enough to confuse users with this option.

However, if you still want to send me the coins, you can send them to 13U9U1wzUnJHbztsnkZiXdi1Qi65DtDLUV which is also the address associated with my keybase account: https://keybase.io/bauerj
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJZMTlpAAoJEITxv5JbH0hNZ/oH/jtbNKgzVWs2TJ0spMs0Mmbv
oSR+DQ4D6xFWpa+GHphr1CqAcW8/eYX4ybc5EDgEX8ocOnmHh7e8GxrloKuIpAq8
NgkAhW6FF5pGtFhzDruSp+L9E9/C8nBqTnJ+4pO5sLXadHAAz+3bMn9PUmyEJOqs
OEm3ZyZEC4QMlPm1DawGx/ZZB+zierwiWVWHDXkfYAoRAUW/K83RgXZLPKw36MiI
63HHjdSJIaocNYPBFEReFREFbCAQ3ljXMs6PsGOMALYHteVvDWYWs3S29gImmN0F
ZTZs73NvGr5D5lBBbwTIii6RDgtkkKWoMLZ5GCEvkEyLNWLQz8ojhlv03FB73/k=
=gp/G
-----END PGP SIGNATURE-----

Rspigler:
Hi @bauerj, I believe you've earned the full bounty regardless, but I do greatly appreciate your candor.

The transaction can be seen here:  https://btc.blockr.io/tx/info/2f9e763141efaa405dccfa4d5b9ea8ec5c9e5eae02d081917e3f20df67a2e05b

Robert

Robot1982:
> I used your locktime fix in [d4e8f4b](https://github.com/spesmilo/electrum/commit/d4e8f4beb630d46cdc196df11dcf57540f0bec71)
> 
> I am not merging the rest of this PR, because I think this should not be part of the wallet, as long as the wallet does not have unbroadcasted transaction in its history. As it is now, the wallet will create transactions that are immediately double spent the next time you use the wallet. That is too confusing.
> 
> I plan to add support for unbroadcasted transactions, though.

How about e726752 (Clarify the "lock time" entry in the preview window). This is useful when you load time locked transactions and you want to sign them with electrum (cold storage). It confirms that the time lock you want is correct (especially if it is a date (> 500000000)). This is useful even if the rest of the PR is not being merged.

ecdsa:
note: Electrum now supports local transactions, so maybe we should reopen this

roth-a:
Pro:
I like this very advanced option, and a nice to have feature. However I think there should be some feature test with testnet or regtest, to make sure the code actually works as intended. Reasoning:
It is harder to check if the network accepts the transactio,  than without locktime, because it is dependent on time or blockheight.


Con:
People, believing there is no immediate risk, store this transaction on Google drive. They of course forget the due date, and when they check they discover that the funds are gone.

Robot1982:
> Con:
> It could create a false sense of security. Example:
> People, believing there is no immediate risk, store this transaction on Google drive. They of course forget the due date, and when they check they discover that the funds are gone.

Maybe this should be enabled only if the "Advanced preview" is enabled (new feature in 4.0.0)? By doing so it will not confuse inexperienced users, but it will still be available for the more advanced users. I think this is a good compromise.

SomberNight:
Yes, in the new Preview dialog, just like RBF can be toggled directly in the dialog, the locktime could also be made customisable (it's shown already anyway).
Further, unlike when this PR was opened, the wallet now allows saving "local" transactions into its history. This seems ideal for future-timelocked txns, as now they can be saved into the wallet itself, and further, in this case, the affected UTXOs will not be "mistakenly" double-spent (invalidating the future-timelocked txn).

patchmaster:
I love this feature. For now I have ported the patch to electrum-3.3.8 and it works for legacy addresses. I don't think it works for segwit though. https://pastebin.com/XEf4XLbp
I would love the option to generate also tx based on the timelocked utxo, knowing that the new tx would be accepted by the network only after a successful acceptance of the timelocked one.

Issue #2377, opened 2017-05-01
Old version translation?


tomas06:
Hi,

488 <-> 501
The crowdin.com are translations with 488 rows.
When I use the development version and "./contrib/make_locale" translations with 501 rows.

What is the template ".pot" update time interval to crowdin.com. Since the release of the new version of electrum?

bauerj:
Hey,

the template on crowdin was last updated 2 months ago. Ideally, that should happen automatically when there are new strings to translate.

PR #2421, opened 2017-05-18
Let Travis upload templates to crowdin

bauerj:
This uploads updated translation strings to crowdin on every push to master.

You need to [add the API key] as a [Travis environment variable] in order for this to work.

As long as you don't merge something that prints the API key, it should be safe:

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

[Travis environment variable]: https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings
[add the API key]: https://travis-ci.org/spesmilo/electrum/settings
Closes: #2377

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ install:
     - pip install tox
 script:
     - tox
+after_success:
+    - if [ "$TRAVIS_BRANCH" = "master" ]; then pip install pycurl requests && contrib/make_locale; fi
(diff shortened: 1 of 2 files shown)


ecdsa:
I uploaded the key.
we'll need to check if this is not too cpu intensive, though

bauerj:
Huh? Too CPU intensive for Travis CI?

ecdsa:
yes; but I guess it's not worse than running tests

bauerj:
It should be fine. They only impose a limit on runtime which is 50 minutes for free projects.

Issue #2462, opened 2017-05-29
Possible regression in transaction hashing/txid calculation


hexafraction:
The hash of the following:

    0200000001191601a44a81e061502b7bfbc6eaa1cef6d1e6af5308ef96c9342f71dbf4b9b5000000006b483045022100a6d44d0a651790a477e75334adfb8aae94d6612d01187b2c02526e340a7fd6c8022028bdf7a64a54906b13b145cd5dab21a26bd4b85d6044e9b97bceab5be44c2a9201210253e8e0254b0c95776786e40984c1aa32a7d03efa6bdacdea5f421b774917d346feffffff026b20fa04000000001976a914024db2e87dd7cfd0e5f266c5f212e21a31d805a588aca0860100000000001976a91421919b94ae5cefcdf0271191459157cdb41c4cbf88aca6240700

appears to be calculated by Electrum 2.8.2 (using load transaction -> from text) as:

    2b25a612b05ec349a7ba53af5c59c3cb2147832385008168780515bca55c138e

All other tools, including an older version of electrum (2.6.something) report it as 

    b97f9180173ab141b61b9f944d841e60feec691d6daab4d4d932b24dd36606fe

which is a [valid, mined transaction on the chain](https://blockchain.info/tx/b97f9180173ab141b61b9f944d841e60feec691d6daab4d4d932b24dd36606fe).

The bytes I am hashing match the raw bytes the block explorer provides [here](https://blockchain.info/tx/b97f9180173ab141b61b9f944d841e60feec691d6daab4d4d932b24dd36606fe?format=hex).

I've also tested this with a clean .electrum directory (by logging in as a different Unix user with their own home folder and importing the tx in a clean wallet). Hashing other random raw confirmed txns from blockchain.info gives correct hashes.

bauerj:
This seems to be a BIP68 transaction. Currently, Electrum doesn't parse the version field which leads to this error.

PR #2468, opened 2017-05-29
Parse transaction version field

bauerj:
The version bit is not always set to 1.

Closes: #2462


ecdsa:
why is the default value 0 ?

bauerj:
Indeed, I think it should be 1. I'll change that later.

bauerj:
Updated.

PR #2505, opened 2017-06-10
Add plugin hook `init_menubar_tools`

chipmixer:
New plugin hook `init_menubar_tools` that allows plugins to add submenu item in Tools menu.

PR #2506, opened 2017-06-10
ChipMixer's plugin for Electrum

chipmixer:
Plugin's readme is here: https://github.com/chipmixer/electrum/tree/plugin/plugins/chipmixer


chipmixer:
Thank you for reply. Please merge hook (https://github.com/spesmilo/electrum/pull/2505). We will distribute it separately, but also it will be useful for other plugin creators.

PR #2521, opened 2017-06-15
Add the packaging metadata to build the electrum snap

elopio:
This package will let you publish the latest electrum in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,7 +16,7 @@ parts:
   electrum:
     source: .
     plugin: python
-    python-version: python
+    python-version: python2
     stage-packages: [python-qt4]
     build-packages: [pyqt4-dev-tools]
     install: pyrcc4 icons.qrc -o $SNAPCRAFT_PART_INSTALL/lib/python2.7/site-packages/electrum_gui/qt/icons_rc.py
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -10,7 +10,7 @@ confinement: strict
 apps:
   electrum:
     command: desktop-launch electrum
-    plugs: [network, network-bind]
+    plugs: [network, network-bind, x11, unity7]
 
 parts:
   electrum:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,21 @@
+name: electrum
+version: master
+summary: Bitcoin thin client
+description: |
+  Lightweight Bitcoin client
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  electrum:
+    command: desktop-launch electrum
+    plugs: [network, network-bind]
+
+parts:
+  electrum:
+    source: .
+    plugin: python
+    python-version: python2
+    stage-packages: [python-qt4]
+    after: [desktop-qt4]
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,6 +16,8 @@ parts:
   electrum:
     source: .
     plugin: python
-    python-version: python2
+    python-version: python
     stage-packages: [python-qt4]
+    build-packages: [pyqt4-dev-tools]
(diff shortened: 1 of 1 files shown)


elopio:
Hello!

Here at ubuntu we are working a new packaging and delivery system. If you haven't heard about snaps before, here's a lot of info: https://snapcraft.io/ 

I used electrum to test our new snapcraft release, and it worked nicely so here it is for you.
I think it would be great if you push it to the store. After landing this PR, you can go to https://build.snapcraft.io/ and it will take care of the building and continuous delivery of new versions to your users.

To test this in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/electrum
    $ cd electrum
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(`--dangerous` because the snap you build doesn't come signed from the store. Once you publish it, your users will just have to do `sudo snap install electrum`).

If you have any questions, please let me know.

pura vida.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,23 @@
+name: electrum
+version: master
+summary: Bitcoin thin client
+description: |
+  Lightweight Bitcoin client
+
+grade: devel # must be 'stable' to release into candidate/stable channels
```
bauerj: Should this be changed later?

elopio: Just as a safeguard during the initial tests. Once it's move to the stable channel in the store it will be installable by millions of users of Ubuntu who can find it in the software center. I'm usually cautious to make sure that there is a commit that marks when the snap is stable. But this is not strictly necessary.

bauerj: That makes sense. Thank you!

tomascw:
Background. I maintain a unofficial Snap of Electrum. I'm not aware of any Bitcoin wallet that's in any Linux repository. So it might be a first. Normally my packages have a ``-tpaw`` suffix. Shuttleworth convinced me to drop the suffixes so I made another Electrum package without it but I recently regretted that in favor of future wanting to proper official distributors to continue where I left off. I have no idea give up claim on the package name. However, I noticed packages are presented as ``package.maintainer`` on the developer dashboard which makes me question if there would be no conflicts.

Anyway, my experience with maintaining the Snap is pretty straightforward and not that complicated. The setup pulls in the upstream binaries and packages the snap as you would it expect. Nothing overlay complicated. Updating from one release has pretty is simple as replacing the version numbers with a simple find and replace in a editor. It all works as you would expect.

elopio:
Making the transfer is easy. All the users @tomascw has will switch to the official publisher. And you can even make him a contributor on the store, this might be handy because he already has some good experience maintaining snaps.

elopio:
Thanks for the merge @ecdsa!
When you have an account in https://dashboard.snapcraft.io/, let me know the name and I will start the transfer between @tomascw and your team.

ecdsa:
@elopio why? I do not plan to do so

elopio:
@ecdsa so you can automate the delivery, get a new version in the edge channel of the store every time you change master, and use the community of testers in Ubuntu to get early feedback.

Also, to control when you send a stable update to the users of this snap. (more than 1000, with ~10 new installs every day, and growing).

ecdsa:
@elopio I have no time to do that. if merging your PR is not useful, let me know and I will remove it.

elopio:
@ecdsa that's the beauty of it, it's the most low maintenance store that you will find. You just need to click a couple of buttons in build.snapcraft.io.
And if you don't have time to click one checkbox to move it to stable, @tomascw can help there. (I think, haven't asked :))

But of course, to get this going, somebody with access to your github repo needs to create an account. That's the only way to get access to this repo.

parity, parity-bitcoin, cpp-ethereum, kovri, and some others already have the snapcraft.yaml in their repo, are taking advantage of the continuous delivery and getting their software exposed to all the Ubuntu users.

hoschi-it:
Actually, there is a [snap in the store](https://uappexplorer.com/snap/ubuntu/electrum) since September 2016 (meaning it's v2.6.4-tpaw0), maintained by Tomas CaseyWilcox. Snap store is not meant to hold deprecated versions but to let users easily get updates...

Issue #2576, opened 2017-07-01
[bug] KeyError: 'type' in Wallet.add_transaction


dabura667:
During initial sync of a wallet that has 78 addresses used, this error appeared in the console and syncing hangs forever.

Restarting results in the same error, except this time the history actually displays partially, but it is completely messed up.

I can send the xpub if necessary.

```
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/electrum/network.py", line 845, in run
    self.wait_on_sockets()
  File "/usr/local/lib/python2.7/dist-packages/electrum/network.py", line 839, in wait_on_sockets
    self.process_responses(interface)
  File "/usr/local/lib/python2.7/dist-packages/electrum/network.py", line 625, in process_responses
    self.process_response(interface, response, callbacks)
  File "/usr/local/lib/python2.7/dist-packages/electrum/network.py", line 577, in process_response
    callback(response)
  File "/usr/local/lib/python2.7/dist-packages/electrum/synchronizer.py", line 146, in tx_response
    self.wallet.receive_tx_callback(tx_hash, tx, tx_height)
  File "/usr/local/lib/python2.7/dist-packages/electrum/wallet.py", line 663, in receive_tx_callback
    self.add_transaction(tx_hash, tx)
  File "/usr/local/lib/python2.7/dist-packages/electrum/wallet.py", line 585, in add_transaction
    is_coinbase = tx.inputs()[0]['type'] == 'coinbase'
KeyError: 'type'
```

dabura667:
AHHHH

Found the cause using -v

```
cannot find address in input script 5355540b6f93598893578893588851
```

I previously used this wallet to receive funds from a P2SH workshop I did with some students, so my history contains a transaction where the input won't be parsed properly.

dabura667:
The address that sent to my wallet https://insight.bitpay.com/address/3MbZjYS1Kjo5An9vVCwZYTd2JeobwjUsFh

The redeemscript: 6f93598893578893588851 (OP_3DUP OP_ADD OP_9 OP_EQUALVERIFY OP_ADD OP_7 OP_EQUALVERIFY OP_ADD OP_8 OP_EQUALVERIFY OP_TRUE)
The scriptSig part before the redeemscript: 535554 (OP_3 OP_5 OP_4)

[...]

PR #2577, opened 2017-07-01
Fix KeyError: 'type' for weird P2SH inputs

dabura667:
I have tested with my wallet, but the first one might mess with coinbase inputs.......

@ecdsa Please let me know what you think. I will allow edits from maintainers.


dabura667:
Fixes #2576

dabura667:
fixed on master

PR #2582, opened 2017-07-03
Add tab icons

bauerj:
This adds an icon to each tab in the main window which will hopefully improve usability.

Screenshot
===
python_2017-07-02_23-52-01

Icon source
===

The icons are part of [_fontawesome_](http://fontawesome.io/license/):

History
http://fontawesome.io/icon/list

Send
http://fontawesome.io/icon/paper-plane

Receive
http://fontawesome.io/icon/download

Addresses
http://fontawesome.io/icon/book

Contacts
http://fontawesome.io/icon/address-book

Coins
http://fontawesome.io/icon/money

Console
http://fontawesome.io/icon/bolt


ecdsa:
not sure if I really like it..
it adds more complexity to the GUI

bauerj:
My opinion is that clear icons make it easier to find the right tab.

ecdsa:
icons

bauerj:
New icons: 
python_2017-07-09_00-29-04

MaxXor:
I'd prefer the colorful icons as they are more easy distinguishable and also match the existing transaction icons.

EagleTM:
I like the colorful icons and think most users would enjoy them.

Compromise could be to make it a config setting, so you can switch it off by selecting "minimal visual effects". This setting could be used for other beautifications in the future which might be seen as less slick to purists.

ecdsa:
thanks; it's better with color

PR #2601, opened 2017-07-12
Fix "Back" button in installwizard

bauerj:
Since this seems to be the first page of the wizard, there is nothing to go back to. Furthermore, clicking "Back" actually takes you to the next step.

This changes the button to "Close" and closes the wizard.


ecdsa:
why did you change return to self.close() ?
that seems to be wrong

ecdsa:
reverted in f33ee01e6456734ab723ee4444b7bbfb706b29be

PR #2615, opened 2017-07-17
Move SSL checkbox to server tab

bauerj:
This is unrelated to the proxy.


ecdsa:
no, it should notbe moved there; 
the server tab is disabled by auto-connect.

PR #2626, opened 2017-07-23
Add locktime cmdline support

neocogent:
This simply exposes tx locktime as a --locktime option for payto command. This may be useful after a fork to allow splitting coins using locktime method. 

Currently only supports by block number. It wouldn't be hard to add date support but not sure what date format it should accept. If that can be decided I could change the arg type to support a date value and convert it.

I tested this only by creating an unsigned tx and then decoding it on http://coin.bin to verify. It gave correct locktime value. I don't currently have a wallet with enough funds to sign and send a test locktime tx. hopefully, someone who sees value in this cmd option can test.

If a ledger wallet then this PR requires the ledger fix PR #2625 just submitted. The ledger cmdline module cannot load with spelling mistake.

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -405,7 +405,7 @@ class Commands:
         sig = base64.b64decode(signature)
         return bitcoin.verify_message(address, sig, message)
 
-    def _mktx(self, outputs, fee, change_addr, domain, nocheck, unsigned, rbf, password):
+    def _mktx(self, outputs, fee, change_addr, domain, nocheck, unsigned, rbf, password, locktime=None):
         self.nocheck = nocheck
         change_addr = self._resolver(change_addr)
         domain = None if domain is None else map(self._resolver, domain)
@@ -417,6 +417,8 @@ class Commands:
 
         coins = self.wallet.get_spendable_coins(domain, self.config)
         tx = self.wallet.make_unsigned_transaction(coins, final_outputs, self.config, fee, change_addr)
+        if locktime != None: 
+            tx.locktime = locktime
         if rbf:
             tx.set_rbf(True)
         if not unsigned:
@@ -424,19 +426,19 @@ class Commands:
         return tx
 
     @command('wp')
-    def payto(self, destination, amount, tx_fee=None, from_addr=None, change_addr=None, nocheck=False, unsigned=False, rbf=False, password=None):
+    def payto(self, destination, amount, tx_fee=None, from_addr=None, change_addr=None, nocheck=False, unsigned=False, rbf=False, password=None, locktime=None):
         """Create a transaction. """
         tx_fee = satoshis(tx_fee)
         domain = [from_addr] if from_addr else None
-        tx = self._mktx([(destination, amount)], tx_fee, change_addr, domain, nocheck, unsigned, rbf, password)
+        tx = self._mktx([(destination, amount)], tx_fee, change_addr, domain, nocheck, unsigned, rbf, password, locktime)
         return tx.as_dict()
 
     @command('wp')
-    def paytomany(self, outputs, tx_fee=None, from_addr=None, change_addr=None, nocheck=False, unsigned=False, rbf=False, password=None):
+    def paytomany(self, outputs, tx_fee=None, from_addr=None, change_addr=None, nocheck=False, unsigned=False, rbf=False, password=None, locktime=None):
         """Create a multi-output transaction. """
         tx_fee = satoshis(tx_fee)
         domain = [from_addr] if from_addr else None
-        tx = self._mktx(outputs, tx_fee, change_addr, domain, nocheck, unsigned, rbf, password)
+        tx = self._mktx(outputs, tx_fee, change_addr, domain, nocheck, unsigned, rbf, password, locktime)
         return tx.as_dict()
 
     @command('w')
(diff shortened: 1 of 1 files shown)


ecdsa:
wallet.make_unsigned_transaction is supposed to set the locktime 
(see lines currently commented out)
your commit would brak that

bauerj:
> I don't currently have a wallet with enough funds to sign and send a test locktime tx. 

You can test it on [testnet](https://testnet.manu.backend.hamburg/faucet), just start Electrum with `--testnet`.

neocogent:
@bauerj  Yes. It's something I haven't used before but I should give that a try. Will do later today.

@ecdsa I did originally pass the locktime down to that function and override the 0 value it sets. Then I decided it's probably better not to affect as many files in the PR and backed it out to simply overriding it as it is. I had a look again just now at latest chgs and seems in make_unsigned_transaction has locktime commented out but would set it to current block height. I'm not aware of what that is about or keepkey behavior (in comment), but either way if the user wants to set a locktime it's going to override regardless and it seems to me it's better to keep it contained in the commands code. But of course I'm a bystander here simply wishing that locktime could be an option in Electrum and the best way to code that is your choice.

bauerj:
> I had a look again just now at latest chgs and seems in make_unsigned_transaction has locktime commented out but would set it to current block height instead of zero (set in tx constructor). I'm not aware of what that is about

See bitcoin/bitcoin#2340

neocogent:
Ah yes, I see now - to prevent future fee sniping. In that case it makes sense the command should pass locktime down to the make_unsigned_transaction rather than assuming default of zero and override the current block default when requested. I should change this and test on testnet before updating the PR.

Seems to me this is another use case for a payto locktime option. It allows someone (expert enough) doing an offline tx to preset a locktime for next blocks to preclude fingerprinting of offline clients.

ecdsa:
@neocogent I think you did not understand my comment.
if a locktime is set in make_unsigned_transaction, your commit will overwrite it with zero.
it should leave it unchanged by default

neocogent:
I've updated the commit now. It uses make_unsigned_transaction and passes None so that default value can be set in that function. It keep current value by default but I updated the commented code so that in future it will set current block.

I tested this on testnet and the tx was created correctly but would not be accepted by the network. Here is the details for my test tx and I currently don't know why the network won't accept it.
Output from payto cmd:
```
./electrum --testnet -w testnet1 payto mjfZCCgT43mv6Loajtr1pQUbFAaDWNENMQ 0.1 -f 0.0005 --locktime 1155076
{
    "complete": true, 
    "final": true, 
    "hex": "0100000001c5f04db5ce9f67da9b22de3cd7aad0a867c10cf3be7377bdecf5c7232b1ec6fa000000006b48304502210096c878a4c628f7a492da298a53d50a8158f595e2de29aae0391623c27b2e106402206c63e362626f4842d07ae8715632a670341849c9f2b24ab702f3702064ce8bc6012103b5edb6daf3359e2de626e8d2f8c9bbbd294c4cd0f2eaf8539207049fa966e966feffffff0280969800000000001976a9142d80f6901c38e3e85b619b868bcb2adec764e66d88acb04a2607000000001976a914a0f06b5092d7b7ba359ac298b1411e5bc225366288ac04a01100"
}
```
I then started the daemon and used broadcast. Returned error:
```
./electrum --testnet -w testnet1 broadcast 0100000001c5f04db5ce9f67da9b22de3cd7aad0a867c10cf3be7377bdecf5c7232b1ec6fa000000006b48304502210096c878a4c628f7a492da298a53d50a8158f595e2de29aae0391623c27b2e106402206c63e362626f4842d07ae8715632a670341849c9f2b24ab702f3702064ce8bc6012103b5edb6daf3359e2de626e8d2f8c9bbbd294c4cd0f2eaf8539207049fa966e966feffffff0280969800000000001976a9142d80f6901c38e3e85b619b868bcb2adec764e66d88acb04a2607000000001976a914a0f06b5092d7b7ba359ac298b1411e5bc225366288ac04a01100
[
    false, 
    "error: The transaction was rejected by network rules.  (64: non-final)\n[0100000001c5f04db5ce9f67da9b22de3cd7aad0a867c10cf3be7377bdecf5c7232b1ec6fa000000006b48304502210096c878a4c628f7a492da298a53d50a8158f595e2de29aae0391623c27b2e106402206c63e362626f4842d07ae8715632a670341849c9f2b24ab702f3702064ce8bc6012103b5edb6daf3359e2de626e8d2f8c9bbbd294c4cd0f2eaf8539207049fa966e966feffffff0280969800000000001976a9142d80f6901c38e3e85b619b868bcb2adec764e66d88acb04a2607000000001976a914a0f06b5092d7b7ba359ac298b1411e5bc225366288ac04a01100]"
]

```
I tried a signtransaction as well and also by loading raw tx in gui and got same results. Not sure why it's rejected as appears to be final and signed when displayed on screen in gui.

First time using testnet so any suggestions welcome. Thx.

ecdsa:
this PR is broken, maybe you made a mistake rebasing.
also, you don't have to modify anything in wallet.py; all I'm saying is that the option should not overwrite the default locktime with zero, as your initial PR did.

dabura667:
:-( git is hard... I think my first time using git was to try and PR to Electrum... 2013 or so.

imo locktime is nearly useless without CHECKLOCKTIMEVERIFY, and its use case for a hard fork scenario is nullified by the two way replay protection existing in Bitcoin Cash.

So I'd say NACK for now.

If you'd like me to help you clean up the rebase, let me know @neocogent

neocogent:
It was ok before. And then I guess I tried to sync my repo up to 2.9.0 and that somehow got relayed thru into the PR. Didn't expect that or even know it did that. I'll see if I can fix it myself.

I wasn't thinking that locktime would be useful for BCC specifically. I wanted this before when BIP148 had the focus and now it seems like it's just something an advanced user ought to be able to do. I've seen various posts saying using locktime has uses even without OP_CLTV. Wouldn't implementing OP_CLTV be much more complicated?

neocogent:
I believe I've rebased now so that it's ok. Also, reverted out changes in wallet.py so now only commands.py is affected.

PR #2666, opened 2017-08-01
Remove usages of deprecated apply() builtin

suut:
Instead of using the deprecated function `apply(f, args, kwargs)` we use the modern Python 3-compatible unpack syntax `f(*args, **kwargs)`, easing the porting effort.


ecdsa:
fixed in 4c2deb6d2bfe4680a2cc9596f15d10823665f58a
next time please test.

suut:
Sorry, I thought it'd work the same as in Python 3.

Issue #2490, opened 2017-06-04
The current fee unit used is BTC/kB. Popular convention is Sat/B [enhancement]


cluelessperson:
My current max fee is 0.0043584 BTC/kB

This is tedious (math is hard) to translate to Sat/B for users, which most sites use by popular convention. This will alleviate further fee confusion and help people conform to the same known glossary.

Examples:

https://bitcoinfees.21.co/
https://blockchain.info/tx/68973cd03a6d264c1396cb1e60f27fe87214f90aa20e70abb39cc201c706a1c5

dabura667:
you can change the issue title and edit your original post... there is no need for a new issue.

juscamarena:
sat/byte should be standard imo.

ecdsa:
I agree.
we could also add the color codes used in https://jochen-hoenicke.de/queue/#24h

neocogent:
I have a commit for adding sat/byte on my repo for my own use (my-tweaks branch). I avoided submitting a PR on this because I thought the general consensus was that it was not wanted. I'd be happy to submit a PR on it so that I don't need to keep my own version. Here is link to commit - only one line changed anyway.

https://github.com/neocogent/electrum/commit/2274c312ea390169ade73fabb1764b3db8fbce7b

au:
Not to preach to the choir, but this is rather necessary. [Mempool](https://jochen-hoenicke.de/queue/#2h) charts and [bitcoin fee](https://bitcoinfees.21.co/) calculators all use sat/byte, which are essential for trying to figure out transactions, unless you want to rely on Blockchain.info's calculator, which is the next best option at the moment.

au:
Personally, that commit would be tipping point for me using Electrum over Blockchain, who has a very friendly fee calculator at the moment.

Would you consider cherry picking that commit into an updated repo, and making a PR for it, @neocogent ?

neocogent:
I've rebased as a fresh branch called fee-sat-byte and created a PR for this one line change. Even if it's not merged soon you can pull this into your local repo as a branch with:

    git fetch upstream pull/2667/head:sat-fee-byte

and then test locally.

Tested on my local wallet and looked ok on my tx dialog preview.

philrosenthal:
IMHO the fee should be set by entering "satoshi per byte", anything else results in tedious math, or overpaying.

dabura667:
> tedious math

0.00123456 BTC/kB = 123 satoshis/byte

[...]

PR #2667, opened 2017-08-01
Add sat/byte fee info in preview.

neocogent:
Adds an extra sat/byte fee value to tx preview dialog, and to slider tool tip.

Requested by some users as sat/byte is commonly felt to be more standard and makes comparison with mempool status sites easier.

See issue #2490.

### DIFF
--- a/gui/qt/fee_slider.py
+++ b/gui/qt/fee_slider.py
@@ -29,7 +29,7 @@ class FeeSlider(QSlider):
 
     def get_tooltip(self, pos, fee_rate):
         from electrum.util import fee_levels
-        rate_str = self.window.format_amount(fee_rate) + ' ' + self.window.base_unit() + '/kB'
+        rate_str = '%.0d sat/byte' % (fee_rate/1e3) if self.window.fee_unit == 0 else '%1.3f mBTC/kB' % (fee_rate/1e5)
         if self.dyn:
             tooltip = fee_levels[pos] + '\n' + rate_str
         else:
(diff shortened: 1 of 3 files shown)


ecdsa:
I don't think it's nice to show both; it is confusing.
there should be a preference for fee unit.

neocogent:
I can add a drop down for "Fee Unit" in prefs.
Should it use Base Unit / kB or fixed mBTC/kB ? As well as sat/byte of course.
Right now it uses Base Unit though I imagine users may not want BTC/kB even if that's their base unit due to how many zeros.

ecdsa:
I think fixed mBTC/kb is better

neocogent:
Updated PR with fee_unit preference dropdown, and fee slider and tx preview/details now use this preference.

I didn't change the max_static_fee preference units to match because it looked like it meant editing another file amount_edit.py and adding a class that uses fee_unit. If you think this edit box really should adapt to fee unit selected I'll go ahead. Just trying not to change too much.

lzsaver:
```
rate_str = '%.0d sat/byte' % (fee_rate/1e3) if self.window.fee_unit == 0 else '%1.3f mBTC/kB' % (fee_rate/1e5)
TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'
```

juscamarena:
@ecdsa I don't anyone who thinks mBTC/kb is better, sat/b is pretty much the standard for human readable fee rates. .

ecdsa:
merged manually after changes

PR #2679, opened 2017-08-03
Fix hide in tray on Qt-based window managers

suut:
In Qt-based window managers, hide in tray doesn't work because the window manager reports two simple-clicks instead of a double-click.

The bug has been searched for on KDE, LXQt, GNOME, GNOME Classic, Cinnamon, LXDE, XFCE, MATE; it appears only on KDE and LXQt.

This commit adds a configuration option (located under the tab _Appearance_, and named `tray_simpleclick` in config file) that enables hide in tray with a simple-click instead of a double-click, and the option is enabled by default on KDE and LXQt.

### DIFF
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -123,7 +123,7 @@ class ElectrumGui:
         self.tray.setIcon(self.tray_icon())
 
     def tray_activated(self, reason):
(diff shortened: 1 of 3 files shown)


ecdsa:
please rework this; do not add Qt code to simple_config.py

PR #2687, opened 2017-08-04
TREZOR: Remove sighash from input signatures

jhoenicke:
TREZOR firmware expects signatures not to contain the trailing sighash
byte.  This fixes issue fyookball/electrum#47 which also applies to
electrum.


ecdsa:
thanks

molecular:
thank you!

Issue #2604, opened 2017-07-13
Request: Can the history command have an optional "after timestamp" argument?


alex-georgiou:
I am aware that on the command line I can filter the output of `electrum history` using `jq`. However if I want to scan the history over the RPC interface, I think it would be really helpful to be able to request only the history after timestamp _t_. Or the history after the _nth_ transaction. This way one could programmatically poll for history changes without retrieving the entire wallet history in situations where websockets are not a good fit. (I'm developing a WordPress plugin.)

I'm new to Electrum so I don't know if this is a good idea, or if there's a better way to do this. Perhaps someone can let me know. Thanks.

PR #2715, opened 2017-08-09
History limits

zebra-lucky:
These changes allow to limit history command output.

From command line limits are follow:
- **--from-height FROM_HEIGHT** &ndash; show history output starting from block FROM_HEIGHT (inclusive)
- **--to-height TO_HEIGHT** &ndash; show history output finishing on block TO_HEIGHT (**NOT** inclusive)
- **--from-time FROM_TIME** &ndash; show history output starting from time FROM_TIME (inclusive)
- **--to-time TO_TIME** &ndash; show history output finishing on time TO_TIME (**NOT** inclusive)

Time can be in epoch seconds or in string in local time for electrum daemon in format "YYYY-mm-dd HH:MM:SS".

RPC calls also can use string or int time formats, and are:
- from_height
- to_height
- from_time
- to_time

Closes: #2604

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -50,6 +50,18 @@ def satoshis(amount):
     return int(COIN*Decimal(amount)) if amount not in ['!', None] else amount
 
 
+def int_epoch_time(data):
+    """Return int epoch time from Y-m-d H:M:S format or simply convert to int"""
+    if type(data) == int:
+        return data
+    try:
+        res = datetime.datetime.strptime(data, '%Y-%m-%d %H:%M:%S')
+        res = int(time.mktime(res.timetuple()))
+    except ValueError:
+        res = int(data)
+    return res
+
+
 class Command:
     def __init__(self, func, s):
         self.name = func.__name__
@@ -444,11 +456,21 @@ class Commands:
         return tx.as_dict()
 
     @command('w')
-    def history(self):
-        """Wallet history. Returns the transaction history of your wallet."""
+    def history(self, from_height=-1, to_height=-1, from_time=-1, to_time=-1):
+        """Wallet history. Returns the transaction history of your wallet
+        with optional limit by block height/time."""
+
+        # Additional conversion for RPC calls
+        from_height = int(from_height)
+        to_height = int(to_height)
+        from_time = int_epoch_time(from_time)
+        to_time = int_epoch_time(to_time)
+
         balance = 0
         out = []
-        for item in self.wallet.get_history():
+        for item in self.wallet.get_history(
+                from_height=from_height, to_height=to_height,
+                from_time=from_time, to_time=to_time):
             tx_hash, height, conf, timestamp, value, balance = item
             if timestamp:
                 date = datetime.datetime.fromtimestamp(timestamp).isoformat(' ')[:-3]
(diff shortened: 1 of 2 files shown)


lib/commands.py
```diff
@@ -721,6 +743,10 @@ def help(self):
     'pending':     (None, "--pending",     "Show only pending requests."),
     'expired':     (None, "--expired",     "Show only expired requests."),
     'paid':        (None, "--paid",        "Show only paid requests."),
+    'from_height': (None, "--from-height", "Starting from height."),
+    'to_height':   (None, "--to-height",   "Finishing on height."),
+    'from_time':   (None, "--from-time",   "Starting from time."),
```
bauerj: This should contain information about the expected format.

lib/wallet.py
```diff
@@ -479,18 +479,32 @@ def get_tx_info(self, tx):
         return tx_hash, status, label, can_broadcast, can_bump, amount, fee, height, conf, timestamp, exp_n
 
 
-    def get_addr_io(self, address):
+    def get_addr_io(self, address, **kwargs):
```
bauerj: I think it would be more clear to explicitly state the keyword arguments here.

lib/commands.py
zebra-lucky: Done

lib/wallet.py
zebra-lucky: Done

bauerj:
Any reason why you still use `kwargs.get` e.g. in `get_history`? Looks good to me otherwise.

zebra-lucky:
I was not sure about other methods, will change in a half a hour.

bauerj:
Okay, if you change one method it makes sense to change the other methods too 😊

zebra-lucky:
I've found some additional logic error when both from_time/from_height is set.
Will commit when resolve it.

zebra-lucky:
Removed kwargs from methods, and fixed error when both min_height/min_time was set.

zebra-lucky:
Found additional logical problem when filtering by timestamp, must study the problem to finish code.

ecdsa:
thanks for the pull request. 
please don't take it bad, but it would be better to submit it once it is ready, and to rebase it as a single commit

zebra-lucky:
Please. Ok, it seems no errors leaved. Will rebase when test again some wallets.

I make significant changes to code of history-limits branch.
- Now **from_time**/**to_time** not advanced as params further of **get_history**, and in **get_history** from time params (Re)calculated height params if it is need. **get_addr_io**, **get_addr_balance**, **get_balance** uses only **to_height** param, and in **get_history** code only **from_height**, **to_height** is used.  This is done due to unstability of block time, it not always increment with block height increment.
- Fix display and calculation of balance on unconfirmed tx. If to_height is not last block with transactions from wallet, then unconfirmed tx not displayed and not used in balance calculation.

zebra-lucky:
Hi!

I'm finally finished with code to history-limits, but not pushed to remote yet.

Added `with_unverified=True` param to `get_addr_io`, `get_addr_balance`, `get_balance`.
That stabilized code and all testing on different wallets (all confirmed, only unconfirmed, mix) works well.

Can I rebase and squash PR branch and then force push it to remote, or not shoul do that?
I do not try this earlier. Maybe github code can't track this.

bauerj:
> Can I rebase and squash PR branch and then force push it to remote

Yes.

zebra-lucky:
Added yet some small fixes. I'll squash tomorrow to see changes.

zebra-lucky:
Rebased and squashed now.

zebra-lucky:
Seems I should do to_time limit not inclusive.

jpentland:
The commit summary contains a typo. height/heigt.

It might be a nice idea to make the commit message more similar to the summary you provided above in the pull request description. Then someone looking at the history can see what was added.

zebra-lucky:
Ok, typo is corrected, will add tests in a week.

zebra-lucky:
Typo is corrected, will add more tests in a week or two.

bauerj:
Hey @zebra-lucky,

you said you were going to add more commits. Are you still working on it or is this ready for review now?

zebra-lucky:
Hi! Still need to write tests. No time at moment. Maybe try to do in two weeks again.

ecdsa:
replaced by 3f954a8b3d39d2cb45927371a856802104f9182c

PR #2716, opened 2017-08-09
Added the ability to create a new address.

block8437:
Added a new RPC call `getnewaddress` which is a direct call to `create_new_address`.


dabura667:
concept ACK

ecdsa:
what is the point of adding a new command?
there is already two commands that can create addresses: getunusedaddress and addrequest

block8437:
@ecdsa this is used when I need to assign users in a service bitcoin addresses that are't always used immediately. If they are assigned to the user and not used immediately they can be reused by other users which leads to address conflicts.

PR #2727, opened 2017-08-11
Add BIP39 seed restore option to validate checksum.

neocogent:
When restoring a seed and BIP39 option selected then another option to "Validate checksum" is available.

This is useful for users who want to generate a BIP39 seed by rolling dice, flipping coins or other external entropy. It's also just good for users who want to validate they are using a seed compatible with other wallets. Currently if you enter random words Electrum will create a wallet but the seed words used will not work with other BIP39 wallets which seem to all prevent non-valid seeds.

I tested this against most of the test case word lists in the BIP39 spec on github, and with language specific seeds created on [iancoleman site](https://iancoleman.github.io/bip39/). Also tried as many variations of user abuse I could think of. It worked well for as far as I could tell.

When a user wants to use external entropy they can enter the words and vary the last word by using nearby (apparently last 4 bits, so 16 words) until they get a valid chksum. I tried that once and it took 12 tries before valid.

~~As far as I could tell there is no way to indicate language for word list. It just assumes default. Maybe it needs a combobox selector or it could try each available in sequence until it finds a word?~~
UPDATE - I modified the PR to autodetect and validate for languages that have a wordlist present (currently English, Japanese , Sim. Chinese, Spanish - couldn't check Portuguese). Many thanks to  dabura667 for help getting Japanese working. 

Question - should checksum validate be on by default and option to disable? Currently off, with option to turn on.

### DIFF
--- a/gui/qt/seed_dialog.py
+++ b/gui/qt/seed_dialog.py
@@ -52,6 +52,7 @@ class SeedLayout(QVBoxLayout):
     #options
     is_bip39 = False
     is_ext = False
+    do_validate = False
 
     def seed_options(self):
         dialog = QDialog()
@@ -61,6 +62,7 @@ class SeedLayout(QVBoxLayout):
             cb_ext.setChecked(self.is_ext)
             vbox.addWidget(cb_ext)
         if 'bip39' in self.options:
+            cb_valid = QCheckBox(_('Validate checksum'))
             def f(b):
                 if b:
                     msg = ' '.join([
@@ -76,17 +78,27 @@ class SeedLayout(QVBoxLayout):
                     ])
                     #self.parent.show_warning(msg)
                 self.seed_type_label.setVisible(not b)
+                cb_valid.setVisible(b)
                 self.is_seed = (lambda x: bool(x)) if b else self.saved_is_seed
+                self.on_edit()
+            def v(b):
+                self.is_seed = (lambda x: bool(x)) if not b else self.validate
+                self.do_validate = b
                 self.on_edit()
             cb_bip39 = QCheckBox(_('BIP39 seed'))
             cb_bip39.toggled.connect(f)
             cb_bip39.setChecked(self.is_bip39)
             vbox.addWidget(cb_bip39)
+            cb_valid.setVisible(self.is_bip39)
+            cb_valid.toggled.connect(v)
+            cb_valid.setChecked(self.do_validate)
+            vbox.addWidget(cb_valid)
         vbox.addLayout(Buttons(OkButton(dialog)))
         if not dialog.exec_():
             return None
         self.is_ext = cb_ext.isChecked() if 'ext' in self.options else False
         self.is_bip39 = cb_bip39.isChecked() if 'bip39' in self.options else False
+        self.do_validate = cb_valid.isChecked() if 'bip39' in self.options else False
 
 
     def __init__(self, seed=None, title=None, icon=True, msg=None, options=None, is_seed=None, passphrase=None, parent=None):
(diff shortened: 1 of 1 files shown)


dabura667:
This is broken for Japanese phrases.

You must normalize the text before attempting to use .split(). This won't affect English, as any ASCII text NFKD normalized is equal to itself.

dabura667:
Japanese spaces are not split() but NFKD normalizing JP spaces replaces it with ASCII space

neocogent:
Yes, I realized that after posting from looking at the other BIP39 chksum PR. I've made changes now to call normalize_text and to check each language in turn until it finds a match or fails. 

Problem I'm having is even after normalize_text the split is still not making words. I get word count = 1 always. Googled a bit and found maybe text.encode('utf-8').split() should work after normalizing but that still didn't help. Trying now with Japanese test vectors linked from BIP39 spec. Current code:

```
        from electrum.mnemonic import Mnemonic, normalize_text, filenames
        text = normalize_text(text)
        words = len(text.split())
        if words not in [12,15,18,21,24]:
            print "bad word count", words
            return False

```
Always gives me bad word count 1. Also with text.encode('utf-8').split(). I'm no expert on Japanese or unicode so if it's obvious please hit me.

dabura667:
iirc .split() with no args only defaults to space for str type and not unicode type.

Try the explicit .split(' ')

neocogent:
I tried that but no better. I added a line to print the text as a list; here's what I got:

`text [u'\u305d', u'\u3068', u'\u3064', u'\u3089', u'\u3042', u'\u307e', u'\u3068', u'\u304a', u'\u304a', u'\u3046', u'\u3042', u'\u3053', u'\u304b', u'\u308c', u'\u308b', u'\u3044', u'\u304f', u'\u3075', u'\u3093', u'\u3051', u'\u3044', u'\u3051', u'\u3093', u'\u3042', u'\u305f', u'\u3048', u'\u308b', u'\u3044', u'\u3088', u'\u304f', u'\u305d', u'\u3068', u'\u3064', u'\u3089', u'\u3042', u'\u307e', u'\u3068', u'\u304a', u'\u304a', u'\u3046', u'\u3042', u'\u304b', u'\u3061', u'\u3083', u'\u3093']`

for the Japanese: そとづら　あまど　おおう　あこがれる　いくぶん　けいけん　あたえる　いよく　そとづら　あまど　おおう　あかちゃん

Seems more complex than just u'\u3042' as there should be 11 whitespaces (12 words).
Oh, I messed with the normalize and removed the last line which alters CJK and then it can count words, but cannot find in word list. Messy stuff.

dabura667:
found the problem

normalize_text adds an extra step that varies from BIP39.

```python
# remove whitespaces between CJK
seed = u''.join([seed[i] for i in range(len(seed)) if not (seed[i] in string.whitespace and is_CJK(seed[i-1]) and is_CJK(seed[i+1]))])
```

This is NOT how BIP39 works, and iirc Thomas justified it as "well, Electrum seeds are not BIP39 compatible anyways"

The reason why the example on the site worked is because it is NOT typed by a Japanese person on a Japanese keyboard. It is generated from a pre-normalized list using join on ASCII spaces to begin with.

here is a naturally typed Japanese phrase non-normalized using a Japanese keyboard:
```
そとづら　あまど　おおう　あこがれる　いくぶん　けいけん　あたえる　いよく　そとづら　あまど　おおう　あかちゃん
```

It looks the same, but try opening your browser javascript console, type quotes and paste YOUR phrase (from your comment) and MY phrase (above, I typed it using my keyboard). and test if they are equal with == you will see false.

The one you pasted was a stupid stupid stupid Pull Request someone made on my test vectors, they said "OMG THE TEST VECTORS ARE NOT NORMALIZED AND THE WORDLIST IS NORMALIZED SO MY STUPID APP IS THROWING ERRORS!" and they changed the words BUT NOT THE SPACES (*facepalm*) to be normalized in the json file... The whole idea of test vectors is to list INPUT FROM ACTUAL USERS which makes no sense why the test vectors should be normalized... and to boot they didn't normalize the Japanese spaces... lol

Your other site was generating phrases using ASCII spaces and the pre-normalized list with is why your test worked without normalize_text.

normalize_text does a lot of Electrum-only stuff, so instead you should only do this:

```python
import unicodedata
unicodedata.normalize('NFKD', unicode(text))
```

neocogent:
Wow. Nice you worked that all out. It's gotta help to have actual Japanese knowledge. I'll try again with this change and see what I can do. So I should use the test vectors linked from BIP39 for testing, or I should not? I probably need to use the sample you manually typed above, right?

It sounds like iancoleman's site needs an issue opened to inform them the Japanese language option on the site won't work for actual Japanese typing. Or maybe I'm still befuddled.

neocogent:
Boom. That did it. Your seed above produces first address, 

1CmjRBbhvbUcS6UkTjvC781j795HUnUpxw

iancoleman site accepts your seed above and produces same addresses. So it's working ok too. Just the crazy test vectors that are screwing things up.

I'll commit and push this change. Just fixed another error as well.

ecdsa:
I've added a warning, which breaks this PR. Anyway, I think the approach in #2130 is better; we don't want to disable the next button because we cannot be sure about the language, and there is no guarantee that the dictionary will remain the same.

ecdsa:
I merged #2130 after some reworking

neocogent:
It only disabled the Next button if you specifically enabled the chksum validation, so I don't see how that could be considered a problem. No one would do that by mistake and get stuck. I guess I'd have to go see how it works now and if I can do what I'd want with it.

PR #2754, opened 2017-08-15
bugfix: certain network config changes did not get saved

### LINKED ISSUE #2747: tor proxy setting does not stay selected
When selecting Use Tor Proxy in the Gui from Tools->Network, settings are not saved after app restart and need to be reselected.

SomberNight:
Fixes #2747

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -471,12 +471,14 @@ class Network(util.DaemonThread):
                 int(proxy['port'])
         except:
             return
-        self.config.set_key('auto_connect', auto_connect, False)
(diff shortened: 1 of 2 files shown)


SomberNight:
I believe the bug was introduced by #2720

ecdsa:
thanks for pointing this out.
I think it's better to revert #2720, because your fix makes things a lot more complex

ecdsa:
reverted in a22cf973b4a43ef03ef08d8d6fa576fb35221a29

Issue #2738, opened 2017-08-12
trezor multisig failure to sign


fireduck64:
After creating a multisig 2of3 wallet with two standard electrum seeds and one trezor cosigner I get the following error when trying to send a transaction that involves change:

Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/electrum_gui/qt/util.py", line 582, in run
    result = task.task()
  File "/usr/local/lib/python2.7/dist-packages/electrum/wallet.py", line 1141, in sign_transaction
    k.sign_transaction(tx, password)
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/plugin.py", line 66, in sign_transaction
    self.plugin.sign_transaction(self, tx, prev_tx, xpub_path)
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/plugin.py", line 242, in sign_transaction
    signed_tx = client.sign_tx(self.get_coin_name(), inputs, outputs, lock_time=tx.locktime)[1]
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/clientbase.py", line 221, in wrapped
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 158, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 824, in sign_tx
    res = self.call(proto.TxAck(tx=msg))
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 158, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 201, in call
    msg = handler(resp)
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/clientbase.py", line 35, in callback_Failure
    raise RuntimeError(msg.message)
RuntimeError: Failed to compile output

However, if I send a transaction that spends the entire wallet with no change, I get no errors.

This is with electrum 2.9.3 and python trezor 0.7.16

ecdsa:
I cannot reproduce that; is your firmware up to date?

fireduck64:
Yeah, 1.5.0

fireduck64:
Also, I have the pin disabled.  Maybe that changes the flow?

BTCgithub:
[...]

PR #2755, opened 2017-08-15
Set bip32 path for multisig change address

jhoenicke:
This fixes bug #2738 

Multisig is still not working perfectly with TREZOR.  Since the xpubkey information is destroyed when a signature is filled, the second signer will not accept the change address.  This is because TREZOR checks that the return address uses the same xpubs as all inputs.  However this is a different issue.


ecdsa:
address_n is already computed line 337, no need to do it again.

jhoenicke:
The address_n is slightly different.  line 337 computes only the path from the public key, but we need to give the full path from the master seed.

ecdsa:
sorry I did not notice that

PR #2760, opened 2017-08-16
[TREZOR] Added Segwit support.

jhoenicke:
This adds segwit support for TREZOR in Electrum.  So far I only tested standard wallet.  

Following changes were necessary outside the TREZOR plugin.
- transaction.py: update_transaction handles segwit transactions.
- keystore.py: added a segwit parameter to bip44_derivation,
  use m/49' instead of m/44' for segwit.

### DIFF
--- a/lib/keystore.py
+++ b/lib/keystore.py
@@ -684,11 +684,10 @@ is_private_key = lambda x: is_xprv(x) or is_private_key_list(x)
 is_bip32_key = lambda x: is_xprv(x) or is_xpub(x)
 
 
-def bip44_derivation(account_id):
-    if bitcoin.TESTNET:
-        return "m/44'/1'/%d'"% int(account_id)
-    else:
-        return "m/44'/0'/%d'"% int(account_id)
+def bip44_derivation(account_id, segwit=False):
+    bip  = 49 if segwit else 44
+    coin = 1 if bitcoin.TESTNET else 0
+    return "m/%d'/%d'/%d'" % (bip, coin, int(account_id))
 
 def from_seed(seed, passphrase):
     t = seed_type(seed)
--- a/lib/transaction.py
+++ b/lib/transaction.py
@@ -421,8 +421,7 @@ def parse_input(vds):
 
 def parse_witness(vds):
     n = vds.read_compact_size()
-    for i in range(n):
-        x = vds.read_bytes(vds.read_compact_size())
+    return list(vds.read_bytes(vds.read_compact_size()).encode('hex') for i in xrange(n))
 
 def parse_output(vds, i):
     d = {}
@@ -548,7 +547,12 @@ class Transaction:
         for i, txin in enumerate(self.inputs()):
             pubkeys, x_pubkeys = self.get_sorted_pubkeys(txin)
             sigs1 = txin.get('signatures')
-            sigs2 = d['inputs'][i].get('signatures')
+            if d.get('witness') is None:
+                sigs2 = d['inputs'][i].get('signatures')
+            else:
+                # signatures are in the witnesses.  But the last item is
+                # the pubkey or the multisig script, so skip that.
+                sigs2 = d['witness'][i][:-1]
             for sig in sigs2:
                 if sig in sigs1:
                     continue
--- a/plugins/trezor/plugin.py
+++ b/plugins/trezor/plugin.py
@@ -259,7 +259,9 @@ class TrezorCompatiblePlugin(HW_PluginBase):
         derivation = wallet.keystore.derivation
         address_path = "%s/%d/%d"%(derivation, change, index)
         address_n = client.expand_path(address_path)
-        client.get_address(self.get_coin_name(), address_n, True)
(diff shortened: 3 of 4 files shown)


jhoenicke:
I use the derivation path to distinguish segwit (segwit over p2sh uses bip-49 instead of bip-44).  I noticed that Electrum uses a different xpub format.  Is that standardized?

Has anyone tested segwit with multisig?  Does it work for non-hardware wallets?

ecdsa:
thanks. 
the xpub format for segwit is not standardized and might not be deployed in mainnet

lib/transaction.py
```diff
@@ -425,8 +425,7 @@ def parse_input(vds):
 
 def parse_witness(vds):
     n = vds.read_compact_size()
-    for i in range(n):
-        x = vds.read_bytes(vds.read_compact_size())
+    return list(vds.read_bytes(vds.read_compact_size()).encode('hex') for i in xrange(n))
```
saleemrashid: Is this a bugfix?

ecdsa: maybe for python3 compatibility

saleemrashid: Doesn't look like it returned anything before though, also `xrange` doesn't exist in Python 3

ecdsa: indeed, it was not returning anything before; it was not used either :)

jhoenicke: It returned None before.  But the update_signature function needs the witnesses to update the signatures.

I guess update_signature isn't used if you sign with the seed in electrum; it is only used for hardware wallets.

jhoenicke:
I updated the patch for current master, in particular the python3 changes.

When I enable segwit for mainnet (needs two trivial changes to electrum), I can sync with the TREZOR segwit wallet.

plugins/trezor/plugin.py
```diff
@@ -26,6 +26,9 @@ class TrezorCompatibleKeyStore(Hardware_KeyStore):
     def get_derivation(self):
         return self.derivation
 
+    def is_segwit(self):
+        return self.derivation.startswith("m/49'/")
```
saleemrashid: I don't think this is a good idea (is there a wallet property for SegWit that we can use?)

lib/keystore.py
```diff
@@ -684,11 +684,10 @@ def is_private_key_list(text):
 is_bip32_key = lambda x: is_xprv(x) or is_xpub(x)
 
 
-def bip44_derivation(account_id):
-    if bitcoin.TESTNET:
-        return "m/44'/1'/%d'"% int(account_id)
-    else:
-        return "m/44'/0'/%d'"% int(account_id)
+def bip44_derivation(account_id, segwit=False):
+    bip  = 49 if segwit else 44
```
saleemrashid: `purpose` and `coin_type` and `account` would mimic the BIP-0044 wording, but it's up to you.

plugins/trezor/plugin.py
ecdsa: inddeed, you should use keystore.is_segwit()
it should not be a property of the plugin, because there is a single plugin object for many wallets

ecdsa: this looks ok, it is a property of the KeyStore

saleemrashid: @ecdsa It would be nice to be able to use a different derivation for whatever reason.

ecdsa: @saleemrashid that is a consequence of the bip39+bip43 design, which was imposed by the trezor development team. I believe it is a bad standard, but we are forced to follow it in the trezor plugin.

saleemrashid: @ecdsa It's not a consequence AFAIK because I could just change that to say any other derivation and it will still work. We should keep track of whether to use SegWit through a different means.

ecdsa: @saleemrashid yes you can, but we do not want to expose users to a system where they have to remember more than their seed. (at least not with electrum seeds)

saleemrashid: @ecdsa But they wouldn't, because this "SegWit property" would be set by Electrum.

ecdsa: @saleemrashid how would electrum know about this property?

saleemrashid: @ecdsa Well, the TREZOR plugin would know when the wallet is imported (it has to choose between `m/44'` and `m/49'` anyway)

ecdsa: @saleemrashid so you're proposing to add a checkbox in the trezor restore dialog, where the user enters the derivation? that would work for me

saleemrashid: @ecdsa We already allow the user to enter the derivation. I'm proposing to allow them to choose between SegWit and non-SegWit for any derivation.

ecdsa: @saleemrashid I am not opposed to that. I will let @jhoenicke decide if he wants to add it.

PR #2813, opened 2017-08-28
Preference item for ignoring dust - min_utxo_value, and highlight in coins list

neocogent:
Added config option min_utxo_value - default 0. 

Sets a threshold value for ignoring dust inputs to prevent wasting fees. 

Also highlights in utxo_list when an utxo is below the threshold so it's easy to tell.

Default is 0 so it has no effect. Since even 1 sat/byte fee would give make an input cost 148 sat it may be sensible to make a higher default. Nowadays, even at a low fee of 50 sat/byte outputs below 7400 sat are costing more to spend than the output is worth.

Tested on testnet and appeared to work ok.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -2444,6 +2444,15 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
             self.config.set_key('rbf_policy', x)
         rbf_combo.currentIndexChanged.connect(on_rbf)
         fee_widgets.append((rbf_label, rbf_combo))
+        
+        min_utxo = self.config.get('min_utxo_value', 0)
+        min_utxo_label = HelpLabel(_('Ignore Outputs Below (sat)') + ':', '')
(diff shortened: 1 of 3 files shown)


neocogent:
No, I didn't know it was checking that. I did look at the CoinChooser and saw it was not adding outputs when below a dust threshold (adds to fees). And I noticed it checks sufficient funds for each bucket. Plus different coinchooser subclasses use different priorities. I just did not see it discarding inputs when they cost more than they provide. I still don't see that after checking again.

I wouldn't have bothered if I'd known as an automated method is better than manually setting a limit. I think it would still be nice to highlight useless coins in the utxo_list. 

I started this after analyzing the blockchain and finding that 68% of 1 satoshi outputs had already been spent - so obviously most wallets are not filtering out the dust. After a check through the code here it seemed like Electrum was likewise not ignoring these outputs.

ecdsa:
strip_unneeded is what discards inputs that cost more than they provide

PR #2817, opened 2017-08-29
Add French & Italian wordlists

julien-boudry:
(copy from https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md)


ecdsa:
note: the GUI currently generate only English seeds, because the Android app keyboard would need to be extended with other languages.

ecdsa:
closing, we will not add this. see #6860

PR #2820, opened 2017-08-29
fix Qt threading issue

benma:
Issue: #2209

Fixes this error when signing with a HW wallet:

``` 
$ ./electrum
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
python3: xcb_io.c:259: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
Aborted (core dumped)
```
The cause is likely with the `show_message()` calls. The proper fix
would be to hunt down where QT Gui elements are touched from a wrong
thread.

The workaround is to call XInitThreads(), as the error message suggests.

### DIFF
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -82,6 +82,7 @@ class ElectrumGui:
         # GC-ed when windows are closed
         #network.add_jobs([DebugMem([Abstract_Wallet, SPV, Synchronizer,
         #                            ElectrumWindow], interval=5)])
+        QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_X11InitThreads)
         self.config = config
         self.daemon = daemon
         self.plugins = plugins


benma:
Note: this fix involved a bit of guess work, and it seems to fix the problem for me. Since the crash is non-deterministic, I can't guarantee this works, so more testers would be welcome (@ecdsa?) .

ecdsa:
I would love to test, but I rarely saw that bug.
how do you trigger it? with difgitalbitbox?

benma:
Yeah, I stumbled upon it when signing txs and messages with the digital bitbox. @btchip and @dieeasy, maybe you could also check if it fixes #2209?

ecdsa:
does the crash still occur with python3?

benma:
Yes.

btchip:
I'll try to reproduce - never really managed before but great news if it's fixed

ecdsa:
are there any known side effects of setting X11InitThreads ?
(eg do you know if it will break binaries on windows, osx)

ecdsa:
note: I doubt the problem is with show_message(), because it sends a signal

benma:
>  (eg do you know if it will break binaries on windows, osx)

I don't know.

> note: I doubt the problem is with show_message(), because it sends a signal

I think when I removed the dialog boxes, the issue also disappeared. Don't take my word for that yet, I'll have to recheck.

ecdsa:
please do. I would prefer to have a proper fix than this one

benma:
I removed the dialog boxes and had no error after 20 message signatures. With the dialogs, I got one within 10 tries three times in a row.

Not very scientific, but it indicates that there is an issue with the `show_message()` calls or related to it.

PR #2821, opened 2017-08-29
fix digital bitbox plugin, port to Python 3

benma:
The digital bitbox plugin didn't work anymore, mostly due to str/bytes
issues.

This patch fixes message and tx signing. Backup restore is also
tested.

Untested: firmware <= v2.1.1 and wallet initialization.


benma:
@douglasbakkum

PR #2829, opened 2017-09-01
python 3 changes in README

SomberNight:
Had problems getting Electrum to work after python3 migration.

Following commands in old README, notably ran into:

```
user@debian:~/wspace/electrum$ ./electrum 
Error: Could not import PyQt4 on Linux systems, you may try 'sudo apt-get install python3-pyqt4'

user@debian:~/wspace/electrum$ ./electrum 
qRegisterResourceData(int, str, str, str): argument 2 has unexpected type 'str'
Error: Could not find icons file.
Please run 'pyrcc4 icons.qrc -o gui/qt/icons_rc.py -py3', and reinstall Electrum
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ Getting started
 Electrum is a pure python application. If you want to use the
 Qt interface, install the Qt dependencies::
 
-    sudo apt-get install python-qt4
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks

PR #2837, opened 2017-09-04
fix urllib.quote py2 to py3

ariard:
urllib.quote no more being, replace by new urrlib.parse.quote


lib/util.py
```diff
@@ -602,7 +602,7 @@ def create_URI(addr, amount, message):
     if message:
         if six.PY2 and type(message) == unicode:
             message = message.encode('utf8')
```
bauerj: Can you remove these lines too while you're at it?

PR #2865, opened 2017-09-12
fix unresolved variable

SomberNight:
Wasn't sure if it was better to keep the change minimal or to be consistent.

i.e. input_script() method has
```
_type = txin['type']
```
so feel free to do that instead


ecdsa:
thanks

Issue #2876, opened 2017-09-15
Native SegWit scriptPubKeys are incorrectly constructed


sipa:
In `lib/bitcoin.py`, the following code is used to construct scriptPubKeys for BIP173 addresses:

```python
def address_to_script(addr):
    if is_segwit_address(addr):
        witver, witprog = segwit_addr.decode(SEGWIT_HRP, addr)
        script = bytes([witver]).hex() + push_script(bytes(witprog).hex())
return script
```

This will produce a script whose first byte is the witness version. However, the correct behaviour according to BIP141, is using `OP_n` for witness version n. This is the same when n = 0, but incorrect otherwise. If this code is used to send to future witness version addresses, it will likely result in funds loss.

Please consider implementing the BIP173 test vectors as unit tests; they would have caught this.

gmaxwell:
The correct code would look something like this:

assert(witver >= 0 and witver<=16)
script = bytes([(witver+0x50)*(witver>0)]).hex() + push_script(bytes(witprog).hex())

(But please do not just take my word for it, add the tests too)

dabura667:
@gmaxwell I like #2877 with the exception that multiplying by a boolean is hard to read.

Please see my comment: https://github.com/spesmilo/electrum/pull/2877#discussion_r139274210

ecdsa:
thanks for this

ecdsa:
see also 6e9b2a4459a108d438a917241c97f315544644c9

PR #2877, opened 2017-09-16
fix bech32 scriptPubKeys (OP_n), and tests

SomberNight:
See #2876

### DIFF
--- a/lib/bitcoin.py
+++ b/lib/bitcoin.py
@@ -338,7 +338,7 @@ def address_to_script(addr):
     if is_segwit_address(addr):
         witver, witprog = segwit_addr.decode(SEGWIT_HRP, addr)
         assert (0 <= witver <= 16)
-        OP_n = (witver+0x50) * (witver > 0)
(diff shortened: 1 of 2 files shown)


lib/bitcoin.py
```diff
@@ -337,7 +337,10 @@ def script_to_p2wsh(script):
 def address_to_script(addr):
     if is_segwit_address(addr):
         witver, witprog = segwit_addr.decode(SEGWIT_HRP, addr)
-        script = bytes([witver]).hex() + push_script(bytes(witprog).hex())
+        assert (0 <= witver <= 16)
+        OP_n = (witver+0x50) * (witver > 0)
```
dabura667: This gets the job done but is hard to read if you don't understand the way a boolean resolved in a mathmatic operation.

I suggest:

```python
OP_n = witver + 0x50 if witver > 0 else 0
```

gmaxwell: I didn't know python had a ternary or I would have used it!

PR #2883, opened 2017-09-17
Fix parsing of witnesses

jhoenicke:
The lengths in the witness structure are `var_int` not pushes.

From BIP-141:
The `witness` is a serialization of all witness data of the transaction.
Each txin is associated with a witness field. A witness field starts
with a `var_int` to indicate the number of stack items for the txin. It is
followed by stack items, with each item starts with a `var_int` to
indicate the length. Witness data is NOT script.

This bug was triggered by tx
d379210d85c1346dafbfd60e3cbc5c5573e50b1f9576d39f177afb2b378f1b98


jhoenicke:
Yes, serialize_witness also needs to be changed.

PR #2901, opened 2017-09-20
Added fiat currency alongside bitcoin quantity in transaction dialog

joechrisellis:
When you opened the transaction windows previously, it'd only show the bitcoin quantity. I've added the equivalent fiat value next to it.


ecdsa:
third, you included a commit from another pull request.
you should always make pull requests atomic, and not mix them

Issue #2899, opened 2017-09-20
server.version not always sent before blockchain.scripthash.subscribe


shsmith:
On each new connection, server.version must be sent in order to negotiate the protocol version with the server.  

When server selection is manual, and the initial server connection times out, the re-try will send blockchain.scripthash.subscribe messages and afterwards send the server.version. This sequence results in the server kicking the session because without a server.version message the blockchain.scripthash.subscribe message is invalid. The cycle seems to repeat forever after this.

I suspect this is due to this condition:
https://github.com/spesmilo/electrum/blob/8da36e3bd17cd4470f55a682ccc36fc52e8ccb2f/lib/network.py#L724

I suspect that the interface.ping_required method returns false on connection retry.

This is happening on testnet using git HEAD with a segwit-enabled wallet and ElectrumX 1.1 on the server side.

SomberNight:
I can confirm this.
To reproduce, block the port on the server after the client is fully synced. The client will time out and open a new connection to the server, then reopen the port on the server.

```
[127.0.0.1] --> ('server.version', ['3.0', '1.1'], 39)
[127.0.0.1] <-- {'result': ['ElectrumX 1.1', '1.1'], 'jsonrpc': '2.0', 'id': 39}
socket errno 11 (resource temporarily unavailable)
[127.0.0.1] --> ('server.version', ['3.0', '1.1'], 40)
[127.0.0.1] <-- {'result': ['ElectrumX 1.1', '1.1'], 'jsonrpc': '2.0', 'id': 40}
socket errno 11 (resource temporarily unavailable)
[127.0.0.1] --> ('server.version', ['3.0', '1.1'], 41)
[127.0.0.1] timeout 1
[Network] connecting to 127.0.0.1:50001:t as new interface
[127.0.0.1] connected
[Network] switching to 127.0.0.1:50001:t
[Network] sending subscriptions to 127.0.0.1:50001:t 0 29
[127.0.0.1] --> ('blockchain.headers.subscribe', [], 42)
[127.0.0.1] --> ('server.banner', [], 43)
[127.0.0.1] --> ('server.donation_address', [], 44)
[127.0.0.1] --> ('server.peers.subscribe', [], 45)
[127.0.0.1] --> ('blockchain.estimatefee', [25], 46)
[127.0.0.1] --> ('blockchain.estimatefee', [10], 47)
[127.0.0.1] --> ('blockchain.estimatefee', [5], 48)
[...]

PR #2902, opened 2017-09-20
send server.version before subscriptions

SomberNight:
See #2899 

I send `server.version` after `blockchain.relayfee` to replicate the order of messages on a new connection.


ecdsa:
I guess it would make sense to send it before

Issue #2878, opened 2017-09-16
Fee estimation only done when client starts?


SuBPaR42:
I tend to keep my client open for weeks at a time and noticed that I had to restart the client to get new fee estimates.

Can we add functionality so the client polls the server for new fee info every XX minutes/hours?

helpau:
Now(since PR?) mining fee is polling, i think this issue can be closed

PR #2906, opened 2017-09-20
Periodically poll fee estimates from server

SomberNight:
See #2878

It is not obvious to me whether introducing fee-specific lock(s) would be needed, I can't see any harmful race conditions at the moment.


ecdsa:
not sure about that.. it would be better to have a subscription model, where the server pushes updates when it changes its estimates.

Issue #2905, opened 2017-09-20
[segwit] wrong tx sizes and fees


lzsaver:
Electrum 3.0 master branch git 77eb62b

Electrum still counts tx sizes with the witness. But it does not go to the blockchain. So we should not take it into account, right? If so - we should count fees without the witness too.

PR #2907, opened 2017-09-21
Moving to virtual tx size instead of total tx size

SomberNight:
This is an attempt at migrating calculations and text fields from using old (total) transaction sizes to virtual transaction sizes.

See #2905


SomberNight:
@ecdsa The change is now less intrusive, as agreed.

ecdsa:
merged manually

Issue #2880, opened 2017-09-16
address_to_scripthash needs to be updated


lzsaver:
```
electrum/bitcoin.py, line 361, in address_to_scripthash
return bytes(reversed(h)).hex()
AttributeError: 'bytes' object has no attribute 'hex'
```

ecdsa:
use python3. see release notes.

wakiyamap:
bytes.hex is newer python3.5
https://docs.python.org/3.5/library/stdtypes.html#bytes.hex

use higher python3.5

lzsaver:
wakiyamap, thanks.

I used python3.4. It was not clear that I have to use python3.5. We need to added this info to the RELEASE-NOTES file.

SomberNight:
Shouldn't we do something with this?
E.g. change
https://github.com/spesmilo/electrum/blob/7b2ff57d0cafba6afa6f77257176e7c77947cada/setup.py#L14-L15

PR #2910, opened 2017-09-22
Require Python 3.5

lzsaver:
Electrum 3.0 uses Python 3.5 features.


SomberNight:
See #2880

Note however that python 3.4 is referenced in multiple locations in the code, e.g.
```
setup-release.py
contrib/build-wine/prepare-wine.sh
```
and possibly others.

ecdsa:
no we should not do thar

PR #2911, opened 2017-09-23
migration to PyQt5

SomberNight:
This would greatly benefit from some testing. Basic functionality seems to work on my machine.

I could not always see a clean way of migrating old-style signals to new-style signals, hence the occasional ugliness.

At the very least, the following is missing:
- the Windows building stuff in `./contrib/build-wine/*`
- updating `./snap/snapcraft.yaml` (also for py3)

### DIFF
--- a/gui/qt/qrcodewidget.py
+++ b/gui/qt/qrcodewidget.py
@@ -97,6 +97,7 @@ class QRDialog(WindowModalDialog):
 
         vbox = QVBoxLayout()
         qrw = QRCodeWidget(data)
+        qscreen = QApplication.primaryScreen()
         vbox.addWidget(qrw, 1)
         if show_text:
             text = QTextEdit()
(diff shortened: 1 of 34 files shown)


gui/qt/qrcodewidget.py
```diff
@@ -109,12 +111,12 @@ def __init__(self, data, parent=None, title = "", show_text=False):
             filename = os.path.join(config.path, "qrcode.png")
 
             def print_qr():
-                p = QPixmap.grabWindow(qrw.winId())
+                p = QScreen.grabWindow(qrw.winId())
```
SomberNight: This does not work it seems. (Same for the one below) Needs a fix.
```
TypeError: grabWindow(self, sip.voidptr, x: int = 0, y: int = 0, width: int = -1, height: int = -1): first argument of unbound method must have type 'QScreen'
```

SomberNight: This does not work it seems. (Same for the one below) Needs a fix.
```
TypeError: grabWindow(self, sip.voidptr, x: int = 0, y: int = 0, width: int = -1, height: int = -1): first argument of unbound method must have type 'QScreen'
```

SomberNight:
As can be seen from all the commits, I keep finding changes that need to be done.
I would be glad if someone else could also spend some time trying to get the GUI to crash (or produce errors).
In any case, I think this PR is more-or-less stable now.

gui/qt/qrcodewidget.py
Wullee: @@ -109,12 +111,12 @@ def __init __ (ตัวเองข้อมูล parent = ไม่มี, title = "", show_text = เท็จ):
             filename = os.path.join (config.path, " qrcode.png " )
 
             def  print_qr ():
+ p =  QScreen .grabWindow (qrw.winId ())

PR #2916, opened 2017-09-23
Update Windows build to Python 3.5 and PyQt5

bauerj:
- PyQt5 needs `sip`, which is not available for Python 3.4
- Python3.5 doesn't provide a single msi installer
- I changed `pywin32` to install from pypi so there are less buttons to click
- While I was at it, I updated NSIS too


bauerj:
I tested it on Wine and it seems to work.

I wasn't able to test it on Windows yet.

If someone wants to test: [electrum-2.9.3-190-gffdafc57.zip](https://github.com/spesmilo/electrum/files/1326678/electrum-2.9.3-190-gffdafc57.zip)

wakiyamap:
I tested on Windows10 64bit.
run it . nice work .

PR #2931, opened 2017-09-26
Create informantion

lulodb75:
are two days that electrum does not send a send .. it is locked in a triangular signal with an esclemative point and says "Replaceable" what can i do? thanks to who responds

### DIFF
--- a/informantion
+++ b/informantion
@@ -1,0 +1,1 @@
+are two days that electrum does not send a send .. it is locked in a triangular signal with an esclemative point and says "Replaceable" what can i do? thanks to who responds


lzsaver:
[Electrum FAQ: My transaction has been unconfirmed for a long time. What can I do?](http://docs.electrum.org/en/latest/faq.html#my-transaction-has-been-unconfirmed-for-a-long-time-what-can-i-do)

Try these methods:
- [preferredminer.com](http://preferredminer.com/)
- [transaction broadcasting](https://en.bitcoin.it/wiki/Transaction_broadcasting)

Please do not commit your questions, use [forums](https://electrum.org/#community).

Issue #2897, opened 2017-09-19
Online addresses + offline seed signing setup does not work


goblin:
I'm having problems with grayed out "Sign" button in 2.9.3. Haven't tried earlier versions.

My scenario:
1. On an offline machine, I created wallet_1 with a seed. Then I copied the first generated bitcoin address to an online machine.
2. I prepared the transaction on the online machine (watch-only wallet_2 with 1 imported bitcoin address)
3. Finally I copied the transaction over back to the offline machine and loaded it from file. The "Sign" button is grayed out.

To actually sign the transaction on the offline machine I had to export the private key from the first address in wallet_1, and then to create a new standard wallet_3 using this (imported) private key. Only on wallet_3 could I sign the transaction.

As I understand it, the "Sign" button may be grayed out because Electrum doesn't know how far in the HD hierarchy to search for the key. However, it should at least scan the first ones that it's displaying in the Address tab. Preferably, if there's no key there, there should be an option to specify the HD index manually when displaying an unsigned transaction.

ecdsa:
you cannot combine online addresses and offline seed, for the reason you explained.

you need one of those:
 - online xpub, offline seed
 - online addresses, offline set of import private keys

goblin:
Could this be left open as an enhancement proposal? I might consider implementing it myself.

The problem with online xpub is that Electrum only allows a single xpub key, so you cannot import multiple keys this way.

One problem with offline set of imported private keys is convenience: why should the user have to create more wallets than necessary to accomplish a simple task of signing a transaction? But the other problem is more severe: on the offline wallet, once you run out of original 20-or-so generated addresses, there's no visible way to create/export more of them. Allowing the user to specify an HD index manually when signing would make it easier. Another option would be to have a button to create more addresses.

goblin:
Yet another good solution for me would be for Electrum to allow to import multiple public EC keys (as opposed to a single xpub) in the online wallet.

ecdsa:
I don't see how you could possibly do this. 
unless you want the user to manually add the HD index, and I believe that would not be a feature.

goblin:
I'll have to familiarize myself with the code a bit more to know the details.

But if the user can sign the transaction by copy&pasting keys around wallets, surely it can be automated instead.

goblin:
I mean the human-language algorithm would be to:
1. iterate over the available addresses (as shown in the Addresses tab) and check if any of them matches an input on the transaction being previewed
[...]

PR #2933, opened 2017-09-26
initial fix for #2897

goblin:
This is my initial proposal for fixing #2897, which was closed prematurely IMO (please reopen until this is merged).

This allows offline-signing a transaction that was prepared using a watch-only address-based online wallet.

Generating the 40 addresses is a bit slow and this could surely be improved, but I'm not sure where the generated keys are stored -- and particularly how to access them from the Keystore. I'll be happy to fix this if you point me at the right direction.

I have successfully signed a testnet transaction using this patch (from 0th Receiving address), but this probably needs more testing than that. I'll be happy to have a look at how tests work too, if that's what you want.

### DIFF
--- a/lib/keystore.py
+++ b/lib/keystore.py
@@ -267,7 +267,18 @@ class Xpub:
         assert len(s) == 2
         return xkey, s
 
+    def get_pubkey_derivation_from_addr(self, x_pubkey):
+        _, needle = xpubkey_to_address(x_pubkey)
+        for c in range(0, 2):
+            for i in range(0, 20):
+                _, candidate = xpubkey_to_address(self.derive_pubkey(c, i))
+                if needle == candidate:
+                    return [c, i]
+        return
+
     def get_pubkey_derivation(self, x_pubkey):
+        if x_pubkey[0:2] == 'fd':
+            return self.get_pubkey_derivation_from_addr(x_pubkey)
         if x_pubkey[0:2] != 'ff':
             return
         xpub, derivation = self.parse_xpubkey(x_pubkey)


goblin:
No idea why the build failed. How do I run the tests on my box?

bauerj:
That's unrelated to your PR.

goblin:
I've rebased my fix onto the latest master

ecdsa:
I disagree with that idea. If it is not possible to find the derivation in the general case, we should not try to do it for the first generated addresses. It is better to let users know right away that this combination does not work, instead of creating a system that will work for a while, and then inexplicably stops working.

goblin:
OK, thanks for the explanation.

I believe it *is* possible to find it in the general case, with user input. It just needs to be presented slightly differently. Firstly, user would need to be aware which HD derivation the original address came from (so these would need to be displayed along with addresses, maybe in the Details view). Secondly, there'd need to be an option to enter that derivation in the Transaction Preview page. And presumably also having a button to "generate more derivations" would be useful too.

It's a bit more work, but I may consider doing it. If you really disagree with this (I do understand it's only usable for more advanced users), I guess I'll be keeping a fork.

ecdsa:
I think it is better to educate users to do the right thing. 
maybe we can add that info in the wizard

ecdsa:
but yes, I disagree with introducing complicated features when they can be avoided

goblin:
So what's the right thing here to do for a user?

My use case is I prepare a lot of master HD keys offline. Then I export their first bitcoin address to a watch-only wallet. When I want to spend from an address, I prepare the transaction on an online watching-only wallet, and then want to be able to sign the transaction on an offline one.

So currently to do this, I need to create an intermediate wallet: either
1. on the offline machine, manually using the key from the HD wallet (and this is impossible to do with a key with ID greater than ~20), or
2. on the online one, by using the public part of the master HD key

Option 2. is better, and would actually work quite well if it was possible to import a lot of master public keys (rather than bitcoin addresses) to a watch-only wallet. I wouldn't need an intermediate wallet in this case.

So maybe that's the better way to go... What would you say to adding support for importing multiple master public keys in a watch-only wallet?

ecdsa:
why can't you have several watching-only wallets?

goblin:
@ecdsa I'm talking dozens, maybe even hundreds or master keys... it's impractical to have so many separate wallets

ecdsa:
and you only watch one address per wallet?

goblin:
Yes, however being able to watch more than 1 would also be nice. So implementing multiple xpubs per wallet would actually be even better for me than my original patch.

ecdsa:
why don't you create two imported wallets, one with the addresses you want to watch, the other with the privkeys?

goblin:
That's a good point. The reason in my case is security. I don't want the computer which signs the transaction to have access to any more private keys than necessary.

I generate the all the addresses on a computer which doesn't have any connectivity (not even via USB pendrives or such). I copy the required privkey (actually a wallet seed) via screen/keyboard to another offline one which then signs the transaction and stores it on USB.

ecdsa:
well, in that case create many wallets with one privkey in each...

goblin:
Yup, that's more or less what I'm doing. I could copy the private key over the screen/keyboard instead of a wallet seed.

But having a full HD wallet is more handy because it then allows to store change on a new address derived from the master key. Of course I now realize the online computer can't create such a transaction from just the bitcoin address, so it needs to know the xpub anyway.

I guess I'll just be copying over the xpub wallet when I want to spend from an address, but being able to import all of them at once to have in one window would be cool.

PR #2943, opened 2017-09-27
tests: message signing

SomberNight:
This would have caught the bug fixed in #2941


SomberNight:
Ironically enough, the Travis tests are failing because they already include this new test, but not the fix from the referenced PR.

Issue #2961, opened 2017-10-02
Address generation in p2sh-segwit wallets is broken


SomberNight:
On git master 031b911dce818c4c005663e770b4784234bc5590.

Try restoring a BIP39 seeded wallet with a `m/49'/0'/0'` derivation path. When the wallet would try to generate its p2sh-segwit receive addresses, Electrum will crash.

```
[profiler] load_transactions 0.0000
[profiler] build_reverse_history 0.0000
[profiler] check_history 0.0000
[profiler] on_update 0.0001
[ElectrumWindow/test_bip39_49_treat] Notifying GUI
[ElectrumWindow/test_bip39_49_treat] using default geometry
Traceback (most recent call last):
  File "/home/user/wspace/electrum/lib/util.py", line 139, in run_jobs
    job.run()
  File "/home/user/wspace/electrum/lib/synchronizer.py", line 189, in run
    self.wallet.synchronize()
  File "/home/user/wspace/electrum/lib/wallet.py", line 1575, in synchronize
    self.synchronize_sequence(False)
  File "/home/user/wspace/electrum/lib/wallet.py", line 1565, in synchronize_sequence
    self.create_new_address(for_change)
  File "/home/user/wspace/electrum/lib/wallet.py", line 1554, in create_new_address
    address = self.pubkeys_to_address(x)
  File "/home/user/wspace/electrum/lib/wallet.py", line 1708, in pubkeys_to_address
    return bitcoin.pubkey_to_address(self.txin_type, pubkey)
  File "/home/user/wspace/electrum/lib/bitcoin.py", line 342, in pubkey_to_address
    scriptSig = transaction.p2wpkh_nested_script(pubkey)
NameError: name 'transaction' is not defined
```

Naively adding `from . import transaction` at the top of `bitcoin.py` will not work, I believe due to making imports circular, it then fails with:
```
$ ./electrum -v
Traceback (most recent call last):
  File "./electrum", line 91, in <module>
    imp.load_module('electrum', *imp.find_module('lib'))
  File "/usr/lib/python3.5/imp.py", line 244, in load_module
    return load_package(name, filename)
  File "/usr/lib/python3.5/imp.py", line 216, in load_package
    return _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 673, in exec_module
[...]

PR #2962, opened 2017-10-02
fix: p2sh-segwit address generation

SomberNight:
See #2961

Perhaps a bit ugly to `import electrum` "inside electrum" but other fixes would likely involve restructuring imports.


ecdsa:
no, you cannot do that

PR #2964, opened 2017-10-03
tests: wallet-keystore integrity (bip39)

SomberNight:
This fails on current master 031b911dce818c4c005663e770b4784234bc5590 (which clearly means it's needed :P), due to #2961. Works with #2962 applied.

Note that I obtained the constants the asserts compare against in
- test_bip39_multisig_seed_bip44_standard()
- test_bip39_multisig_seed_bip49_p2sh_segwit()

from Electrum itself, hence they might not be correct... Would be nice to get an independent confirmation. Still, even without that, these tests can be used to confirm self-consistency.

### DIFF
--- a/lib/base_wizard.py
+++ b/lib/base_wizard.py
@@ -302,10 +302,7 @@ class BaseWizard(object):
         self.on_keystore(k)
 
     def on_bip43(self, seed, passphrase, derivation):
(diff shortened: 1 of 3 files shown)


SomberNight:
(no longer fails on current master 3d54011c5d53c405ad22ace08e038bf3ec327e91)

PR #2979, opened 2017-10-04
Fixed syntax error with utf-8 encoding argument

jrkong:
…ead of "utf-8". This will cause the reader to read the files using default encoding rather then utf-8.


SomberNight:
For future reference, `utf-8`, `utf8`, and `utf_8` should all work.

https://stackoverflow.com/questions/35383506/difference-between-encoding-utf-8-and-utf8-in-python-3-5
https://docs.python.org/3/library/codecs.html#standard-encodings

Besides, there are 20+ other uses of `utf8` throughout the project.

PR #2996, opened 2017-10-05
digitalbitbox: import mobile pairing config

benma:
- menu option to verify addresses securely
 - p2pkh transaction verification

Next steps: p2sh tx verification and full 2FA.

### DIFF
--- a/plugins/digitalbitbox/digitalbitbox.py
+++ b/plugins/digitalbitbox/digitalbitbox.py
@@ -5,7 +5,8 @@
 
 try:
     import electrum
-    from electrum.bitcoin import TYPE_ADDRESS, var_int, msg_magic, Hash, verify_message, pubkey_from_signature, point_to_ser, public_key_to_p2pkh, EncodeAES, DecodeAES, MyVerifyingKey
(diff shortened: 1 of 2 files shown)


ecdsa:
@douglasbakkum can you review this? the code includes a function that pushes a payload to a digitalbitbox URL

plugins/digitalbitbox/digitalbitbox.py
```diff
@@ -207,6 +217,45 @@ def seed_device_dialog(self):
                 return
         self.isInitialized = True
 
+    def mobile_pairing_dialog(self):
+        dbb_user_dir = None
+        if sys.platform == 'darwin':
```
benma: I tested only on linux. Would be good to do test the other platforms as well.

douglasbakkum:
Tested ACK.

@ecdsa This is a useful addition because it allows using our mobile app to verify transaction data (as is done with our desktop app currently). The payload goes to our relay server (https://github.com/digitalbitbox/smart-verification-server) that is used for communicating to the mobile app, with the communication encrypted for user privacy. (FYI @benma is a member of our company.)

PR #2999, opened 2017-10-05
 Fix low value error message 

gdvine:


gdvine:
Oops I think I did this wrong

Issue #3021, opened 2017-10-11
'Imported_KeyStore' object has no attribute 'get_xpubkey'


qirtaiba:
Under the latest Electrum 3.0 from github when I try to send from either the command online or the text GUI, I get "'Imported_KeyStore' object has no attribute 'get_xpubkey'".

SomberNight:
What type is the wallet? `type(wallet)`
I don't think the problem is related to command-line/text UI usage.

qirtaiba:
Multisig, imported

ecdsa:
how did you create that wallet?

qirtaiba:
To use it with Electrum I restored it using the private key. I originally created it using coinb.in if I recall correctly.

ecdsa:
how did you restore a multisig wallet with "the private key" ?

qirtaiba:
Eh you're right I guess it's not a multisig wallet is it?  It's just a normal wallet.

SomberNight:
Again, could you check either the output of `type(wallet)` from the `Console` tab in the QT GUI, or the value of `wallet_type` in the wallet file?

If it is `Standard_Wallet` / `standard`, then the v14 conversion did not run for some reason; and if it is `Imported_Wallet` / `imported`, then either something went horribly wrong, or you are not on latest master.

qirtaiba:
It's imported, but I am on the latest master. To be sure, I installed from source again just five minutes ago.

ecdsa:
v14 conversion requires load_wallet to be called, which is not the case if he uses the command line

qirtaiba:
seed_version says 14

ecdsa:
can you paste the content of the wallet file (obfuscate whatever is private)

qirtaiba:
OK here it is and I made a mistake! It says "wallet_type: standard". I mistook "keystore type: imported" for "wallet_type".

`{`
`    "addr_history": {`
`        "13FEd6gUFvhk--------SyA65LfABTU7Dw": [`
`...`
`        ]`
`    },`
`    "addresses": {`
`        "change": [],`
`        "receiving": [`
`            "13FEd6gUFvhk--------SyA65LfABTU7Dw"`
`        ]`
`    },`
`    "keystore": {`
`        "keypairs": {`
`            ---`
`        },`
`        "type": "imported"`
`    },`
`    "pruned_txo": {},`
`    "seed_version": 14,`
`    "stored_height": 488434,`
`    "transactions": {`
`...`
[...]

PR #3025, opened 2017-10-12
storage: upgrade format version in storage.load_data()

SomberNight:
I have found that there are quite a few places in the code where a `WalletStorage` is created. It is evidently error-prone to rely on the programmer to remember to call `storage.upgrade()` in every case. See #3021

This is an alternative solution.

Another solution would be to call `storage.upgrade()` in the constructor of `Abstract_Wallet`.
Even another could be to call `storage.upgrade()` in `storage.put()` and in `storage.get()`; but I think that needs serious locking (~2 locks and maybe even thread.local variables?)...

### DIFF
--- a/gui/qt/installwizard.py
+++ b/gui/qt/installwizard.py
@@ -243,7 +243,7 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
                     return
 
         path = self.storage.path
-        if self.storage.requires_split():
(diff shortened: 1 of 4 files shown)


SomberNight:
squashed and rebased

SomberNight:
Made changes so that this can possibly be used as basis for agreed upon solution.
- removed optional param from WalletStorage constructor
- split_accounts() and upgrade() are now called automatically

PR #3026, opened 2017-10-12
fix: missing import 'stat' in network.py and simple_config.py

SomberNight:
```
$ ./electrum -v
  File "./electrum", line 364, in <module>
    config = SimpleConfig(config_options)
  File "/home/user/wspace/electrum/lib/simple_config.py", line 74, in __init__
    self.path = self.electrum_path()
  File "/home/user/wspace/electrum/lib/simple_config.py", line 96, in electrum_path
    os.chmod(path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
NameError: name 'stat' is not defined
```

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -29,6 +29,7 @@ from __future__ import unicode_literals
 import time
 import queue
 import os
+import stat
 import errno
 import sys
 import random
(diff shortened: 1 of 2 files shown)


ecdsa:
there might be an issue with android here..
I forgot about that when I merged #2320

PR #3047, opened 2017-10-16
fix: storage upgrade of pre-2.8 imported key wallets, and tests

SomberNight:
- fixed upgrading of pre-2.8 wallets that store imported private keys
- added tests for single-account Trezor wallets
- added tests for multi-account Trezor wallets
- added tests for 2-of-2 multisig wallets (one seed, one xpub)
- added tests for address-watching wallets

The number of tests is probably overkill, there hadn't been a storage structure change between every major version of the client. However I've found it harder to see when there were changes than to simply add more tests.


SomberNight:
Hmm, I see the Travis test failed.
`python3 -m unittest discover -s lib/tests` worked locally
Let me see...

SomberNight:
[The Travis build still fails](https://travis-ci.org/spesmilo/electrum/jobs/288388939), currently with
`ImportError: No module named 'PyQt5'`

I need parts of the Trezor plugin loaded to be able to test Trezor wallets... Currently I load the Trezor plugin for this reason (and actually, because it was a lot easier, all the plugins), but as the existing unit tests did not do anything similar, dependencies are missing from the environment of the Travis build.

PR #3061, opened 2017-10-18
[QT] reuse items in address list

### LINKED ISSUE #2968: [3006932] QTreeWidgetItem
```
  File "/home/user/electrum/lib/python3.5/site-packages/electrum_gui/qt/address_list.py", line 133, in <lambda>
    menu.addAction(_("Copy %s")%column_title, lambda: self.parent.app.clipboard().setText(item.text(col)))
RuntimeError: wrapped C/C++ object of type QTreeWidgetItem has been deleted
```

### LINKED ISSUE #3060: [QT] Address list buttons/refresh bugged
On git master 0a648e2b1cdb7852d80c0626c62984ef01c1c7e9.

Using the QT GUI,
- open the `Addresses` tab
- change the size of the window so that all your addresses fit inside (potentially by reducing gap limit beforehand)
- make sure there is no scrollbar after previous step
- select an address (e.g. via left-click on mouse)
- click `Receiving`

Expected behaviour:
- now the `change` addresses will be displayed

Actual behaviour:
- the address list does not get properly updated, and almost all shown addresses will still be `receiving` addresses

This is on Debian 9.2
```
Qt version: 5.7.1
SIP version: 4.18.1
PyQt version: 5.7
```

SomberNight:
This fixes #2968, as the references to items will remain valid across updates.
Also fixes #3060.

I've found #3060 to be related to `QTreeWidget.clear()`. When I had initially tried to call
`self.clear()` inside `on_update()` (in the case of `clear==True`), #3060 occured, but without calling `self.clear()`, it does not.

### DIFF
--- a/gui/qt/address_list.py
+++ b/gui/qt/address_list.py
@@ -68,7 +68,7 @@ class AddressList(MyTreeWidget):
     def toggle_change(self):
         self.show_change = not self.show_change
         self.set_change_button_text()
-        self.on_update()
+        self.on_update(clear=True)
 
     def set_change_button_text(self):
         s = [_('Receiving'), _('Change')]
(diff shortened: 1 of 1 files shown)


ecdsa:
at this point I cannot reproduce #3060. I do not like this pull request because it seems to add a lot of complexity just for #2968 

Another point is that we still use a child class of QTreeWidget, because the address tab used to be rendered as a tree. Now that we render it as a list, we should derive our lists from QListWidget, it would reduce complexity. I just did not have time to d it yet.

Before fixing #2968 and #3060, I would like to know if they still occur when we stop using QTreeWidget

PR #3067, opened 2017-10-19
kivy - replace requests tab by address

ariard:


ecdsa:
this crashes because address.kv is missing
did you rename requests.kv ?

PR #3074, opened 2017-10-20
build scripts: https

SomberNight:


bauerj:
What's the advantage of changing from `git://` to `https://` URLs?

PR #3077, opened 2017-10-20
Update commands: output str(Decimals), not floats

danuker:
The commands `listunspent` and `history` were using float.
Patched to consistently return strings of Decimal objects.


ecdsa:
thanks

danuker:
Cheers! :beer:

PR #3079, opened 2017-10-20
building on wine: check hash of NSIS, and GPG sig of Python

SomberNight:
Despite the "not so simple" look, this should be low maintenance.
Whenever the NSIS URL is updated, the committer should change the corresponding hash; and in case of a signature verification failure for the Python binaries, one should check if they added a new release-signing-key.

While all the `pip` downloads and `setuptools` downloads are still there as is, I guess the point of this PR is to reduce the attack surface.

### DIFF
--- a/contrib/build-wine/prepare-wine.sh
+++ b/contrib/build-wine/prepare-wine.sh
@@ -2,6 +2,7 @@
 
 # Please update these carefully, some versions won't work under Wine
 NSIS_URL=https://prdownloads.sourceforge.net/nsis/nsis-3.02.1-setup.exe?download
+NSIS_SHA256=736c9062a02e297e335f82252e648a883171c98e0d5120439f538c81d429552e
 PYTHON_VERSION=3.5.4
 
 ## These settings probably don't need change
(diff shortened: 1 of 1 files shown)


contrib/build-wine/prepare-wine.sh
```diff
@@ -11,6 +12,30 @@ export WINEPREFIX=/opt/wine64
 PYHOME=c:/python$PYTHON_VERSION
 PYTHON="wine $PYHOME/python.exe -OO -B"
 
+
+# based on https://superuser.com/questions/497940/script-to-verify-a-signature-with-gpg
+verify_signature() {
+    local file=$1 keyring=$2 out=
+    if out=$(gpg --no-default-keyring --keyring "$keyring" --status-fd 1 --verify "$file" 2>/dev/null) &&
+       echo "$out" | grep -qs "^\[GNUPG:\] VALIDSIG "; then
```
dabura667: This doesn't work on my machine as my locale is JA, and the output is all in Japanese.

Can there be an alternative method?

dabura667: perhaps run gpg command with `LANG=en_US.UTF-8 ` before it? @SomberNight @ecdsa

dabura667: (If you run with LANG=ja_JP.UTF-8 you can see what I see)

Issue #2838, opened 2017-09-04
Feature request: show fee in fiat in Send tab


charleshsu168:
I am using 2.8.3
please include option to show fee in USD next to the slider bar.
currently only showing fee in btc. not very helpful for those of us who think fee in terms of dollars. 
thank you.
![untitled](https://user-images.githubusercontent.com/5799190/30038724-a911eda6-917c-11e7-93cd-480a55e02173.png)

pmukherjeee:
Yeah, this is useful. I second this.

cdodd:
Or more usefully, display it in the fiat currency that the user has selected in the preferences.

ecdsa:
the fiat fee is already shown in the confirmation dialog.
how many times should it be shown?

Engelberg:
@ecdsa Ideally, it should be shown on the page with the slider, where you are selecting what level of fee to select, before you click send.  So putting the fiat fee as specified in the picture above would be superior to putting it in the confirmation dialog (which the user has no reason to assume will appear after hitting send, with a fiat fee given).

Engelberg:
My preference would be to show the fiat fee to the right of the bitcoin fee in the above photo, rather than having it *instead* of the bitcoin fee.  Both are useful to see.

pydlv:
@ecdsa I don't see any downside to showing the fee in fiat on this screen. Support.

fresheneesz:
I also want to reference this feature request, which partially duplicates this one: https://github.com/spesmilo/electrum/issues/2788 . The alignment of the amount and fee fields is pretty confusing. At very least, the fee slider and the BTC fee field should swap positions.

jonathancross:
Displaying the fee in fiat (next to the fee input) would be useful to prevent nasty surprises, but I'd suggest making it read-only.  Proper fees for a bitcoin transaction can only be calculated in BTC. If the fee seems too high, the user might choose to simply delay a transaction until there is less network congestion.

leafcutterant:
> the fiat fee is already shown in the confirmation dialog.
> how many times should it be shown?

@ecdsa It should be shown in both places because this way one has to continuously press "Send" and then go back  just to adjust fees. Other than being time-consuming, this is also frustrating and unnerving because one doesn't like to hit "Send" on something that they don't know the full costs of. This especially applies to new users who don't know how Electrum works. Experienced ones will know that fees will be displayed later and that a password entry will be needed before the transaction is sent, but even then it's unnerving.

[...]

PR #3098, opened 2017-10-23
reorganize fee line in send tab

janhn:
This is an attempt to improve how fee is presented in the send tab and implement #2838.

Changes:
- reorder to line up the columns with Amount
  [ amount, optional fiat, controls ]
  [ fee, optional fee in fiat, controls ]
- make fee amount always visible (regardless of whether editable in settings)
- visually distinguish editable and non-editable with border & color
- fix Clear button to respect the editable setting

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -1060,6 +1060,17 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
               + _('A suggested fee is automatically added to this field. You may override it. The suggested fee increases with the size of the transaction.')
         self.fee_e_label = HelpLabel(_('Fee'), msg)
 
+        self.fee_e = BTCAmountEdit(self.get_decimal_point)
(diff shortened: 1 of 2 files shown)


bauerj:
Can you post a screenshot?

janhn:
@bauerj sure: screenshot with fiat enabled and fee amount editable. 

send_tab_fee

lzsaver:
What about to merge?

ecdsa:
@lzsaver I do not like it, I will not merge it. 

I think the slider should be on a separate line from the fee entry.
the line with the fee entry should be optional, and display 3 things:
[tx_size] x [fee_rate] = [fee]
with fee_rate and fee editable

that way users will stop asking why the fee is so high when they spen small inputs

ecdsa:
also, the "replaceable" checkbox will go away from that tab

SomberNight:
What @ecdsa describes above has now been implemented in master.

Issue #3101, opened 2017-10-23
Importing private keys takes way too long


daneos:
I have a file with over 8000 private keys backed up from numerous versions of Satoshi client using `dumpwallet` command.
I'm trying to create new Electrum wallet from this file.
It's currently running for over 24h, constantly using 100% of one processor core and I have no way of telling if it just hanged or if it will finish someday.

Electrum version is 2.9.3 running on Linux Mint 17.3.

fr3aker:
I noticed this problem, too, in an earlier version. It appears to be a GUI problem. It seems the complexity of adding private keys scales exponentially with the amount of keys added. Adding them in batches of 500 worked but adding even a single more key to the text field would again cause Electrum to hang for a couple of seconds.

PR #3106, opened 2017-10-24
Utility script for bulk import of private keys

daneos:
This is a workaround for #3101.
The script `privkey2electrum` can be used to quickly create import wallet with large number of private keys.
Script takes two arguments, first being a text file with privkeys, one per line, and second being the path for newly created wallet.
I was able to import over 8000 keys in just over 20min.

I'm totally new to this project and I couldn't find the problem in the source, so this is not a complete solution, only the workaround.

Tested on Linux Mint 17.3 with Python 3.4.3 and Electrum 3.0 (latest master).

### DIFF
--- a/privkey2electrum
+++ b/privkey2electrum
@@ -1,0 +1,38 @@
+#! /usr/bin/env python3
+
+import sys
+from time import time
+from electrum import WalletStorage, Wallet
+from electrum.keystore import Imported_KeyStore
+
+
+if __name__ == "__main__":
+	if len(sys.argv) < 3:
+		print("USAGE: %s <source_file> <wallet_file>" % sys.argv[0])
+		print("       source_file should containt a list of private keys, one key per line")
+		print("       wallet_file is a path to your new wallet")
+		sys.exit()
+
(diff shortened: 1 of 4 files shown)


privkey2electrum
```diff
@@ -0,0 +1,40 @@
...
+import sys
+from time import time
+from electrum import WalletStorage, Wallet
+from electrum.keystore import Imported_KeyStore
+
+
+if __name__ == "__main__":
+	if len(sys.argv) < 3:
+		print("USAGE: %s <source_file> <wallet_file>" % sys.argv[0])
+		print("       source_file should containt a list of private keys, one key per line")
```
dabura667: Contain

setup.py
```diff
@@ -79,7 +79,7 @@
             'locale/*/LC_MESSAGES/electrum.mo',
         ]
     },
-    scripts=['electrum'],
+    scripts=['electrum', 'privkey2electrum'],
```
dabura667: Leave this out and move the script to the scripts folder.

PR #3111, opened 2017-10-24
fix: file not found if both testnet and local dir options

ariard:
`electrum -P --testnet create` if not already a local dir then FileNotFoundError


ecdsa:
janus confirmed this.
however, the fix should not rely on 'portable' because the bug aklso occur with --dir

PR #3113, opened 2017-10-25
efficiency: private key imports, address imports

SomberNight:
See #3101 

Importing a single private key without this PR takes linear time in the size of the keystore (~number of already imported private keys), so importing n private keys has quadratic time complexity.
With this PR, I think, importing a single private key takes amortized constant time (but certainly sub-linear).

-----

With this PR, in a VM, for me, importing
- 500 private keys takes 35-40 seconds -- instead of 40-45 seconds
- 5000 private keys takes ~7.5 minutes -- instead of ~15 minutes
- 10000 private keys takes ~16 minutes -- instead of ~44 minutes
- 5000 addresses takes 50-60 seconds -- instead of ~3.5 minutes

Hmm. I was expecting a larger improvement...

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -2364,7 +2364,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         self.warn_if_watching_only()
         self.show_transaction(tx)
 
-    def _do_import(self, title, msg, func):
+    def _do_import(self, title, msg, import_func, finalize_func):
         text = text_dialog(self, title, msg + ' :', _('Import'))
         if not text:
             return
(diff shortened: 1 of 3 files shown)


ecdsa:
this looks unnecessarily complicated.
I think things would be better encapsulated if wallet had a method to import many keys/addresses

arnezami:
Hi,

I tried this patch, but in any case, import opertion takes a lot of time. In my wallet I have ~52345 imported addresses. Yesterday I tried to import ~4000 addresses and this took ~3.5 hours. When I push Import button, the application is hang and I can't do nothing.

My suggestion is to change approach. We can make import like the task on background and make this on another thread. In the list of the addresses we can show special icon near of each address that is in the proccess of importing. This will give the possibility to continue work with wallet and don't wait when import will be finished. In application bottom bar we can show import proccess bar  {imported _count} / {total}. What do you think about it?

SomberNight:
Try now on master.
Importing ~10k addresses should only take a few seconds now.
If you have libsecp256k1 installed, same for private keys; otherwise python-ecdsa becomes the bottleneck.

SomberNight:
but note that most electrum servers have a limit of subscribing to 50k addresses per session

Issue #2857, opened 2017-09-09
Invalid TX generated by v2.9.3 (Non-canonical DER signature)


SomberNight:
User "Xeron" on #electrum IRC (on freenode) retrieved a bitcoin private key from an old (2010) wallet.dat using Bitcoin Core 0.14.2.
He then tried to sweep the coins from the privkey using Electrum 2.9.3, using several different servers.
He also tried creating a standard wallet directly from the privkey.

In all cases, when trying to broadcast the signed transaction, he was presented with the error
```
error: The transaction was rejected by network rules. (16: mandatory-script-verify-flag-failed (Non-canonical DER signature))
```

Unsigned TX:
```
0100000002ccae9778042adb58ac388aa0a2463eb0e68ba9e34dfcca115706510e54458b38010000004401ff41049fd7cfe066a814030c538d1ef4cbc412dec86c0215c364bca46ed267165f3cfde66e55db5738bd781ab84f69085a2c27909f20afaeae65f1d341944c5ded3c60fdffffff11e5c3079d60db1a5496bc98ad4d7fe884bae7a20eac8cb2941664bd70546697000000004401ff41046619ac0ae726b306297d85bd5e877d7e9a9922094db131297dc7e3d16b93e6e01620b57c503439a0f03d39735e9fc56c7d2e644ba963f43451b4e47306947e4afdffffff018cdfd2470100000017a91479ef7e9aec1e5943862790344d23843f41b109de8700000000
```

Signed TX:
```
[...]

PR #3125, opened 2017-10-25
try to sweep p2pk outputs from old type WIF privkeys

SomberNight:
See #2857 

Now that the most widely used server implementation, ElectrumX, with [version 1.2](https://github.com/kyuupichan/electrumx/blob/1da8cb547d733fe3da9924a0d31187a6d5bb58a1/README.rst#version-12), properly separates P2PK and P2PKH scripts, we can implement a way for users to spend P2PK outputs.

In this PR, I propose enabling the `sweep` functionality to redeem these coins. Unfortunately, the WIF format for private keys gives no hint if P2PK scripts are to be used. What I do, is search for both P2PK and P2PKH scripts whenever an "old type" (pre-segwit) WIF is swept from. The `compressed` byte is assumed to be correct, the same boolean is used for both script types.

I think the fact that (in some sense) duplicate queries are sent to servers (hence e.g. increasing load/bandwidth), can be overlooked, given sweeping is a rare operation. IMHO it is more important for users to be able to access these coins than to avoid looking up both script types at all costs. So while it might not be desirable to always subscribe to both script types (say) in an imported wallet, this one-time cost can be afforded.

I've tested this on regtest with ElectrumX 1.2.

### DIFF
--- a/lib/bitcoin.py
+++ b/lib/bitcoin.py
@@ -405,6 +405,9 @@ def address_to_script(addr):
 
 def address_to_scripthash(addr):
     script = address_to_script(addr)
+    return script_to_scripthash(script)
+
+def script_to_scripthash(script):
     h = sha256(bytes.fromhex(script))[0:32]
     return bh2u(bytes(reversed(h)))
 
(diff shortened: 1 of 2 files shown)


ecdsa:
note: that will not work on all servers

PR #3137, opened 2017-10-27
tests: have bip39 multisig wallet unit test mirror hw wallet behaviour

SomberNight:
When I wrote the original unit test I did not realise that Electrum uses a different derivation for (multisig) hardware wallets. I think it's nicer to have the unit test mirror what Electrum does. (Though with software bip39 seeds any derivation can be specified.)

### DIFF
--- a/lib/tests/test_wallet_vertical.py
+++ b/lib/tests/test_wallet_vertical.py
@@ -165,20 +165,20 @@ class TestWalletKeystoreAddressIntegrity(unittest.TestCase):
         self.assertEqual(w.get_change_addresses()[0], 'bc1q89s4j4k3ghdmscjz0pklz2fl24mc9ptvwyg38xgqyw8f5vk29ccskzcf2y')
 
     @mock.patch.object(storage.WalletStorage, '_write')
-    def test_bip39_multisig_seed_bip44_standard(self, mock_write):
+    def test_bip39_multisig_seed_bip45_standard(self, mock_write):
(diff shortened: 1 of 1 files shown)


hotbits420:
I AM NOT THE DEVELOPER OR BUGGY BOUNCER TECH, YET ,HOWEVER I AM VERY BUGGED BY BLOCKCHAIN,I SMELL A RAT I M READY TO CALL IT IN A NUTSHELL AND I REFUSE TO LET ONE BAD APPLE POISON THE KEG BUT I PERSONALY HAVE EXPERIENCED THE TRUTH OF THEIR CO.TITLE THE BRUTAL REALITY OF A"BLOCKED CHAIN"THAT NAME VERY LIKELY MAY BE THE FIRST LINE OF DEFENSE OF DEFENCE.PLAINLY IT DESCRIBES WHAT THEY DO,BLOCK THEIR CLIENTS ACCESS TO THE CHAIN   WHILE CONVINCING THEM THEY EARNED FEES FOR  SERVICE ,IT APPEARS THEIR CLIENTS PROVIDE BY RUNNING HARDWARE TO INCREASE MINING ACTIVITY.I WILL NOT LET THIS DESTROY MY SPIRIT FOR THIS IMPORTANT MOVEMENT.I HOPE TO BECOME MOREXREFINED AND FOCUS ON PEOPLE UNDERSTANDING AND NOT HAVING TO WASTE TIME AND GET NEEDLESSLY FRUSTRATED.

PR #3141, opened 2017-10-28
txn: set locktime to local height for RBF, CPFP, sweep

SomberNight:
nLockTime is currently not set when
- sweeping
- using RBF
- using CPFP

I set it analogous to https://github.com/spesmilo/electrum/blob/6f3c822867bd512e276ef63ad7b8c2af2fa40927/lib/wallet.py#L858-L860
as it's already done in case of `make_unsigned_transaction()`.


ecdsa:
this would modify the locktime of an existing RBF transaction.
I do not think you can do that (although I have not tested)

SomberNight:
I've read https://bitcoincore.org/en/faq/optin_rbf/ and [BIP-0125](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki), and couldn't find a specific mention of this.
So I ended up testing it on testnet, where I've successfully got transactions confirmed that replaced another RBF tx with a different locktime. Tested to change the locktime in both directions, and both worked (succeeded in both increasing and decreasing it).

PR #3145, opened 2017-10-29
Add cluelessperon's server to the default lists.

cluelessperson:


ecdsa:
please do not create pull requests for files that are generated automatically.

Issue #3144, opened 2017-10-29
Ledger native segwit test failed


qazasds:
Hello,
I tested Ledger native segwit but it failed. Is it fully supported now? Thank you!

Test environment & steps:
1. Electrum 3.0 (master branch)
2. Ledger Nano S Firmware 1.3.1; Bitcoin App 1.1.10
3. Set env: export LEDGER_NATIVE_SEGWIT=1
4. Create new wallet->Standard Wallet->Hardware Wallet
5. The PATH dialogue popped up, I left it as default (m/44/0) (I thought it should be native segwit related value?)

6. Error message:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/Electrum-3.0-py3.6.egg/electrum_plugins/ledger/ledger.py", line 88, in get_xpub
    xpub = bitcoin.serialize_xpub(xtype, nodeData['chainCode'], publicKey, depth, self.i4b(fingerprint), self.i4b(childnum))
  File "/usr/local/lib/python3.6/dist-packages/Electrum-3.0-py3.6.egg/electrum/bitcoin.py", line 903, in serialize_xpub
    xpub = xpub_header(xtype) + bytes([depth]) + fingerprint + child_number + c + cK
  File "/usr/local/lib/python3.6/dist-packages/Electrum-3.0-py3.6.egg/electrum/bitcoin.py", line 894, in xpub_header
    return bfh("%08x" % XPUB_HEADERS[xtype])
KeyError: 'segwit'

SomberNight:
Hi. Can you try with #3146 ? I don't have a ledger.

qazasds:
Hello,
It works for p2wpkh, thank you!
I am going to do further tests for multisig wallet with ledger native segwit later.

qazasds:
Hello,
I tested 2-of-2 multisig wallet (Ledger + Electrum). If xtype=p2wpkh (#3146), we will get error "Wrong key type p2wpkh". And I changed xtype=p2wsh, it works.
It looks like we need individual xtype for standard wallet (p2wpkh) and multisig wallet (p2wsh).

Thank you!

ecdsa:
are you able to spend from p2wpkh with your ledger?

qazasds:
Hi,
I tested to spend from p2wpkh (Ledger plugin with patch #3146 ) to p2sh-p2wpkh on testnet, and it failed:

[...]

PR #3146, opened 2017-10-29
fix xtype for ledger native segwit

SomberNight:
see #3144


qazasds:
Hi, I just added a comment on #3144. Unfortunately this patch may not work for multisig wallet (p2wsh).

Issue #3181, opened 2017-11-02
Manual/Automatic wallet storage upgrades


SomberNight:
On version 3.0.0.

With the current state of the code, (roughly) storage upgrades are intended to
- prompt the user when using the QT GUI to accept (with the option to cancel)
- when not using the QT GUI, happen silently for "trivial" upgrades
- otherwise, fail with an exception

However, even when using the QT GUI, in many cases, the "trivial" upgrades, by mistake (relative to the intended behaviour) happen silently.

For example:
- if you start Electrum (with QT) and it directly opens a wallet (most recently used one), it upgrades it **silently**
- if you then,  with a wallet already open, do `File`>`Open` to open another wallet, it upgrades it **silently**

- but for example if you have a wallet open, and do `File`>`New/Restore`, and select a wallet, it **will offer** to upgrade
- or if when you start Electrum it can't find the most recently used wallet, then after you select a wallet, it **will offer** to upgrade

PR #3182, opened 2017-11-02
wallet storage upgrade will now happen silently whenever possible

SomberNight:
See #3181

This PR changes the behaviour to consistently offer upgrades to the user when using the QT GUI (in all four scenarios listed in above linked issue). Users of other GUIs/CLIs/daemon should be unaffected by this change, I think, and upgrades should happen for them silently as before.

### DIFF
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -192,7 +192,7 @@ class ElectrumGui:
                 d.exec_()
                 return
             if not wallet:
-                storage = WalletStorage(path)
+                storage = WalletStorage(path, manual_upgrades=True)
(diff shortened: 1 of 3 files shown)


ecdsa:
the intended behaviour is that upgrades that do not split the wallet file should happen silently

SomberNight:
Ok. I'll look into how to change this PR to accomplish that.

SomberNight:
I've pushed another commit.
I believe that with this change,
- when using the QT GUI, non-split upgrades happen silently; splits prompt the user
- when using any other UI, non-split upgrades happen silently; splits fail with an exception telling the user she needs to split the wallet

Issue #3295, opened 2017-11-13
Not prompted to set password when creating imported private key wallet


AbdussamadA:
When you create a wallet using imported private keys you are not prompted to set a password during the wallet creation process. Tested in 3.0.1.

(WIF private keys not xprvs. You are prompted when creating a wallet with xprvs)

SomberNight:
"bug A" should be solved on master now, due to https://github.com/spesmilo/electrum/pull/3346

I haven't tested the others recently, but I did back when the issue was opened, they are probably still TODOs.

AbdussamadA:
Thanks! The other two bugs are still there in 3.0.6. I'm going to name the last one bug c so it's clearer.

PR #3297, opened 2017-11-13
wizard: prompt for password for Imported_Wallet (with privkeys)

SomberNight:
See #3295 - "bug A"
Allow setting password/encryption for imported privkey wallets during creation in wizard.

It is a tiny bit hackish... :)

### DIFF
--- a/lib/base_wizard.py
+++ b/lib/base_wizard.py
@@ -150,9 +150,13 @@ class BaseWizard(object):
         elif keystore.is_private_key_list(text):
             k = keystore.Imported_KeyStore({})
             self.storage.put('keystore', k.dump())
-            self.wallet = Imported_Wallet(self.storage)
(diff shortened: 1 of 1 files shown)


ecdsa:
in another thread, we mentioned that we could add an encryption password to watching-only wallets.
wouldn't it make the present patch obsolete?

SomberNight:
Indeed that is possible. Something similar will probably still be needed.

There is encryption of private keys/xprv/seed (1); and there is encryption of storage (2).
(1) is not applicable to watch-only wallets, clearly. So e.g. imported wallets with addresses, and imported wallets with private keys would still need separate handling/logic.

But yes, that change is likely to affect this code.
I can look into implementing that instead. I probably won't have sufficient time in the next ~2 days, but after that, yes.

SomberNight:
Closing, as #3346 includes this, with minor changes.

PR #3316, opened 2017-11-16
tests: transaction signing

SomberNight:
Not sure how resistant these tests would be to future code changes.

It seems hard to come up with useful transaction signing tests that are zero-maintenance.

In case of some potential serialisation format changes (of unsigned/partial txs), the resulting errors (in the tests) might be fixed by recalculating the new raw transactions from the dictionaries (deserialised txs); or the other way around, if the dictionary format changes... :)


ecdsa:
I wanted to change something in the serialization format: the NO_SIGNATURE byte seems to be useless, I believe it could be replaced with an empty string. I never really had time to verify that it can be done, though.

SomberNight:
this is now too outdated to be useful, and also redundant due to newer tests

PR #3326, opened 2017-11-18
fix: ledger clear_dialog() threading issues

### LINKED ISSUE #3264: Electrum 3.0 crashes when signing transactions with Ledger
For the first time, I used my ledger nano s to send BTC using electrum but after pressing confirm on the device, Electrum freezes and doesn't respond. I tried on both MacOs and Win7, with the same exact behaviour, both running v3.0.1. I encountered no problems when using the ledger wallet.

SomberNight:
fixes #3264

The crash happens at https://github.com/spesmilo/electrum/blob/219a5b02a63a9ee227cff1c83516d003736abf4b/plugins/hw_wallet/qt.py#L161

Delegating the `clear_dialog()` calls to the GUI thread (via signals) seems to fix it. I would appreciate it if someone who understands QT better would enlighten me why... :P

It took me way too long to debug this...

### DIFF
--- a/plugins/ledger/ledger.py
+++ b/plugins/ledger/ledger.py
@@ -238,7 +238,7 @@ class Ledger_KeyStore(Hardware_KeyStore):
         except Exception as e:
             self.give_error(e, True)
         finally:
-            self.handler.clear_dialog()
+            self.handler.finished()
         self.signing = False
         # Parse the ASN.1 signature
         rLength = signature[3]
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks. this is because Qt operations should never be called outside of the GUI thread

Wullee:
Yes, me or wood, what do I do?

Issue #3183, opened 2017-11-02
Enable (optional) encryption for watch-only wallet files


zzAD26KAH59cYpL:
The user should be prompted to set a password to encrypt watch-only wallet files.

The user should also be prompted to set a password to encrypt watch-only wallet files created when an hardware wallet is used with Electrum.

Without encryption for watch-only wallet files the user's privacy is potentially exposed because is possible for an adversary to access the master public keys, this is possibly a security issue too as reported here:

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
> One weakness that may not be immediately obvious, is that knowledge of a parent extended public key plus any non-hardened private key descending from it is equivalent to knowing the parent extended private key (and thus every private and public key descending from it).

zzAD26KAH59cYpL:
Another possible security implication is the possibility for an adversary to spoof addresses editing the watch-only (not encrypted) wallet files.
Maybe this should be addressed for not encrypted wallet files too, possibly checking the single addresses against the public master key when opening a wallet file.

AbdussamadA:
If an adversary has file system access then at that point in time your computer is compromised and it doesn't matter what you do. 

The reason whole wallet file encryption was introduced is to prevent people from receiving money to a wallet they can't spend from. With a watch-only wallet even if you know the password there is no guarantee that you can spend from this wallet since the seed is in another wallet file. 

So overall I don't agree that this is a feature worth adding.

FomgbdsLQeC2LPq:
@AbdussamadA 
> The reason whole wallet file encryption was introduced is to prevent people from receiving money to a wallet they can't spend from. 
Do you care to explain how exactly wallet files encryption can protect users from receiving funds to an address they can't spend from? 

Encryption is by definition a tool to reduce the risks of private information leaks, the master public key is an information that you should keep private as much as possible, and an option to encrypt this information is a simple and useful tool that can help user's to protect their own privacy.
Keep in mind that a watch-only wallet is intended to be used in less secure and maybe shared on-line environments (while using cold storage or hardware wallets to protect your private keys). Being able to (optionally) encrypt a file containing privacy sensitive data can help to mitigate at least a part of the privacy and security problems in this context.

AbdussamadA:
>Do you care to explain how exactly wallet files encryption can protect users from receiving funds to an address they can't spend from?

[...]

Issue #2929, opened 2017-09-25
TREZOR wallets stored in plaintext


schrectacular:
TREZOR wallets are stored in plaintext, with no option to encrypt. This means full public keys are exposed, along with address lists.

I would expect that a TREZOR wallet would be stored encrypted unless the TREZOR was connected. Alternatively, a separate Electrum encryption password option.

lzsaver:
Probably the best option is to encrypt such wallets using Trezor itself.

schrectacular:
I believe other wallets have solved this by allowing two import options for hardware wallets:
- A watch-only wallet for observing transactions, with the option to sign them if the hardware wallet is connected.
- A "regular" wallet import which is only viewable with the device connected.
Food for thought. Your solution sounds fine as well.

ecdsa:
it is not possible using the trezor itself; they have disabled message decryption

ecdsa:
we could encrypt with a password, but it would be unrelated to trezor

lzsaver:
Yes, it does not make sense.

slush0:
> it is not possible using the trezor itself; they have disabled message decryption

@ecdsa We disabled one particular scheme using elliptic curves, but message encryption is clearly possible with TREZOR. Electrum can use for example CipherKeyValue as described in SLIP 11 (https://github.com/satoshilabs/slips/blob/master/slip-0011.md). We use the same for encrypting transaction labels in our wallet (scheme described in SLIP 15).

ecdsa:
@SomberNight see slush's comment

PR #3346, opened 2017-11-21
allow encrypting watch-only wallets

SomberNight:
This PR
- implements #3183: encryption of watch-only wallets
- supersedes #3297: in wizard, prompt for password for Imported_Wallet
- implements #2929: allow encrypting wallet files for hardware wallets
    - but the password is unrelated to the device :(

As it doesn't really make sense to encrypt only the keystore (and not the storage) of watch-only wallets (there isn't even a keystore for imported addresses wallets), I've disabled that option for this case. I am not too happy about how this change (`force_disable_encrypt_cb`) looks in the code, so suggestions welcome.

### DIFF
--- a/electrum
+++ b/electrum
@@ -192,6 +192,8 @@ def init_daemon(config_options):
         print_msg("Type 'electrum create' to create a new wallet, or provide a path to a wallet with the -w option")
         sys.exit(0)
     if storage.is_encrypted():
+        if storage.is_encrypted_with_hw_device():
+            raise NotImplementedError("CLI functionality of encrypted hw wallets")
         if config.get('password'):
             password = config.get('password')
         else:
@@ -236,6 +238,8 @@ def init_cmdline(config_options, server):
     # commands needing password
     if (cmd.requires_wallet and storage.is_encrypted() and server is None)\
        or (cmd.requires_password and (storage.get('use_encryption') or storage.is_encrypted())):
+        if storage.is_encrypted_with_hw_device():
+            raise NotImplementedError("CLI functionality of encrypted hw wallets")
         if config.get('password'):
             password = config.get('password')
         else:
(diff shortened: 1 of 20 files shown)


SomberNight:
@ecdsa I've added commits to implement encryption of wallet files for single-signer hardware wallets as agreed.

The "password" for the encryption is an ecdsa pubkey on a private bip32 derivation path. The derivation path is static and has a BIP43-like "purpose".

Sorry for the less-than-clear intermediate commits, I originally intended those to be temporary until I realised that would make it even harder to review.
I tested this... quite a bit, though I did not repeat all the tests at the current state of the PR.

CLI usage of encrypted hw wallets is not implemented. If we want this, I might need some guidance how to approach.

ecdsa:
thanks. you can squash all your commits into one and rebase.
if you dont known how to do it I will do it before merging

SomberNight:
Ok, done.

SomberNight:
Rebased and fixed conflicts.

chrisrico:
I know this is old but can I ask why only single-signer wallets are supported? Perhaps I'm missing something obvious, but I don't see the concern encrypting my 2-of-2 wallet with a password or hardware keystore.

SomberNight:
Any wallet can be encrypted.
The single-signer restriction is only if you want to encrypt "with a hardware device". For all other wallets, you can encrypt with a password.

From above, encrypting with a hardware device works as follows:
> The "password" for the encryption is an ecdsa pubkey on a private bip32 derivation path. The derivation path is static and has a BIP43-like "purpose".

It gets confusing if you wanted to apply this for a multisig wallet.
How would the user experience look? Say it's a 2-of-2 wallet with a trezor and a ledger. Is one of those "special"? Can you only decrypt the wallet with e.g. the trezor; and it just fails with the ledger? Storing plaintext metadata for a helpful error message would need a different scheme. Or would both work? That needs a different scheme.

This change was already huge, did not want to make it even more complicated.

chrisrico:
>The single-signer restriction is only if you want to encrypt "with a hardware device".

Ah ok, I misunderstood.

>It gets confusing if you wanted to apply this for a multisig wallet.
>How would the user experience look? Say it's a 2-of-2 wallet with a trezor and a ledger. Is one of those "special"?

I'd thinking asking the user which wallet should be used to encrypt is ok. But I think you're forgetting about possibly a more common multi-sig case, where multiple individuals each provide a partial signature. In that case, it's likely there's only one hardware wallet to choose from for encryption.

PR #3351, opened 2017-11-21
better segwit tx size estimation

### LINKED ISSUE #3347: Fee slider is calculating fee lower than specified
For example set slider to 100 sat/byte and sign transaction (1 input, 2 outputs):
> Size: 168 bytes
> Fee: 0.00014236 (84 sat/byte)

Large 16 sat/byte difference.
Expected result - never lower than specified (at least 0.00016800 in this example).

Electrum 3.0.2
P2SH-P2WPKH BIP49 wallet

SomberNight:
fixes #3347

### DIFF
--- a/lib/coinchooser.py
+++ b/lib/coinchooser.py
@@ -91,8 +91,9 @@ class CoinChooserBase(PrintError):
             buckets[key].append(coin)
 
         def make_Bucket(desc, coins):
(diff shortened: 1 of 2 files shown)


SomberNight:
I'll have a look at the failing tests in a few hours.

PR #3361, opened 2017-11-23
Show notifications for incoming transactions

bauerj:
I wanted to implement showing notifications for incoming transactions but quite surprised I found out that there already is such a feature.

Apparently, the feature only worked when the transaction was received while Electrum was offline, since `ElectrumWindow.notify_transactions` was only called once from `load_wallet`.

Calling the method from the `on_network` callback makes the notification appear instantly for new transactions.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -89,6 +89,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
 
     payment_request_ok_signal = pyqtSignal()
     payment_request_error_signal = pyqtSignal()
+    notify_transactions_signal = pyqtSignal()
     new_fx_quotes_signal = pyqtSignal()
     new_fx_history_signal = pyqtSignal()
     network_signal = pyqtSignal(str, object)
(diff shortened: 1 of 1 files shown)


gui/qt/main_window.py
```diff
@@ -565,11 +568,11 @@ def notify_transactions(self):
                       self.tx_notifications.remove(tx)
                       is_relevant, is_mine, v, fee = self.wallet.get_wallet_delta(tx)
                       if(v > 0):
-                          self.notify(_("New transaction received. %(amount)s") % { 'amount' : self.format_amount_and_units(v)})
+                          self.notify(_("New transaction received: %(amount)s") % { 'amount' : self.format_amount_and_units(v)})
 
     def notify(self, message):
         if self.tray:
-            self.tray.showMessage("Electrum", message, QSystemTrayIcon.Information, 20000)
+            self.tray.showMessage("Electrum", message, QIcon(":icons/electrum_dark_icon"), 20000)
```
SomberNight: Have you tested this? Because `QSystemTrayIcon.Information` looks to be an integer.
Anyway, it crashes for me on Debian.

```
TypeError: showMessage(self, str, str, icon: QSystemTrayIcon.MessageIcon = QSystemTrayIcon.Information, msecs: int = 10000): argument 3 has unexpected type 'QIcon'
Aborted
```
[Apparently it's non-trivial to set a custom icon here](https://stackoverflow.com/questions/31045413/qsystemtrayicon-notification-message-with-custom-icon)

SomberNight: Oh I see the problem. This requires Qt 5.9.
https://doc.qt.io/qt-5/qsystemtrayicon.html#showMessage-1

Issue #3379, opened 2017-11-26
[QT] pay-to-many option should be disabled if payto_e is frozen


SomberNight:
The `Tools`>`Pay to many` option should be greyed out and disabled when the `Pay to` field is frozen (read-only) in the `Send` tab.

### Example
Paying for a payment request
![send_amount_frozen_paytomany1](https://user-images.githubusercontent.com/29142493/33237310-0cf754f8-d270-11e7-9970-4ac6e3eac57a.png)
After clicking `Tools`>`Pay to many`
![send_amount_frozen_paytomany2](https://user-images.githubusercontent.com/29142493/33237311-0eba6ce4-d270-11e7-8409-93c3a01bb664.png)

ecdsa:
I have been thinking that the payto field should not be multiline by default, because it is a rare use case.
I think it should be a QLineEdit, and replaced by a multiline only if the user hits the paytomany menu.

Issue #2179, opened 2017-02-19
when sweeping, no way to set a fee


ivanhoe1982:
it gets a default 50 satoshi per byte, which in current circumstances results in 8-12 hour confirmation times
(ubuntu)

gits7r:
Go to Preferences and use dynamic fees and enable option edit fees manually.

ivanhoe1982:
I did just that even before posting the bug. I am testing on 2.7.18. After sweep, I get a transaction window with the usual transaction id, ins and outs, copy, save, sign (greyed out since it is already signed with the pk) and broadcast. No obvious (or non-obvious) way to edit the fee. The fee message actually says "unknown". Am I missing something?

LawnboyMax:
When you press `sweep`, a sweep function from wallet.py is called with fee field set as `None` no matter what, which creates a transaction with an estimated fee (by using estimate_fee function from wallet.py) which is determined by the dynamic fee setting. 

This means that if the fee is to be changed when sweeping, it has to be done somehow before you press `sweep` and create a transaction, similar to how you can change the fee when sending. 

I've also had problems with default default dynamic fees, when my transactions got stuck as unconfirmed for more than a day.

I think something has to be changed about how the dynamic fees are calculated or, even better, a fee slider added to the sweep window (where you are prompted to enter private keys).

ivanhoe1982:
Yup, I agree, I did exactly the same trace yesterday... Sweep window is a very special 'transaction' window, and I feel like the sweep function should just hand off to the transaction window, instead of being this weird standalone case. Then one could use functionality already built into transaction.

JazzTp:
Newbie question, I could most likely find an answer from docs:

How does sweep differ from just creating a transaction and choosing Max
amount?

JazzTp:
Ooops, maybe the answer is that you only use sweep to extract all funds
from a wallet which private key you imported into an Electrum wallet, not
to send all funds out of an Electrum wallet.

schildbach:
I can attest that when sweeping you cannot set the fee. Also you cannot RBF the transaction.

ecdsa:
as a workaround, you can create a temporary wallet with the privkeys that you want to sweep.
then, all the advanced fee and rbf options are available.

donnlee:
Yes, please allow editing of the fee when sweeping a private key.
Thanks to google websearch for finding this github issue and confirming that I wasn't missing a checkbox setting somewhere.

[...]

PR #3380, opened 2017-11-26
use the send tab when sweeping (qt gui)

SomberNight:
This might not be perfect as is, I haven't done very extensive testing yet (just a bit). However I wanted to get feedback regarding the approach/idea in general.

The primary motivation is #2179

Notes:
1. `Amount` is frozen and set to max. This is to avoid dealing with the user trying to sweep only "some of the coins" (where to put the change in that case?)
2. `Pay to` field is frozen and set to the single address the user entered in the sweep dialog. This is done to block the user from using the advanced functionalities of the `Pay to` field, as then she would be able to set a smaller amount directly or via pay-to-many (see previous point)
3. If the user selects `Pay to many` from the menu, the `Pay to` field will become empty and read-only. To recover she needs to use `Clear`. The `Pay to many` functionality should probably be disabled (#3379).
4. The `Send` button is hidden for watch-only wallets at the moment, but the `Preview` button is there, so the user can work around this and do the sweep. https://github.com/spesmilo/electrum/blob/9425319dcd4c16a63ee3559bde433472b44ec165/gui/qt/main_window.py#L1772
5. For passworded wallets, the user is prompted for a password to sign, but this is of course pointless.
6. Using `Spend` in the `Coins` tab, or `Spend from` in the `Addresses` tab, when the `Send` tab is filled with the sweep details, does not play nicely at the moment... the user needs to use `Clear` to recover.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -118,6 +118,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         self.require_fee_update = False
         self.tx_notifications = []
         self.tl_windows = []
+        self.tx_external_keypairs = {}
 
         self.create_status_bar()
         self.need_update = threading.Event()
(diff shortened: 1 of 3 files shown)


ecdsa:
note: this needs to be reworked because I moved the sweep method out of the wallet class.
it would be nice to display the From field with the set of utxos

SomberNight:
I've rebased this on current master.

> it would be nice to display the From field with the set of utxos

Isn't that already the case? It should be set by `self.spend_coins(coins)`. Or maybe I misunderstand what you mean. This is how a "sweep" looks like:

electrum_sweep_in_send_tab

ecdsa:
oh ok, thanks.

PR #3397, opened 2017-11-27
Add script for deterministic builds

bauerj:
`build.sh` should be used to run the whole pipeline now. It includes a workaround for .pyc timestamps in the pyinstaller bundle which might be fixed by pyinstaller/pyinstaller#3007.

This also sets the versions of the hardware wallet libraries to a fixed version. Maybe we want to move that to setup.py or requirements.txt in the future in order to use it with Mac builds.


contrib/build-wine/build.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+# Lucky number
+export PYTHONHASHSEED=22
+
+here=$(dirname "$0")
+
+echo "Clearing $here/build and $here/dist..."
+rm $here/build/* -rf
```
lzsaver:
```
# here=$(dirname "$0")
# ...
# rm $here/build/* -rf
# rm $here/dist/* -rf
```

This script looks dangerous. If user does not have coreutils or if there is another dirname executable in the local bin directory - this script can remove /build/* and /dist/* in the root of user's filesystem.

```
$ here=$(nothing "$0")
$ echo rm $here/build/* -rf
$ echo rm $here/dist/* -rf
```

We have to check that _${here}_ is not empty at least. Also do we really need it or can we just use _${PWD}_?

Issue #3072, opened 2017-10-20
"Failed to compile output" after click confirm button on Trezor wallet


lkthomas:
Electrum 2.9.3 --testnet
Ubuntu 14.04 64bit

[profiler] estimated_size 0.0002
[CoinChooserOldestFirst] change: [(0, u'2MuD9bCeGay5nx1cUKhzgErZAPApYnRqP1z', 129899600)]
[CoinChooserOldestFirst] using 1 inputs
[CoinChooserOldestFirst] using buckets: [u'60f1f864f4a157f71388b98e73c98d400620245d5d0b2446b7307e5eb3eaea29:0']
[profiler] estimated_size 0.0006
[profiler] estimated_size 0.0008
[14.3.140.101] failed to connect timed out
[DeviceMgr] getting client for keystore
[DeviceMgr] scanning devices...
[DeviceMgr] end client for keystore
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/electrum_gui/qt/util.py", line 582, in run
    result = task.task()
  File "/usr/local/lib/python2.7/dist-packages/electrum/wallet.py", line 1141, in sign_transaction
    k.sign_transaction(tx, password)
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/plugin.py", line 66, in sign_transaction
    self.plugin.sign_transaction(self, tx, prev_tx, xpub_path)
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/plugin.py", line 242, in sign_transaction
    signed_tx = client.sign_tx(self.get_coin_name(), inputs, outputs, lock_time=tx.locktime)[1]
  File "/usr/local/lib/python2.7/dist-packages/electrum_plugins/trezor/clientbase.py", line 221, in wrapped
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 158, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 824, in sign_tx
    res = self.call(proto.TxAck(tx=msg))
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 158, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/trezorlib/client.py", line 201, in call
    msg = handler(resp)
[...]

PR #3407, opened 2017-11-29
fix: keepkey sending to p2sh output

SomberNight:
fixes (parts of ???) https://github.com/spesmilo/electrum/issues/3072
see https://github.com/spesmilo/electrum/issues/3072#issuecomment-347751703

This PR reverts https://github.com/spesmilo/electrum/commit/020f0637d7d8ec412df3762cfa9485a36574d274


ecdsa:
Note: I might have merged this too fast. I believe that 020f063 is needed (or will be needed in the future) for trezor to be able to send to bech32 addresses. @slush0 can you confirm?

Issue #3363, opened 2017-11-23
sweep paper wallet fails


wessjan:
I'm still in Testnet. 
When I try to sweep the private key of a paper wallet, electrum (3.0.2) finds the funds, creates a transaction and signs it. But every time when I hit Broadcast, the error "**min relay fee not met**" occurs. I've not seen a way to set the fee at all.  
![electrum001](https://user-images.githubusercontent.com/15079538/33190777-e8696d80-d0b0-11e7-8245-336e758f933c.PNG)

SomberNight:
This happens when we originally underestimate the transaction size, and the intended sat/byte fee is really low. On testnet, the dynamic fee estimate for a 5 block confirmation target is often ~1 sat/vbyte, which is also the minimum default relay fee used by servers.

Also, unfortunately, there is currently no straightforward way to set a fee when sweeping. ( https://github.com/spesmilo/electrum/issues/2179 )

A trivial example for Electrum underestimating the transaction size is when a WIF with an uncompressed pubkey is being swept.
https://github.com/spesmilo/electrum/blob/9fc946492bb329f274431c9af6105bccc1560fea/lib/transaction.py#L620-L623

SomberNight:
The user can now set a fee when sweeping: https://github.com/spesmilo/electrum/pull/3380. This applies from version 3.0.3 going forward.
I think this can now be closed.

PR #3421, opened 2017-11-30
Fee option in Sweep window and fee showing up in transaction window instead of 'unknown'

sunnyville01:
This pull request offers a solution to [2179](https://github.com/spesmilo/electrum/issues/2179) and also, as a consequence, to [3363](https://github.com/spesmilo/electrum/issues/3363) as well. 

The first problem was that the fee for the sweep transaction was getting calculated automatically using the dynamic fee estimations, without giving an option to the user to choose a suitable fee themselves, like through the two fees settings in `Preferences`, for example. This often resulted in very low fee when the transaction size was under estimated. 

The second problem was that after clicking the `Sweep` button, the transaction window that appeared had fee shown as `unknown`.

The first problem is solved in the [first commit](https://github.com/spesmilo/electrum/commit/03e07efe2bcc9eefa07f46cebe36c54fca2d8c0e) and second one in the [second commit](https://github.com/spesmilo/electrum/commit/10582d7d3e341e7f570deda0adcaa3917671f635). This was achieved by:

- Adding a `FeeSlider` to Sweep dialog box (`sweep_key_dialog()`), and additionally showing a `BTCAmountEdit` box, if `Edit fees manually` was chosen.
- Passing the right value of fee to the `sweep` function in `wallet.py`.
- Adding a `tx_type` attribute to certain functions and using that to display the correct fee in the transaction window.

I've tested this reasonably well and tried to catch all the exceptions where the the fee is handled. I've tried to mimic how it is done in the `Send` tab, but have done it very cautiously. 

Here are some screenshots:

sweep-1

The slider and the box work as expected with the 'Use dynamic fees' and 'Enter fees manually' settings.

sweep-2

The transaction window shows the correct fee and not 'unknown' as was the case before.

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -125,11 +125,12 @@ def sweep(privkeys, network, config, recipient, fee=None, imax=100):
         raise BaseException(_('No inputs found. (Note that inputs need to be confirmed)'))
     total = sum(i.get('value') for i in inputs)
     if fee is None:
+        if not network.config.has_fee_estimates():
+            raise BaseException('Dynamic fee estimates not available')
         outputs = [(TYPE_ADDRESS, recipient, total)]
         tx = Transaction.from_io(inputs, outputs)
         fee = config.estimate_fee(tx.estimated_size())
-        if config.fee_per_kb() is None:
-            raise BaseException('Dynamic fee estimates not available')
+
     if total - fee < 0:
         raise BaseException(_('Not enough funds on address.') + '\nTotal: %d satoshis\nFee: %d'%(total, fee))
     if total - fee < dust_threshold(network):
@@ -441,14 +442,17 @@ class Abstract_Wallet(PrintError):
             delta += v
         return delta
 
-    def get_wallet_delta(self, tx):
+    def get_wallet_delta(self, tx, tx_type=None):
         """ effect of tx on wallet """
         addresses = self.get_addresses()
         is_relevant = False
         is_mine = False
         is_pruned = False
         is_partial = False
+        is_sweep = False
         v_in = v_out = v_out_mine = 0
+        if tx_type == 'sweep':
+            is_sweep = True
         for item in tx.inputs():
             addr = item.get('address')
             if addr in addresses:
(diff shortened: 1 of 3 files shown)


ecdsa:
thanks, but I think #3380 is a better solution

lib/wallet.py
```diff
@@ -125,11 +125,12 @@ def find_utxos_for_privkey(txin_type, privkey, compressed):
         raise BaseException(_('No inputs found. (Note that inputs need to be confirmed)'))
     total = sum(i.get('value') for i in inputs)
     if fee is None:
+        if not network.config.has_fee_estimates():
+            raise BaseException('Dynamic fee estimates not available')
```
SomberNight: Not sure why this change. You forgot to handle the case where the user has dynamic fees disabled and there are no fee estimates available. He should be able to sweep none-the-less (as he is using static fees anyway). The line with `config.fee_per_kb()` below, that you have deleted, handled that.

SomberNight:
The second commit "Transaction window shows the correct fee instead of unknown" could still make sense to merge; or at least the problem it's trying to solve should be at least considered.

When the user is sweeping, the tx preview window shows an unknown fee, and that commit would fix that. It exploits that sweep uses `u = network.synchronous_get(('blockchain.scripthash.listunspent', [sh]))` which sets `'value'` for the inputs.

lib/wallet.py
sunnyville01: I was actually not able to use `config.fee_per_kb()` the way it was used before, probably because the `sweep` function was taken out of the `Abstract_Wallet` class in a recent commit. So I just used `network.config.has_fee_estimates()` while over looking the case that you correctly stated.

sunnyville01:
@SomberNight Yes, I do think that the issue of showing 'unknown' in the transaction window should be fixed. In my [second commit](https://github.com/spesmilo/electrum/pull/3421/commits/10582d7d3e341e7f570deda0adcaa3917671f635). The problem is that the function `get_wallet_delta` in `wallet.py` sets `is_mine` variable to `False`, if it finds that none of the addresses used to get the inputs in the transaction is owned by the user's wallet. In this case It sets the `fee` to None, overlooking the case of `sweeping`. And this is too deep in the function chain, as this value of the fee is then used by `get_tx_info`, which is then used in the `TxDialog` class called by `show_transaction` function inside `transaction_dialog.py`. I was able to fix that by passing the `tx_type` argument in these sets of functions.

Also, could you elaborate more on what you meant when you mentioned `u = network.synchronous_get(('blockchain.scripthash.listunspent', [sh]))` in your last comment? Are you proposing an alternative solution? I don't quite understand what that bit of code does.

SomberNight:
> Also, could you elaborate more on what you meant when you mentioned u = network.synchronous_get(('blockchain.scripthash.listunspent', [sh])) in your last comment? Are you proposing an alternative solution? I don't quite understand what that bit of code does.

No, sorry for the misunderstanding, I wasn't proposing an alternative solution. I was just remarking that that's where `'value'` gets set for the input UTXOs. That code is a networking call that queries an electrum server to find UTXOs (coins) associated with the addresses. I just made that remark because I myself had not immediately seen where `'value'` gets set.

Issue #3404, opened 2017-11-28
App crashes if it can't find wallet


reubano:
Was pulling my hair trying to figure out why the app kept crashing. Turns out, I forgot to open the sparse disk file that contains my wallet. It'd be useful if the app gave a nice error message that it can't find the wallet.

SomberNight:
- Are you using the QT GUI? Or how did you start Electrum?
- Did Electrum crash on startup upon trying to open the most recently used wallet (but the path for that was unavailable)?
- What OS is this on?
- Is there an exception trace if you start the app with the `-v` flag?

Can you please give more details in general?

SomberNight:
I've found a crash that happens if the QT install wizard encounters an IOError for the path you enter and you still click `Next`.

```
[WalletStorage] wallet path /home/user/.electrum/wallets/wallet_112
Traceback (most recent call last):
  File "/home/user/wspace/electrum/gui/qt/main_window.py", line 443, in new_wallet
    self.gui_object.start_new_window(full_path, None)
  File "/home/user/wspace/electrum/gui/qt/__init__.py", line 199, in start_new_window
    wallet = wizard.run_and_get_wallet()
  File "/home/user/wspace/electrum/gui/qt/installwizard.py", line 225, in run_and_get_wallet
    if not self.storage.file_exists():
AttributeError: 'NoneType' object has no attribute 'file_exists'
Aborted
```

reubano:
> Are you using the QT GUI? Or how did you start Electrum?

Downloaded img... double clicked icon

> Did Electrum crash on startup upon trying to open the most recently used wallet (but the path for that was unavailable)?

Correct

> What OS is this on?

macOS 10.13

> Is there an exception trace if you start the app with the -v flag?

N/A. app opened via icon, not cli

SomberNight:
I can't reproduce this on Debian, nor on Windows 10. Might be specific to Mac.

reubano:
@SomberNight try renaming the wallet folder. My tests show that renaming the file works fine, but if the folder is missing, it crashes.

SomberNight:
I think this is not applicable anymore. The startup has been made much more robust to errors over recent years.

PR #3430, opened 2017-12-02
Show an error if Electrum cannot open the wallet file

sunnyville01:
This pull request is motivated by [3404](https://github.com/spesmilo/electrum/issues/3404), where the app crashes if for any reason Electrum cannot open the wallet file chosen from the Open File dialog. With this pull it will now show a helpful message instead. 

This was achieved by having `self.show_error('Error message')` instead of a simple `return` inside `if not filename:` condition block. I have currently only implemented this for the `open_wallet()` function, but I think we can use this in other places as well, like `import_invoices()` and `import_contacts()`, for example.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -389,7 +389,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         wallet_folder = self.get_wallet_folder()
         filename, __ = QFileDialog.getOpenFileName(self, "Select your wallet file", wallet_folder)
         if not filename:
-            return
+            self.show_error('Error opening the file. Please check if the file is accessible.')
         self.gui_object.new_window(filename)
 
 


SomberNight:
> the app crashes if for any reason Electrum cannot open the wallet file chosen from the Open File dialog

Can you give more details how to reproduce this? Why does Electrum crash? Do you have a stack trace? Does the crash only affect the QT GUI (or does it also need to be handled for e.g. CLI usage)?

gui/qt/main_window.py
```diff
@@ -389,7 +389,7 @@ def open_wallet(self):
         wallet_folder = self.get_wallet_folder()
         filename, __ = QFileDialog.getOpenFileName(self, "Select your wallet file", wallet_folder)
         if not filename:
-            return
+            self.show_error('Error opening the file. Please check if the file is accessible.')
```
SomberNight: Are you sure that the `return` can be safely removed? Obviously, removing it means that the line below will execute. I can't say for certain if it's ok, because I am not even sure why the "crash" even occurs...

ecdsa: IIRC getOpenFileName returns None if the user cancels the dialog.
in that case it is unsafe to remove

PR #3441, opened 2017-12-04
add ripemd test

cluelessperson:


cluelessperson:
class name "Test_ripemd" is ugly.
So is "TestRipemd" :?

bauerj:
Please remove these lines from the ripemd module then. No need to duplicate it.

ecdsa:
I dont see the point of addding that test

bauerj:
I do: It would be nice to know when future Python changes break this code. Arguable the code isn't changed that often but it's still good to know that it works.

What's the harm in merging this (once the code duplication is removed)?

ecdsa:
this code is just not part of Electrum.

it is there only to provide ripemd hashing on Android because python-crytax does not have it.
it is not supposed to stay once python for android have fixed their hashlib module.

the harm merging that is that I get too many requests and that it takes too much of my time to evaluate them. adding requests of that kind is not productive at all. 

adding tests that actually test the software is difficult. it requires as much skills that writing the code itself.

PR #3455, opened 2017-12-06
Send change path to Ledger so user isn't prompted for change output

mzhou:
This way the user can trust based on the Ledger confirmation that change is going to their own address without trusting Electrum on their computer. Currently the change address and amount is prompted as an output.
Only implemented for segwit, since input amounts are signed, guaranteeing that the Ledger prompt represents the total non-change outputs, including fees.


SomberNight:
Tested ACK.

Tested with a p2wpkh-p2sh wallet,
- sending to a single output of different script types
- with complex transactions sending to ~5 outputs of mixed script types

plugins/ledger/ledger.py
```diff
@@ -371,7 +371,11 @@ def sign_transaction(self, tx, password):
             if segwitTransaction:
                 self.get_client().startUntrustedTransaction(True, inputIndex,
                                                             chipInputs, redeemScripts[inputIndex])
-                outputData = self.get_client().finalizeInputFull(txOutput)
+                if changePath:
+                    outputData = self.get_client().finalizeInput(output, format_satoshis_plain(outputAmount),
```
SomberNight: `output` is `str` here but `btchip-python` expects `bytes`. You can import `to_bytes` from `electrum.util` to convert it. Not sure if it's better to convert here or where its value gets set initially.

SomberNight:
I've now had another look at `finalizeInput` in [btchip-python](https://github.com/LedgerHQ/btchip-python/blob/c2bb2166cdaec95795517811c6f6e16038401f1d/btchip/btchip.py#L266), and while this PR works, I have doubts if that method should be used like this... It seems that that method *might* fall back to creating a tx with two outputs (one using `outputAddress`, the other `changePath`) -- though this does not seem to happen for some reason.

PR #3485, opened 2017-12-09
QR Library modified to work with Mac OS X. 

emanuelelaface:
I know that ThomasV does not agree to have OpenCV in Electrum (and I can understand that is an overkill just to have the QRCode working on Mac) but I haven't find another quick method to implement it on Mac, so I do anyway the pull request just for who is interested.
The Dependencies are: OpenCV, zbarlib, PIL.

### DIFF
--- a/lib/qrscanner.py
+++ b/lib/qrscanner.py
@@ -23,61 +23,92 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import os
 import sys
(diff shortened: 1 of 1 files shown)


ecdsa:
I see that you are using zbarlight instead of libzbar. 
is that related to osx?

emanuelelaface:
yes, I changed it because libzbar was crashing continuously on my Mac. Maybe is only my local setup, I haven't spent much time on it.

ecdsa:
deprecated by #4870

PR #3496, opened 2017-12-11
CoinChooser: privacy prefers confirmed and is default

SomberNight:
- "privacy" is now the default coinchooser policy, and "priority" has been removed
- "privacy" prefers to spend from confirmed coins
- coinchooser choice has been hidden in QT, the same should be done in kivy

-----

One remark regarding the changed privacy policy, which actually needs detailing the changes...

Any bucket can be:
    1. "confirmed" if it only contains confirmed coins; else
    2. "unconfirmed" if it does not contain coins with unconfirmed parents; else
    3. "unconfirmed parent"

In the current state of the PR, we try to only use buckets of type 1, and if the coins there are not enough, try to use the next type but while also selecting **all buckets of all previous types** (A).
This could be trivially changed to instead simply expand the set of allowed buckets, but do not automatically choose all buckets of all previous types (B).

I am not sure which option is better.

-----

Note: I have been researching ideas for alternative coinchooser policies that could be added (later probably), and found https://github.com/bitcoin/bitcoin/issues/7664 and http://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf
A potential "fee saver" policy could try to select inputs that are "an exact match" for the target send, and hence not need to create a change output.

### DIFF
--- a/lib/coinchooser.py
+++ b/lib/coinchooser.py
@@ -68,7 +68,7 @@ class PRNG:
             x[i], x[j] = x[j], x[i]
 
 
-Bucket = namedtuple('Bucket', ['desc', 'size', 'value', 'coins'])
+Bucket = namedtuple('Bucket', ['desc', 'size', 'value', 'coins', 'min_height'])
 
 def strip_unneeded(bkts, sufficient_funds):
     '''Remove buckets that are unnecessary in achieving the spend amount'''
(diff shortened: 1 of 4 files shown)


lib/coinchooser.py
```diff
@@ -267,23 +247,58 @@ def bucket_candidates(self, buckets, sufficient_funds):
...
+
+        This method tries to only use buckets of type 1, and if the coins there
+        are not enough, tries to use the next type but while also selecting
+        all buckets of all previous types.
+        """
+        conf_buckets = [bkt for bkt in buckets if bkt.min_height > 0]
+        unconf_buckets = [bkt for bkt in buckets if bkt.min_height == 0]
+        unconf_par_buckets = [bkt for bkt in buckets if bkt.min_height == -1]
+
+        bucket_sets = [conf_buckets] + [unconf_buckets] + [unconf_par_buckets]
```
kyuupichan: Isn't this just a simple array initialization of 3 entries?

SomberNight: It is; I'll change it to that explicitly to make it clearer.

Issue #3445, opened 2017-12-04
Feature Request: Testnet support for 2FA wallets


SomberNight:
Currently, 2FA wallets can't be used on testnet.
There seems to be no support for it on the server-side by TrustedCoin, and some small bits are likely missing on the client side from Electrum.

I can code the client side, but I would need TrustedCoin to make the server side changes.

In the docs for the normal `/1/` API provided by TrustedCoin, there seem to be support for testnet.
[The docs say](https://api.trustedcoin.com/#/docs) that `POST /1/cosigner` supports an optional `'network'` property.
However Electrum uses a custom API on `/2/`.

I tried including the `'network'` property in the HTTP request the client sends but it is ignored by the server.

Patch:
```patch
diff --git a/plugins/trustedcoin/trustedcoin.py b/plugins/trustedcoin/trustedcoin.py
index 7cd2976b..bb8efed6 100644
--- a/plugins/trustedcoin/trustedcoin.py
+++ b/plugins/trustedcoin/trustedcoin.py
@@ -74,7 +74,7 @@ class TrustedCoinException(Exception):
 class TrustedCoinCosignerClient(object):
     def __init__(self, user_agent=None, base_url='https://api.trustedcoin.com/2/'):
         self.base_url = base_url
-        self.debug = False
+        self.debug = True
         self.user_agent = user_agent
 
     def send_request(self, method, relative_url, data=None):
@@ -125,6 +125,7 @@ class TrustedCoinCosignerClient(object):
             'xpubkey1': xpubkey1,
             'xpubkey2': xpubkey2,
             'billing_plan': billing_plan,
+            'network': 'testnet',
         }
         return self.send_request('post', 'cosigner', payload)
```

Log:
```
post https://api.trustedcoin.com/2/cosigner
{
  'xpubkey2': 'xpub69EkcMgnjkJnqwAWbRSeFxR3ZpA5C7ohw42MSQRhirvTTkvshpyKqQ7EfN2rLSPKZYpHc3h22HAyUt6rUUypfWQUeL5QzrQQyAEPNaRNejs',
  'network': 'testnet',
  'email': 'testing@asdads.com',
  'billing_plan': 'electrum-per-tx-otp',
[...]

PR #3514, opened 2017-12-13
make 2fa wallets work on testnet

SomberNight:
Closes https://github.com/spesmilo/electrum/issues/3445

The tpub <-> xpub conversions are a bit messy... :/


ecdsa:
note: this needs to be updated, tpub conversions are not needed anymore.

PR #3538, opened 2017-12-17
add ability to show a receiving address on ledger screen

mzhou:


SomberNight:
Tested, and works :)

However, there is equivalent functionality for Trezor (and KeepKey and Digital Bitbox), implemented differently. I think it would be better if this PR was made to be similar to that.

https://github.com/spesmilo/electrum/blob/3851e78454f559ec4368a4b04dd61c2b9a1f9dc1/gui/qt/address_list.py#L163
https://github.com/spesmilo/electrum/blob/3851e78454f559ec4368a4b04dd61c2b9a1f9dc1/plugins/trezor/qt_generic.py#L189-L197

plugins/ledger/ledger.py
```diff
@@ -435,6 +435,20 @@ def sign_transaction(self, tx, password):
         tx.raw = tx.serialize()
         self.signing = False
 
+    def show_address(self, sequence, txin_type):
+        self.signing = True
+        # prompt for the PIN before displaying the dialog if necessary
+        client = self.get_client()
```
marceloneil: Is this variable necessary?

PR #3545, opened 2017-12-17
requirements.txt support for user only pip requirements

lillypad:
Some linux distros require pip to use the `--user` switch. The use case for installing dependencies in this case would be `pip3 install --user -r requirements.txt` then `sudo python3 setup.py install`.

Small change but useful for others.

Cheers!🥂

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,9 @@
+pyaes>=0.1a1
+ecdsa>=0.9
+pbkdf2
+requests
+qrcode
+protobuf
+dnspython
+jsonrpclib-pelix
+PySocks>=1.6.6
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,9 @@ import platform
 import imp
 import argparse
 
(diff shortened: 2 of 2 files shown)


SomberNight:
There's already `contrib/requirements.txt` -- actually I'm not sure why the duplication between setup.py and that at the moment. Could we not just use that?

bauerj:
Thanks!

@SomberNight We have a script to update `contrib/requirements.txt`: `contrib/freeze_packages.sh`. The requirements file there is only intended to be used when creating binaries because the versions there are fixed.

Since it's bootstrapped using setup.py, it doesn't make much sense to use the output file there.

lillypad:
That's good I just thought it may be a good suggestion. I use Gentoo so my use of pip and setup tools varies from Debian based distros slightly.

ecdsa:
closing, replaced by #3604

SomberNight:
@ecdsa I think this is a somewhat independent change from #3604.

Issue #3239, opened 2017-11-08
On Tails 3.2, Electrum 3.0.1 segmentation fault opening QRcode scanner, OK on Linux Ubuntu 16.04.


JazzTp:
On Tails 3.2, Electrum 3.0.1 segmentation fault opening QRcode scanner, while it opens fine on Linux Ubuntu 16.04.

Related issue: #2527

ysangkok:
A stack trace would be useful. If the Python interpreter is crashing, start GDB using `gdb $(which python3)` and write `run <path of the electrum script>`. Electrum should start, you can reproduce the crash and after the crash you can write `bt` to get a stack trace.

JazzTp:
```bash: gdb: command not found```

((( I use to boot from DVD, no idea how to install stuff on Tails. Actually I chose to switch to Ubuntu Linux for the watching only part to avoid the webcam problem. I have the cold storage running on an Older Tails 2.12 where there's no webcam problem, and BTW I haven't tested so far if Electrum 3.0.1 based on python 3 can run on it [using the wallet very seldom admittedly]. )))

**Comment by yairEO** [comment]
On Puppy Linux I get this also

**Comment by JazzTp** [comment]
Same thing with Electrum 3.0.2 on Tails 3.3 (released 2017-11-14).

**Comment by jeffersoncarpenter** [comment]
Is this the same issue as #3224?

**Comment by JazzTp** [comment]
If "clicking QR code" means opening the QR code scanner, apparently it
would be the same issue.

I got there by "tools | load transaction | from QR code".

**Comment by potato20** [comment]
I'm getting this too. I'm using pyenv with python 3.6.3. Here's the backtrace in gdb

```
#0  0x00007fffbc4620dc in XUnmapWindow () from /usr/lib/x86_64-linux-gnu/libX11.so.6
#1  0x00007fffeabdebc4 in ?? () from /usr/lib/x86_64-linux-gnu/libzbar.so.0
#2  0x00007fffeabbcac6 in zbar_processor_set_visible () from /usr/lib/x86_64-linux-gnu/libzbar.so.0
#3  0x00007ffff4234e40 in ffi_call_unix64 () from /usr/lib/x86_64-linux-gnu/libffi.so.6
#4  0x00007ffff42348ab in ffi_call () from /usr/lib/x86_64-linux-gnu/libffi.so.6
[...]

PR #3546, opened 2017-12-17
qrscanner: Fix segfault when using a specific camera device

cb8:
The path to the camera device is not encoded before it gets passed to the library. This results in `zbar_processor_init` returning an error code. However, as this code is not checked, a SEGFAULT is thrown after calling further methods on `proc`.

Fixes: #3239


SomberNight:
Note that if you close the popup window (of the camera), then it crashes with:
```
  File "/home/user/wspace/electrum/gui/qt/paytoedit.py", line 263, in qr_input
    if data.startswith("bitcoin:"):
AttributeError: 'NoneType' object has no attribute 'startswith'
Aborted
```
But this is unrelated for the PR. I'll fix this in a moment.

PR #3557, opened 2017-12-19
add compilation guide on windows

zoowii:
I add a guide to build electrum on pure-Windows environment, without usage of Wine


bauerj:
Hey,

thanks for this guide. We don't use Windows to build the binaries so we can't say whether this is correct or not. Furthermore, we will not be able to update this guide with future changes as that would require too much time for testing.

However, since this seems to be useful for anyone wanting to build it on Windows, you can consider to add this to the community wiki here on Github if you add a disclaimer that this is not officially supported.

PR #3575, opened 2017-12-21
fix: address filters for Imported_Wallet: wallet.is_used()

SomberNight:
I don't see the reason for this override, and it's causing bugs here:

https://github.com/spesmilo/electrum/blob/4bab8b63e192078be98ebbf0e84581b2056b5dc5/gui/qt/address_list.py#L88-L93


ecdsa:
IIRC the reason is that the receive tab requires an unused address.
but that is not clean, so better merge this

PR #3583, opened 2017-12-22
fee ui: if user sets fee_e, stick to that

SomberNight:
- if user sets fee_e, stick to that
- show tooltip re fee precision when applicable
    - alternatively, instead of setting the tooltip of fee_e, we could show/hide an info icon to the right of fee_e, and that info icon could have the tooltip


SomberNight:
Rebased.

ecdsa:
I do not think this is how we should handle the precision issue.
I think the satoshis added for rounding should be displayed explicitly in the fee details.
something like : 
fee_rate x size + rounding = fee

SomberNight:
Closing. Superseded by https://github.com/spesmilo/electrum/pull/3643

PR #3599, opened 2017-12-25
fix some crashes when the underlying QT (C/C++) object no longer exists

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/2968#issuecomment-337762741

Also a similar crash in the `History` tab with `Edit Description`.
And another with message decryption.
And another with message signing.


ecdsa:
I'm merging this because we don't have #3856 yet.
we should not forget to cleanup that code later, though

PR #3604, opened 2017-12-26
freeze hw wallet dependencies into separate file

SomberNight:
The versions in `prepare-hw.sh` were getting a bit outdated, and this restructuring might be better for deterministic builds in any case. For example, without using pip freeze for these dependencies, I think if one of the hidden dependencies (e.g. needed by trezor) got an update, builds might not be reproducible.

It is a separate file from `requirements.txt` as that is used by `make_packages`.

@bauerj


bauerj:
This is a good idea but I would propose a different approach:

1. Move hardware-wallet dependencies to a requirements file similar to #3545.
2. The requirements.txt with pinned versions should be moved to e.g. `/contrib/deterministic-build/` so people are not confused why there are two files.
3. Add an [extra](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) "hardware" to setup.py that uses the dependencies from `/dependencies-hw.txt`.
4. Document how to install Electrum with hardware-wallet support in README.txt and on electrum.org

What do you think?

SomberNight:
Ok, sounds good. I'll have a go at that a bit later.

SomberNight:
Squashed and rebased on master.

Issue #3373, opened 2017-11-24
Provide link to documentation in the "Derivation" section of hardware wallet setup


fresheneesz:
When setting up a hardware wallet (eg Trezor) in a new electrum wallet (I'm on v3.0.2), you get to a section called "Derivation" that explains absolutely nothing about what it is, what it means, or where to find more information about it. From a usability standpoint, this is as bad as it gets. The only thing that could make this section worse is if it didn't pre-populate that field with anything. But since it *does* prepopulate that field, in the case you're trying to use a segwit address, the prepopulated value is **incorrect**.

Is there documentation for this value (ie how to construct the value, what its parts mean, etc)? I couldn't find any documentation of this. But if it does exist, that documentation should be linked to from this section. If it doesn't, then someone should write and publish that documentation. 

But, having the user type a bizarre string here can't be the best user interface for this, right?

SomberNight:
It's a complex topic.

[BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki), [BIP43](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki), [BIP49](https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki)
(all these implicitly use [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki))

I guess the default derivation could be changed to use BIP49 (segwit-p2sh)... that would be beneficial to **new users**, BUT existing users, especially casual ones, might freak out and **think they lost their coins**, and blame it on Electrum. This is why, as far as I know, this project considers BIP44 and related BIPs to be bad design, but hardware wallets and some other software still use it... hence Electrum is also forced to support it.

When you use Electrum seeds, all this is completely abstracted away from the user, as this information is encoded in the seed words themselves. This is why you need to generate new seed words if you want to use segwit -- because it is encoded in the older type seed words not to use it.

fresheneesz:
@SomberNight Thanks for the info! Could we add this info somewhere in the electrum docs, and then add a link to that documentation page in the Derivation section when setting up? We could then add to it from there.

I would suggest adding a dropdown here where the user can select from "standard (BIP44)", "segwit (BIP49)", and "custom" where if you select custom, it gives you a blank input to type in your derivation path.

ecdsa:
I guess we could have a buttton that sets default values for 44 and 49, while still allowing other derivations.
[...]

PR #3610, opened 2017-12-27
wizard: add derivation presets for bip44/49

SomberNight:
Closes https://github.com/spesmilo/electrum/issues/3373

bip44_der_buttons


fresheneesz:
I would suggest we add the actual derivation path that it will be changed to next to those buttons. Either that or labeling the current derivation path with one of those textual descriptions (if it matches). One of those things would allow users to much more easily tell if the derivation path they have is already one of those. I'd actually perfer that second method.

lichtamberg:
Please also add a description if ledger or trezor is using one of the mentioned methods

PR #3611, opened 2017-12-27
avoid shorter seeds 'by luck'

SomberNight:
I am not sure if this is deliberate or not, but currently Electrum might generate shorter seeds in rare cases (fewer words).

I only noticed it when the GUI gave me `sniff slender rifle rack express anxiety ecology believe heavy accuse blue`, which is only 11 words.
Suddenly we can longer tell people that "there's no way your seed contains only 9 words", rather "it's extremely unlikely..." :)

### DIFF
--- a/lib/mnemonic.py
+++ b/lib/mnemonic.py
@@ -171,7 +171,10 @@ class Mnemonic(object):
         n_custom = int(math.ceil(math.log(custom_entropy, 2)))
         n = max(16, num_bits - n_custom)
         print_error("make_seed", prefix, "adding %d bits"%n)
-        my_entropy = ecdsa.util.randrange(pow(2, n))
+        my_entropy = 1
+        while my_entropy < pow(2, n - bpw):
+            # try again if seed would not contain enough words
+            my_entropy = ecdsa.util.randrange(pow(2, n))
         nonce = 0
         while True:
             nonce += 1


AbdussamadA:
I believe you are reducing the entropy of the seed by doing this. I once asked nullc whether omitting the first 2^32 and last 2^32 when generating private keys was a good idea because I thought those private keys were easily calculated. He said it was not a good idea and all I was doing was reducing the entropy of my private keys.

SomberNight:
> I believe you are reducing the entropy of the seed by doing this.

This change would eliminate 1/2048 of all possible seeds, that's true. Those seeds however use fewer words than usual seeds (e.g. 11 instead of 12). All in all, it probably doesn't matter much either way.

AbdussamadA:
By default all values for myentropy that are less than 2^121 are disallowed. Only values between 2^121 and 2^132 are allowed. This reduces the keyspace a lot.

SomberNight:
Note that e.g. the 2fa seed logic breaks down if this happens:
https://github.com/spesmilo/electrum/blob/563aae0dbd6f29c74e18bb0b2ca18ca340f40706/plugins/trustedcoin/trustedcoin.py#L399-L412

Fortunately, the wallet will not be allowed to be created, Electrum crashes after confirming the seed in the wizard.

SomberNight:
Actually, before version 2.7, the trustedcoin code did not raise an exception for shorter seeds:
https://github.com/spesmilo/electrum/blob/7dbb23e8c6acfa40795d861b192c205dbb4b4268/plugins/trustedcoin/trustedcoin.py#L356-L359
Which means from version 2.0 until and excluding 2.7 users could generate shorter 2fa seeds. A seed of any length in fact. :/

PR #3621, opened 2017-12-29
Make Electrum work with trezorlib 0.9.0

slush0:
In upcoming days we'll release trezorlib 0.9.0 which removes dependency to protobuf library by Google (and uses pure-python lightweight implementation). Thanks to this, some internal structures have changed.

Please review, but not merge yet; trezorlib 0.9.0 is not deployed to pip yet.


bauerj:
Are these changes backwards-compatible? If not, can you add a check for the trezorlib version?

slush0:
@bauerj Fair point. I didn't find an example how to add version requirement into __init__.py (all other plugins just mention dependency without any specific version. Is there any recommendation?

ecdsa:
sorry, I did not see that it should not be merged yet. hopefully trezorlib will be released soon

slush0:
Yes, it probably will, in following days. I wanted to make another cosmetic changes, but I'll open another PR.

slush0:
OT: I just removed myself from Spesmilo group. I don't need direct commit rights and with my git-fu it will be much safer this way :-).

bauerj:
@slush0 How long will it take for 0.9.0 to be released? As far as I can [see](https://pypi.org/project/trezor/) not even 0.8.0 is released yet.

Issue #3551, opened 2017-12-18
bech32 with Trezor


someone235:
Trezor [just implemented](https://blog.trezor.io/expanded-cryptos-cryptocurrency-support-firmware-1-6-0-3825b5853470) bech32 support for receiving address. Is there a way to use it with Electrum?

bauerj:
Are you sure this is for receiving addresses? I thought they only support sending to bech32 addresses yet.

SomberNight:
Not at the moment, no.

It is not clear for me whether only sending to native segwit outputs was implemented in the Trezor fw or receiving too.
See https://github.com/spesmilo/electrum/issues/2976#issuecomment-340871611

Even if the Trezor firmware (1.6.0) has full receiving support now,
1. the Trezor plugin in Electrum would need some changes to support this
2. Electrum is unlikely to expose this functionality to users until there is a bip43 purpose allocated (or similar, at least a reasonable proposal) for the respective scripts. Something like bip49 would be needed (which specifies a bip32 derivation path along which p2wpkh-p2sh scripts should be created)

sneurlax:
So... closed?

bauerj:
Yes, this has been merged.

PR #3630, opened 2017-12-31
bip84 for sw seeds

SomberNight:
This enables software bip39 seeds to use p2wpkh scripts to receive on along the derivation path described in soon-to-be bip84: https://github.com/bitcoin/bips/pull/620

~~Best not to merge until bip84 gets into draft status (above PR gets merged).~~
EDIT: bip84 is now in draft status

-----

For the time being **I've disabled this** (p2wpkh output scripts for receiving) **for hardware wallets**.
- The Trezor plugin needs some changes for it to work, but the fw might be ready. https://github.com/spesmilo/electrum/issues/3551
    - also atm Trezor on master is "broken" due to merging https://github.com/spesmilo/electrum/pull/3621 too fast :)
- Ledger (Nano S) seems to work already with btchip-python 0.1.24, and latest fw
- KeepKey and dbb are not ready for this (though I haven't tested)
    - KeepKey needs changes in plugin, but even before that, I think python-keepkey needs some changes (and their fw too)
    - not sure about dbb but the respective plugin, again, would at least need changes

Hence, the most reasonable path forward would be to get Trezor to work ( https://github.com/spesmilo/electrum/issues/3551 ), and then test Trezor and Ledger a bit more, before enabling this for them. (This PR can get merged before that though.)

### DIFF
--- a/lib/base_wizard.py
+++ b/lib/base_wizard.py
@@ -231,14 +231,15 @@ class BaseWizard(object):
             self.derivation_dialog(f)
 
     def derivation_dialog(self, f):
-        default = bip44_derivation(0, False)
+        default = bip44_derivation(0, bip43_purpose=44)
         message = '\n'.join([
             _('Enter your wallet derivation here.'),
             _('If you are not sure what this is, leave this field unchanged.')
         ])
         presets = (
-            ('legacy BIP44', bip44_derivation(0, False)),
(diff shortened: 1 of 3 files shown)


prusnak:
BIP84 has been merged into the BIPS repo: https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki

PR #3631, opened 2017-12-31
change default relay fee to 1000 sat/kbyte

SomberNight:
Not sure why this is 5000 atm.
https://github.com/bitcoin/bitcoin/blob/v0.15.1/src/policy/policy.h#L32


dabura667:
ACK

It was 5000 before BlueMatt added the variable min_relay_tx_fee logic, which included the bump back down to 1000...

PR #3634, opened 2018-01-01
Avoid modifying self.transactions in prepare_for_verifier

achow101:
In python3, the `.keys()` function returns an iterator, not a list,
so to get a list that can be iterated over, use `list()` instead to
avoid modification of a list while in use.


SomberNight:
Right, looks good.
There's been many cases of this during the py3 migration, though `d.keys()` was usually just wrapped with a `list` cast, so `list(d.keys())`, but `list(d)` should be equivalent.

PR #3643, opened 2018-01-03
fee ui: rounding

SomberNight:
- display info icon with tooltip
    - current icon from icons8 pack (which is already credited), but can be changed ofc
- show pre-rounding values in UI


ecdsa:
note: it would be better to display the final fee in the dialog; the tooltip will be missed by most users.

Issue #3374, opened 2017-11-24
Password protect the JSONRPC interface


jsmad:
The JSONRPC interface is currently completely unprotected, I believe it should be a priority to add at least some form of password protection.

Scans for the JSONRPC interface of Ethereum wallets have already started:
https://www.bleepingcomputer.com/news/security/theres-some-intense-web-scans-going-on-for-bitcoin-and-ethereum-wallets/

Optimizingroman:
Even if a hacker got access to your wallet somehow it's still encrypted with strong password i hope.

jsmad:
The JSONRPC interface is not about your personal wallet at home. It is mostly used by web servers for remotely executing commands, like handling a wallet via a web interface or accepting web payments.

While the electrum daemon is running, someone on a different virtual host of  the web server could easily access your wallet via the local RPC port.

Currently, there is no security/authentication, giving someone access to the RPC port full access to the wallet.

taviso:
Hello, I'm not a bitcoin user, a colleague pointed me at this bug report because localhost RPC servers drive me crazy 😛.

I installed Electrum to look, and I'm confused why this isn't being treated as a critical and urgent vulnerability? If this bug wasn't already open for months, I would have reported this as a vulnerability, but maybe I misunderstand something.

The JSON RPC server is enabled by default, it does use a random port but a website can simply scan for the right port in seconds.

I made you a [demo](https://lock.cmpxchg8b.com/electrum.html). It's very basic, but you get the idea.

If you did set a password, some misdirection is required, but it's still game over, no?

Here is how I reproduced:

- Install Electrum 3.0.3 on Windows.
- Create a new wallet, all default settings. I left the wallet password blank - the default setting.
- Visit https://lock.cmpxchg8b.com/electrum.html in Chrome.
- Wait a few seconds while it guesses the port, then an alert() appears with: `seed: {"id": 0.7398595146147573, "result": "pony south strike horror throw acquire able afford pen lunch monster runway", "jsonrpc": "2.0"}`

devrandom:
Looks like electrum unconditionally returns a positive CORS pre-flight check.  Looks like a serious issue.

mithrandi:
The code doing this dates back to the original implementation: https://github.com/spesmilo/electrum/blob/2c67de8f642a182dd76c706b65c14eeeeee64d03/gui/jsonrpc.py#L32-L42

I suspect this should just be removed, but there may be a reason it's there?

[...]

PR #3659, opened 2018-01-06
Stop allowing CORS for the JSON-RPC server

mithrandi:
As far as I can tell, there is no need to allow this, and doing so poses severe security risks (see #3374).


ecdsa:
thanks for that patch.

PR #3677, opened 2018-01-08
Fix Blockchain.get_target()

pooler:
After d1b8a6fae6d44e78c8093b01321f8d430442b687, `get_target()` should always return just a target.

With this patch, synchronization from scratch should start working again.


SomberNight:
Thank you

PR #3679, opened 2018-01-09
Better ledger version comparison

marceloneil:
Currently the ledger version comparison is a bit lacking. If the bitcoin ledger app was updated to 1.2.0, electrum would assume incompatibility with multi-output, segwit and native segwit.

Instead of adding any new dependencies for this simple task (ie. semver), I simply compare the versions as tuples.

Let me know if I should change anything.

### DIFF
--- a/plugins/ledger/ledger.py
+++ b/plugins/ledger/ledger.py
@@ -29,6 +29,9 @@ MSG_NEEDS_FW_UPDATE_GENERIC = _('Firmware version too old. Please update at') + 
                       ' https://www.ledgerwallet.com'
 MSG_NEEDS_FW_UPDATE_SEGWIT = _('Firmware version (or "Bitcoin" app) too old for Segwit support. Please update at') + \
                       ' https://www.ledgerwallet.com'
+MULTI_OUTPUT_SUPPORT = '1.1.4'
+SEGWIT_SUPPORT = '1.1.10'
+SEGWIT_SUPPORT_SPECIAL = '1.0.4'
 
 
 class Ledger_Client():
@@ -53,6 +56,9 @@ class Ledger_Client():
 
     def i4b(self, x):
         return pack('>I', x)
+
+    def versiontuple(self, v):
+        return tuple(map(int, (v.split("."))))
 
     def get_xpub(self, bip32_path, xtype):
         self.checkDevice()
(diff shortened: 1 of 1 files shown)


plugins/ledger/ledger.py
```diff
@@ -118,12 +123,12 @@ def supports_native_segwit(self):
...
-            self.multiOutputSupported = int(firmware[0]) >= 1 and int(firmware[1]) >= 1 and int(firmware[2]) >= 4
-            self.segwitSupported = (int(firmware[0]) >= 1 and int(firmware[1]) >= 1 and int(firmware[2]) >= 10) or (firmwareInfo['specialVersion'] == 0x20 and int(firmware[0]) == 1 and int(firmware[1]) == 0 and int(firmware[2]) >= 4)
-            self.nativeSegwitSupported = int(firmware[0]) >= 1 and int(firmware[1]) >= 1 and int(firmware[2]) >= 10
+            firmware = firmwareInfo['version']
+            self.multiOutputSupported = self.versiontuple(firmware) >= self.versiontuple(MULTI_OUTPUT_SUPPORT)
+            self.nativeSegwitSupported = self.versiontuple(firmware) >= self.versiontuple(SEGWIT_SUPPORT)
+            self.segwitSupported = self.nativeSegwitSupported or (firmwareInfo['specialVersion'] == 0x20 and self.multiOutputSupported)
 
-            if not checkFirmware(firmware):
+            if not checkFirmware(firmwareInfo):
```
SomberNight: Why? Looks like this was `firmwareInfo['version'].split(".")`.
Though it seems that function just returns True anyway: https://github.com/LedgerHQ/btchip-python/blob/c2bb2166cdaec95795517811c6f6e16038401f1d/btchip/btchipFirmwareWizard.py#L20-L21

plugins/ledger/ledger.py
```diff
@@ -118,12 +123,12 @@ def supports_native_segwit(self):
...
         try:
             firmwareInfo = self.dongleObject.getFirmwareVersion()
-            firmware = firmwareInfo['version'].split(".")
-            self.multiOutputSupported = int(firmware[0]) >= 1 and int(firmware[1]) >= 1 and int(firmware[2]) >= 4
-            self.segwitSupported = (int(firmware[0]) >= 1 and int(firmware[1]) >= 1 and int(firmware[2]) >= 10) or (firmwareInfo['specialVersion'] == 0x20 and int(firmware[0]) == 1 and int(firmware[1]) == 0 and int(firmware[2]) >= 4)
-            self.nativeSegwitSupported = int(firmware[0]) >= 1 and int(firmware[1]) >= 1 and int(firmware[2]) >= 10
+            firmware = firmwareInfo['version']
+            self.multiOutputSupported = self.versiontuple(firmware) >= self.versiontuple(MULTI_OUTPUT_SUPPORT)
+            self.nativeSegwitSupported = self.versiontuple(firmware) >= self.versiontuple(SEGWIT_SUPPORT)
+            self.segwitSupported = self.nativeSegwitSupported or (firmwareInfo['specialVersion'] == 0x20 and self.multiOutputSupported)
```
SomberNight: `self.multiOutputSupported` is `1.1.4` but before this change this was `1.0.4`. Is this change intentional?

SomberNight: I mean regarding the second branch of `self.segwitSupported`

plugins/ledger/ledger.py
marceloneil: In the event that btchip implements the `checkFirmware` function, I'm assuming that they would take the firmware value in the same format that btchip-python returns. Do you think that I should revert it?

plugins/ledger/ledger.py
marceloneil: Oh no! You're totally correct, I missed that. Good catch!

marceloneil:
It might be better to have a `compare_version(v1, v2)` function, it would save some space, thoughts?

SomberNight:
Thanks, I think this is fine as is.

PR #3698, opened 2018-01-11
Check python version

marceloneil:
Raise a readable exception if the user is using an incorrect version of python

### DIFF
--- a/electrum
+++ b/electrum
@@ -26,6 +26,9 @@
 import os
 import sys
 
+if sys.version_info < (3, 5):
+    raise BaseException("Must be using Python 3.5 or higher")
+
 # from https://gist.github.com/tito/09c42fb4767721dc323d
 import threading
 try:


SomberNight:
Note that there is already a hardcoded version here https://github.com/spesmilo/electrum/blob/d83d9277810ab1810ea8df4cec6d620e06fcfcc3/setup.py#L14
This check should be reached for example when using `pip install`.

What is the scenario in which you are trying to trigger that exception?

Also, btw, at the moment python 3.4 should be enough I think, but after merging the asyncio branch, it will have to be raised to at least 3.5.

marceloneil:
Oh, I didn't know that. Somehow two users managed to run electron-cash using python2, so I figured it might be useful here as well: https://github.com/fyookball/electrum/issues/420

ecdsa:
closing since 3.5 is not required now

Issue #3678, opened 2018-01-09
why is it possible to execute arbitrary system commands via python code through the electrum console


n0bd:
wondering why it is currently possible to execute arbitrary system commands via python code through the electrum console, and why there is no whitelisting/blacklisting of allowed methods/functions as this seems like a major security issue
  
inserting a payload such as below in the console, then checking /tmp for the file creation shows the file was successfully created
>> lol=__import__('os');lol.system('/bin/sh -c "touch /tmp/test"');print(lol.path.isfile('/tmp/test')); 
True

attritionorg:
Dupe to https://github.com/spesmilo/electrum/issues/3374 and fixed by 3.0.5 per changelog?

n0bd:
ehh @attritionorg this is not a dupe of that issue by any means

AbdussamadA:
so you can execute python code in the python console? why do you think this is a security risk? anyone who has access to that can just as well run python directly.

n0bd:
@AbdussamadA its a security issue because there is no sandboxing/whitelisting of allowed or disallowed methods to be called, and you most definitely should not be able to execute system level commands from a wallet which is designed to store your funds in

AbdussamadA:
I recommend this be closed because it's just another junk issue.

n0bd:
@AbdussamadA has no understanding of this and no clue as to what hes talking about, and this is definitely not another junk issue. perhaps you should go read one of the many papers on practices for execution of untrusted code before you make such remarks buddy.

blackwat3r:
i can see many relevant reasons as to why this is a security issue, but it seems a bit obvious @AbdussamadA hasnt the first clue in regards to his aforementioned claims.

SmileyChris:
For someone read on the practices for execution of untrusted code, one would hope you also understand responsible disclosure procedures and wouldn't just make a public ticket about what you perceive to be a security vulnerability :tired_face: 

Is there is any possibility for remote execution?

ysangkok:
I don't think there is any indication of remote execution being possible.

Many applications have a scripting console, for example FreeCAD. For example, if you have online banking, there is a JavaScript console in your browser, which is then having full rights to do whatever you can do with your online banking solution.

Sandboxing is a solution for running untrusted code. The code you write in the Python console is trusted. Therefore sandboxing is not applicable.

So I would argue that the Python console is accepting arbitrary Python as it should.

[...]

PR #3700, opened 2018-01-11
Add a warning about harmful code to the console

bauerj:
This adds a warning about malicious code to the console window. It should help prevent users from executing code they found online.

## Screenshots

This warning is certainly hard to miss:
python_2018-01-12_00-08-25

Clicking the message indeed hides it:
python_2018-01-12_00-08-31

Related: #3678

### DIFF
--- a/gui/qt/console.py
+++ b/gui/qt/console.py
@@ -7,6 +7,7 @@ from PyQt5 import QtCore
 from PyQt5 import QtGui
 from PyQt5 import QtWidgets
 from electrum import util
+from electrum.i18n import _
 
 
 if platform.system() == 'Windows':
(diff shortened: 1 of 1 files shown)


SomberNight:
Tested, looks good.


Though depending on QT version, the last line is sometimes not shown; and the text is not wrapped.
Not sure these are worth fixing, unless they are one-liners.

-----
All pics on Debian 9.

### PyQt 5.7

pic3
pic4

### PyQt 5.9

pic1
pic5

bauerj:
Thanks for testing! I didn't find a way to automatically increase the size of the label based on the content size. However, I've enabled word wrapping and increased the height slightly so it should be at least better now.

ecdsa:
do we want the hidden state to be remembered?

bauerj:
Not sure about that. The console is seldom used so the benefit wouldn't be big.
But users might forget that they disabled that warning. I'd rather have the warning after every restart.

dabura667:
No.

People forget.

I don’t mind that Facebook always greets me with a huge DONT PASTE STUFF HERE message.

As long as it doesn’t cover up server banners I say persist for the session until the GUI is closed

SomberNight:
@ecdsa Do you mean clicking it once would hide it forever? It's very easy to click it even by mistake.

@dabura667 Well it covers up server banners somewhat (now more than on above screenshots). But I think the behaviour in the PR of "click it once, and it won't be shown until you re-open Electrum" is fine.

PR #3701, opened 2018-01-11
Fix crash when exporting private keys

marceloneil:
Prevents electrum from crashing when canceling a private key export. This error was caused when the window closed, but a thread continued to try to write to it.

Special thanks to @rt121212121 for helping me fix this one

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -2269,25 +2269,38 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         private_keys = {}
         addresses = self.wallet.get_addresses()
         done = False
+        cancelled = False
         def privkeys_thread():
             for addr in addresses:
                 time.sleep(0.1)
-                if done:
+                if done or cancelled:
                     break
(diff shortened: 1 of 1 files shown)


gui/qt/main_window.py
```diff
@@ -2269,25 +2269,38 @@ def export_privkeys_dialog(self, password):
...
         def show_privkeys():
             s = "\n".join( map( lambda x: x[0] + "\t"+ x[1], private_keys.items()))
             e.setText(s)
             b.setEnabled(True)
             self.show_privkeys_signal.disconnect()
+            nonlocal done
+            done = True
+
+        def on_dialog_closed(*args):
+            nonlocal done
```
SomberNight: I guess `done` does not need to be `nonlocal` here.

marceloneil: Trust me I tried (I attempted to keep it simple), and it didn't seem to work at all

marceloneil:
Yeah its pretty crappy code, but somewhat necessary if we don't want a full rewrite. Even then, there is probably a better way of doing it: but this works

SomberNight:
Thanks.

Issue #3706, opened 2018-01-13
Without --nossl,  -s *:t option is silently ignored


nothingmuch:
When specifying e.g. `-s 127.0.0.1:8000:t` in order to connect to a private server w/o SSL, the CLI option is silently ignored (even if `-1` is specified, and no mention with `-v`), and instead proceeds with automatic server selection as if `-s` was never specified at all.

When *both* the `:t` suffix and `--nossl` are specified, then the `-s` directive takes effect as expected. `--nossl` on its own is not sufficient if `:t` is not also added to the address.

Given the potential privacy implications of connecting to unspecified servers, as well as the lack of diagnostic output (I initially thought the option was simply removed and didn't realize it was still being parsed but wasn't in effect), I think the correct behavior when given an explicit `:t` suffix would be to exit with an error message indicating that `--nossl` is now mandatory, especially is this may silently change peoples' configured setups, by starting to connect to the public servers instead.

See also #3689 and dc388d4c7c541fadb9869727e359edace4c9f6f0 which introduced this change

SomberNight:
I agree that if the user has a selected server (i.e. the `server` config variable is set), we really SHOULD NOT silently connect to a different server.

I believe the current behaviour is by accident, and it was probably missed here
https://github.com/spesmilo/electrum/blob/dc388d4c7c541fadb9869727e359edace4c9f6f0/lib/network.py#L177-L181
that if the `assert` triggers, that is catched by the `except` block; in https://github.com/spesmilo/electrum/commit/dc388d4c7c541fadb9869727e359edace4c9f6f0.

The question is: what should happen if the `server` config variable is in contradiction with `nossl`?

Unfortunately, it would require major changes to properly pass a notification/popup about any of this to the GUI, as the daemon starts the network before the GUI.

I propose keeping the server (host) but changing the protocol (enforcing the value of the `nossl` config var), silently (with only a log message).

SomberNight:
This should now be fixed by https://github.com/spesmilo/electrum/pull/3733 and https://github.com/spesmilo/electrum/pull/3732

PR #3707, opened 2018-01-13
network: (nossl) handle_protocol_contradiction

SomberNight:
closes https://github.com/spesmilo/electrum/issues/3706

see https://github.com/spesmilo/electrum/issues/3706#issuecomment-357404512

### DIFF
--- a/gui/qt/network_dialog.py
+++ b/gui/qt/network_dialog.py
@@ -389,18 +389,6 @@ class NetworkChoiceLayout(object):
         if protocol != self.protocol:
             self.protocol = protocol
 
-    def change_protocol(self, use_ssl):
-        p = 's' if use_ssl else 't'
-        host = self.server_host.text()
-        pp = self.servers.get(host, NetworkConstants.DEFAULT_PORTS)
-        if p not in pp.keys():
-            p = list(pp.keys())[0]
-        port = pp[p]
-        self.server_host.setText(host)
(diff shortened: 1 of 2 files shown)


ecdsa:
that seems too complicated... 
how about not having conflicts in the first place, by simplifying the syntax?
we could drop the protocol letter in the command line: "electrum -s server:port --nossl"

if a default server is set in config, and it conflicts with the --nossl flag, we can simply fail at connecting

SomberNight:
> we could drop the protocol letter in the command line: "electrum -s server:port --nossl"

Yes that would be fine.

> if a default server is set in config, and it conflicts with the --nossl flag, we can simply fail at connecting

Fail how? Consider the following scenario:
- a GUI user previously unchecked the SSL checkbox, and also right-click "Use as server" -selected a server
- he now upgrades and ... ?

SomberNight:
We could "config-upgrade" the `server` variable to force-drop the protocol letter from there; combined with your proposal.

ecdsa:
well, yes, we can drop the protocol letter from the config, but that will still result in a connection failure, because we don't know the correct port number. I dont think we should try to do better, it is fine to fail in that case.

SomberNight:
Oh right, I see what you mean. Failing to connect is okay. What I wanted to avoid is failing -> raising an exception -> "crashing", or similar.

PR #3710, opened 2018-01-13
QT: Only set if hasattr(QtCore.QCoreAplication, "AA_ShareOpenGLContexts")

nothingmuch:
If the attribute does not exist (presumably in older versions of Qt),
this change prevents the following error:

  AttributeError: type object 'Qt' has no attribute 'AA_ShareOpenGLContexts'

### DIFF
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -92,7 +92,8 @@ class ElectrumGui:
         #network.add_jobs([DebugMem([Abstract_Wallet, SPV, Synchronizer,
         #                            ElectrumWindow], interval=5)])
         QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_X11InitThreads)
-        QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_ShareOpenGLContexts)
+        if hasattr(QtCore.Qt, "AA_ShareOpenGLContexts"):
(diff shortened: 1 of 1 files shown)


nothingmuch:
Please note, I have *not* tested this on a version that does have `AA_ShareOpenGLContexts` so I can't confirm this does not break 4f5bc9c but it does eliminate a startup error that was introduced by that commit on my setup

gui/qt/__init__.py
```diff
@@ -92,7 +92,8 @@ def __init__(self, config, daemon, plugins):
         #network.add_jobs([DebugMem([Abstract_Wallet, SPV, Synchronizer,
         #                            ElectrumWindow], interval=5)])
         QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_X11InitThreads)
-        QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_ShareOpenGLContexts)
+        if hasattr(QtCore.QCoreApplication, "AA_ShareOpenGLContexts"):
```
SomberNight: it should be `if hasattr(QtCore.Qt, "AA_ShareOpenGLContexts"):`

nothingmuch: oh crap, so I did have it right the first time, but then I second guessed myself into breaking it =D

I'll rebase a99f237 on top of master and push --force that to this PR

Issue #1752, opened 2016-04-06
[SUGGESTION] Transaction History/Historical Prices


RustyNomad:
When exporting the transaction history the details of the transaction is exported WITHOUT the historical fiat prices. I understand that the historical fiat prices are provided by the plugin and that they are essentially not part of the transaction but it would be a very nice feature if this could be included in the export.

ecdsa:
I agree. this is needed for accounting.

ecdsa:
note: this plugin is going to be merged with the main code

ecdsa:
note: the plugin is merged now.

PR #3724, opened 2018-01-15
Add fiat value to history export

bauerj:
This includes the historical fiat value in the csv/json history export.

Closes: #1752


ecdsa:
replaced by 3f954a8b3d39d2cb45927371a856802104f9182c

PR #3732, opened 2018-01-16
Config upgrades

SomberNight:
~~Depends on https://github.com/spesmilo/electrum/pull/3731
Config version 2 removes the protocol letter from `server` in config.~~

- remove system config
- implement config upgrades
    - can only meaningfully upgrade user_config (not CLI)

EDIT: config version 2 now changes the protocol to `:s` in config


SomberNight:
(Travis fails because I introduced `PrintError.print_stderr` in util in the other PR, and using it here.)

Issue #3670, opened 2018-01-08
Merchant: simple-websocket-server needs update to python 3


ser:
https://github.com/ecdsa/simple-websocket-server needed for merchant to function:

```
byte-compiling /home/pay3/.virtualenvs/electrum/lib/python3.5/site-packages/SimpleWebSocketServer/SimpleExampleServer.py to SimpleExampleServer.cpython-35.pyc
  File "/home/pay3/.virtualenvs/electrum/lib/python3.5/site-packages/SimpleWebSocketServer/SimpleExampleServer.py", line 31
    print self.address, 'connected'
             ^
SyntaxError: Missing parentheses in call to 'print'
```
 
Hint: python3 is supported upstream

ecdsa:
do we still need that repo?

ser:
yes we need - I'll create a PR with an updated version

ser:
I have problems with running the newest upstream on Electrum - debugging now, so it will take more time

ser:
@ecdsa - do you remember why did you have to patch the upstream with timeout?

ecdsa:
iirc, to ensure proper termination of electrum

ser:
@ecdsa - in verbose daemon run i am getting a sequence:
```
connected ('10.100.100.100', 53848)
message received id:0361a81cf1
response {'id': 72, 'params': ['mhjQ2fg3wmo1PNaYkN63hrMtiYjMdGsEyW'], 'method': 'blockchain.address.subscribe', 'jsonrpc': '2.0', 'result': None}
```
and the developer console shows that WS was initiated.

But after payment, electrum says:
```
response {'method': 'blockchain.address.subscribe', 'jsonrpc': '2.0', 'params': ['mhjQ2fg3wmo1PNaYkN63hrMtiYjMdGsEyW', 'c324fce82eac87474f556ecdc63efe41222051a7649b9a5584bf085de4ff1fbf']}
```
and developer console in the browser does not react in any way, looks like nothing is getting received on client side. 

Does the exchange look normally in your opinion? I have no idea how to trace it, as additionally everything is encapsulated in SSL which makes any efforts to monitor transfer more complex.

SomberNight:
Btw, I've tested and the fork for simple-websocket-server is no longer needed.

ecdsa:
thats great, I'm going to delete my fork

Issue #3437, opened 2017-12-03
Does URL notification really work?


igrishaev:
I tried to reproduce the whole scenario when an address of mine got paid and a URL notification should be called (in testnet mode). However, it did not happen. Here are the steps I made:

1. start daemon and load wallet:
```
electrum --testnet daemon start
electrum --testnet daemon load_wallet
```

2. Create payment request
```
electrum --testnet addrequest 1
{
    "URI": "bitcoin:mmZJ2SR7Eji7Qe7SgWFVVh3p8HJ1Hv9MtW?amount=1.",
    "address": "mmZJ2SR7Eji7Qe7SgWFVVh3p8HJ1Hv9MtW",
    "amount": 100000000,
    "amount (BTC)": "1.",
    "exp": null,
    "id": "3799c3948d",
    "memo": "",
    "status": "Pending",
    "time": 1512320573
}
```

3. Add notification with RequestBin service that captures all the requests to the URL:
```
electrum --testnet notify mmZJ2SR7Eji7Qe7SgWFVVh3p8HJ1Hv9MtW https://requestb.in/w9bp68w9
```

4. Get paid to that address using Testnet Sandbox service: https://testnet.manu.backend.hamburg/faucet

The result: calling `--testnet getbalance` displays the new balance within updated wallet history. However, the target URL was not called. The https://requestb.in/w9bp68w9?inspect page does not show any caught calls.

Can you give me any tips on what did I do wrong?

ecdsa:
some other issues have been reported about it.
you should have a look

PR #3734, opened 2018-01-16
websocket: migrate to scripthashes

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/3670

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -604,6 +604,9 @@ class Network(util.DaemonThread):
                 elif method == 'blockchain.scripthash.subscribe':
                     response['params'] = [params[0]]  # addr
                     response['result'] = params[1]
+                elif method == 'blockchain.address.subscribe':
+                    response['params'] = [params[0]]
+                    response['result'] = params[1]
                 callbacks = self.subscriptions.get(k, [])
 
             # update cache if it's a subscription
--- a/lib/network.py
+++ b/lib/network.py
@@ -604,9 +604,6 @@ class Network(util.DaemonThread):
                 elif method == 'blockchain.scripthash.subscribe':
                     response['params'] = [params[0]]  # addr
                     response['result'] = params[1]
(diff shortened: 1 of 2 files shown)


lib/network.py
```diff
@@ -604,6 +604,9 @@ def process_responses(self, interface):
                 elif method == 'blockchain.scripthash.subscribe':
                     response['params'] = [params[0]]  # addr
                     response['result'] = params[1]
+                elif method == 'blockchain.address.subscribe':
```
bauerj: Why is this needed?

SomberNight: I guess the other parts of this commit make it redundant. That part missing/changed was the reason it wasn't working in the first place, but then I decided to just migrate it :P

However, if other parts of the code still use `blockchain.address.subscribe` then it is needed.
/me goes off looking after an example...
such as the `notify` command
https://github.com/spesmilo/electrum/blob/954897c281eecb45b4f690919b4f623aca57ac28/lib/commands.py#L665

SomberNight: So... possibly related? https://github.com/spesmilo/electrum/issues/3437

bauerj: Would it make sense to switch the other parts to scripthash too instead?

SomberNight: Yes, of course. Still, case in point :D

SomberNight: Ok, I've migrated and fixed the notify command in https://github.com/spesmilo/electrum/commit/d808bf057e21ef728e71f926523ad58056f8c526
Do you think I shoul delete this part of the change?

bauerj: Yes, we should try to remove all traces from the client before the server drops support 👍

ser: Amazing, guys :) I can continue work on the websockets javascript client, then :+1:

ser:
I can confirm that I have tested WS and it is working correctly now. Thanks again for help, @SomberNight

PR #3737, opened 2018-01-17
fix: sweeping into same wallet

SomberNight:
Sweeping UTXOs owned by the same wallet does not work; it starts signing and then finishes without actually signing.

`wallet.add_input_sig_info` changes the pubkeys from 02-04 to ff (for deterministic wallet)

https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/wallet.py#L1735-L1740

and then `Transaction.sign` won't find the pubkeys:

https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/transaction.py#L933-L942

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -887,8 +887,9 @@ class Abstract_Wallet(PrintError):
         if fixed_fee is None and config.fee_per_kb() is None:
             raise NoDynamicFeeEstimates()
 
-        for item in inputs:
-            self.add_input_info(item)
+        if not is_sweep:
+            for item in inputs:
+                self.add_input_info(item)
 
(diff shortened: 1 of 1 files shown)


ecdsa:
currently the is_sweep parameter is passed but it does nothing. why is that?

SomberNight:
The method is overridden in the 2FA child class and the argument is used there
https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/plugins/trustedcoin/trustedcoin.py#L251-L256

PR #3739, opened 2018-01-18
Intruduction of getfee command, fixes #3704

### LINKED ISSUE #3704: RFC: new command: getfee
# Request For Comments

### Problem

When you use Electrum as a merchant, you want to suggest your customers a correct fee to avoid problems. Electrum already has this information, why then use an another external source to create another point of failure?

### Proposed Solution

Creation of another command ```getfee``` - which will operate via RPC as well. It's code will be identical to currently displayed fee when you issue a command line ```electrum daemon status```.

### Who will write it

If concept is agreed, I will propose #PR for this.

ser:


lib/commands.py
```diff
@@ -672,6 +672,12 @@ def is_synchronized(self):
         """ return wallet synchronization status """
         return self.wallet.is_up_to_date()
 
+    @command('')
```
bauerj: Shouldn't this use `'n'` instead?

ser: Fixed.

ecdsa:
can you rename it "getfeerate" ?

ecdsa:
I had to merge manually because commits were not rebased

ser:
Sorry, next time I will remember.

PR #3750, opened 2018-01-20
Add default keyserver (prepare-wine.sh)

lzsaver:
@SomberNight, can you approve?

It may be required to run _prepare-wine.sh_ twice if there is no _--keyserver_ option and no _~/.gnupg_ directory. This should fix it.

Follow-up: 0dca33e2 (PR #3079).


SomberNight:
I've tested running the current build script in a ~fresh Ubuntu 16.04 VM, and indeed it errored out with
```
gpg: no keyserver known (use option --keyserver)
gpg: keyserver receive failed: bad URI
```
It worked with the change.
Thanks.

PR #3760, opened 2018-01-22
Use python3 for virtualenv

lzsaver:
It will not create virtualenv for python3 if python2 is used by default on the system. This should fix it.

### DIFF
--- a/contrib/freeze_packages.sh
+++ b/contrib/freeze_packages.sh
@@ -6,8 +6,8 @@ contrib=$(dirname "$0")
 
 which virtualenv > /dev/null 2>&1 || { echo "Please install virtualenv" && exit 1; }
 
-rm $venv_dir -rf
-virtualenv $venv_dir
+rm "$venv_dir" -rf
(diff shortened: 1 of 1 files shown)


bauerj:
That should only be true if virtualenv is installed with Python 2 too, right?

contrib/freeze_packages.sh
```diff
@@ -6,8 +6,8 @@ contrib=$(dirname "$0")
 
 which virtualenv > /dev/null 2>&1 || { echo "Please install virtualenv" && exit 1; }
 
-rm $venv_dir -rf
-virtualenv $venv_dir
+rm "$venv_dir" -rf
+python3 -m virtualenv $venv_dir
```
bauerj: Please change this to use e.g. -p $(which python3) as that seems to be the preferred way to do this.

lzsaver:
Yes.

contrib/freeze_packages.sh
lzsaver: Not sure about that. But as you wish.

Issue #3738, opened 2018-01-17
Imported Addresses wallet can only "create" txns with p2pkh inputs


SomberNight:
Offline signing with imported wallets only works for p2pkh inputs.

https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/wallet.py#L1560-L1561

^ that line, it all starts with that line :)
The point is that an Imported Addresses wallet does not know what `txin_type` should be.

So...
the input script will be constructed as if it was p2pkh
https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/transaction.py#L718-L719

and the partial txn will be serialised using the legacy format because the input will never be seen as segwit

https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/transaction.py#L684-L690

https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/transaction.py#L805-L806

https://github.com/spesmilo/electrum/blob/0dfaf9b97013ce9221bc8589ad806e52643e727d/lib/transaction.py#L815-L821

which also means it won't contain `value` for the input.

SomberNight:
Maybe we should scrap transaction.py and implement https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki :)

ecdsa:
bip174 is too big and ambitious. 
implementing it would slow down other high priority projects.

PR #3762, opened 2018-01-23
change partial txn serialization format for imported addresses txins

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/3738
fixes https://github.com/spesmilo/electrum/issues/2897
fixes https://github.com/spesmilo/electrum/issues/3918

- if any inputs have `address` txin_type, serialize the txn in segwit format
    - make sure the partial txns contain `value` for all these inputs
- put `OP_INVALIDOPCODE` + `OP_0` + `push_script(fd_extended_pubkey)` into the `scriptSig`
    - I meant `OP_0` as an internal version number but it could be left out I guess
- the `address` txin_type, unlike before, now deliberately has a distinct serialization hence, and the type is also kept as `address` when deserialized. the idea is that the offline signing wallet can fill in the data for the txin after deserializing the txn
    - this is currently done in `wallet.add_input_info_to_all_inputs`

### DIFF
--- a/gui/qt/transaction_dialog.py
+++ b/gui/qt/transaction_dialog.py
@@ -72,7 +72,7 @@ class TxDialog(QDialog, MessageBoxMixin):
         # Take a copy; it might get updated in the main window by
         # e.g. the FX plugin.  If this happens during or after a long
         # sign operation the signatures are lost.
-        self.tx = copy.deepcopy(tx)
+        self.tx = tx = copy.deepcopy(tx)
         try:
             self.tx.deserialize()
         except BaseException as e:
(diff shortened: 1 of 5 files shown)


SomberNight:
Rebased on master.

SomberNight:
Rebased and cleaned up a bit.

SomberNight:
Actually maybe changing the serialization format itself is not strictly necessary.
The format on master should be able to handle what we want thanks to `fd` extended pubkeys.
We would need to distinguish between `p2pkh` and `address` in `parse_scriptSig`:
https://github.com/spesmilo/electrum/blob/86e2f3f561bf651cfcf270324d5c2b794e491dcf/lib/transaction.py#L339-L356
but this would involve some guessing..

So I think it's better to have the partial format clearly distinguishable, as in this PR.

Also, this just made me realise that with this approach, `fd` pubkeys are no longer needed.... so those parts of the code might be removed...

SomberNight:
note: currently there is a bug where an incorrect txid is shown in the GUI for unsigned segwit txns.
This is due to `can_sign` having a side-effect of adding extra info for the txn, and `can_sign` not being called early enough... Maybe this magic should not be a part of `can_sign`.
https://github.com/spesmilo/electrum/blob/86e2f3f561bf651cfcf270324d5c2b794e491dcf/gui/qt/transaction_dialog.py#L217-L225
^ txid calculation is before `can_sign` is called

SomberNight:
note: this actually resolves https://github.com/spesmilo/electrum/issues/2897 too, in a relatively nice way, as `add_input_info` and `add_input_sig_info` set `txin['x_pubkeys']` to the `ff`/`fe` pubkey

PR #3777, opened 2018-01-27
fix wallet.check_history

SomberNight:
https://github.com/spesmilo/electrum/commit/5be78950ca31ba903b4bd0f575511eeb732e114c broke wallet.check_history AFAICT.
See here https://github.com/spesmilo/electrum/commit/5be78950ca31ba903b4bd0f575511eeb732e114c#diff-9a32635d4a93b6e41c91fec9e4001d4aL200

This PR restores the behaviour to before the above mentioned commit.

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -272,10 +272,15 @@ class Abstract_Wallet(PrintError):
     @profiler
     def check_history(self):
         save = False
-        mine_addrs = list(filter(lambda k: self.is_mine(self.history[k]), self.history.keys()))
-        if len(mine_addrs) != len(self.history.keys()):
+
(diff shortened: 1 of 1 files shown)


SomberNight:
I added another commit to optimise time complexity.
`wallet.is_mine` is incredibly inefficient. It usually does linear operations (in the number of addresses in the wallet) but sometimes even more (for Imported Wallet, it even sorts the addresses!).

For a standard wallet with 50k addresses and no txns, time taken (in seconds) without this PR:
```
[profiler] check_history 59.0495
```
TIme taken with this PR:
```
[profiler] check_history 0.0660
```

Feel free to just cherry pick the first commit if the second looks ugly but we should do something with `wallet.is_mine` in any case. It is a major factor in the time taken when opening a large wallet.

PR #3781, opened 2018-01-27
fine tuning the fix for 1-sat fees on ledger nano s

alfonumeric:
if u've already used electron  cash for bch and ledger nano-s for bch or eth then allow yourself at least an hour to apply this fix which hooks up the ledger and electron-cash wallets. 

If you're totally new to electron-cash & ledger  nano-s  than allow a full-on crypto-weekend to learn the special ropes needed for detecting low flying 1-sats..

set up your bch wallet to operate on the split chain. install bitcoin cash wallet 1.1.8 from the chrome browser and launch it..

click both ledger buttons together on the bitcoin cash window to open your bch wallet in ledger, check that browser-support = no on the device...toggle between "use wallet to view" and "settings" then dual-click on settings

then install electroncash.org as follows

1) check u have the latest version here https://github.com/fyookball/keys-n-hashes/tree/master/sigs-and-sums ... if youre on windows, download the 7-zip tool http://www.7-zip.org/ ..it has a checksum tool

2) file / new-restore / standard / add hardware wallet / next ...

3) the native electron wallet is called default-wallet, when the ledger connects the second wallet on electron it will be called wallet-1 or wallet-2

4) after the first connection, installing electron subsequently will remember wallet-1 or 2 and take u directly there where u can see the ledger addresses

5) when u send from ledger u will see the tx show up in electron way before ledger picks it up, so u must not disconnect electron until your funds go through via 6 confirmations

TIPS FOR ESTABLISHING THE CONNECTION

1) everytime u hit an error on electroncash.org reboot, open the bch wallet with the code even though your timeout period maybe 10 mins

2) try a different usb slot or a different cable

3) if new/restore takes u straight into wallet-1 or 2 and u can see your ledger addresses called "my account" inside electron, then the connection is ok, even though u get no confirmation message and even though u cant see you additional ledger adresses..

4) if u get the "install wizard" dialog when u click new/restore , then click choose to see what wallets u have got if u see that wallet-1 and wallet-2 have both been created then delete both and start again

5) if u get the "install wizard" dialog when u click new/restore check and u see either the wallet-1 or wallet-2 screen, then check that u are seeing your ledger addreses on wallet-1 or 2

TIPS FOR HELPING TX SUCCEED

1) [need help here] it seems electron cash supports only 1 account on ledger
do we send/ receive via electron cash, nanos or either? 

2) if u have more than 1 account and u reset app data...via settings on chrome browser support, click on tools/ reset application data. Next time u go into ledger your bch balance will look like zero. 
but when u recoonect with electroncash, the funds in your main account will show up first and then several hours later the funds in your additional accounts should also be recovered.


bauerj:
I'm not sure what you want to do but we don't want to merge all the changes from Electron Cash back.

PR #3782, opened 2018-01-27
Optimisations in wallet.py

SomberNight:
`wallet.is_beyond_limit` is slow because is looks up addr indices in a list and because it copies sublists

This is a problem as this method gets called inside a for loop when opening a wallet:
https://github.com/spesmilo/electrum/blob/7ab9fa5be48cf5aefec98d94657241c73a6e280c/gui/qt/address_list.py#L109

`wallet.is_mine` is also slow. It usually does linear operations (in the number of addresses in the wallet) but sometimes even more (for Imported Wallet, it even sorts the addresses!).

This is also a problem for the same reason:
https://github.com/spesmilo/electrum/blob/7ab9fa5be48cf5aefec98d94657241c73a6e280c/lib/wallet.py#L275

-----

- make `wallet.get_address_index` faster by storing an addr->index dict, and use it
- make one fewer list copy
- make `wallet.is_mine` faster by looking up in dicts

-----

Opening a wallet with 50k addresses without PR:
```
[WalletStorage] wallet path /home/user/.electrum/testnet/wallets/test_large_wallet
[profiler] load_transactions 0.0001
[profiler] build_reverse_history 0.0070
[profiler] save_transactions 0.0652
[profiler] check_history 57.0266
[profiler] on_update 0.1118
[profiler] load_wallet 31.1191
[profiler] on_update 0.0912
```

With PR:
```
[WalletStorage] wallet path /home/user/.electrum/testnet/wallets/test_large_wallet
[profiler] load_transactions 0.0000
[profiler] build_reverse_history 0.0064
[profiler] check_history 0.0687
[profiler] on_update 0.1119
[profiler] load_wallet 1.6227
[profiler] on_update 0.0851
```

-----

This PR depends on https://github.com/spesmilo/electrum/pull/3777

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -62,6 +62,7 @@ from .fee_slider import FeeSlider
 
 from .util import *
 
+from electrum.util import profiler
 
 class StatusBarButton(QPushButton):
     def __init__(self, icon, tooltip, func):
@@ -325,6 +326,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
             self.print_error('close_wallet', self.wallet.storage.path)
         run_hook('close_wallet', self.wallet)
 
+    @profiler
     def load_wallet(self, wallet):
         wallet.thread = TaskThread(self, self.on_error)
         self.wallet = wallet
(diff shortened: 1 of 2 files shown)


SomberNight:
Note: the tests fail because https://github.com/spesmilo/electrum/pull/3777 is needed.

PR #3787, opened 2018-01-28
Enable sorting of list widgets

bauerj:
This allows users to sort the list widgets by column value simply by clicking on the column they want to sort by.

It is based on fyookball/electrum#512 by @ContrarianUnderscoreUnderscore.

---
bildschirmfoto 2018-01-28 um 13 11 34
 This is the default sorting (status, conf)

---
bildschirmfoto 2018-01-28 um 13 11 53
This is sorted by amount.

### DIFF
--- a/gui/qt/address_list.py
+++ b/gui/qt/address_list.py
@@ -38,6 +38,7 @@ class AddressList(MyTreeWidget):
         MyTreeWidget.__init__(self, parent, self.create_menu, [], 1)
         self.refresh_headers()
         self.setSelectionMode(QAbstractItemView.ExtendedSelection)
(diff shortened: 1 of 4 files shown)


ecdsa:
Currently MyTreeWidget is a child class of QTreeWidget. We do not need that complexity anymore, because we no longer render addresses as a tree. The only place wehere we display a tree is in the network dialog, and it does not need to inherit from MyTreeWidget.

If we use a child class of QListWidget, we will have sortable columns for free, and this PR will not need to define an extra class.

ecdsa:
merged, after fixing conflicts manually

PR #3810, opened 2018-01-31
conflicting transactions

SomberNight:
Merging the `local_tx` branch introduced the possibility of having conflicting transactions in history.

atm this PR uses a new data structure `spent_outpoints` instead of extending `pruned_txo`

### DIFF
--- a/gui/qt/transaction_dialog.py
+++ b/gui/qt/transaction_dialog.py
@@ -179,10 +179,13 @@ class TxDialog(QDialog, MessageBoxMixin):
         self.main_window.sign_tx(self.tx, sign_done)
 
     def save(self):
-        self.wallet.add_transaction(self.tx.txid(), self.tx)
+        if not self.wallet.add_transaction(self.tx.txid(), self.tx):
+            self.show_error(_("Transaction could not be saved. It conflicts with current history."))
+            return
         self.wallet.save_transactions(write=True)
 
-        self.main_window.history_list.update()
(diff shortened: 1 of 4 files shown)


SomberNight:
Questions/scenarios:
1. user bumps fee of an RBF txn but does not broadcast it, rather saves it locally. what should happen?
2. should conflicting txns be allowed in wallet if they are local
    - note: maybe revert https://github.com/spesmilo/electrum/commit/8a71e46e9b9940932e1d25b75047ea36f535541c -- it's not perfect anyway (saving local txns currently "does not work" due to that commit)
3. should wallet try to spend from local txns?
    - note: coinchooser needs changes. e.g. here https://github.com/spesmilo/electrum/blob/5c1da002add5a8280737e67a5d9de7979bf08783/lib/coinchooser.py#L295-L297

ecdsa:
1. this means the locally saved tx would conflict with a mempool tx. this should not be allowed, per point 2. in case of conflict, mempool should have priority.
2. conflicting transactions should never be allowed. Anything else would open a UX nightmare. 8a71e46 should be reverted.
3. yes we should be able to spend outputs from local tx

SomberNight:
Don't merge yet, I'm hunting down a bug...

ecdsa:
ok, can you squash your commits?

SomberNight:
done

ecdsa:
thanks

PR #3821, opened 2018-02-01
add block explorer: blockonomics.co

SomberNight:
Sorry for the horrible diff, but blockonomics.co structures their URLs differently... :/

### DIFF
--- a/lib/util.py
+++ b/lib/util.py
@@ -431,39 +431,41 @@ def time_difference(distance_in_time, include_seconds):
         return "over %d years" % (round(distance_in_minutes / 525600))
 
 mainnet_block_explorers = {
-    'Biteasy.com': ('https://www.biteasy.com/blockchain',
-                        {'tx': 'transactions', 'addr': 'addresses'}),
-    'Bitflyer.jp': ('https://chainflyer.bitflyer.jp',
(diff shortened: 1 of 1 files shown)


lib/util.py
```diff
@@ -431,39 +431,41 @@ def time_difference(distance_in_time, include_seconds):
...
 }
 
 testnet_block_explorers = {
-    'Blocktrail.com': ('https://www.blocktrail.com/tBTC',
-                       {'tx': 'tx', 'addr': 'address'}),
-    'system default': ('blockchain:',
-                       {'tx': 'tx', 'addr': 'address'}),
+    'Blocktrail.com': ('https://www.blocktrail.com/tBTC/',
+                       {'tx': 'tx/', 'addr': 'address/'}),
+    'system default': ('blockchain://000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943/',
```
bauerj: This doesn't look right.

SomberNight: I changed that based on this https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki#examples

bauerj: Alright then, good catch!

PR #3823, opened 2018-02-02
show address on trezor for multisig

farmerwampum:
Naive removal of the code to decide whether to show menu option.
And Trezor must be the first seed in the wallet.

### DIFF
--- a/plugins/trezor/qt_generic.py
+++ b/plugins/trezor/qt_generic.py
@@ -188,7 +188,6 @@ class QtPlugin(QtPluginBase):
 
     @hook
     def receive_menu(self, menu, addrs, wallet):
-        keystore = wallet.get_keystore()
         if len(addrs) != 1:
             return
         for keystore in wallet.get_keystores():
--- a/plugins/trezor/qt_generic.py
+++ b/plugins/trezor/qt_generic.py
@@ -189,10 +189,14 @@ class QtPlugin(QtPluginBase):
     @hook
     def receive_menu(self, menu, addrs, wallet):
         keystore = wallet.get_keystore()
-        if type(keystore) == self.keystore_class and len(addrs) == 1:
-            def show_address():
-                keystore.thread.add(partial(self.show_address, wallet, addrs[0]))
-            menu.addAction(_("Show on %s") % self.device, show_address)
+        if len(addrs) != 1:
+            return
+        for keystore in wallet.get_keystores():
+            if type(keystore) == self.keystore_class:
+                def show_address():
+                    keystore.thread.add(partial(self.show_address, wallet, keystore, addrs[0]))
+                menu.addAction(_("Show on %s") % self.device, show_address)
+                break
 
     def show_settings_dialog(self, window, keystore):
         device_id = self.choose_device(window, keystore)
--- a/plugins/trezor/qt_generic.py
+++ b/plugins/trezor/qt_generic.py
@@ -188,8 +188,6 @@ class QtPlugin(QtPluginBase):
 
     @hook
     def receive_menu(self, menu, addrs, wallet):
-        if type(wallet) is not Standard_Wallet:
-            return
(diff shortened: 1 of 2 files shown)


SomberNight:
Why does Trezor need to be the first cosigner?

farmerwampum:
Not sure why it won't show the option.

plugins/trezor/qt_generic.py
```diff
@@ -188,8 +188,6 @@ def create_handler(self, window):
 
     @hook
     def receive_menu(self, menu, addrs, wallet):
-        if type(wallet) is not Standard_Wallet:
-            return
         keystore = wallet.get_keystore()
         if type(keystore) == self.keystore_class and len(addrs) == 1:
```
SomberNight: `wallet.get_keystore()` returns the first keystore in the wallet, that is why it only works if Trezor is first

SomberNight:
You should do something like this:

```patch
diff --git a/plugins/trezor/plugin.py b/plugins/trezor/plugin.py
index 9bdf7903..1f45e2f3 100644
--- a/plugins/trezor/plugin.py
+++ b/plugins/trezor/plugin.py
@@ -246,16 +246,16 @@ class TrezorCompatiblePlugin(HW_PluginBase):
         raw = bh2u(signed_tx)
         tx.update_signatures(raw)
 
-    def show_address(self, wallet, address):
-        client = self.get_client(wallet.keystore)
+    def show_address(self, wallet, keystore, address):
+        client = self.get_client(keystore)
         if not client.atleast_version(1, 3):
-            wallet.keystore.handler.show_error(_("Your device firmware is too old"))
+            keystore.handler.show_error(_("Your device firmware is too old"))
             return
         change, index = wallet.get_address_index(address)
-        derivation = wallet.keystore.derivation
+        derivation = keystore.derivation
         address_path = "%s/%d/%d"%(derivation, change, index)
         address_n = client.expand_path(address_path)
-        script_gen = wallet.keystore.get_script_gen()
+        script_gen = keystore.get_script_gen()
         if script_gen == SCRIPT_GEN_NATIVE_SEGWIT:
             script_type = self.types.InputScriptType.SPENDWITNESS
         elif script_gen == SCRIPT_GEN_P2SH_SEGWIT:
diff --git a/plugins/trezor/qt_generic.py b/plugins/trezor/qt_generic.py
index 0fff2019..a4b9c07c 100644
--- a/plugins/trezor/qt_generic.py
+++ b/plugins/trezor/qt_generic.py
@@ -188,11 +188,14 @@ class QtPlugin(QtPluginBase):
 
     @hook
     def receive_menu(self, menu, addrs, wallet):
-        keystore = wallet.get_keystore()
-        if type(keystore) == self.keystore_class and len(addrs) == 1:
-            def show_address():
-                keystore.thread.add(partial(self.show_address, wallet, addrs[0]))
-            menu.addAction(_("Show on %s") % self.device, show_address)
+        if len(addrs) != 1:
+            return
+        for keystore in wallet.get_keystores():
+            if type(keystore) == self.keystore_class:
+                def show_address():
+                    keystore.thread.add(partial(self.show_address, wallet, keystore, addrs[0]))
+                menu.addAction(_("Show on %s") % self.device, show_address)
+                break
 
     def show_settings_dialog(self, window, keystore):
         device_id = self.choose_device(window, keystore)
```

Feel free to do it another way though.

plugins/trezor/qt_generic.py
```diff
@@ -188,13 +188,15 @@ def create_handler(self, window):
 
     @hook
     def receive_menu(self, menu, addrs, wallet):
-        if type(wallet) is not Standard_Wallet:
-            return
         keystore = wallet.get_keystore()
```
SomberNight: This line is no longer needed as the loop below overwrites the value of `keystore`.

SomberNight:
Thank you.

PR #3826, opened 2018-02-02
migrated some commands from addr to scripthash

SomberNight:
Note: how should `getproof` be handled? We could just delete the method completely.


ecdsa:
yes, we can delete getproof

Issue #3237, opened 2017-11-07
Electrum crashes on bad localization message


pokazef:
When a localization message omits a substitution variable, Electrum crashes.

For example, in the FR locale, "Connected to %d nodes" was translated to "Connecté aux noeuds".
In this case, upon opening the network dialog, Electrum 3.0.1 terminates with this trace:

    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/dist-packages/electrum_gui/qt/main_window.py", line 85, in onPress
        self.func()
      File "/usr/local/lib/python3.5/dist-packages/electrum_gui/qt/main_window.py", line 1761, in <lambda>  
        self.status_button = StatusBarButton(QIcon(":icons/status_disconnected.png"), _("Network"), lambda: self.gui_object.show_network_dialog(self))
      File "/usr/local/lib/python3.5/dist-packages/electrum_gui/qt/__init__.py", line 169, in show_network_dialog
        self.network_updated_signal_obj)
      File "/usr/local/lib/python3.5/dist-packages/electrum_gui/qt/network_dialog.py", line 47, in __init__ 
        self.nlayout = NetworkChoiceLayout(network, config)
      File "/usr/local/lib/python3.5/dist-packages/electrum_gui/qt/network_dialog.py", line 322, in __init__
        self.update()
      File "/usr/local/lib/python3.5/dist-packages/electrum_gui/qt/network_dialog.py", line 358, in update  
        status = _("Connected to %d nodes.")%n if n else _("Not connected")
    TypeError: not all arguments converted during string formatting

Obviously, the right fix here is to correct the localized message, but could there be a way to make this kind of error non fatal?

SomberNight:
I am wondering if this could be fixed in a generic way by somehow modifying the method "doing the translation"

https://github.com/spesmilo/electrum/blob/d599f11cceb0b74d092befff019097d0f0d3f2ac/lib/i18n.py#L35-L37

such that it would (for example) compare if `x` and `language.gettext(x)` are "compatible" format strings and if they aren't, return `x`.

maxit77:
Thanks. I solved momentarily the problem by switching to english language.

ecdsa:
closing this. I will fix other occurences as they are found. 
(I don't think we can fix it in a generic way, at least not easily)

pokazef:
> maybe we should replace all occurences of % with string.format()

Right, Python seems to simply ignore missing replacement fields with the new string formatting API. That should work, thanks :)

PR #3834, opened 2018-02-04
use string.format instead of old style (%) formatting

SomberNight:
Found more translated strings using % formatting.

### DIFF
--- a/gui/kivy/main_window.py
+++ b/gui/kivy/main_window.py
@@ -631,7 +631,7 @@ class ElectrumWindow(App):
             if not self.wallet.up_to_date or server_height == 0:
                 status = _("Synchronizing...")
             elif server_lag > 1:
-                status = _("Server lagging (%d blocks)"%server_lag)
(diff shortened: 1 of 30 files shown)


dabura667:
Not to make a comment on whether this change is good or bad. Just adding a perspective of areas affected by this PR:

1. Crowdin will reset all of these strings to untranslated for all languages. But since the %s versions will be in the TM, it should be easy for the translators to go through and change them... but it's still a bit of mundane work, as Crowdin isn't as powerful as grep etc.

I wonder if there would be a way to download all languages current states, replace the %s with {}, then replace them in all the language files as well and re-upload them all without needing to bother the translators... I am not too familiar.

ecdsa:
@dabura667 this change is made in order to protect Electrum from crashes caused by Crowdin translators. I understand their work is affected by this, but this is the best solution we have found.

SomberNight:
More context: https://github.com/spesmilo/electrum/issues/3237

PR #3841, opened 2018-02-04
address filter: add "all" filter option besides "receiving" and "change"

### LINKED ISSUE #3356: Minor UI issue: Can't "spend from" receive & change addresses at same time
Under the new 3.0.2 ui, you can't see "receive" and "change" addresses at the same time, making it impossible to select a group of "spend from" addresses that include both types at once. 

The 2.x UI displayed this as a tree view, allowing picking a mixture.  Maybe 3.0.3 could add an "all" option to the "receive/change" filter at the top, allowing this ability again?

SomberNight:
address filter: (qt gui) add "all" filter option besides "receiving" and "change"

@shsmith asked for this on IRC

Closes #3356

### DIFF
--- a/gui/qt/address_list.py
+++ b/gui/qt/address_list.py
@@ -38,11 +38,11 @@ class AddressList(MyTreeWidget):
         MyTreeWidget.__init__(self, parent, self.create_menu, [], 1)
         self.refresh_headers()
         self.setSelectionMode(QAbstractItemView.ExtendedSelection)
-        self.show_change = False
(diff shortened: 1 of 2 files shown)


SomberNight:
Also closes https://github.com/spesmilo/electrum/issues/3356

SomberNight:
(do we want this for kivy too?)

ecdsa:
thanks. yes we could have that for kivy too

PR #3846, opened 2018-02-05
Add script to check and sign executables on Windows

bauerj:
This is based on a proposal by @EagleTM 

Signing Windows builds using the Authenticode key is simple enough. This script integrates the deterministic builds into the process by checking if the files match with someone else's files.

osslsigncode needs to be installed in order to use this script.

### DIFF
--- a/contrib/build-wine/README.md
+++ b/contrib/build-wine/README.md
@@ -36,3 +36,30 @@ $ wine --version
 2. Make sure `/opt` is writable by the current user.
 3. Run `build.sh`.
 4. The generated binaries are in `./dist`.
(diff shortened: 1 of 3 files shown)


ecdsa:
is it possible to check that the signed binaries are derived from the deterministic binaries?
I tried 'osslsigndecode remove-signature', but I do get the initial file back

ecdsa:
looking at the diff, there seems to be a timestamp difference, and some trailing zeroes at the end of the file (paddding)

ecdsa:
merged manually after fixes

PR #3851, opened 2018-02-05
Reset scroll position after updating lists

bauerj:


ecdsa:
I  merged it, but this is not a good solution, because we currently use a dirty hack to refresh lists:
we call list.clear() and then regenerate the whole list.

Qt has a way to keep a list synchronized with a model, see QAbstractItemModel and QModelIndex
The proper way to display lists would be to use these classes.

Issue #3255, opened 2017-11-09
Fee selection not satoshi accurate. Fee is round up


Raulo:
3.0.1 version.

Fee selection from GUI (but surprisingly also from command line) is not accurate up to satoshi level. E.g,, for 142 byte transaction, lowest fee possible is 200 satoshi. If one selects max funds, full satoshi resolution is available. Tested both standard and segwit wallets.

SomberNight:
Could you clarify what "142 byte" means in your particular example?

https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations

If you've obtained that number from Electrum, it should be "virtual transaction size":
https://github.com/spesmilo/electrum/blob/ccc94155429241c3d344183a4efdddbbcf6736cf/lib/transaction.py#L831-L839

Raulo:
Yes, virtual size. But for non-segwit transactions (that also are not satoshi level accurate) size=virtual size. 

For a 224 byte (non-segwit) transaction, specifying 2.24 bits fee results in rounding up to 3.00 bits. Also, e.g. 3.5 bits fee is rounded up to 4 bits

SomberNight:
As a workaround, you can set `wallet.relayfee = (lambda: 0)` in the console (if using the QT GUI). This will allow you to broadcast a tx with an arbitrarily low fee but if it is below the relay fee of the Electrum server, you will be presented with an error message from the server.

SomberNight:
Disregard above comment, I see the issue.

SomberNight:
Indeed, fees can only be set to 100 sat precision (in most cases), due to this code snippet:
https://github.com/spesmilo/electrum/blob/ce4be1f8f236cf1ad33ee1aae2c4599bcb8e8e5a/lib/coinchooser.py#L147-L151

Changing this would have privacy implications. Though a choice could be given to the user by e.g. introducing a new CoinChooser, possibly.

Raulo:
I see. It explains why if one chooses Max (no change), there is full precision. I realized now that if you send amount with full satoshi accuracy, the fee is up to satoshi level, too.  

And indeed replacing amounts.append(amount) with amounts.append(remaining) fixes the problem 

But the benefits of the code seem too elusive to me to institute a tax to miners. Sure if I send, say, 6786 bits, the change may look like the amount sent. But if I send 1 BTC, rounding to 1 bit will not mask the obvious change.

principiante:
[...]

PR #3878, opened 2018-02-08
coinchooser: make output value rounding configurable

SomberNight:
- introduce a config variable `coin_chooser_output_rounding` that enables/disables output value rounding by the coinchooser
- by default it is disabled
- expose control of config variable to qt gui
- also; changed how the rounding works if there is only one change output
    - the rounding logic seems to have been written with multiple change outputs in mind; where it makes sense to allow a change output to have more-or-less precision than other outputs
    - with only one change output, the ideal outcome should be to have a change output that doesn't have the most precision; with this change we aim for as much precision as the otherwise most precise output

closes https://github.com/spesmilo/electrum/issues/3255


SomberNight:
updated.

Issue #3871, opened 2018-02-08
Windows build script fails: trezor->pyblake2->vc++


SomberNight:
Trezor now depends on pyblake2: https://github.com/trezor/python-trezor/commit/1ab602423c6f53916e6fdd35841aaa19bfeae4a7

The Windows build script fails:

```
  Downloading trezor-0.9.0.tar.gz (119kB)
    100% |################################| 122kB 704kB/s ta 0:00:01
Requirement already satisfied: urllib3==1.22 in c:\python3.5.4\lib\site-packages (from -r ../../deterministic-build/requirements-hw.txt (line 18))
Requirement already satisfied: setuptools>=19.0 in c:\python3.5.4\lib\site-packages (from hidapi==0.7.99.post21->-r ../../deterministic-build/requirements-hw.txt (line 7))
Installing collected packages: hidapi, btchip-python, click, Cython, mnemonic, keepkey, libusb1, pyblake2, trezor
  Running setup.py install for btchip-python ... done
  Running setup.py install for mnemonic ... done
  Running setup.py install for keepkey ... done
  Running setup.py install for libusb1 ... done
  Running setup.py install for pyblake2 ... error
    Complete output from command c:\python3.5.4\python.exe -u -c "import setuptools, tokenize;__file__='C:\\users\\user\\Temp\\pip-build-llmiyezh\\pyblake2\\setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record C:\users\user\Temp\pip-je1yc8kx-record\install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_ext
    building 'pyblake2' extension
    error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": http://landinghub.visualstudio.com/visual-cpp-build-tools
    
    ----------------------------------------
[...]

PR #3883, opened 2018-02-10
Test Windows build using Travis

bauerj:
This makes Travis run the Windows build script on every commit.

This way we notice if the build fails because:
* We broke something
* An upstream package broke something
* A new Wine release broke something

It also prints the resulting hashes for the binaries which might be useful for a quick comparison.


bauerj:
So due to #3871 the build should fail.

PR #3886, opened 2018-02-11
Various requirements.txt related changes

bauerj:
This will:

* Merge #3545 
* Move requirements.txts to contrib/requirements
* Introduce a requirements-binaries.txt which contains requirements exclusive for our binaries
* Freezes versions for the remaining packages in Wine/macOS builds


SomberNight:
One last thing; it would be nice to include pycryptodomex in the linux binaries. :P

PR #3887, opened 2018-02-11
Workaround for PyBlake2 build issues

bauerj:
This is a workaround for the issue described in #3871.

It will install a binary wheel of pyblake2 instead of the source package. Hopefully they will distribute the wheels themselves in the future.


bauerj:
There is also #3873 to fix the same issue but I don't like that approach since it complicates the build and introduces new error possibilities.

SomberNight:
I think this is a better solution than https://github.com/spesmilo/electrum/pull/3873 because of being a lot simpler. But as @bauerj said, it would be best if we didn't have to use our own binary rather there would be an "official" one.

SomberNight:
This workaround was removed in https://github.com/spesmilo/electrum/commit/6ab1ba52b05e0cd454ac0a513fc3cd44653a31f8#diff-76154fa219ee4d9785f711a5cf36b462
It's no longer needed.

PR #3893, opened 2018-02-12
Prevent crash after pressing "spend max"

Rav3nPL:
Error log w/o this change
```
  File "/home/rav3n/electrum/gui/qt/main_window.py", line 1232, in spend_max
    self.do_update_fee()
  File "/home/rav3n/electrum/gui/qt/main_window.py", line 1293, in do_update_fee
    size = tx.estimated_size()
  File "/home/rav3n/electrum/lib/util.py", line 226, in <lambda>
    return lambda *args, **kw_args: do_profile(func, args, kw_args)
  File "/home/rav3n/electrum/lib/util.py", line 222, in do_profile
    o = func(*args, **kw_args)
  File "/home/rav3n/electrum/lib/transaction.py", line 866, in estimated_size
    weight = self.estimated_weight()
  File "/home/rav3n/electrum/lib/transaction.py", line 914, in estimated_weight
    total_tx_size = self.estimated_total_size()
  File "/home/rav3n/electrum/lib/transaction.py", line 896, in estimated_total_size
    return len(self.serialize(True)) // 2 if not self.is_complete() or self.raw is None else len(self.raw) // 2  # ASCII hex string
  File "/home/rav3n/electrum/lib/transaction.py", line 810, in serialize
    nLocktime = int_to_hex(self.locktime, 4)
  File "/home/rav3n/electrum/lib/bitcoin.py", line 212, in int_to_hex
    return rev_hex(s)
  File "/home/rav3n/electrum/lib/bitcoin.py", line 205, in rev_hex
    return bh2u(bfh(s)[::-1])
ValueError: non-hexadecimal number found in fromhex() arg at position 6
Aborted
```

### DIFF
--- a/lib/bitcoin.py
+++ b/lib/bitcoin.py
@@ -208,6 +208,7 @@ def rev_hex(s):
 def int_to_hex(i, length=1):
     assert isinstance(i, int)
     s = hex(i)[2:].rstrip('L')
+    s = s.lstrip('0x')
     s = "0"*(2*length - len(s)) + s
     return rev_hex(s)
 


lib/bitcoin.py
```diff
@@ -208,6 +208,7 @@ def rev_hex(s):
 def int_to_hex(i, length=1):
     assert isinstance(i, int)
     s = hex(i)[2:].rstrip('L')
+    s = s.lstrip('0x')
```
SomberNight: `hex(i)[2:]` is trying to do the same thing above. what is `i` in your malformed case?

SomberNight:
What commit were you on when producing the crash?

Rav3nPL:
On commit https://github.com/akcelerator/electrum/commit/0dfaf9b97013ce9221bc8589ad806e52643e727d
Trying to reproduce it now, but can not...
Added some logging and it is cutting 0x properly w/o my line.
Maybe it somehow get negative value?

SomberNight:
Note: You can get an almost identical trace to appear in the log on current master (but the exception is caught) if you set `Amount` to an absurdly large number.

SomberNight:
Can't reproduce and present change would probably not fix it anyway.

Issue #3884, opened 2018-02-11
Add support for Importing Testnet master keys (tpub/tprv)


vzelenko:
There is no (apparent to me) way to restore a testnet wallet from exported public or private key.

I am able to "import" a Testnet deterministic wallet by importing "word phrase", which is a sufficient workaround for me.  I can also import a non-deterministic wallet using Private/Public key.  

Looking at the "keystore.py" code, Xpub logic does not account for testnet types...

If no one has the bandwidth to make the changes, perhaps expanding the docs with instructions on how to import the Testnet wallets would be helpful.

ecdsa:
yes, we need to switch back to t,u,v letters for testnet.

SomberNight:
Related: https://github.com/spesmilo/electrum/issues/3539

Issue #3539, opened 2017-12-17
Electrum standard wallet on testnet give a xpub (3.0.3)


NicolasDorier:
If you create a wallet on testnet with 3.0.3, the master seed is a xpub instead of tpub.

ecdsa:
This is expected, see the 3.00 release notes.

I decided to use the bip32 version bits to encode the script type, and not the network, because there are potentially an infinity of networks and altcoins, so that does not scale. however, it seems that some wallets want to use letters {t,u,v} for testnet. I am open to such a change.

SomberNight:
[Samourai-Wallet](https://github.com/Samourai-Wallet/bitcoinj/commit/52f03cfb134728b854c3946d9f764ccb3407fb94) seems to be using t/u/v for testnet.

Besides them, ypub/zpub has also been adopted by at least [blockonomics.co](https://blog.blockonomics.co/a-segwit-blockexplorer-47cd516dd8c5), [Trezor](https://blog.trezor.io/explore-the-new-send-dialogue-in-trezor-wallet-feae1253443a), and your [BTCPay](https://github.com/btcpayserver/btcpayserver/commit/24ce325e3117a5f9791672ca9940f11e98ed4baf).

I guess it would be nice to reach a consensus whether to use separate version bytes for testnet. I guess it could be {t, u, U, v, V} in addition to [{x, y, Y, z, Z}](https://github.com/spesmilo/electrum/blob/3851e78454f559ec4368a4b04dd61c2b9a1f9dc1/lib/bitcoin.py#L56-L69).

shivaenigma:
I also prefer different prefixes for testnet. Few reasons:

- Visual identification: I have multiple xpubs, lying around, I don't want to keep another table tracking if this is tesnet/mainet
- Users may try to recover/import testnet xprv into mainnet (or vice versa) and not find the funds
- Sharing of xpub/xprv account with merchants to receive recurring funds or other such workflows. Incorrect derivation can lead to funds going to the wrong address

If we are worrying about running out of letters, we can go back from _t_ :smirk: 
_{t,s,r,...}_

ecdsa:
let me add @SamouraiDev to this conversation

SamouraiDev:
For our internal TestNet version of Samourai, we are using 0x044A5262 for 'upub' and 0x044A4E28 for 'uprv'.

We display 'ypub' in Samourai and accept them in Sentinel (from Samourai, Trezor & Ledger).

[...]

PR #3898, opened 2018-02-13
bip32 version numbers (xpub headers): use t/u/U/v/V for testnet

SomberNight:
Based on https://github.com/spesmilo/electrum/issues/3539, it seems some people want (or use already) separate version bytes for testnet (mostly upub; besides tpub obviously). No one had a strong opinion against this; hence to try to be compatible with other wallets/services, we too can distinguish networks.

This PR changes bip32 extended key version bytes for testnet to use `{t,u,U,v,V}|{pub,prv}`.

The constants for `u` and `v` are what [Samourai-Wallet](https://github.com/Samourai-Wallet/bitcoinj/commit/52f03cfb134728b854c3946d9f764ccb3407fb94) and [iancoleman's website](https://github.com/iancoleman/bip39/blob/fcc7d121097189600a14d5f8247735cc88a9dbb2/src/js/segwit-parameters.js#L38) uses (only `u` is defined on the latter).

I could not find constants for `U` and `V`, so I calculated them myself.

Closes: https://github.com/spesmilo/electrum/issues/3884

### DIFF
--- a/lib/bitcoin.py
+++ b/lib/bitcoin.py
@@ -47,28 +47,6 @@ def read_json(filename, default):
     return r
 
 
-
-
-# Version numbers for BIP32 extended keys
-# standard: xprv, xpub
-# segwit in p2sh: yprv, ypub
-# native segwit: zprv, zpub
-XPRV_HEADERS = {
-    'standard': 0x0488ade4,
-    'p2wpkh-p2sh': 0x049d7878,
-    'p2wsh-p2sh': 0x295b005,
-    'p2wpkh': 0x4b2430c,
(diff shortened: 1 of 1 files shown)


SomberNight:
Note that if we don't want to break existing (Electrum) testnet wallet files, we could add a new storage version and corresponding version upgrade.
Maybe this is not worth it though; this is testnet afterall. Users can recreate wallets from seed words; or advanced users can change the version bytes and recalc the checksum.

ecdsa:
it's fine to break testnet :)

Issue #3904, opened 2018-02-14
Importing contacts fails if source wallet is fully encrypted


AbdussamadA:
So first of all the procedure for importing contacts is a little unintuitive. You can't export contacts only import them and the way to import them apparently is by selecting the wallet file you want to import from. If that wallet has full wallet encryption then importing fails silently. Even when running electrum with the -v option there are no error messages displayed. def import_file in contacts.py [just returns](https://github.com/spesmilo/electrum/blob/master/lib/contacts.py#L55) when the json in the file doesn't validate. Ideally it should prompt for the wallet password, decrypt it and then import the contacts.

SomberNight:
I can offer https://github.com/spesmilo/electrum/commit/909c063eb1922d0a7e4b61d5538cb5b34edbf1e7 for now. The actual implementation of importing from an encrypted wallet is up for grabs.

PR #3910, opened 2018-02-15
Added exporting contacts

AbdussamadA:
Export contacts just like labels

### DIFF
--- a/gui/qt/contact_list.py
+++ b/gui/qt/contact_list.py
@@ -54,10 +54,10 @@ class ContactList(MyTreeWidget):
         self.parent.set_contact(item.text(0), item.text(1))
 
     def import_contacts(self):
-        import_meta_gui(self.parent, 'contacts', self.parent.contacts.import_file, self.on_update)
+        import_meta_gui(self.parent, _('contacts'), self.parent.contacts.import_file, self.on_update)
 
     def export_contacts(self):
-        export_meta_gui(self.parent, 'contacts', self.parent.contacts.export_file)
+        export_meta_gui(self.parent, _('contacts'), self.parent.contacts.export_file)
 
     def create_menu(self, position):
         menu = QMenu()
--- a/gui/qt/contact_list.py
+++ b/gui/qt/contact_list.py
@@ -23,6 +23,8 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import webbrowser
+import os
+import json
 
 from electrum.i18n import _
 from electrum.bitcoin import is_address
@@ -33,7 +35,7 @@ from PyQt5.QtCore import *
 from PyQt5.QtWidgets import (
     QAbstractItemView, QFileDialog, QMenu, QTreeWidgetItem)
 from .util import MyTreeWidget
-import os, json
+
 
 class ContactList(MyTreeWidget):
     filter_columns = [0, 1]  # Key, Value
@@ -51,7 +53,7 @@ class ContactList(MyTreeWidget):
         if column == 0:  # Remove old contact if renamed
             self.parent.contacts.pop(prior)
         self.parent.set_contact(item.text(0), item.text(1))
-       
+
     def import_contacts(self):
         filename = self.parent.getOpenFileName(_("Open contacts file"), "*.json")
         if not filename:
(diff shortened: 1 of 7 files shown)


AbdussamadA:
this is for #3904

gui/qt/contact_list.py
```diff
@@ -33,7 +33,7 @@
 from PyQt5.QtWidgets import (
     QAbstractItemView, QFileDialog, QMenu, QTreeWidgetItem)
 from .util import MyTreeWidget
-
+import os, json
```
bauerj: Please follow [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) for the import order:

> Imports should be grouped in the following order:
>
> 1. standard library imports
> 2. related third party imports
> 3. local application/library specific imports

So this should be moved up to `import webbrowser`.

AbdussamadA: done

lib/contacts.py
```diff
@@ -63,6 +64,14 @@ def import_file(self, path):
         self.update(d)
         self.save()
 
+    def export_file(self, fileName):
+        try:
+            with open(fileName, 'w+') as f:
+                json.dump(self, f, indent=4, sort_keys=True)
+        except (IOError, os.error) as reason:
+            traceback.print_exc(file=sys.stderr)
+            raise reason
```
SomberNight: Not sure how much sense it makes to catch these exception and just re-raise them like this, here.
Could you maybe do it similar to `import_file`? Define `FileExportFailed` and raise that here.

lib/contacts.py
```diff
@@ -63,6 +64,14 @@ def import_file(self, path):
         self.update(d)
         self.save()
 
+    def export_file(self, fileName):
+        try:
+            with open(fileName, 'w+') as f:
+                json.dump(self, f, indent=4, sort_keys=True)
```
SomberNight: btw you can add a "success" pop-up as you originally did, I think that's a good idea. A clean way to do it is with a `try-except-else`, in the `else` branch. If you do, add that to that importing too. :P

gui/qt/contact_list.py
```diff
@@ -63,12 +64,22 @@ def import_contacts(self):
...
         self.on_update()
 
+    def export_contacts(self):
+        filename = self.parent.getSaveFileName(_("Select file to save your contacts"), 'electrum_contacts.json', "*.json")
+        if  not filename:
+            return
+        try:
+            self.parent.contacts.export_file(filename)
+        except (IOError, os.error) as reason:
+            self.parent.show_critical(_("Electrum was unable to export your contacts.") + "\n" + str(reason))
```
SomberNight: btw you can add a "success" pop-up as you originally did, I think that's a good idea. A clean way to do it is with a `try-except-else`, in the `else` branch. If you do, add that to that importing too. :P
(deleted the other comment regarding this, it was at the wrong pos.)

lib/contacts.py
```diff
@@ -54,15 +55,23 @@ def import_file(self, path):
         try:
             with open(path, 'r') as f:
                 d = self._validate(json.loads(f.read()))
-        except json.decoder.JSONDecodeError:
+        except ValueError:
```
SomberNight: What's the reason for this change?

lib/contacts.py
```diff
@@ -54,15 +55,23 @@ def import_file(self, path):
...
             traceback.print_exc(file=sys.stderr)
-            raise FileImportFailedEncrypted()
+            raise FileImportFailedInvalidJSON()
```
SomberNight: You are right, the file might not be encrypted, all we know is that JSON decoding failed. `FileImportFailedEncrypted` was just a guess. However I don't think we should start introducing all kinds of  `File*Failed*` exceptions. 
How about keeping only `FileImportFailed` and `FileExportFailed`, and extending `FileImportFailed` with a message the same way you have done for `FileExportFailed`?

lib/contacts.py
AbdussamadA: https://stackoverflow.com/a/44714607/672396

lib/contacts.py
AbdussamadA: ok will do. also shall i do this same treatment for invoices? invoices are payment requests right?

AbdussamadA: yes they are the same as payment requests. i'll do invoice exporting as well.

SomberNight: Invoices are what you should pay; i.e. the stuff at the bottom of the `Send` tab (when visible).

lib/contacts.py
SomberNight: Ok. Can you add a comment that reflects this? like `# backwards compatibility for JSONDecodeError ` or similar

PR #3911, opened 2018-02-15
Build: Pyinstaller installation

bauerj:
This changes the macOS build script to use our patched pyinstaller. Also, its installation in the Wine build scripts is simplified.


SomberNight:
Looks good.

But; so that reproducible builds would stay reproducible some time after a release, maybe we should pin the exact commit of pyinstaller?

PR #3922, opened 2018-02-18
network: test genesis hash of interface/server

SomberNight:
testnet_servers
atm testnet servers are reporting some mainnet servers among their peers.
Testnet clients connect to the mainnet servers too and then don't disconnect.

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -714,6 +714,8 @@ class Network(util.DaemonThread):
         interface.mode = 'default'
         interface.request = None
         self.interfaces[server] = interface
+        # ask for genesis hash first and disconnect if unexpected
+        self.request_header(interface, 0)
         self.queue_request('blockchain.headers.subscribe', [], interface)
         if server == self.default_server:
             self.switch_to_interface(server)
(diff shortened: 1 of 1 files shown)


ecdsa:
I cannot reproduce the issue. do you still have it?

SomberNight:
If it happens once, `recent_servers` gets "corrupted" and it will happen all the time; so yes, it still happens with that datadir :P

ecdsa:
earlier versions of the code requested the genesis block
I removed it because it creates an unnecessary request.
The client should be able to figure out it is not on the right chain and disconnect.

In your screenshot, if your client is connected to mainnet servers, why are they not in the list?

SomberNight:
> if your client is connected to mainnet servers, why are they not in the list?

On testnet currently:
```
>> network.get_interfaces()
['kirsche.emzy.de:50002:s', 'electrum.poorcoding.com:50002:s', 'testnet.qtornado.com:51002:s', 'node.arihanc.com:50002:s', 'E-X.not.fyi:50002:s', 'testnet.hsmiths.com:53012:s', '88.198.43.231:50002:s', 'electrumx.kekku.li:51002:s', 'electrum.achow101.com:50002:s', 'tardis.bauerj.eu:50002:s']

>> [(k, [interface.server for interface in v]) for (k,v) in network.get_blockchains().items()]
[(0, ['testnet.qtornado.com:51002:s', 'testnet.hsmiths.com:53012:s', 'electrumx.kekku.li:51002:s'])]
```

The list is populated with interfaces that are associated with `network.get_blockchains()`; it seems only testnet servers are such.
https://github.com/spesmilo/electrum/blob/b7b592fd6eb59049896ec644bdfb86cc7bf1685b/gui/qt/network_dialog.py#L103-L119
The total count displayed is `len(self.network.get_interfaces())`, which includes both testnet and mainnet servers.
https://github.com/spesmilo/electrum/blob/b7b592fd6eb59049896ec644bdfb86cc7bf1685b/gui/qt/network_dialog.py#L353-L355

ecdsa:
yes, ok, that's normal.
the issue is that mainnet servers are not disconnected.

ecdsa:
a wrong parameter was passed to connection_down, see my commit.

It might make sense for servers to send the genesis block hash during the handshake (server.version), so that clients can disconnect faster from altcoin servers. However, I do not want to add an extra genesis block request after the handshake, because it would waste time, such a request will be useless 99% of the time, and also because asking for the genesis block is not robust to attacks, and ineffective in the case of bitcoin forks.

PR #3923, opened 2018-02-19
local tx: restructure exception handling wrt wallet.add_transaction and QT

SomberNight:


ecdsa:
I think it would be cleaner if save_transaction_into_wallet was a method of main_window

SomberNight:
Ok, moved `save_transaction_into_wallet` from `util` to `main_window`.
Rebased and squashed.

PR #3967, opened 2018-02-24
qt save_payment_request catch exception

### LINKED ISSUE #3965: PermissionError: [Errno 13] Permission denied: '/var/www/html/r/req/d'
Crash Report
============

Traceback
--------------

```Python traceback
  File "/home/user/wspace/electrum/gui/qt/main_window.py", line 907, in save_payment_request
    self.wallet.add_payment_request(req, self.config)
  File "/home/user/wspace/electrum/lib/wallet.py", line 1563, in add_payment_request
    os.makedirs(path)
  File "/usr/lib/python3.5/os.py", line 231, in makedirs
    makedirs(head, mode, exist_ok)
  File "/usr/lib/python3.5/os.py", line 231, in makedirs
    makedirs(head, mode, exist_ok)
  File "/usr/lib/python3.5/os.py", line 241, in makedirs
    mkdir(name, mode)

PermissionError: [Errno 13] Permission denied: '/var/www/html/r/req/d'
```

Reporter
------------

| Electrum Version  | Operating System  | Wallet Type  | Locale |
|---|---|---|---|
| 3.1  | Linux-4.9.0-3-amd64-x86_64-with-debian-9.3 | standard | en_GB |

Additional Information
------------------------

SomberNight:
fix #3965 

It's not exactly clear how an exception such as #3965 should be handled...
By the time the exception is raised, the unsigned pr is already in storage; only writing to disk fails.
So maybe we could still sign it? Anyway, in this PR we don't; and we also don't revert what has already been done.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -904,11 +904,17 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         i = self.expires_combo.currentIndex()
         expiration = list(map(lambda x: x[1], expiration_values))[i]
         req = self.wallet.make_payment_request(addr, amount, message, expiration)
-        self.wallet.add_payment_request(req, self.config)
-        self.sign_payment_request(addr)
-        self.request_list.update()
-        self.address_list.update()
-        self.save_request_button.setEnabled(False)
+        try:
+            self.wallet.add_payment_request(req, self.config)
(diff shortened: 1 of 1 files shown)


ecdsa:
note: at some point we should remove the dns signing feature from the code, nobody uses it

PR #3968, opened 2018-02-24
show headers download in GUI

### LINKED ISSUE #3966: when electrum is run for the first time without headers, display in the GUI that download activity is undergo
When starting Electrum for the first time, without any blockchain headers, it needs to download them. This is not a huge amount of data but on slow connections or behind custom setups (like slow proxies, Tor network, etc.) it can take quite a few minutes. During this time, there is nothing shown in the GUI to indicate it. 

For example, I just started on Tails an Electrum 3.0.6 instance from scratch, with a brand new wallet. It immediately shown 0 mBTC in bottom left side, like this was the balance in fully synced mode. In bottom right side there is a blue icon at network settings (ok, since i am using localhost:9050 socks5 proxy) and only if I click on that to further open the network dialog i can see the blockchain: x blocks increasing gradually as more headers are downloaded.  The background activity (headers download) works good.

Users who don't know how this works might have a bad experience. I think it would be useful to display something that will let the user know that something is happening and that electrum needs more time to catch-up with blockchain headers.

Related is #2650, might check if we find useful implementing something from there as well.

SomberNight:
Closes #3966

`< -1` instead of `< 0` as
- this should mostly eliminate the possibility of displaying the message when we are synced but a new block was just mined
- it shouldn't matter much anyway

### DIFF
--- a/gui/kivy/main_window.py
+++ b/gui/kivy/main_window.py
@@ -630,6 +630,8 @@ class ElectrumWindow(App):
             server_lag = self.network.get_local_height() - server_height
             if not self.wallet.up_to_date or server_height == 0:
                 status = _("Synchronizing...")
+            elif server_lag < -1:
+                status = _("Headers download ({} left)").format(-1 * server_lag)
             elif server_lag > 1:
                 status = _("Server lagging ({} blocks)").format(server_lag)
             else:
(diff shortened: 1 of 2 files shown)


ecdsa:
I am not sure if it is useful to display that.
The user who requested this feature is probably using an older version, before we introduced checkpoints.
In addition, `server_lag` is not a good predictor of whether we are downloading headers, because chunks may be requested in arbitrary order.

PR #3970, opened 2018-02-25
remove unneeded 'pass' statement

romanz:


romanz:
The error above seems to be unrelated to the PR - it also happens on the original master branch:
https://travis-ci.org/romanz/electrum/builds/345947651

PR #3972, opened 2018-02-26
Update build-wine's README.md

Lastrellik:
While trying to build Windows executables I ran into quite a few problems related to Wine that could have been helped with this documentation. I imagine this could help someone in the future if they have issues similar to mine.


ecdsa:
note: I ran into the same issue with 7zip

bauerj:
Thanks!

I removed the part about fixing a broken Wine installation since I don't think that should happen to a lot of people.

PR #3984, opened 2018-02-27
Create file caching system

Lastrellik:
After having problems with Sourceforge being down/slow a few days ago I decided to modify this script to allow for caching of the files that would have otherwise been downloaded every time this script runs. This way we can still generate binaries even if Sourceforge is down.

### DIFF
--- a/contrib/build-wine/prepare-wine.sh
+++ b/contrib/build-wine/prepare-wine.sh
@@ -38,7 +38,7 @@ verify_signature() {
 verify_hash() {
     local file=$1 expected_hash=$2
     actual_hash=$(sha256sum $file | awk '{print $1}')
-    if [ "$actual_hash" = "$expected_hash" ]; then
+    if [ "$actual_hash" == "$expected_hash" ]; then
         return 0
     else
         echo "$file $actual_hash (unexpected hash)" >&2
--- a/contrib/build-wine/prepare-wine.sh
+++ b/contrib/build-wine/prepare-wine.sh
@@ -1,13 +1,16 @@
 #!/bin/bash
 
 # Please update these carefully, some versions won't work under Wine
-NSIS_URL=https://prdownloads.sourceforge.net/nsis/nsis-3.02.1-setup.exe?download
+NSIS_FILENAME=nsis-3.02.1-setup.exe
+NSIS_URL=https://prdownloads.sourceforge.net/nsis/$NSIS_FILENAME?download
 NSIS_SHA256=736c9062a02e297e335f82252e648a883171c98e0d5120439f538c81d429552e
 
-ZBAR_URL=https://sourceforge.net/projects/zbarw/files/zbarw-20121031-setup.exe/download
+ZBAR_FILENAME=zbarw-20121031-setup.exe
+ZBAR_URL=https://sourceforge.net/projects/zbarw/files/$ZBAR_FILENAME/download
 ZBAR_SHA256=177e32b272fa76528a3af486b74e9cb356707be1c5ace4ed3fcee9723e2c2c02
 
-LIBUSB_URL=https://prdownloads.sourceforge.net/project/libusb/libusb-1.0/libusb-1.0.21/libusb-1.0.21.7z?download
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks. no need to replace == with = in the test

Lastrellik:
That's what I thought, and I'm sure there's something about it I don't understand, but the == was returning false when the provided checksums were actually equal. It might be worth checking out from your end.

contrib/build-wine/prepare-wine.sh
```diff
@@ -91,22 +100,22 @@ $PYTHON -m pip install -r ../../deterministic-build/requirements-binaries.txt
...
-wget -q -O nsis.exe "$NSIS_URL"
-verify_hash nsis.exe $NSIS_SHA256
-wine nsis.exe /S
+download_if_not_exist $NSIS_FILENAME "$NSIS_URL"
+verify_hash $NSIS_FILENAME "$NSIS_SHA256"
+wine "$PWD/$NSIS_FILENAME" /S
+
+download_if_not_exist $LIBUSB_FILENAME "$LIBUSB_URL"
+verify_hash $LIBUSB_FILENAME "$LIBUSB_SHA256"
+7z x -olibusb $LIBUSB_FILENAME -aos
```
bauerj: What is `-aos`?

contrib/build-wine/prepare-wine.sh
```diff
@@ -70,8 +79,8 @@ KEYSERVER_PYTHON_DEV="hkp://keys.gnupg.net"
 gpg --no-default-keyring --keyring $KEYRING_PYTHON_DEV --keyserver $KEYSERVER_PYTHON_DEV --recv-keys $KEYLIST_PYTHON_DEV
 for msifile in core dev exe lib pip tools; do
     echo "Installing $msifile..."
-    wget "https://www.python.org/ftp/python/$PYTHON_VERSION/win32/${msifile}.msi"
-    wget "https://www.python.org/ftp/python/$PYTHON_VERSION/win32/${msifile}.msi.asc"
+    wget -nc "https://www.python.org/ftp/python/$PYTHON_VERSION/win32/${msifile}.msi"
```
bauerj: Why not `download_if_not_exist` too?

contrib/build-wine/prepare-wine.sh
Lastrellik: The -aos tells 7Zip to skip over existing files.

contrib/build-wine/prepare-wine.sh
Lastrellik: The -nc prevents wget from downloading if the file already exists, but cannot be used with -O for the output file or else it will download every time, making the caching feature useless. I didn't run into that problem using -nc with these files because we didn't need to specify an output file.

ecdsa:
why did you close this pull request?

Lastrellik:
Oh! My mistake. I must have misclicked. >.< I didn't mean to close it

Lastrellik:
Actually what happened was I accidentally reset my commit while trying to sync my fork with the main repo >.< This should have fixed it. My mistake.

ecdsa:
can you try == with double brackets? see https://stackoverflow.com/questions/2237080/how-to-compare-strings-in-bash

Lastrellik:
@ecdsa that also resulted in an "unexpected hash."
```
prepare-wine.sh: 41: prepare-wine.sh: [[: not found
zbarw-20121031-setup.exe 177e32b272fa76528a3af486b74e9cb356707be1c5ace4ed3fcee9723e2c2c02 (unexpected hash)
```
I was nervous about the ```=``` when I first tried it because of its different meaning in other languages, but I found that it has the desirable effect even when the hashes are not equal.

bauerj:
`[[` is a bash-builtin. Are you sure that /bin/bash is actually bash?

Lastrellik:
@bauerj you found my problem! I've been using ```sh``` this whole time. Thanks for pointing that out and I went ahead and changed that in my latest commit.

PR #3985, opened 2018-02-27
qt addresses tab: new column for type

SomberNight:
addr_list2

I also set the formatting of the `Balance` column to be decimal-separator-aligned (both `Addresses` and `Coins` tab).

### DIFF
--- a/gui/qt/address_list.py
+++ b/gui/qt/address_list.py
@@ -24,18 +24,19 @@
 # SOFTWARE.
 import webbrowser
 
-from .util import *
 from electrum.i18n import _
 from electrum.util import block_explorer_URL
 from electrum.plugins import run_hook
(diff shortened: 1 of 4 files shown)


SomberNight:
Note that there is an interesting detail in this image: address `tb1qjx423y0j4a8aunsw43w4x2kkd4lnue4uw6fpem` is funded but has `0` transactions related to it.
It was funded by a `local` transaction, and the column shows the number of txns in the server-provided history:
https://github.com/spesmilo/electrum/blob/8fe315beabc94e18de6a3bf4ea0beb409fd3d3e6/gui/qt/address_list.py#L87

ecdsa:
thanks for the PR.
the Tx column is not very useful, I think we could replace it with a 'status' that would match the second filter (used, funded, etc)

Issue #3996, opened 2018-03-02
Cannot access Trezor Model T with passphrases enabled


mocmocamoc:
Using the latest electrum and trezorlib code, the Trezor Model T works in Electrum as long as passphrases are not enabled.

If passphrases are enabled, then the device asks you to choose whether to enter the passphrase on the device or the "host" (i.e. Electrum).
-If you choose to enter the passphrase on the device, the device screen goes blank and Electrum shows a prompt for a passphrase. When anything is entered Electrum gives an error: "Passphrase entered when it should not be"
-if you choose to enter the passphrase on the host. the device shows "Please type passphrase on connected host" and Electrum shows a prompt for a passphrase.  When anything is entered Electrum gives an error: "Got <class 'trezorlib.messages.PassphraseStateRequest.PassphraseStateRequest>, expected (<class ' trezorlib.messages.PublicKey.PublicKey'>,)"

I am able to use trezorlib directly to get addresses from the device while passphrases are enabled, in which case passphrases are correctly accepted on either the device or host as per the user's choice.

PR #3997, opened 2018-03-02
trezorT: fix/implement passphrases

SomberNight:
fix https://github.com/spesmilo/electrum/issues/3996

Note that you need python-trezor master for this, 0.9.0 is not enough.
I think probably https://github.com/trezor/python-trezor/commit/bccd61cb23a5259e02d6606ac320093d6637d203 is needed specifically; but in any case I've been testing on https://github.com/trezor/python-trezor/commit/e37e9bfebda2c2c5c6bc2bf60aafe6dc65cc6546


prusnak:
I think trezor 0.9.1 from PyPI should be enough. cc @matejcik

Issue #4004, opened 2018-03-03
Digital Bitbox does not work on testnet


SomberNight:
dbb needs xpub <-> tpub conversions to work on testnet..

https://github.com/spesmilo/electrum/blob/7816edc3428f7148de1b2c4b567408b68c1d25cb/plugins/digitalbitbox/digitalbitbox.py#L89-L100

PR #4008, opened 2018-03-03
refactor network constants

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/4004

### DIFF
--- a/electrum
+++ b/electrum
@@ -89,6 +89,7 @@ if is_local or is_android:
 
 
 from electrum import bitcoin, util
+from electrum import constants
 from electrum import SimpleConfig, Network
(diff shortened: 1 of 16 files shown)


ecdsa:
I agree with moving constants to a separate module, but I think we do not need an extra NetworkConstants class in it; references will be simplified if net is an attribute of the module itself. Also, you should name it constants.py; we already have a network.py file, network_constants.py is going to mess with autocompletions.

SomberNight:
All right, I've changed it a bit. :)

We should make sure not to do something like this though:
```
from electrum.constants import *
net = BitcoinTestnet
```

Issue #1116, opened 2015-03-31
display suggestions when typing / submitting a recovery seed


I3ck:
A friend of mine wrote down his seed by hand and tried to recover his wallet.  
He couldn't hit next when doing so.  
After some research I found the word list used when creating the seeds and sent it to him.  
He then was able to figure out his typo and recover the wallet.  
# suggestions

(it may not make sense to implement all of them at once, but a combination)
- add a proper warning that the seed may be incorrect instead of just disabling the next button  
- auto-suggest the valid words when entering the seed
- don't disable the next button and then prompt the user with a corrected suggestion for the seed

SomberNight:
Let me detail one of the suggestions you mention, for people who might implement it:

In the QT GUI, when restoring a wallet from seed words, display word suggestions. That is, given the current partial word, offer suggestions of possible words to the user.

The suggestions MUST be optional, i.e. the user must have the option to ignore them completely; as seeds can contain any string (e.g. `9dk` is a valid segwit-type Electrum seed). (but the seeds the current version of the software generates only contain words from the BIP39 English word list)

Note that the kivy GUI (that is used in the Android app) has this feature, so one might want to have a look at the implementation there.

I would suggest sticking to the English dictionaries, probably the union of the BIP39 English word list and the old Electrum mnemonic word list.

Example from the kivy GUI:
![seed_word_suggestions](https://user-images.githubusercontent.com/29142493/36751022-43cc7f7a-1bff-11e8-814c-e38aba4c9ba6.PNG)

Note that the current partial word is `bloo`, and the suggestions are `bloody`, `bloom`, `blood`.

SomberNight:
This was done in https://github.com/spesmilo/electrum/pull/4033

PR #4019, opened 2018-03-04
Add suggestions when restoring from seed

Lastrellik:
In reference to [#1116](https://github.com/spesmilo/electrum/issues/1116), this feature displays a list of suggestions from the seed wordlist file according to which word the user is entering in at the time. I added the new file to the lib folder because I would like for the seed restoration auto suggestion in the kivy code to utilize this feature as well. This may come in a later pull request depending on the feedback I get from this one.

### DIFF
--- a/gui/qt/seed_dialog.py
+++ b/gui/qt/seed_dialog.py
@@ -27,6 +27,7 @@ from PyQt5.QtGui import *
 from PyQt5.QtCore import *
 from PyQt5.QtWidgets import *
 from electrum.i18n import _
+from electrum.seed_suggestion import SeedSuggestion
 
 from .util import *
 from .qrtextedit import ShowQRTextEdit, ScanQRTextEdit
@@ -100,6 +101,9 @@ class SeedLayout(QVBoxLayout):
             self.is_seed = is_seed
             self.saved_is_seed = self.is_seed
             self.seed_e.textChanged.connect(self.on_edit)
+            self.suggestion = SeedSuggestion("./lib/wordlist/english.txt")
+            self.suggestion_label = QLabel()
+            self.addWidget(self.suggestion_label)
         self.seed_e.setMaximumHeight(75)
         hbox = QHBoxLayout()
         if icon:
@@ -135,7 +139,17 @@ class SeedLayout(QVBoxLayout):
         text = self.seed_e.text()
         return ' '.join(text.split())
 
+    def display_suggestions(self):
+        suggestion_char_length = 45
+        self.suggestion_label.setText("Suggestions: ")
+        if self.seed_e.text() == "" or self.seed_e.text()[-1] == ' ':
+            return
+        for word in self.suggestion.get_suggestions(self.get_seed().split(' ')[-1],
+                                                    suggestion_char_length):
+            self.suggestion_label.setText(self.suggestion_label.text() + word + " ")
+
     def on_edit(self):
+        self.display_suggestions()
         from electrum.bitcoin import seed_type
         s = self.get_seed()
         b = self.is_seed(s)
@@ -149,6 +163,7 @@ class SeedLayout(QVBoxLayout):
             label = 'BIP39' + ' (%s)'%status
         self.seed_type_label.setText(label)
         self.parent.next_button.setEnabled(b)
+
 
 
 class KeysLayout(QVBoxLayout):
--- a/lib/seed_suggestion.py
+++ b/lib/seed_suggestion.py
@@ -1,0 +1,45 @@
+# Electrum - lightweight Bitcoin client
+# Copyright (C) 2012 thomasv@ecdsa.org
+#
+# Permission is hereby granted, free of charge, to any person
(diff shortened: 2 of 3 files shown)


bauerj:
python_2018-03-04_21-57-22
Nothing is typed in, nothing is suggested. ✓

---- 

python_2018-03-04_21-57-09
Typing a word shows correct suggestions. ✓

----

python_2018-03-04_21-57-29
While nothing is printed before starting to type, as soon as something is typed, "Suggestions" is written above the text edit. This should be removed if there are no suggestions.

----

python_2018-03-04_22-40-57
Typing something that doesn't start a correct seed word, `[none]` is printed. Instead, a warning should be printed.

----

python_2018-03-04_22-44-36
When enabling BIP39, the same suggestions are printed. Since we don't know what wordlist was used to generate the seed in the first place, we should not use our wordlist to make suggestions.

gui/qt/seed_dialog.py
```diff
@@ -100,6 +101,9 @@ def __init__(self, seed=None, title=None, icon=True, msg=None, options=None, is_
             self.is_seed = is_seed
             self.saved_is_seed = self.is_seed
             self.seed_e.textChanged.connect(self.on_edit)
+            self.suggestion = SeedSuggestion("./lib/wordlist/english.txt")
```
bauerj: Not sure if this will work in the binaries.

`lib` will be installed as `electrum` so the path might be different.

bauerj:
Also: Please ignore the failing check from Travis CI. Currently, all our builds fail because Sourceforge is offline.

SomberNight:
@bauerj 
> When enabling BIP39, the same suggestions are printed. Since we don't know what wordlist was used to generate the seed in the first place, we should not use our wordlist to make suggestions.

I don't know... maybe. Our wordlist IS the bip39 English wordlist. Of course you are right that the user might have a seed using a different wordlist (most likely using a different language), but I think we could consider this a minority case, and display English suggestions. After all, these are just suggestions that the user can ignore. One thing that we can do differently in the bip39 case though is only consider the new wordlist and not the old one.

SomberNight:
@bauerj 
> `self.suggestion = SeedSuggestion("./lib/wordlist/english.txt")`
> Not sure if this will work in the binaries.

> ignore the failing check from Travis CI. Currently, all our builds fail because Sourceforge is offline.

Actually it is failing due to the path :P

```

self = <lib.seed_suggestion.SeedSuggestion object at 0x7f8f61ced390>
wordlist_file = '../wordlist/english.txt'
    def __init__(self, wordlist_file):
>       with open(wordlist_file) as wordlist_file:
E       FileNotFoundError: [Errno 2] No such file or directory: '../wordlist/english.txt'
lib/seed_suggestion.py:28: FileNotFoundError
```

bauerj:
Yes, I saw that after posting it 😅

bauerj:
> I don't know... maybe. Our wordlist IS the bip39 English wordlist. Of course you are right that the user might have a seed using a different wordlist (most likely using a different language), but I think we could consider this a minority case, and display English suggestions. After all, these are just suggestions that the user can ignore. One thing that we can do differently in the bip39 case though is only consider the new wordlist and not the old one.

Alright, maybe instead we should disable suggestions in case the user already typed a word that's not in the word list.

SomberNight:
> maybe instead we should disable suggestions in case the user already typed a word that's not in the word list

Oh. Good idea!

Lastrellik:
Thank you for your responses! 

@SomberNight I did spend some time looking at the kivy implementation when I wrote this and thought that a lot of what they're doing in that file can be taken care of in the new class that I created. The reason why I didn't yet is because I wanted to get some feedback first on this whole idea and I haven't installed anything related to kivy. I'm hoping we can get rid of all duplicated code with this library. 

Also, it shouldn't be too hard to implement more than one wordlist into the suggestion feature, although it would probably use a lot more memory. I can fix the relative path and the suggestion thing that @bauerj suggested. Is there anything else you guys would like to see in this feature?

ecdsa:
This is not very useful if you do not let the user use the suggestions.
Qt has a class for that, QCompleter. 
QCompleter is used in paytoedit.py, for example.

lib/seed_suggestion.py
```diff
@@ -0,0 +1,45 @@
+# Electrum - lightweight Bitcoin client
+# Copyright (C) 2012 thomasv@ecdsa.org
```
SomberNight: I suggest changing this to
```
# Copyright (C) 2018  The Electrum developers
```

SomberNight:
Hmm indeed QCompleter would be nice... Please take a look at that >.<

send_tab_qcompleter

Lastrellik:
Ahh.... Yeah that is much nicer... And easier >.< I'll see what I can come up with tomorrow. It shouldn't be too hard :) Thanks for the feedback.

ecdsa:
ok, closing this PR; please resubmit a new one

PR #4033, opened 2018-03-05
Text completer

Lastrellik:
After learning about the QCompleter tool from @ecdsa, I came up with a CompletionTextEdit class that has been utilized in both paytoedit.py and seed_dialog.py. This feature allows the user to optionally select from a list of words when restoring from seed in order to make the process a little easier and less error prone. 

Before merging this pull request, if someone could make sure that the send functionality in the main window still auto populates, that would be great :) I know that it doesn't crash, but I didn't have anyone in my address book that I could test it with :) I appreciate your patience.


gui/qt/paytoedit.py
```diff
@@ -23,14 +23,12 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from PyQt5.QtCore import *
 from PyQt5.QtGui import *
-from PyQt5.QtWidgets import QCompleter, QPlainTextEdit
-from .qrtextedit import ScanQRTextEdit
-
+from .completion_text_edit import CompletionTextEdit
```
bauerj: Please move this import down to the other imports from Electrum.

gui/qt/seed_dialog.py
```diff
@@ -131,6 +132,13 @@ def __init__(self, seed=None, title=None, icon=True, msg=None, options=None, is_
             self.seed_warning.setText(seed_warning_msg(seed))
         self.addWidget(self.seed_warning)
 
+    def initialize_completer(self):
+        english_path = os.path.join(os.path.dirname(__file__), "../../lib/wordlist/english.txt")
```
bauerj: Why don't you use a `Mnemonic` object?

    Mnemonic('en').wordlist

Lastrellik:
@bauerj that last commit should take care of the suggestions in your review. I have to be done for the day, but I can work on disabling the suggestions tomorrow, if you'd like.

Lastrellik:
I forgot to mention, this is in reference to [#1116](https://github.com/spesmilo/electrum/issues/1116)

gui/qt/seed_dialog.py
```diff
@@ -131,6 +133,13 @@ def __init__(self, seed=None, title=None, icon=True, msg=None, options=None, is_
             self.seed_warning.setText(seed_warning_msg(seed))
         self.addWidget(self.seed_warning)
 
+    def initialize_completer(self):
+        english_list = Mnemonic('en').wordlist
+        old_list = electrum.old_mnemonic.words
+        wordlist = english_list + list(set(old_list) - set(english_list))
```
SomberNight: maybe this is easier to read:
`wordlist = list(set(old_list) | set(english_list))`

SomberNight: but if you are doing that to have the words of the new list appear before the words of the old, then disregard

gui/qt/completion_text_edit.py
```diff
@@ -0,0 +1,108 @@
...
+        cr.setWidth(self.completer.popup().sizeHintForColumn(0) + self.completer.popup().verticalScrollBar().sizeHint().width())
+        self.completer.complete(cr)
+
+    def is_special_key(self, e):
+        if self.completer != None and self.completer.popup().isVisible():
+            if e.key() in [Qt.Key_Enter, Qt.Key_Return]:
+                return True
+        if e.key() in [Qt.Key_Tab]:
+            return True
+        if e.key() in [Qt.Key_Down, Qt.Key_Up] and not self.is_multiline():
```
SomberNight: `is_multiline` was defined in `PayToEdit` but not here

gui/qt/seed_dialog.py
Lastrellik: It was more to concatenate both lists without having duplicates. Because we're not doing any sort of binary search or tree data structure, the order of the list doesn't matter.

SomberNight: What does the order of the suggestions presented to the user depend on? Is that independent of the order in this list here?

gui/qt/paytoedit.py
```diff
@@ -177,9 +177,6 @@ def get_outputs(self, is_max):
     def lines(self):
         return self.toPlainText().split('\n')
 
-    def is_multiline(self):
```
SomberNight: this method is used in this class, around line 206 :P

Lastrellik:
@SomberNight thanks for your feedback! I went ahead and made those changes in my last commit.

gui/qt/paytoedit.py
Lastrellik: I saw that. My thought is that it does extend CompletionTextEdit, so that functionality should still be available. Unless I need to use ```parent.is_multiline()``` instead of ```self.is_multiline()``` on line 206

SomberNight: It should be fine; sorry, I was not paying enough attention. See EDIT above.

Lastrellik:
I should have been paying closer attention :) The is_multiline only returns a value that is more than one if the text contains ```\n```. This only applies to paytoedit.py. So in this last commit I moved ```is_multiline()``` back to paytoedit.py and I also added a feature where the suggestions are disabled if the user enters a word that is not in the wordlist, as requested by @bauerj

gui/qt/seed_dialog.py
```diff
@@ -160,6 +160,13 @@ def on_edit(self):
         self.seed_type_label.setText(label)
         self.parent.next_button.setEnabled(b)
 
+        # to account for bip39 seeds
+        for word in self.get_seed().split(" ")[:-1]:
+            if word not in self.wordlist:
+                self.seed_e.disable_suggestions()
+                break
+            else:
+                self.seed_e.enable_suggestions()
```
SomberNight: it might make more sense for the `else` to be one indentation layer "outer"; so that it accompanies the `for` loop instead of the `if`

Lastrellik:
@SomberNight thanks for the suggestion :) I went ahead and made that change with that last commit. I appreciate everyone's patience while I get familiar with Electrum and its code base :)

SomberNight:
If the current partial word is a valid word (present in the dict) and you press enter, it will duplicate the word.
Can you make the suggestions disappear and not do anything in that case?

<details>
 <summary>pics</summary>

seed_word_suggestions2
seed_word_suggestions3
</details>

Also, by the way, I think for the seed word case, it might make sense to also add a whitespace when the user auto-completes a word, so they can start entering the next word right away. (whitespaces are normalized out from seeds so don't worry about not adding a whitespace after the last word or stuff like that)

In fact, it might be easier to solve the first thing while solving the second:
maybe you can append a whitespace after every word in the dictionary or something like that;
or do it some other way :P

Lastrellik:
@SomberNight again, thank you for your suggestions! I think we're all good now. What do you think?

SomberNight:
Great; thanks!
@ecdsa I think this can get merged

Issue #3916, opened 2018-02-17
New wallet wizard loading page


bauerj:
While looking at #3914 I got the impression that the current dialog (especially with the new button there) looks a bit confusing on first sight:

![image](https://user-images.githubusercontent.com/598790/36345473-6fb8c818-142b-11e8-9587-41d00b0031d1.png)

I think the current design could be changed to increase the user experience and add a more natural option to create a new wallet:

______

![1](https://user-images.githubusercontent.com/598790/36345419-2e5c0eda-142a-11e8-8617-f451107584e4.png)
A list of recently opened wallets is shown. On the button of that list there are buttons to "extend" the list by either loading a wallet from file or creating a new one.

_____

![2](https://user-images.githubusercontent.com/598790/36345420-2e732a84-142a-11e8-906e-5cc08ab3be8c.png)
The user wants to open a wallet without encryption.

_____

![3](https://user-images.githubusercontent.com/598790/36345421-2e88add2-142a-11e8-9d97-a23d5d66c078.png)
After creating "New Wallet", the button disappears, just like in your example.

___

In my opinion, this gives the dialog a tidier appearance while still providing more information. The list of recent wallets should help people find the right wallet. The "Last opened" date can be obtained by reading the file modification as closing Electrum seems to save the wallet file.

What do you think? Is this better than what we have now?

SomberNight:
It looks good.

One thing: we want to make it unambiguous that the `load wallet` button is actually for choosing a different wallet, not to load the currently selected one... Do you think that is clear to the user?
Maybe it could be `change wallet` or `different wallet` or `load other wallet`?

bauerj:
Yes, "Load other wallet" sounds good.

SomberNight:
There was another user on IRC today who got stuck because he forgot the wallet password and could not figure out how to restore from seed.

fresheneesz:
I would recommend that we don't differentiate the "current wallet" from other wallets. I think we should treat all existing wallets the same (perhaps sorted by use). I also recommend we get rid of the "Next" and "Cancel" buttons, since buttons closer to the context would be more intuitive. And if you need to say "Click next to continue", you might as well just put the Next button there instead.

![electrumHomeView](https://user-images.githubusercontent.com/149531/71536090-89442d80-28c0-11ea-819b-3057eacd027f.PNG)

PR #3914, opened 2018-02-17
wizard: (qt) add dedicated button to create new wallet

SomberNight:
I find newbies often get stuck on this screen if they don't know their password and want to create a new wallet. Some even get stuck if they know their password but want to create a new wallet... :)

open_wallet_enter_pwd

-----

I propose:

open_wallet_enter_pwd_offer_new
open_wallet_no_pwd_offer_new
open_wallet_no_existing

PR #4069, opened 2018-03-07
New wallet wizard redesign

bauerj:
This implements the new design presented in #3916.

Also:
 * `HistoryList`'s `format_date` was moved to util
 * New icon for "Create new file"
 * Icon for "Open file" was changed to something more modern

Before:
python_2018-03-07_22-41-12
After:

Moving image

Closes: #3916
Closes: #3914


bauerj:
Hey,

I couldn't reproduce the width issue on my device but I think it should be better now. Can you test again?

SomberNight:
My previous remarks have been addressed; thanks. New ones here :P

This is how it looks for me now:
pr_4069_take2_pic2

If I change the min window height to 500 as in `InstallWizard.setMinimumSize(600, 500)`, then it becomes:
pr_4069_take2_pic1
(maybe that's too much, I don't know, but there should not be a vertical scrollbar in the outer box)

-----

The text in the fields of the "recent wallets" table should not be editable.

bauerj:
Addressed again 😎

SomberNight:
Thanks.
As for functionality, I like it. :)
However, I don't really like the fact that at this point `run_and_get_wallet` is ~240 lines... >.< Creating the layout is about 90 lines.. do you think it would be possible to move the layout-creating part into a separate method?

bauerj:
Sure, I'll refactor it next week.

gui/qt/installwizard.py
```diff
@@ -148,38 +146,116 @@ def __init__(self, config, app, plugins, storage):
...
+        palette.setColor(QPalette.Disabled, QPalette.Text, palette.color(QPalette.Normal, QPalette.Text))
+        self.wallet_name.setPalette(palette)
+        self.wallet_opened = QLabel(_('Last opened:'))
+
+        layout_wallet_info = QHBoxLayout()
+        layout_wallet_info.addWidget(self.wallet_name)
+        layout_wallet_info.addWidget(self.wallet_opened)
+        layout_wallet.addLayout(layout_wallet_info)
+
+        self.wallet_message = QLabel('Enter your password or choose another wallet.')
```
SomberNight: translate string _()

SomberNight:
Maybe the name of the selected wallet should always stay a QLineEdit.
Currently, the changing between input field/static text, has weird consequences.
If you have a wallet named `wallet6`, and would want to create a wallet named `wallet61`, the field will change from an input field to static text after typing `wallet6`.
Further, if there is some existing string in the input field, and you try to modify it in the middle, after every keypress (textEdited callback I guess), the cursor will be placed at the end.

bauerj:
I agree, I didn't think about wallet names that start with the name of an existing wallet. Both should be fixed now.

PR #4076, opened 2018-03-08
Fix #4073

### LINKED ISSUE #4073: same wallet can be opened multiple times via InstallWizard
duplicate_wallets

In Qt GUI, open wallet file, `File`>`New/Restore`.
The same wallet file can be opened multiple times simultaneously.

Lastrellik:
This should fix #4073.

### DIFF
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -198,7 +198,7 @@ class ElectrumGui:
             storage = WalletStorage(path, manual_upgrades=True)
             wizard = InstallWizard(self.config, self.app, self.plugins, storage)
             try:
-                wallet = wizard.run_and_get_wallet(self.daemon)
+                wallet = wizard.run_and_get_wallet(self.daemon.get_wallet)
             except UserCancelled:
                 pass
             except GoBack as e:
@@ -208,6 +208,7 @@ class ElectrumGui:
                 return
 
             if not self.daemon.get_wallet(wallet.storage.path):
+                # wallet was not in memory
                 wallet.start_threads(self.daemon.network)
                 self.daemon.add_wallet(wallet)
         try:
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -197,6 +197,8 @@ class ElectrumGui:
             wizard = InstallWizard(self.config, self.app, self.plugins, storage)
             try:
                 wallet = wizard.run_and_get_wallet()
+                if self.daemon.get_wallet(wallet.storage.path) is not None:
+                    wallet = self.daemon.get_wallet(wallet.storage.path)
             except UserCancelled:
                 pass
             except GoBack as e:
@@ -205,7 +207,7 @@ class ElectrumGui:
             if not wallet:
                 return
 
-            # to account for if the wallet is already open
+            # to account for if the wallet window is already open
             for w in self.windows:
                 if w.wallet.storage.path == wallet.storage.path:
                     w.bring_to_top()
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -184,11 +184,8 @@ class ElectrumGui:
         return w
 
     def start_new_window(self, path, uri):
-        # to account for if the wallet window is already open
-        for w in self.windows:
-            if w.wallet.storage.path == path:
-                w.bring_to_top()
-                return
+        '''Raises the window for the wallet if it is open.  Otherwise
+        opens the wallet and creates a new window for it'''
         try:
             wallet = self.daemon.load_wallet(path, None)
         except BaseException as e:
@@ -201,9 +198,7 @@ class ElectrumGui:
             storage = WalletStorage(path, manual_upgrades=True)
             wizard = InstallWizard(self.config, self.app, self.plugins, storage)
             try:
-                wallet = wizard.run_and_get_wallet()
-                if self.daemon.get_wallet(wallet.storage.path):
-                    wallet = self.daemon.get_wallet(wallet.storage.path)
+                wallet = wizard.run_and_get_wallet(self.daemon)
             except UserCancelled:
                 pass
             except GoBack as e:
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -184,43 +184,43 @@ class ElectrumGui:
         return w
 
     def start_new_window(self, path, uri):
-        '''Raises the window for the wallet if it is open.  Otherwise
-        opens the wallet and creates a new window for it.'''
-        for w in self.windows:
-            if w.wallet.storage.path == path:
-                w.bring_to_top()
-                break
-        else:
+        try:
+            wallet = self.daemon.load_wallet(path, None)
+        except BaseException as e:
+            traceback.print_exc(file=sys.stdout)
+            d = QMessageBox(QMessageBox.Warning, _('Error'),
+                            _('Cannot load wallet:') + '\n' + str(e))
+            d.exec_()
+            return
+        if not wallet:
+            storage = WalletStorage(path, manual_upgrades=True)
+            wizard = InstallWizard(self.config, self.app, self.plugins, storage)
             try:
-                wallet = self.daemon.load_wallet(path, None)
-            except BaseException as e:
-                traceback.print_exc(file=sys.stdout)
-                d = QMessageBox(QMessageBox.Warning, _('Error'),
-                                _('Cannot load wallet:') + '\n' + str(e))
-                d.exec_()
-                return
+                wallet = wizard.run_and_get_wallet()
+            except UserCancelled:
+                pass
+            except GoBack as e:
+                print_error('[start_new_window] Exception caught (GoBack)', e)
+            wizard.terminate()
             if not wallet:
-                storage = WalletStorage(path, manual_upgrades=True)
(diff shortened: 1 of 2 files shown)


gui/qt/__init__.py
```diff
@@ -184,43 +184,43 @@ def create_window_for_wallet(self, wallet):
...
+            wizard = InstallWizard(self.config, self.app, self.plugins, storage)
             try:
-                wallet = self.daemon.load_wallet(path, None)
-            except BaseException as e:
-                traceback.print_exc(file=sys.stdout)
-                d = QMessageBox(QMessageBox.Warning, _('Error'),
-                                _('Cannot load wallet:') + '\n' + str(e))
-                d.exec_()
-                return
+                wallet = wizard.run_and_get_wallet()
```
SomberNight: The InstallWizard will happily create a storage and then a wallet for any path.

gui/qt/__init__.py
```diff
@@ -184,43 +184,43 @@ def create_window_for_wallet(self, wallet):
...
-            except BaseException as e:
-                traceback.print_exc(file=sys.stdout)
-                d = QMessageBox(QMessageBox.Warning, _('Error'),
-                                _('Cannot create window for wallet:') + '\n' + str(e))
-                d.exec_()
                 return
+
+            # to account for if the wallet is already open
+            for w in self.windows:
+                if w.wallet.storage.path == wallet.storage.path:
```
SomberNight: Just because there is no window for it, a wallet might already be "open".

Lastrellik:
I don't understand... If we have a wallet window open and then close it, what do we want to happen if we try to open that wallet again? Shouldn't the ```start_new_window()``` method start a new window? Or should it start a window with the existing wallet thread if it already exists?

SomberNight:
The point is that we do not want duplicate Wallet objects. We definitely do not want duplicate Wallet objects with threads running.

However, I misread the control flow a bit; at the very beginning of the method,
`wallet = self.daemon.load_wallet(path, None)` should handle the case I was concerned about. Sorry.

In any case there is still a newly introduced bug here: if the above call returns a wallet, you don't check if there is an existing window for it, because you moved the check inside the `if not wallet`.

SomberNight:
Well actually my original comments still stand too.. because when we return from `wizard.run_and_get_wallet()`, the InstallWizard might have created a wallet for any path... argh, it's a bit complicated >.<

Lastrellik:
Hmm.. Running the code it looks like it does the behavior we're looking for, except I can't tell if there are any wallet threads leftover. I would like to understand this so I can be a better contributor. Like if there is a method somewhere that returns a list of all running wallet threads and a method that safely terminates a thread.

gui/qt/__init__.py
```diff
@@ -184,43 +184,43 @@ def create_window_for_wallet(self, wallet):
...
+                    w.bring_to_top()
+                    return
+
+            wallet.start_threads(self.daemon.network)
+            self.daemon.add_wallet(wallet)
+        try:
+            w = self.create_window_for_wallet(wallet)
```
SomberNight: As I said in a comment, if `wallet = self.daemon.load_wallet(path, None)` gets back a wallet then we will create a window for it here, even if there is already a window for this wallet open, which is a new bug.

SomberNight:
The references for open wallets are in `daemon.wallets`; this is a superset of the wallets for which there are windows open. So e.g. if `daemon.get_wallet(path)` returns a wallet then we should use that instead of creating a new Wallet object.

Note that I said "superset". Your check "to account for if the wallet is already open" is insufficient. Just because there is no window for a wallet; we might still have a reference to it in `daemon.wallets`.
Further, by the time this check runs, the new Wallet object is already created, along with a WalletStorage of course, which means we possibly modified the file on the filesystem for this wallet that we already have open.

I am not saying to kill wallet threads or similar; I am saying that we have references for wallets in `daemon.wallets` and we should use them.

Lastrellik:
That last commit modified ```run_and_get_wallet``` in installwizard.py to only return a new wallet if a thread for it isn't already open. Hopefully this should fulfill all the requirements of fixing this bug, the one thing that is a little weird is that if an encrypted wallet is already open and we try to open a new one using ```File>New/Restore``` it still asks for a password instead of just bringing the open wallet to the front. I don't really think that's a problem and it doesn't ask for a password if we just try to open it with ```File>Open```. 

During debugging I had the daemon print out which wallets were open and under no circumstance did it have duplicate wallets open at any given time. If this isn't a full solution, we're definitely pretty close.

SomberNight:
> it still asks for a password

You can test whether the wallet is in memory when setting up the message and password fields. You should also test in the while loop if the wallet is from memory and not decrypt the storage in that case, in fact you can just return the wallet there.

Also, I think it might be better to pass `daemon.get_wallet` to the `run_and_get_wallet` instead of `daemon`; but this may be subjective.

So, I was thinking something along the lines of:
<details>
 <summary>patch</summary>

```patch
diff --git a/gui/qt/__init__.py b/gui/qt/__init__.py
index 8744d64f..221fdf1d 100644
--- a/gui/qt/__init__.py
+++ b/gui/qt/__init__.py
@@ -198,7 +198,7 @@ class ElectrumGui:
             storage = WalletStorage(path, manual_upgrades=True)
             wizard = InstallWizard(self.config, self.app, self.plugins, storage)
             try:
-                wallet = wizard.run_and_get_wallet(self.daemon)
+                wallet = wizard.run_and_get_wallet(self.daemon.get_wallet)
             except UserCancelled:
                 pass
             except GoBack as e:
@@ -208,6 +208,7 @@ class ElectrumGui:
                 return
 
             if not self.daemon.get_wallet(wallet.storage.path):
+                # wallet was not in memory
                 wallet.start_threads(self.daemon.network)
                 self.daemon.add_wallet(wallet)
         try:
diff --git a/gui/qt/installwizard.py b/gui/qt/installwizard.py
index eab6f76a..3cd22ea4 100644
--- a/gui/qt/installwizard.py
+++ b/gui/qt/installwizard.py
@@ -148,7 +148,7 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
         self.raise_()
         self.refresh_gui()  # Need for QT on MacOSX.  Lame.
 
-    def run_and_get_wallet(self, daemon):
+    def run_and_get_wallet(self, get_wallet_from_daemon):
 
         vbox = QVBoxLayout()
         hbox = QHBoxLayout()
@@ -181,9 +181,10 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
 
         def on_filename(filename):
             path = os.path.join(wallet_folder, filename)
+            wallet_from_memory = get_wallet_from_daemon(path)
             try:
-                if daemon.get_wallet(path):
-                    self.storage = daemon.get_wallet(path).storage
+                if wallet_from_memory:
+                    self.storage = wallet_from_memory.storage
                 else:
                     self.storage = WalletStorage(path, manual_upgrades=True)
                 self.next_button.setEnabled(True)
@@ -195,7 +196,7 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
                     msg =_("This file does not exist.") + '\n' \
                           + _("Press 'Next' to create this wallet, or choose another file.")
                     pw = False
-                else:
+                elif not wallet_from_memory:
                     if self.storage.is_encrypted_with_user_pw():
                         msg = _("This file is encrypted with a password.") + '\n' \
                               + _('Enter your password or choose another file.')
@@ -207,6 +208,10 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
                     else:
                         msg = _("Press 'Next' to open this wallet.")
                         pw = False
+                else:
+                    msg = _("This file is already in memory.") + '\n' \
+                          + _("Press 'Next' to create/focus window.")
+                    pw = False
             else:
                 msg = _('Cannot read file')
                 pw = False
@@ -231,6 +236,9 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
                 return
             if not self.storage.file_exists():
                 break
+            wallet_from_memory = get_wallet_from_daemon(self.storage.path)
+            if wallet_from_memory:
+                return wallet_from_memory
             if self.storage.file_exists() and self.storage.is_encrypted():
                 if self.storage.is_encrypted_with_user_pw():
                     password = self.pw_e.text()
@@ -255,7 +263,7 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
                             _('Failed to decrypt using this hardware device.') + '\n' +
                             _('If you use a passphrase, make sure it is correct.'))
                         self.stack = []
-                        return self.run_and_get_wallet(daemon)
+                        return self.run_and_get_wallet(get_wallet_from_daemon)
                     except BaseException as e:
                         traceback.print_exc(file=sys.stdout)
                         QMessageBox.information(None, _('Error'), str(e))
@@ -302,7 +310,8 @@ class InstallWizard(QDialog, MessageBoxMixin, BaseWizard):
             self.run(action)
             return self.wallet
 
-        return Wallet(self.storage) if not daemon.get_wallet(path) else daemon.get_wallet(path)
+        self.wallet = Wallet(self.storage)
+        return self.wallet
 
     def finished(self):
         """Called in hardware client wrapper, in order to close popups."""
```
</details>

Lastrellik:
@SomberNight so this is a really nice solution :) I appreciate all your help. I can tell that you're very good at this. One day I'll catch up :P

PR #4131, opened 2018-03-15
Add testnet mode to .desktop file

yura-pakhuchiy:
Easy way to start testnet version from linux DE

### DIFF
--- a/electrum.desktop
+++ b/electrum.desktop
@@ -14,4 +14,8 @@ StartupNotify=false
 Terminal=false
 Type=Application
 MimeType=x-scheme-handler/bitcoin;
+Actions=Testnet;
 
+[Desktop Action Testnet]
+Exec=electrum --testnet %u
+Name=Testnet mode


bauerj:
Can you show a screenshot? How is this displayed?

yura-pakhuchiy:
It depends on DE, in GNOME3 it looks like this (right click menu on the launcher icon):
screenshot from 2018-03-17 23-15-42

ecdsa:
I don't think this is a good idea..  testnet is for testing, should not be exposted to all users

yura-pakhuchiy:
Testnet can be used not only for testing, it is a great way to try various bitcoin and electrum features without risk of losing money. Bitcoin can be quite confusion to new users, some people prefer learning by trying and testnet works great for that.

bauerj:
It is also exposed to Windows users - why isn't it a problem there?

yura-pakhuchiy:
Besides windows version creates shortcut for testnet mode, why linux version should not?

ecdsa:
@bauerj maybe it is a problem there...

yura-pakhuchiy:
@ecdsa could you explain why testnet should not be available for all users? It is quite useful for learning.

yura-pakhuchiy:
Besides in linux testnet mode appears only in right click menu, not a separate item in list of programs. I do not feel that it is a lot of exposure.

ecdsa:
@yura-pakhuchiy there is a distinction between "not available" and "not availble in the desktop icon menu".
Adding a menu item for every function is how you make an application clogged and unusable.

ecdsa:
anyway, I will still merge it for consistency

PR #4149, opened 2018-03-20
setup.py: fix icon path for user installation

toxeus:
Before, the user installation used the wrong directory
to store the application's icon resulting in the desktop
environment falling back to an ugly default icon. Now,
the correct icon is displayed in the menu.

"By default, apps should look in $HOME/.icons (for backwards
compatibility), in $XDG_DATA_DIRS/icons and in /usr/share/pixmaps
(in that order)."
https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,15 +27,17 @@ if platform.system() in ['Linux', 'FreeBSD', 'DragonFly']:
     parser.add_argument('--root=', dest='root_path', metavar='dir', default='/')
     opts, _ = parser.parse_known_args(sys.argv[1:])
     usr_share = os.path.join(sys.prefix, "share")
+    icons_dirname = 'pixmaps'
     if not os.access(opts.root_path + usr_share, os.W_OK) and \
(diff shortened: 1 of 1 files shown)


bauerj:
That sounds right but doesn't your commit put it in `~/.local/share/icons` if `$XDG_DATA_HOME` is not set? According to your quote it should look in `~/.icons` in stead.

toxeus:
Nope. `~/.icons` should not be used by new apps that's why they explicitly mention it as backwards compatibility. `~/.local/share` is the XDG spec's fallback if `$XDG_DATA_HOME` is not set.

[spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):
>  $XDG_DATA_HOME defines the base directory relative to which user specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

PR #4161, opened 2018-03-22
allow loading plugins from custom folder

SomberNight:
The usecase is importing 3rd party plugins when using the binaries.
E.g. using the Windows standalone exe, currently you can't use custom plugins.

The idea is that the user creates a folder somewhere, puts all custom plugins into that folder; puts an empty `__init__.py` there too (so that python can import it as a package); and then selects the folder using the GUI. The path to the folder is persisted in the config.


ecdsa:
I think this would be an avenue for stealware. Of course, in theory this is not more dangerous than existing attack vectors, such as keyloggers. However, in practice, it will make stealing coins a lot easier, because the attacker will no longer need to run malware while electrum is running. 

It is true that my initial plan was to make plugins easy to add by users. However, there is no way we can ensure plugins are safe, other than reviewing their code. I have also been thinking about an "app store" for plugins, but that still implies that we review and approve them...

PR #4166, opened 2018-03-23
electrum.desktop: improve user installation

toxeus:
Before, the desktop entry file would only work correctly
if the user has added `$HOME/.local/bin` to his or her
PATH. The PATH variable is an implementation detail and
the user shouldn't be exposed to it.

Now, the shell is used to look for executables in the
aforementioned folder. System-wide installations remain
intact.

### DIFF
--- a/electrum.desktop
+++ b/electrum.desktop
@@ -3,7 +3,7 @@
 
 [Desktop Entry]
 Comment=Lightweight Bitcoin Client
-Exec=electrum %u
+Exec=sh -c 'PATH="$HOME/.local/bin:$PATH" electrum %u'
 GenericName[en_US]=Bitcoin Wallet
 GenericName=Bitcoin Wallet
 Icon=electrum
@@ -17,5 +17,5 @@ MimeType=x-scheme-handler/bitcoin;
 Actions=Testnet;
 
 [Desktop Action Testnet]
-Exec=electrum --testnet %u
+Exec=sh -c 'PATH="$HOME/.local/bin:$PATH" electrum --testnet %u'
 Name=Testnet mode


gileri:
Isn't it dangerous to allow this on system-wide .desktop files ?

gileri:
I may be wrong, but adding a user-owned directory into `PATH` in a system-wide `.desktop` file is dangerous as it could allow any user-owned process to add an executable script/binary in `~/.local/bin/electrum` and it will be executed by the user when launching electrum, even if the user only use system-wide, root-owned binaries.

It is not a huge flaw, but I think it it can help another exploit.

Now if the user do a local installation of electrum (in `~/.local/bin/`) there is of course no (additional) harm in having `~/.local/bin/` in `PATH` of the `.desktop` file

toxeus:
Thank you for the explanation :+1:

What you're describing is correct. However, if on your system a user-owned process is trying to trick you to execute something else instead of the `electrum` you're intending to execute then it's game over anyway. Basically, this user-owned process is a malware already. Why should it launch a second malware process that is executed instead of the genuine `electrum`? That doesn't really make a lot of sense. It can already log your keystrokes, upload your private keys to some remote host, manipulate `PATH` in your `.bashrc` or `.profile`, etc.

Last but not least, there exists a simple fix for the "flaw" you're describing: `"PATH=\"\\$HOME/.local/bin:\\$PATH\"` needs to be changed to `"PATH=\"\\$PATH:\\$HOME/.local/bin\"` [here](https://github.com/spesmilo/electrum/blob/629b9cb3b54cd067da9673c8fb84d49e417903bf/electrum.desktop#L6) and [here](https://github.com/spesmilo/electrum/blob/629b9cb3b54cd067da9673c8fb84d49e417903bf/electrum.desktop#L20). Feel free to open a PR if you feel strongly about this :wink:

gileri:
>What you're describing is correct. However, if on your system a user-owned process is trying to trick you to execute something else instead of the electrum you're intending to execute then it's game over anyway. Basically, this user-owned process is a malware already.

Not necessarily : a process with a vulnerability could for example only allow writing files but not executing arbitrary code.

I agree that this is far from the lowest hanging fruit, but I think it is not sound to add this.

>Last but not least, there exists a simple fix for the "flaw" you're describing: "PATH=\"\\$HOME/.local/bin:\\$PATH\" needs to be changed to "PATH=\"\\$PATH:\\$HOME/.local/bin\"

I will make a PR after #4300, thanks for the suggestion.

PR #4178, opened 2018-03-25
Make .dmg creation deterministic

bauerj:
Since it seems to be impossible to create deterministic images with `hdiutil`, we can use `genisoimage` and `libdmg-hfsplus` instead (like Bitcoin does for their builds).

The build script was tested on a Ubuntu VM.

In addition to being deterministic, the new script also features a fancy folder design:

bildschirmfoto 2018-03-25 um 22 21 01


ecdsa:
thanks. what is contrib/build-osx/template/.DS_Store?

bauerj:
That file contains the directory metadata, e.g. the window size, icon size and position.

ecdsa:
I had to do this to make it run:
```

-dmg dmg Electrum_uncompressed.dmg $build_dir/Electrum.dmg || fail "Unable to create compressed dmg"
+dmg dmg Electrum_uncompressed.dmg $build_dir/Electrum.dmg || fail "Unable to create compressed dmg"

```

ecdsa:
also, I suppose this would be better: 
```
-build_dir=$(dirname "$0")
+build_dir=$(dirname "$1")
```
however when I changed that, I noticed that the final hash differs.
so it seems to be dependent on build_dir

ecdsa:
ok, I got it to work, but the Applications icon does not work for me

ecdsa:
merged after changes

PR #4194, opened 2018-03-28
trezor: pass transport parameter explicitly as a keyword argument

romanz:
Following https://github.com/trezor/python-trezor/pull/241.


SomberNight:
Why do we pass transport to ProtocolMixin anyway?

matejcik:
maybe because ProtocolMixin calls super() ?

I dunno. The mixin architecture is going away soon, this is madness.

PR #4210, opened 2018-04-01
change testnet p2wsh-p2sh prefix to 0x024289ef to reflect Upub

fivepiece:
Currently testnet p2wsh-p2sh master pubkey prefix is Ups4, this changes it to Upub.

### DIFF
--- a/lib/tests/test_wallet_vertical.py
+++ b/lib/tests/test_wallet_vertical.py
@@ -5,6 +5,7 @@ import lib.bitcoin as bitcoin
 import lib.keystore as keystore
 import lib.storage as storage
 import lib.wallet as wallet
+from lib import constants
 
 from plugins.trustedcoin import trustedcoin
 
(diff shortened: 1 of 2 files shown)


fivepiece:
Added new test for p2sh-p2wsh 2of2 multisig on testnet, had to add :
```python
from lib import constants
```
To be able to call `constants.set_testnet()` then back to `constants.set_mainnet()`

SomberNight:
Arghhh.. I even tested those constants... not sure what happened. :/
Thank you.
At least it's testnet; we will just break compat...

lib/tests/test_wallet_vertical.py
```diff
@@ -280,3 +281,26 @@ def test_bip39_multisig_seed_p2sh_segwit(self, mock_write):
 
         self.assertEqual(w.get_receiving_addresses()[0], '35LeC45QgCVeRor1tJD6LiDgPbybBXisns')
         self.assertEqual(w.get_change_addresses()[0], '39RhtDchc6igmx5tyoimhojFL1ZbQBrXa6')
+
+    @mock.patch.object(storage.WalletStorage, '_write')
+    def test_bip39_multisig_seed_p2sh_segwit_testnet(self, mock_write):
+        constants.set_testnet()
```
SomberNight: This is not really OK in general because tests might run in parallel.
If you try to do this in e.g. test_bitcoin.py, there will be dozens of failures.
I'll do a bit of refactoring in a few hours and fix this though, so just leave it.

SomberNight:
Related changes in https://github.com/spesmilo/electrum/commit/1eae16aa3e78a435d9de1de4f85bb4929cda6d5e https://github.com/spesmilo/electrum/commit/1eae16aa3e78a435d9de1de4f85bb4929cda6d5e https://github.com/spesmilo/electrum/commit/ff57c198b4f7acfb981b95bac6e69e769c0a75c4

I've also added more tests for this in https://github.com/spesmilo/electrum/commit/58273fc841c8ca80698cdafe790f8152f264183c

lib/tests/test_wallet_vertical.py
fivepiece: Ah yes I had a feeling I was being a bit hacky with that, cheers :)

PR #4233, opened 2018-04-06
digitalbitbox: enable native segwit

benma:


benma:
@ecdsa there is a regression where the tx, which is passed to the plugin for signing, is missing the 'pubkeys' field of an input. I worked around it the commit (https://github.com/spesmilo/electrum/pull/4233/commits/d1ee8ec8f7fbda8aa6905afd05c8af3aa0544438), but I wonder when and where the pubkey field should be populated properly.

SomberNight:
Can you give more details regarding the 'pubkeys' field missing? I can't reproduce.

benma:
I sent funds to a segwit wallet (either wrapped in p2sh or native), and when sending the funds out by using the Send button, the transaction that is passed for signing doesn't have this key:

`tx.inputs()[0]['pubkeys']`.

Note that it works with the Preview/Sign workflow, as the Preview also serializes the transaction and populates the missing fields.

The proper way to fix it is to have it be present before it reaches the signing functions, but I am a bit confused about the whole workflow.

SomberNight:
It should be ok now with https://github.com/spesmilo/electrum/commit/4ff3791cb93fcdda0b534bc2be575121f4d59705

Can you remove the "hack" commit?

SomberNight:
I've cherry-picked the first and third commits manually.
https://github.com/spesmilo/electrum/commit/cf98c8f2c15e5b0fdec0ce2150afc8993601f21a
https://github.com/spesmilo/electrum/commit/f7d1d5a7c75b2b16b7d2c2c79f82f892bcdb5ca5
Thank you. :)

benma:
No, thank you :P Sorry for the delay!

Issue #4241, opened 2018-04-08
Regtest support


fivepiece:
What are your thoughts about a PR adding regtest support for electrum?
The use case would be to have an easy and reliable way of testing a few edge cases regarding incoming and outgoing payemnts that are hard to achieve even on testnet without mining hardware and some luck involved such as blockchain re-orgs and transaction double spending and replacement both while the client is online and offline.
The instantaneous nature, disposability and the ability to `invalidateblock` and `reconsiderblock` in regtest helps a lot with testing these scenarios.

AFAICT there could be two approaches to make it work.  The change would either be extensive or minimal, depending on which (if any) approach will be deemed as the correct way to go.

First I'll list the changes required for the naive approach to work.  This will be done by treating regtest as a second testnet (`TESTNET = True`), changing a few minimal parameters:

1. Add a `--regtest` option, much like `--testnet` it will create a `regtest` subdirectory in the datadir to hold the relevant files
2. Change the human readable part for bech32 addresses from `tb` to `bcrt`
3. Set an empty checkpoints parameter: `CHECKPOINTS = []`
4. Set the genesis block hash to `0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206`
5. Create a new `servers_regtest.json` only `localhost` as the server and use it for setting `DEFAULT_SERVERS`
6. Make sure that `--regtest` and `--testnet` are exclusive

As electrum disregards PoW target on testnet, these 6 changes would be enough for it to play along with a regtest chain.  The thing is, regtest does have well defined rules for PoW validation that aren't as crazy as testnet. 
Specifically considering electrum's verification of headers, the retarget period would be 144 blocks (with a two weeks timespan), and the target itself never changes and always stays at it's very high initial value (much higher than mainnet).

The less naive and more extensive approach to adding regtest support would be to add all the changes from the first list, and also implement regtest's PoW rules.  Personally, I'm not sure if it's worth while as the point behind this change is not chain validation.

Another caveat that is worth mentioning is that aside from the Digital Bitbox, I don't believe any hardware wallet could play along with the `bcrt` human readable part of bech32 addresses.  Again, this change is not really intended for testing hardware wallets so perhaps this is a non issue.

Myself, I'm happy with the naive change as it suits my needs but I'm happy to hear your thoughts before I waste your time with a PR :)

Cheers and thanks for reading

SomberNight:
[...]

PR #4242, opened 2018-04-08
add regtest support

fivepiece:
See #4241 for meta discussion

1. Add a `--regtest` option
2. Change the human readable part for bech32 addresses from `tb` to `bcrt`
3. Set an empty checkpoints parameter: `CHECKPOINTS = []`
4. Set the genesis block hash to `0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206`
5. Create a new `servers_regtest.json` only `localhost` as the server and use it for setting `DEFAULT_SERVERS`
6. Make sure that `--regtest` and `--testnet` are exclusive


lib/servers_regtest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "127.0.0.1": {
+        "pruning": "-",
+        "s": "50002",
+        "t": "50001",
```
SomberNight: I guess these should rather be `'t': '51001', 's': '51002'`, which is also what you specified as `BitcoinRegtest.DEFAULT_PORTS`
https://github.com/kyuupichan/electrumx/blob/fbf0ec4d9b9bb5c4c6bdcbf27c3fd4d9def91670/lib/coins.py#L571

fivepiece: Yep, changing now.

fivepiece:
thanks.  ping me when you want me to squash the commits

lib/constants.py
```diff
@@ -93,6 +93,34 @@ class BitcoinTestnet:
     }
 
 
+class BitcoinRegtest:
```
SomberNight: Actually, I'm thinking it might be better to inherit from `BitcoinTestnet` so some of the fields don't need to be duplicated :P

fivepiece: No problem :)

fivepiece:
Updated to address @SomberNight comments

PR #4257, opened 2018-04-13
contrib: detect builds with uncommitted changes

toxeus:
The `--dirty` flag causes `git describe` to append
`-dirty` to the version string it outputs. This is
useful to detect if a build was - intentionally or
not - created with uncommitted changes.

https://git-scm.com/docs/git-describe#git-describe---dirtyltmarkgt


bauerj:
Thanks!

Issue #2447, opened 2017-05-24
Fee estimation on command line


bsbits:
Electrum provide fee estimation on the GUI with excellent features like target blocks.

But is there anything similar or usable via command line ? I do not want to use an external service to compute or estimate the required fees.

ecdsa:
if you run the daemon, it will use the same fee estimates as the GUI.
you need to open the GUI to set your fee preferences.

bsbits:
Except I will be using electrum on servers as a hot wallet for my websites, so no GUI.

ecdsa:
well, you can check what are the relevant configuration variables on another machine, and set them using setconfig

dabura667:
You could add your own CLI command to commands.py:

```python
@command('n')
def getestimatedfee(self, number_of_blocks):
    """Returns the fee estimation from the server.
    smallest value is 2.
    """
    number_of_blocks = int(number_of_blocks)
    if number_of_blocks < 2:
        raise BaseException('number_of_blocks must be >= 2')
    return self.network.synchronous_get(('blockchain.estimatefee', [number_of_blocks]))
```
usage
```bash
electrum getestimatedfee
```

dabura667:
Oh, you want to set the dynamic fees without opening GUI? Sorry, I misread. @bsbits 

`electrum setconfig dynamic_fees True`
to set the dynamic fees on. only need to be done once, and it will save to your config file.

`electrum setconfig fee_level 3`
To set the dynamic target to the 4th tick. (4 is the fastest, 3 is a 2 block estimate, 0 is the slowest dynamic fee setting)

Also just in case something gets sent by non-dynamic fees due to a bug, I might just set:

`electrum setconfig fee_per_kb 300000`
Maybe do this before the other two commands.

Now as long as you use the same user to run electrum (these commands modify the OS user's config, not global config) it should send using the dynamic fees you want.

bsbits:
@dabura667 : amazing.

TJC:
I have been using the command-line version of Electrum, and always manually calculating the fee to pass to the payto command. (ie. `payto -f amount`).

I've messed this up before, though, when I sent a transaction that was twice as many bytes as usual -- and so my fee was effectively half the usual. It took about twenty hours to get confirmed.

While I now know from this issue about how to configure dynamic fees, there's still work to be done. I'd love it if the "help payto" command could let the user know how to enable dynamic fees.

[...]

PR #4264, opened 2018-04-15
Getfeerate cmd options

SomberNight:
closes https://github.com/spesmilo/electrum/issues/2447#issuecomment-381360086

```
$ ./electrum getfeerate
26000
$ ./electrum getfeerate --fee_method "static" --fee_level 0.1
10000
$ ./electrum getfeerate --fee_method "static" --fee_level 0.4
50000
$ ./electrum getfeerate --fee_method "static" --fee_level 0.8
150000
$ ./electrum getfeerate --fee_method "eta" --fee_level 0.1
1081
$ ./electrum getfeerate --fee_method "eta" --fee_level 0.4
10705
$ ./electrum getfeerate --fee_method "eta" --fee_level 0.8
21190
$ ./electrum getfeerate --fee_method "mempool" --fee_level 0.1
1000
$ ./electrum getfeerate --fee_method "mempool" --fee_level 0.4
1000
$ ./electrum getfeerate --fee_method "mempool" --fee_level 0.8
20000
```

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -657,10 +657,23 @@ class Commands:
         return self.wallet.is_up_to_date()
 
     @command('n')
-    def getfeerate(self):
-        """Return current optimal fee rate per kilobyte, according
-        to config settings (static/dynamic)"""
-        return self.config.fee_per_kb()
+    def getfeerate(self, fee_method=None, fee_level=None):
+        """Return current suggested fee rate (in sat/kvByte), according to config
+        settings or supplied parameters.
+        """
+        if fee_method is None:
+            dyn, mempool = None, None
+        elif fee_method.lower() == 'static':
+            dyn, mempool = False, False
+        elif fee_method.lower() == 'eta':
+            dyn, mempool = True, False
+        elif fee_method.lower() == 'mempool':
+            dyn, mempool = True, True
+        else:
+            raise Exception('Invalid fee estimation method: {}'.format(fee_method))
+        if fee_level is not None:
+            fee_level = Decimal(fee_level)
+        return self.config.fee_per_kb(dyn=dyn, mempool=mempool, fee_level=fee_level)
 
     @command('')
     def help(self):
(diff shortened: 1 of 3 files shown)


lib/commands.py
```diff
@@ -657,10 +657,23 @@ def is_synchronized(self):
...
+        settings or supplied parameters.
+        """
+        if fee_method is None:
+            dyn, mempool = None, None
+        elif fee_method == 0:  # static
+            dyn, mempool = False, False
+        elif fee_method == 1:  # ETA
+            dyn, mempool = True, False
+        elif fee_method == 2:  # mempool
+            dyn, mempool = True, True
```
haarts: Why not dispense with the magic integers and just use the strings 'static'/'ETA'/'mempool'?

lib/commands.py
```diff
@@ -657,10 +657,23 @@ def is_synchronized(self):
         return self.wallet.is_up_to_date()
 
     @command('n')
-    def getfeerate(self):
-        """Return current optimal fee rate per kilobyte, according
-        to config settings (static/dynamic)"""
-        return self.config.fee_per_kb()
+    def getfeerate(self, fee_method=None, fee_level=None):
```
haarts: Does it make sense to have `fee_level` default to `None`. Doing so complicates the code with several if statements. Also passing a default signals the argument is optional which it does seem to be.

lib/commands.py
```diff
@@ -657,10 +657,23 @@ def is_synchronized(self):
...
+        """Return current suggested fee rate (in sat/kvByte), according to config
```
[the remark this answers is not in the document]
SomberNight: kilo-virtual-byte >.< looks weird, I know
1 vByte = 4 weight
1 kvByte = 4 kiloweight

lib/commands.py
SomberNight: both arguments are optional. if they are not passed, we deduce their value from the config.
this also makes this change backwards compatible

lib/commands.py
SomberNight: good point

PR #4265, opened 2018-04-15
qt pay to script: append hex as is, instead of pushing to script

SomberNight:
It's not always the case that `push_script` is what the user wants.


Platinumwrist:
Great that's excellent ⏳

SomberNight:
(maybe we should combine https://github.com/spesmilo/electrum/blob/3312dd152fd80ab09bb53e143d0507f20f167aab/lib/bitcoin.py#L202 and `push_script`; have it on the master branch too; instead of this PR)

SomberNight:
this is no longer relevant with https://github.com/spesmilo/electrum/commit/e13183ea7a4d43c861a973f4792980d637005074#diff-b026615fe6f31bfcbd95d0860025eccdR209

PR #4272, opened 2018-04-17
added therocktrading exchange rate

mhitcher:


bauerj:
The website seems to be offline right now:

chrome_2018-04-17_22-33-06

Are you affiliated with therocktrading?

mhitcher:
https://www.therocktrading.com/ is not offline. it is the oldest, most trusted and most used exchange in italy.

lib/exchange_rate.py
```diff
@@ -328,6 +327,11 @@ def get_rates(self,ccy):
     def history_ccys(self):
         return ['BRL']
 
+class TheRockTrading(ExchangeBase):
+
+    def get_rates(self, ccy):
+        json = self.get_json('api.therocktrading.com', '/v1/funds/BTC{}/ticker'.format(ccy))
+        return {ccy: Decimal(json['last'])}
```
SomberNight: Could you, as un-intuitive as it sounds, hardcode the `EUR` ccy, instead of using the `ccy` argument?
The reason is that `currencies.json` is actually a generated file, and during generation `get_currencies` is called, which calls `get_rates`  with an empty string ccy, and expects a dictionary where the keys are the supported currencies. That's why all the other exchanges do that. It's a bit crazy, and could probably use a rewrite...
i.e. as is, the auto-generation would not include `TheRockTrading`

lib/exchange_rate.py
```diff
@@ -328,6 +327,11 @@ def get_rates(self,ccy):
     def history_ccys(self):
         return ['BRL']
```
SomberNight: While at it, could you also remove this `history_ccys` method for `NegocieCoins` too? :)
It is not useful without a `request_history` method.

lib/exchange_rate.py
mhitcher: ok, now TheRockTrading is working, but others are giving me error,

BitStamp ok
Unocoin error
itBit ok
Bitso ok
Coinsecure error
Bitmarket ok
Kraken ok
MercadoBitcoin ok
NegocieCoins ok
WEX ok
BitcoinAverage ok
BlockchainInfo ok
BitFlyer ok
BTCParalelo error
BitcoinVenezuela error
BitPay ok
Coinbase ok
LocalBitcoins ok
Zaif error
BTCChina error
TheRockTrading ok
Winkdex error
Bitcointoyou ok
Bitbank ok
CoinDesk ok
Foxbit ok
Bitvalor ok

SomberNight:
thanks

Issue #4279, opened 2018-04-19
Consider restructuring the project modules (lib/gui/plugins)


toxeus:
This is a minor issue but still worth reporting.

`test_electrum_seed_2fa` was added in this [commit](https://github.com/spesmilo/electrum/commit/6d88eab005c306eb9fda6249cf387656c56a2122) which imports `plugins.trustedcoin` which imports [`electrum` and its submodules](https://github.com/spesmilo/electrum/blob/f6cb26be1fa1d1b3de92e5fce6c24f60d70d357a/plugins/trustedcoin/trustedcoin.py#L33-L44). As a consequence breaking changes in `electrum` or its submodules will not break the test until the `electrum` package is reinstalled. Even worse, a `git pull` might suddenly result in the test failing, assuming that the test and trustedcoin were changed.

toxeus:
@SomberNight in my personal projects I use the following project structure which isn't affected at all by the problems electrum currently faces.

```sh
repo
├── script
├── pkg1
├── pkg2
└── ...
```

The [hacks in the main script](https://github.com/spesmilo/electrum/blob/376a815458d7765c3e3b4c22b74552bd4702b163/electrum#L68-L72) are not needed and the unit tests use the newest code from the repo because Python puts [the directory containing the input script](https://docs.python.org/3.6/tutorial/modules.html#the-module-search-path) on top of its module search path (consequently `pytest` needs to be executed from the projects root dir; not sure if this is already the case).

For the electrum project the structure would therefore look like the following (only listing relevant files and folders).

```sh
electrum
├── electrum.py
├── electrum
├── electrum_gui
└── electrum_plugins
```

SomberNight:
@ecdsa @ysangkok @bauerj 
probably of interest to all of you

SomberNight:
(context: https://github.com/spesmilo/electrum/pull/4286#issuecomment-383161468 )

ysangkok:
@toxeus Try this: https://github.com/ysangkok/electrum/tree/file_layout_change

What do you think? I haven't used your suggested layout since it requires more hacks when there is no module that contains all relevant sources. This still needs to be tested on all platforms except Linux.

toxeus:
@ysangkok sorry for the late reply. I don't know how to review 265 commits efficiently.

![screenshot_2018-07-07 ysangkok electrum 1](https://user-images.githubusercontent.com/7702854/42409168-96c1c2fc-81d6-11e8-8b65-c1831d231916.png)

[...]

Issue #4259, opened 2018-04-13
Feature request: fractional fees


l29ah:
Now i can't input a fee less than one satoshi per byte, while the network would gladly accept one satoshi for the whole transaction currently!

PR #4286, opened 2018-04-20
Format the transaction window fee rate with 1 decimal place

Empact:
For consistency with the inline display of sat/b in the unconfirmed list view.
See lib/wallet.py, line 1173.
https://github.com/spesmilo/electrum/blob/master/lib/wallet.py#L1173

This is particularly important when the fees/b are low and the fractional amount is a significant percentage of the total fee, e.g. between 1 and 2 sat/b.

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -649,7 +649,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         return text
 
     def format_fee_rate(self, fee_rate):
-        return format_satoshis(fee_rate/1000, self.num_zeros, 0) + ' sat/byte'
+        return format_satoshis(fee_rate/1000, self.num_zeros, 0, precision=1) + ' sat/byte'
 
     def get_decimal_point(self):
         return self.decimal_point
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -44,8 +44,8 @@ from electrum.bitcoin import COIN, is_address, TYPE_ADDRESS
 from electrum import constants
 from electrum.plugins import run_hook
 from electrum.i18n import _
-from electrum.util import (format_time, format_satoshis, PrintError,
-                           format_satoshis_plain, NotEnoughFunds,
+from electrum.util import (format_time, format_satoshis, format_fee_satoshis,
+                           format_satoshis_plain, NotEnoughFunds, PrintError,
                            UserCancelled, NoDynamicFeeEstimates, profiler,
                            export_meta, import_meta, bh2u, bfh, InvalidPassword)
 from electrum import Transaction
@@ -649,7 +649,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         return text
 
     def format_fee_rate(self, fee_rate):
-        return format_satoshis(fee_rate/1000, self.num_zeros, 0, precision=1) + ' sat/byte'
+        return format_fee_satoshis(fee_rate/1000, self.num_zeros) + ' sat/byte'
 
     def get_decimal_point(self):
         return self.decimal_point
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -639,7 +639,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
             self.require_fee_update = False
 
     def format_amount(self, x, is_diff=False, whitespaces=False):
-        return format_satoshis(x, is_diff, self.num_zeros, self.decimal_point, whitespaces)
+        return format_satoshis(x, self.num_zeros, self.decimal_point, is_diff=is_diff, whitespaces=whitespaces)
 
     def format_amount_and_units(self, amount):
         text = self.format_amount(amount) + ' '+ self.base_unit()
@@ -649,7 +649,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
         return text
 
     def format_fee_rate(self, fee_rate):
-        return format_satoshis(fee_rate/1000, False, self.num_zeros, 0, False)  + ' sat/byte'
+        return format_satoshis(fee_rate/1000, self.num_zeros, 0) + ' sat/byte'
 
     def get_decimal_point(self):
         return self.decimal_point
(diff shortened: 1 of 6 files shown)


gui/qt/main_window.py
```diff
@@ -649,7 +649,7 @@ def format_amount_and_units(self, amount):
         return text
 
     def format_fee_rate(self, fee_rate):
-        return format_satoshis(fee_rate/1000, False, self.num_zeros, 0, False)  + ' sat/byte'
+        return format_satoshis(fee_rate/1000, False, self.num_zeros, 1, False)  + ' sat/byte'
```
SomberNight: This is not doing what you think it does. You are effectively changing units here.
For example, instead of `6. sat/byte`, it will return `0.6 sat/byte`.

SomberNight:
A working and non-hacky solution to this would be welcome.
A full solution would be to properly allow fractional fees everywhere, resolving also https://github.com/spesmilo/electrum/issues/4259

gui/qt/main_window.py
Empact: Ah yeah, thanks.

Empact:
My apologies for a sloppy attempt at a first contribution.

Empact:
Kind of a shame that there's no automated test suite for the project. Do you have intentions on adding one?

Empact:
@SomberNight updated with another attempt at it, this time with tests. Note there's no limit on the `format_fee_rate` precision currently, and I imagine there should be.

Empact:
Also I get this test failure which I don't have any leads on:
```
======================================================================
 ERROR: gui.qt (unittest.loader._FailedTest)
 ----------------------------------------------------------------------
 ImportError: Failed to import test module: gui.qt
   File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/loader.py", line 462, in _find_test_path
     package = self._get_module_from_name(name)
   File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
     __import__(name)
   File "/Users/ben/Development/Bitcoin/electrum/gui/qt/__init__.py", line 63, in <module>
   File "/Users/ben/Development/Bitcoin/electrum/gui/qt/main_window.py", line 47, in <module>
     from electrum.util import (format_time, format_fee_rate, format_satoshis_amount,
 ImportError: cannot import name 'format_fee_rate'
```

SomberNight:
Thanks, will have a look later.

---

Re the test failure, probably similar to https://github.com/spesmilo/electrum/issues/4279
Electrum is comprised of three modules: `lib` or `electrum`, `gui` or `electrum_gui`, and `plugins` or `electrum_plugins` (first of each pair is the local/source name, second is the installed module name).
This is causing problems sometimes..
Depending on the unit testing framework, when you import `electrum` you might import "outdated" code :/ This is of course not a problem for Travis, who just installs the latest code every time.

There is a bunch of hacks in the main script so that the same problem does not (usually..) affect development itself; notably:
https://github.com/spesmilo/electrum/blob/23c29715af210f49899d5a68f948b4f594135a52/electrum#L67-L72

Empact:
@SomberNight alright I squashed this into a relatively minimal change that avoids hitting the aforementioned test failure.

Empact:
For the record, the test command I was running was `python3 -m unittest discover` from the root of the project.

gui/qt/main_window.py
```diff
@@ -639,7 +639,7 @@ def timer_actions(self):
             self.require_fee_update = False
 
     def format_amount(self, x, is_diff=False, whitespaces=False):
-        return format_satoshis(x, is_diff, self.num_zeros, self.decimal_point, whitespaces)
+        return format_satoshis(x, self.num_zeros, self.decimal_point, is_diff=is_diff, whitespaces=whitespace)
```
SomberNight: typo `whitespaces=whitespace` (missing `s`)

lib/util.py
```diff
@@ -434,6 +432,8 @@ def format_satoshis(x, is_diff=False, num_zeros = 0, decimal_point = 8, whitespa
         result = " " * (15 - len(result)) + result
     return result
 
+FEE_PRECISION=1
```
SomberNight: Can you add a comment to explain this? e.g. "number of digits after decimal separator"

lib/simple_config.py
```diff
@@ -367,7 +367,7 @@ def get_fee_text(self, pos, dyn, mempool, fee_rate):
         text is what we target: static fee / num blocks to confirm in / mempool depth
         tooltip is the corresponding estimate (e.g. num blocks for a static fee)
         """
-        rate_str = (format_satoshis(fee_rate/1000, False, 0, 0, False)  + ' sat/byte') if fee_rate is not None else 'unknown'
+        rate_str = format_satoshis(fee_rate/1000, 0, 0, precision=FEE_PRECISION) + ' sat/byte' if fee_rate is not None else 'unknown'
```
SomberNight: I think we should replace the inline `if` with a multi-line standard `if`.
It always takes me a few seconds to think about operator precedence when seeing long expressions without brackets.. :) but also, the line is getting long, so let's do standard `if`

Empact: Yeah. I never liked python's "ternary", given it buries the lede.

Empact:
@SomberNight updated

lib/wallet.py
```diff
@@ -1170,7 +1170,7 @@ def get_tx_status(self, tx_hash, height, conf, timestamp):
             if fee is not None:
                 size = tx.estimated_size()
                 fee_per_byte = fee / size
-                extra.append('%.1f sat/b'%(fee_per_byte))
+                extra.append(format_satoshis(fee_per_byte, precision=FEE_PRECISION) + ' sat/b')
```
SomberNight: this won't work
in `format_satoshis` you have
```
result = ("{:" + decimal_format + "f}").format(x / pow (10, decimal_point)).rstrip('0')
```
which divides `fee_per_byte` by 10**8.
So at the very least you should set `decimal_point` to `0` here.
I guess this is a sign that `format_satoshis` is a bit confusing. :/

Empact: Yeah that's interesting it's something of a formatting and conversion function in one - sat -> BTC -> mBTC on the one hand, zeroes / padding / precision on the other. Maybe that's the line to divide it along.

SomberNight:
Thanks

PR #4287, opened 2018-04-20
Proposal: Track fee rates in sat/b to avoid conversion

Empact:
This is to propose a gradual transition toward tracking fees in sat/B consistently, and provide an example of how doing so could simplify the code.

Maybe not worth the effort, but it seems useful given that is the natural scale people have gravitated to, e.g. https://jochen-hoenicke.de/queue/#1,24h


Empact:
Fair enough

PR #4308, opened 2018-04-25
exchange_rate: prefer relative imports

toxeus:
As this project is currently structured, absolute
imports have the drawback that they'll load modules
from installed paths rather then from the local repo
when running unit tests. This somehow breaks the main
idea of unit tests which is being able to quickly test
newest development changes. Therefore, use a relative
import here.


toxeus:
related: https://github.com/spesmilo/electrum/issues/4279

PR #4313, opened 2018-04-26
electrum: cleanup plugin imports

toxeus:
`electrum_plugins` is imported in the `electrum` script
without being used - I assume it's done to make PyInstaller
find that package. But this can be achieved by avoiding
the usage of `__import__()` when importing `electrum_plugins`
where it's effectively used.


lclc:
utACK.

Looks like a smart move :)

Issue #3364, opened 2017-11-24
Feature request: Add "Show on TREZOR" icon on the receive tab


ejose19:
Currently, you need to go to Addresses tab, look the address you want to receive, right click on it and then show on trezor, however it's more practical to have an icon (could be an eye) next to the copy icon to show the current receiving address in the device.

This feature makes the verification of the address more easier and security is improved by more users verifiying their addresses before giving out them.

ejose19:
Any updates regarding this?

PR #4316, opened 2018-04-27
[Qt] receive tab: show address on hw wallet

SomberNight:
closes https://github.com/spesmilo/electrum/issues/3364

Image for multisig wallet:
receive_tab_show_address2

TODOs:
- [x] replace icon
- [x] test more

note: the icon is from the icons8 pack, which is already credited


SomberNight:
Note: given this introduces a new icon, https://github.com/spesmilo/electrum-icons will need an update

PR #4320, opened 2018-04-28
Remove cryptohead.de

schildbach:
The owner of the domain asked me to remove it.

### DIFF
--- a/lib/servers.json
+++ b/lib/servers.json
@@ -74,11 +74,6 @@
         "s": "995",
         "version": "1.2"
     },
-    "cryptohead.de": {
-        "pruning": "-",
-        "s": "50002",
-        "version": "1.2"
-    },
     "daedalus.bauerj.eu": {
         "pruning": "-",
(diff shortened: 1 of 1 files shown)


SomberNight:
This is a generated file, so if we remove the server manually, it might get added back when the file is re-generated.
ElectrumX has a peer discovery protocol: servers announce themselves to each other, and they keep a list of their peers. Servers can be queried for their peers, and this is how this file is generated:
https://github.com/spesmilo/electrum/blob/40e9e8bd632bd3d600d8913788221d60a3801cc1/scripts/servers#L8-L9
the script queries one of the servers for their list of peers.

Server operators can disable announcing themselves to other servers:
https://electrumx.readthedocs.io/en/latest/environment.html#envvar-PEER_ANNOUNCE
https://github.com/kyuupichan/electrumx/blob/90f28314d2239961b4cc7728bf6416d4aedd4adf/docs/environment.rst#peer-discovery

PR #4188, opened 2018-03-27
WIP: use coincurve if available instead of python-ecdsa

SomberNight:
closes https://github.com/spesmilo/electrum/issues/3926

(should be mostly OK already; but only pushed to see if tests pass with travis too)

PR #4358, opened 2018-05-12
coincurve take2. crypto refactoring.

SomberNight:
- use coincurve if available instead of python-ecdsa
- move ecc crypto to ecc.py, abstract away underlying module (python-ecdsa/coincurve) from rest of the code
- move symmetric crypto and hash functions to crypto.py

this supersedes https://github.com/spesmilo/electrum/pull/4188


SomberNight:
After this gets merged, we will need to rebase lightning, and start using the new functions/classes there too.

ofek:
Question: does this also attempt to install Coincurve when Electrum is installed?

SomberNight:
> does this also attempt to install Coincurve when Electrum is installed?

Yes, with the current PR, coincurve is bundled for Windows and MacOS.
As for Linux, coincurve is in the `fast` setuptools extra, and I am thinking we could change the install guides (download page on electrum.org, README in this project), so that they install the `fast` extra.

It is not bundled for Android, as that is more difficult; but at some point after this is merged I plan to do that too.

python-ecdsa is slow, and in fact it is too slow for lightning, so we need this.

ofek:
This is so, so awesome!

ecdsa:
I would rather not use coincurve, because it brings too many dependencies.
A better approach would be to use ctypes; see for example https://github.com/richardkiss/pycoin
maybe we could replace python-ecdsa and some of our low-level code with pycoin; adding @richardkiss to conversation.

ofek:
@ecdsa I'm very sorry to hear that. That is quite surprising to me though as there are only 2 dependencies:

- `asn1crypto` - Which is pure Python (and is only used for optional DER serialization).
- `cffi` - Python's new standard to interface with C which, like coincurve itself, also has binary wheels for [every platform](https://pypi.org/project/cffi/#files), is known to build fine on Android, Alpine linux, etc., fully supports PyPy (recommended actually), and is faster than ctypes.

Also, as Richard (hello again man!) will tell you, the [experimental code](https://github.com/richardkiss/pycoin#too-slow) to call 2 OpenSSL functions requires users to first set an environment variable and based on testing it [doesn't work on Windows](https://github.com/ofek/bit/issues/4#issuecomment-288273686). Please in mind as well that coincurve/libsecp256k1 in still an order of magnitude faster than OpenSSL. I can provide benchmarks if you wish!

ecdsa:
adding coincurve means adding 4 dependencies, not 2

I was not suggesting to use openssl, but libsecp256k1, with ctypes.
I see two ways of doing that: 
1. importing pycoin. since pycoin has a native python of ecdsa, we could remove our python-ecdsa dependency.
2. using ctypes as in https://github.com/richardkiss/pycoin/blob/master/pycoin/ecdsa/native/secp256k1.py

SomberNight:
@ofek We are considering if it would be better to use pycoin. We know that coincurve has precompiled dll/so files.

If we used pycoin:
- as it has a pure python ecdsa implementation as well, we can remove `python-ecdsa` as a dependency
- we would need to compile (and bundle) `libsecp256k1`
    - for Windows (in the Wine script?)
    - in the Mac build script
    - for Android, cross-compile with buildozer
- not clear how Linux users could get the .so file, perhaps they would need to manually compile the lib
- [libsecp256k1 usage](https://github.com/richardkiss/pycoin/blob/01b1787ed902df23f99a55deb00d8cd076a906fe/pycoin/ecdsa/native/secp256k1.py) is not actually in the latest release of `pycoin`, only on master

If we used coincurve:
- introduces new dependencies: `cffi`, `pycparser` (from `cffi`), `asn1crypto`, as well as `coincurve` itself of course 
_// maybe `asn1crypto` could be made an optional dependency of `coincurve` btw_
- `libsecp256k1` is already precompiled, which makes it convenient, but it also means we would need to trust these binary blobs
- we would still need to put some work into Android, re cross-compiling with buildozer

@bauerj How feasible do you think it is for us to compile `libsecp256k1` in the Wine and MacOS build scripts? Wine would probably need MSVC..

@richardkiss Any idea when a new pycoin release from master would come?
Also, [I saw in setup.py](https://github.com/richardkiss/pycoin/blob/d98f70ff67c836d180032cb6f7a7491a81cfd2eb/setup.py#L55) that the development status is "alpha". Would you say the ecdsa submodule is production-ready?

ofek:
Is the main issue that you want coincurve to have a pure Python ECDSA fallback option?

richardkiss:
The master trunk is in a bit of a mess right now as I work toward altcoin support in a less hacky way. That said, the ecdsa submodule is quite mature and in decent shape, and uses OpenSSL and libsecp256k1 for speed-ups if they are available. The code that implements these optimizations is a little gross, but the interface should remain stable. If necessary, I can tag a check-in on the master branch as 0.81a.

ofek:
I compiled a (probably incomplete) list of downstream users: [https://github.com/ofek/coincurve#users](https://github.com/ofek/coincurve#users)

It may be worthwhile to get feedback from the Ethereum folks on how they are liking Coincurve for use in their client/app.

SomberNight:
> Is the main issue that you want coincurve to have a pure Python ECDSA fallback option?

- cffi is a comparately large dependency that is not strictly needed; ctypes is already in the stdlib
- asn1crypto is again a relatively large dependency that we would not even use at all
- the setup.py file of coincurve is.. daunting :)
- also I, personally, would like it more if we either built libsecp256k1 ourselves, or used a binary provided e.g. by a linux distribution (btw ubuntu [17.10](https://packages.ubuntu.com/artful/libsecp256k1-0) and [18.04](https://packages.ubuntu.com/bionic/libsecp256k1-0), and [debian9](https://packages.debian.org/stretch/libsecp256k1-0) have binaries).
- turns out we actually need to keep using python-ecdsa anyway, as we do EC operations on other curves for dnssec validation

so my current idea is to use libsecp256k1 directly through ctypes, if available;
build and bundle the dll/so for Win/Mac, so it would be available there;
and not do anything with it on Linux for now, hope that more distributions will have it (but an Ubuntu LTS version having it is already great)

richardkiss:
For the record, pycoin supports other EC curves, and uses OpenSSL to speed up some operations; see https://github.com/richardkiss/pycoin/blob/master/pycoin/ecdsa/secp256r1.py for an example.

PR #4371, opened 2018-05-17
Remove explicit protocol mentions

haarts:
I'm endeavoring to understand the code base. Ultimately I would like to see if the backend (electrumX) could be loosely coupled with the rest.

In order to achieve this I'm hoping to push all explicit mentions of the protocol strings into the network module. This PR is a tiny step into that direction.

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -647,8 +647,7 @@ class Commands:
                 util.print_error('Got Response for %s' % address)
             except BaseException as e:
                 util.print_error(str(e))
-        h = self.network.addr_to_scripthash(address)
-        self.network.send([('blockchain.scripthash.subscribe', [h])], callback)
+        self.network.subscribe_to_addresses([addr], callback)
         return True
 
     @command('wn')
--- a/lib/network.py
+++ b/lib/network.py
@@ -625,7 +625,7 @@ class Network(util.DaemonThread):
             # Response is now in canonical form
             self.process_response(interface, response, callbacks)
 
-    def overload_cb(self, callback):
+    def map_scripthash_to_address(self, callback):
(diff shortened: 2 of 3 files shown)


lib/commands.py
```diff
@@ -647,8 +647,7 @@ def callback(x):
                 util.print_error('Got Response for %s' % address)
             except BaseException as e:
                 util.print_error(str(e))
-        h = self.network.addr_to_scripthash(address)
-        self.network.send([('blockchain.scripthash.subscribe', [h])], callback)
+        self.network.subscribe_to_addresses([addr], callback)
```
SomberNight: should be `address` not `addr`

lib/network.py
```diff
@@ -641,13 +635,15 @@ def cb2(x):
         return cb2
 
     def subscribe_to_addresses(self, addresses, callback):
-        hashes = [self.addr_to_scripthash(addr) for addr in addresses]
-        msgs = [('blockchain.scripthash.subscribe', [x]) for x in hashes]
-        self.send(msgs, self.overload_cb(callback))
+        hash2address = {bitcoin.address_to_scripthash(address): address for address in addresses}
+        self.h2addr.update(hash2address)
+        msgs = [('blockchain.scripthash.subscribe', [x]) for x in hash2address.keys()]
```
SomberNight: note: as a result of this, the subscriptions for `addresses` will be in a different than before order (sorted) on the wire, but I believe this is OK.

lib/commands.py
haarts: will fix today

haarts: Ah. You already did. :) Thanks for catching that!

PR #4374, opened 2018-05-21
Remove explicit protocol mentions from scripts

haarts:
See the commit message for additional information!

I'd be willing to investigate if the `scripts/util.py` module can be removed in favor of the plain network handling as is done in `scripts/watch_address` for example. 

You'll be seeing more of these small PRs from me soon, I've opted to keep them small in order to give you an opportunity to comment often.

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -1083,6 +1083,18 @@ class Network(util.DaemonThread):
             return False, "error: " + out
         return True, out
 
+    def history_for_scripthash(self, hash):
+        command = 'blockchain.scripthash.get_history'
+        return (command, [hash])
+
+    def subscribe_to_headers(self):
+        command = 'blockchain.headers.subscribe'
+        return (command, [])
+
+    def subscribe_to_address(self, address):
+        command = 'blockchain.address.subscribe'
+        return (command, [address])
+
     def export_checkpoints(self, path):
         # run manually from the console to generate checkpoints
         cp = self.blockchain().get_checkpoints()
--- a/lib/network.py
+++ b/lib/network.py
@@ -1062,7 +1062,7 @@ class Network(util.DaemonThread):
     def get_local_height(self):
         return self.blockchain().height()
 
-    def synchronous_get(self, request, timeout=30):
+    def synchronous_send(self, request, timeout=30):
         q = queue.Queue()
         self.send([request], q.put)
         try:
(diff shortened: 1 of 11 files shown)


lib/network.py
```diff
@@ -1076,13 +1077,45 @@ def synchronous_get(self, request, timeout=30):
...
+
+    def subscribe_to_headers(self):
+        command = 'blockchain.headers.subscribe'
+        return (command, [])
+
+    def subscribe_to_address(self, address):
+        command = 'blockchain.address.subscribe'
+        return (command, [address])
+
+    def get_merkle_for_transaction(transaction_hash, transaction_height):
```
SomberNight: what about `self`? or maybe you want this to be a `staticmethod`?

lib/network.py
```diff
@@ -1076,13 +1077,45 @@ def synchronous_get(self, request, timeout=30):
...
+        command = 'blockchain.transaction.get_merkle'
+        return (command, [transaction_hash, transaction_height])
+
+    def subscribe_to_scripthash(scripthash):
```
SomberNight: `self` again

lib/network.py
```diff
@@ -1076,13 +1077,45 @@ def synchronous_get(self, request, timeout=30):
...
+        command = 'blockchain.scripthash.subscribe'
+        return (command, [scripthash])
+
+    def get_transaction(transaction_hash):
```
SomberNight: `self` again

lib/network.py
```diff
@@ -1076,13 +1077,45 @@ def synchronous_get(self, request, timeout=30):
...
+        command = 'blockchain.transaction.get'
+        return (command, [transaction_hash])
+
+    def listunspent_for_scripthash(scripthash):
```
SomberNight: `self` again

lib/network.py
```diff
@@ -1076,13 +1077,45 @@ def synchronous_get(self, request, timeout=30):
...
+        command = 'blockchain.scripthash.listunspent'
+        return (command, [scripthash])
+
+    def get_balance_for_scripthash(scripthash):
```
SomberNight: `self` again

lib/network.py
haarts: You are 100% right. I realized that when my work progressed and I've addressed that in #4376, https://github.com/spesmilo/electrum/pull/4376/files#diff-08f47fc4236c4931079fa1058b996bb2R1128

SomberNight:
Oh wait, so https://github.com/spesmilo/electrum/pull/4376 includes all these commits?
Then there is no reason for this PR anymore.

PR #4376, opened 2018-05-22
Remove explicit send calls

haarts:
This is a continuation of #4371.
The goal here was to make the code base easier to understand and more decoupled. All the ElectrumX protocol mentions are now in the network module. Except for two instances in the websockets module.

Functionally almost nothing changed. The only thing which changed is that in the commands module an optional `timeout` parameter could be passed. That parameter has been removed in favor of simpler code.

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -1035,17 +1035,6 @@ class Network(util.DaemonThread):
     def get_local_height(self):
         return self.blockchain().height()
 
-    def synchronous_send(self, request, timeout=30):
-        q = queue.Queue()
-        self.send([request], q.put)
-        try:
-            r = q.get(True, timeout)
-        except queue.Empty:
-            raise util.TimeoutException(_('Server did not answer'))
-        if r.get('error'):
-            raise Exception(r.get('error'))
-        return r.get('result')
-
     def __wait_for(it):
         """Wait for the result of calling lambda `it`."""
         q = queue.Queue()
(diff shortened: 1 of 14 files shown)


lib/network.py
```diff
@@ -1062,27 +1037,131 @@ def follow_chain(self, index):
...
+    def broadcast_transaction(self, transaction, callback=None):
+        command = 'blockchain.transaction.broadcast'
+        invocation = lambda c: self.send((command, [str(transaction)]), c)
+
+        if callback:
+            invocation(callback)
+
         try:
-            out = self.synchronous_get(('blockchain.transaction.broadcast', [str(tx)]), timeout)
+            out = Network.__wait_for(invocation)
```
kyuupichan: This is weird.  __wait_for is not a class method.

haarts: No need for annotations here (and if I would I'd use `@staticmethod`):

```
class Sheep():
    def sound(self):
        Sheep.__default_sound()
        
    def __default_sound():
    	print("beh")
              
Sheep().sound()  # 'beh'
Sheep.__default_sound()  # AttributeError 
```
(a bit of a silly example of cource)

haarts:
(rebased again)

haarts:
What can I do to move this PR forward?

lib/websockets.py
```diff
@@ -95,17 +95,18 @@ def run(self):
...
             method = r.get('method')
-            params = r.get('params')
+            scripthash = r.get('params')[0]
             result = r.get('result')
             if result is None:
                 continue    
             if method == 'blockchain.scripthash.subscribe':
-                self.network.send([('blockchain.scripthash.get_balance', params)], self.response_queue.put)
+                self.network.subscribe_to_scripthash(
+                        scripthash, self.response_queue.put)
```
SomberNight: This is a state machine; you missed what it is doing.
When receiving a reply to a `blockchain.scripthash.subscribe`, send a `blockchain.scripthash.get_balance` request.
Your change breaks it; it will never reach the next state.

haarts:
I just found that out too. I will push a fix Monday.

lib/websockets.py
haarts: This was annoying me. :) Pushed a fix.

lib/network.py
```diff
@@ -1060,27 +1035,131 @@ def follow_chain(self, index):
...
```
SomberNight: I'm a bit confused. How is `broadcast_transaction` supposed to behave when it is given a callback? It looks to me that `invocation` will be called twice.

haarts: You're right. It is missing a `return` statement. I'll amend it.

haarts:
I found a bug in this PR. I'll update this (or an other PR) soon.

Issue #22, opened 2012-08-26
changing font size for balance label causes gui elements to shift around


genjix:
when you enter values in the amount field it causes shifting of the elements which is ugly.

cosmetic bug.

PR #4379, opened 2018-05-25
 Fix: bug #22

Anilkumar18:
Added https://chainz.cryptoid.info/ftc/ as an option in the Preferences->Appearance section.

screenshot from 2018-05-25 11-03-47


bauerj:
This is only for Bitcoin.

Issue #3926, opened 2018-02-19
Replace python-ecdsa with faster alternative


SomberNight:
As in https://github.com/spesmilo/electrum/issues/3778#issuecomment-360976713

Potential options:
- https://github.com/ofek/coincurve (python wrapper for libsecp256k1)
- https://github.com/ludbb/secp256k1-py (python wrapper for libsecp256k1)
- https://github.com/Legrandin/pycryptodome

ofek:
I can guide anyone with the use of `coincurve` :smile:

SomberNight:
@ofek 
How should one sign and then verify a message? The signature types are incompatible..?
I would like to use `sig = PrivateKey.sign_recoverable(...)` which gives a 64 byte (+recid) sig.
Both `PublicKey.verify` and `verify_signature` expect a DER encoded signature.

I would like to use something like `cdata_to_der` but that wants a pointer, not bytes? e.g.
```
der_sig = coincurve.ecdsa.cdata_to_der(sig)
```
So currently I am using python-ecdsa as:
```
r, s = ecdsa.util.sigdecode_string(sig, SECP256k1.order)
der_sig = ecdsa.util.sigencode_der_canonize(r, s, SECP256k1.order)
```

ofek:
Sorry, I was ill this week. You want `recoverable_convert` which you can then pass to `cdata_to_der`:
https://github.com/ofek/coincurve/blob/865db90c3e8541ebe8bf20b89e3acebfd5deb4d1/coincurve/ecdsa.py#L126

SomberNight:
oh right. I was not really looking under the warning :)
https://github.com/ofek/coincurve/blob/865db90c3e8541ebe8bf20b89e3acebfd5deb4d1/coincurve/ecdsa.py#L79

Engelberg:
In 3.2.2 I'm now getting a warning when running under Tails "[ecc] warning: libsecp256k1 library not available, falling back to python-ecds".  Is this a concern, or is the fallback option perfect viable?

SomberNight:
it's only printed when running in verbose (-v) mode, right? it can be ignored; python-ecdsa is what all pre-3.2 versions of electrum used

Engelberg:
I received the warning just when invoking "python3 electrum" to run electrum, so I wasn't explicitly triggering a verbose mode.  But thanks for letting me know it can be ignored.

PR #4381, opened 2018-05-25
crypto refactoring take3

SomberNight:
closes https://github.com/spesmilo/electrum/issues/3926

- move ecc crypto to ecc.py, abstract away underlying implementation from the rest of the code
- use libsecp256k1 directly (through ctypes) to speed up some ec operations (based on [pycoin](https://github.com/richardkiss/pycoin/blob/d98f70ff67c836d180032cb6f7a7491a81cfd2eb/pycoin/ecdsa/native/secp256k1.py)); specifically, python-ecdsa is being monkey-patched when operating on the bitcoin curve
- build libsecp256k1 dll for windows (based on [coincurve](https://github.com/ofek/coincurve/blob/82b6547754fa4baf1b782cfe8da39566b995d5d8/.travis/build_windows_wheels.sh)) and bundle it in the binaries
- travis runs ecc-related tests with both python-ecdsa and monkey-patched-python-ecdsa; travis is installing libsecp256k1 from a 3rd party ppa (I don't think we need to compile it ourselves for CI)
- move symmetric crypto and hash functions to crypto.py

This is an alternative approach to https://github.com/spesmilo/electrum/pull/4358

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ jobs:
         - sudo apt-key add Release.key
         - sudo apt-add-repository https://dl.winehq.org/wine-builds/ubuntu/
         - sudo apt-get update -qq
-        - sudo apt-get install -qq winehq-stable dirmngr gnupg2 p7zip-full
+        - sudo apt-get install -qq winehq-stable dirmngr gnupg2 p7zip-full mingw-w64
       before_script: ls -lah /tmp/electrum-build
       script: ./contrib/build-wine/build.sh
       after_success: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@ language: python
 python:
     - 3.5
     - 3.6
+addons:
+  apt:
+    sources:
+      - sourceline: 'ppa:tah83/secp256k1'
+    packages:
+      - libsecp256k1-0
 install:
   - pip install -r contrib/requirements/requirements-travis.txt
 cache:
--- a/contrib/build-wine/README.md
+++ b/contrib/build-wine/README.md
@@ -14,13 +14,14 @@ Usage:
  - gpg
  - 7Zip
  - Wine (>= v2)
+ - mingw-w64
 
 
 For example:
 
 
 ```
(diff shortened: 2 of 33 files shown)


markblundeberg:
@SomberNight any interest in having a 'second fallback' to OpenSSL? It's still much much faster than the pure-python ecdsa. This would give fast EC capabilities to nearly all tarball-users (I can't imagine a platform lacking openssl) without them having to install another dependency. I am inclined to code this up but I am wondering if there are any security concerns that would make this a bad idea.

SomberNight:
@markblundeberg I think we will likely make libsecp256k1 a required dependency "soon". (mainly because lightning needs to do too many EC operations, e.g. to verify the gossip)

At some point after that, I think it would be desirable to minimise what python-ecdsa is used for, and only use libsecp256k1 (not with python-ecdsa monkey-patching but directly) for bitcoin related stuff
(atm we could not remove python-ecdsa completely because some DNSSEC stuff also uses it)

Using libsecp everywhere might make implementing Schnorr sigs easier too; we certainly cannot use python-ecdsa for that.

markblundeberg:
Re: Schnorr sigs it's actually the other way around unfortunately... with libsecp256k1 you'll have to wait for the library maintainers to add it. Users will then be obliged to upgrade to that version, so that means you probably can't rely on the system-distributed version (like in Ubuntu) and a binary will have to be provided.

In contrast, with python-ECDSA (or OpenSSL) you get direct access to curve operations so it's trivial to do the Schnorr signing/verifying operation ([example here using openssl](https://github.com/markblundeberg/Electron-Cash/blob/schnorr/lib/schnorr.py#L164), it's even more easy with python-ecdsa). With libsecp256k1, the maintainers strangely did not think it would be helpful to expose API that lets people do secp256k1 operations.

cculianu:
Interesting. I am all in favor of the openssl fallback.  And am curious if there are any security concerns related to openssl. Definitely openssl is as common as atoms of nitrogen in the atmosphere out there in the wild.. so.. it would make packaging easier.

Am curious what the verdict is. I lack the security background to assess this. I guess we can do research on it...

Issue #3779, opened 2018-01-27
Make it configurable the number of confirmations required to make Tx confirmed


Yuki-Inoue:
Currently, electrum treats a Tx to be confirmed if it is included in any Block. (Has any height).

On actual operation, one may want to treat Tx to be confirmed only after certain amount (e.g. 6) of confirmations be done.  In other words, one wants to treat a Tx to be confirmed only if that Tx height is certain amount smaller than the current block height.

So question:

* Is making it configurable the number of confirmations needed for Tx to be confirmed, is it a good idea or not?
* If it is good, could this be added as a feature request (if there's such category that is...?)

bauerj:
Hey,

what exactly do you mean? The GUI doesn't treat transactions with one confirmation as "confirmed".  Instead, it shows a series of icons based on the number of confirmations: 

![clock1](https://user-images.githubusercontent.com/598790/35470606-c7769dfe-034c-11e8-93d4-6d34b15f7339.png)![clock2](https://user-images.githubusercontent.com/598790/35470607-c792b67e-034c-11e8-8f35-bd44c8a5a2b0.png)![clock3](https://user-images.githubusercontent.com/598790/35470608-c7adda80-034c-11e8-88a1-3d92b99b96b1.png)![clock4](https://user-images.githubusercontent.com/598790/35470609-c81097ce-034c-11e8-831a-16870541dc7f.png)![clock5](https://user-images.githubusercontent.com/598790/35470611-c858c3f0-034c-11e8-82cc-98aa3aaf84d8.png)

Are you using the command line interface?

Yuki-Inoue:
@bauerj Oh yeah, sorry not giving enough details.

> Are you using the command line interface?

Yes, I'm using the CUI.  When kicking command such as `electrum balance` or `electrum payto` commands, the electrum program seems to consider a Tx is confirmed as soon as it gets included in chain (has any height information).

I thought making it configurable is somewhat common request.  My first question is that whether such request is good or not in the first place.

bauerj:
Would you prefer it to show the balance like in the GUI? (e.g. 1BTC (0.2BTC unconfirmed))

Yuki-Inoue:
@bauerj 

I'd imagine `electrum getbalance` giving json output like:

```
{
    "confirmed": "0",
    "unconfirmed": "1"
}
```

Or, in more descriptive output:

```
{
    "confirmed": "0",
    "unconfirmed": [
        {
            "confirmations": 0,
[...]

PR #4391, opened 2018-05-31
Add minimum confirmations for balance

jesobreira:
I think it would be good if users were able to decide how many confirmations they need in order to declare a transaction actually confirmed (recommended is 6, Electrum uses 1).

This also required a PR on ElectrumX server as well (kyuupichan/electrumx#482) and shall not be accepted if ElectrumX PR does not get accepted. Unfortunately I don't have enough storage for deploying and testing my PR to Electrum and therefore I wasn't able to test this as well, so please consider this as an untested idea.

Related: #3779

### DIFF
--- a/lib/commands.py
+++ b/lib/commands.py
@@ -322,7 +322,8 @@ class Commands:
         server query, results are not checked by SPV.
         """
         sh = bitcoin.address_to_scripthash(address)
-        out = self.network.synchronous_get(('blockchain.scripthash.get_balance', [sh]))
+        minconf = self.config.get('minconf', 1)
+        out = self.network.synchronous_get(('blockchain.scripthash.get_balance', [sh, int(minconf)]))
         out["confirmed"] =  str(Decimal(out["confirmed"])/COIN)
         out["unconfirmed"] =  str(Decimal(out["unconfirmed"])/COIN)
         return out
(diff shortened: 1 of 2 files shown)


lib/wallet.py
```diff
@@ -695,16 +695,17 @@ def get_addr_received(self, address):
     def get_addr_balance(self, address):
         received, sent = self.get_addr_io(address)
         c = u = x = 0
+        minconf = config.get('minconf', 1)
         local_height = self.get_local_height()
         for txo, (tx_height, v, is_cb) in received.items():
             if is_cb and tx_height + COINBASE_MATURITY > local_height:
                 x += v
-            elif tx_height > 0:
```
SomberNight: tx_height is a block height where the tx was included. 0 and -1 have special meanings signalling the tx is unconfirmed. your change won't do what you want.

lib/commands.py
```diff
@@ -322,7 +322,8 @@ def getaddressbalance(self, address):
         server query, results are not checked by SPV.
         """
         sh = bitcoin.address_to_scripthash(address)
-        out = self.network.synchronous_get(('blockchain.scripthash.get_balance', [sh]))
+        minconf = self.config.get('minconf', 1)
+        out = self.network.synchronous_get(('blockchain.scripthash.get_balance', [sh, int(minconf)]))
```
SomberNight: This change, as it affects the protocol, should probably be done separately from the rest. In most cases, the client can just filter transactions based on confirmations and calculate whatever, but not if you outsource the request to the server. :)

ecdsa:
this should not require any server side change.

Issue #3595, opened 2017-12-24
Unconfirmed transactions disappearing


schildbach:
Electrum keeps forgetting unconfirmed low-fee, but wallet-relevant transactions. This has cost me several thousand of Euros, because if Electrum forgets them (and I relied on it) I can't even do CPFP to make them confirm.

I now need to effectively save all incoming transactions to a file because I don't trust Electrum any more in keeping them. On some of those cases I tested for double spends – no one was double-spent by the sender, so they were still perfectly valid.

AbdussamadA:
It depends on servers for the data relevant to your wallet. Servers drop transactions that pay low fees. You could try switching to a different server until you find one that has your tx in its mempool. Then you can do a cpfp or rbf.

BTW shouldn't you be using your own wallet?

schildbach:
I'm quite surprised that Electrum doesn't do its own persistence. I always thought it only uses the servers for learning about new payments, not for rememering them. I think a wallet should never forgot transactions that are relevant to it. Even if transactions become invalid they should stay (but marked as invalid), so that the user can get notified. Silently deleting transaction is a terrible user experience!

I even lost an outgoing transaction (200 sat/byte CPFP), that doesn't make any sense at all to me as a user.

(I'm always using multiple different wallets in parallel.)

SomberNight:
> I'm quite surprised that Electrum doesn't do its own persistence

Electrum is persisting these transactions, but deletes them if it realizes that the server does not know about them.
https://github.com/spesmilo/electrum/blob/13bf539d175d7d97b9d55453892d380107a4fa8b/lib/wallet.py#L745-L754

fresheneesz:
I agree with schildbach, Electrum should not delete them, but rather mark them as removed from the mempool, ideally with an easy option for resubmitting manually.

ecdsa:
I agree that we could keep them in history. 
the same feature would be needed to do more than one offline transactions

barnacs:
I've had a low fee outgoing transaction sitting in the mempool for weeks and it often disappears/reappears when restarting Electrum (depending on which server it connects to, based on the above comments).

Marking transactions that are unknown to the current server in the user interface and adding the option to manually resubmit would be an option but
- marking and unmarking transactions this way between restarts could still be confusing to users without a deep enough understanding of Electrum's architecture
- as a user, I wouldn't know when to resubmit manually because there's still no distinction between a transaction that's missing from the current server and one that's been evicted from the network altogether

[...]

PR #4405, opened 2018-06-04
remove "from addresses"

SomberNight:
- remove "from addresses" dependency from wallet logic
- change partial txn format
    - partial txns now include a header consisting of 5 fixed magic bytes and 1 byte version number
- we only parse scriptSig/witness of partial txns; no more guessing of input addresses
    - this resolves the gloomy FIXME from https://github.com/spesmilo/electrum/pull/3794
- in the Qt GUI, transaction dialog, only `is_mine` input addresses are shown
- removed "`is_mine` restriction" from `wallet.spent_outpoints`
    - as a result, the local transaction feature becomes more powerful: any txn that drops from the mempool will become local. this resolves https://github.com/spesmilo/electrum/issues/3595
    - `wallet.spent_outpoints` is now persisted in the wallet file; there is a corresponding v17 version upgrade. wallet files are expected to become ~10% larger on disk

### DIFF
--- a/lib/wallet.py
+++ b/lib/wallet.py
@@ -332,7 +332,7 @@ class Abstract_Wallet(PrintError):
                     continue
                 tx = self.transactions.get(tx_hash)
                 if tx is not None:
-                    self.add_transaction(tx_hash, tx)
+                    self.add_transaction(tx_hash, tx, allow_unrelated=True)
                     save = True
         if save:
             self.save_transactions()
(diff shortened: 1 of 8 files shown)


lib/wallet.py
```diff
@@ -914,59 +907,57 @@ def add_transaction(self, tx_hash, tx):
...
-                    self.pruned_txo.pop(ser)
-                    dd = self.txi.get(next_tx, {})
-                    if dd.get(addr) is None:
-                        dd[addr] = []
-                    dd[addr].append((ser, v))
-                    self._add_tx_to_local_history(next_tx)
+                    # give v to txi that spends me
+                    next_tx = self.spent_outpoints[tx_hash].get(n)
+                    if next_tx is not None:
+                        dd = self.txi.get(next_tx, {})
```
rt121212121: get -> setdefault

PR #4408, opened 2018-06-05
Remove explicit send calls, part deux

haarts:
This is a continuation of #4376. The only thing changed is one commit fixing a bug in the original PR, I thought I could reopen the PR but I can't after I rebased the commits.  Ping @SomberNight, @kyuupichan.

From the original:
The goal here was to make the code base easier to understand and more decoupled. All the ElectrumX protocol mentions are now in the network module. Except for two instances in the websockets module.

Functionally almost nothing changed. The only thing which changed is that in the commands module an optional timeout parameter could be passed. That parameter has been removed in favor of simpler code.

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -1045,6 +1045,20 @@ class Network(util.DaemonThread):
             raise Exception(r.get('error'))
         return r.get('result')
 
+    def wait_for(it):
+        """Wait for the result of calling lambda `it`."""
+        q = queue.Queue()
+        it(q.put)
+        try:
+            result = q.get(block=True, timeout=30)
+        except queue.Empty:
+            raise util.TimeoutException(_('Server did not answer'))
+
+        if result.get('error'):
+            raise Exception(result.get('error'))
+
+        return result.get('result')
+
     def request_header(self, interface, height):
         #interface.print_error("requesting header %d" % height)
         self.queue_request('blockchain.block.get_header', [height], interface)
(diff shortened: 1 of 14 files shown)


lib/network.py
```diff
@@ -1059,27 +1034,132 @@ def follow_chain(self, index):
     def get_local_height(self):
         return self.blockchain().height()
 
-    def synchronous_get(self, request, timeout=30):
+    def __wait_for(it):
```
SomberNight: Even if it's not needed, could you make this a `@staticmethod`?

lib/network.py
```diff
@@ -1059,27 +1034,132 @@ def follow_chain(self, index):
...
-        return r.get('result')
 
-    def broadcast(self, tx, timeout=30):
-        tx_hash = tx.txid()
+        if result.get('error'):
+            raise Exception(result.get('error'))
+
+        return result.get('result')
+
+    def __with_default_synchronous_callback(invocation, callback):
```
SomberNight: same; `staticmethod`?

lib/network.py
```diff
@@ -1059,27 +1034,132 @@ def follow_chain(self, index):
...
+
+    def get_transaction(self, transaction_hash, callback=None):
+        command = 'blockchain.transaction.get'
+        invocation = lambda c: self.send([(command, [transaction_hash])], c)
+
+        return Network.__with_default_synchronous_callback(invocation, callback)
+
+    def get_transactions(self, transaction_hashes, callback=None):
+        command = 'blockchain.transaction.get'
+        messages = [(command, tx_hash) for tx_hash in transaction_hashes]
```
SomberNight: `(command, [tx_hash])`

lib/network.py
haarts: This is indeed not needed but I'll gladly conform.

lib/network.py
haarts: idem

lib/network.py
haarts: good catch.

PR #4412, opened 2018-06-06
Do not actually compute the nested p2wsh-p2sh script if just estimating size

devrandom:
Causes a 1+ second lag for medium size wallets when typing the amount in the send tab.

### DIFF
--- a/lib/transaction.py
+++ b/lib/transaction.py
@@ -825,7 +825,10 @@ class Transaction:
             scriptSig = bitcoin.p2wpkh_nested_script(pubkey)
             return push_script(scriptSig)
         elif _type == 'p2wsh-p2sh':
-            witness_script = self.get_preimage_script(txin)
+            if estimate_size:
+                witness_script = ''
+            else:
+                witness_script = self.get_preimage_script(txin)
             scriptSig = bitcoin.p2wsh_nested_script(witness_script)
             return push_script(scriptSig)
         elif _type == 'address':


SomberNight:
The lag depends on the number of inputs in the transaction, right? How many inputs do you have? What is m and n for the multisig? I'll try to reproduce later.

With this change, the tx size will be off of course.

devrandom:
I have about 5 inputs and a 2-of-3 multisig.

Seems to also be ameliorated with the recent secp256k1 native library in master, but this might be worth it anyway.

dabura667:
ws_len = len(txin[“x_pubkeys”]) * 34 + 3
witness_script = ‘x’ * ws_len

Make a dummy string that is as long as the witness script should be.

It’s slow because it’s deriving all the pubkeys from the xpubs. We know pubkeys are 33 bytes and use a single push op, so just calculate the length and make a dummy string.

devrandom:
@dabura667 - no need to have `witness_script` be any specific length.  `p2wsh_nested_script` does a sha256, so the length of `witness_script` does not affect the size of `scriptSig`.  Agree with the rest of your analysis.

dabura667:
But in order to estimate size, you need to count witness size too... witness script is pushed onto the witness stack.

I could be wrong, I haven’t kept up with the repo lately.

devrandom:
Looks to me that the witness size is estimated in `serialize_witness`, not here.  That method seems to already optimize away constructing the pubkeys.

dabura667:
ok.

Further reviewing the code, I agree that this PR should not have an adverse affect on size estimates.

concept ACK

SomberNight:
Indeed, I was wrong, and the size estimation should not be affected.
I just tested and there really is a noticeable lag on master without libsecp256k1.

PR #4416, opened 2018-06-07
trezor: do not rely on deprecated ckd_public

matejcik:
electrum has its own xpub parsing method; use that in favor of Trezor's which is now considered private and moved away

### DIFF
--- a/plugins/trezor/trezor.py
+++ b/plugins/trezor/trezor.py
@@ -3,7 +3,7 @@ import traceback
 import sys
 
 from electrum.util import bfh, bh2u, versiontuple, UserCancelled
-from electrum.bitcoin import (b58_address_to_hash160, xpub_from_pubkey,
+from electrum.bitcoin import (b58_address_to_hash160, xpub_from_pubkey, deserialize_xpub,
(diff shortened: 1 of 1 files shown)


plugins/trezor/trezor.py
```diff
@@ -268,6 +266,12 @@ def _initialize_device(self, settings, method, device_id, wizard, handler):
             client.load_device_by_xprv(item, pin, passphrase_protection,
                                        label, language)
 
+    def _make_node_path(self, xpub, address_n):
+        xtype, depth, fingerprint, child_num, chain_code, key = bitcoin.deserialize_xpub(xpub)
```
SomberNight: you need
```python
fingerprint = struct.unpack('>I', fingerprint)[0]
child_num   = struct.unpack('>I', child_num)[0]
```

SomberNight: `bitcoin` is not imported

SomberNight:
In multisig wallets that involve a trezor cosigner.

matejcik:
this seems to work then

SomberNight:
Great; thanks.

Issue #4406, opened 2018-06-04
Failure to sweep Bitcoin Core P2SH segwit private key: "No inputs found."


rapodaca:
All on testnet:

On Bitcoin Core 0.16:

- created a new default wallet
- accepted payment from the Receive tab
- obtained address by double-clicking (P2SH-segwit, starts with "2")
- obtained private key from Console with dumpprivkey <address>
- block explorer shows address has funds

On Electrum 3.1.3:

- created a new default wallet
- Wallet>Private Keys>Sweep
- pasted private key from Bitcoin Core
- error: "No inputs found. (Note that inputs need to be confirmed)"

I was expecting Electrum to sweep. Instead, it gave an error.

Electrum wallet created with import defaults to P2PKH, giving wrong address.

My interpretation:

Bitcoin Core is creating segwit-P2SH addresses by default. However, Electrum doesn't know this and is attempting to scan for P2PKH based off of private key.

Electrum finds no P2PKH outputs matching private key, so reports an error.

Is it possible to import into Electrum a Bitcoin Core 0.16.0 default private key?

SomberNight:
Prefix the WIF key with `p2wpkh-p2sh:`.

Related:
https://github.com/spesmilo/electrum/pull/3925
https://github.com/spesmilo/electrum/blob/f0daf2d37bb7043105230d32ca3b179e06b17b15/RELEASE-NOTES#L88-L91

rapodaca:
Worked - thanks!

PR #4425, opened 2018-06-10
Qt: show WIF help in import(x2) and sweep dialogs

SomberNight:
Our custom prefix types, and their existence, is word of mouth on e.g. reddit atm.

Related: https://github.com/spesmilo/electrum/issues/4406 

wif_help
wif_help2
wif_help3


bauerj:
I think it looks a bit odd having a "?" button just below the "?" button of the window.

Also: Can't we just assume that users only import keys of the same type in one batch? If so, we could add a drop-down menu to choose the type for all keys in the list.

SomberNight:
> I think it looks a bit odd having a "?" button just below the "?" button of the window.

Hmm, yeah, ok, I must have selective vision because I did not notice until you said so, but now it bothers me a bit :D

> Can't we just assume that users only import keys of the same type in one batch?

Well, on the sweep screen, that would affect the number of on-chain txns needed.
In the wizard, it would suggest that imported wallets cannot have mixed script types (only "suggest"; as after creation, the user could import other keys).

SomberNight:
Now it looks like this:
wif_help4

PR #4427, opened 2018-06-11
Extract electrumx module

haarts:
This pull request further clarifies the network module API.

A couple of major things have happened here (non of which impacts functionality):
- A class Triggers is defined and used as a mixin in the network module. This class deals with GUI notifications.
- A class ElectrumX is defined and used *instead of* the Network class. This class holds (almost) all the previously defined ElectrumX specific API calls. The methods therein are just copied over from the Network class.
- All of the helper methods in the Network class have been prefixed with an underscore as to communicate their private nature.

I would understand if these changes are considered controversial. Read with care.

### DIFF
--- a/lib/network.py
+++ b/lib/network.py
@@ -41,12 +41,10 @@ import socks
 from . import util
 from .util import print_error
 from . import bitcoin
-from .bitcoin import COIN
 from . import constants
 from .interface import Connection, Interface
 from . import blockchain
 from .version import ELECTRUM_VERSION, PROTOCOL_VERSION
-from .i18n import _
 
 
 NODES_RETRY_INTERVAL = 60
@@ -640,13 +638,13 @@ class Network(util.DaemonThread):
         elif method == 'blockchain.estimatefee':
             if error is None and result > 0:
                 i = params[0]
-                fee = int(result*COIN)
+                fee = int(result*bitcoin.COIN)
                 self.config.update_fee_estimates(i, fee)
                 self.print_error("fee_estimates[%d]" % i, fee)
                 self.notify('fee')
         elif method == 'blockchain.relayfee':
             if error is None:
-                self.relay_fee = int(result * COIN) if result is not None else None
+                self.relay_fee = int(result * bitcoin.COIN) if result is not None else None
                 self.print_error("relayfee", self.relay_fee)
         elif method == 'blockchain.block.headers':
             self.on_block_headers(interface, response)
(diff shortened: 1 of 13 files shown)


kyuupichan:
You seem to have lost some of the changes I recently made

haarts:
I've identified 4 commits of your to which you are referring:
1. https://github.com/spesmilo/electrum/commit/bc83fa8d68763afe85e904871da94d6374a033ff
1. https://github.com/spesmilo/electrum/commit/e0a6b082d2c608e644872632705338d073d382b2
1. https://github.com/spesmilo/electrum/commit/b164cc4bba25693daedbb4519ff90ef9967737ca
1. https://github.com/spesmilo/electrum/commit/1900e58e88010eef0b4b88ec13175311c50f91a9

Number 1 is addressed in https://github.com/spesmilo/electrum/pull/4427/commits/e2898808141fe913307d9a6b7234b0aa304f14cd and https://github.com/spesmilo/electrum/pull/4427/commits/071f7826d5bdf33c6600a044cb38cac973fb89c6. 
Number 2 is rebased properly afaict.
Number 3 I indeed changed slightly and I might have missed something there. I have omitted the server version call in the `scripts/block_headers` as I believe that that message has already been sent in the initialization of the network class here: https://github.com/haarts/electrum/blob/dbea86f39da57fad4f7677b25635ffa94b445682/lib/network.py#L697
Number 4 I did miss passing `True` once here: https://github.com/spesmilo/electrum/pull/4427/files#diff-cb3587f39324c05e12dc545e531b2300R120. I will fix that.

I might very well be missing other commits or patches. I appreciate the feedback!

kyuupichan:
Grepping for get_chunk still has hits.  There are none in master.

Otherwise looks good

lib/network.py
```diff
@@ -525,7 +525,7 @@ def _add_recent_server(self, server):
 
     def _process_response(self, interface, response, callbacks):
         if self.debug:
-            self.print_error("<--", response)
+            self.print_error(interface.host, "<--", response)
```
SomberNight: did you test this? I think it's redundant, it happens already.
see https://github.com/spesmilo/electrum/blob/12c5474cf126f8bcb01a5cf90b582189f1297cb0/lib/util.py#L172

lib/synchronizer.py
```diff
@@ -120,7 +120,7 @@ def on_address_history(self, response):
         tx_fees = dict(filter(lambda x:x[1] is not None, tx_fees))
         # Note if the server hasn't been patched to sort the items properly
         if hist != sorted(hist, key=lambda x:x[1]):
-            self.network.interface.print_error("serving improperly sorted address histories")
+            self.print_error("error: serving improperly sorted address histories")
```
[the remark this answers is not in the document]
haarts: Yeah... I know what I did there. The reason I did this was so I could get rid of the last remaining 'interface' reference. I deemed the upside bigger than the downside.

lib/verifier.py
```diff
@@ -38,29 +38,31 @@ def run(self):
...
-            if (tx_height > 0) and (tx_height <= lh):
-                header = blockchain.read_header(tx_height)
-                if header is None:
-                    index = tx_height // 2016
-                    if index < len(blockchain.checkpoints):
-                        self.network.request_chunk(interface, index)
-                else:
-                    if (tx_hash not in self.requested_merkle
-                            and tx_hash not in self.merkle_roots):
+            if tx_height == 0 or tx_height > local_height:
```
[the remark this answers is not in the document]
haarts: Ah, that is surprising. Thanks for pointing that out. Will fix.

haarts:
What needs to be done to get this branch merged?

ecdsa:
@haarts I did not have time to look at it. 
note that I also plan to update the network code to use asyncio, so I will look at it at that time.

haarts:
This PR can be seen as a prelude to converting the network code to use asyncio. 
As a matter of fact, the code I'm currently work on is yet an other step into that direction. I have kept the network module a thread but only to contain the changes to the network module (otherwise much more intensive refactoring is required). This again is a good step in the direction of making it asyncio compatible.

It would help me a lot if this PR got merged, otherwise my works diverges more and more from master. You can follow my progress on my own branch: https://github.com/haarts/electrum/tree/add-pylibbitcoin

SomberNight:
Given you are subclassing Network to do e-x specific stuff, I am guessing you won't allow libbitcoin/e-x heterogeneous networks. What is your plan for libbitcoin? Will the client connect to multiple libbitcoin instances like with e-x currently, or will it only connect to one which we assume to be trusted?

haarts:
Yes, you are correct. I currently do not allow for hetrogeneous networks. Doing the sub classing as implemented however makes it easy to switch implementations. Either `network = ElectrumX()` or `network = Libbitcoin()`. Both implementations have the same interface so the Electrum code can use it as a drop in replacement. That's why spent so much time reducing the API of the Network class. This makes the implementation of the Libbitcoin class doable.

The Libbitcoin implementation supports multiple backends and multiple chains.

SomberNight:
I guess you probably already thought this through but is it easier to modify the client to support both backends (e-x and libbitcoin) than, say having some middleware between the electrum client and libbitcoin translate the electrum protocol calls into libbitcoin rpc (or similar)?

lib/websockets.py
```diff
@@ -69,22 +72,27 @@ def make_request(self, request_id):
         d = json.loads(s)
         addr = d.get('address')
         amount = d.get('amount')
-        return addr, amount
+        scripthash = bitcoin.address_to_scripthash(addr)
+        hash2address[scripthash] = address
+        return scripthash, amount
```
SomberNight: `bitcoin` is not imported
instead of `hash2address`, you want `self.hash2address`
`addr` instead of `address`

haarts:
As to your middleware question; the thought occurred to me. 
And in a way this is what is happening in my next PR, see [protocol.py](https://github.com/haarts/electrum/blob/add-pylibbitcoin/lib/libbitcoin/protocol.py) on my development branch.
I feel that which ever road I would have taken this would still have lead to the changes I've done up until this point. All I've done so far is refactor. All with a dual goal; understanding the code base and clearly defining an API. As a pleasant side effect I'm writing tests and gently polish the code somewhat. 
Does this make sense?

lib/websockets.py
```diff
@@ -34,6 +34,7 @@
     sys.exit("install SimpleWebSocketServer")
 
 from . import util
+from lib import bitcoin
```
SomberNight: use a relative import; especially as the package structure is different when running from source and when having the lib installed
( https://github.com/spesmilo/electrum/issues/4279 )

haarts: I did not know that. Will fix.

haarts:
I've rebased this PR again.

Would it perhaps help if I split this PR up in pieces? It has grown quite a bit.

haarts:
I can still split it up if you guys find that easier. I'll be sure to do so in the future at least. :)

haarts:
I'm closing this one in favor of smaller PR's.

Issue #62, opened 2012-12-18
 TypeError: a float is required


gllmhyt:
I'm getting an error with the git 72b4ac9  revision and  the 1.5.6 version.

``` python
Traceback (most recent call last):
  File "/usr/bin/electrum", line 254, in <module>
    gui.main(url)
  File "/usr/lib/python2.7/site-packages/electrum/gui.py", line 1247, in main
    ew = ElectrumWindow(self.wallet, self.config)
  File "/usr/lib/python2.7/site-packages/electrum/gui.py", line 535, in __init__
    self.create_history_tab()
  File "/usr/lib/python2.7/site-packages/electrum/gui.py", line 973, in create_history_tab
    self.update_history_tab()
  File "/usr/lib/python2.7/site-packages/electrum/gui.py", line 1178, in update_history_tab
    details = self.wallet.get_tx_details(tx_hash)
  File "/usr/lib/python2.7/site-packages/electrum/wallet.py", line 436, in get_tx_details
    time_str = datetime.datetime.fromtimestamp(timestamp).isoformat(' ')[:-3]
TypeError: a float is required
```

PR #4428, opened 2018-06-11
Added a validation for Blank message field

Anilkumar18:
Included a validation for empty message field in Encrypt/Decrypt message window. Please refer the bug [#62](https://github.com/Feathercoin-Foundation/electrum-ftc/issues/62) for detailed description


SomberNight:
What if I would like to encrypt the empty string? Why do you want to make this not possible?

PR #4429, opened 2018-06-11
Added a meaningful message on invalid public key

Anilkumar18:
Error message is not in understandable format on entering a invalid/blank public key in Encrypt/decrypt window. Please refer the bug [#64](https://github.com/Feathercoin-Foundation/electrum-ftc/issues/64) for description/steps

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -2248,7 +2248,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
             encrypted_e.setText(encrypted.decode('ascii'))
         except BaseException as e:
             traceback.print_exc(file=sys.stdout)
-            self.show_warning(str(e))
+            self.show_warning("Invalid Public key")
 
     def encrypt_message(self, address=''):
         d = WindowModalDialog(self, _('Encrypt/decrypt Message'))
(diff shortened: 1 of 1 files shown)


gui/qt/main_window.py
```diff
@@ -2248,7 +2248,7 @@ def do_encrypt(self, message_e, pubkey_e, encrypted_e):
             encrypted_e.setText(encrypted.decode('ascii'))
         except BaseException as e:
             traceback.print_exc(file=sys.stdout)
-            self.show_warning(str(e))
+            self.show_warning("Invalid Public key")
```
SomberNight: Let's allow translators for this string to be translated:
`self.show_warning(_('Invalid Public key'))`

gui/qt/main_window.py
```diff
@@ -2248,7 +2248,7 @@ def do_encrypt(self, message_e, pubkey_e, encrypted_e):
             encrypted_e.setText(encrypted.decode('ascii'))
```
SomberNight: I think only the first line (`public_key = ...`) actually needs to be protected against `BaseException`. Given you are now showing a warning that specifically says the public key is invalid, let's move the other two lines out of the try, and `return` in the `except` branch after showing the warning (or use an `else` branch).

Anilkumar18: Yes. Altered the code.

Issue #60, opened 2012-12-14
TextUI: Unability to type some chars


slush0:
I cannot enter special chars (!@#$%^&*()) in the text UI, especially to password field when sending transaction.

Latest git version, Ubuntu 12.10, gnome terminal.

PR #4430, opened 2018-06-11
Empty Message Field validation added in Sign/Verify

Anilkumar18:
Signature gets generated with Blank Message field in sign/verify window. Please see the bug [#60](https://github.com/Feathercoin-Foundation/electrum-ftc/issues/60) for steps


SomberNight:
As in https://github.com/spesmilo/electrum/pull/4428#issuecomment-396245088
What if I would like to sign the empty string? Why do you want to make this not possible?

ecdsa:
signing empty string might be needed

Issue #40, opened 2012-09-24
Backup button, as in satoshi client


Mqrius:
The satoshi client has a button to backup the wallet file. In my opinion this would be a nice addition to electrum as well.

The usecase would be for new users getting started on Bitcoin. The idea in electrum is that one memorizes the passphrase, however, since the passphrase is never used, one will forget it over time.
A techy user will use spaced repetition to commit it to long term memory, but when I help someone getting started, I would like to have them set up securely and instantly, without having to remind them to learn their passphrase.

I would tell them to put a password on their wallet (=encrypt it), then backup the wallet and send it to themselves via mail, or save it via dropbox (or send it to me so I'll add it to my backups).

PR #4432, opened 2018-06-11
Removed the Max button disable code in reset_max

Anilkumar18:
Max button gets disabled on editing the requested amount code. Please refer the bug [#40](https://github.com/Feathercoin-Foundation/electrum-ftc/issues/40)

### DIFF
--- a/gui/qt/main_window.py
+++ b/gui/qt/main_window.py
@@ -1206,7 +1206,6 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, PrintError):
 
         def reset_max(t):
             self.is_max = False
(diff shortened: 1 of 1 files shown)


SomberNight:
I think the current behaviour is the intended one, at least based on  https://github.com/spesmilo/electrum/commit/8cc8f8a5e1fe2313163de53a0c4590c4a1b250a9
If you want to click max "again", you can press the `Clear` button, or manually clear the amount field.

ecdsa:
the current behavious is intended; the goal is to prevents users from accidentally sending all their coins after typing an amount.

PR #4435, opened 2018-06-11
Add SVG icons resembling old raster icons

ysangkok:
I made all SVG's default to 512x512 px but that can easily be changed with a text editor. The Viewbox will make the image contents scale to image dimensions.

The status orbs were manually created using an eye dropper and the old icons.

The sources of the preferences icon and the checkmark icon are in the SVG files themselves, they are licensed permissively.


SomberNight:
`status_waiting.png` is cropped

PR #4436, opened 2018-06-12
check SPV proof inner nodes not to be valid transactions

SomberNight:
This attempts to implement protection against the attack recently described [on the mailing list](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-June/016105.html).

Before merging, I would like to measure the performance impact with a relatively large wallet on mainnet.
So far, I could only measure on testnet, but blocks are mostly empty there, so merkle proofs are short.

On testnet, I synced a wallet (A) that has ~350 txns.
Irrespective of this PR, the sync took ~10 seconds.
Without the patch, the verifier spent about `19 ms` altogether in `hash_merkle_root`.
With the patch, the verifier spent about `66 ms` altogether in `hash_merkle_root`.

On testnet, I synced a wallet (B) that has ~10.9k txns.
The sync took ~3.5 minutes. (with the patch, it was about 4 seconds longer, but that is within error probably; I only measured this once)
Without the patch, the verifier spent about `733 ms` altogether in `hash_merkle_root`.
With the patch, the verifier spent about `2014 ms` altogether in `hash_merkle_root`.

### DIFF
--- a/lib/verifier.py
+++ b/lib/verifier.py
@@ -20,8 +20,12 @@
 # ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-from .util import ThreadJob
+from .util import ThreadJob, bh2u
 from .bitcoin import Hash, hash_decode, hash_encode
+from .transaction import Transaction
+
+
+class InnerNodeOfSpvProofIsValidTx(Exception): pass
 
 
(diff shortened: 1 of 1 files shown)


SomberNight:
I've found this address on mainnet: 1NcXyC1m3XE1s3u9rBZawt3sCZkeGEgoH9,
which has made 9733 transactions, all in the last 6 weeks.

When syncing a wallet for the above address, I benchmarked the verifier:
```
Without the patch, the verifier spent about `647 ms` altogether in `hash_merkle_root`.
With the patch, the verifier spent about `1946 ms` altogether in `hash_merkle_root`.
```

PR #4440, opened 2018-06-15
Add oxt.me as a Online Block Explorer

Rayserzor:
https://oxt.me/ 

Oxt.me have support to bech32 transactions.


lakraa51 (commented):
No think much

PR #4367, opened 2018-05-15
Improving Imports

fjhheras:
I am trying to understand the code. As part of my effort, I have improved the module imports, making them more explicit.

PR #4448, opened 2018-06-19
Improving imports

fjhheras:
Removing some wildcard imports.

Superseeds merge request #4367, that is now too outdated

### DIFF
--- a/lib/blockchain.py
+++ b/lib/blockchain.py
@@ -24,9 +24,9 @@ import os
 import threading
 
 from . import util
-from . import bitcoin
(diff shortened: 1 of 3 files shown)


bauerj:
Hey,

there are currently a lot of changes in our lightning branch and we merge from master regularly so I don't think this is the best time to do this refactoring.

SomberNight:
blockchain.py and verifier.py should be mostly independent of the lightning branch changes
and the removing the "import *" from network.py should be mostly fine too
Overall this is a small PR, it should be fine I think.

PR #4463, opened 2018-06-26
Daemon: Fix preflight CORS request handling

maschwanden:
Problem:
When Electrum is started in daemon mode it calls an authenticate method for every incoming request. However, when a POST request from another domain is conducted, the actual POST request is always preceded by an OPTIONS request (Cross-Origin Resource Sharing; CORS, see https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS). This OPTIONS request should/can not contain any credential information. Thus, in the current state the Electrum daemon is not accessible from other domains.

Fix:
Circumvent the authentication method when the request method is "OPTIONS".

### DIFF
--- a/lib/jsonrpc.py
+++ b/lib/jsonrpc.py
@@ -58,6 +58,9 @@ class VerifyingJSONRPCServer(SimpleJSONRPCServer):
                 # first, call the original implementation which returns
                 # True if all OK so far
                 if SimpleJSONRPCRequestHandler.parse_request(myself):
+                    # Do not authenticate OPTIONS-requests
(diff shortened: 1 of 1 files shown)


SomberNight:
Looks okay.
[This post](https://stackoverflow.com/a/15734032/7499128) also says "OPTIONS" requests should be exempt from authentication.
Have you tested this?

maschwanden:
Yes I tried it out. With this fix the authentication for OPTIONS requests is circumvented and, if implemented, do_OPTIONS() is called without any problems. Before the fix do_OPTIONS() is never called due to an authentication problem.

PR #4473, opened 2018-06-29
ln: enable adding peer in gui

ysangkok:
do not add_peer in open_channel, but add_peer in gui when opening channel to unknown peer. This does remove the automatic add_peer in open_channel, but that doesn't belong there. This means that there is no longer any add_peer in the CLI, but this can be added separately.


ecdsa:
This will create two classes of peers: peers that are run in aiosafe, and peers that are not.

ecdsa:
done in c4bb185f9857bc50fce6119d1f3bf37b5952c617

PR #4485, opened 2018-07-01
Expand Github issue template

lclc:
This will hopefully help lowering the amount of new issues and improve their quality.


ecdsa:
I think it will decrease their readability too. 
issues will all have the same section titles, whether or not the user actually writes useful information in them

Issue #3844, opened 2018-02-05
Mark 2fa fee outputs


AbdussamadA:
Users [are often](https://www.reddit.com/r/BitcoinBeginners/comments/7vchn7/electrum_tx_between_non_segwit_to_segwit_adress/) confused by 2fa fee outputs. They have no way of identifying the address as trusted coin's. It would be better if the output was marked as being for 2fa fees in the transaction details on the history tab.

Looking [at the code](https://github.com/spesmilo/electrum/blob/master/plugins/trustedcoin/trustedcoin.py#L308) I see the addresses are deterministically generated so Electrum knows what past 2fa fee addresses are and can identify them as such to the user.

PR #4490, opened 2018-07-02
colour 2fa billing outputs in Qt tx dialog

SomberNight:
resolves https://github.com/spesmilo/electrum/issues/3844

tx_dialog_color_2fa

Billing addresses are persisted in the wallet file.
When receiving a new billing index from TrustedCoin, which is a counter from 0 up, we generate all missing billing addresses up to that index (to cover the case of restoring from seed).

This information is used to colour the output addresses in the Qt tx dialog; and also to make sure when RBF bumping a transaction the billing output is not decreased.

Note that when restoring from seed and disabling 2fa, i.e. keeping two xprvs in the wallet, TrustedCoin is not contacted for billing info, and hence the billing addresses will not be calculated, and e.g. the colouring will not work.


bauerj:
Maybe it would be even clearer if the description would be added to the same line instead of a tooltip? There should be enough horizontal space in that text edit.

Issue #4508, opened 2018-07-06
Docker: installation of wine fails


toxeus:
[This](https://github.com/spesmilo/electrum/blob/27fc715182bf816033056a97e2e2a6d27d463f76/contrib/build-wine/docker/Dockerfile#L18) line fails with
```txt
E: Unable to correct problems, you have held broken packages.
```
By removing the `-qq` flag from `apt-get install` the following error message can be revealed
```txt
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 winehq-stable : Depends: wine-stable (= 3.0.1~bionic)
E: Unable to correct problems, you have held broken packages.
```
I'm assuming that the maintainer of the wine deb repository did break something when he updated it recently.

The installation is successful if the version specifier for the wine package is removed.

toxeus:
Also, see [travis](https://travis-ci.org/spesmilo/electrum/jobs/400540086).

toxeus:
Agreed. However, I really like the new docker approach! :+1: 

Do you think that the version of wine needs to be pinned down too? Maybe this package is not relevant for reproducibility.

SomberNight:
Well, I am not sure what needs to be pinned down, so I just tried to pin down all the top level packages we install. I was also thinking of pinning down the whole tree but did not do that yet; also not sure what the best way would be.
atm I am trying to pin down some of the dependencies of wine, I think that will fix this issue at least

PR #4509, opened 2018-07-06
Dockerfile: don't suppress error message

toxeus:
The `-qq` flag will suppress helpful output in case of an error.

Related: https://github.com/spesmilo/electrum/issues/4508


toxeus:
Compare the error output on travis from [this PR](https://travis-ci.org/spesmilo/electrum/jobs/400947694) with [this one](https://travis-ci.org/spesmilo/electrum/jobs/400540086) without my patch.

PR #4510, opened 2018-07-06
rename plugins package to electrum_plugins

toxeus:
The advantages are:

* local and installed packages are imported the [same way](https://github.com/spesmilo/electrum/pull/4510/files#diff-8b09c684b41764f7f9aa0150d4ecf441L49)
* unit tests do test modified code in [local repo](https://github.com/spesmilo/electrum/pull/4510/files#diff-5ba89e7dd9b4904f5b8eb3f22e771cbfL14)
* absolute imports instead of relative imports [can](https://github.com/spesmilo/electrum/pull/4427#discussion_r198757088) be used

I have opened this PR to illustrate my proposal in https://github.com/spesmilo/electrum/issues/4279 because it was easy to do after https://github.com/spesmilo/electrum/pull/4313 got merged. If this PR gets merged I'll create follow-up PRs where the other packages are renamed as well to keep the package naming consistent.


toxeus:
@ysangkok wrote:
> I would prefer if the restructuring was one big PR, because it would be unfortunate to move all these files if it turns out later that it doesn't enable us to run tests without installing.

I don't like big PRs. It makes them hard to review and hard to make correct. The changes to the packages are orthogonal. Python isn't black magic. It'll "turn out" as described in the documentation. If not then we have found out a serious bug :wink: (in this particular case I know for a fact that it works as expected).

> How is "import electrum" going to work without installing if "lib" is not renamed to "electrum"?

See above where I wrote: "If this PR gets merged I'll create follow-up PRs where the other packages are renamed as well to keep the package naming consistent."

> I don't see how you can (with this approach not renaming lib) be able to run the tests without installing. Which is a major reason for doing this reachitecturing!

Again, read my comment above. I'm not going to invest too much time into a change if I'm not sure that the project maintainers will merge it. More explicit acknowledgement or refusal of my proposal in the [issue](https://github.com/spesmilo/electrum/issues/4279) would have certainly helped to make things more clear to me. Therefore, I'd chosen to implement the quickest and simplest renaming change of the three ones. It already clearly demonstrates that we don't need to differentiate between a local and a installed case anymore.

> This is what we currently have, and I'd like to see a restructuring fix this (as my proposal does, which you didn't comment on :P):

Sorry, I was a bit overwhelmed by your branch. I have now commented on it in the issue. Also, I'm aware that `test_wallet_vertical` fails without an installation because [I was the one to report it](https://github.com/spesmilo/electrum/issues/4279#issue-315779304). This PR doesn't brake anything and doesn't fix anything but it *illustrates* what my proposal was. Again, if you read my comment above then you can see that my plan is to also rename the `lib` package.

ysangkok:
Ok, your explanation makes sense to me, and I'd be all for it. I'll discuss this with Thomas next week and we'll get back to you.

toxeus:
Great, thanks!

ysangkok:
We wanted to put have gui and plugins in electrum, so we have merged a large restructuring instead, it should have the advantages you list as reasons for this PR.

toxeus:
No worries, I do understand that different engineers have different approaches to software engineering :+1:  But looking at all the follow-up commits fixing your big commit confirms that it's hard to make a big commit correctly, as I said above :wink:

PR #4521, opened 2018-07-09
docker: simplify README.md

toxeus:


SomberNight:
Can you add a note what the current working directory should be when executing the commands?

Also, I guess there's no point in setting the context to `.` -- I did not notice originally, but now `.` is over 1 GB for me, so it takes time to compress it :P
However, if I simply change the command to
```
sudo docker build --no-cache -t electrum-wine-builder-img -f contrib/build-wine/docker/Dockerfile -
```
then it hangs for me indefinitely(?).
This however works:
```
sudo docker build --no-cache -t electrum-wine-builder-img - < contrib/build-wine/docker/Dockerfile
```

toxeus:
> Can you add a note what the current working directory should be when executing the commands?

Sure.

> Also, I guess there's no point in setting the context to . -- I did not notice originally, but now . is over 1 GB for me, so it takes time to compress it :P

Good catch! I wasn't aware that the context is also pulled into the image. My understanding was that only files that are added using `COPY` or `ADD` would end up in the image. I suggest the following command then
```sh
sudo docker build --no-cache -t electrum-wine-builder-img contrib/build-wine/docker
```
because having the Dockerfile in the context enables docker's caching which can sometimes be useful.

toxeus:
@SomberNight I missed that. I'm preparing a PR that will change that line anyway. Therefore, you don't need to bother fixing it.

PR #4524, opened 2018-07-09
build-wine: allow local testing

toxeus:
Before, it was only possible to test commits that are
on Github (pull request or merged). Now, changes
can be tested locally too.

This introduces the risk that a release could be built
containing uncommitted changes which by definition breaks
deterministic builds. Fortunately, this will always be
detected because the version string is created using
[`git describe --tags --dirty`](https://github.com/spesmilo/electrum/blob/dc6fb04ffe9ddd4243ca5b001edc57ee8305783e/contrib/build-wine/build-electrum-git.sh#L44).


SomberNight:
Nice. :)
So are you setting `$TARGET` to an "empty string" to build the local dir?

toxeus:
@SomberNight exactly.

toxeus:
Good point. Moreover, this change actually obsoletes the need for the `$TARGET` variable because the commit, that needs to be built, can just be checked out. I'd remove the relevant code. What do you think?

SomberNight:
This is causing several problems during the release process at the moment.

1. The binaries are being marked dirty as a consequence of the locale being built at https://github.com/spesmilo/electrum/blob/bc35c82619c86dcef8b8e74bb5118b4afae59324/contrib/build-wine/build-electrum-git.sh#L36-L40
We had to add a gitignore to the locale submodule in https://github.com/spesmilo/electrum/commit/bc35c82619c86dcef8b8e74bb5118b4afae59324

2. Some of the untracked files in the repository, not sure which, on Thomas' machine were making the hash not reproducible. We needed to run `git clean -fxd` to get the same hashes as Travis.

3. The build fails on this line https://github.com/spesmilo/electrum/blob/bc35c82619c86dcef8b8e74bb5118b4afae59324/contrib/build-wine/build-electrum-git.sh#L43
when it is trying to recursively touch all files.
Previously this was done in a clean clone, but with this PR, there can be anything here (e.g. untracked files).
Specifically, it is failing on some broken symlink in `.buildozer`, which are present if one has built the Android binary.

	```
	/opt/wine64/drive_c/electrum /opt/wine64/drive_c/electrum/contrib/build-wine/tmp
	Last commit: 3.2.3-dirty
	/opt/wine64/drive_c/electrum/contrib/deterministic-build/electrum-locale /opt/wine64/drive_c/electrum /opt/wine64/drive_c/electrum/contrib/build-wine/tmp
	/opt/wine64/drive_c/electrum /opt/wine64/drive_c/electrum/contrib/build-wine/tmp
	touch: cannot touch './.buildozer/android/platform/build/build/venv/local/include': No such file or directory
	touch: cannot touch './.buildozer/android/platform/build/build/venv/local/bin': No such file or directory
	touch: cannot touch './.buildozer/android/platform/build/build/venv/local/lib': No such file or directory
	touch: cannot touch './.buildozer/android/platform/build/build/other_builds/python3crystax-version3.6/armeabi-v7a/python3crystax/Lib/site-packages/README': No such file or directory
	```

toxeus:
@SomberNight Thank you for the feedback. 

1. I have a proper fix for this. Will open PR soon.
1. This might be fixed by the other fixes for 1. and 3.
1. I have a proper fix for this. Will open PR soon. I didn't notice this because I've never built electrum for android.

toxeus:
1. https://github.com/spesmilo/electrum/pull/4685
2. let's see if needed
3. https://github.com/spesmilo/electrum/pull/4684

Issue #4519, opened 2018-07-08
"scripts" assume that 'server' in config is set


aghamir:
I tried to use [scripts](https://github.com/spesmilo/electrum/tree/master/scripts) in your repo. However, I got this error:
```
Traceback (most recent call last):
  File "estimate_fee", line 3, in <module>
    peers = util.get_peers()
  File "/home/mohammad/Desktop/Applications/electrum/scripts/util.py", line 55, in get_peers
    interfaces = get_interfaces([server])
  File "/home/mohammad/Desktop/Applications/electrum/scripts/util.py", line 18, in get_interfaces
    connecting[server] = Connection(server, socket_queue, config.path)
  File "/usr/local/lib/python3.5/dist-packages/electrum/interface.py", line 53, in Connection
    host, port, protocol = server.rsplit(':', 2)
AttributeError: 'NoneType' object has no attribute 'rsplit'
```
After debugging, I found this line returns `None`:
https://github.com/spesmilo/electrum/blob/c6ccdbef90de0a96b5354125d090fc4dea8443cd/scripts/util.py#L53
Is there something that I missed?
Thanks a lot

bauerj:
Sure, why not.

aghamir:
Hi @SomberNight ,
Thanks a lot. Your respond is very helpful for me.
I got another error while explore scripts:
```Traceback (most recent call last):
  File "estimate_fee", line 4, in <module>
    results = util.send_request(peers, 'blockchain.estimatefee', [2])
  File "/home/mohammad/Desktop/Applications/electrum/scripts/util.py", line 72, in send_request
    interfaces = get_interfaces(peers)
  File "/home/mohammad/Desktop/Applications/electrum/scripts/util.py", line 19, in get_interfaces
    connecting[server] = Connection(server, socket_queue, config.path)
  File "/usr/local/lib/python3.5/dist-packages/electrum/interface.py", line 53, in Connection
    host, port, protocol = server.rsplit(':', 2)
ValueError: not enough values to unpack (expected 3, got 1)
```
I think interface.py assume there is port and protocol in server name. However, server name is like this `ozahtqwp25chjdjd.onion`.

SomberNight:
Which script are you trying to run? and what is `server` set to in config (because that should include port and protocol like in my example)

aghamir:
I ran `estimate_fee`. I've set server like you said before:
`electrum setconfig server testnet.qtornado.com:50002:s`

PR #4526, opened 2018-07-10
Fix util.py script

aghamir:
Hi,
This is a fix to issue #4519. I don't know why, but 't' protocol generates error. Thus, I removed them.

### DIFF
--- a/scripts/util.py
+++ b/scripts/util.py
@@ -1,11 +1,25 @@
-import select, time, queue
-# import electrum
-from electrum import Connection, Interface, SimpleConfig
-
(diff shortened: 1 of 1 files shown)


ysangkok:
Thank you for trying, but don't send pull requests that include unnecessary changes like pprint everywhere! I know we have some debug output, but it should at least always be prefixed with a description of what is being output, so that it is easy to find out where it is coming from. The issue you are trying to fix was just due to a lack of a call to `filter_protocol`, I pushed a commit with that fix.

Don't get discouraged and thank you for trying to make Electrum better.

Issue #3859, opened 2018-02-06
Change addresses beyond default gap limit will stay beyond gap limit


SomberNight:
Suppose the user uses `create_new_address` to create change addresses.
https://github.com/spesmilo/electrum/blob/d56dba80392bd5e7d679ab9483b2cbfd50d4c506/lib/wallet.py#L1813
(Alternatively, the ~same can be triggered by changing `gap_limit_for_change`)

Let's say she creates 50 new change addresses.
The wallet, when creating a new txn, will choose from the last 6 (hardcoded `gap_limit_for_change`) change addresses.
https://github.com/spesmilo/electrum/blob/d56dba80392bd5e7d679ab9483b2cbfd50d4c506/lib/wallet.py#L999-L1007
When all among these last 6 are used, at some point, new ones will be generated.
https://github.com/spesmilo/electrum/blob/d56dba80392bd5e7d679ab9483b2cbfd50d4c506/lib/wallet.py#L1832-L1835

As a consequence, an existing "gap" will never be overcome.

ecdsa:
is that really a concern? 
we cannot address all the possible scenarios that result from a user typing commands in the console..

SomberNight:
Ok, fair enough. I just ran into this and it took some time to diagnose why I was missing some coins after a restore. :)

PR #4530, opened 2018-07-10
wallet: try to plug gap limit for change addresses

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/3859

Sometimes there are legitimate reasons for creating new change addresses / increasing the change gap limit. (such as verifying/identifying change addresses when offline signing)

If the user does that, current master will tend to leave a gap of unused change addresses when making new transactions, which can lead to not discovering funds when restoring. [0]
The gap limit for change is hardcoded at 6.
The gap is due to the wallet always choosing change addresses from the last 6.

This PR keeps a list of unused change addresses, and use it such that the wallet will always prefer choosing the first unused one (or more in case of the multiple change outputs preference).

-----
[0] This has been irking me for some time, as whenever a user asks for help to generate more change addresses, I have to remember to give a disclaimer of this issue.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -179,6 +179,8 @@ class Abstract_Wallet(AddressSynchronizer):
         self.fiat_value            = storage.get('fiat_value', {})
         self.receive_requests      = storage.get('payment_requests', {})
 
+        self.calc_unused_change_addresses()
+
         # save wallet type the first time
         if self.storage.get('wallet_type') is None:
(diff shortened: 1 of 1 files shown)


jlopp:
Concept ACK

We ran into gap limit issues fairly often at BitGo and this was something that I suggested years ago.

Is 6 the standard for change? I thought the standard was a gap limit of 20 for both receiving and change derivation. IIRC at BitGo we increased it to 100 on both in our recovery tool.

PR #4542, opened 2018-07-12
Add electrumx class

haarts:
This PR (once part of #4427) introduces the ElectrumX class. This class deal solely with API calls to the ElectrumX backend thus splitting responsibilities between:
- actual calls (now ElectrumX class)
- network socket maintenance (still Network class)
- trigger sending (still Network class)
- local blockchain management (still Network class)

It also holds the odd new docstring and linter fixes.

### DIFF
--- a/lib/websockets.py
+++ b/lib/websockets.py
@@ -23,7 +23,9 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import queue
-import threading, os, json
+import threading
+import os
+import json
 from collections import defaultdict
 try:
     from SimpleWebSocketServer import WebSocket, SimpleSSLWebSocketServer
@@ -34,7 +36,9 @@ except ImportError:
 from . import util
 from . import bitcoin
 
+
 request_queue = queue.Queue()
+
 
 class ElectrumWebSocket(WebSocket):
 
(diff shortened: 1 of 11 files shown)


lib/websockets.py
```diff
@@ -98,26 +107,26 @@ def run(self):
             method = r.get('method')
             result = r.get('result')
             if result is None:
-                continue    
+                continue
             if method == 'blockchain.scripthash.subscribe':
                 addr = r.get('params')[0]
```
SomberNight: As I said on IRC, I am fairly certain this will fail here now, as you no longer give the address to the network.
`r.get('params')[0]` is now a scripthash

SomberNight:
How will other subclasses of `Network` look like? Are they going to be implementing the same public methods?

lib/websockets.py
haarts: Wait. I checked that. I still have your commit open: https://github.com/spesmilo/electrum/commit/8591587f9ed9e055edd0205ab45e9412e52c0f63
The issue is that, if the method is `scripthash.subscribe` the params contain an address not a scripthash. I must have misinterpreted the commit.

haarts:
Yes. They are going to implement the exact same public methods.

lib/websockets.py
SomberNight: the issue stems from if you call `network.subscribe_to_scripthash` then you will get back a scripthash as param, and if you call `network.subscribe_to_addresses` then you will get back an address

gabriel-samfira:
I think if you are planning on abstracting away some functionality, it's worth creating a ```metaclass```, See:

haarts:
As discussed on IRC with @ecdsa we've decided to drop the ElectrumX class. Instead I'll limit myself to improvements to the Network class.

PR #4554, opened 2018-07-16
docker: capture full build context into image

toxeus:
The advantages are:

 * The build process is even more simplified to one single
   command.

 * A build using an outdated builder image is prevented
   because the steps of creating the build environment
   and building the electrum exes are now coupled/atomic.

 * The developer's local repo remains untouched.

 * The resulting image contains the full build context
   including the build artifacts. It can be archived locally
   or on Docker Hub and used for retrospective analysis for
   builds that are not yet deterministic or not yet proven
   to be deterministic.

 * Docker's caching feature can be taken advantage of. Docker
   very reliably detects when to invalidate caches given the
   full build context.

The disadvantage is:

 * The image size increases.


SomberNight:
Rebuilding the image takes several minutes, which would now be needed to be done for every build.
If one is debugging an issue with Windows binaries, builds take noticeably longer this way.
But maybe for that use case, one can use build.sh directly, without docker..

toxeus:
Just leave away the `--no-cache` flag for such a case. Also, I have one more quite effective caching optimization in the pipeline but I didn't want to intermingle it in this PR.

Issue #4540, opened 2018-07-12
Electrum's revealer plugin UI suggestions


Engelberg:
1. There should be a way to access the plugin's ability to generate revealers and encrypted seeds or arbitrary messages without displaying your seed.
2. Right now, the UI is completely ambiguous about whether the button to encrypt your seed with a revealer includes any seed extension words or not. Does it? (I'm guessing not, but I think this should be made clearer in the interface).

SomberNight:
2. Indeed, seed extension is not included.

ecdsa:
@tiagotrs

tiagotrs:
for 1. we could hook the Plugin from the main window, next to the seed icon? Then it would need to unlock the seed in case the user wants to encrypt it and it's password locked. 
About 2. I think it makes sense to encrypt only the seed. The use case of seed extensions is to have different extensions for the same seed right? I could add a warning that the extension is not included in the backup or I could add an explicit option to include it.

SomberNight:
I don't think the seed extension should be included in the text on the revealer. I consider the seed extension another "factor"/thing you need to access your coins. Adding a warning/note that it's not included is fine.

Engelberg:
Agree.  Thanks for considering.

tiagotrs:
also about 1. in the future I plan to integrate the ledger app (https://github.com/LedgerHQ/bolos-app-revealer/), so it makes sense that the revealer icon is not on the seed dialog.  Is a icon on the main window acceptable? otherwise, where should I hook it from?

PR #4555, opened 2018-07-16
add warning that seed extension will not be included in the backup 

tiagotrs:


SomberNight:
related: https://github.com/spesmilo/electrum/issues/4540

PR #4557, opened 2018-07-16
Backport from Electron Cash (macOS) -- fix missing URI types

cculianu:
After you guys upgraded to the new building scheme for macOS (which we ported over to our code) -- the Info.plist for macOS became spartan and bare.  One of the thing's it's missing is support for bitcoin: (or bitcoincash: in our case) URI types.

This adds that.

~Also I noticed your build system insists on Homebrew.. when it also works with either MacPorts or Homebrew - so I added supporting that too.~

~I personally use MacPorts and consider it pretty awesome so I try to add support for it to any project.  Since it's 0 cost for you guys to support it (it provides identical tools really) -- here it is, acknowledged as being a viable way to build.~

Tested and works on my macOS system.

### DIFF
--- a/contrib/build-osx/make_osx
+++ b/contrib/build-osx/make_osx
@@ -15,7 +15,7 @@ cd $src_dir/../..
 export PYTHONHASHSEED=22
 VERSION=`git describe --tags --dirty`
 
-which brew > /dev/null 2>&1 || fail "Please install brew from https://brew.sh/ to continue"
+which brew > /dev/null 2>&1 || which port > /dev/null 2>&1 || fail "Please install either Homebrew (from https://brew.sh/) or MacPorts (from https://www.macports.org/) to continue"
 
 info "Installing Python $PYTHON_VERSION"
 export PATH="~/.pyenv/bin:~/.pyenv/shims:~/Library/Python/3.6/bin:$PATH"
@@ -89,6 +89,12 @@ done
 
 info "Building binary"
 pyinstaller --noconfirm --ascii --clean --name $VERSION contrib/build-osx/osx.spec || fail "Could not build binary"
+
+info "Adding bitcoin URI types to Info.plist"
+plutil -insert 'CFBundleURLTypes' \
+	-xml '<array><dict> <key>CFBundleURLName</key> <string>bitcoin</string> <key>CFBundleURLSchemes</key> <array><string>bitcoin</string></array> </dict></array>' \
+	-- dist/$PACKAGE.app/Contents/Info.plist \
+	|| fail "Could not add keys to Info.plist. Make sure the program 'plutil' exists and is installed."
 
 info "Creating .DMG"
 hdiutil create -fs HFS+ -volname $PACKAGE -srcfolder dist/$PACKAGE.app dist/electrum-$VERSION.dmg || fail "Could not create .DMG"
--- a/contrib/build-osx/make_osx
+++ b/contrib/build-osx/make_osx
@@ -15,7 +15,7 @@ cd $src_dir/../..
 export PYTHONHASHSEED=22
 VERSION=`git describe --tags --dirty`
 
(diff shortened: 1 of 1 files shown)


contrib/build-osx/make_osx
```diff
@@ -15,7 +15,7 @@ cd $src_dir/../..
 export PYTHONHASHSEED=22
 VERSION=`git describe --tags --dirty`
 
-which brew > /dev/null 2>&1 || fail "Please install brew from https://brew.sh/ to continue"
+which brew > /dev/null 2>&1 || which port > /dev/null 2>&1 || fail "Please install either Homebrew (from https://brew.sh/) or MacPorts (from https://www.macports.org/) to continue"
```
bauerj: We call brew directly later in this script so I don't think it works with port installed unless you provide an alternative there too.

cculianu: really? it worked for me!! I don't even have brew installed.. whaaa??

Ok double-checking now.. it totally worked!

Will get back to you in a minute..

cculianu: Yeah ok -- i know why it worked.  Because I already have autoconf, automake, etc installed. So:
```
# brew install autoconf automake
```

Is a no-op on my system (it did say command not found).

Man you guys are persecuting us MacPorts users!!! Ha ha :)

I have 2 options:

1. Modify your script further to cope with macports.. or 
2. just leave you in our brew-centric existance and get rid of that MacPorts thing.

Hmm.  What to do?

cculianu:
Ok, changes made.  Got rid of MacPorts support.  Maybe I'll do that as a separate PR.

bauerj:
Sorry for the long delay, I finally got around to test it now 😞

cculianu:
Ha! No worries! You got a master's thesis to finish up! 👍

SomberNight:
Could this instead go to
https://github.com/spesmilo/electrum/blob/1af225015a19e0697f3ca31535d74f6f646ea7ac/contrib/build-osx/osx.spec#L100-L103
?

If not, why not?

cculianu:
@SomberNight  Aye, I investigated this at the time.  Obviously this would be the preferred method rather than the less elegant "call plutil from a script" that this PR does.

Unfortunately pyinstaller spec files don't support anything but a very simple Info.plist structure.  Just a basic name=value dict is supported (where value is a string, bool, int).  'Value' cannot be something more complex such as a nested arrays or dict.  This URI type stuff on MacOS requires an array of dicts be added to the Info.plist (if you look at the actual XML -- you an see it's a rather convoluted array of dicts!).

So it can't be done using the spec file. :/

If you want to read more --  from the Pyinstaller docs: 

https://pythonhosted.org/PyInstaller/spec-files.html#spec-file-options-for-a-mac-os-x-bundle  (see a few paragraphs down in this section for the relevant discussion).

SomberNight:
All right, I see. :/ Thanks for explaining.

SomberNight:
Any idea if this file is used at all? https://github.com/spesmilo/electrum/blob/60f8cf665eb167ae315645c849fee81596d455e8/Info.plist

cculianu:
I seem to vaguely recall that it's cruft left over from a previous incarnation of the osx build scheme.  Let me double-check that to be sure.  I could be wrong. 

I will get back to you on that today... 💯

cculianu:
@SomberNight Good catch.  I just verified. That file is definitely cruft.  It was used before you guys upgraded the build system for OSX to be more streamlined (which we also ported over to Electron Cash).  

It's no longer used by anything and should be removed from the project because all it does is sow the seeds of confusion to anyone maintaining this.

I even double-checked this myself by removing it and re-building Electrum git head from scratch on a clean dir (sans Info.plist in root), just to be super-sure. 

That file is cruft.

I'm going to remove it from Electron-Cash right now.

I can issue a PR right now to you guys to remove it, too. 👍

cculianu:
Ok, added. @SomberNight  please see #4773

PR #4563, opened 2018-07-18
remove pbkdf2 dependency, use stdlib instead

ysangkok:


ysangkok:
Needs testing with DigitalBitbox.

PR #4565, opened 2018-07-18
Trezor: fix spending coinbase outputs

yura-pakhuchiy:
Attempt to spend coinbase output results in error:
a bytes-like object is required, not 'str'

Note: I have not tested this with Bitcoin since obtaining coinbase outputs is not easy even on Bitcoin testnet. Originally I have found, tested and fixed this problem in Zcoin fork of Electrum and just backporting this fix back to upstream.

### DIFF
--- a/electrum/plugins/keepkey/keepkey.py
+++ b/electrum/plugins/keepkey/keepkey.py
@@ -288,7 +288,7 @@ class KeepKeyPlugin(HW_PluginBase):
         for txin in tx.inputs():
             txinputtype = self.types.TxInputType()
             if txin['type'] == 'coinbase':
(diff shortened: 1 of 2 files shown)


SomberNight:
Thanks, looks ok.
I think `b"\x00"*32` would be clearer; can you set it to that instead?
Also, this code is duplicated for keepkey, please fix it there too.
https://github.com/spesmilo/electrum/blob/e5661156f0d4667a91ea60cae5cd8189e0770a92/electrum/plugins/keepkey/keepkey.py#L291

yura-pakhuchiy:
@SomberNight updated!

SomberNight:
Thanks.

Issue #4595, opened 2018-07-31
Create  a new wallet with only one address


lightlizm:
Hi , I want to create  a new wallet with only one address for users, 
without changed address ,  just only one address,  And user can  get account balance , all transactions in this only address.

 Can I modify some code to achieve this goal when I create a new wallet?

Thanks.

SomberNight:
You can create an imported wallet from a WIF private key.

lightlizm:
I create a new wallet through CLI ， So  have a command that creating an imported wallet from a WIF private key?
Thanks

SomberNight:
You can do something like this
```
SEED=$(electrum --testnet make_seed --segwit)
electrum --testnet -w /home/user/.electrum/testnet/wallets/wallet_299 restore "$SEED"
ADDR=$(python3 -c "l=$(electrum --testnet -w /home/user/.electrum/testnet/wallets/wallet_299 listaddresses); print(l[0])")
PRIV=$(electrum --testnet -w /home/user/.electrum/testnet/wallets/wallet_299 getprivatekeys $ADDR)
electrum --testnet -w /home/user/.electrum/testnet/wallets/wallet_300 restore $PRIV
```

Unfortunately the two `restore` commands will prompt you for a password that cannot be provided as an argument -- I'll try to fix that on master -- but you get the idea.

SomberNight:
> If all unspent outputs as next inputs to create transaction, It appears problem that if last transaction's confirmations is 0 , it can't create next transaction. because now the balance is 0 . When last transaction has more than 1 confirmation, the balance has amount and it can create next transaction.

Make sure `Tools`>`Preferences`>`Transactions`>`Spend only confirmed coins` is unchecked.

> Now I payto 0.5 tBTC , anyone unspent is enough to pay 0.5 tBTC, Why need two unspend outputs as inputs to create a transaction ?

The automatic coin chooser is trying to preserve your privacy, by spending all-or-none coins that are on the same address. It is optimised for deterministic wallets, or at least imported wallets with many addresses. Whatever your usecase is, having only a single address and reusing that for every transaction is in general not a good idea.

PR #4600, opened 2018-08-01
cli 'restore': allow providing password as argument

SomberNight:
To streamline scripts; see e.g. https://github.com/spesmilo/electrum/issues/4595#issuecomment-409499489


SomberNight:
Note: instead of this, we might as well make `restore` and `create` available via RPC

PR #4606, opened 2018-08-04
fix ecc_fast.py for win64

zebra-lucky:
This patch fixes next error on win64:
```ctypes.ArgumentError: argument 1: <class 'OverflowError'>: int too long to convert```
When trying to make signature or sweep master key to new wallet.


SomberNight:
Thanks.

Note: the failure only happens when using 64-bit python with a 64-bit libsecp256k1 build;
the release builds (at least for windows, not sure about Mac...) use 32-bit python with 32-bit libsecp256k1.

zebra-lucky:
Yes, only Win 64-bit was concerned.

BTW there is on-travis 64-bit build of secp256k1:
https://github.com/zebra-lucky/secp256k1/releases

PR #4631, opened 2018-08-15
Removed "Spend from" for frozen adresses

cmm-31:
Before, the user was able to choose the "Spend from" option for frozen
adresses. This is now removed.


cmm-31:
As your solution is the better one I changed the commit to yours.

bauerj:
I wonder if this is the right behaviour. Suppose there were multiple addresses selected. Currently, if any of those is frozen the "Spend from" option will be hidden.

Wouldn't it make more sense to hide it only if all of the associated UTXOs are frozen?

cmm-31:
I wasn't really aware of that, good point tho. 
But if there are, f.e. several frzen and several unfrozen addresses - what should we show? Since if there is only one unfrozen address, it's obvious to have the spend from option for that particular address.

ecdsa:
closing; current master does not add frozen coins to the list

PR #4647, opened 2018-08-21
setup.py: add 'gui' extra. potentially build Qt icons files

SomberNight:
Rotonen suggested on IRC that it would be nice to be able to just pip install from github and have a working installation with GUI; something like:
```
pip install -Ue git+https://github.com/spesmilo/electrum.git@3.2.2#egg=electrum[fast,gui]
```

For this,
- a `gui` extra is added that contains pyqt5
- we try to build the `icons_rc.py` file if it's not present and if pyqt5 is installed

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,12 +2,15 @@
 
 # python setup.py sdist --format=zip,gztar
 
-from setuptools import setup, find_packages
 import os
 import sys
 import platform
 import imp
 import argparse
+import subprocess
+
+from setuptools import setup, find_packages
+from setuptools.command.install import install
 
 with open('contrib/requirements/requirements.txt') as f:
     requirements = f.read().splitlines()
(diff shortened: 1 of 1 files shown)


Rotonen:
Tried with `pip install -Ue git+https://github.com/SomberNight/electrum.git@setuptools_pyrcc#egg=electrum[fast,gui]` and it fails to find the icon file. It'll also need to add it to the package resources / `MANIFEST.in` upon the extra.

Additionally the old guard responded to me on distutils-sig about how to detect extras gracefully:
https://mail.python.org/mm3/archives/list/distutils-sig@python.org/thread/Z3LLXCCWWFHNRIHQTVFSHSYEJYNXSVXX/#Z3LLXCCWWFHNRIHQTVFSHSYEJYNXSVXX

Rotonen:
Seems one should have compiled it before `sdist.read_manifest()` kicks in, so it gets picked up along for the ride.

https://github.com/pypa/setuptools/blob/0254a2fda8e8bd4f289d01e2179191e936517f04/setuptools/command/sdist.py#L180

So hook into `sdist` instead of `install` and compile the file before you run the original `sdist.read_manifest()`.

Rotonen:
Sorry for the noise, user error.

`pip install -U git+https://github.com/SomberNight/electrum.git@setuptools_pyrcc#egg=electrum[fast,gui]` works.

The `-e` flag for pip makes the egg editable, which runs `develop` instead of `install`.

https://github.com/pypa/setuptools/blob/0254a2fda8e8bd4f289d01e2179191e936517f04/setuptools/command/develop.py#L109

Issue #4641, opened 2018-08-18
Windows 7 64bit > Portable versions 3.2.* > CANNOT START PROPERLY


digital-outsider:
Houston, we have a problem.
After upgrading to 3.2 version, the portable exe, keeps running in background without starting the GUI.
I can see the process in windows activity management panel ... but nothing happens.
I've tried with all 3 versions ( 3.2.0\1\2 ) also installing the KB2999226 update as suggested.

![windows activity](https://user-images.githubusercontent.com/42490251/44297940-404a9500-a2da-11e8-9f4d-191f405f08e9.jpg)

**The big problem is that upgrade also modify wallet data so i cannot downgrade to 3.1.3** !!!
You need to fix this asap.
THANKS

* Attached you find also a dump of the process.

[electrum-3.2.1-portable-DUMP.zip](https://github.com/spesmilo/electrum/files/2299324/electrum-3.2.1-portable-DUMP.zip)

SomberNight:
> The big problem is that upgrade also modify wallet data so i cannot downgrade to 3.1.3 !!!

You can always just restore from your seed words.

Also, the upgrade does not happen without GUI interaction (even if your wallet does not have a password, before an upgrade, the wizard gets displayed, and you need to click "Next"). Hence, I don't understand how your wallet could have upgraded if the GUI does not even appear -- or did that window appear?

What version of Electrum were you using before?

If I built a custom binary for you that printed some debug logs, would you be willing to run that?

SomberNight:
Ok, try with this binary then. It should open an additional window with debug logs while running.

[electrum-3.2.2-debug-portable.exe.zip](https://github.com/spesmilo/electrum/files/2299397/electrum-3.2.2-debug-portable.exe.zip)
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

8a13a224523845172c3cc6446a6879c5b55931f37cde846e7527561b8d219775	electrum-3.2.2-debug-portable.exe
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEEbXohFtqQngCsIRCLsztfIyxicekFAlt3+HIACgkQsztfIyxi
celxDw/+M1uZkpnygN55vVlO3/Ze5D2YXAgiao0JXX67UMQbeDTIpgsyJmC7QAwh
DSYUNMh165Ywo3nWXj87cQw7q0d11T2EQvnhMfjwmGre4w1zjDvW2nlEA2vuMCF/
[...]

PR #4648, opened 2018-08-21
storage upgrades: make a copy of the wallet before upgrading

SomberNight:
This might be useful for the user should anything go wrong.

For example, in https://github.com/spesmilo/electrum/issues/4641#issue-351807160, the user ran into a problem with the newer version but could not go back to the older one, because of the upgrade.


SomberNight:
Note that it's not necessarily the upgrade itself that might be problematic. If there is a problem with a new release, and there was an upgrade, the user cannot go back to the previous one.

In any case, valid point regarding the user confusion and questions.

PR #4655, opened 2018-08-23
Add signature Low R grinding to match with Bitcoin Core

junderw:
Ref: https://github.com/bitcoin/bitcoin/pull/13666

Depends on python-ecdsa pull request to allow for extra_entropy
Ref: https://github.com/warner/python-ecdsa/pull/92

### DIFF
--- a/electrum/ecc.py
+++ b/electrum/ecc.py
@@ -373,12 +373,15 @@ class ECPrivkey(ECPubkey):
         if sigdecode is None:
             sigdecode = get_r_and_s_from_sig_string
         private_key = _MySigningKey.from_secret_exponent(self.secret_scalar, curve=SECP256k1)
-        sig = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sigencode)
+        def sig_encode_r_s(r, s, order):
+            return r, s
+        r, s = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sig_encode_r_s)
         counter = 0
-        while ord(sig[0]) > 127: # grind for low R value https://github.com/bitcoin/bitcoin/pull/13666
+        while r >= 2**255: # grind for low R value https://github.com/bitcoin/bitcoin/pull/13666
             counter += 1
-            extra_entropy = bfh("%064x" % counter)[::-1]
-            sig = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sigencode, extra_entropy=extra_entropy)
+            extra_entropy = int.to_bytes(counter, 32, 'little')
+            r, s = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sig_encode_r_s, extra_entropy=extra_entropy)
+        sig = sigencode(r, s, CURVE_ORDER)
         public_key = private_key.get_verifying_key()
         if not public_key.verify_digest(sig, data, sigdecode=sigdecode):
             raise Exception('Sanity check verifying our own signature failed.')
--- a/electrum/ecc.py
+++ b/electrum/ecc.py
@@ -374,6 +374,11 @@ class ECPrivkey(ECPubkey):
             sigdecode = get_r_and_s_from_sig_string
         private_key = _MySigningKey.from_secret_exponent(self.secret_scalar, curve=SECP256k1)
         sig = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sigencode)
+        counter = 0
+        while ord(sig[0]) > 127: # grind for low R value https://github.com/bitcoin/bitcoin/pull/13666
+            counter += 1
+            extra_entropy = bfh("%064x" % counter)[::-1]
+            sig = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sigencode, extra_entropy=extra_entropy)
         public_key = private_key.get_verifying_key()
         if not public_key.verify_digest(sig, data, sigdecode=sigdecode):
             raise Exception('Sanity check verifying our own signature failed.')


SomberNight:
Note: there are about 20 tests in `test_bitcoin.py` and `test_wallet_vertical.py` that exploit the deterministic nature of signatures -- these would need to be updated.

Also, interestingly this PR got [flagged ](https://travis-ci.org/spesmilo/electrum/requests)by Travis CI as "abuse", that's why the tests did not run.

junderw:
- [ ] warner/python-ecdsa#92 is merged and version updated
- [ ] Update tests in test_bitcoin.py and test_wallet_vertical.py

Thanks for informing me. I will keep an eye out for python-ecdsa. Then modify this PR accordingly to update to the newer ecdsa...

Discussion Topic: Do we really want to even do this with Electrum. If we don't follow, it becomes easier to fingerprint transactions on the blockchain based on which wallets support and don't support.

electrum/ecc.py
```diff
@@ -374,6 +374,11 @@ def sign(self, data: bytes, sigencode=None, sigdecode=None) -> bytes:
             sigdecode = get_r_and_s_from_sig_string
         private_key = _MySigningKey.from_secret_exponent(self.secret_scalar, curve=SECP256k1)
         sig = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sigencode)
+        counter = 0
+        while ord(sig[0]) > 127: # grind for low R value https://github.com/bitcoin/bitcoin/pull/13666
+            counter += 1
+            extra_entropy = bfh("%064x" % counter)[::-1]
```
ysangkok: wouldn't `int.to_bytes(counter, 32, 'little')` be equivalent, clearer and faster?

electrum/ecc.py
```diff
@@ -374,6 +374,11 @@ def sign(self, data: bytes, sigencode=None, sigdecode=None) -> bytes:
             sigdecode = get_r_and_s_from_sig_string
         private_key = _MySigningKey.from_secret_exponent(self.secret_scalar, curve=SECP256k1)
         sig = private_key.sign_digest_deterministic(data, hashfunc=hashlib.sha256, sigencode=sigencode)
+        counter = 0
+        while ord(sig[0]) > 127: # grind for low R value https://github.com/bitcoin/bitcoin/pull/13666
```
SomberNight: First, `sig` is bytes, so instead of `ord(sig[0])`, you want `sig[0]`.
But anyway, `sig` is in many cases DER encoded! It depends on `sigencode`. So this `while` will not be entered of course.
You should define something like
```
def sig_encode_r_s(r, s, order):
    return r, s
```
and pass that when to python-ecdsa, and only when r has the desired property, use the provided sigencode:
```
sig = sigencode(r, s)
```

junderw:
Just to confirm: I just found out `int.to_bytes` is not available in Python2, but that's ok, correct?

SomberNight:
yes. current minimum we want is 3.4

junderw:
No problems here :-D
```
$ python
Python 3.5.6 (default, Aug 24 2018, 18:49:10) 
[GCC 5.4.0 20160609] on linux
>>>
```

SomberNight:
It's bothering me somewhat that we still don't have this because python-ecdsa hasn't had a release from its master branch. (and last release was in Feb 2015, 4.5 years ago...)

Maybe we should try to remove python-ecdsa completely, and just use libsecp256k1... (although @ysangkok pointed out that pubkey-recovery is not compiled into libsecp256k1 by default, so that might result in complications...)

Or we could monkey-patch python-ecdsa to have [the required PR](https://github.com/warner/python-ecdsa/pull/92); we would probably also need to pull in `src/ecdsa/rfc6979.py` to electrum...

-----

In any case, another thing that I realised (which we would have caught when trying to have the tests pass) is that this grinding needs to be duplicated for the libsecp256k1 monkey-patches.
Something like this would probably do:

```patch
diff --git a/electrum/ecc_fast.py b/electrum/ecc_fast.py
index aec7f33aa..db4af28ff 100644
--- a/electrum/ecc_fast.py
+++ b/electrum/ecc_fast.py
@@ -179,14 +179,19 @@ def _prepare_monkey_patching_of_python_ecdsa_internals_with_libsecp256k1():
         nonce_function = None
         sig = create_string_buffer(64)
         sig_hash_bytes = hash.to_bytes(32, byteorder="big")
-        r = _libsecp256k1.secp256k1_ecdsa_sign(
-            _libsecp256k1.ctx, sig, sig_hash_bytes, secret_exponent.to_bytes(32, byteorder="big"), nonce_function, None)
-        if not r:
-            raise Exception('the nonce generation function failed, or the private key was invalid')
-        compact_signature = create_string_buffer(64)
-        _libsecp256k1.secp256k1_ecdsa_signature_serialize_compact(_libsecp256k1.ctx, compact_signature, sig)
-        r = int.from_bytes(compact_signature[:32], byteorder="big")
-        s = int.from_bytes(compact_signature[32:], byteorder="big")
+        counter = -1
+        while counter == -1 or r >= 2**255:  # grind for low R value https://github.com/bitcoin/bitcoin/pull/13666
+            counter += 1
+            extra_entropy = counter.to_bytes(32, byteorder="little")
+            ret = _libsecp256k1.secp256k1_ecdsa_sign(
+                _libsecp256k1.ctx, sig, sig_hash_bytes, secret_exponent.to_bytes(32, byteorder="big"),
+                nonce_function, extra_entropy)
+            if not ret:
+                raise Exception('the nonce generation function failed, or the private key was invalid')
+            compact_signature = create_string_buffer(64)
+            _libsecp256k1.secp256k1_ecdsa_signature_serialize_compact(_libsecp256k1.ctx, compact_signature, sig)
+            r = int.from_bytes(compact_signature[:32], byteorder="big")
+            s = int.from_bytes(compact_signature[32:], byteorder="big")
         return ecdsa.ecdsa.Signature(r, s)
 
     def verify(self: ecdsa.ecdsa.Public_key, hash: int, signature: ecdsa.ecdsa.Signature) -> bool:
```

Note that while libsecp256k1 is an optional dependency, we cannot have vanilla python-ecdsa and libsecp256k1 generate different signatures, as that would make at least testing immensely complicated, so either both or neither needs to grind R values.

junderw:
@SomberNight do note that in Bitcoin Core, the way it works is:

```
1. Sign once with no extra entropy (null)
2. If high R, then repeat with little endian 32 byte counter of 1 (not 0)
3. Repeat 2 increasing the counter (2,3,4,5,6,7...) until low R is found.
```

Your patch will start the first try with an extra entropy of 0, which is not technically the same as Bitcoin Core (or my Pull Request for that matter)

Just a heads up.

SomberNight:
python-ecdsa 0.14 is now released with your required patch
I opened a new PR that extends this one: https://github.com/spesmilo/electrum/pull/5820

junderw:
Closed in favor of #5820

I checked out the diff and it LGTM, though I haven't tested.

PR #4668, opened 2018-08-29
try fixing rare Qt crash when closing window

RGauthamRam:
Resolving the issue #4363


SomberNight:
Do you have a way to reproduce that crash?
I am still not clear how it could happen.

But ok, I guess this could not hurt much.

PR #4669, opened 2018-08-30
test_bitcoin: fix decorators for running tests twice

toxeus:
This bug is triggered if `ecc_fast._libsecp256k1` and/or `crypto.AES` are not present.

Before, if the first test would have raised an exception it would have been implicitly caught by returning from the `finally` block and hence this effectively masks the test's outcome. Now, the exception is properly propagated causing the test to fail if an exception is raised.

The effect of the `return` statement in a `finally` block can be observed using the following snippet.
```py
def foo():
   try:
       raise Exception()
   finally:
       return

try:
    foo()
except:
    print("Exception?")
else:
    print("No exception!!!")
```
Contributed by FDF.

### DIFF
--- a/electrum/tests/test_bitcoin.py
+++ b/electrum/tests/test_bitcoin.py
@@ -48,10 +48,10 @@ def needs_test_with_all_ecc_implementations(func):
             # first test without libsecp
             func(*args, **kwargs)
         finally:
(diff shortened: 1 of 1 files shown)


SomberNight:
Thanks.

PR #4684, opened 2018-09-04
docker: ignore .buildozer folder

toxeus:
The `.buildozer` folder is not needed for the Windows build. Moreover, it causes the `touch` command to fail. Therefore, it should be ignored from docker's context.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.buildozer


toxeus:
@SomberNight wrote:
> It's a bit annoying that this should go into the root folder.

I fully agree. It's also annoying that build systems shamelessly clutter the repo with build artefacts. I guess the excuse is that it's the way `make` works in C/C++ projects. `CMake` has partially cleaned up that behaviour. In Go the source files and the build artefacts are strictly separated. At least some build tools do it right.

> What if at some point we want to use docker to build the Android binaries? That build script would need this directory.

It would be very desirable to start using docker for the Android binaries! Then, the build artefacts would be captured inside the container rather than cluttering Thomas' repo on his host's filesystem. We could then remove the `.dockerignore` file again. This of course only works if you decide to use docker as intended #4554.

> There also does not seem to be an option to specify a path for this file on the CLI (moby/moby#12886).

The docker CLI always used to lag behind some very obvious use cases. I started using docker in 2014 and back then there was no `-f` flag to specify a custom `Dockerfile`. So you could only have a single `Dockerfile` per project and it had to be in the project's root (or whatever the root of your build context is). That's a huge oversight! If electrum would do the Android build using docker then the first thing you'd need is support for a second `Dockerfile` :laughing: 

> Also, while atm the failure was due to touching files in .buildozer, in general there could be other files
in the project's folder that are not needed for the build and that are causing problems --- though I guess those could be added to this blacklist too.

I agree. IMO it's worth to become aware of what the build system is doing to your code base and then deliberately fix issues - especially if deterministic builds are a requirement.

SomberNight:
Does this assume https://github.com/spesmilo/electrum/pull/4554 ?

I've tested and this change does not solve the original problem; the `.buildozer` folder is still present.
Presumably due to us mounting the whole folder into the container at
https://github.com/spesmilo/electrum/blame/ecbbfdd10c9fdfefabc94ccd871266a9117b34d5/contrib/build-wine/docker/README.md#L34

toxeus:
Yes, this does assume #4554. I seem to have forgotten that it's not merged yet.

toxeus:
One more thing that came to mind: provided that we have the build system as in #4554 then we can also do a `git clean -fdx` in the very beginning of the build process and hence become robust to repos in any state of "dirtiness". Personally, I don't really like brute force approaches like this but I thought I'll share that idea nevertheless.

PR #4685, opened 2018-09-04
 build-wine: avoid untracked changes in submodule 

toxeus:
https://github.com/spesmilo/electrum-locale/commit/27e36687f4b0fbd126628bdde80758b63ade7347 should be reverted after this has been merged.


toxeus:
This patch eliminates the built files from the submodule. So there's no need to ignore anything.

SomberNight:
Yes, it's redundant. Still, I don't think there's any harm in keeping it :P

PR #4697, opened 2018-09-07
aiorpcx: detect lost connection

ysangkok:
By looking whether the `pm_task` (process messages task) has exited, we can detect whether the connection was lost.

Submitting this for review since I think this might belong in NotificationSession. But what would be the API? To handle the case of a lost connection and still process incoming messages, it seems necessary to wait for two events concurrently. In this PR, I do that by using a WaitGroup with wait=any.

Note that TimeoutAfter results in a CancelledError propagating:
```
import aiorpcx
import asyncio
async def f():
    async with aiorpcx.curio.TimeoutAfter(1):
        await asyncio.sleep(2)

asyncio.get_event_loop().run_until_complete(f())
```

This example dies because of CancelledError, not a 'TaskTimeout', 'TimeoutCancellationError', or 'UncaughtTimeoutError'.

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -31,7 +31,6 @@ import threading
 import traceback
 import aiorpcx
 import asyncio
-import concurrent.futures
 from aiorpcx import ClientSession, Notification
 
 import requests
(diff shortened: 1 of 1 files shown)


kyuupichan:
If you need to detect that, inherit the session class and override the connection_lost method.  Be sure to call the base class method.

kyuupichan:
I believe TaskTimeout is inherited from CancelledError.  It should be a TaskTimeout exception.

kyuupichan:
See https://github.com/kyuupichan/aiorpcX/blob/master/tests/test_curio.py#L706-L717

SomberNight:
I've pushed https://github.com/spesmilo/electrum/commit/7500b1fbee57cd34bcdaaec7846e88a97cc8b829 instead -- I think that is clean.

Issue #4215, opened 2018-04-02
missing link to specific revision breaks deterministic build


toxeus:
Due to the non-deterministic behavior of `pyrcc5` the `icons_rc.py` file is pre-generated and stored in an external repo https://github.com/spesmilo/electrum-icons. However, the build scripts blindly [checkout](https://github.com/spesmilo/electrum/blob/10a0b0ad7f97666143e3172542970f765c54ecfd/contrib/build-wine/build-electrum-git.sh#L30) the [master branch](https://github.com/spesmilo/electrum/blob/10a0b0ad7f97666143e3172542970f765c54ecfd/contrib/build-osx/make_osx#L46) instead of a specific revision. This breaks deterministic builds.

Example:
commit https://github.com/spesmilo/electrum-icons/commit/a1a986ceed2d4546cf719ebfd9ab3f98df2ce979 was added for the 3.1 release. Since then the build of previous versions will not result in the same build as it used to be at release time of the previous versions.

```bash
cd /electrum/contrib/build-wine
build-electrum-git.sh 3.0.6 # uses wrong `icon_rc.py` now
```

The correct fix would be to use git submodules to create a commit-specific link between the main repo and the other repo. Then, `git clone --recursive` can be used to checkout the dependency repos at the correct revision.

Alternatively, for the special case of `icon_rc.py`, you might consider version controlling it in the main repo since it's just a single file.

Last but not least, the very same issue is experienced with https://github.com/spesmilo/electrum-locale.

toxeus:
This has been fixed in the PR referenced above. Closing.

PR #4711, opened 2018-09-13
gui: use existing icons_rc.py

toxeus:
This cleans up all sorts of custom build steps or manual user action and is fully compatible with
```sh
pip install git+https.
```
The `pyrcc5` command now only needs to be used by developers that wish to update the `icons_rc.py`
file.

Let's stop pretending that `icons_rc.py` is not available because since the move to submodules
it is :wink: 

Contributed by FDF.

### DIFF
--- a/electrum/gui/qt/__init__.py
+++ b/electrum/gui/qt/__init__.py
@@ -53,11 +53,13 @@ from .installwizard import InstallWizard
 
 
 try:
-    from . import icons_rc
+    from .icons import icons_rc
 except Exception as e:
     print(e)
     print("Error: Could not find icons file.")
-    print("Please run 'pyrcc5 icons.qrc -o electrum/gui/qt/icons_rc.py'")
+    print("If you're running electrum from the repo then please make sure that")
+    print("the submdules are initialized.")
+    print("git submodule update --init")
     sys.exit(1)
 
 from .util import *   # * needed for plugins
--- a/setup.py
+++ b/setup.py
@@ -7,10 +7,8 @@ import sys
 import platform
 import imp
 import argparse
-import subprocess
 
 from setuptools import setup, find_packages
-from setuptools.command.install import install
 
 with open('contrib/requirements/requirements.txt') as f:
     requirements = f.read().splitlines()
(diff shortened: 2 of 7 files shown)


toxeus:
Also, contrary to what this [statement](https://github.com/spesmilo/electrum/pull/4647#issuecomment-414842779), installing using `pip install git+https` doesn't work for me. According to [docs](https://github.com/pypa/pip/blob/master/NEWS.rst) in the pip repo this is expected:
> Installing from a local directory or a VCS URL now builds a wheel to install, rather than running setup.py install. Wheels from these sources are not cached. (https://github.com/pypa/pip/pull/4501)

This PR fixes it.

electrum/gui/qt/__init__.py
```diff
@@ -53,11 +53,18 @@
 
 
 try:
-    from . import icons_rc
+    import importlib
+    from pathlib import Path
+    path = str(Path(__file__).parent.joinpath("icons/icons_rc.py"))
+    spec = importlib.util.spec_from_file_location("icons_rc", path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
```
toxeus: By adding `__init__.py` into the icons submodule it would become a python package and could be imported using a simple `import` statement. Also, pyinstaller would then automatically pick up that file. If I leave this as is then the `*.spec` files would need some adjustment. I prefer the icons package solution. What do you think?

setup.py
```diff
@@ -88,6 +69,7 @@ def run(self):
         'electrum': [
             'wordlist/*.txt',
             'locale/*/LC_MESSAGES/electrum.mo',
+            'gui/qt/icons/icons_rc.py',
```
toxeus: This hack can be removed if icons becomes a python package.

electrum/gui/qt/__init__.py
```diff
@@ -53,11 +53,18 @@
...
 except Exception as e:
     print(e)
     print("Error: Could not find icons file.")
-    print("Please run 'pyrcc5 icons.qrc -o electrum/gui/qt/icons_rc.py'")
+    print("If you're running electrum from the repo then please make sure that")
+    print("the submdules are initialized.")
+    print("git submodule init")
```
bauerj: I think `git submodule update` is necessary in this situation too.

toxeus: I will change it to `git submodule update --init`.

electrum/gui/qt/__init__.py
toxeus: @bauerj any opinion on this?

bauerj: Although a package containing only a single module looks wrong I think it does make sense in this context.

SomberNight:
This would mean the icons submodule needs to be kept up to date with this repo. This is potentially extra maintenance work. With this PR, I guess we should update it every time `icons.qrc` or `icons/` changes.

Currently the electrum-icons repository is only updated just before there is a release. It's only there to make the binaries reproducible. If we could make the built `icons_rc.py` reproducible, I would much rather we killed the electrum-icons repo.

toxeus:
> This would mean the icons submodule needs to be kept up to date with this repo.

I don't see this to be a hard requirement. You can still keep your workflow as before.

> I would much rather we killed the electrum-icons repo.

I [like](https://github.com/spesmilo/electrum/issues/4215#issue-310451578) that idea too :+1:

toxeus:
@SomberNight, @bauerj so how do we proceed here?

PR #4715, opened 2018-09-14
build-wine: fix locale path

toxeus:
`$i` already contains `locale/`.


bauerj:
Why didn't this cause the CI test in #4685 to fail?

PR #4723, opened 2018-09-18
qt history list: hack to update faster for large wallets

SomberNight:
Typically half the CPU time of `HistoryList.on_update` is spent in `wallet.get_full_history`.
`HistoryList.on_update` can take several seconds for very large wallets, and this runs in the GUI thread.
This PR moves `wallet.get_full_history()` into a separate thread.

### DIFF
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -26,6 +26,7 @@
 import webbrowser
 import datetime
 from datetime import date
+from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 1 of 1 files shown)


cculianu:
Hmm. Interesting.  This definitely would have the effect of keeping the GUI peppy.  

Only question I have is have you considered a scheme to disallow more than 2 updates form being queued simultaneously?  

Rationale: an entire queue of update requests is a bit reduntant. Only the latest update ultimately "wins" anyway.  So in pathological cases one can imagine a dozen updates being enqueued and they all eat up the CPU as they get processed in series, when only the latest update "matters" anyway.

PR #4724, opened 2018-09-19
use system language by default

un1t:
Default language is english and settings menu is not available until you create first wallet. For not english speaking users it very hard to accomplish.

It would be better to use system language at first start.
My commit is applied to qt interface. I'm not sure where to set default language for kivy interface.

### DIFF
--- a/electrum/i18n.py
+++ b/electrum/i18n.py
@@ -26,6 +26,8 @@ import os
 
 import gettext
 
+from PyQt5.QtCore import QLocale
+
 LOCALE_DIR = os.path.join(os.path.dirname(__file__), 'locale')
 language = gettext.translation('electrum', LOCALE_DIR, fallback=True)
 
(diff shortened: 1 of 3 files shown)


electrum/i18n.py
```diff
@@ -26,6 +26,8 @@
 
 import gettext
 
+from PyQt5.QtCore import QLocale
```
SomberNight: you must not assume that qt is available, in the core lib.

un1t: You right. I also noticed this, after CI failed.

bauerj: Can you move the function to e.g. `electrum.gui.qt.util` in stead?

electrum/i18n.py
```diff
@@ -41,6 +43,11 @@ def set_language(x):
         language = gettext.translation('electrum', LOCALE_DIR, fallback=True, languages=[x])
 
 
+def get_default_language():
+    system_locale = QLocale.system().name()
+    return languages.get(system_locale, 'en_UK')
```
bauerj: Wouldn't this return something like `"German"` if `system_locale` was `de_DE`?

I think something like 

```python
return system_locale if system_locale in languages else "en_UK"
```

would make more sense.

un1t: Oops, yeah, you right!

un1t: I made this fix and moved this function to gui.qt.utils.

PR #4727, opened 2018-09-19
qt: refresh gui with "F5"

SomberNight:
Refresh Qt GUI tabs when user presses `F5` key.


bauerj:
When should this be used? I thought the GUI would always keep itself up to date.

SomberNight:
Yes, it should; but if the functionality is already there for "ctrl+R", why not "f5"? :P

Recently I've been debugging refresh bugs, and this would have been convenient to see whether an issue is really about not updating e.g. the history list or if the problem is with the underlying model/data. Did not remember that we already had this.

rayyyoubecos:
Thanks

PR #4765, opened 2018-10-10
cli/rpc: 'restore' and 'create' commands are now available via RPC

SomberNight:
supersedes https://github.com/spesmilo/electrum/pull/4600

note: not clear if I should raise Exceptions or return dicts with error keys
note: no clear way to communicate messages to user (how to replace `print_msg`?)


SomberNight:
Force-pushed -- as discussed, raising Exceptions instead of returning `{'error': ...}` dicts

PR #4769, opened 2018-10-12
build-wine: update git version

wakiyamap:
The following error occurred at build time.
`E: Version '1:2.17.1-1ubuntu0.1' for 'git' was not found`

It seems that git has updated.
https://packages.ubuntu.com/en/bionic/git


SomberNight:
Indeed; thanks.

Issue #4760, opened 2018-10-09
OpenCAP Alias Support


lane-c-wagner:
OpenCAP aliases allow users to send and receive crypto using a human readable alias. Electrum could integrate very easily, you just need to implement some REST API calls. OpenCAP is open source and decentralized.

protocol: https://github.com/opencap/protocol

server implementation: https://ogdolo.com

PR #4772, opened 2018-10-13
Added OpenCAP alias support

wagslane:
#4760 

Added OpenCAP alias support. Simply type the OpenCAP alias in the "send" box and a lookup will be done to get the current address.

I'm not super familiar with the design patterns of this repo, feel free to refactor a bit. It was only a bit of added code, the majority of the logic is done in the PIP opencap package.

I'm sorry my IDE formatted the whole files because I have "lint on save"... if you don't like that linter maybe make one standard for this repo? Or if there is one I apologize I didn't know about it.

### DIFF
--- a/electrum/gui/qt/paytoedit.py
+++ b/electrum/gui/qt/paytoedit.py
@@ -24,6 +24,8 @@
 # SOFTWARE.
 
 import re
+import json
+import requests
 from decimal import Decimal
 
 from PyQt5.QtGui import *
@@ -35,6 +37,7 @@ from electrum.transaction import TxOutput
 from .qrtextedit import ScanQRTextEdit
 from .completion_text_edit import CompletionTextEdit
 from . import util
+from opencap import *
 
 RE_ALIAS = '(.*?)\s*\<([0-9A-Za-z]{1,})\>'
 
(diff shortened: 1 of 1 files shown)


electrum/gui/qt/paytoedit.py
```diff
@@ -1,252 +1,278 @@
...
+
+        if not dnssec:
+            # maybe warn here
+            pass
+
+        response = requests.get(
+            "https://" + host + "/v1/addresses?alias="+username+"$"+domain)
+        json_data = json.loads(response.text)
+        for addressObject in json_data:
+            if addressObject["address_type"] == 100 or 101:
```
SomberNight:
```
>>> if False or 101:
...   print("asd")
...
asd
```

lane-c-wagner:
Okay, I'll fix. Sorry about that, didn't realize till I finished haha.

lane-c-wagner:
Got it fixed. Should be a lot easier to see what changed xD

electrum/gui/qt/paytoedit.py
```diff
@@ -82,7 +85,30 @@ def parse_address_and_amount(self, line):
...
```
SomberNight:
```
>>> if False or 101:
...   print("this will always evaluate to True")
...
this will always evaluate to True
```

electrum/gui/qt/paytoedit.py
```diff
@@ -35,6 +37,7 @@
 from .qrtextedit import ScanQRTextEdit
 from .completion_text_edit import CompletionTextEdit
 from . import util
+from opencap import *
```
SomberNight: do not import star, unless absolutely necessary

electrum/gui/qt/paytoedit.py
SomberNight: The [documentation is inconsistent](https://github.com/opencap/protocol/tree/5a8d42b4848c84413c4f1a745979b885a9377dbb/AddressTypes) but shouldn't 102 be also allowed?

ecdsa:
indeed, they should not introduce a new dependency

SomberNight:
> What are you referring to when you talk about verifying the root though?
Are you talking about the OpenCAP pip package?

Yes, I am saying that the `opencap` python package is not properly validating dnssec. You are only doing a self-signing test, which in itself is insufficient. That is the same level of security as self-signed https provides (without cert pinning): none. By root, I mean the dnssec root zone ksk.

lane-c-wagner:
Gotcha, thanks. Okay, I'll look at your DNSSEC stuff.

Issue #3190, opened 2017-11-03
v3 OSX Info.plist missing version number (0.0.0)


jonathancross:
`Info.plist` is missing version number:
 
```xml
<key>CFBundleShortVersionString</key>
<string>0.0.0</string>
```

Looks like this `0.0.0` is a default version added by PyInstaller: https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/building/osx.py#L45

Version was correctly set in previous versions.

Hopefully this is as simple as adding the `version` here:
https://github.com/spesmilo/electrum/blob/master/contrib/osx.spec#L65

Screenshot:
![image](https://user-images.githubusercontent.com/5115470/32356674-fc5b06c6-c036-11e7-8c9d-93083bdc734e.png)

ecdsa:
I have restored Info.plist in the 3.0.1 builds.
can you check if that works?

jonathancross:
No, unfortunatly the same issue is still here in 3.0.1 release.  See attached file: [Info.plist.txt](https://github.com/spesmilo/electrum/files/1447855/Info.plist.txt) (added `.txt` extension to make github happy)

[osx.spec](https://github.com/spesmilo/electrum/blob/master/contrib/osx.spec#L65) doesn't seem to have changed, did you try that?

Dedados:
[Fix](https://apple.stackexchange.com/questions/69038/open-in-low-resolution-checkbox-on-retina-macbook-pro-checked-and-disabled)

jonathancross:
@Dedados This has nothing to do with #3178 or the [Fix](https://apple.stackexchange.com/questions/69038/open-in-low-resolution-checkbox-on-retina-macbook-pro-checked-and-disabled) you posted.

Dedados:
Sorry, my mistake!

ecdsa:
@jonathancross it is possible to set NSHighResolutionCapable in the .spec file, according to the documentation: https://pythonhosted.org/PyInstaller/spec-files.html
so, fixing info.plist should fix both issues
I will try that with the next builds

jonathancross:
@ecdsa Great!  So, I guess there is a connection between the two issues afterall  :-)

ecdsa:
please try https://download.electrum.org/test/electrum-3.0.1-40-g8fcf165.dmg 
Let me know if the low resolution issue is fixed too

JustinTArthur:
That DMG fixes the low resolution issue (#3178) for me.

[...]

PR #4773, opened 2018-10-14
Removal of macOS Info.plist. It isn't being used by anything.

cculianu:
As per the discussion with @SomberNight [here](https://github.com/spesmilo/electrum/pull/4557#pullrequestreview-152242918) -- this file is no longer used.

I double and triple checked this.  This file was left over from a previous incarnation of the macOS build scripts.

The new pyinstaller-based build scripts do not use this file.  Instead the Info.plist is dynamically generated by a combination of what's specified in `contrib/build-osx/osx.spec` and some shell commands executed in `contrib/build-osx/make_osx`.

As such, this file is cruft and should be removed.


cculianu:
FYI -- The reason this was left in is discussed here #3190.

However, since then the pyinstaller spec file has evolved and this Info.plist is definitely no longer used and not picked up anymore.  

I double checked this. Proof:

screen shot 2018-10-14 at 1 16 43 pm

SomberNight:
Thanks

PR #4775, opened 2018-10-15
Show description (label) in TxDialog screen when opened from History

cculianu:
This was a pet peeve of mine that we fixed on Electron Cash.

Basically, the TxDialog window never gets the 'label' when a user double-clicks a tx in the History List.  They see the historical tx, but no "Description:" field, even if a label is defined for the tx.

This PR fixes that.

See screenshots:

### User selects transaction from History ###
screen shot 2018-10-15 at 11 35 27 am

### Notice how the TxDialog now has the correct tx `Description:`.  (If it lacks a description, then the behavior is as it was before -- no `Description:` field shown). ###

screen shot 2018-10-15 at 11 35 51 am


SomberNight:
Thanks

cculianu:
You are welcome sir

PR #4779, opened 2018-10-16
OpenCAP with requested changes

lane-c-wagner:
So I tried to follow the same pattern as OpenAlias here... the only problem is I'm not sure exactly how the UI/UX is supposed to work with OpenAlias, so it was hard to make sure that OpenCAP will work the same way.

Could someone that has some understanding make sure that this was done correctly?

The goal here is to be able to send Bitcoin to an OpenCAP alias, and maybe save it to the contact list.

#4760 
#4772 
@SomberNight thanks for the review of my last PR


SomberNight:
This looks better than before :)
However, as said in https://github.com/spesmilo/electrum/pull/4772#pullrequestreview-164463574, we would rather not introduce a new dependency for this. I understand that you might want to have a python package that you can reuse between different projects as a dependency, but I don't think it makes much sense for the ~30 lines of code that is actually being used from there. You should just move that code into Electrum, and not have a dependency here.

lane-c-wagner:
Okay sounds good I'll do that. Not a problem, I wasn't sure if you guys
just didnt want DNSSEC stuff or what.

I know this PR is still incomplete, this OpenCAP code I added needs to tie
into the UI right? Do you know where that happens?

SomberNight:
Note: I've fixed this so that it now works. To test, in Qt, try filling the `Pay to` field in the `Send` tab with e.g. `donate$ogdolo.com`. However, whether it will get merged will need further deliberation.

lane-c-wagner:
I'll take a look at it this evening. Thanks!

lane-c-wagner:
@SomberNight has there been any news as far as this getting merged?

Here is some info that may help with the decision:

Interview by Nano team: https://medium.com/nanocurrency/weekly-spotlight-opencap-8d7a10119c25

OpenCAP vs OpenAlias: https://blog.goodaudience.com/crypto-alias-comparison-opencap-vs-openalias-3e94f9daef94

OpenSource server is in beta: https://github.com/opencap/go-server

lane-c-wagner:
Any conversation or decisions on this PR?
open-source server now comes with full home setup tutorial: https://github.com/opencap/go-server
third-party server for testing: https://ogdolo.com

ecdsa:
No,we are not going to merge this. 
We already support OpenAlias, and it never reached a high adoption.

In addition, this kind of alias encouraging address reuse.
I would prefer to see aliases to pubkeys, which people would use to sign requests.

Issue #4790, opened 2018-10-22
standard wallet has mixed script type addresses. wallet file corrupted or malware


guillehaddad:
I'm trying to send some funds from my wallet but i keep getting the error below. I'm using electrum 3.2.3... A weird thing i've noticed is i have adresses starting with 3 (pe 3ENstJYCK9eyhEeXrq5XNE685RrMKi6WmP) and some starting with 1 (pe 1H1L1emw7Uyq4R9afp81BvyCE1xKge3PUd), thought i should have only one or the other...

error: {'code': 1, 'message': 'the transaction was rejected by network rules.\n\nmandatory-script-verify-flag-failed (Script evaluated without error but finished with a false/empty top stack element) (code 16)

SomberNight:
In the menu bar, go to `Wallet` > `Information`.
What is `Wallet type` and `Script type`?

guillehaddad:
standard and p2pkh

SomberNight:
Like you said, in that kind of wallet, there should not be a mix of "1.." and "3.." addresses. Specifically, all addresses should be "1.." addresses in this case. Your wallet file has either been corrupted somehow (but I doubt Electrum itself did this), or I would guess there is some kind of malware injecting its own addresses into your wallet file, hoping you receive coins there.
You should check the GPG signature or share the hash of the Electrum binary you've downloaded (what OS?). You should restore from seed, preferably on another machine.

guillehaddad:
i did check gpg sig before intalling it... i m using linux and even though i m no expert, i m not a new to btc and i ve been using electrum for years now... i can see the wallet got corrupted some how now looking at the wallet file... it started with p2pkh and all of the sudden it changes to p2wpkh... funds havent moved yet and i do keep my os update and dont install garbage on it, so i dont believe its some kind of malaware... the thing is as u said, all addresses should be p2pkh, but i ve deposited in a p2wpkh address... ive been trying to recover it using seed or xprivkey but havent had any success so far... is there anything  someone can think of that can help recover this funds? thanks a lot anyways

SomberNight:
Have you managed to recover these coins? If not, do you mind sharing the address where they are stuck?

ecdsa:
@guillehaddad 
 - why are you mentioning p2wkph? p2wpkh addresses do not start with a "3", they start with "bc1"
 - do you still have the wallet file?
 - can you post the complete error message (with the hexadecimal transaction)

ecdsa:
[...]

Issue #4462, opened 2018-06-25
Electrum error - 'message': 'the transaction was rejected by network rules.


guaji61:
Below is the error - this error made me to LOSE the funds - need help - I have the seed but because of the error I can't save the private keys!

This is not correct because the seed need to open my funds - the same electrum-3.1.2-portable program receives payments - but ONLY the wallet with the funds - DISAPPEARED!

error: {'message': 'the transaction was rejected by network rules.\n\n16: mandatory-script-verify-flag-failed (Script failed an OP_EQUALVERIFY operation)\n[01000000011b3d1c0cce91507fa7003f5d3c24e2fef200e303eb2de10bcff22f50cd665759000000006a473044022030fc5f836e7fc31f004850e5cfbaa59d5b48e182ad743b826f7814c4583ffe54022060d82df810f3aa09d86974d251f378b88ec83533685775fd8c6718c05d4d34e4012103a0328a951557897d0782945a7ca0e027c218e0387f478e07c2bd603d5280bfa6fdffffff02d0470000000000001976a91450aed898bf35087ea331cfabc3f03ed2dd1a797888ac66792b02000000001976a9149f321cac8e09435ca7856e26f6023966acd1c3a788ac1f000800]', 'code': -1}

SomberNight:
Hi. It's hard to understand the situation from your description.
What do you mean you can't save the private keys?
What do you mean by
> ONLY the wallet with the funds - DISAPPEARED

Have you tried restoring the wallet from seed words? Or is that what you are describing? Do you recognise the addresses in the restored wallet?

guaji61:
I already restored the wallet using the seed words - but as I said - only the wallet with the funds - DISAPPEARED!

Its hard to understand also this situation - but its exactly as I described. I received the funds - then I tried to send the funds to another wallet.

I got the error:

[...]

PR #4828, opened 2018-11-05
wallet: try detecting internal address corruption

SomberNight:
It looks like there might be malware out there which modifies wallet files of users, injecting foreign bitcoin addresses in hope of the user receiving coins there [0]
This is intended as naive defense against that.

- try to re-derive (from xpub) the first 10 addresses of the wallet, plus 10 additional random ones on every wallet-open (only if libsecp256k1 is available as otherwise it would be too slow)
- try to re-derive address (from xpub) in following cases
  - address shown in `Receive` tab
  - address to sweep to
  - user right-click copies address in `Addresses` tab
  - "change" addresses used in newly created transactions

The idea is that now the malware author needs to also change the xpub saved in the wallet to pass these checks; as well as changing most of the addresses stored in the wallet (to the ones actually derived from the foreign xpub). If the original wallet has any history at all, the transactions will disappear from the history, and the user SHOULD notice that something is going on.

-----

[0] see issues:
- https://github.com/spesmilo/electrum/issues/4790
- https://github.com/spesmilo/electrum/issues/4462
- https://bitcoin.stackexchange.com/questions/78813/same-seed-from-electrum-but-generate-different-bitcoin-addresses
- https://github.com/spesmilo/electrum/issues/4891

Also, I see many transactions failing on my server with the following error:
`'the transaction was rejected by network rules.\n\n16: mandatory-script-verify-flag-failed (Script failed an OP_EQUALVERIFY operation)`
which might point to this type of attack

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -9,9 +9,9 @@ import threading
 
 from electrum.bitcoin import TYPE_ADDRESS
 from electrum.storage import WalletStorage
-from electrum.wallet import Wallet
+from electrum.wallet import Wallet, InternalAddressCorruption
 from electrum.paymentrequest import InvoiceStore
-from electrum.util import profiler, InvalidPassword
+from electrum.util import profiler, InvalidPassword, send_exception_to_crash_reporter
 from electrum.plugin import run_hook
 from electrum.util import format_satoshis, format_satoshis_plain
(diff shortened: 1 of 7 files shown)


ecdsa:
addresses also need to be checked when:
- coins are sent to a change address
- an address is copied from the addresses tab

SomberNight:
covered those cases too now, and updated opening comment

PR #4829, opened 2018-11-07
add new blockstream.info as explorer option - supports bech32 and more

neocogent:
Pretty simple. Tested on both main and testnet without issues. 
Nice that it does support bech32 addresses.


SomberNight:
Thanks, was just talking about this on IRC when you PR-ed >.<

greenaddress:
thanks

Issue #3147, opened 2017-10-29
weak password hashing used


undeath:
Electrum is using relatively weak password hashing to protect the encrypted wallet.

https://github.com/spesmilo/electrum/blob/f7b14cb27f42700672168c250fd561e8f00e17d5/lib/storage.py#L122

It does use the (kind of outdated) PBKDF2 key derivation with no salt and a quite low iteration count. Wouldn't it be more sensible to save PBKDF2 settings in the wallet file, especially including a random salt and ideally a user-adjustable high iteration count (>10000)?

Omitting a random salt makes Electrum wallets prone to rainbow table attacks and the low, fixed iteration count is making the creation of such tables even easier. I realise there is still the work of doing the EC decryption but this is allowing unnecessary shortcuts for an attacker.

dabura667:
This is only protecting the metadata. The private keys are protected with a different function.

undeath:
You are correct. The snipped quoted above is only used if the wallet is encrypted. But the same password is used to encrypt the wallet (storage) itself and its private key(s) (keystore):
https://github.com/spesmilo/electrum/blob/f03cb757131cd5bcef376ccc77b9214e5837a230/lib/base_wizard.py#L354-L357

The function to encrypt the private keys themselves is indeed even weaker and only using two iterations of SHA-256 without any salting (used in pw_encode/pw_decode):
https://github.com/spesmilo/electrum/blob/f03cb757131cd5bcef376ccc77b9214e5837a230/lib/bitcoin.py#L244-L247

SomberNight:
(from https://github.com/spesmilo/electrum/pull/4838)
> Currently, when keystore encryption is enabled, the symmetric key used to encrypt the seed/private keys is sha256d(password). (If storage encryption is also enabled, then the whole wallet file is encrypted again, there the key derivation uses PBKDF2 with 1024 iterations of sha512)

Making the KDF more expensive is blocked on https://github.com/spesmilo/electrum/issues/4909
It's more complicated than I had thought... :/

Perhaps it would make sense to use the same KDF in both cases.
Anyway, to change the storage KDF, metadata needs to be put in the header of the encrypted file.
To change the keystore KDF, metadata needs to be put in the (decrypted) storage.

SomberNight:
>  Anyway, to change the storage KDF, metadata needs to be put in the header of the encrypted file.

See https://github.com/spesmilo/electrum/issues/5999.
[...]

PR #4838, opened 2018-11-10
keystore: stronger pbkdf for encryption

SomberNight:
Currently, when keystore encryption is enabled, the symmetric key used to encrypt the seed/private keys is `sha256d(password)`. (If storage encryption is also enabled, then the whole wallet file is encrypted again, there the key derivation uses PBKDF2 with 1024 iterations of sha512)

This PR introduces versioning for the keystore encryption, and the existing behaviour becomes version 1.
Version 2 uses PBKDF2 with 50k iterations of hmac-sha256 with a random 16 byte salt.

Newly created wallets will use version 2 automatically; and for existing wallets, if they change the password, they will get upgraded to version 2.

see: https://github.com/spesmilo/electrum/issues/3147

### DIFF
--- a/electrum/crypto.py
+++ b/electrum/crypto.py
@@ -32,6 +32,7 @@ from typing import Union
 import pyaes
 
 from .util import assert_bytes, InvalidPassword, to_bytes, to_string
+from .i18n import _
 
 
 try:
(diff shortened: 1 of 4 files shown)


SomberNight:
note: maybe the random 16 byte per-wallet salt is overkill as the secret (which is a seed or xprv or WIF key) should be high entropy.

If an imported wallet has a popular leaked WIF key imported but also other high value privkeys then without the salt, the leaked popular WIF key could compromise the whole wallet if an attacker has precomputed rainbow tables for it -- but this is such a stretch...

In any case, it is low cost for us to implement, so why not.

undeath:
Why are you planning to switch to a deprecated kdf like pbkdf2 instead of a gpu-resistant kdf like (ye)scrypt or argon2?

SomberNight:
- Mainly because it's in python stdlib so we don't need to introduce a new dependency.
- Even if we introduced a new dependency, Electrum is at the moment still (by default) a pure python application, which introduces further complication.
- Also, whatever we choose will also be run by the Android app, which uses python-for-android, where not even everything that would normally be in the stdlib is available (e.g. scrypt is in python stdlib starting at python 3.6, but it's not available on python-for-android (at least not with crystax NDK, not sure about the newer developments there)).

But if you have an alternative in mind that might conceivably work, we could use that.

undeath:
That makes it indeed hard to use other kdfs. Since most newer ones are optimized to run fast on x86-64 CPUs a pure python implementation will slow them down considerably.

A pure python implementation for scrypt existis (https://github.com/ricmoo/pyscrypt) but given the further complexity and additional external dependency it may not be worth choosing over pbkdf2 in this case.

AbdussamadA:
hashlib supports scrypt since python 3.6. https://docs.python.org/3/library/hashlib.html#key-derivation .

PR #4861, opened 2018-11-22
blockchain: generalise fork handling and follow most work chain

SomberNight:
- id forks by first blockhash instead of forkpoint height
- swap child fork with parent based on work instead of length

### DIFF
--- a/electrum/blockchain.py
+++ b/electrum/blockchain.py
@@ -79,26 +79,67 @@ def hash_raw_header(header: str) -> str:
     return hash_encode(sha256d(bfh(header)))
 
 
-blockchains = {}  # type: Dict[int, Blockchain]
-blockchains_lock = threading.Lock()
+# key: blockhash hex at forkpoint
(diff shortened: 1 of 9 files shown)


SomberNight:
Well, @rt121212121 had said he has something locally that probably is more or less equivalent to https://github.com/spesmilo/electrum/pull/4861/commits/141ff99580192c920bc6bb7f6bbc9d35449daea8
but otherwise it's applicable and could be ported, yes

but re the other commit, https://github.com/spesmilo/electrum/pull/4861/commits/65ce3deeaa33828407cf3a873c7ce5c48fa0b6d4, that is specific to Bitcoin, as BCH has different difficulty adjustment rules now

rt121212121:
My version is a part of a refactoring that has a tentative alternate sqlite3 backend. The end goal would be dropping the flat files and going with the much simpler database backend.  However, it's not ready yet, it seems to work, but I need to write more unit tests.

PR #4870, opened 2018-11-26
[macOS] Add QR Reader. Uses native obj-c helper app (included). Automatic code signing also added.

cculianu:
Note this PR includes within it #4869.  As far as I know it's impossible to make PR's depend on one another so I submitted #4869 first, then this (which requires it).

This is similar to https://github.com/Electron-Cash/Electron-Cash/pull/977, although this Electrum version builds from source each time and does not include a compiled binary.

--- 
Sample screenshot taken from Electron Cash:

me_ec_test

---

This feature adds much-needed QR scanning capability to Electrum on macOS.

Note that zbar does **not** work on macOS for reading from the webcam.  So -- a different approach was needed.

I did some digging and most other approaches I could find have external dependencies that just make building and packaging too cumbersome.  (See: https://github.com/spesmilo/electrum/pull/3485 for an example of a cumbersome approach).

However -- macOS since 10.7 comes with very good video processing capabilities right in the OS.  This includes the ability to decode QR codes! Why bundle heavy-handed libs when the OS already provides them? (And -- they are **hardware accelerated**, to boot!).

So -- I wrote a small helper app that Electrum can execute to read QR codes from a camera.  The entire app is 80KB compiled -- so it's tiny and on the order of the size of other bundled binaries included on macOS such as `libusb.dylib`. 

The workflow for this app is identical to the 'zbar' approach used on Linux and Windows.

This PR also takes care of bundling and signing the helper app in the built .app & .DMG for macOS.  

The helper app basically creates a window and reads from the default camera device on the system. It will continue to run until either the window is closed by the user or a QR image is read.

#### Rationale for using a Helper App rather than a dylib

1. macOS Mojave and above require special Info.plist keys and annoying permissions from the user to access the camera. It was far simpler to do this as a self-contained app, than make Electrum adapt to this insanity.
2. It's far less invasive to have a standalone app than to build a dylib when it comes to creating GUI windows -- no interference with Electrum's GUI event loop. Everything is self-contained.

#### Workflow for the helper app is as follows (note this is identical to zbar):

1. If a QR image is scanned, it will print the decoded string to stdout and exit.
2. If there is an error detecting the camera, it will show an error message and wait.
3. If the user closes the window without having scanned a QR code (because of an error or s/he changed his/her mind), it will print nothing to stdout and exit.

As such, `lib/qrscanner.py` was modified to call the helper app using `subprocess.Popen` in the 'darwin' case. The 'windows' and 'linux' cases were left unmodified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ electrum/gui/kivy/theming/light.atlas
 .cache/
 .coverage
 .pytest_cache
+
+# Xcode project per-user data
+xcuserdata/
+
--- a/contrib/CalinsQRReader/CalinsQRReader.xcodeproj/project.pbxproj
+++ b/contrib/CalinsQRReader/CalinsQRReader.xcodeproj/project.pbxproj
@@ -1,0 +1,339 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 50;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		FAA6C5A221AA195800ACDD8F /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = FAA6C5A121AA195800ACDD8F /* AppDelegate.m */; };
+		FAA6C5A721AA195A00ACDD8F /* MainMenu.xib in Resources */ = {isa = PBXBuildFile; fileRef = FAA6C5A521AA195A00ACDD8F /* MainMenu.xib */; };
+		FAA6C5AA21AA195A00ACDD8F /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = FAA6C5A921AA195A00ACDD8F /* main.m */; };
+		FAA6C5B321AA1A4400ACDD8F /* AVFoundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FAA6C5B221AA1A4400ACDD8F /* AVFoundation.framework */; };
+/* End PBXBuildFile section */
+
+/* Begin PBXFileReference section */
+		FAA6C59D21AA195800ACDD8F /* CalinsQRReader.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = CalinsQRReader.app; sourceTree = BUILT_PRODUCTS_DIR; };
+		FAA6C5A021AA195800ACDD8F /* AppDelegate.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = AppDelegate.h; sourceTree = "<group>"; };
+		FAA6C5A121AA195800ACDD8F /* AppDelegate.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = AppDelegate.m; sourceTree = "<group>"; };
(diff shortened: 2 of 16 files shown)


ecdsa:
thanks for this.
I think we should move the CalinsQRReader directory to a separate repository

cculianu:
Ah, I see

Ok -- so it wouldn't exist inside Electrum and just be cloned at build-time?

Where would the repository live, ideally? A new `spesmil`o repo? Or shall I host it on my account `cculianu`?

I'm ok with whatever you like and I can submit another PR that does that. Just let me know where to put the repo.

After that you can close this PR and I can submit another one that:

1. Clones the repo at build-time (from the make_osx script) <-- should I hard-code a specific 'tag'?
2. Compiles it in make_osx
3. Adds it to .app bundle
3. Optionally signs everything (using the just-merged-in #4869).

Sound good?

Let me know!

ecdsa:
we can host the code in spesmilo.
how should the repo be named?

cculianu:
I went with the name 'CalinsQRReader' as a bit of pseudo-narcissistic self-advertising.  If that's ok with you I'll go with that name. Muahaha. 

But if not -- let me know and we can pick something more generic-sounding or whatever you prefer. :)

After you create the repo I can issue a PR that basically contains the code in the CalinsQRReader subdirectory and we can take it from there....

cculianu:
... or I can also create the repo on my end and you can fork it.  Either/or work.

ecdsa:
CalinsQRReader is fine. 
I created a repo here: https://github.com/spesmilo/CalinsQRReader
you can create a PR for that repo.
in the build script, I think that we should define it as a submodule

cculianu:
Awesome.  Will issue a PR to that repo today.

> in the build script, I think that we should define it as a submodule

What does that mean, exactly? Is that a pyinstaller thing? Or? Can you elaborate a little?

ecdsa:
we have two submodules already, that are used for icons and translations, see: 
https://github.com/spesmilo/electrum/tree/master/contrib/deterministic-build

see also https://gist.github.com/gitaarik/8735255

ecdsa:
no, the make_osx script should compile the app, as it is currently done in your PR

cculianu:
Ok, fair enough! :)

Will get to work on this today and hopefully have a PR ready soon.  Thanks for the guidance -- you've been a really good help in steering this in the "right" direction.

I may go back and do it this way for Electron Cash too.. now that I know what submodules are. :)

Thanks again!

cculianu:
@ecdsa  Hey so github won't let me issue a pull request to spesmilo/CalinsQRReader because it's empty.  

I did create a repo here though: https://github.com/cculianu/CalinsQRReader which is pretty much 100% ready for you to slurp up into spesmilo/CalinsQRReader.

Let me know when that's done -- as my upcoming PR will have to reference your repo when it's ready! :)

ecdsa:
ok I forked it

cculianu:
Awesome.

One last question while you're here -- where should the submodule 'live'?

`contrib/`? `contrib/deterministic-build`?  Any other place ?

ecdsa:
contrib is fine. we'll move it if we need.

cculianu:
Ok, closing this PR as the new one has already been submitted.

I'm very happy with the submodule approach! I think I will backport this approach to Electron Cash -- and I'm going to start using it in other projects going forward.  Very organized! (I feel sort of dumb for not knowing about it but.. live & learn!).

PR #4873, opened 2018-11-27
android docker build

SomberNight:
build android apk in docker

I also intend to have travis run this build script, though not necessarily as part of this PR.


ecdsa:
note: we need to be able to pass parameter to make_apk

ecdsa:
crashes with `KeyError: 'ndk_api'`

Issue #4817, opened 2018-10-31
backwards-compatibility for API breakage in python-trezor 0.11


matejcik:
Not an issue in Electrum as it stands, more of a call for assistance.

I'm planning to do a breaking release 0.11 of python-trezor sometime in November.

I'll contribute fixes to the trezor plugin, it should allow simplification of at least some parts of the code. With that I'd like to bump required version to 0.11 for future Electrum releases, because keeping compatibility between 0.10 and 0.11 is going to be a pain.

**This bug** is about possibilities for an upgrade path, esp. in situations when someone upgrades python-trezor on a system with pre-existing Electrum installation.
Unfortunately, Electrum has a lower-bounds check for library versions but not an upper-bounds check. And given the nature of changes, the incompatibility will manifest at client instantiation time and keeping compatibility wrappers in python-trezor would be tricky.
(mixins are gone, only the `TrezorClient` class is supported, UI is not controlled by callback methods anymore but instead through a separate `ui` instance. I could keep around enough objects so that the import statements don't fail, but at some point things are going to break.)

Instead, I was thinking of raising some sort of descriptive exception at some point, which would tell the users that they need to either downgrade python-trezor or upgrade Electrum.

What would be a good injection point for that?

Another option would perhaps be to detect Electrum in python-trezor code and monkey-patch the `check_libraries_available` method set - hopefully in a way that also works in other Electrum forks.

Thoughts, comments, ideas?

SomberNight:
Note: with python-trezor master, there's an error during plugin init, which is not shown to the user (in the GUI) at all atm.
```
[Plugins] loaded coldcard
[Plugins] loaded digitalbitbox
[Plugins] loaded keepkey
[Plugins] loaded ledger
[Plugins] loaded safe_t
Traceback (most recent call last):
  File "...\electrum\electrum\plugin.py", line 149, in get_hardware_support
    p = self.get_plugin(name)
  File "...\electrum\electrum\plugin.py", line 177, in get_plugin
    self.load_plugin(name)
  File "...\electrum\electrum\plugin.py", line 101, in load_plugin
    plugin = p.Plugin(self, self.config, name)
  File "...\electrum\electrum\plugins\trezor\trezor.py", line 90, in __init__
    from . import client
  File "...\electrum\electrum\plugins\trezor\client.py", line 11, in <module>
    TrezorClientBase.wrap_methods(TrezorClient)
[...]

Issue #4404, opened 2018-06-04
Interference when using two Trezors


matejcik:
I'm trying to have two wallets open, secured by two separate Trezor devices.
This *mostly* works. However, it's possible to cause interference in the following scenario:

1. in wallet 1, go to Addresses tab
2. in wallet 2, prepare and Preview a transaction
3. in wallet 1, right-click an address and "Show on TREZOR"
4. while the "please check your trezor" dialog is open, click "Sign" in the transaction preview

expected: second Trezor starts signing

actual: Electrum goes haywire

if the "show address" device is a T1, then the sign attempt switches its display to QR view (which is actually a T1 bug). That allows recovery, because when the address view is confirmed on device, Electrum resumes the sign operation on the other Trezor.

If the "show address" is a TT, there's a bug (possibly TT's) that clicking Cancel raises an "unexpected message". In this case, Electrum never works again. You can close all windows and the UI is responsive (i.e., not frozen), but nothing does anything anymore and the only way out is force close.

I couldn't figure out where the interference is coming from - are the Trezor connections somehow shared under the hood? I _think_ it's not in trezorlib, but it's still possible, I suppose.

SomberNight:
Ahh it is due to `scan_devices`, that breaks the separation..
I introduced the bug in https://github.com/spesmilo/electrum/commit/85b36e027f8cada43aa755fdcef9291ab45a5e06 :/
wallet1 will get pinged when you start signing on wallet2

SomberNight:
The original problem that commit is trying to fix affects all the hardware wallets except trezor; presumably due to the custom enumeration stuff.
I've tested with Trezor One (no), Trezor T (no), Ledger Nano S (yes), Ledger HW.1 (yes), KeepKey (yes), DigitalBitbox (yes).

SomberNight:
Further interference in case of tx signing is here:
https://github.com/spesmilo/electrum/blob/32af83b7aede02f0b9bb3e8294ef2dc0481fb6de/electrum/plugins/trezor/trezor.py#L316-L317

(TrezorPlugin is singleton; shared among the wallets)

matejcik:
i'm fixing that one as part of #4817

matejcik:
Still not sure what to do about the original bug. It might go away by itself if I manage to clean up exception handling.

Or if the problem that's solved through `scan_devices` (which I don't understand very well) doesn't affect Trezor, maybe the solution is just to return True from `has_usable_connection`?

[...]

Issue #3131, opened 2017-10-26
Unexpected magic characters


874gt837g4:
trying to use my trezor with electrum...weird...i can change name by clicking on trezor in bottom right, and name change goes to trezor

but plugged in...i try to sign txs,,,,it says "Unexpected magic characters" and i cannot spend my coins...

any help please?

874gt837g4:
i tried already using diff usb cables, trying restart computer, try usb port different, not working

SomberNight:
What Electrum version?

davidkaufman:
This happened to me a couple weeks ago too after I updated my Trevor firmware to access my bitcoin cash wallet, transfer it to real bitcoin, and then transfer everything to a new segwit address wallet.  At that point Electrum stopped being able to open the tremor wallet.  You have to connect to your Trevor using https://wallet.trezor.io/ now until Electrum gets segwit support

rroopreddy:
I get the same issue

Electrum 3.0.2
Trezor 1.6 (tries with 1.5.2 also)
trezor-0.7.16 (pip3 install trezor on mac)

SomberNight:
Another report of this on IRC. Seems to be MacOS related. Has anyone experienced it on another OS?

```
<kyuupichan> jddebug: have you tried changing the USB port and closing all electrum windows?
<jddebug> I have. I think it may be the way electrum communicates to the trezor on mac
<kyuupichan> I've used on on my mac without issue
<jddebug> What OS version? I'm using a MacPro.
<ghost43> are you using the official 3.0.2 binary jddebug?
<ghost43> or source
<jddebug> I am using the exicutable.
<jddebug> executable
<kyuupichan> I'm running from source on latest MacOSX update.  trezor 0.7.16 hidapi 0.7.99.post21
<ghost43> jddebug: any chance you can try running from source
<kyuupichan> I did see the magic error once, not sure if for trezor or keepkey, but it went away
<ghost43> kyuupichan: or that you try running the binary? :D
...
<kyuupichan> "Unexpected magic characters" when running the binary
<ghost43> kyuupichan: thanks. my working theory is that the trezorlib included in the binary might be an older version
<jddebug> yes
<ghost43> because the version is not fixed im the repository for Mac
<ghost43> only for Windows
<kyuupichan> Also when running from source :(  No clue.
<kyuupichan> However it does seem to function normally
<kyuupichan> It's the init_device call on line 458 of client.py that is causing the error
[...]

Issue #2440, opened 2017-05-23
Sending BTC with 'use multiple change addresses' and trezor asks for confirmation of unexplained movements


artvogel:
- Using Electrum 2.8.2 with Trezor, 
- select 'use multiple change addresses' 
-  initiate a bitcoin-sending transaction 

Observed behavior:
 - Trezor asks user to confirm transfers to change addresses never seen before. User must manually go to the 'Console' pane and run 'listaddresses()' to confirm that the change transfers are to addresses in their control.
 
Expected behavior:
 -  Electrum program will explain all btc transfers in the transaction, so user can easily confirm on the trezor that change is going to the correct addresses.

ptilancelot:
0x7068D71569E9c514805c71008dA69D8F3445225F

ecdsa:
shall we always open the 'preview' window if there are multple change addresses?

prusnak:
For the record, we changed the behaviour of the Trezor firmware to allow multiple change-outputs. See https://github.com/trezor/trezor-firmware/pull/1100 - for 2 change-outputs there is no extra screen, for 3 and more change-outputs we show an extra confirmation screen.

I think this can be closed as the issue will resolve itself once a user has FW 2.3.2+ or 1.9.2+.

PR #4875, opened 2018-11-27
WIP: Trezor 0.11

### LINKED ISSUE #4806: OS X wallet import hardware ID issue
This one is a head scratcher but I'll do my best to describe how to reproduce.

Electrum 3.2.3 on testnet
OS X Sierra
Create 3-of-5 p2sh wrapped segwit wallet
First pubkey: import from Ledger Nano S with derivation path m/49/1/0
Second pubkey: import from Trezor One with derivation path m/49/1/0
Third pubkey: import from Trezor One with derivation path m/49/1/0
Fourth pubkey: import master upub
Fifth pubkey: import master ubpub

The problem occurs when importing the third pubkey (the second Trezor) - I unplug the first trezor immediately after importing its pubkey and plug in the second Trezor. Electrum says it detects a Trezor but it displays the label of the first Trezor rather than the second. If you proceed to import, it seems to import successfully. In fact, the wallet initializes as expected and even derives the correct addresses and finds transactions and balances.

The problem occurs when trying to spend from the wallet. You can successfully sign with the Ledger Nano S and the first Trezor, but then when trying to apply the third signature with the second Trezor, the operation fails because Electrum says it is expecting the first Trezor.

So it seems that the xpubs are all stored correctly, but something is stored incorrectly regarding the actual hardware device ID.

matejcik:
Posting this PR now for review. This works against unreleased python-trezor version 0.11.0, available from git master. There might be bugs.

Fixes:
- #4817
- #4404 (in case it wasn't resolved already)
- python-trezor update should resolve #4806
- #4083 due to rearchitecting the plugin
- #2440 and #3131 presumably gone
- ~~#4130 should be fixed in trezord 2.0.25~~

#4389 might still exist, albeit in a different form

Please wait with merge until 0.11.0 is released.

I also intend to improve outdated firmware detection.
Would you be OK with pinging our server on launch and checking for latest mandatory firmware update? If not, there is also going to be a hardcoded firmware version check in python-trezor.
(Alternately, I plan to implement something like `find_latest_firmware`, we could add a button to the settings dialog that would allow users to fetch latest firmware version.)


SomberNight:
Nice list of fixes! :)

> Would you be OK with pinging our server on launch and checking for latest mandatory firmware update?

You mean this code would be in python-trezor?
In any case, no, please don't do this.

> If not, there is also going to be a hardcoded firmware version check in python-trezor.

Ok, sure. There is already a hardcoded fw version check in the plugin code though? But I guess you want more control over it (and enforce it for all users of python-trezor), so that's fine.

> Alternately, I plan to implement something like find_latest_firmware, we could add a button to the settings dialog that would allow users to fetch latest firmware version.

That's fine too, just don't connect to a central server silently by default.

matejcik:
Understood.

The point of the new firmware checks is to enforce that people don't use old FWs with security problems, that's why I suggested auto-checking. python-trezor itself will not do silent pings. But I'll bundle latest-known-good FW images and check if yours is older than that.
As I intend to do it now, too-old firmware will raise exceptions by default. I'm thinking of making it possible to ignore them but having noisy warning boxes in the UI?

I'm also thinking that maybe opt-in checks would be OK? But implementing that is probably out of my current scope.
(also wondering if we could put firmware information on the blockchain so that you could check anonymously. But that's _really_ out of scope :)) )

SomberNight:
> I'm also thinking that maybe opt-in checks would be OK? But implementing that is probably out of my current scope.
(also wondering if we could put firmware information on the blockchain so that you could check anonymously. But that's really out of scope :)) )

We're not even checking that the version of Electrum itself is up-to-date atm to avoid pinging a central server. These same points you mention came up re that. :P

> As I intend to do it now, too-old firmware will raise exceptions by default. I'm thinking of making it possible to ignore them but having noisy warning boxes in the UI?

I guess it's ~ok not to be able to ignore the error (I don't have a strong opinion), but the error should be displayed in the GUI and not crash the whole app of course.

ecdsa:
hello, what is the ETA for trezor 0.11 ?

matejcik:
@ecdsa most likely next week

matejcik:
status update: python-trezor 0.11 is feature-complete, we're waiting to release it in sync with firmware 2.0.10

Obsolete firmware now raises UserFacingExceptions, which seems to do what we want:
- show device in "select device" list
- disallow creating wallets with it
- disallow any operations when an existing wallet is open

I had to remove the firmware check from `create_client` because then ... then the client would not get created, Electrum would think that there is nothing connected and do all sorts of special handling for "nothing is connected" state. It would be nice if "device exists but we don't want it" was a recognized state in itself.

in addition, turns out #4130 still exists

ecdsa:
note: we probably need to replace client.py with trezor.py in the .spec files, for pyinstaller to include trezorlib

matejcik:
@ecdsa won't it get pulled in through qt -> trezor -> clientbase ?

matejcik:
python-trezor 0.11.0 is out \o/

Leaving the pyinstaller thing up to you, I have no idea how it's supposed to work.
But let me know if I can help

ecdsa:
thanks! we will test it soon

SomberNight:
Thank you for the PR; I think the code became clearer than it was before! :)

Somewhat related, I also pushed https://github.com/spesmilo/electrum/commit/8c3920a0db10d5519bc2435a5354c13aadb5ed1c to master, to improve the error msg the user sees.

PR #4882, opened 2018-11-28
[macOS] QR scanner facility using a platform-native helper app

cculianu:
Note: This PR deprecates #4870, as well as #3485.

Note 2: This PR adds a submodule in `contrib/CalinsQRReader` which basically points to [spesmilo/CalinsQRReader](https://github.com/spesmilo/CalinsQRReader) (helper app Obj-C source code).

This is similar to https://github.com/Electron-Cash/Electron-Cash/pull/977, although this Electrum version builds from source each time and does not include a compiled binary.

--- 
Sample screenshot taken from Electron Cash:

me_ec_test

---

This feature adds much-needed QR scanning capability to Electrum on macOS.

Note that zbar does **not** work on macOS for reading from the webcam.  So -- a different approach was needed.

I did some digging and most other approaches I could find have external dependencies that just make building and packaging too cumbersome.  (See: https://github.com/spesmilo/electrum/pull/3485 for an example of a cumbersome approach).

However -- macOS since 10.7 comes with very good video processing capabilities right in the OS.  This includes the ability to decode QR codes! Why bundle heavy-handed libs when the OS already provides them? (And -- they are **hardware accelerated**, to boot!).

So -- I wrote a small helper app that Electrum can execute to read QR codes from a camera.  The entire app is 80KB compiled -- so it's tiny and on the order of the size of other bundled binaries included on macOS such as `libusb.dylib`. 

The workflow for this app is identical to the 'zbar' approach used on Linux and Windows.

This PR also takes care of bundling and signing the helper app in the built .app & .DMG for macOS.  

The helper app basically creates a window and reads from the default camera device on the system. It will continue to run until either the window is closed by the user or a QR image is read.

#### Rationale for using a Helper App rather than a dylib

1. macOS Mojave and above require special Info.plist keys and annoying permissions from the user to access the camera. It was far simpler to do this as a self-contained app, than make Electrum adapt to this insanity.
2. It's far less invasive to have a standalone app than to build a dylib when it comes to creating GUI windows -- no interference with Electrum's GUI event loop. Everything is self-contained.

#### Workflow for the helper app is as follows (note this is identical to zbar):

1. If a QR image is scanned, it will print the decoded string to stdout and exit.
2. If there is an error detecting the camera, it will show an error message and wait.
3. If the user closes the window without having scanned a QR code (because of an error or s/he changed his/her mind), it will print nothing to stdout and exit.

As such, `lib/qrscanner.py` was modified to call the helper app using `subprocess.Popen` in the 'darwin' case. The 'windows' and 'linux' cases were left unmodified.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "contrib/deterministic-build/electrum-locale"]
 	path = contrib/deterministic-build/electrum-locale
 	url = https://github.com/spesmilo/electrum-locale
+[submodule "contrib/CalinsQRReader"]
+	path = contrib/CalinsQRReader
+	url = https://github.com/spesmilo/CalinsQRReader
--- a/contrib/build-osx/base.sh
+++ b/contrib/build-osx/base.sh
@@ -21,7 +21,7 @@ function DoCodeSignMaybe { # ARGS: infoName fileOrDirName codesignIdentity
     identity="$3"
     deep=""
     if [ -z "$identity" ]; then
-        # we are ok with them not passing anything -- master script calls us always even if no identity is specified
+        # we are ok with them not passing anything; master script calls us unconditionally even if no identity is specified
         return
     fi
     if [ -d "$file" ]; then
--- a/contrib/build-osx/make_osx
+++ b/contrib/build-osx/make_osx
@@ -16,6 +16,7 @@ export PYTHONHASHSEED=22
 VERSION=`git describe --tags --dirty --always`
 
 which brew > /dev/null 2>&1 || fail "Please install brew from https://brew.sh/ to continue"
+which xcodebuild > /dev/null 2>&1 || fail "Please install Xcode and xcode command line tools to continue"
 
 # Code Signing: See https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html
 APP_SIGN=""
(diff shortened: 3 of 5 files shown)


ecdsa:
thanks. It would be nice if you could reduce the impact of your commit in qrscanner.py

You do not need to right-indent all the code in scan_barcode.
Instead you could rename that function, and define scan_barcode conditionally:

`scan_barcode = scan_barcode_osx if sys.platform == 'darwin' else scan_barcode_ctypes`

That makes the pull request easier to read, because one does not have to wonder if you modified something in scan_barcode_ctypes

cculianu:
Yeah it bothered me too that the pull request looked so invasive when viewing the diff.  That's a good approach to it -- it didn't occur to me to do that.  Will do it that way and add a commit!

cculianu:
Ok, new commit pushed.  The diff looks much more aesthetically pleasing now. 👍

cculianu:
hey @ecdsa -- the Travis checks on Windows failed. The build timed out.  I don't think this has anything to do with my changes and/or this PR.  It could just be idiosyncrasies related to travis (I've seen it time out especially on Windows/Wine before).  

I'm going to push out another basically no-op commit to get it to do another build check.

Hopefully it will work this time!

cculianu:
Yay. It worked! :)

ecdsa:
thanks. it looks good

ecdsa:
merged after squashing commits

PR #4885, opened 2018-11-30
Relax the restriction against hardware keystore encryption for multisig wallets

chrisrico:
Only slightly, when there is a single hardware keystore for the wallet.


ecdsa:
if you created a wallet with several hardware devices, you won't be able to open it without the device used for encryption. So, a '2of3' multisig wallet might not be 2of3 anymore :)

chrisrico:
So not only when there's a single hardware keystore, but also when `n == m`?

Or do you have any other suggestions?

SomberNight:
Related: https://github.com/spesmilo/electrum/pull/3346#issuecomment-442262912

I think you might be misunderstanding each other.

In any case...
Note that there is currently no choice given to the user re the type of encryption. So if hardware device encryption is available then you either use that or nothing.
Further complication is "keystore encryption" (i.e. encrypting software keys in the wallet file).

scenario1: 2of2 wallet with electrum seed + trezor. With current PR, you either encrypt with the trezor, or if not, the seed is plaintext in the wallet file.

scenario2: 2of3 wallet with electrum seed + trezor + xpub. If you encrypt with trezor then there is no way to spend without it at all. If you don't, then the seed is plaintext in the wallet file.

IMO encrypting with a hardware device that is also one of the cosigners does not fit well in the general multisig case.

I suppose your use case is a multisig where one cosigner is a hardware wallet and all the others are xpubs. Maybe we could support this special case only.

SomberNight:
> My suggestion to only apply this new logic where m == n means that you'd always need the hardware wallet to spend anyway.
>What I suggest is to encrypt the multisig wallet with the hardware keystore if...
>There is only ONE hardware keystore AND (n == m OR n-1 keystores are xpubs)

I still think we should only allow `ONE hardware keystore AND (n-1 keystores are xpubs)` in the multisig case.
One argument is that it's not necessarily obvious which keystore gets to be "special" (most likely the "first" one, but users don't even necessarily know there is a first one).

I also think that it might be better to "expand" this conservatively; as it's hard to backtrack.
Which actually reminds me: what if someone already has a password encrypted hw multisig wallet (1 hw + xpubs)? I assume they will be able to open it but not sure whether the password dialog will do anything unexpected.

jessicaamber:
Bitcoin recovery, and cryptocurrency recovery in general, requires strong knowledge of blockchain technology, especially in circumstances where your hope is to recover funds from a forked coin. The specific assistance you will require can vary depending upon the reasons your coins are inaccessible, but in every case, having computer experts working for you is key to recovery.

Quick Funds Recovery Co. offers something other Bitcoin recovery services don’t. Because of the collaboration between Bukh Law Firm and CyberSec, you will have some of the world’s most notorious hackers using their knowledge of decryption and password recovery tools to access your locked wallet or your lost coins. Recover all funds lost to binary options scam, Bitcoin scam or basically any stolen money. Contact quickfundsrecovery @ gmail . com
Text/WhatsApp: +1(262)872-0558

PR #4888, opened 2018-11-30
qt: use QStandardItemModel

ysangkok:
This is also fixing a few bugs, like showing the status icon in the invoice list, and making preventing OpenAlias names from being edited (we tried doing this before, but it wasn't comparing the type, but the name itself with the string 'openalias'). Also, descriptions are no longer editable in the invoice list (can still edit in the form), since memos and labels were being updated inconsistently.


ecdsa:
thanks

ecdsa:
merged

Issue #4909, opened 2018-12-09
tx signing with keystore hashing v2 became noticeably slower


SomberNight:
When signing a transaction, if the keystore is encrypted, the key to sign each input is independently decrypted using the password.

Since https://github.com/spesmilo/electrum/pull/4838, this now became noticeable.
The PBKDF is redone for every key.

This is due to calling `self.get_private_key` in the loop below:
https://github.com/spesmilo/electrum/blob/762082e13d7e9db7c1ca603a04d9dc1aebe8b0af/electrum/keystore.py#L113-L124

This is easily fixable for seeded wallets but not for imported.
For seeded wallets, all leaf keys are derived from the encrypted xprv, and so the decrypted xprv can be moved out of the loop.
For imported, this trick does not apply. The random nonce for each key seemingly kills any possible optimisation.

SomberNight:
The slowness is not just about signing. There is a more general issue at hand. The password is being passed around everywhere in the code, and the symmetric cipher key is being derived every time it is needed. Making the KDF expensive is problematic because of this.

All operations that interact with private keys are affected. E.g. importing/exporting/signing.
https://github.com/spesmilo/electrum/blob/c59ac49feadeb6905aef17990b0f93414fd93574/electrum/keystore.py#L162-L173
https://github.com/spesmilo/electrum/blob/c59ac49feadeb6905aef17990b0f93414fd93574/electrum/keystore.py#L178-L184

https://github.com/spesmilo/electrum/blob/c59ac49feadeb6905aef17990b0f93414fd93574/electrum/wallet.py#L1416-L1418
https://github.com/spesmilo/electrum/blob/c59ac49feadeb6905aef17990b0f93414fd93574/electrum/gui/qt/main_window.py#L2489-L2494

SomberNight:
I believe the simplest solution would be to instead of passing around the password (str) everywhere, define a class and pass that around. This class would store the password, but it could also be used to cache the result of the KDF.

ecdsa:
I would rather do the KDF outside of pw_decode, and pass the result where we currently pass the password

ecdsa:
> I would rather do the KDF outside of pw_decode, and pass the result where we currently pass the password

Note: For this to work, we need either no salt or to reuse the same salt across keys. #4838 used a per-key salt.

mr-seifi:
> > I would rather do the KDF outside of pw_decode, and pass the result where we currently pass the password
> 
[...]

PR #4910, opened 2018-12-09
keystore: in sign_transaction, only decrypt once for seeded wallets

SomberNight:
see https://github.com/spesmilo/electrum/issues/4909

this handles the deterministic wallet case


ecdsa:
deprecated by #4937

Issue #4911, opened 2018-12-09
QTreeView + sort model + QStandardItemModel is not efficient performance


wo01:
If there are a few thousands of histories, it takes many seconds to update the list.
Therefore, the window freezes every update.
```
[profiler] HistoryList.update 30.9912
```

I found the following problem.
https://forum.qt.io/topic/55272/qtreeview-sort-model-qstandarditemmodel-the-most-efficient-way-to-update-the-model

ecdsa:
are you on the latest HEAD? how many lines are in your history?

ecdsa:
try the `qabstractitemmodel` branch. we will merge it soon.

wo01:
Yes, I am using latest HEAD.
There are about 3000 lines.
OK, thanks. I will try the branch.

PR #4915, opened 2018-12-10
use QAbstractItemModel in History tab

SomberNight:
fix https://github.com/spesmilo/electrum/issues/4911

### DIFF
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -92,6 +92,7 @@ class HistoryModel(QAbstractItemModel):
         tx_item = self.transactions[index.row()]
         tx_hash = tx_item['txid']
         conf = tx_item['confirmations']
+        txpos = tx_item['txpos_in_block'] or 0
         try:
             status, status_str = self.tx_status_cache[tx_hash]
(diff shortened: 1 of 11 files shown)


electrum/gui/qt/history_list.py
```diff
@@ -247,8 +247,15 @@ def update_fiat(self, row, idx):
         tx_item.update(fiat_fields)
         self.dataChanged.emit(idx, idx, [Qt.DisplayRole, Qt.ForegroundRole])
 
-    def update_item(self, *args):
-        self.refresh('update_item')
+    def update_tx_mined_status(self, tx_hash: str, tx_mined_info: TxMinedInfo):
+        self.tx_status_cache[tx_hash] = self.parent.wallet.get_tx_status(tx_hash, tx_mined_info)
+        try:
+            row = self.transactions.pos_from_key(tx_hash)
```
SomberNight: note: we are not using [QStandardItemModel.findItems()](https://doc.qt.io/qt-5/qstandarditemmodel.html#findItems) as it is not defined for QAbstractItemModel and also as that takes linear time in the number of elements in the list

PR #4916, opened 2018-12-10
 new interface ref. #4540 fix #4676

### LINKED ISSUE #4676: Revealer: long seed has letters obscured by logo in upper-left corner
I created a "digital revealer" and entered a long custom seed. The first two letters of the seed were obscured by the revealer logo in the upper-left corner of the generated png.

tiagotrs:
new interface

ref. #4540 
fix #4676

### DIFF
--- a/electrum/plugins/revealer/qt.py
+++ b/electrum/plugins/revealer/qt.py
@@ -79,6 +79,8 @@ class Plugin(BasePlugin):
             self.cseed = keystore.get_seed(password)
             if keystore.get_passphrase(password):
                 self.extension = True
+            else:
+                self.extension = False
         except Exception:
             traceback.print_exc(file=sys.stdout)
             return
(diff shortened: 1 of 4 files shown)


electrum/plugins/revealer/qt.py
```diff
@@ -28,20 +28,20 @@
 
 from .hmac_drbg import DRBG
 
+from btchip.btchipComm import getDongle
```
SomberNight: This import will fail if the user does not have btchip-python installed, e.g. on Linux. This would mean they cannot use the Revealer plugin at all.

You should import inside a `try ... except`. It seems atm `getDongle` is only used in one place, so the import could be moved there. (in general it's preferred to do imports at the module level at the top of the file, but conditional imports are an exception I would say)

tiagotrs: Thanks! and sorry, the ledger support commit is not ready and shouldn't be in the PR at this stage. follow up commits coming.

SomberNight: then if you want the other stuff merged faster, you could force-push remove the ledger commit

SomberNight:
I've force-push removed the merge-commit, and made some minor fixes/clean-up in a follow-up commit.

SomberNight:
not directly related, but also cleaned up path handling a bit in https://github.com/spesmilo/electrum/commit/91ef367176eeb516e6baee2923222078d938f702

PR #4944, opened 2018-12-19
Increase timeout when using --oneserver and --server

chris-belcher:
Usually if an Electrum protocol message times out then the client will switch to another server, which is hopefully faster. But when --oneserver and --server are both used then the client will reconnect back to the same slow server. The timeout interrupts the protocol messages and doesn't fix the problem. What's needed is for the client to tolerate a slow server if the user specifically configured for that single server to be used.

My motivation for this patch is because of an issue with Electrum Personal Server (see https://github.com/chris-belcher/electrum-personal-server/issues/46), where many people run the software on slow hardware and/or connect via a slow tor hidden service or VPN. The patch shouldn't affect any other kind of user.

Let me know if there's a better way to do this, and I'll be happy to recode the patch. I did consider creating a new command line flag called something like `--tolerateslowserver`.

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -59,6 +59,7 @@ class NotificationSession(RPCSession):
         self.subscriptions = defaultdict(list)
         self.cache = {}
         self.in_flight_requests_semaphore = asyncio.Semaphore(100)
+        self.tolerate_slow_server = False
 
     async def handle_request(self, request):
         # note: if server sends malformed request and we raise, the superclass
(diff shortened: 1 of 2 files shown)


SomberNight:
Thanks for this. There is actually another timeout used when searching block headers (if tip did not connect; binary search / exponential backward search); and another re transaction broadcast. I think we should collect them into one place and then it's also easier to do what you suggest.

SomberNight:
see https://github.com/spesmilo/electrum/pull/4945

chris-belcher:
I'm closing this because #4945 is better.

PR #4945, opened 2018-12-20
network: put NetworkTimeout constants together

SomberNight:
based on https://github.com/spesmilo/electrum/pull/4944


chris-belcher:
Thanks for this. I've reviewed the code and I think it's good. I've also tested it by adding print statements to print out the timeout values and running Electrum with and without the combinations of `--oneserver` and `--server`.

PR #4950, opened 2018-12-20
Replace EncumberedTransaction with LightningTransaction

ysangkok:
This was tested on top of the ln-ctx-consistency branch.


ysangkok:
@ecdsa I have added one commit to remove LightningTransaction.

ecdsa:
1. Why does this pull request insert  `if sweep_tx is None: return None` before `return sweep_tx`? Did you take the time to review your diff? 
2. Why does lnwatcher contain cryptic variable names such as 'e_tx' and 'encumbered_sweep_txns'? don't you understand that such mysterious names are impainring reading, now that the EncumberedTransaction type has been removed?
3. If you had taken the time to squash the multiple commits of this PR into a single commit, maybe 1). would not have happened (because you might have examined the diff after squashing)

PR #4957, opened 2018-12-22
Add flag --skipmerklecheck

chris-belcher:
The `--skipmerklecheck` optional flag makes Electrum tolerate invalid merkle proofs from the server. This is useful for building Electrum servers that need a minimum amount of storage, though of course users should only enable it if they completely trust the connected server.

This is primarily motivated so that Electrum Personal Server can be compatible with pruning. Users who have a pruned full node would enable `--skipmerklecheck`. Also it would make it much easier to create a "no history" mode where the rpc call `scantxoutset` (https://github.com/chris-belcher/electrum-personal-server/issues/27) is used to find a wallet balance without needing a long rescan.

The way this works is based on an IRC conversation with ghost43 who suggested this method, saying it's a little bit hacky but is easy to test and maintain, and is preferable to modifying the code in 10 different places for something that wouldn't be tested very often.


chris-belcher:
Could I get a concept ACK ?

SomberNight:
Tested this now; looks okay.

Note that when this option is enabled, the wallet file will get modified as if the present transactions have been SPV verified (these transactions will not be rechecked even if the user disables the option and connects to different servers). The assumption is that the user only uses this option when connected to a server they fully trust.

PR #4967, opened 2018-12-26
wallet: randomise locktime of transactions a bit. also check if stale.

SomberNight:
more complex logic for setting locktime of new transactions
- check if chain tip is stale
  see https://github.com/bitcoin/bitcoin/pull/15039
- randomise locktime occasionally
  see https://github.com/bitcoin/bitcoin/blob/df73c23f5fac031cc9b2ec06a74275db5ea322e3/src/wallet/wallet.cpp#L2595-L2600

### DIFF
--- a/electrum/blockchain.py
+++ b/electrum/blockchain.py
@@ -426,6 +426,11 @@ class Blockchain(util.PrintError):
             return None
         return deserialize_header(h, height)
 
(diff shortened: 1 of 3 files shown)


chris-belcher:
I've reviewed the code and it looks good to me.

PR #4985, opened 2019-01-01
[Qt] Add optional update notification

bauerj:
This adds an option to check for updates to Electrum through the Qt GUI:
* automatically on startup (Opt-In)
* manually, through Help -> Check for updates

The current source of information for new versions is electrum.org/version which was previously used by the update check mechanism which has been removed in 2.5 due to privacy concerns.

I propose to re-add such a feature because I think the security benefits outweigh the potential privacy implications. In case of security updates it is crucial that users install them but currently they have to find out about the update through some other channel: by following Electrum on Twitter or Reddit or by randomly checking the website.

Since the feature is opt-in, users can decide themselves whether they want this.

An alternative would be to store the information on the blockchain instead and retrieve it through the Electrum server connection.

Screenshots:
===========

Checkbox added to preferences. Appearance tab renamed to general

-----

This is only being shown when the user manually opens the update check through the Help menu. If the check fails for whatever reason a message is being shown.

----

This is only being shown when the user manually opens the update check through the Help menu. While the version information is retrieved, a progress bar indicates that it's still running.

-----

This will be shown both through the help menu and by clicking the notification in the status bar.

-----

image


bauerj:
This will only notify the user about a new version, the download location is hard-coded into Electrum. A malicious response could only annoy users by showing a version that is not available yet.

doremon1:
Looks like you misunderstood me.

Consider a scenaria where electrum.org got hacked and the attacker replaced binaries with malware and also https://electrum.org/version to a heigher number inorder to bring users to download the malware from electrum.org. Now since we didn't have any mechanism to check weather this version number on electrum.org is legitimate people think that theres really a new version and they'll visit the official hacked site to download. 

The good thing is this attack fails when users verify GPG signs but consider people directly running binaries without verifying.

What i'm suggesting is either encrypt this version with private keys of developers and hardcode public key in client. So that the client can verify weather this version is really relased by developers.

ecdsa:
GPG is too complex for most users. Maybe we could sign binaries and new version announcements with an EC key, so that the signature could be verified by Electrum in a user friendly way.

SomberNight:
See https://github.com/spesmilo/electrum/compare/update_notifications branch, that adds authentication for the update notifications.
Specifically https://github.com/spesmilo/electrum/commit/08499af9f168e82bd34d117e49009ddba3aa2b68

we would hardcode one or more bitcoin addresses (or ec pubkeys) in the source
messages would be signed by one or more of these
(having a list of signatures allows rolling the key; i.e. if one gets compromised, new message can be signed with both old and new key)

bauerj:
Looks good! I still think it would make sense to add functionality that downloads and verifies installer binaries too.

SomberNight:
merged in https://github.com/spesmilo/electrum/compare/5613f9b903c6da799ec461d3a06b25ea0d0ee577...0bfda7c8c74757d261bbc7e24eee44fa09965e85

PR #4991, opened 2019-01-06
Fix ledger transaction version

tomix1024:
Use the correct transaction version when signing using a ledger hardware wallet.
By default, transactions created by electrum are currently version `1`, which is also the default transaction version assumed by `startUntrustedTransaction`.

However, when signing a version `2` transaction with a ledger device, the correct version should be passed through in order to obtain a valid signature for the transaction.


SomberNight:
Thanks.

PR #4997, opened 2019-01-09
Give visual feedback while starting

sajolida:
I'm working for Tails (https://tails.boum.org/), a privacy-oriented
Linux distribution which includes XXX.

While doing usability testing for Tails or getting feedback from
users, we realized how important it was to give feedback very quickly
after the user chooses to start an application. The usual mechanism
for this in GNOME is to change the mouse cursor into a spinner as soon
as the user chooses the application from a menu or from the activities
overview.

For example, we've seen time after time people opening the same
application several times when its launcher lacked this feedback.

XXX doesn't provide this feedback but I think it's easy to fix.

Even when an application usually starts fast, there might be some
circumstances when it will be a bit slower to start (live operating
systems like Tails is one of them but it might happen on any busy
system).

The Freedesktop specification for such feedback is to uses a
combination of `StartupNotify` and `StartupWMClass`.

See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html


bauerj:
What is XXX?

ecdsa:
I guess XXX is 'Electrum' . The message with this PR might be a generic message, posted on all projects that do not implement StartupNotify

sajolida:
Hey! Indeed, I sent PR to several projects included in Tails and forgot to edit my template properly.

I fixed this now in the text. We use a lot of Git at Tails but never GitHub :)

Sorry for the confusion!

bauerj:
No worries, I'm all for working efficiently!

I validated the file with `desktop-file-validate` so it seems like the syntax is correct. Can anyone test this?

SomberNight:
I've tried to test this but could not notice any difference.
I think the spinner does not show up as the application process starts very fast (even though in some cases the GUI window takes some time to appear) -- is there supposed to be a spinner while the window is not visible?

Looking at the spec, the change looks okay; and based on [this askubuntu thread](https://askubuntu.com/questions/367396/what-represent-the-startupwmclass-field-of-a-desktop-file), StartupWMClass also seems ok (note that when running from source it's actually `run_electrum` but in that case you wouldn't use the .desktop launcher I guess).

In any case, I think we can merge this.

ecdsa:
I also could not notice any difference. I guess merging was ok

PR #5010, opened 2019-01-17
add command for creating change addresses

jlopp:
This has come in handy for me several times while recovering development funds that were outside of gap limits.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -659,6 +659,11 @@ class Commands:
         return self.wallet.create_new_address(False)
 
     @command('w')
+    def createnewchangeaddress(self):
+        """Create a new change address, beyond the gap limit of the wallet"""
+        return self.wallet.create_new_address(True)
+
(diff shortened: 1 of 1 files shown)


SomberNight:
How did you get coins outside the gap limit in the first place?
Would it perhaps be more useful to be able modify the gap limit for change?

jlopp:
Mycelium ended up increasing their gap limit from 1 to 20, though it's more of a band-aid than a real fix. I expect that this command would only be needed by developers anyway.

SomberNight:
_I think Electrum would not create gaps in change in such a way. New change addresses are only created after existing ones have funds with 3+ confirmations, so you would need reorgs for a gap to develop._

jlopp:
For sure, though recall that we sometimes see reorgs in excess of 1,000 blocks on testnet - I'm pretty sure one of those crazy events is what screwed up my wallet.

Issue #5033, opened 2019-01-25
Document things to do before tagging a release


JeremyRand:
There isn't currently any documentation of what steps need to be taken prior to tagging a new release; documenting this would be useful.  Off the top of my head, I would guess that updating the server list and updating the checkpoints would both be on the list; I might be forgetting some other things though.

AbdussamadA:
delete any zero byte files. there's one in the tarball under the electrum subdir called voegtlin@electrum.2717:1544126902

JeremyRand:
Presumably bumping the version would also belong on the list.

JeremyRand:
Running `./contrib/sign_version` also presumably belongs on the list.

Issue #5030, opened 2019-01-25
Release process for GNU/Linux binaries not documented


JeremyRand:
The release process for GNU/Linux binaries isn't clearly documented.  It would be highly useful to clearly document what terminal commands to run, starting with `git clone` and ending with the `.tar.gz` file that is uploaded to the website.

bauerj:
You can check out https://github.com/spesmilo/electrum/blob/master/contrib/make_tgz, the process is fairly simple.

Issue #4424, opened 2018-06-10
Detritus in release tarball


mithrandi:
The release tarball seems to have some junk:

```
scripts/__pycache__/lntest.cpython-36.pyc
scripts/__pycache__/util.cpython-35.pyc
scripts/util.pyc
```

Could we clean this up?

ecdsa:
fixed in 3.2.0

PR #5040, opened 2019-01-28
contrib/make_tgz: small improvements

SomberNight:
~~The release tarballs should contain the compiled Qt icons file.
The need for this was previously not articulated anywhere AFAICT.~~
 
Also, occasionally some garbage files from the build machine get included into the tarball (https://github.com/spesmilo/electrum/issues/5033#issuecomment-457930965 and https://github.com/spesmilo/electrum/issues/4424), so dry-run a `git clean` and show the user the results.

related: https://github.com/spesmilo/electrum/issues/5030

### DIFF
--- a/contrib/make_tgz
+++ b/contrib/make_tgz
@@ -1,11 +1,31 @@
 #!/bin/bash
 
-contrib=$(dirname "$0")
-packages="$contrib"/../packages/
+set -e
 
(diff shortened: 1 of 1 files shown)


SomberNight:
Removed the Qt icon building part as it is no longer needed after https://github.com/spesmilo/electrum/pull/5053
The rest is still useful though.

PR #5067, opened 2019-02-04
android_all_currency and add execution files for win, android

kpstar:


SomberNight:
What's the rationale?
Why would we merge binaries?

ecdsa:
don't even ask why...

PR #5069, opened 2019-02-05
move lightning message encoding to new lnmsg module

ysangkok:


ysangkok:
used singleton pattern [like in cpython](https://github.com/python/cpython/blob/master/Lib/random.py#L779) and pushed in 7229982d4d8c14ae98ac2a0ad824bc5e5f83fb1d

PR #5111, opened 2019-02-12
Anti-Phishing: It appears loading tx from blockchain was vulnerable.

cculianu:
It printed server messages to client.  See screenshot:

screen shot 2019-02-12 at 5 15 00 pm

This hole has now been plugged using the normal scheme: transalate server messages into user messages, falling back to something generic if not recognized.

screen shot 2019-02-12 at 5 15 42 pm


SomberNight:
indeed, thanks. @kyuupichan disclosed this yesterday to us
I am working on this atm; trying to solve it in a more general way...

cculianu:
Yeah ideally it would be down in the network layer that way all UI versions (iOS, kivy, etc) could benefit (sort of the way the broadcast tx code works).

I already merged this into EC but if you guys come up with something clever will steal. ;)

PR #5133, opened 2019-02-17
Sync android multisig wizard with desktop

noone2019:
On Electrum desktop, up to 15 keys can be specified in the multisig wizard. This changes syncs android version with desktop version.

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/installwizard.py
+++ b/electrum/gui/kivy/uix/dialogs/installwizard.py
@@ -137,7 +137,7 @@ Builder.load_string('''
             text: _('From {} cosigners').format(n.value)
         Slider:
             id: n
-            range: 2, 5
+            range: 2, 15
             step: 1
             value: 2
         Label:


SomberNight:
I am not sure why it is currently limited to 5 on Android but I imagine there is a reason. @ecdsa any idea?
My guess: the assumption is that mostly QR codes are used to transfer the partial transactions, and QR codes simple cannot fit too much data. (the more cosigners, the larger the partial txn is)

Do you have a multisig wallet you would like to use on Android but can't because of this?

noone2019:
Yes. I checked desktop version, it supported up to 15 keys, I decided to use it to create a multisig wallet. Then I tried to use Android version, and then I found that it only supports up to 5 keys.

noone2019:
Hey, will you consider merging this?

If the problem is QRCode, a solution is to display a warning to the user on the creation wizard that QRCode might not work. Another solution might be to hide QRCode button when data is larger than a certain amount and instead show a export/copy/share button to export the transaction data.

ecdsa:
QR code size is indeed a problem.
Also, the slider in the install wizard would become more difficult to use.
I would be curious about the use cases for multisig wallets with more than 5 keys

noone2019:
Hey @ecdsa (what a cool username!),

First, thanks for Electrum. Great program.

The slider on mobile is currently in front of the label, which make is half the width of the screen. It can be moved to the line below the label and take the whole screen width. 

Anything I say for the reason will not convince you. However, if you think more than 5 keys is useless, why Electrum on desktop have a max of 15?

On desktop I can create up to 15 multisig wallets, and it works, and when I want to use the QRCode for a large transaction, it shows me an error indicating that it is too large for QRCode. It's fair and I can export the text data or save it to a file.

This can happen on mobile too. Users who don't know what they are doing, will have a hard time anyway. And users who want to use a feature in the protocol, can use it.

I opened this because I have created a wallet on desktop, but now I can't use it on mobile, and QRCode is not a problem if I just copy the hex output. This is not to promote more signatures, but to make Electrum on mobile have the same functionality of the desktop version.

Issue #5061, opened 2019-02-02
Ignoring/hiding/blacklisting/fighting dust attack/spam inputs/UTXO


birdie-github:
Sometimes your receive transactions like this one:

https://www.blockchain.com/btc/tx/6e41f6acb7541cf4ffeddf126dbd87a01092be8d021a897350a6a332ba67a071

And it's quite expensive to spend them (actually **more** expensive than not to spend them at all).

It would be great if Electrum allowed to blacklist/hide/disable such inputs/UTXOs.

iPherian:
I agree. Or let you know by highlighting the transaction red or something.

wtoung:
please help restore my stolen fund due to electrum wallet which i didn't open for months.
shocking to see my small lost from a PC, which i didn't touch for months.

or is that electrum itself steal our funding !! sorry, i may not know culture of dev team, but want to trust electrum developers are smart, good and honest people.

https://www.blockchain.com/btc/tx/bb08b3ed52955da3c98562638e5d1486328995e029f8b33ce35320deeddcd0b3

any clue restoring? Thanks.

AbdussamadA:
Are you sure the cost of spending these utxos is high enough to worry about? I mean if they are included with other utxos in a transaction then it might not matter much. 

If you want to exclude them here's how. You can select specific inputs to a transaction by going to view menu > show coins and ctrl+clicking them on that tab to select them, right click and choose to spend. So simply spend any other utxos sent to the address where the dust was sent and then on the addresses tab right click on that address and choose to freeze it. Freezing means coins sent to that address are ignored by electrum. This way the dust will never be spent.

birdie-github:
@AbdussamadA 

Your workaround looks overly complicated for average people. It would be so much better to be able to go to the addresses tab, right click a bad transaction and choose something like freeze/hide/blacklist/etc. so that you won't be able to spend it ever.

Also your workaround implies that the user has to move his funds on the blockchain in order to combat dust spam. This is just inefficient/complicated/prone to errors or/and funds loss and generally should be avoided.

AbdussamadA:
- Ordinary people don't care about dust spam. 

- Addresses tab doesn't list transactions. 

- You don't have to move your coins now. You can simply do this the next time you spend your bitcoins.

If you want to be able to freeze specific utxos then you're asking for a feature to be added to electrum. In that case you should edit the OP and the issue title to reflect that.

SomberNight:
[...]

PR #5152, opened 2019-02-24
Freeze individual UTXOs

SomberNight:
Implements new feature where users can freeze individual UTXOs.
This is orthogonal to addresses being frozen, and for the wallet to automatically select a coin to spend, the coin must be unfrozen both ways.

somewhat based on https://github.com/Electron-Cash/Electron-Cash/pull/891

Not clear how to make UI intuitive; for now I've opted to use foreground colour for UTXO frozenness, and background colour for address frozenness.

capture

related: https://github.com/spesmilo/electrum/issues/5061

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -25,7 +25,7 @@ import threading
 import asyncio
 import itertools
 from collections import defaultdict
-from typing import TYPE_CHECKING, Dict, Optional
+from typing import TYPE_CHECKING, Dict, Optional, Set, Tuple
 
 from . import bitcoin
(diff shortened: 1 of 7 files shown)


SomberNight:
> Not clear how to make UI intuitive; for now I've opted to use foreground colour for UTXO frozenness, and background colour for address frozenness.

Instead of that, I've now changed to using the existing blue background notation, but in the "Address" column if the address is frozen; and in the "Output point" column if the UTXO is frozen.

Capture

cculianu:
Ha, I had the same dilemma as to how to make it intuitive. Glad I'm not the only one.  Looks good to me.. ;)

Issue #5157, opened 2019-02-25
Android Fiat Currency "Some currency not existing as desktop".


kpstar:
Desktop has all currencies in setting/Fiat Currency with history options.
But android doesn't have some currencies because history option is set True by default.

I am about to use 'COP' currency but can't find in android. How can I use? Thanks.

SomberNight:
> But android doesn't have some currencies because history option is set True by default.

Yes, you are correct.

What exactly is the suggestion? To allow toggling the history option like in Qt?

kpstar:
![screenshot_28](https://user-images.githubusercontent.com/36706893/53345738-8b5ff780-3950-11e9-925b-47f463ba7c9b.png)

kpstar:
Thanks for your reply, @SomberNight 
I tried to change like this as attached image and installed my phone.
It seemed there was no error regarding this action.
I am just wondering why you disable history option.

I am also a mobile developer and want to contribute to your project.
I wish you would give me some opinions. Thanks.

SomberNight:
In Kivy, on the "Balance" tab, if you tap your balance at the top, all the values change between BTC and fiat.
This is only possible with historical conversion rates. Thus, for simplicity, only currency/exchange combos that offer historical rates can be selected.

kpstar:
Isn't there any way for all currencies in android?

kpstar:
For example, setting checkbox for history option like desktop.
Because I use 'COP' currency and mobile is essential.

SomberNight:
You could add the same "show fiat history rates" option to settings like in Qt, enabled by default.

accumulator:
qml client uses FX like desktop, closing.

PR #5162, opened 2019-02-26
android all currencies with history checkbox

kpstar:
I inserted history checkbox option for all currencies in android.


SomberNight:
note: this is regarding https://github.com/spesmilo/electrum/issues/5157

purpurato:
I would like this feature, when  are you guys planning to include it?

SomberNight:
Thank you for the contribution.

PR #5210, opened 2019-03-19
fix sign message not accept rich text

zebra-lucky:
When copying text of message to sign from rich text editors, there is rich formating copied together with text and confusions possible during verify process. This patch disables `acceptRichText` on QTextEdit field for message.


SomberNight:
Should this also be applied to message encryption/decryption?

zebra-lucky:
Maybe, I will check it now.

PR #5228, opened 2019-03-28
Strip whitespace from amounts when copying. 

AbdussamadA:
Fix bug in request list where address was copied instead of column specific data


SomberNight:
Thank you

PR #5296, opened 2019-04-27
use stdlib logging module instead of print_error

SomberNight:
- use stdlib logging module
- remove PrintError and print_error
- log to file by default, with loglevel `DEBUG`
    - can be disabled with flag `disablefilelogging`
    - one file per process lifecycle; latest ~10 files kept
- log to stderr; by default only `WARNING` and higher
    - if `-v` is used then `DEBUG`
- old style `-v` works
    - but more advanced filtering options are available:
        ```
        -v=debug,network=error,interface=error      // blacklists network and interface
        -v=warning,network=debug,interface=debug    //  whitelists network and interface
        ```

@cluelessperson your original commit https://github.com/cluelessperson/electrum/commit/106e20815c6f6967c9bbc495ec8d2eee5e054df7 is not properly attributed by github
do you want me to amend it so that it includes your email? (if so, what email?)

### DIFF
--- a/electrum/__init__.py
+++ b/electrum/__init__.py
@@ -1,5 +1,5 @@
 from .version import ELECTRUM_VERSION
-from .util import format_satoshis, set_verbosity
+from .util import format_satoshis
 from .wallet import Wallet
 from .storage import WalletStorage
 from .coinchooser import COIN_CHOOSERS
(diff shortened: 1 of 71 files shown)


SomberNight:
Note: sometimes we log pubkeys/addresses, so with file logging being default on, some might consider this a privacy leak. Should mention this in changelog, along with `disablefilelogging` option.

PR #5299, opened 2019-04-27
new domain for luggs servers

luggs-co:
Updating server address to new domain, from elec.luggs.co to ex.lug.gs

### DIFF
--- a/electrum/servers.json
+++ b/electrum/servers.json
@@ -121,7 +121,7 @@
         "t": "50001",
         "version": "1.4"
     },
-    "elec.luggs.co": {
+    "ex.lug.gs": {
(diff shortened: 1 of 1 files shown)


bauerj:
Why this change? Is the server still available at the old domain?

luggs-co:
> 
> 
> Why this change? Is the server still available at the old domain?

it won't be no.

bauerj:
Are you still in control of the old domain?

luggs-co:
> 
> 
> Are you still in control of the old domain?

no

bauerj:
I assume you sold your domain to the ones that run the current phishing attack against Electrum users. Is that correct?

PR #5311, opened 2019-05-04
Allow retrieving transaction fee also for confirmed transactions

romanz:
Currently, it shows the fee from `JsonDB` only for mempool transactions.
Please let me know if there is a better way to resolve this issue (e.g. let Electrum query the server for input transactions and calculate the fee by itself).

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -364,6 +364,9 @@ class Abstract_Wallet(AddressSynchronizer):
         can_bump = False
         label = ''
         tx_hash = tx.txid()
+        if fee is None:
+            fee = self.db.get_tx_fee(tx_hash)
+            self.logger.debug('%s transaction fee: %d sat', tx_hash, fee)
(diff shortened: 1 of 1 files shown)


SomberNight:
> let Electrum query the server for input transactions and calculate the fee by itself

Yes... well the issue there is that it might significantly increase bandwidth usage / sync times.
Trusting the server with fee data is a compromise.

I don't believe this patch actually solves the issue.
Check out the protocol docs for [blockchain.scripthash.get_history](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-scripthash-get-history).
The fees are only included for mempool transactions.
So this won't always work.

romanz:
Got it, many thanks for the explanation.

Since I am running my own Electrum server, I wanted to [add](https://github.com/romanz/electrs/compare/fees) a `fee` field to the `blockchain.scripthash.get_history` API response, but I understand that this solution is problematic for public Electrum servers.

Issue #2094, opened 2017-01-04
improve wallet synchronization


ecdsa:
synchronization of large wallets is slow.
synchronizaion could be separated in stages, with the existing protocol:
1. address generation. this requires only address.get_status
2. request UTXOs for each address
3. request transactions with UTXOs
4. request merkle branches
5. request full histories
6. request spent-only transactions
the wallet balance can be displayed when 3 is completed.

bauerj:
Once this is done, we can also show a progress bar to the user, indicating how far the synchronization has proceeded.

SomberNight:
in the context of potential protocol changes to help this, tangentially related:
https://github.com/kyuupichan/electrumx/issues/82
https://github.com/kyuupichan/electrumx/issues/348

PR #5312, opened 2019-05-05
synchronizer: show progress in GUI (take 1) (num_addrs, num_txs)

SomberNight:
I think the user should be shown some details regarding the progress/state of the history sync.
This has become even more relevant recently, with servers getting DDOS-ed and as a consequence becoming slower/less responsive.

Instead of `Synchronizing...`,
we now display `Synchronizing... ({num_subs} addrs, {num_txs} txs)`
- where `num_subs` is the number of addresses for which the client sent `blockchain.scripthash.subscribe` and received the initial scripthash status from the server,
- and `num_txs` is the number of "raw hex" transactions that are already stored in the wallet 

sync1
sync2

(I know, the kivy version is somewhat ugly; I can try to improve it)

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -582,6 +582,13 @@ class AddressSynchronizer(Logger):
     def is_up_to_date(self):
         with self.lock: return self.up_to_date
 
+    def get_history_sync_state_details(self) -> Tuple[Optional[int], int]:
+        num_subs = None
+        if self.synchronizer:
+            num_subs = self.synchronizer.num_subscriptions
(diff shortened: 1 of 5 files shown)


bauerj:
Wow, cool! Any reason not to show a progress bar?

SomberNight:
> Any reason not to show a progress bar?

What would you consider 100%? You don't know how many addrs/txs there will be :/
Showing that progress is being made is still a step in the good direction at least :D

ecdsa:
related: #2094

SomberNight:
I think https://github.com/spesmilo/electrum/pull/5319 is superior, closing.

PR #5336, opened 2019-05-12
Script to create a debian package

rllola:
Added some script to create a debian package.

### DIFF
--- a/contrib/build-linux/pkg-debian/DEBIAN/prerm
+++ b/contrib/build-linux/pkg-debian/DEBIAN/prerm
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-rm /usr/bin/ZeronetBrowser
+rm /usr/bin/electrum
--- a/contrib/build-linux/pkg-debian/DEBIAN/prerm
+++ b/contrib/build-linux/pkg-debian/DEBIAN/prerm
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+rm /usr/bin/ZeronetBrowser
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ contrib/build-wine/tmp/
 contrib/build-wine/fresh_clone/
 contrib/build-linux/appimage/build/
 contrib/build-linux/appimage/.cache/
+
+contrib/build-linux/pkg-debian/usr/
+contrib/build-linux/pkg-debian/DEBIAN/control
--- a/contrib/build-linux/pkg-debian/DEBIAN/postinst
+++ b/contrib/build-linux/pkg-debian/DEBIAN/postinst
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+ln -s /usr/share/Electrum/run_electrum /usr/bin/electrum
--- a/contrib/make_deb
+++ b/contrib/make_deb
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+echo "====== GET VERSION ======"
+VERSION="$(python3 -c "import electrum.version; print(electrum.version.ELECTRUM_VERSION)")"
+echo $VERSION
+
+echo "====== COPY SOURCES FILES ======"
+mkdir -p contrib/build-linux/pkg-debian/usr/share
+mkdir -p contrib/build-linux/pkg-debian/usr/share/applications/
+cp electrum.desktop contrib/build-linux/pkg-debian/usr/share/applications/
+unzip -q dist/Electrum-$VERSION.zip -d contrib/build-linux/pkg-debian/usr/share/
+mv contrib/build-linux/pkg-debian/usr/share/Electrum-$VERSION contrib/build-linux/pkg-debian/usr/share/Electrum
+
+echo "====== GENERATE CONTROL FILE ======"
+cat > $PWD/contrib/build-linux/pkg-debian/DEBIAN/control << EOF
+Package: Electrum
+Version: ${VERSION}
+Architecture: all
+Essential: no
+Section: web
+Priority: optional
+Maintainer: Thomas Voegtlin
+Description: Lightweight Bitcoin Wallet.
+EOF
+
+echo "====== CREATE DEBIAN PACKAGE ======"
+dpkg -b contrib/build-linux/pkg-debian dist/Electrum-$VERSION.deb


contrib/build-linux/pkg-debian/DEBIAN/prerm
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+rm /usr/bin/ZeronetBrowser
```
JeremyRand: I assume this wasn't intentional, since it's not related to Electrum?

rllola: Indeed, my bad I have done it for another project and did a copy/past. This has been corrected now. Thanks

SomberNight:
Thanks, but do we need this? What is the advantage compared to the tar.gz?

JeremyRand:
> Thanks, but do we need this? What is the advantage compared to the tar.gz?

@SomberNight `.deb` files are the standard way to install software in Debian and all downstream GNU/Linux distros such as Ubuntu.  They integrate with package managers, can automatically install Start menu shortcuts, allow uninstalling and upgrading easily, don't force users to remember what directory they extracted the .tgz to, etc.  It's similar to why Windows users often prefer an NSIS installer compared to a `.zip` file.

rllola:
It will automatically install the source files in `/usr/share/Electrum` and the `electrum.desktop` file in the correct emplacement so it can be started from the launcher.
It reduce the number of manipulations for the installation and have an uninstall script to keep it clean.

Is it really needed? I would say no. But is it convenient to have it? I would say yes as it is a current way to distribute package and contribute to the one step install concept.

JeremyRand:
> Is it really needed? I would say no. But is it convenient to have it? I would say yes

I'd go slightly further than this, as I have users downstream of me (GNU/Linux distro maintainers) who want to bundle Electrum-based software with their distro by default, and for whom `.deb` packages are a requirement for their workflow.  That doesn't exactly *require* that the packaging scripts live in the Electrum repo, but it would certainly be a lot cleaner if they do.

JeremyRand:
@rllola High level review: it's not clear to me whether this approach to producing `.deb` packages is the best one.  Why not use `setup.py` to create deb/rpm packages?  (Not saying setup.py is better, just that I'm curious what you think the advantages/disadvantages are.)

rllola:
> @rllola High level review: it's not clear to me whether this approach to producing `.deb` packages is the best one. Why not use `setup.py` to create deb/rpm packages? (Not saying setup.py is better, just that I'm curious what you think the advantages/disadvantages are.)

I am afraid I don't have a good answer for that. I just learned this approach for another project and thought I could contribute with this recent acquire knowledge to improve electrum. I am open to explore the use of `setup.py` for it.

ldz1:
@SomberNight: “Thanks, but do we need this?”

I would like to ask the same question, but I would put the focus elsewhere. I understand the advantages of the deb packages, but what have these packages to do with Electrum developers / with Electrum source code? 

Should these packages serve as template for deb-based distributions? I believe they should not, because the distro-maintainers know their stuff much better and they are responsible for the package integration with their distros. I believe the task of electrum-developers is to produce the electrum source code and to provide general information about the build process. They don’t need to prepare build information for other (maybe 200 or 300) distros. 

I am Gentoo Linux user. There is a package manager called portage and the gentoo community has their own electrum ebuilds. In general electrum community knows nothing about these ebuilds and I think that’s a good thing, because they do not need it. 

I would imagine a brief piece of information in the form of a small documentary on how to install the package on other systems. In my opinion that should be everything.

JeremyRand:
> I am afraid I don't have a good answer for that. I just learned this approach for another project and thought I could contribute with this recent acquire knowledge to improve electrum. I am open to explore the use of setup.py for it.

@rllola The following links look usefully relevant:

https://shallowsky.com/blog/programming/python-debian-packages-w-stdeb.html
https://pypi.org/project/stdeb/

For rpm packages, it's apparently easier: `python3 setup.py bdist --format=rpm`, see https://docs.python.org/3/distutils/builtdist.html

JeremyRand:
@ldz1 I agree that it's usually better to use packages supplied by the distro maintainer unless there's a specific reason not to.  That said, there are some valid reasons not to.  One case that seems relevant here is that Debian doesn't actually package Electrum, which (based on perusing the Debian bugtracker) seems to be because of concerns about `aiorpcx` having an unstable API.  The concerns are reasonable from Debian's perspective given the standards they apply to what they package, but there exist Debian users who don't care about those concerns and would rather have a `.deb` package from the Electrum devs, even if it doesn't meet Debian's packaging standards.

Whether you personally fall into the set of users who would benefit from this is, of course, your business, but it's definitely the case that there exist such users.

JeremyRand:
Correction: I actually don't see any aiorpcx-related bugs in Debian's bugtracker, so maybe Debian would accept an aiorpcx package if someone volunteered to add one to Debian.  Seems a worthwhile line of inquiry, but I don't think there's anything wrong with Electrum distributing deb packages in the meantime -- I suspect that the set of people who would do a setup.py-based script for Electrum's repo and the set of people who would create an official Debian package for aiorpcx are mostly disjoint.

JeremyRand:
And, just to be clear -- in the specific case of non-disjoint sets, i.e. people who are familiar with creating packages for official Debian inclusion, I think fixing the official Debian packaging situation would be a better usage of time.  (Doesn't mean this issue should go away, since the sets are definitely not fully overlapping.)

rllola:
I have no experience in creating official Debian package but I am happy to give a go. Do you have any recommendation on where I should start?

JeremyRand:
> I have no experience in creating official Debian package but I am happy to give a go. Do you have any recommendation on where I should start?

@rllola Searching the Debian wiki for `python packaging` should give you some good starting points.  In addition, I asked @josephbisch a while back about who in Debian would be a good contact for Bitcoin/Namecoin software, and he said:

> the bitcoin packaging group is very small, but i think they would be the group most likely to be interested in namecoin software.  namecoin would probably have to do the work and then someone from one of those groups who is a DD would upload the package

Looks like there's a "Python Apps Packaging Team" as well; they might also be worth contacting in addition to the Bitcoin packaging team..

AFAIK the main blocker isn't Electrum itself, but rather some of its dependency libraries which aren't packaged for Debian.  So probably packaging those dependencies first would be the right approach, and then look into packaging an updated version of Electrum.

Hope this helps.

relativisticelectron:
@rllola : I like the idea of a deb build script. (sorry, I have not tested your build script yet.) Deb packages have the advantages that:

1. you can uninstall it easily
2. once installed the files cannot be changed without sudo rights. 
3. SInce a lot of people use ubuntu/debian it would make it easier for them

I have a question though: Which binary would it wrap? The appimage?

rllola:
Good question... It has been a long time since I made this PR.

According to the script it will take whatever is in the .zip in the `dist` folder.
https://github.com/rllola/electrum/blob/debian-package/contrib/make_deb#L11

And I believe inside there is the python files...

accumulator:
Closing this as outdated. Electrum is packaged on recent Debian, one could extract the packaging scripts from there.

PR #5356, opened 2019-05-18
Spreading malware

shyrwall:
Confirmed as bad


SomberNight:
Double-checked.
Thanks.

PR #5413, opened 2019-06-08
add easy support for op return bytes

jcktm:
I need this for a token sale, feel free to merge. Not sure if ppl even want this feature but it would be super handy for me if it was  included in the next release ^_^

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -453,7 +453,7 @@ class Commands:
         message = util.to_bytes(message)
         return ecc.verify_message_with_address(address, sig, message)
 
-    def _mktx(self, outputs, fee, change_addr, domain, nocheck, unsigned, rbf, password, locktime=None):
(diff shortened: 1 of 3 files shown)


jcktm:
Screen Shot 2019-06-09 at 12 40 08 am
Screen Shot 2019-06-09 at 12 40 58 am
Screen Shot 2019-06-09 at 12 40 35 am

jcktm:
I guess it's not a good UX for non-technical folks though, so maybe the UI needs to change. Add a checkbox or something...

SomberNight:
Thanks; but this functionality already exists :)
Clearly it's not advertised well enough.

You can already write stuff like this into the `Pay to` field:
```
tb1qmgfge5npg3cptmc06czl09altj4e4vx37clmdc, 0.001
OP_RETURN deadbeef, 0
OP_CHECKSIG 02b057b04a768053e450967fe75e2c3daacc38a2e9c3cd53f239beb3fc760b0614, 0.001
```
pr_5413

jcktm:
Oh wow! Yeah I couldn't find it. This is great news, thanks!

PR #5438, opened 2019-06-20
Update block explorer URL for btc.com

nachunjae:


streetcrypto7:
@SomberNight  Can you merge this?

SomberNight:
I recently tried to test this but btc.com did not load properly from two different IPs.
Does it load for you?

Further, what about the testnet btc.com a few lines below. Does that need to be changed?

nachunjae:
@SomberNight I've removed testnet btc.com cause they've closed for testnets.

Mainnet BTC.com explorer works fine, Binance exchange use their exchange.

SomberNight:
Ok. Well it still doesn't load for me but maybe they just rate limited all the proxies I have tried.

PR #5440, opened 2019-06-20
Add multisig support for Coldcard plugin

peter-conalgo:
In Wallet > Information dialog:
- add a "Export for Coldcard" button ... required a new hook, sorry
- improved information display, so the keystore labels for each cosigner is visible

Sign P2SH inputs using Coldcards
- can do online signing between multiple Coldcards, or export PSBT at any point in signing process
- ~~bugfix in `transaction.py`: a transaction with 2-of-4 signatures needed, but has more than that, perhaps 3 or 4 signatures is still "complete"~~

Under Tools > Load Transaction, add "From PSBT File or Files"
- reads PSBT files, and displays the transaction details (general purpose)
- can combine a number of PSBT (all must be about same transaction) and combine their signatures (only implements P2SH signatures for now)
- can export the combined PSBT, or transmit if signed enough, or do further online signing
- add PSBT parser and re-serializer (new file: `basic_psbt.py`)
- new file`build_psbt.py` isolates code to build PSBT from transactions, and new combining code

Other notes:
- corrects endian of XFP (fingerprint) as shown to users
- tooltip changed for "show address" so when there are 15 devices attached, it's more clear which one you are trying to activate
- multi-vendor multisig should work, but haven't tried
- requires version 2.1RC0 or higher on Coldcard
- start Coldcard simulator with "-m" flag to get a 2-of-4 wallet on simulator ready to go, using BIP39 passphrases: "Me", "Myself", "And I", and empty string.

Happy to make changes to make this large diff easier to merge.


peter-conalgo:
My change to transaction.py, 75b6b663e ... was a bad idea and wrong. Reverted that.

justinmoon:
I just tested this and everything seemed to work. Was able to:
- create a 2/2 multisig with trezor and coldcard
- fund it
- spend from it
- "Export for Coldcard" button also worked

nvk:
Merge conflict fixed.

SomberNight:
How difficult do you think it would be to make the PSBT support independent of coldcard, and move it into the main electrum code?

peter-conalgo:
Well I didn't do that because I was trying hard to be a good plugin author and keep my diffs isolated... But it shouldn't be hard to move this out of plugins and into shared code areas.

Certainly I'd like to see Electrum support PSBT across the board. I'm willing to put in the work, but I'm worried it might not get merged for some time and our users need support. 

What is the timeline for the next Electrum release? I'm happy to work with that.

Alternatively, merge this PR now and I'll move the PSBT stuff out from that point. Might get some more eyes testing the newer code that way.

SomberNight:
Ok, I will try to review this this week (at quick read through it looks good); and then merge it.
After that, it would be great if the PSBT stuff could be mostly moved out of the plugin and integrated with the rest of the codebase.

There isn't really a timeline for a next release. The timeline atm is basically "if there is something important on master not yet released, make a release", otherwise master is on back-burner basically and we are  focusing on lightning. So perhaps this could get merged, then we can do a minor release; and PSBT integration could get merged after that.

peter-conalgo:
Great, let me know if you need any changes as you review it, I will try to be speedy.

tepidStorage:
I am trying to setup an airgapped 2-of-3 multisig wallet using 3 ColdCards and Electrum.

On each Cold Card, I created a new seed and exported the XPUB. On the third ColdCard I created the wallet and exported the el-CC-2-of-3.json file for Electrum (per the instructions at https://coldcardwallet.com/docs/multisig).

I then cloned the https://github.com/Coldcard/electrum repo and built the "multisig" branch. I had to install the libusb headers and the ckcc_protocol module, but I eventually persuaded Electrum to read my exported JSON wallet. Electrum is showing me my receive addresses and I sent a small amount of BTC to one of them.

Now I'm trying to create a spend transaction. I used the "Save PSBT" button on the Electrum Preview screen to save the PBST to my SD card.

On the ColdCard, I went to "Ready to sign". It briefly flashes an error on screen (so fast that I had to video the process with my smartphone and go frame-by-frame to see the error!) The error is "Failure: Invalid PBST: Need witness script for input..." (I cannot see the rest of the text because I cannot scroll). It looks like this error comes from line 738 of the firmware (see: https://github.com/Coldcard/firmware/blob/master/shared/psbt.py)

What am I doing wrong?

(Also posted on Reddit: https://old.reddit.com/r/coldcard/comments/ckcfpj/airgapped_multisig_with_electrum_failure_invalid/)

nvk:
Make sure Coldcard is not set to testnet. You might want to email support not sure if this is the place.

tepidStorage:
In "Settings" > "Blockchain" on the ColdCard, I do have "Bitcoin" selected (not "Testnet: BTC"). In txt file that ColdCard exported for the multisig wallet is m/48'/0'/0'/2'.

Posted here because it seems like it should be a deal breaker for this PR if Electrum isn't constructing PSBHs correctly for airgapped multisig wallets.

tepidStorage:
Another observation: When trying to load the PSBH ("Tools > Load Transaction") back into the Electrum instance that just created it, I get: "No partial signatures found for input #0".

nvk:
Ok. Will look at it in the morning.

peter-conalgo:
Good morning @tepidStorage ...

Nothing wrong with your process there, and I bet Electrum built the PSBT right. It's probably a bug on Coldcard side. Regardless, we want to fix it, so need some answers:
- Please: send the PSBT if you are willing to share it (here as base64, or support@coinkite.com)
- what address type was the wallet?The error implies it was `P2WSH-P2SH`, but the derivation path above implies `P2WSH`
- do you know what UTXO was being spent?

RE: disappearing error message. This is a know bug that we are working on, sorry.

RE: "No partial signatures found" .. that's a warning because you've imported an PSBT with no new signatures. This is to be expected when you save and then re-import the unchanged PSBT. No progress was made towards signing, so we see that as a concern. IIRC, it still lets you view the transaction.

tepidStorage:
Hi @peter-conalgo and thank you for looking at my issue.

I cannot share this specific PSBT (it would harm my privacy).  I do want to help you get to the root cause though as I'm upgrading my cold storage and hoping to make use of ColdCards!  Maybe I can help by answering your questions and clarifying repro steps.  Failing that I'll see if I have some dust BTC or some testnet BTC somewhere that I can use to create a repro that I can share with you.

Firstly, clarification of the repro steps (assume I'm accepting the defaults for any parameters not explicitly stated):
1. Get 3 new ColdCards with firmware 2019-07-03T1316-v2.1.1-coldcard.dfu and 1 SD card
2. Insert SD card into CC1. Initialize CC1 with a new seed, go to "Settings" then "Multisig Wallets" then "Export XPUB"
3. Insert SD card into CC2. Initialize CC2 with a new seed, go to "Settings" then "Multisig Wallets" then "Export XPUB"
4. Insert SD card into CC3. Initialize CC3 with a new seed, go to "Settings" then "Multisig Wallets" then "Export XPUB"
5. Still on CC3: Go to "Settings" then "Multisig Wallets" then "Create Airgapped" (this creates el-CC-2-of-3.json on the SD card)
6. Go to a Linux computer and build Electrum from the "multisig" branch of https://github.com/Coldcard/electrum
7. Run Electrum and open el-CC-2-of-3.json
8. Go to the receive tab and grab the bc1q.... address (Address_1)
9. Use some other wallet software to send BTC to Address_1
10. Wait for it to confirm
11. Go to the receive tab and grab a new bc1q.... address for this wallet (Address_2)
12. Go to the "Send" tab and try to send the BTC now in the wallet to Address_2, click the "Preview" button and then "Save PSBT", save the file to the SD card
13. Put the SD card back in CC3 to sign it; observe that it is rejected with the error "Failure: Invalid PBST: Need witness script for input..."

In my case, the ColdCard reports that the multisig wallet uses P2WSH addresses and derivation path m/48'/0'/0'/2' (as mentioned, I used default settings when creating the wallet).

The UTXO being spent is a pretty boring transaction: It came from an instance of Wasabi Wallet (no multisig, no hardware wallets, bc1... style addresses).

Let me know if you have any other questions.  Or need me to create a repro with a PSBT I'd be willing to share.

peter-conalgo:
I can reproduce this now. Pretty sure it's an issue on the Coldcard side, but still researching.

peter-conalgo:
We've fixed some issues on both sides: 
- Please use only the latest version of the Coldcard firmware (published today): [v2.1.2](https://coldcardwallet.com/docs/upgrade). That fixes a bunch of issues and greatly improves the display of PSBT errors to aid in debug.
- the _'No partial signatures found for input'_ message is improved now (text change)
- full list of [firmware changes is here](https://github.com/Coldcard/firmware/blob/master/releases/ChangeLog.md)
- In addition, I found a case where the `witness_script` was not being added to the PSBT file by the Electrum plugin, and some issues when merging/combining the resulting PSBT. For that fix, please pull for the latest on this branch.


In your case @tepidStorage, you'll need to update all the Coldcard in your group. No need to remake the shared wallet.

tepidStorage:
Awesome, I'll try this out now...

tepidStorage:
Works great!

peter-conalgo:
Great to hear that worked for you and solved that issue.

@SomberNight I think this is ready to merge.

mflaxman:
This worked great for me, but the output displaying the paths on a testnet multisig address seems broken?:
IMG_4018
IMG_4016
IMG_4017

justinmoon:
I can replicate ^^

peter-conalgo:
That is by-design. The question mark is shown for sections of the derivation that it's not possible for the Coldcard to verify itself. We have to take on faith the hardened derivation of our co-signers.

mflaxman:
Ah, that makes sense. I'd suggest `_` or `-` for this, as the `?` looks like a bug! Also, `Expected Paths`, `Unconfirmed Paths`, or `Cosigner Paths`, might also be better copy in the previous screen if it fits.

peter-conalgo:
Yes, we'll change that to be `-` (dash).

mflaxman:
Any plans to merge this in? I've tested and it's awesome.

I'm working on a multi-hardware wallet multisig tutorial using Electrum that requires this branch.

Thanks!

ecdsa:
please add PSBT support as a separate pull request. (not in the plugin)

SomberNight:
I have run into multiple issues while testing this, and while some I can fix, some are related to either the PSBT logic or the firmware itself.

I am testing with coldcards using fw 2.1.2

I have to say that the testing matrix is huge, getting near the limit of what a human can manually test
(`script_types` x `whether_there_is_change` x `whether_coldcard_has_wallet_description_txt_file_imported` x `coldcard_using_usb_vs_sdcard`, for multisig one needs to test at least 2of2 and 2of3; also when one cosigner is a different manufacturer hw device; and not sure what else)
Hopefully not every element of the Cartesian product needs testing though but there are unexpected correlations.

below tests using USB cable, and wallet description txt imported on device prior:

- even without this PR, with p2wpkh-p2sh scripts, I am getting a change fraud failure if there is change:
https://github.com/spesmilo/electrum/issues/5612
- with the PR, I am also getting a change fraud failure for p2wsh-p2sh scripts
- with the PR, with p2wsh scripts both 2of2 and 2of3 wallet, if there is a change output, I am getting `Failure\nInvalid PSBT\n\n(psbt.py:390)`

Also, regarding importing the wallet description txt, is it expected that the following two files overwrite each other?? Kind of annoying. :P
<details>
 <summary>test_ms_p2wsh_2of2_cc3132_cc3133-cc-export.txt</summary>

```
# Exported from Electrum
Name: test_ms_p2wsh_2of2_c
Policy: 2 of 2
Format: P2WSH
Derivation: m/48'/1'/0'/2'

B2768D2F: Vpub5n3MbBwqiSaeqw9QrVzaRtRdJ6G9V9T44BoahqKKQVF23FTzY11q7nmbFKkzjkj1c2E7CZMXvncLsU5954SqkbTppiJ3mGRGzQhrHZE4ubZ

E6F393AF: Vpub5koKpLamdEGUXjxbJzNrsaqTxwXvYAxbnj1WT4x1HaUJF3KSzomRsixH428yyoUyjZn6LXecMWxxDtVw12iG54A1on3gQrvamfzyBowRHBt
```
</details>

<details>
 <summary>test_ms_p2wsh_p2sh_2of2_cc3132_cc3133-cc-export.txt</summary>

```
# Exported from Electrum
Name: test_ms_p2wsh_p2sh_2
Policy: 2 of 2
Format: P2WSH-P2SH
Derivation: m/48'/1'/0'/1'

B2768D2F: Upub5TD6HXGvZm3Aw7at4jndBnHaLgssRpKAJx4o3TEgDRyLnHvMQcstDkg6Phm7E2UyqpmtjrtR1vK7k3RaL5Q68axY7FWbUAkeuSimtRLuuz4

E6F393AF: Upub5Ry4WfurUYizeniDVrNS5jzS2cc9P54JXA5kVeX2vYign6TZXMuDNXagXLxCQzYSQ4SHYZvnrpRQJmTPvXtVD74XKiQ76n85phESxTuMgR8
```
</details>

nvk:
> Ah, that makes sense. I'd suggest _ or - for this, as the ? looks like a bug! Also, Expected Paths, Unconfirmed Paths, or Cosigner Paths, might also be better copy in the previous screen if it fits.

@mflaxman This was jut addressed on firmware release 2.1.3 https://coldcardwallet.com/docs/upgrade

nvk:
> I am testing with coldcards using fw 2.1.2

Try today's 2.1.3 release.

> Bugfix: P2SH-P2WPKH change outputs incorrectly flagged as fraudulent (regression from v1.1.0)
> Bugfix: Wanted redeem script, but should be witness script for P2WSH change outputs.

SomberNight:
> Try today's 2.1.3 release.

I have redone some of my tests.
Indeed the p2wpkh-p2sh change output "change fraud" is now fixed. However the other listed issues are not.

> - with the PR, I am also getting a change fraud failure for p2wsh-p2sh scripts
> - with the PR, with p2wsh scripts both 2of2 and 2of3 wallet, if there is a change output, I am getting Failure\nInvalid PSBT\n\n(psbt.py:390)

I am testing using a USB cable, with the wallet description txt files imported prior.
I have tested the following combinations this time:
- legacy p2sh 2of2 wallet
	- single external output: OK
	- one external, one change output: OK
- p2wsh-p2sh 2of2 wallet
	- single external output: OK
	- one external, one change output: **change fraud**
- p2wsh 2of2 wallet
	- single external output: OK
	- one external, one change output: `Failure\nInvalid PSBT\n\n(psbt.py:391)`

Please test the failing cases yourself.

peter-conalgo:
I've found an issue in the 2.1.3 firmware if the change output is multisig. Coldcard incorrectly flags them as fraud because we lack the code to properly decode the `scriptPubKey`.

Nonetheless, v2.1.3 does fix `p2wpkh-p2sh` change outputs.

I'm working on a fix for multisig output change, but first revamping our clearly insufficient  test cases.

With respect to this branch... I have another issue w/ change outputs for PSBT; seems the API of `Keystore` object has changed, and lost a function `get_derivation` (or something like that).

SomberNight:
> I'm working on a fix for multisig output change, but first revamping our clearly insufficient test cases.

Great, thank you.

> With respect to this branch... I have another issue w/ change outputs for PSBT; seems the API of Keystore object has changed, and lost a function get_derivation (or something like that).

Could you give some more context? I've diffed the latest release against master, and no function was removed from keystore.py. Are you saying there is an issue with how the PSBT is constructed re change outputs even in this branch (not just decoding in the firmware)? Is this "removed function" something you would now want to use; or is the PR already using it?

peter-conalgo:
I think it's just bogus old code on my part. On line 162 of `build_psbt.py`:

```python
  # XFP + fixed prefix for this keystore
  ks_prefix = packed_xfp_path(xfp_for_keystore(ks), ks.get_derivation()[2:])
```
Where `ks` is a `Keystore` object. In other spots of the same file, I'm doing `dd = getattr(ks, 'derivation', 'm')`  so I think this might be a remnant of much earlier stuff.

SomberNight:
It's here https://github.com/spesmilo/electrum/blob/215dc96de73b507160885fb37d2698c473194898/electrum/plugins/coldcard/coldcard.py#L291-L292

Looks like all subclasses of `Hardware_KeyStore` define it; exactly like this; so maybe it should be moved to that class :) anyway, I am planning some refactoring after this PR is merged...

peter-conalgo:
Right. Some of the co-signers in a multisig may not be a hardware wallet (ie. just a BIP32 keystore), so that's why I used `getattr` in that one spot... You're right it needs a fix at the base class.

peter-conalgo:
Version 2.1.4 of Coldcard firmware is public now; contains fixes for multisig change outputs.

I've pushed a small change to this branch to help a user who has a `BIP32Keystore` as a cosigner in his multisig wallet.

mflaxman:
I only tested ps2sh 2-of-3 (1 coldcard, 1 trezor, 1 tpub) but this immediately failed for me on both displaying a receive address:
```
E | plugins.coldcard.coldcard.Coldcard_KeyStore | Error showing address
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/coldcard.py", line 486, in show_p2sh_address
    dev_addr = client.show_p2sh_address(M, xfp_paths, script, addr_fmt=addr_fmt)
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/coldcard.py", line 208, in show_p2sh_address
    return self.dev.send_recv(CCProtocolPacker.show_p2sh_address(*args, **kws), timeout=None)
  File "/home/satoshi/workspace/electrum_coldcard_pr/.venv3/lib/python3.6/site-packages/ckcc/client.py", line 159, in send_recv
    return CCProtocolUnpacker.decode(resp)
  File "/home/satoshi/workspace/electrum_coldcard_pr/.venv3/lib/python3.6/site-packages/ckcc/protocol.py", line 182, in decode
    return d(msg)
  File "/home/satoshi/workspace/electrum_coldcard_pr/.venv3/lib/python3.6/site-packages/ckcc/protocol.py", line 196, in err_
    raise CCProtoError("Remote Error: " + str(msg[4:], 'utf8', 'ignore'), msg[4:])
ckcc.protocol.CCProtoError: Remote Error: pk#1 wrong, tried: (m=0918C47F)
```
As well as signing a transaction:
```
E | gui.qt.main_window.[2of3_p2sh] | on_error
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/gui/qt/util.py", line 718, in run
    result = task.task()
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/wallet.py", line 1112, in sign_transaction
    k.sign_transaction(tx, password)
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/coldcard.py", line 318, in wrapper
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/coldcard.py", line 400, in sign_transaction
    raw_psbt = build_psbt(tx, self.my_wallet)
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/build_psbt.py", line 162, in build_psbt
    ks_prefix = packed_xfp_path_for_keystore(ks)
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/build_psbt.py", line 91, in packed_xfp_path_for_keystore
    return packed_xfp_path(xfp_for_keystore(ks), derv[2:], int_path=int_path)
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/build_psbt.py", line 56, in packed_xfp_path
    x = int(x)
ValueError: invalid literal for int() with base 10: ''
```

I'm using 5a3b001903d46fdea4c6d3a9c9941ba365646bb1 with firmware 2.1.4.

mflaxman:
Also, hitting the `Save PSBT` button on a TX doesn't work:
```
E | gui.qt.exception_window.Exception_Hook | exception caught by crash reporter
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/qt.py", line 96, in <lambda>
    btn.clicked.connect(lambda unused: self.export_psbt(dia))
  File "/home/satoshi/workspace/electrum_coldcard_pr/electrum/plugins/coldcard/qt.py", line 114, in export_psbt
    assert type(keystore) == self.keystore_class
AssertionError
```

peter-conalgo:
> I still get `Change Fraud\nOutput#1: Change output is fraudulent` for p2wsh-p2sh (2of2).
> The "Show address" function for the exact change address in question works, the device displays the change address along with its derivation path, so clearly the issue is with the validation code (or the PSBT could be constructed incorrectly too I guess).
> 
> PSBT (unsigned 59b833d9b4ea42fd77c545394a3251ccfc2b8f4031e48165fe5ca3b18b204f20)
> ```
> cHNidP8BAHMCAAAAAaUG2AdoI6uDIQHrtlIcQDePgM24HgsAlUWD/i3TmeXJAAAAAAD9////AqCGAQAAAAAAF6kUGsiaZEieHCCR5SHpmBDsnI2rstCHtPwcAAAAAAAXqRQIhXSEDYJqoKecZ8rZZzjBKr4jYIfIFBgATwECQonvBK1TixuAAAABC+LJV8NlRCFNSK1TihAUEHM9TMh3PyVVHvooI4R5QJsC/9bxDFtRkCwA26T6XoNaeRvTxIT9zzzC7Pdf6GlHL1IUsnaNLzAAAIABAACAAAAAgAEAAIBPAQJCie8EBF0XjYAAAAHBmltSKpv7MZjNDUpjW6f4dpPy61XlQGFc0GfYGDHsWwK0L8FME5nfeUxSZn0Mw5+VfK9OKo++4OP7PctvYHBNIRTm85OvMAAAgAEAAIAAAACAAQAAgAABASCAhB4AAAAAABepFGAK7PhGb/ELxMq7EmJRXiCUdyh/hwEFR1IhAh/JrAAq59fCqXqRlaZQ020twlg0eXasE13VG5eEnJ7FIQN85whPAH/Wq0+Xb7WCeks/H55OmjfGsEsgf6qJiVmlSFKuIgYCH8msACrn18KpepGVplDTbS3CWDR5dqwTXdUbl4ScnsUc5vOTrzAAAIABAACAAAAAgAEAAIAAAAAAAAAAACIGA3znCE8Af9arT5dvtYJ6Sz8fnk6aN8awSyB/qomJWaVIHLJ2jS8wAACAAQAAgAAAAIABAACAAAAAAAAAAAAAAAEAR1IhAgFUqG8Ysz/Fp1SbVrvEnSOYpIbuEKEkStxaquDlW3GfIQMm+OLz8xNbUNOrc44zP+GRCdSwuro09Kiy+L3QEKBWyFKuIgICAVSobxizP8WnVJtWu8SdI5ikhu4QoSRK3Fqq4OVbcZ8csnaNLzAAAIABAACAAAAAgAEAAIABAAAAAAAAACICAyb44vPzE1tQ06tzjjM/4ZEJ1LC6ujT0qLL4vdAQoFbIHObzk68wAACAAQAAgAAAAIABAACAAQAAAAAAAAAA
> ```

The Coldcard doesn't like this change output because it has an `OUT_REDEEM_SCRIPT`  and not an `OUT_WITNESS_SCRIPT`. When considering the outputs of a transaction, I don't know of any other way to know if an output will be claimed by a witness script or legacy script. You can see the logic here: [coldcard/firmware/../psbt.py line 399](https://github.com/Coldcard/firmware/blob/master/shared/psbt.py#L399-L404)

When you use "show address" feature, the address type is part of the USB protocol message, so no ambiguity there.

We could test both hash values, but that seems less secure. Instead, I will change the plugin to generate the PSBT with `OUT_WITNESS_SCRIPT` for these outputs, when we know it's a `p2wsh-p2sh` output.

SomberNight:
>> p2wsh-p2sh (2of2)

> The Coldcard doesn't like this change output because it has an OUT_REDEEM_SCRIPT and not an OUT_WITNESS_SCRIPT. When considering the outputs of a transaction, I don't know of any other way to know if an output will be claimed by a witness script or legacy script

@peter-conalgo Looking at BIP174, I am not fully certain what it expects but my interpretation would be that there should be BOTH a witnessScript and a redeemScript for such a change output, as that is what will be in the final signed transaction that will later spend the output (see https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#p2wsh-nested-in-bip16-p2sh)

peter-conalgo:
To be honest, I am very confused about where witness/redeem or maybe both are required inside the PSBT file format. There are a few spots now in the Coldcard code where we just pick whichever one is defined by the PSBT.

SomberNight:
```
[2019-09-16 17:42:48 +0200] <ghost43>	achow101: regarding bip174 per-output fields, PSBT_OUT_REDEEM_SCRIPT and PSBT_OUT_WITNESS_SCRIPT follow what the consensus mandates for the spending signed bitcoin transaction to have, right? so e.g. for a p2wsh-p2sh multisig change output, the psbt should ideally specify both, and PSBT_OUT_REDEEM_SCRIPT is the witness_version + witness_program; while PSBT_OUT_WITNESS_SCRIPT is the last push of the future witness (i.e. the actual multisig script), 
[2019-09-16 17:42:48 +0200] <ghost43>	right?
[2019-09-16 17:43:09 +0200] <ghost43>	see https://github.com/spesmilo/electrum/pull/5440#issuecomment-531817595
[2019-09-16 17:47:05 +0200] <achow101>	ghost43: yes. those fields should contain the same thing that the redeem_script and witness_script fields would have in the input spending those outputs
[2019-09-16 17:47:30 +0200] <ghost43>	thanks
[2019-09-16 17:48:19 +0200] <achow101>	none of the per-output fields are actually required, they're just useful for identifying change
```
so we need both in the p2wsh-p2sh; also your latest commit I believe actually broke the behaviour for p2wpkh-p2sh

my understanding then is:
p2pkh: no redeemScript, no witnessScript
p2wpkh-p2sh: only redeemScript (which is witness version + witness program `0x0014{20-byte-key-hash}`)
p2wpkh: no redeemScript, no witnessScript
p2sh multisig: redeemScript that contains the actual multisig script
p2wsh-p2sh multisig: redeemScript (which is witness version + witness program `0x0020{32-byte-hash}`) and witnessScript (which contains the actual multisig script)
p2wsh multisig: only witnessScript (which contains the actual multisig script)

note that in ambiguous cases you can also look at the scriptPubKey in the actual outputs in the network serialized transaction (`PSBT_GLOBAL_UNSIGNED_TX`); to e.g. distinguish p2pkh and p2wpkh

peter-conalgo:
Ok, I will adjust things to that standard. Thanks for bringing in @achow101

peter-conalgo:
b20dd825a should be putting the right scripts in now, based on the summary above. I'm not sure how it will behave on 2.1.4 firmware because I've also changed the Coldcard's verifications to align with this. Updated CC firmware (2.1.5) is incoming shortly.

Re: @mflaxman issues: one is fixed (untested) regarding `ValueError: invalid literal for int` but the others will require a test setup with similar multi-vendor complexities.

Generally, I need a way to test the many combinations that are possible between Coldcards, single keys and other vendor's wallets. This stuff is too complex for manual testing. Not sure yet if I should do that on this branch, or maybe in a new repo for the purpose.

peter-conalgo:
Version 2.1.5 firmware now [available here](https://github.com/Coldcard/firmware/raw/master/releases/2019-09-17T1310-v2.1.5-coldcard.dfu).

SomberNight:
Ok, p2wsh-p2sh multisig worked too now.

**Merged in** https://github.com/spesmilo/electrum/commit/c81f5395af88f929e423b0a54c960888128d7ab7

Squashed lots of stuff and added some follow-up commits; diff:
https://github.com/spesmilo/electrum/compare/bd83ca02863920479fefd41f1e54b0315fc6a318...c81f5395af88f929e423b0a54c960888128d7ab7

I expect there are still some bugs but this is now good enough.
Did some cross-vendor multisig tests too; that worked too; saw some minor issues only.

I will now try to integrate PSBT support into master, as part of that and after that, hopefully this will work even better, and we can then remove some of the hacks.

SomberNight:
@peter-conalgo ISTM the coldcard only supports either all-multisig inputs or no-multisig inputs (cannot be mixed); which is an ~OK limitation. However, it seems if any global xpub fields are present then the fw assumes the tx to spend multisig inputs:

https://github.com/Coldcard/firmware/blob/b28cdb3966d49b62a39afd90213e2058e8b695fe/shared/psbt.py#L1075-L1077

Could this limitation be removed?

I would like Electrum to populate the global xpub fields regardless of script types.

Specifically, we would populate global xpubs just before signing, and remove them just after signing. This way, the signer can use them if they want; but if e.g. the user just exports a tx it would not contain an xpub, greatly improving privacy (compared to current electrum partial txns).

(Re using the coldcard with an sdcard, there would be a separate "export for coldcard" export option for that, that would include xpubs)

peter-conalgo:
Yes, this limitation could be removed, and it was in our to-do list. However, there are hazards in adding XPUB data into PSBT files that don't need it. (On the Coldcard, no  PSBT file needs the XPUB, since it does not require that data even for multisig.)

In particular, a MiTM can add additional inputs to bump up the miner fee, and could also add change outputs. If you do both at the same time, you could divert funds into an unexpected change output and perhaps ransom knowledge of the change address. The most recent Coldcard firmware (3.0.2) prevents this with a bunch of heuristics.

If they don't have the XPUB, then an MiTM cannot generate the public key values needed to do any of that. It's also a problem for privacy as unrelated inputs are revealed if the PSBT is found at any stage of its lifecycle.

SomberNight:
> However, there are hazards in adding XPUB data into PSBT files that don't need it. [...]

Thanks; interesting attack.

What I outlined above is that we would populate global xpubs just before signing and remove them just after signing. (so not in the export-flows) In the case of hw wallets, this would be the scenario when the python process communicates with them via hidapi or libusb. Well I guess that could still be MiTM'ed -- depends on the threat model. (you as hw fw developer should certainly not trust the host-libusb, but we might...)

> On the Coldcard, no PSBT file needs the XPUB, since it does not require that data even for multisig

Well, except if the user had not exported a multisig wallet setup file, then they are needed.
In general, stateless signers need the global xpubs, at the very least in case of multisig scripts.
I expect - for example - Trezor will want them for multisig.

-----

re the export flows, this is what I have in mind atm:
tx_dialog_export
So, 3 types of exports (x3 mediums):
- main/common type is "for-signing": include everything we know, except global xpubs and derivation prefixes (so this is reasonably privacy-friendly)
- "for coinjoin": strip all globals, all output info, and bip32 paths re inputs
- "for coldcard": include everything we know, including global xpubs and full derivation paths

SomberNight:
> In particular, a MiTM can add additional inputs to bump up the miner fee, and could also add change outputs. If you do both at the same time, you could divert funds into an unexpected change output and perhaps ransom knowledge of the change address. The most recent Coldcard firmware (3.0.2) prevents this with a bunch of heuristics.

I thought a bit more about this; and I think if your threat model is a MiTM then you should just display a txid on the device screen. This is exactly the problem segwit solved: a txid describes all important parts of a transaction, and it is available before signing. The user can manually compare txid between client and hardware screen.

chriswebstin:
At first, I suspected the call was illegitimate, but I was eventually worn down by the tactics used by the fraudsters. They knew some of my personal information, and one of the caller's numbers showed up as "York Regional Police" on my cell phone. Eventually, I was convinced to withdraw $17000 from my bank and take it to a convenience store in Richmond Hill, where i deposited the money into a Bitcoin ATM. The machine converted the cash to Bitcoin and transferred the funds to the caller. Have never heard of Bitcoin before receiving the call. They made me believe that it was a machine used by Revenue Canada to transfer money quickly, after few hours I began to realize what had happened contact: gmail @  ULTIMATEFUNDRECOVERY
WHATSAPP:+14079061584 
TEXT:  +14079061584

nopara73:
Is there any chance to get a minor release that includes this fix?

peter-conalgo:
Version 3.1.1 of Coldcard firmware allows XPUB data in the PSBT, even if it's a single-signer PSBT. Just out today.

Issue #5437, opened 2019-06-18
KeepKey keine Verbindung mit Electrum 3.3.6 unter Windows 10 Build 1903


TheNightProducer:
Keine Verbindung mehr möglich unter neuer Windows Build vorheriges Build ging es noch. Im Geräte Manager wird der KeepKey erkannt und mit Originaler App funktioniert es.

SomberNight:
This is likely related to https://github.com/spesmilo/electrum/issues/5420

TheNightProducer:
https://puu.sh/DHDkL/90337ff378.png

EchterAgo:
I've diagnosed this issue a bit today and it seems that `hidapi` is not enumerating KeepKey anymore starting with Windows 10 Build 1903. `libusb1` and other Windows tools enumerate it just fine, and it shows up under HID devices in device manager. Trezor is likely similarly affected but I can't test that. The difference between Trezor and KeepKey is that Trezor registers an enumeration function that uses `hidapi` and `libusb1` to enumerate devices while KeepKey does not and only relies on the `hidapi` enumeration done by the device manager.

The fix for this would be to either to register a custom enumerate function with device manager that uses the KeepKey enumeration functions or use `libusb1` to enumerate in device manager. I'm not sure what the better solution is.

EchterAgo:
So from reading the other issue it seems like just using libusb will not fix this. I will investigate this more tomorrow.

EchterAgo:
Running as admin does work. I was also able to make it work by using a custom enumeration function in the KeepKey plugin. This probably only fixes the WebUSB KeepKeys (USB PID 0x0002) but leaves the HID ones (USB PID 0x0001) broken.

cculianu:
Trezor broke too with windows 10 1903.  But running as admin it works again.

They probably changed their permissions or something in their USB layer after adding universal 2fa support in 1903? (This is just a hypothesis/guess).

But running as admin does make it work again.

Oh microsoft...

EchterAgo:
So Windows 10 version 1903 blocks attempts to open USB HID interfaces that have a U2F usage. Only admin users are allowed to open those.

https://github.com/node-hid/node-hid/issues/312

Ledger seems to address this by dropping U2F (and maybe HID too?):

https://www.ledger.com/windows-10-update-sunsetting-u2f-tunnel-transport-for-ledger-devices/

[...]

PR #5460, opened 2019-06-26
keepkey: use libusb to enumerate devices instead of hid

SomberNight:
Currently the KeepKeys are enumerated through hidapi, and then used through either hidapi or libusb.
With this change (old fw) hidapi KeepKeys are enumerated and used through hidapi,
while (new fw) webusb KeepKeys are enumerated and used through libusb.

This is needed to handle https://github.com/spesmilo/electrum/issues/5437
See https://github.com/Electron-Cash/Electron-Cash/pull/1471 for more context

This PR currently fails with
```
I | plugin.DeviceMgr | scanning devices...
E | plugin.DeviceMgr | custom device enum failed. func <bound method KeepKeyPlugin.enumerate of <electrum.plugins.keepkey.qt.Plugin object at 0x0000024360098EB8>>, error LIBUSB_ERROR_ACCESS [-3]
  File "...\electrum\electrum\plugin.py", line 606, in scan_devices
    new_devices = f()
  File "...\electrum\electrum\plugins\keepkey\keepkey.py", line 110, in enumerate
    for dev in WebUsbTransport.enumerate():
  File "...\python-keepkey\keepkeylib\transport_webusb.py", line 94, in enumerate
    dev.getProduct()
  File "C:\Users\User\AppData\Local\Programs\Python\Python37\lib\site-packages\usb1\__init__.py", line 2033, in getProduct
    return self._getASCIIStringDescriptor(self.device_descriptor.iProduct)
  File "C:\Users\User\AppData\Local\Programs\Python\Python37\lib\site-packages\usb1\__init__.py", line 2018, in _getASCIIStringDescriptor
    return self.open().getASCIIStringDescriptor(descriptor)
  File "C:\Users\User\AppData\Local\Programs\Python\Python37\lib\site-packages\usb1\__init__.py", line 2068, in open
    mayRaiseUSBError(libusb1.libusb_open(self.device_p, byref(handle)))
  File "C:\Users\User\AppData\Local\Programs\Python\Python37\lib\site-packages\usb1\__init__.py", line 133, in mayRaiseUSBError
    __raiseUSBError(value)
  File "C:\Users\User\AppData\Local\Programs\Python\Python37\lib\site-packages\usb1\__init__.py", line 125, in raiseUSBError
    raise __STATUS_TO_EXCEPTION_DICT.get(value, __USBError)(value)
usb1.USBErrorAccess: LIBUSB_ERROR_ACCESS [-3]
```
So the line triggering the failure is https://github.com/keepkey/python-keepkey/blob/636a053bb80edc1260ced951425a2a029d804ba2/keepkeylib/transport_webusb.py#L94

The Trezor plugin behaves essentially the same and has the same test in python-trezor:
https://github.com/trezor/trezor-firmware/blob/f3e3ebe1d8539dbdc04262c687a4fd3abbfe9497/python/trezorlib/transport/webusb.py#L127

I don't know why KeepKey behaves differently and raises this libusb exception. I would like present PR to work.

@keepkeyjon Please have a look at this and the linked issues.

I think the failing line might be superfluous and that test could be removed. Just commenting out that single line makes everything work AFAICT. Please consider either removing that test, making it optional, or making it easy to monkey-patch out. Alternatively please make the test not fail just like with Trezor (again, no idea what the difference is... already spent too much time on this and if it is at firmware level then as-is that's above my level of understanding).


keepkeyjon:
@SomberNight which firmware version are you seeing this with?

SomberNight:
fw 6.1.0
On Windows 10 (mostly tested on 1803 so 10.0.17134; but a bit on 1903 too)
libusb 1.0.22

EchterAgo:
The `LIBUSB_ERROR_ACCESS` is due to `libusb` failing in the Windows backend `composite_open`. When opening the HID interface of a composite device fails, it won't let you use just the WinUSB backend that succeeded. I have an experimental branch of libusb here that allows you to ignore access denied errors when opening a device.

https://github.com/EchterAgo/libusb/tree/win1903_hid_access
https://github.com/EchterAgo/libusb/tree/win1903_hid_access_default

The difference between those two branches is that the _default branch sets the flag by default. I also have a build system integration for Electron Cash here:

https://github.com/EchterAgo/Electron-Cash/tree/libusb_build

Note that this is WIP.

EchterAgo:
We managed to fix it in https://github.com/Electron-Cash/Electron-Cash/pull/1556

This also fixed the issues we had with Trezor WebUSB on Windows 10 1903.

keepkeyjon:
I'm curious if it behaves differently on firmware `v6.2.0`.

@EchterAgo in your debugging, could you tell why opening the composite device fails? Is it Windows being picky about U2F + Winusb?

EchterAgo:
@keepkeyjon Firmware 6.2.0 doesn't cause any change in behaviour. I'm testing with Electron Cash though and can't speak for Electrum. Here's what our releases do on Windows 10 1903:

* 4.0.6: No detection due to the code using only the hid module.
* 4.0.7: We added switched to keepkeylib WebUsbTransport enumeration. This made it work but requred reconnecting the device on every operation because your enumeration code opens the device. On Windows it is not allowed to open a WebUSB device twice, so the `getProduct` call fails, causing the device manager to think the device is disconnected. Not sure why this works with Trezor. Note that the `getProduct` call was only needed for libusb < 1.0.22 as those versions returned the WinUSB and the HID device as separate devices. libusb version 1.0.22 fixed this problem. At this time we didn't upgrade from 1.0.21 to 1.0.22 as that made it fail completely. See https://github.com/Electron-Cash/Electron-Cash/commit/78d14f2ee980f04b18571c290b091836820601c1
* 4.0.9-dev5: We determined that the reason why 1.0.22 fails is because when opening a composite device, it always fails when it can't also open the hid device, see https://github.com/libusb/libusb/blob/0034b2afdcdb1614e78edaa2a9e22d5936aeae5d/libusb/os/windows_winusb.c#L3804 . We patched libusb to make ignore access denied errors when opening the hid part of a composite device, see https://github.com/EchterAgo/libusb/commit/6200347d16534980baa604725d1de385e3ed436f . We also had to change to our own enumeration function that doesn't call getProduct, see https://github.com/Electron-Cash/Electron-Cash/commit/aa442dbb7a289bf4c4848fb5eaf48be15f812710

The reason why the open fails is indeed Windows being picky. On Windows 10 >= 1903 HID devices with U2F usage can only be opened by Administrators. The WebUSB part is a different interface though and can be opened without problem. The problem is that libusb *only* succeeds opening the device when it can open *both* interfaces.

EchterAgo:
Related PRs for `keepkeylib`:

* https://github.com/keepkey/python-keepkey/pull/67
* https://github.com/keepkey/python-keepkey/pull/68

EchterAgo:
Note that you also can't use `getSerialNumber` during enumeration as that also opens the device. You can just use the path instead.

keepkeyjon:
Hmm. That suggests there's merit to an idea I've been tossing around: make the U2F feature and associated usage opt-in, and remove them from the bootloader. If we defaulted it to off, users who found that windows won't open their device could wipe & re-initialize, and then not use the U2F feature.

keepkeyjon:
Unfortunately I don't have a Windows machine that reproduces these issues, so anything I try here would be a blind fix.

EchterAgo:
I really think libusb shouldn't completely fail when only part of the device can not be opened. Or it needs to allow explicitly selecting which backend is used to open a device.

As for reproducing, I can repro it in a VM running 1903 too.

EchterAgo:
Related PR for `libusb`: https://github.com/libusb/libusb/pull/603

EchterAgo:
The libusb fixes are now merged. @keepkeyjon, are you going to release a new version too?

EchterAgo:
We will use libusb 1.0.23-rc2 in Electron Cash:

https://github.com/Electron-Cash/Electron-Cash/pull/1572

I also posted a binary there, libusb doesn't seem to publish binaries for release candidates. The binary can be built reproducible with this:

https://github.com/Electron-Cash/Electron-Cash/blob/master/contrib/build-wine/_build.sh#L242

keepkeyjon:
>  are you going to release a new version too?

Sure. Next week most likely.

EchterAgo:
@keepkeyjon a release would be really useful, even if it just includes those fixes. Everything is working and libusb released version 1.0.23 which includes the fixes.

3rdIteration:
Just ran a quick test and this works well with Windows 10 1906 and the Keepkey 6.2.2. firmware too. No admin required :)

EchterAgo:
I can confirm, it is working on W10 1906 with libusb 1.0.22 again.

SomberNight:
Do you mean 1909? build number 18363? If not, what is the build number for that?

3rdIteration:
Typo on my part.

Windows 1909, build 18363.476

3rdIteration:
So I bumped to the latest master 4.0.0a0 and it all works nicely with these changes merged too...

Also, re: client.py so that the Trezor-style recovery works again too, @keepkeyjon is there a reason why this shouldn't be re-enabled? (As it was disabled from python-keepkey too) (eg trivial change @ https://github.com/3rdIteration/electrum/commit/524b30bab78c8c30987202cf27a8387732dfec8a)

keepkeyjon:
> Also, re: client.py so that the Trezor-style recovery works again too, @keepkeyjon is there a reason why this shouldn't be re-enabled? (As it was disabled from python-keepkey too) (eg trivial change @ [3rdIteration@524b30b](https://github.com/3rdIteration/electrum/commit/524b30bab78c8c30987202cf27a8387732dfec8a))

Firmware v6.3.0 and later won't have it.

SomberNight:
Note: this PR is waiting on a pypi release of python-keepkey that contains @EchterAgo's fixes.
Please ping me when there is one.

keepkeyjon:
https://pypi.org/project/keepkey/6.3.0/

Sorry that took so long.

SomberNight:
Thanks everyone.

Issue #4645, opened 2018-08-20
Distribute source code using .tar.xz archives


ghost:
Currently Electrum source code is distributed using .tar.gz archives. In my opinion Electrum should use .tar.xz archives instead.

On my system:

    $ du -k Electrum-3.2.2.tar.??
    5488    Electrum-3.2.2.tar.gz
    3924    Electrum-3.2.2.tar.xz

By using the XZ format we can save about 29% of disk space and bandwidth.

What do you think?

SomberNight:
See discussion in https://github.com/spesmilo/electrum/pull/5465.
TL;DR: the files aren't large in either case, `tar.gz` might be better supported, and it's extra complexity (/potential maintenance) to add a new format.

PR #5465, opened 2019-06-27
Build: create a .tar.xz archive

JeremyRand:
Fixes https://github.com/spesmilo/electrum/issues/4645

### DIFF
--- a/contrib/make_tgz
+++ b/contrib/make_tgz
@@ -27,5 +27,5 @@ fi
     # we could build the kivy atlas potentially?
     #(cd electrum/gui/kivy/; make theming) || echo "building kivy atlas failed! skipping."
 
-    python3 setup.py --quiet sdist --format=zip,gztar
+    python3 setup.py --quiet sdist --format=zip,gztar,xztar
 )


SomberNight:
I get that xz compresses better than gz but is there really a point in having 3 distributables?
Can we maybe eliminate at least one of maybe even two of these?
Are there compatibility issues with some of these formats on some platforms?

JeremyRand:
> I get that xz compresses better than gz but is there really a point in having 3 distributables?
> Can we maybe eliminate at least one of maybe even two of these?
> Are there compatibility issues with some of these formats on some platforms?

@SomberNight To the best of my knowledge, all recent GNU/Linux distros support all 3 (though I think `.zip` is a bad idea for GNU/Linux because it doesn't preserve all the permission metadata), and Windows only supports `.zip`.  So it's probably safe to remove `.tar.gz`.  I haven't verified this though, and I have no idea what the situation on macOS (or other more rare OS's) is.  Would you like me to remove `.tar.gz` as part of this PR?

JeremyRand:
According to StackExchange macOS 10.9 and higher support `.tar.xz` natively.

JeremyRand:
AFAICT, Debian Stretch KDE uses the `ark` package for decompressing stuff, which recommends `unar`, which supports both `gz` and `xz`.  So I think `.tar.xz` should be roughly the same as `.tar.gz` in terms of compatibility for GNU/Linux.

JeremyRand:
I recently learned that the Ubuntu packaging team recommends that developers distribute source code tarballs as `.tar.gz`.  So, in order to make it easier for distro packagers' workflow, we probably shouldn't drop `.tar.gz`.  Mildly annoying having 3 archives, but probably doesn't do much harm.

SomberNight:
Note: we stopped distributing `.zip` source archives in https://github.com/spesmilo/electrum/commit/c5c8ea15bbb2df18f9ac600a9f64bd43f05f81af, since it is difficult to build them reproducibly.

> I recently learned that the Ubuntu packaging team recommends that developers distribute source code tarballs as `.tar.gz`

I haven't tested reproducibility of `.tar.xz` sdists, but regardless, AFAICT the main advantage would be saving storage/bandwidth for users that download them, due to better compression. The size difference is <2x, with files in the 10 MB ballpark.
Not sure adding the extra format in addition to or replacing the current one (`tar.gz`) is worth it.

PR #5467, opened 2019-06-28
Travis CI: Use flake8 to find Python syntax errors and undefined names

cclauss:
Also the __sudo:__ tag is now deprecated in Travis CI.

Identity is not the same thing as equality in Python:

$ python
```python
>>> platform = "andr"
>>> platform += "oid"
>>> platform == "android"
True
>>> platform is "android"
False
```

### DIFF
--- a/electrum/plugins/ledger/qt.py
+++ b/electrum/plugins/ledger/qt.py
@@ -47,7 +47,7 @@ class Ledger_Handler(QtHandlerBase):
         else:
             self.word = str(response[0])
         self.done.set()
-    
+
     def message_dialog(self, msg):
         self.clear_dialog()
         self.dialog = dialog = WindowModalDialog(self.top_level_window(), _("Ledger Status"))
@@ -66,18 +66,18 @@ class Ledger_Handler(QtHandlerBase):
         dialog.exec_()
         self.word = dialog.pin
         self.done.set()
-                    
+
     def get_auth(self, data):
         self.done.clear()
         self.auth_signal.emit(data)
         self.done.wait()
         return self.word
-        
+
     def get_setup(self):
         self.done.clear()
         self.setup_signal.emit()
         self.done.wait()
-        return 
-        
+        return
+
     def setup_dialog(self):
         self.show_error(_('Initialization of Ledger HW devices is currently disabled.'))
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: true
 dist: xenial
 language: python
 python:
@@ -29,7 +28,6 @@ jobs:
   include:
     - stage: binary builds
       name: "Windows build"
-      sudo: true
       language: c
       python: false
       env:
@@ -72,7 +70,6 @@ jobs:
       after_script: ls -lah dist && md5 dist/*
       after_success: true
     - name: "AppImage build"
-      sudo: true
       language: c
       python: false
       services:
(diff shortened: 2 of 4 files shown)


electrum/plugins/ledger/qt.py
```diff
@@ -1,6 +1,6 @@
 from functools import partial
 
-#from btchip.btchipPersoWizard import StartBTChipPersoDialog
```
SomberNight: This line was commented out on purpose.
`StartBTChipPersoDialog` was only used from unreachable code, and it was made unreachable because `btchip.btchipPersoWizard` would need updating for it to work. (specifically `btchip.btchipPersoWizard` used PyQt4, and we migrated to PyQt5 around 2 years ago)

I have now removed this stuff in https://github.com/spesmilo/electrum/commit/f405c3fbddc813837403f7cfae3ccb6c1145f838

SomberNight:
I'd merge the `sudo: true` removals, and the `Use ==/!= to compare str, bytes, and int literals` commits, but the flake8 stuff needs more rationale and explanation what the specific tests are.

cclauss:
On the test selection, I have no interest in "_style violations_" (the majority of flake8 error codes that [__python/black__](https://github.com/python/black) can autocorrect).  Instead these tests focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).
* F72 tests syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

JeremyRand:
Concept ACK on E902 and E999, which I infer are the only E9 checks run by flake8.  Concept ACK on F631, F632, and F633, which I infer are the only F63 tests.  Concept ACK on F721, F722, and F723, which I infer are the only F72 tests.  Concept ACK on F821, F822, and F823, which I infer are the only F82 tests.

See https://gitlab.com/pycqa/flake8/blob/master/docs/source/user/error-codes.rst for the list of error codes and what checks they correspond to.

(Personally I'd like to see a larger number of checks, but the list of checks I Concept ACKed above is certainly better than no checks, so it's fine with me to defer the question of adding additional checks for a follow-up PR.)

This PR appears to contain a merge commit; I'd prefer that the PR be rebased so that the history is cleaner.

SomberNight:
I'm not fully convinced we want flake8 but the tests here seem minimal. I guess we will see how useful it is going forward :) Thanks

cclauss:
Two additional codes that you might consider adding to the __—select__ list:
```
89    E722 do not use bare 'except'
256   F401 '.util.format_satoshis' imported but unused
```

JeremyRand:
> Two additional codes that you might consider adding to the —select list:

@cclauss Mind creating a new issue about those, so it doesn't get lost/forgotten?

JeremyRand:
Retrospective +1 on this PR.  In the 11 days since it was merged, it's already been responsible for catching 1 bug in Namecoin's fork of Electrum, as well as 2 bugs in a pending PR I filed here, all of which were likely to remain unnoticed for a while without this PR.

cclauss:
I noticed that the _Identity is not the same thing as equality in Python_ is going to throw SyntaxWarnings in Python >= 3.8 so it is good that we already have them cleaned up.

PR #5534, opened 2019-07-27
gui/kivy: fix installwizard back button behaviour

zebra-lucky:
fix android back button behavior in gui/kivy/uix/dialogs/installwizard.py

As in https://github.com/akhavr/electrum-dash/issues/88

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/installwizard.py
+++ b/electrum/gui/kivy/uix/dialogs/installwizard.py
@@ -533,6 +533,7 @@ class WizardDialog(EventsDialog):
     crcontent = ObjectProperty(None)
 
     def __init__(self, wizard, **kwargs):
+        self.auto_dismiss = False
         super(WizardDialog, self).__init__()
         self.wizard = wizard
         self.ids.back.disabled = not wizard.can_go_back()
@@ -540,7 +541,8 @@ class WizardDialog(EventsDialog):
         self.run_next = kwargs['run_next']
         _trigger_size_dialog = Clock.create_trigger(self._size_dialog)
         Window.bind(size=_trigger_size_dialog,
-                    rotation=_trigger_size_dialog)
+                    rotation=_trigger_size_dialog,
+                    on_keyboard=self.on_keyboard)
         _trigger_size_dialog()
         self._on_release = False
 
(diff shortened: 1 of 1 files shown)


electrum/gui/kivy/uix/dialogs/installwizard.py
```diff
@@ -562,6 +565,20 @@ def add_widget(self, widget, index=0):
         else:
             self.crcontent.add_widget(widget, index=index)
 
+    def on_keyboard(self, instance, key, keycode, codepoint, modifier):
+        if key == 27:
+            if self.can_go_back:
```
SomberNight: `can_go_back` is a method, so this will always be True

SomberNight: ahh no, disregard. there are two of them :)
`self.wizard.can_go_back()` and `self.can_go_back`
is `self.can_go_back` needed?

zebra-lucky: Yes, maybe I should use wizard method.
I'll post changes in a 5 minutes.

zebra-lucky:
Removed can_go_back, use wizard method.
Rebased and squashed changes.

electrum/gui/kivy/uix/dialogs/installwizard.py
zebra-lucky: Removed can_go_back, use wizard method.

ecdsa:
how can the initial issue be reproduced?

zebra-lucky:
On Android fresh install exit from installwizard on back button.
On Linux same behaviour can be tested with `Esc` button.

ecdsa:
@SomberNight could you reproduce the issue?

zebra-lucky:
Yes, I reproduced the problem on Android and Linux on 3.3.8.

zebra-lucky:
Not sure about freezing, however.
Only exit from installwizard was reproduced and fixed.

ecdsa:
@zebra-lucky can you post the traceback?

zebra-lucky:
No, I not reproduced any tracebacks.
Only clean exit from app on this event from popup:

```python
    def on_dismiss(self):
        app = App.get_running_app()
        if app.wallet is None and not self._on_release:
            app.stop()
```

ecdsa:
oh I see. I was not understanding the issue. yes, I can reproduce.

SomberNight:
Callbacks bound to Window need to be unbound too. See https://github.com/spesmilo/electrum/commit/95ae42b998f6abbe8667d9032f2084f78ad009bf
This was missing even before this PR but it was harmless then.
Without fix, pressing back button after creating a new wallet would go back to the wizard (after it was supposedly done).

zebra-lucky:
Thanks, I missed unbinding when doing patch.

zebra-lucky:
Found another bug. Need to call `self.close()` before go back. As in:
```python
        if key == 27:
            if self.wizard.can_go_back():
                self.close()
                self.wizard.go_back()
```

And there is a problem with `App.on_stop`.
If in pause mode and app window is closed, then `on_stop` method is not called.
Maybe should call `save_last_wallet` from `on_pause`
```
        if self.wallet:
            self.electrum_config.save_last_wallet(self.wallet)
```
And maybe also should call `network.stop` from `on_pause` and `network.start` from `on_resume`

Will do some PR in couple of hours.

PR #5543, opened 2019-08-04
Added ELECTRUMDIR env variable

Jorengarenar:
Simple way for allowing users to change localization of Electrum directory from ~/.electrum to somewhere else

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -469,7 +469,9 @@ def bh2u(x: bytes) -> str:
 
 
 def user_dir():
-    if 'ANDROID_DATA' in os.environ:
+    if "ELECTRUMDIR" in os.environ:
+        return os.environ["ELECTRUMDIR"]
+    elif 'ANDROID_DATA' in os.environ:
         return android_data_dir()
     elif os.name == 'posix':
         return os.path.join(os.environ["HOME"], ".electrum")


SomberNight:
You can already set the location of the data directory ("Electrum directory") on the CLI as an option:
https://github.com/spesmilo/electrum/blob/e34afd62cecf7ef0e191fb12e5ebcfe2a1dfb7cb/electrum/commands.py#L937
Do you think this would still be useful?

Jorengarenar:
> You can already set the location of the data directory ("Electrum directory") on the CLI as an option:
> 
> https://github.com/spesmilo/electrum/blob/e34afd62cecf7ef0e191fb12e5ebcfe2a1dfb7cb/electrum/commands.py#L937
> 
> Do you think this would still be useful?

Yes, of course it still would be useful. This works in CLI, so I need to create alias to not type it all the time. And this works only in CLI, right? What about menu? Menus won't read aliases. I need to either edit existing `.desktop` file (what I shouldn't do, because of other users) or create a new one. Later I can also decide to install it in another path, so I need to modify created `.desktop` file. Instead of `alias+file` one env variable would be much more convenient to mange. Also it just works no matter if I decide to install or run without installation (again alias and file changing). When I decide to move dir, I need to remember to modify one place instead of two.

And I think I should have written this all at the very beginning...

Issue #5227, opened 2019-03-28
pip install desktop integration broken. icon/launcher is copied to incorrect path


SomberNight:
Split off from https://github.com/spesmilo/electrum/issues/4768

> I found the only was was to use the steps in the README which were:
`pip3 install .[fast]`
>But this still does not add a menu item or icons as far as I can tell so you have to copy the icon manually from archive and then add a menu or shortcut manually.

If I run
```
python3 -m pip install --user Electrum-3.3.4.tar.gz[fast]
```
the `data_files` passed to `setup()` in setup.py has the value:
```
data_files = (
    [('/home/user/.local/share/applications/', ['electrum.desktop']), 
     ('/home/user/.local/share/icons', ['electrum/gui/icons/electrum.png'])]
)
```
but pip/setuptools turns these absolute paths into relative compared to `site-packages`, e.g.
`~/.local/lib/python3.6/site-packages/home/user/.local/share/applications/`,
and that is where the files end up being copied to.
Which in turn, of course, breaks the desktop-integration.

This looks to be this issue: https://github.com/pypa/wheel/issues/92
And for us specifically it "broke" when we started instructing people to use
`python3 -m pip install` instead of `sudo python3 setup.py install`

ThomasLandauer:
Workaround:

```bash
mv ~/.local/lib/python3.6/site-packages/home/<USERNAME>/.local/share/applications/electrum.desktop ~/.local/share/applications/electrum.desktop
mkdir ~/.local/share/icons
mv ~/.local/lib/python3.6/site-packages/home/<USERNAME>/.local/share/icons/electrum.png ~/.local/share/icons/electrum.png
```

SomberNight:
Note: https://github.com/pypa/wheel/issues/92#issuecomment-410664879 says,

> here's a few things I think you have to watch out for when testing your solution:
>
>    - different python versions
>    - wheel vs no-wheel
>    - some distributions handle site-packages differently (for example, compare Arch vs Ubuntu. I think ubuntu uses dist-packages?)

And indeed the bug described here is only present if the `wheel` package is installed.
Without `wheel`, the absolute paths in `data_files` are respected.

PR #5553, opened 2019-08-10
Logic change in setup and fixed icon location

A6GibKm:
This should be `or` since in some cases only one of the dirs is read only, e.g. for `root_path=/`, `usr_share=/usr/share` in flatpak-builder.

The correct path for icons is one of `{XDG_DATA_HOME,/usr/share,~/.local/share}/icons/hicolor/RESOLUTION/apps` according to freedeskop standars https://www.freedesktop.org/wiki/Specifications/.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,10 +37,9 @@ if platform.system() in ['Linux', 'FreeBSD', 'DragonFly']:
     parser.add_argument('--root=', dest='root_path', metavar='dir', default='/')
     opts, _ = parser.parse_known_args(sys.argv[1:])
     usr_share = os.path.join(sys.prefix, "share")
-    icons_dirname = 'pixmaps'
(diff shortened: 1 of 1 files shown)


SomberNight:
Does this block of code in question even work now? See https://github.com/spesmilo/electrum/issues/5227
Did you test that the icon/launcher are being copied to the expected paths?

A6GibKm:
I made some tests, 
> Does this block of code in question even work now?
It depends of the path were python installs stuff, with sudo or without sudo it installs everything just fine (under `./local/share` or `/usr/share`). If I set `$PYTHONUSERBASE=~/.python` the icons get installed under
`/home/user/.python/lib/python3.7/site-packages/home/use/.local/share/icons/hicolor/128x128/apps/electrum.png` which is weird to say the least.

PR #5572, opened 2019-08-19
Kivy installwizard back button fix2

zebra-lucky:
- kivy: dismiss previous wizard window on back button (additional fix on #5534)
- ~kivy: stop network on pause, save last wallet, start network on resume~

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -896,6 +896,8 @@ class ElectrumWindow(App):
         # pause nfc
         if self.nfcscanner:
             self.nfcscanner.nfc_disable()
+        if self.wallet:
+            self.electrum_config.save_last_wallet(self.wallet)
         return True
 
     def on_resume(self):
--- a/electrum/gui/kivy/uix/dialogs/installwizard.py
+++ b/electrum/gui/kivy/uix/dialogs/installwizard.py
@@ -607,6 +607,8 @@ class WizardDialog(EventsDialog):
     def on_keyboard(self, instance, key, keycode, codepoint, modifier):
         if key == 27:
             if self.wizard.can_go_back():
+                self._on_release = True
+                self.dismiss()
                 self.wizard.go_back()
             else:
                 app = App.get_running_app()
@@ -631,7 +633,7 @@ class WizardDialog(EventsDialog):
 
     def on_release(self, button):
         self._on_release = True
-        self.close()
+        self.dismiss()
         if not button:
             self.parent.dispatch('on_wizard_complete', None)
             return


ecdsa:
please note that network now has more jobs than self.fx.run: watchtower and jsonrpc
although these jobs are not wanted in the context of an android app, it would be cleaner to stop and restart the list of jobs.

zebra-lucky:
What about the wallet?
I do minimal changes. Need to do more job on pause.

zebra-lucky:
Changes in first commit dismiss current wizard page, when switching back.
Found current page is remains open (In `on_release` method current page is
closed by `close`/`dismiss`.

Second commit need to be revised. I Found after `on_pause`, `on_stop` is not called,
so possible situations when last working code in app is `on_pause`.
Maybe need add more code to write all changes to wallet.

About network part, Dash version saves additonal data on network stop,
so I added net stop/start.
And if app go in pause mode, it is logically to stop network, as timeouts will
happen on network connections after some time, and on resume reconnection
still need.

zebra-lucky:
I can remove second commit and rebase code to current master, to simplify merge.

Other changes I need to revise and possibly make another PR

zebra-lucky:
I've removed second commit and rebased code on current master branch.

> Do you have a way to trigger some weird behaviour that gets fixed by this change?
No, it`s not gave effects for me, only found by code review.

zebra-lucky:
P.S.: I have small android practice and misunderstand pause mode. After reading docs and debugging I understand that only activity thread is paused, other threads continue to work, so need to stopping network in pause mode was a mistake.

zebra-lucky:
P.P.S: But second problem is leaves. When on_pause is called and app is closed by UI (cross button), then on_stop is not called. So I added commit to save last wallet.

SomberNight:
@ecdsa looking at https://github.com/spesmilo/electrum/commit/b6f63e1abf2cc3faa9c504e73af656889b9fab9e, this PR looks related :P

ecdsa:
> @ecdsa looking at [b6f63e1](https://github.com/spesmilo/electrum/commit/b6f63e1abf2cc3faa9c504e73af656889b9fab9e), this PR looks related :P

Yes, it is just a pity that the OP was not able to properly explain what the issue was, it would have saved some time..
The bug was introduced in e953eebdb96e8657886e0daa3e7a8224c982cad5, where `on_keyboard` is added.

ecdsa:
@zebra-lucky regarding the other commit in this PR, note that  `save_last_wallet` does not save the wallet file, it only saves the name of the last wallet in the config file. From your comments I supppose that you actually intended to save the wallet file.

PR #5582, opened 2019-08-25
Use NamedTuple notation for TxOutput in test_lnchannel

JeremyRand:
This makes the code more resilient in case additional members are added to TxOutput later.


JeremyRand:
(Triggering a rebuild....)

PR #5593, opened 2019-08-28
Fix travis error

wakiyamap:
I fix this error.
``(standard_in) 1: syntax error``
https://travis-ci.org/spesmilo/electrum/jobs/577905117#L481


ecdsa:
thanks!

PR #5616, opened 2019-09-04
Split out common JSON-RPC server code.

domob1812:
This adds back a `jsonrpc.py` file, holding common code for JSON-RPC servers as they are used both for the `WatchTowerServer` and the core `Daemon`.  By doing so, we avoid duplicating that code (even if it is not so much), and ease possible future maintenance.

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -41,13 +41,14 @@ from jsonrpcserver import response
 from jsonrpcclient.clients.aiohttp_client import AiohttpClient
 
 from .network import Network
-from .util import (json_decode, to_bytes, to_string, profiler, standardize_path, constant_time_compare)
+from .util import json_decode, to_bytes, to_string, profiler, standardize_path
 from .util import PR_PAID, PR_EXPIRED, get_request_status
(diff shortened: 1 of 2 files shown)


domob1812:
Travis failures seem unrelated

ecdsa:
thanks

ecdsa:
not going to merge this, it does not save enough lines with the current code.
note that the handle methods are now different.

domob1812:
> not going to merge this, it does not save enough lines with the current code.

That's your call of course.  But note that I see another benefit (apart from saving code) in separating the logic that "just does a JSON-RPC server" from the "actual logic".  IMHO that makes the code easier to read, as you don't have to keep the JSON-RPC server stuff in your mind when reading `daemon.py`.  (And I assume that most readers / changes will happen to `daemon.py` itself, not the JSON-RPC server.)

> note that the handle methods are now different.

I don't see a difference in the current `master` branch (except for details about error reporting and the lack of authentication for the `WatchTowerServer`, which could probably be unified anyway).

PR #5628, opened 2019-09-09
Digital BitBox: Fix sending to self

TheCharlatan:
Make sure that a pubkey is only appended to the checkpub array if it corresponds to a change address. Signing will fail otherwise, if a non-change pubkey is sent as part of the checkpub list to the Digital BitBox.


SomberNight:
Is this due to a change in recent firmware? I think this used to work.

douglasbakkum:
Yes, there was a change in firmware that limited the range of BIP44 key paths that can be accessed: https://github.com/digitalbitbox/mcu/pull/266/files#diff-d62a720ed7d6ea48bbf4687b4dc4e7b3R157

The rationale is given in this blog post: https://medium.com/shiftcrypto/bitbox-desktop-app-4-6-0-with-firmware-6-0-3-release-ec46937afe7c

SomberNight:
Okay.
Tested.
Thank you.

PR #5631, opened 2019-09-12
wallet: organise get_tx_fee. store calculated fees. storage version 19.

SomberNight:
- in wallet file, under `tx_fees` key is now a dictionary:
  `txid -> (fee, is_calculated_by_us, num_inputs)`
- new wallet file version: 19
    - upgrade function simply deletes old `tx_fees` key
- we now store transaction fees that we calculate, in the wallet file, in above structure
    - see in particular `AddressSynchronizer.get_tx_fee` which is now kind of complicated... it handles all the logic for all current use cases


ecdsa:
thanks

PR #5640, opened 2019-09-13
config: introduce ConfigVar enum. default values are no longer duped

SomberNight:
On current master, every time we lookup the value of a config variable, we potentially duplicate the corresponding default value:
https://github.com/spesmilo/electrum/blob/c7346c1eb8f790ad71d14c634097fa67d2839ac5/electrum/simple_config.py#L184
Further, almost every time we get/set a config variable, we use a new inline string literal for the config key; which means the config keys are also duplicated all over the place.

This PR introduces a `ConfigVar` enum, which lists almost all configuration variables in a single place in the code, along with default values and config key strings on the same line for each.

### DIFF
--- a/electrum/base_crash_reporter.py
+++ b/electrum/base_crash_reporter.py
@@ -36,7 +36,6 @@ from .logging import describe_os_version, Logger
 
 class BaseCrashReporter(Logger):
     report_server = "https://crashhub.electrum.org"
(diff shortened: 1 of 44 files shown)


SomberNight:
closing in favour of https://github.com/spesmilo/electrum/pull/5649

PR #5649, opened 2019-09-20
config: introduce ConfigVar. 2nd approach

SomberNight:
This is an alternative approach to https://github.com/spesmilo/electrum/pull/5640

Usage:

```python
>>> from electrum.simple_config import SimpleConfig
>>> c = SimpleConfig()
>>> c.NETWORK_OFFLINE
<electrum.simple_config.ConfigVarWithConfig object at 0x0000023258A1DF60>
>>> c.NETWORK_OFFLINE.value
False
>>> c.NETWORK_OFFLINE.get_key_str()
'offline'
>>> SimpleConfig.NETWORK_OFFLINE.get_key_str()
'offline'
>>> c.get('offline')
False
>>> c.NETWORK_OFFLINE.is_modifiable()
True
>>> c.NETWORK_OFFLINE.is_defined()
False
>>> c.NETWORK_OFFLINE.value = 1
>>> c.NETWORK_OFFLINE.value
1
>>> c.get('offline')
1
>>> c.NETWORK_OFFLINE.is_defined()
True
```

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -5,7 +5,7 @@ import os
 import stat
 import ssl
 from decimal import Decimal
-from typing import Union, Optional
+from typing import Union, Optional, Any, Dict
 from numbers import Real
 
 from copy import deepcopy
@@ -13,7 +13,8 @@ from copy import deepcopy
 from . import util
 from . import constants
 from .util import (user_dir, make_dir,
-                   NoDynamicFeeEstimates, format_fee_satoshis, quantize_feerate)
+                   NoDynamicFeeEstimates, format_fee_satoshis, quantize_feerate,
+                   DECIMAL_POINT_DEFAULT)
 from .i18n import _
 from .logging import get_logger, Logger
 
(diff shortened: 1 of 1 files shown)


ecdsa:
```
class my_property(property):

    def __init__(self, key, default):
        self._key = key
	self._default = default
        property.__init__(self, self.__get_x_, self.__set_x_)
        print(dir(self))

    def __get_x_(self, x):
        return x.properties.get(self._key, self._default)

    def __set_x_(self, x, y):
        x.properties[self._key] = y
        x.save()

    def key(self):
        return self._key

lass Config(object):

    properties = {}
    not_modifiable = ['blah']
    use_rbf = my_property('use_rbf2', 45)

    def save(self):
	print(self.properties)

    def is_set(self, key):
	return key in self.properties

    def is_set2(self, name):
	klass = getattr(Config, name)
	key = klass.key()
	return key in self.properties

    def is_modifiable(self, key):
	return key not in self.not_modifiable

c = Config()
print(c.use_rbf)
#c.use_rbf = 42                                                                                                                                                                               
print(c.is_modifiable(Config.use_rbf.key()))
print(c.is_set(Config.use_rbf.key()))
print(c.is_set2('use_rbf'))

```

ecdsa:
@SomberNight I simplifiedthe code,  `is_set` and `is_modifiable` take the class instead of the key

SomberNight:
> `c.is_modifiable(Config.use_rbf)`

Note: that is now similar to what the 1st PR did:
`config.is_modifiable(ConfigVar.WALLET_USE_RBF)`

(except you propose only to use it for certain less frequent methods; not for get/set)

richardsj:
` "history_rates": "checked"`

Set from versions < 4.4.0 (e.g. Electrum 4.3.4) causes an error "Cannot load wallet(2) :
TypeError("setChecked(self, bool): argument 1 has unexpected type 'str'")"

Temporary work-around is to **remove this line from the wallet config** manually i.e. delete the line above from the file `~/.electrum/config` but this error should not occur. Additionally the wallet is upgraded, thereby making the older Electrum no longer load it.

See #8367 for detail

SomberNight:
@richardsj thanks but it would have been better to comment that in https://github.com/spesmilo/electrum/issues/8367
This is a generic PR revamping how the config works, and your comment is about a specific bug regarding one config variable.

Issue #5629, opened 2019-09-09
SimpleConfig is supposed to be a singleton but we also create temporary SimpleConfig instances in some cases


SomberNight:
There is a process-global `config` object at any given time:
https://github.com/spesmilo/electrum/blob/e5502a58ba8fb7302f9b869981f6b23637daa6f9/electrum/simple_config.py#L34-L45
Every time `SimpleConfig.__init__` runs, we update the global `config` reference to `self`:
https://github.com/spesmilo/electrum/blob/e5502a58ba8fb7302f9b869981f6b23637daa6f9/electrum/simple_config.py#L123-L124

There is some code that uses `get_config` expecting to get the "correct" `SimpleConfig` object:
(for example)
https://github.com/spesmilo/electrum/blob/e5502a58ba8fb7302f9b869981f6b23637daa6f9/electrum/wallet.py#L1326-L1328

And there are also places where we create a temporary/transient `SimpleConfig` object.
For example, every time a CLI command runs:
https://github.com/spesmilo/electrum/blob/e5502a58ba8fb7302f9b869981f6b23637daa6f9/electrum/daemon.py#L429-L432

Clearly this can lead to bugs.

Note above examples are just examples; i.e. there might be more!

ecdsa:
I believe the instance creation in run_cmdline is not needed anymore.

PR #5652, opened 2019-09-22
config: no longer singleton. it is passed to Wallet.__init__

SomberNight:
SimpleConfig is no longer a singleton.

Everywhere we previously used SimpleConfig.get_instance(), now either get passed a config instance, or we try to get a reference to something else that has a reference to a config.
(see wallet, lnsweep, qt/qrcodewidget, qt/qrtextedit)


SomberNight:
related: https://github.com/spesmilo/electrum/issues/5629

ecdsa:
ah thanks for the pointer

PR #5660, opened 2019-09-26
Optional verification in gettransaction command

JeremyRand:
It is likely that a significant subset of users of the `gettransaction` command would like to not trust the server.  This PR adds an optional `verify` argument (disabled by default) to `gettransaction`, which when enabled verifies the transaction via the SPV verifier.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -49,6 +49,7 @@ from .paymentrequest import PR_PAID, PR_UNPAID, PR_UNKNOWN, PR_EXPIRED
 from .synchronizer import Notifier
 from .wallet import Abstract_Wallet, create_new_wallet, restore_wallet_from_text
 from .address_synchronizer import TX_HEIGHT_LOCAL
+from .verifier import SPV
 from .mnemonic import Mnemonic
 from .lnutil import SENT, RECEIVED
 from .lnpeer import channel_id_from_funding_tx
@@ -679,17 +680,29 @@ class Commands:
         return out
 
     @command('n')
-    async def gettransaction(self, txid, wallet: Abstract_Wallet = None):
+    async def gettransaction(self, txid, verify=False, height=None, wallet: Abstract_Wallet = None):
         """Retrieve a transaction. """
+        if verify:
+            if height is None:
+                raise Exception("Missing height")
+            verifier = SPV(self.network, None)._request_and_verify_single_proof(txid, height)
         tx = None
         if wallet:
             tx = wallet.db.get_transaction(txid)
         if tx is None:
-            raw = await self.network.get_transaction(txid)
+            raw_getter = self.network.get_transaction(txid)
+            if verify:
+                _, raw = await asyncio.gather(verifier, raw_getter)
+            else:
+                raw = await raw_getter
             if raw:
                 tx = Transaction(raw)
             else:
                 raise Exception("Unknown transaction")
+        elif verify:
+            await verifier
+        if verify and tx.txid() != txid:
+            raise Exception("Wrong txid")
         return tx.as_dict()
 
     @command('')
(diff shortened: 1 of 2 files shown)


JeremyRand:
(Feel free to take your time on reviewing this if you're busy with Lightning etc.; it's not super high priority for me at the moment.)

SomberNight:
I would suggest simply adding a `tx.txid() == txid` check at the end of the current method; which would be a ~2 line change. That can get merged instantly, as it is a no-brainer.
The current method does not test that the returned transaction actually corresponds to the input txid, which should absolutely be fixed, but regarding the addition of an SPV check, I am not sure what its purpose is.

Could you elaborate on at least one example usecase, where the SPV check in addition to the txid check would make sense?

The current method has an API: txid -> raw_tx (broadly speaking)
By adding the SPV check and providing `height` as a new argument, I would argue you are doing too many things in this method. See e.g. the `get_tx_status` command; I guess you are trying to achieve something like that (except that only works for in-wallet transactions atm) but as part of this existing kind-of unrelated method.

electrum/commands.py
```diff
@@ -679,17 +680,29 @@ def _mktx(self, wallet: Abstract_Wallet, outputs, *, fee=None, feerate=None, cha
...
+            else:
+                raw = await raw_getter
             if raw:
                 tx = Transaction(raw)
             else:
                 raise Exception("Unknown transaction")
+        elif verify:
+            await verifier
+        if verify and tx.txid() != txid:
+            raise Exception("Wrong txid")
```
SomberNight: As in comment; why can't we just do this; without the other checks?

SomberNight:
> I would suggest simply adding a tx.txid() == txid check at the end of the current method; which would be a ~2 line change. That can get merged instantly, as it is a no-brainer.
The current method does not test that the returned transaction actually corresponds to the input txid, which should absolutely be fixed, but regarding the addition of an SPV check, I am not sure what its purpose is.

I've pushed just this fix in https://github.com/spesmilo/electrum/commit/f72bb03af6d7f06a5d3fc911c806662ff0b4cd36

JeremyRand:
> Could you elaborate on at least one example usecase, where the SPV check in addition to the txid check would make sense?

@SomberNight very sorry for the delayed reply on this; things have been hectic here and they're probably not going to get much less hectic for at least a few more weeks.  Anyway, the general class of use cases is that Alice wants to show Bob some transaction in the blockchain, and Bob wants to see the transaction and also wants to be sure that the transaction really is in the blockchain like Alice said.  The transaction could be, for example, a transaction with an interesting scriptPubKey (e.g. the famous SHA1 pinata that Peter Todd created), or it could be a transaction whose signature encodes a hash that is being timestamped (which is what OpenTimestamps does).  In cases like this (particularly the latter example with a timestamp), knowing that the returned transaction really is in the blockchain is critical to the result being actually useful to Bob.

> See e.g. the `get_tx_status` command; I guess you are trying to achieve something like that (except that only works for in-wallet transactions atm) but as part of this existing kind-of unrelated method.

I actually had not noticed that method while writing this patch.  That said, the intended use case for this patch includes situations where the user explicitly does not want to save the transaction to the wallet, so `get_tx_status` in its current form isn't going to be adequate.  Maybe it could be refactored to apply to non-wallet transactions too; not totally sure.

Does the example use case I provided makes sense?  Do you think there's a better way to organize the functionality?

SomberNight:
> Alice wants to show Bob some transaction in the blockchain, and Bob wants to see the transaction and also wants to be sure that the transaction really is in the blockchain like Alice said
> Does the example use case I provided makes sense? Do you think there's a better way to organize the functionality?

I think we should introduce a new command that takes a txid (and maybe a block height), and returns whether the txid could be SPV verified -- specifically I think returning the number of confirmations might be more useful than a boolean (and return either zero or raise if SPV fails).

ecdsa:
this makes me think that the SPV class should probably not have a reference to wallet

Issue #5418, opened 2019-06-10
Implement Shamir backup ability for Trezor Model T


slush0:
Upcoming release of Trezor Model T will contain ability to create Shamir Backup and recover the seed into Trezor from such shares. It would be awesome to have support for Shamir Backup ability in create wallet / restore wallet for Trezor in Electrum, so users won't need to use official web wallet.

I kindly nominate @matejcik to implement this feature :-).

fresheneesz:
Not sure if this might be helpful, but I found this very old post with instructions on how to do fragmented backups with electrum: https://bitcointalk.org/index.php?topic=526381.0

SomberNight:
This is now implemented with https://github.com/spesmilo/electrum/pull/5692 merged

Issue #5419, opened 2019-06-10
Implement support for SLIP 39 (Shamir backup)


slush0:
This is related to #5418.

It would be awesome to have the ability to (at least) recover SLIP-39 (Shamir Backup) directly into Electrum, as a disaster recovery option for users who lost an access to Trezor.

I understand Thomas is not a fan of BIP-39 and he'll probably dislike SLIP-39 for the same reasons, so I'm not proposing *creating* SLIP-39 shares by Electrum, but having the ability to recover into Electrum would be just amazing.

I kindly nominate @matejcik for this task :-).

fresheneesz:
> Thomas is not a fan of BIP-39 and he'll probably dislike SLIP-39 for the same reasons

What reasons are those?

SomberNight:
>> Thomas is not a fan of BIP-39 and he'll probably dislike SLIP-39 for the same reasons

> What reasons are those?

I guess this has the potential to quickly turn into off topic... :) see https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2016-August/013061.html

fresheneesz:
Those seem like well thought out complaints. Is there any alternative proposal that doesn't have those issues?

SomberNight:
As OP suggests, just as Electrum doesn't support creating BIP39 mnemonics, but does support recovering from them, we could have support for SLIP39 in a similar fashion.
A pull request that implements recovering from SLIP39 backups would be welcome.

Note that during recovery/restoring, with Electrum seeds and BIP39 seeds, there is only "one" seed to enter. With SLIP39, there are usually multiple shamir shares.
Hence this will probably involve some GUI work.

andrewkozlik:
Has anyone started working on this? @brianddk, @matejcik? If not, then I might have a shot at this myself.
> Note that during recovery/restoring, with Electrum seeds and BIP39 seeds, there is only "one" seed to enter. With SLIP39, there are usually multiple shamir shares.
> Hence this will probably involve some GUI work.

The text box for seed entry is multiline, so we can simply ask the user to type one share per line. There is already a status label below the text box, which we can use to inform the user about the number of missing shares, invalid checksum, mismatching shares etc. So I don't think any GUI work is necessary to get this up and running. Of course if anyone wants to play around with the GUI to create a multidialog walkthrough, then that can easily go into another PR later on.

brianddk:
[...]

PR #5692, opened 2019-10-10
Trezor: support for Shamir backup and recovery

matejcik:
implements #5418

please let me know if there's a code style tool I should be using?

### DIFF
--- a/contrib/requirements/requirements-hw.txt
+++ b/contrib/requirements/requirements-hw.txt
@@ -1,4 +1,4 @@
-trezor[hidapi]>=0.11.0
+trezor[hidapi]>=0.11.5
 safet[hidapi]>=0.1.0
 keepkey>=6.0.3
 btchip-python>=0.1.26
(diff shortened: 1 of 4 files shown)


matejcik:
rebased and retriggered CI check (which seemed to failed for unrelated reasons before).

SomberNight:
Do you really want Shamir to be the default if it's available?

SomberNight:
Otherwise, the diff itself look okay. :)

One concern I have is that if a user generates a Shamir backup on their Trezor, and they lose their device, they currently can't restore from the backup in software. This is part of the reason I am thinking Shamir should not be default for now.

matejcik:
you are right, we decided to keep the default the same as the current Wallet, even if Shamir is available.

> One concern I have is that if a user generates a Shamir backup on their Trezor, and they lose their device, they currently can't restore from the backup in software.

also true, and #5419 is still far on the roadmap, at least from my side.

made the change and rebased on top of today's master

SomberNight:
I've discussed with @ecdsa, and we think until it is implemented to restore from backup in software (without hw device), the Shamir options should be less visible.

In https://github.com/spesmilo/electrum/pull/5692/commits/18209fc782c08bb4b41d878d8fdedce86d9fa309, I've made these only visible after the user clicks "expert settings".

After https://github.com/spesmilo/electrum/issues/5419, we can make them more visible or even the default.

SomberNight:
Thank you for your work.

PR #5700, opened 2019-10-15
fix expected returned peer address values when opening channel.

fiatjaf:
2019-10-14-225221_549x219_scrot


SomberNight:
Thanks. I made some minor related clean-ups in a follow-up.

PR #5712, opened 2019-10-16
segwit_addr: Use normal comparison for tuple literal

### LINKED ISSUE #5711: SyntaxError in segwit_addr with Python 3.8
```
electrum\segwit_addr.py:121: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  assert decode(hrp, ret) is not (None, None)
```

This error seems to be present since [merging into Electrum](https://github.com/spesmilo/electrum/commit/e853dde1113f5d3fb96b864aca283781abbd6f8e) even though it is not present in [sipa's code](https://github.com/sipa/bech32/blame/bdc264f84014c234e908d72026b7b780122be11f/ref/python/segwit_addr.py#L121). Weird.

ysangkok:
Fix #5711

### DIFF
--- a/electrum/segwit_addr.py
+++ b/electrum/segwit_addr.py
@@ -1,3 +1,4 @@
+
 # Copyright (c) 2017 Pieter Wuille
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


SomberNight:
I think we should do instead what the current upstream is doing:
https://github.com/sipa/bech32/blob/bdc264f84014c234e908d72026b7b780122be11f/ref/python/segwit_addr.py#L121
as that suggest `decode(hrp, ret) == (None, None)` might actually happen

SomberNight:
nvm, I looked at where we actually call this, and maybe the assertion failure is better

PR #5717, opened 2019-10-21
fix potential vulnerability with get_sorted_pubkeys

sidhujag:
txin['pubkeys'] is not always None it is sometimes == [] where len(pubkeys) == 0, so catch it as such. Note it is set up as an array here:
https://github.com/spesmilo/electrum/blob/master/electrum/transaction.py#L454

### DIFF
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -638,7 +638,7 @@ class Transaction:
             return [], []
         x_pubkeys = txin['x_pubkeys']
         pubkeys = txin.get('pubkeys')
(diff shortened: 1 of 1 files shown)


SomberNight:
I don't see a vulnerability here. At best there might be a bug.
Do you have a way to reproduce some issue that this change fixes?

sidhujag:
Yup defensive programming, just a bug. I did find it an issue while syncing, but I am also on a fork, I had to add that to fix it and noticed it was array. Definitely a bug.

Issue #4615, opened 2018-08-08
Implement BIP174 (PSBT) for serialising partial and unsigned transactions


SomberNight:
[BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) is now no longer in "draft" state. It has also been merged into Bitcoin Core (in e.g. https://github.com/bitcoin/bitcoin/pull/13557), and it looks like they will have support for it in 0.17.

We should consider ditching our custom format and implement it as well, as it would be good to be compatible with Bitcoin Core and potentially others.

Related:
https://github.com/spesmilo/electrum/issues/4514
https://github.com/spesmilo/electrum/issues/4614

peter-conalgo:
If it helps, the Coldcard wallet plugin already can make a PSBT: see [function `built_psbt()`](https://github.com/Coldcard/electrum/blob/0c0fbdc72731dbae254b83f6886475533a53a149/electrum/plugins/coldcard/coldcard.py#L368) in the PR <https://github.com/spesmilo/electrum/pull/4470>.

We do somewhat cheat because we didn't do any merging or combining, and just assume the result is a finalized transaction coming back from the Coldcard. (That's a safe assumption given the role of Coldcard in this case as a hardware keystore and signer.) We didn't need a PSBT parser.

Another limitation is the "Save as PSBT" button on the transaction dialog, appears only if the transaction is associated with a Coldcard based wallet. It should be possible for Electrum to offer PSBT output there for almost any unsigned transaction.

Naturally as PSBT support matures in Electrum, we will use those features as much as possible for this plugin. Always happy to delete code!

PS: You may find this debug tool, [`psbt_dump`](https://github.com/peter-conalgo/psbt_dump), to be useful, and @achow101 has been really helpful as well.

SomberNight:
note: @asfin has said he is interested in working on this

roznalex:
Are there any updates?

SomberNight:
there are WIP changes at https://github.com/asfin/electrum/tree/bip174-2

SomberNight:
Note: I am now working on this. Trying to reuse code already written as part of the coldcard plugin, and some stuff from https://github.com/spesmilo/electrum/pull/4883

craigraw:
I think this would be a valuable feature for many. Although I've only tested it with the coldcard, the plugin PSBT support works well, although it does lose partial sigs when loading PSBTs (fix proposed here: #5671). 

[...]

Issue #4514, opened 2018-07-07
unsigned or partially signed transaction from Electrum cannot be interpreted by Bitcoin Core


gits7r:
Question: why does an unsigned transaction or partially signed non-final transaction generated by Electrum cannot be interpreted by Bitcoin Core? Do we change the serialization somehow or add / remove bits of data?

Example:
Unsigned tx by  Electrum:
45505446ff000100000001d61c4afee1080779a2a4957d76fd25f265a4c38aaeb3f2e85274ee9dc13d11ac010000005701ff4c53ff0488b21e0000000000000000003c094e637b4dcf44f6ac41a80ae632c0e5d7c845173572c5e7f24642096dead4032c2f71bad3efa41faf273ad858c663296fafa28ffdebc7ef1f72a4f0572516c000000801fdffffff029a7f1c03000000001976a9145837b116b0dda66ba3ea938ca313d580939765e688ac00879303000000001976a9145837b116b0dda66ba3ea938ca313d580939765e688acbf190800

Bitcoin Core returns:
error code: -22
error message:
TX decode failed

Same error is returned by Bitcoin  Core for partially signed non-final transaction generated by Electrum.

This question is related to #4471 , we should answer this before we decide  a good way to implement that.

gits7r:
@SomberNight -- thank you very much for this. Yes, that is the reason it's not working. I noted that somewhere but apparently missed it last night while doing some tests.

I understand the complexity of reversing this, but instead is there a way for the command line electrum to take a hex unsigned or partially signed transaction and decode it to show:
- outputs
- fee
- unsigned, partially signed, signed
- locktime
- its size in kb and /or weight

Issue #4322, opened 2018-04-28
wallet.is_mine behaviour when offline signing beyond gap limit is undesired


SomberNight:
When offline signing with deterministic wallets, on the offline machine, the tx inputs/outputs will not be recognised as is_mine (if beyond the gap limit).
The signing itself will work, but this still has some consequences.

Notably:
- the tx dialog will not colour the inputs/outputs
- in case of hw wallet, **the hw device will treat all outputs as foreign** (so e.g. change will have to be confirmed as well)
https://github.com/spesmilo/electrum/blob/530980ea37b9627b0ab26b63acc8866b799b23bc/lib/wallet.py#L1477-L1488

See e.g. below picture, where the `Sign` button is enabled, and in fact all inputs/outputs belong to this HD wallet.
![offline_signing_is_mine](https://user-images.githubusercontent.com/29142493/39401042-a2c9154e-4b3c-11e8-9638-a6a4087c3ec8.PNG)

Related: https://github.com/spesmilo/electrum/issues/3302

SomberNight:
This is hard to fix :(
We could easily learn from the serialised transaction the is_mine inputs; as for the signing to work, the inputs already contain derivation paths.
However, the outputs are problematic. The serialisation format does not contain any extra info about them.
Actually, the outputs would be the more important thing to fix, not the inputs, for hardware wallets (so that they don't prompt the user to confirm). It is not really worth it to fix just the inputs.

Suggested workaround for users: manually increase the gap limit on the offline machine.

SomberNight:
To increase the gap limit on the offline machine one can e.g. do
```
wallet.change_gap_limit(100)
wallet.gap_limit_for_change = 100
wallet.synchronize()
```
(note however that after changing the gap limit for change that wallet file should not be used to construct new transactions due to https://github.com/spesmilo/electrum/issues/3859 (but this wallet is the offline wallet anyway, presumably it won't create txns))

SomberNight:
Implementing [BIP-174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) would help with this, as that partial transaction serialisation format can include the derivation path for the is_mine outputs (https://github.com/achow101/bitcoin/commit/d1957f4a68d3b04cecf3aeb112521195aac17c9c).

Related: https://github.com/spesmilo/electrum/issues/3738#issuecomment-358185706

jonathancross:
BIP-174 support would be great, should there be a feature request created for this?

[...]

Issue #3302, opened 2017-11-14
Offline signing with hardware wallet


badikumar:
Using Electrum 3.0.2

### **Machine Setup**

- I have two machines. One offline and one online. 
- Online machine has a watch-only standard wallet I created with 1 xpub.
- The 1 xpub was generated by a standard wallet on the offline computer using a Ledger device (Nano S).

### **Creating a transaction (online computer)**

1. I create a transaction on the online computer 
2. Copy the transaction onto a file.
3. Transfer this file over to the offline computer

### **Signing a transaction (offline computer)**

1. Load the transaction created from the previous step
2. Click on "Sign"
3. Electrum says "Signing transaction". The prompt to confirm never appears on the Ledger device. After a while, it says "Server not reachable" 

### **Code**

Looking at the code, it looks like the `add_hw_info` in `wallet.py` is being called by `sign_transaction` if `isinstance(k, Hardware_KeyStore)`. That means you **have** to be online to sign a transaction using the Ledger device which defeats the purpose of a cold storage device.

Please advise.

ecdsa:
@btchip could it be related to the websocket thing?

btchip:
no, this shouldn't be used with a Nano S, only with a Nano / HW.1. Not sure what add_hw_info is, doesn't ring a bell

badikumar:
@btchip add_hw_info is in wallet.py code. It looks that code wants to get online before signing

ecdsa:
I see, it needs to request the input transactions in order to verify the fee.
that is not needed anymore for segwit wallets, and we can remove that.
for older wallets we are not going to fix that

badikumar:
What does old wallet mean? Non-segwit or wallets created with older version of Electrum?

badikumar:
I see you've clarified that in your message.

But what I don't understand is why does it require the previous tx if it is hardware keystore and not if it isnt?

badikumar:
What if I have a mult-sig wallet? Will this be solved for that case too?

SomberNight:
I've implemented this for Trezor in https://github.com/spesmilo/electrum/pull/4005.

-----

I have had a look at Ledger as well, but looking at the way how we pass amounts to btchip-python, and then to the ledger fw, they don't seem to be optional for segwit. Or at least it's very unclear how we can avoid needing to pass them.

[...]

PR #5721, opened 2019-10-23
integrate PSBT support natively. WIP

SomberNight:
This PR integrates [PSBT](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) support into the main codebase.

Design decisions taken:
- We completely drop support for the existing custom partial tx format.
  - Users will need to upgrade all their cosigners at the same time.
  - This greatly simplifies the code (and the UI/UX).
- Allow insecure signing of legacy inputs without full previous tx. This retains small QR code sizes (and QR codes being possible in the first place!) for txns with legacy inputs.
  - Note that full previous txns can be rather large, just consider trying to spend from a batched exchange withdrawal.
  - With the current custom partial tx format Electrum has been using for many years, full previous txns were never included. The GUI has been displaying "fee" as "unknown".
  - When importing a PSBT that has a legacy input and no corresponding NON_WITNESS_UTXO, we allow signing (against bip174 spec).
    - The Qt GUI
      - If there is a WITNESS_UTXO, shows the fee along with a warning icon with an explanatory tooltip.
      - If there is no WITNESS_UTXO, shows fee as "unknown".
    - The Kivy GUI shows fee as "unknown".
  - When exporting a PSBT, the spec is followed; except when "showing" a QR code, in which case all NON_WITNESS_UTXOs are converted to WITNESS_UTXOs.
- To try to preserve user privacy in serialized txns, the PSBT_GLOBAL_XPUB fields are rarely populated. Further, in PSBT_IN_BIP32_DERIVATION and PSBT_OUT_BIP32_DERIVATION fields, the derivation prefix (typically the hardened part of the path) is often not included.
  - When exporting a PSBT (clipboard, file, qrcode), there are three options offered (in the Qt GUI).
    - The "main"/typical export option (the only option in Kivy/CLI) does not put any xpubs into the psbt, and does not include derivation prefixes.
    - The "coinjoin" export option strips every GLOBAL field and every per-output field, and also strips the PSBT_IN_BIP32_DERIVATION field from the inputs.
    - The coldcard plugin adds another export option that puts xpubs into the psbt, and includes full derivation paths.
  - At the beginning of the tx signing flow, we put global xpubs into the transaction, and also put full derivation paths for the inputs and outputs where we know them. At the end of the signing flow, the xpubs are removed and the full paths are replaced with derivation suffixes.
    - This is done as hardware wallets need the full paths during signing.
    - Further, some signers, e.g. hardware wallets, might want the global xpubs, e.g. when signing multisig inputs. Signing multisig inputs for stateless signers is not secure without xpubs (specifically the change detection part).
- Related to the previous point, note that in the wallet creation wizard, there has been no option to specify a derivation prefix or a root fingerprint. This is still the case, and dealing with this is now postponed for later. (See https://github.com/spesmilo/electrum/issues/5715) Further note that even if we introduced a way to specify these, it is important not to make it mandatory, as this is actually a privacy leak between cosigners.
  - If we do not have a derivation prefix or root fingerprint for a key, we use the fingerprint of the intermediate bip32 node, and put the derivation suffix as the path. Technically, this is not against the spec, and if another Updater/Signer is smart enough, they can detect what is going on (Electrum of course does detect this).
- Upgrading wallet files. Again, related to the previous point(s), existing wallet files do not store root fingerprints and derivation prefixes.
  - If the depth (of the intermediate xpub) is shallow enough, 0 or 1, such as with Electrum seeds, then we can, just from the xpub, reconstruct the root fp and the der prefix.
  - For deeper depths, the missing fields are explicitly set to None, and the code handles None values for these.
  - For hardware wallets, existing wallets store the derivation prefix (as hww has always needed this) but not the root fingerprint. The root fp is opportunistically populated the next time the user connects their device.
- Encoding.
  - When importing files/text/qrcodes, we accept many possibilities, including raw bytes, hex strings, base64 text irrespective of medium (notably base64 text in a file is accepted, which should probably also be mandated by the spec itself...)
  - When exporting, we respect the spec, and clipboard gets base64 text, while saving to file uses raw bytes.
  - As before, transactions are encoded as base43 when displayed in a QR code. (see https://github.com/spesmilo/electrum/issues/5746)
    - This rule is relaxed on the QR-reader side now: when decoding, accept anything we recognise.
- Input finalizer. This implementation can only [finalize inputs](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer) that are "ismine" to the wallet. Relatedly, transaction size estimation when there are not-yet-finalized not-ismine inputs is not really good.
  - This is because input finalizing and input size estimation relies on the wallet side-effecting tx objects by adding all kinds of data to the inputs (in `add_input_info`).

-----

- [x] figure out something about https://github.com/spesmilo/electrum/issues/5715
- [x] add tests for trustedcoin legacy tx serialization
- [x] re-add CLI command "serialize"
- [ ] ~~some performance issues with large wallets atm~~ (maybe later)
- [x] fix offline signing beyond gap limit
- [x] opportunistically learn root fingerprint of existing hardware keystores
- [x] remove global xpubs and bip32 der prefixes from psbt after signing

closes https://github.com/spesmilo/electrum/issues/4615
closes https://github.com/spesmilo/electrum/issues/4514
closes https://github.com/spesmilo/electrum/pull/5671
closes https://github.com/spesmilo/electrum/pull/4883
closes https://github.com/spesmilo/electrum/pull/5717
closes https://github.com/spesmilo/electrum/issues/4322
closes https://github.com/spesmilo/electrum/issues/3302


SomberNight:
Note: one "feature" of the present implementation is that we can only [finalize inputs](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer) that are "ismine". Input finalizing works through the wallet side-effecting tx objects by adding all kinds of data to inputs (in `add_input_info`).
To be able to finalize not ismine inputs, we would need to hardcode templates in the transaction logic to recognise known script types (p2pkh, p2wpkh, multisig types, etc).
Another thing that currently relies on the wallet side-effecting ismine inputs is input size estimation. We only have "good" estimates for ismine inputs.

I don't believe this is a crucial issue to be fixed, rather a nice-to-have, hence it is not a blocker.

ecdsa:
Thanks a lot for this outstanding work.
I have one more question:
does Coldcard require a global xpub in order to sign non-multisig inputs?

SomberNight:
see https://github.com/spesmilo/electrum/pull/5721/commits/1017fefdc95738aeb99a165e80104a0efe35992a

craigraw:
Picked up a couple of issues when testing:

- Tools -> Load Transaction -> From file does not accept binary PSBT files (trivial fix)
- Error when loading a fully signed (multisig) PSBT: "SerializationError(error('unpack_from requires a buffer of at least 4 bytes'))"

Should the *.txn option be removed from the file selection dialogs?

SomberNight:
@craigraw Thanks for testing!
I've pushed some changes to address your findings:
https://github.com/spesmilo/electrum/commit/85a4811742e641828ee8fa45115e8ef99ca9bdbd
https://github.com/spesmilo/electrum/commit/9ff7d2c5a70298da885e5917b39f79cccb16dbaa
https://github.com/spesmilo/electrum/commit/365aa189f274970ef2f4446f6e89b429e41b4ae5

> Should the *.txn option be removed from the file selection dialogs?

Well, with what file extension do you suggest we save fully signed (non-psbt) transactions?
We have been using .txn in the past, it seemed to make sense to keep using that for these.
Further, these fully signed transactions are saved as hex strings, not raw bytes; just like before.
Hex strings are more human-friendly, both for debugging and for easy copy-pasting, and as they are strings they are even easier to interact with in shell.
We can change this though if you have a better suggestion.
@ecdsa 

> Error when loading a fully signed (multisig) PSBT: "SerializationError(error('unpack_from requires a buffer of at least 4 bytes'))"

I am not fully sure about this; it might be already fixed by the above commits; if not, please give more details or share the psbt.

craigraw:
Thanks, I now understand that a .psbt will be exported for a partially signed transaction, and a .txn for a fully signed one. The file selection dialog had me a little confused - it's now much clearer with your recent commits. 

One note is that you can't currently load a .txn from the Load Transaction -> From file (it fails with the SerializationError I mentioned before) - it does however work using Load Transaction -> From text.

SomberNight:
> One note is that you can't currently load a .txn from the Load Transaction -> From file (it fails with the SerializationError I mentioned before) - it does however work using Load Transaction -> From text.

Thanks again. Should work now with https://github.com/spesmilo/electrum/commit/bf8a58c0b49528573b9651b7a27a4168ec5985f7

craigraw:
Thanks, confirmed fixed.

PR #5742, opened 2019-11-04
plugins/coldcard/qt.py: bugfix for API change on main_window.show_transaction

peter-conalgo:
Fix needed after recent API change in: 5c1340b7bd8bab9a5beb8e15b7b15ea26a334695


SomberNight:
Thanks

PR #5752, opened 2019-11-10
CoinChooser: avoid NotEnoughFunds if zero buckets are sufficient

JeremyRand:
The coin chooser incorrectly raises `NotEnoughFunds` when the following conditions are both true when `make_tx` is called:

* `len(coins) == 0`
* The combined value of `inputs` is sufficient to fund the transaction.

This PR explicitly checks whether an empty bucket is sufficient to fund the transaction, and if so, suppresses the `NotEnoughFunds`.


SomberNight:
Thank you for the report and the code contribution.
I've adapted your change a bit; I think it's more natural to do the check in `bucket_candidates_any`.

PR #5755, opened 2019-11-11
Add @lukechilds servers

lukechilds:
:)


ecdsa:
we do not accept pull requests for default servers.
your server will be added to the list if we find that it is reliable over time.

ecdsa:
@EagleTM

lukechilds:
@ecdsa Ah ok, makes sense.

PR #5758, opened 2019-11-14
Mnemonic performance improvements

mbarkhau:


SomberNight:
Nice!
We should have benchmarked this and noticed :)
Indeed, this method is in a really hot code path.

> list.index(word) is O(n)
dict[word] is O(log(n))

Actually, I think dicts are not implemented as balanced trees but rather as hash maps, so lookups are amortised constant time.

From a few quick tests, it seems `make_seed` became 3x faster.
Thank you.

mbarkhau:
Cheers.

Issue #5750, opened 2019-11-08
Color coding addresses in electrum android 


roth-a:
The color coding of addresses belonging to the wallet (when viewing/signing a transaction) is a very useful feature and I would like to see it on Electrum android too. 

Electrum-android version 3.3.7

roth-a:
@SomberNight can you give me an idea where I would need to change the code?  Then I could start experimenting with electrum myself (for the first time).  Does the component listing the output transactions:  https://github.com/spesmilo/electrum/blob/master/electrum/gui/kivy/main.kv#L110  even support colored entries? Or would the component need to be changed?

SomberNight:
The `OutputList` you link to, contains `OutputItem`s, which contains `TopLabel`s.
I expect you should be able to set `color` and/or `background_color` for these, such as at e.g.
https://github.com/spesmilo/electrum/blob/6e3875ceabf75ee96e722358bcff33f883dddbf5/electrum/gui/kivy/main.kv#L365
See https://kivy.org/doc/stable/api-kivy.uix.label.html

SomberNight:
Note that you will probably want to set the colour dynamically from python code.
An example of modifying a property of a widget from python code is here:
https://github.com/spesmilo/electrum/blob/6e3875ceabf75ee96e722358bcff33f883dddbf5/electrum/gui/kivy/uix/dialogs/label_dialog.py#L53

roth-a:
@SomberNight Thanks! That was very clear and helped a lot.

roth-a:
Have a working implementation: https://github.com/roth-a/electrum/commit/aeb4afed4be3cd8a0a365083feeb3796c4ae832c   (still need to clean up code)

roth-a:
@SomberNight : Is there a central file for kivy color schemes? Or is it all hard-coded?

roth-a:
Added pull request: https://github.com/spesmilo/electrum/pull/5774

roth-a:
Any chance of merging this in soon?

PR #5774, opened 2019-11-17
Color-Code Addresses in Kivy UI

roth-a:
- Created a kivy component BackgroundTopLabel, which allows custom background colors
- Transferred the logic for color coding from the qt version, which gives identical color coding of output addresses in the kivy gui

Closes: #5750

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/__init__.py
+++ b/electrum/gui/kivy/uix/dialogs/__init__.py
@@ -213,16 +213,17 @@ class OutputList(RecycleView):
         res = []
         for o in outputs:
             value = self.app.format_amount_and_units(o.value)
-            res.append({'address': o.get_ui_address_str(), 'value': value, 'background_color': (0.3, 0.3, 0.3, 1), 'color': (1, 1, 1, 1)})
+            res.append({
+                    'address': o.get_ui_address_str(),
+                    'value': value,
+                    'background_color': (0.3, 0.3, 0.3, 1),
+                    'color': (1, 1, 1, 1)})
         self.data = res
 
 
 class TopLabel(Factory.Label):
     pass
 
-
-class TopLabelExtended(TopLabel):
-    pass
 
 
 class RefLabel(TopLabel):
--- a/electrum/gui/kivy/uix/dialogs/__init__.py
+++ b/electrum/gui/kivy/uix/dialogs/__init__.py
@@ -213,7 +213,7 @@ class OutputList(RecycleView):
         res = []
         for o in outputs:
             value = self.app.format_amount_and_units(o.value)
-            res.append({'address': o.get_ui_address_str(), 'value': value, 'background_color': (0.3, 0.3, 0.3, 1)})
+            res.append({'address': o.get_ui_address_str(), 'value': value, 'background_color': (0.3, 0.3, 0.3, 1), 'color': (1, 1, 1, 1)})
         self.data = res
 
 
--- a/electrum/gui/kivy/uix/dialogs/__init__.py
+++ b/electrum/gui/kivy/uix/dialogs/__init__.py
@@ -213,11 +213,15 @@ class OutputList(RecycleView):
         res = []
         for o in outputs:
             value = self.app.format_amount_and_units(o.value)
-            res.append({'address': o.get_ui_address_str(), 'value': value})
+            res.append({'address': o.get_ui_address_str(), 'value': value, 'background_color': (0.3, 0.3, 0.3, 1)})
         self.data = res
 
 
 class TopLabel(Factory.Label):
+    pass
+
+
+class TopLabelExtended(TopLabel):
     pass
 
 
--- a/electrum/gui/kivy/uix/dialogs/__init__.py
+++ b/electrum/gui/kivy/uix/dialogs/__init__.py
@@ -213,11 +213,7 @@ class OutputList(RecycleView):
         res = []
         for o in outputs:
             value = self.app.format_amount_and_units(o.value)
-            res.append({
-                    'address': o.get_ui_address_str(),
-                    'value': value,
-                    'background_color': (0.3, 0.3, 0.3, 1),
-                    'color': (1, 1, 1, 1)})
+            res.append({'address': o.get_ui_address_str(),'value': value})
         self.data = res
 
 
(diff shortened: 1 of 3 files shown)


roth-a:
@ghost43: can you review the pr?

electrum/gui/kivy/uix/dialogs/tx_dialog.py
```diff
@@ -181,6 +182,31 @@ def update(self):
...
+        def text_format(addr):
+            """
+            Chooses the appropriate text color and background color to 
+            mark receiving, change and billing addresses.
+
+            Returns: color, background_color
+            """
+            
+            # modified colors (textcolor, background_color) from electrum/gui/qt/util.py
+            GREEN = ("#000000", "#8af296")
```
bauerj: In order not to duplicate the color definitions, it might be cleaner to:

1. Move the color definitions from `electrum.gui.qt.util` to `electrum.util`
2. Make ColorScheme in `electrum.gui.qt.util` extend the one in `electrum.util`.
3. Reference the definition in `electrum.util` here.

roth-a: I completely agree that it would be cleaner to have a central place for colors. I didn't do it for the following reasons:

1. Since this is my first contribution I am not confident to make changes across multiple guis.
2. The color in https://github.com/spesmilo/electrum/blob/master/electrum/gui/qt/util.py#L766  contains 2 colors. A fg_color and bg_color. However I need to also specify the text_color in 
and https://github.com/spesmilo/electrum/pull/5774/files#diff-773ff6da8745ec9ae8fec1cbf69f196f .  So essentially the information in electrum.gui.qt.util is not enough, and would need to be extended.

electrum/gui/kivy/uix/dialogs/tx_dialog.py
```diff
@@ -181,6 +182,31 @@ def update(self):
...
+
+            colors = DEFAULT
+            if self.wallet.is_mine(addr):
+                colors = YELLOW if self.wallet.is_change(addr) else GREEN
+            elif self.wallet.is_billing_address(addr):
+                colors = BLUE
+            return (get_color_from_hex(color) for color in colors)
+
+        for data in self.ids.output_list.data:
+            data['color'], data['background_color'] = text_format(data['address'])
```
roth-a: It would probably be better to choose a different variable name.

roth-a: Fixed with commit https://github.com/roth-a/electrum/commit/67bab545635798f3f413d1725bcb1d912eb81b61

electrum/gui/kivy/uix/dialogs/tx_dialog.py
bauerj: You are right, I initially assumed that the Qt GUI prints foreground colored text on background color. However, it uses the default text color instead. So you are totally right about point 2.

I also agree with you on point 3 but I have to do some nit-picking here. Now that I look at it, the color-choosing mechanism in https://github.com/spesmilo/electrum/commit/5c2beb4ab5d2ec604b4f69878e6be8a4d91944f9#diff-a36dff186a5eff20d9f84c1572ea0acbR605 is quite hard to read but the colors you are using here are only returned as a background color when using a dark theme (like the one kivy uses). In that case, the default text color is assumed to be white(-ish).

image

In my opinion, while still readable, a black text provides less contrast than a white text on all of these backgrounds. Thus my suggestion would be to use a white text in all cases. What do you think?

roth-a: The display with the color bars is interesting. However the colors seem to be off:
The background green #8af296 is:
Screenshot_20191120_205754
Such that the coloring (electrum testnet, kivy on linux) looks like:
Screenshot_20191120_205953

With white it would look like:
Screenshot_20191120_210143

bauerj: Well, I still think the way to go would be to use the darker version of each color with white text 😊 The Qt UI could be changed accordingly.

But of course both could also be changed later since this might be out of scope of your PR.

Does anyone else have an opinion on this?

roth-a:
I now also build an apk and successfully tested the color coding on Android.

electrum/gui/kivy/uix/dialogs/__init__.py
```diff
@@ -213,13 +213,18 @@ def update(self, outputs: Sequence['TxOutput']):
         res = []
         for o in outputs:
             value = self.app.format_amount_and_units(o.value)
-            res.append({'address': o.get_ui_address_str(), 'value': value})
+            res.append({
+                    'address': o.get_ui_address_str(),
+                    'value': value,
+                    'background_color': (0.3, 0.3, 0.3, 1),
+                    'color': (1, 1, 1, 1)})
```
SomberNight: why are these colors needed to be specified here?
when are they used?

electrum/gui/kivy/uix/dialogs/tx_dialog.py
SomberNight: The darker bg colors and white text might look good too; but this looks good too already. I think it's fine as is.

SomberNight:
Besides that one nit above, it looks good and the change is reasonably simple.
(reviewed at 048285f95e966351a19b51b46864030da21f5465)

electrum/gui/kivy/uix/dialogs/__init__.py
roth-a: Thanks for pointing this out. I couldn't find the definition of the kivy default colors, and therefore had to state them here again. I would like to improve this if I can. Do you know where the kivy default colors are defined?

roth-a: I removed these 2 lines and it still works.  That seems cleaner.

roth-a: @SomberNight : done.

electrum/gui/kivy/uix/dialogs/__init__.py
```diff
@@ -221,5 +221,6 @@ class TopLabel(Factory.Label):
     pass
 
 
+
```
SomberNight: nit again; but could you remove this line :P
as otherwise there are no changes in this file

roth-a: yes. I will.

SomberNight:
Thanks for your work!

roth-a:
> Thanks for your work!

Thanks for your and @bauerj 's review.

PR #5775, opened 2019-11-18
Fix missing wallet argument to _resolver

JeremyRand:
`_mktx` wasn't properly passing the `wallet` argument to `_resolver`; this caused errors when it was called with non-`None` values for `domain_addr`.  This PR fixes that.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -34,6 +34,7 @@ import operator
 import asyncio
 import inspect
 from functools import wraps, partial
+from itertools import repeat
 from decimal import Decimal
 from typing import Optional, TYPE_CHECKING, Dict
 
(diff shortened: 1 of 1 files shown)


SomberNight:
Thanks.

Nit: maybe could use a partial or lambda instead and pass `wallet` that way. `nocheck` could also be passed that way, the fact that we currently side-effect the whole `Commands` object with `self.nocheck = nocheck` is really ugly

JeremyRand:
> Nit: maybe could use a partial or lambda instead and pass `wallet` that way. `nocheck` could also be passed that way, the way that we currently side-effect the whole `Commands` object with `self.nocheck = nocheck` is really ugly

@SomberNight FWIW, my understanding is that the Python docs for `repeat` explicitly say that its intended purpose is for supplying invariant arguments to `map`.  So, hard to get more Pythonic than that.  But sure, if you can find a cleaner approach, by all means go for it.

PR #5787, opened 2019-11-21
Return 401 from RPC server for missing auth

domob1812:
When no (supported) authentication is passed to the JSON-RPC server, return a 401 HTTP error code instead of 403.  This indicates to the client that authentication is required, and also requests that to be sent using the "basic" method.  The previously-returned code 403 is now only returned if authentication is passed but not valid.

There are some JSON-RPC clients out there that only send authentication after a 401 code requested it.  Those fail to connect to the Electrum RPC interface even if the correct password is configured.  Those same clients can e.g. connect to Bitcoin Core successfully, which already implements logic matching this change.

See also https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses.

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -259,6 +259,12 @@ class PayServer(Logger):
 class AuthenticationError(Exception):
     pass
 
+class AuthenticationInvalidOrMissing(AuthenticationError):
+    pass
+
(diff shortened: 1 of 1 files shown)


SomberNight:
Thanks

Issue #4232, opened 2018-04-06
assert spending_tx_hash in self.transactions


SomberNight:
On git master 0180abc1791a169cd6c493a922900b143492bb6e.

I have a wallet file that somehow managed to get into an inconsistent state.
It has remnants of a transaction (A) that conflicts with another (B) I just made.

<details>
 <summary>Log</summary>

```
user@debian:~/wspace/electrum$ python3.6 ./electrum -v --testnet --oneserver
[SimpleConfig] electrum directory /home/user/.electrum/testnet
[get_server] [Errno 111] Connection refused
[Plugins] registering hardware digitalbitbox: ('hardware', 'digitalbitbox', 'Digital Bitbox wallet')
[Plugins] registering hardware keepkey: ('hardware', 'keepkey', 'KeepKey wallet')
[Plugins] registering hardware ledger: ('hardware', 'ledger', 'Ledger wallet')
[Plugins] registering hardware trezor: ('hardware', 'trezor', 'TREZOR wallet')
[Plugins] registering wallet type ('2fa', 'trustedcoin')
[profiler] __init__ 0.0027
cannot connect fork_0_1249088
cannot connect fork_0_1202455
cannot connect fork_0_1227602
cannot connect fork_0_1210726
cannot connect fork_0_1231244
cannot connect fork_0_1251380
cannot connect fork_0_1230884
cannot connect fork_0_1202487
cannot connect fork_0_1253281
cannot connect fork_0_1230299
cannot connect fork_0_1210502
cannot connect fork_0_1251316
cannot connect fork_0_1210720
cannot connect fork_0_1230929
cannot connect fork_0_1210707
cannot connect fork_0_1210467
cannot connect fork_0_1251387
cannot connect fork_0_1196375
cannot connect fork_0_1251366
cannot connect fork_0_1253410
cannot connect fork_0_1251305
cannot connect fork_0_1231249
cannot connect fork_0_1251344
cannot connect fork_0_1230934
[...]

PR #5809, opened 2019-11-28
wallet: better (outgoing) invoice "paid" detection

SomberNight:
- no more passing around "invoice" in GUIs, invoice "paid" detection is now handled by wallet logic
- a tx can now pay for multiple invoices
- an invoice can now be paid by multiple txs (through partial payments)
- new data structure in storage: prevouts_by_scripthash
  - type: scripthash -> set of (outpoint, value)
  - also, storage upgrade to build this for existing wallets

Note that this is regarding "outgoing" invoices;
"incoming" ones are currently called "receive requests" (in the code) and they are handled differently.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -26,7 +26,7 @@ import threading
 import asyncio
 import itertools
 from collections import defaultdict
-from typing import TYPE_CHECKING, Dict, Optional, Set, Tuple, NamedTuple, Sequence
+from typing import TYPE_CHECKING, Dict, Optional, Set, Tuple, NamedTuple, Sequence, List
 
 from . import bitcoin
 from .bitcoin import COINBASE_MATURITY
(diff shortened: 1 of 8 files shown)


electrum/address_synchronizer.py
```diff
@@ -327,6 +329,13 @@ def remove_from_spent_outpoints():
             self.db.remove_txi(tx_hash)
             self.db.remove_txo(tx_hash)
             self.db.remove_tx_fee(tx_hash)
+            self.db.remove_verified_tx(tx_hash)
+            self.unverified_tx.pop(tx_hash, None)
+            if tx:
```
ecdsa: when is tx None?

SomberNight: Hmmm I guess if the wallet is not corrupted then it should never happen. Seems hard to come up with a case where we would try to call `remove_transaction` before `add_transaction` for a given tx.
I had decided to handle the case simply because it is already considered possible inside remove_transaction:
https://github.com/spesmilo/electrum/blob/cfbd83c43230ad6b31e98120daff8336c3ac0b7a/electrum/address_synchronizer.py#L303-L305

Also, there is `remove_local_transactions_we_dont_have`, which if gets called it would certainly trigger this.
Then again, that method should only get called for (partially) corrupted wallet I think:
https://github.com/spesmilo/electrum/commit/1935d7d1654b7b90f35167e342bcfc0732c69bce
https://github.com/spesmilo/electrum/issues/4232

PR #5810, opened 2019-11-28
Readme improvement: Added android-sdk to the apt-get installation instruction 

roth-a:
When following the Readme instruction I noticed that the android-sdk is necessary. Adding this to the Readme could make it easier for other readers.


SomberNight:
For which step is it needed?
Or do you mean the `adb` invocations in the FAQ section?

roth-a:
If remember correctly it was necessary for a successful dockers build of electrum. But I could test that again.

SomberNight:
Well that sounds weird as the build happens inside the container, so whether something is installed outside it should be irrelevant.

roth-a:
You are right. It is not needed. Sorry for the confusion.

PR #5811, opened 2019-11-28
add SW wallets qt console password lock

zebra-lucky:
It is preferable to lock console on SW wallets with keystore encryption


SomberNight:
Why do you think this is needed?

zebra-lucky:
If wallet leaved open, someone else can get aceess to sensitive info.

SomberNight:
Such as? Ciphertext of seed words? Some examples might be nice.
Much privacy compromising info is already accessible in such a scenario without using the console; such as your xpub.

zebra-lucky:
GC sometime can have xprv.

zebra-lucky:
I'm sorry, it was possible false about xpriv, trying to find again and noting was found.
But in garbage collector there is huge amount of data, maybe somethen else can be found.

ecdsa:
this is the wrong approach.
if private keys remain in memory after they have been used, then we should clear them.
building a fence around the python console is not going to make things safer

zebra-lucky:
About a false positive: I'm switching password on-off and make a dump in a file.
Now seems that xprv was from a paswordless variant, so xprv does not remains in gc.
Maybe some data from qt widgets can be found, not sure about it.

zebra-lucky:
found a bug, when console is not show in menu, widget is displayes, will add fix today

SomberNight:
If there is secret material in memory left uncleared, that is e.g. accessible from the console, we should try our best to fix that and clear it, as said already in the thread.

The solution proposed here feels hackish. Not in implementation but in idea. Putting a Qt widget on top of the console only protects against a very limited set of attack vectors, namely the scenario where the wallet owner left Electrum running with a wallet open, they left the OS unlocked, and an attacker gains complete control of the OS (by e.g. physically controlling the mouse/keyboard).

If someone physically accesses your computer, they might just dump the memory of the python process.

PR #5820, opened 2019-12-05
ECDSA signatures: grind low R to match with Bitcoin Core (take 2)

SomberNight:
Supersedes and closes https://github.com/spesmilo/electrum/pull/4655

This matches with what Bitcoin Core implements: https://github.com/bitcoin/bitcoin/pull/13666


SomberNight:
Note: somewhat annoyingly, [Bitcoin Core does not grind R when signing messages](https://github.com/bitcoin/bitcoin/blob/d8a66626d63135fd245d5afc524b88b9a94d208b/src/key.cpp#L252); while we do with this change (as we would always grind)

SomberNight:
After some consideration, I think it has no significance that Bitcoin Core does not grind R when signing messages.
It's good to try to imitate Bitcoin Core regarding on-chain behaviour; but not much benefit re message-signing.

It's not even just message-signing vs transaction-signing. E.g. we also sign data for LN gossip messages (e.g. channel updates). I would rather not draw an arbitrary line and say when we grind R and we don't. It is simply cleaner to do it all the time.

SomberNight:
Note: as a sanity check, I've tested signing a few PSBTs with both Electrum and Bitcoin Core 0.19, and they generated identical signatures as expected.

JeremyRand:
> Note: somewhat annoyingly, [Bitcoin Core does not grind R when signing messages](https://github.com/bitcoin/bitcoin/blob/d8a66626d63135fd245d5afc524b88b9a94d208b/src/key.cpp#L252); while we do with this change (as we would always grind)

@SomberNight Did the Bitcoin Core devs provide any justification for this decision (if it was intentional)?

SomberNight:
@JeremyRand I could not find one.

Note that grinding for low R makes the sig generation take 2x more time on average,
and the motivation for having a low R is to save 1 byte when encoding the signature in DER.

1. signatures for on-chain transactions (atm) are all DER encoded -- they will not be in witness v1 txns in the future (which use Schnorr signatures (BIP-340))
2. signatures for text messages use "compact sigs" so no savings there
3. signatures for lightning gossip messages use "compact sigs" so no savings there

- Bitcoin Core implements (1) and (2), and [only grinds for (1)](https://github.com/bitcoin/bitcoin/blob/d8a66626d63135fd245d5afc524b88b9a94d208b/src/key.cpp#L252).
- Electrum implements (1) and (2) and (3), and grinds for all.
- c-lightning implements (1) and (3), and [grinds for all](https://github.com/ElementsProject/lightning/pull/3220).

If one is really pedantic, it only really makes sense to grind for low R for (1), and only when those sigs will be DER encoded. OTOH I think it's nicer for the signing code not to care about what the signature will be used for.

JeremyRand:
@SomberNight Thanks for explaining.  Prior to this PR, the message signatures in the unit tests could be reproduced in Bitcoin Core.  If I understand correctly, this is no longer the case.  How can the message signatures in the tests be reproduced now?

JeremyRand:
@SomberNight It just seems weird to use signatures produced by Electrum in the tests for Electrum.  Sort of a circular correctness proof.  It makes more sense to me to produce the signatures in Bitcoin Core.  I suppose I could just produce signatures in Electrum and verify them in Bitcoin Core to make sure that they're valid without a circular correctness proof; is that the best available option?

Issue #5710, opened 2019-10-16
Receive tab needs to display only the address as well, not just the payment request URI


gits7r:
Currently, the withdraw-to-payment request is not implemented on most of the services, particularly exchange services that allow you to buy BTC and withdraw to your wallet. None of the biggest ones support withdrawing to payment request URI, instead, they only ask for the address. Same as for multiple services that offer bitcoin related services. All ask for a plain and simple 'receiving address' (if it's bech32 or not this is a different subject, but still they require just the address, not the payment request URI).

Users are used to "handing out a bitcoin address" in order to receive bitcoins from a friend or withdraw bitcoin from an exchange service, etc. It's not ideal to have them select just the address from the payment request URI and copy it because it's annoying and opens the door to mistakes.

Another downside is that our redesigned receive tab changed the long time behavior we had in previous versions where we would display the address in its own text field. On latest git master I can see there is a box in the right side of the receive tab where we can toggle from QR to Text, but it also displays the payment request URI instead of the plain and simple sole address.

I propose we should add 2 more tabs before and have a total of 4 possible representations:

1) Variant 1
a) Tab 1: Only Address (text), label = Address
b) Tab 2: Only Address (QR Code), label = Address (QR)
c) Tab 3: Payment request URI (text), label = Payment Request
d) Tab 4: Payment request URI (QR Code), label = Payment Request (QR)
The Tab displayed by default when is Tab 1: Address.

or
2) Variant 2
We add a right - left slider button with two positions: left and right, and describe left as "Address" and right as "Payment request URI". Default position is left (Address). Depending on each position of this slider, the tabs can be:

2.1 Slider set to Address (left) [DEFAULT POSITION]
a) Tab 1: Only Address (text), label = Address
b) Tab 2: Only Address (QR Code), label = Address (QR)

2.2 Slider set to Payment Request URI (right)
a) Tab 1: Payment request URI (text), label = Payment Request
b) Tab 2: Payment request URI (QR Code), label = Payment Request (QR)

Feel free to vote whichever you think will provide the best user experience. I am for Variant 1, so users have at hand the address if they need just the address as well as the URI if they need it. The right click menu from the bottom of the payment request window also includes a copy address link which is great and should stay there, so I think this should be sufficient.

SomberNight:
- This also affects the kivy GUI. It also currently displays bip21 URIs and QR codes. Imagine going to a bitcoin ATM and it rejecting your QR codes...
[...]

PR #5822, opened 2019-12-07
qt receive tab: show plain bitcoin address

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/5710

pic3

<details>
 <summary>More pics</summary>

pic1
pic2
pic4
pic5

</details>

@gits7r @ecdsa

### DIFF
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -227,7 +227,7 @@ class AddressList(MyTreeView):
             menu.addAction(_('Details'), lambda: self.parent.show_address(addr))
             persistent = QPersistentModelIndex(addr_idx)
             menu.addAction(_("Edit {}").format(addr_column_title), lambda p=persistent: self.edit(QModelIndex(p)))
-            menu.addAction(_("Request payment"), lambda: self.parent.receive_at(addr))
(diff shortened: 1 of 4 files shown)


SomberNight:
There is a lot of unused space on the left :/
Maybe the "incoming payments" list could occupy that; it could take up less X-axis space but more Y-axis space, if we force it on the left side of the QR code/receive address.

gits7r:
First or second one look good to me.
In the QR image box, by default the QR code representing the full payment request is shown, correct? And a second 'QR' button below the raw receiving address field needs to be pushed in order to QR display just the address, right? And will that be a pop-up or change the existing QR image above?

SomberNight:
> In the QR image box, by default the QR code representing the full payment request is shown, correct? 

Yes, the QR encodes the full BIP21 payment request.

> And a second 'QR' button below the raw receiving address field needs to be pushed in order to QR display just the address, right?

Yes

> And will that be a pop-up or change the existing QR image above?

It is a popup.

However, do note that I have checked the behaviour on 3.3.8 and the QR code in the Receive tab is encoding the full payment request there too. I think this has been the behaviour for a long time.

Issue #3531, opened 2017-12-15
Improvement request: Improve address filter + Add Address button in send


kristian:
As an improvment of the "Spend from" functionallity in the Addresses / Coins tab, I would love to see a "Spend From Multiple Addresses" functionallity in the Send tab, to select more than one addresses to send from. It would be way more convinient if I could add a address in the "Send" UI directly, instead of juming back and forth.

Additionally a "Funded or Unused" Filter would be exactly what I would like to see in the Addresses tab, these are the addresses I am interested the most (because I will either send from them, or receive coins on them).

SomberNight:
> I am currently unable to spend from two addresses, in case one is a Receiving and one is a Change address

This point is a duplicate of https://github.com/spesmilo/electrum/issues/3356
As mentioned there, you can choose from the `Coins` tab.

kristian:
@SomberNight thanks for pointing it out. I adapted my original feature request.

ecdsa:
looks like the current GUI has those features

PR #5823, opened 2019-12-07
Qt address list: add "funded or unused" filter option

Chris-777:
I've noticed that old used addresses can make an address list unwieldy over time. I would like a filter for "Funded & Unused" that filters out old zero balance addresses.

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/addresses.py
+++ b/electrum/gui/kivy/uix/dialogs/addresses.py
@@ -102,9 +102,9 @@ Builder.load_string('''
                 spacing: '5dp'
                 AddressButton:
                     id: search
-                    text: {0:_('All'), 1:_('Unused'), 2:_('Funded'), 3:_('Used')}[root.show_used]
+                    text: {0:_('All'), 1:_('Unused'), 2:_('Funded'), 3:_('Used'), 4:(_('Funded')+'|'+_('Unused'))}[root.show_used]
                     on_release:
-                        root.show_used = (root.show_used + 1) % 4
+                        root.show_used = (root.show_used + 1) % 5
                         Clock.schedule_once(lambda dt: root.update())
             AddressFilter:
                 opacity: 1
@@ -270,6 +270,8 @@ class AddressesDialog(Factory.Popup):
             if self.show_used == 2 and balance == 0:
                 continue
             if self.show_used == 3 and not is_used_and_empty:
+                continue
+            if self.show_used == 4 and is_used_and_empty:
                 continue
             card = self.get_card(address, balance, is_used_and_empty, label)
             if search and not self.ext_search(card, search):
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -43,6 +43,7 @@ class AddressUsageStateFilter(IntEnum):
     UNUSED = 1
     FUNDED = 2
     USED_AND_EMPTY = 3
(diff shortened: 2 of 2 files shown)


SomberNight:
Maybe "funded or unused" would be a more intuitive name?
It's a set theoretic OR you are doing here; and especially in the code, as there is already a "used and empty", which is a set theoretic AND, I find your naming a bit confusing.

Chris-777:
The filter is effectively "Display if NOT is_used_and_empty" correct? The result being that "Funded" AND "Unused" addresses are both displayed. If we said OR then it would surely imply showing "Funded" OR "Unused" but not both at the same time?

SomberNight:
> The filter is effectively "Display if NOT is_used_and_empty" correct?

Yes but that is not equivalent to `not used and not empty` (which you suggest when you say it is "unused and funded").

`not is_used_and_empty` = `not (used and empty)` = `not used or not empty` = `unused or funded`

> The result being that "Funded" AND "Unused" addresses are both displayed. 

All the addresses in the "Funded" set and all the addresses in the "Unused" set are displayed, yes, but this means that the correct operation is "OR".

Disregarding the order of addresses for now, these are sets.
The set of "funded" addresses, the set of "used" addresses, etc.
If you consider them as sets, then "and" means "intersection"; and "or" means "union".

This is also how `USED_AND_EMPTY` is interpreted in the code: intersection of the "used" set and the "empty" set.

See these Venn diagrams:
ven gif

Chris-777:
Thanks for taking the time to thoroughly clarify for me. You are of course perfectly correct regarding the set theory.

How would you suggest we label the drop down item? Thinking of it from the perspective of a user (who is presumably unaware of the actual logical operation) this is simply a concatenation of two separate options already in the drop down menu.

> Maybe "funded or unused" would be a more intuitive name?
I could could probably get used to it but or still seems very counter intuitive to me.

SomberNight:
> How would you suggest we label the drop down item? Thinking of it from the perspective of a user (who is presumably unaware of the actual logical operation) this is simply a concatenation of two separate options already in the drop down menu.

I think everyday written language might use "and/or" in this case; though not sure if I like that.
I guess we could keep what you already have.

Chris-777:
Would 'Funded + Unused' be better?

ecdsa:
I too believe that "or" would be clearer than "and". 

I am not sure if we want to go down the route of adding a filter for each possible combination of the basic flags. What is the use case for that?

SomberNight:
> I am not sure if we want to go down the route of adding a filter for each possible combination of the basic flags. What is the use case for that?

This particular combination of "funded or unused" I believe is actually useful, as these are the addresses power users are typically interested in. You either want to spend your UTXOs or copy and give out an unused address to receive.

It has also been requested multiple times AFAIR, although I could only find one example on github: https://github.com/spesmilo/electrum/issues/3531#issue-282533934

Chris-777:
> these are the addresses power users are typically interested in

This has been my experience with using the address list. Used and empty addresses are of very little interest most of the time, currently the only way to remove this ever growing list is by using either the Funded or Unused filters.

This new filter creates a neat summary of the addresses I'm most interested in. It would be my personal choice for the default view of the address list. It also works for new wallets and discourages address reuse.

An alternative naming suggestion: 'Hide Used'.

Chris-777:
Is there anything preventing this request from being merged?

DvalCaruso (approved):

PR #5832, opened 2019-12-10
wizard: (qt) add dedicated button to create new wallet

SomberNight:
This PR is a revival of https://github.com/spesmilo/electrum/pull/3914
_(Had to create a new PR: I force-pushed the old one while it was closed and now it can no longer be re-opened.)_

The underlying issue keeps coming back; we should fix this.


ecdsa:
merged manually

PR #5843, opened 2019-12-16
notify when local tx is in address history

zebra-lucky:
When local transaction is broadcasted and received in address history notify ~with network "new_transaction" notification~ notify by some means.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -304,8 +304,11 @@ class AddressSynchronizer(Logger):
             # add to local history
             self._add_tx_to_local_history(tx_hash)
             # save
+            is_new_tx = (tx_hash not in self.db.transactions)
             self.db.add_transaction(tx_hash, tx)
             self.db.add_num_inputs_to_tx(tx_hash, len(tx.inputs()))
+            if is_new_tx and not self.is_local_tx(tx_hash) and self.network:
+                self.network.trigger_callback('new_transaction', self, tx)
             return True
 
     def remove_transaction(self, tx_hash: str) -> None:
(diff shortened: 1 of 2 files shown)


zebra-lucky:
Seen tests failed, tested on 3.3.8, will redo now for master

zebra-lucky:
Maybe "new_transaction" is not best notification for the situation and some other notification should be sent. Posslible variatn "tx_broadcasted"/"tx_local" pair. If wallet switches from/to some lagging server, then possible switching from/to local tx state.

zebra-lucky:
I'll try to modify code now, to move `new_transaction` in `wallet.add_transaction` method and add two more notifications `tx_on_history`, `tx_off_history`.

zebra-lucky:
Added modified code.

Two new notifications `tx_on_history`, `tx_off_history` added.
And `new_transaction` notification sent from `wallet.add_transaction`.
More code changes, but seem more consistent.

I'm sorry for the first variant, it was quick try to fix.

SomberNight:
What are you trying to achieve specifically?
Are you trying to fix the scenario where
- user has local tx A in history
- server sends notification re history update for some address -> client downloads new history -> tx A is in new history -> client sees it already has tx A
- there is no OS notification about tx A

Is it only the lack of OS notification that you are trying to fix?

-----

Note that, e.g. during an initial history sync, `wallet.add_transaction` is often called many times for the same tx, so with your current changeset it will schedule multiple OS notifications for some txs.

zebra-lucky:
Not only os notifcation.

On dash fork I try implement some automatic transactions processing, and need more info on tx state. Local transactions is created, and then no additional means to detect when transaction state changes, only verified notification. So I think this code can add more consistency to notifications and send to upstream.

> Note that, e.g. during an initial history sync, wallet.add_transaction is often called many times for the same tx, so with your current changeset it will schedule multiple OS notifications for some txs.

in add_transaction `new_transaction` notification is sent olny when tx is not found in db.transactions (`is_new_tx` flag).

zebra-lucky:
Added small fix to order of tx_on_history notifications

zebra-lucky:
Seems found some bugs in notifications, need to test again tomorrow.
Reproduced when pay many small inputs to other open wallet window.

zebra-lucky:
Not reproduced again, I'll do more tests.

zebra-lucky:
I'm sorry for a delay.

After some time I was realized that `tx_on_history`, `tx_off_history` is the wrong approach, as history is related to addresses. I do rewrite of code to add `new_transaction` notification when local tx appears on history and completely removed `tx_on_history`, `tx_off_history` code.

PR #5850, opened 2019-12-19
kivy: allow generic passwords for wallets

SomberNight:
<details>
 <summary>pics</summary>

pic1
pic2
^ This is on desktop Linux; on Android the system keyboard would also show up "over" the bottom
pic3
^ This is on desktop Linux; on Android the system keyboard would also show up "over" the bottom
</details>


relativisticelectron:
After testing it seems that it does not remember that a generic password was used for this password.  This means that one needs to click "generic password" each time you open the wallet.

ecdsa:
@relativisticelectron this PR is not going to be merged as is, but some of its code will be used.
We need the capability to save a backup of the wallet, for lightning. Since the backup will not be in internal storage, it requires strong encryption. Wallets that are in internal storage should keep using a PINI code.

ecdsa:
deprecated by #5951

Issue #5878, opened 2020-01-08
Make the copy outpoint feature relevant again


bilthon:
This is actually something that was present in v3.3.8 of Electrum, but which I can see has changed on 4.0.0. The feature itself has to do with the copy to clipboard of outpoint data in the "coins" section.

UTXOs in the "coins" section are displayed in an abbreviated form, for instance:

> 9269f2c6a8c04bb5...:1

And that's still true for 4.0.0 and I agree that it is the right thing to do. However it used to be that when one right-clicked a specific outpoint and selected the "copy" option, the full data "txid:index" (without the shortening & ellipsis) ended up in the clipboard. What I see now after compiling 4.0.0 (at commit 29cf01524a286b3a6c884a6a05df12a8b759c12f) is that the clipboard data is also the abbreviated version.

I am experimenting with a side-project that would greatly benefit if the actual data being copied to clipboard is the full outpoint and not the abbreviated version, which is kind of useless if you come to think about it. The reasoning here is that if you went ahead and selected a specific outpoint and "copied" it to clipboard, you're likely interested in its full content, not just the abbreviated version.

I'm not that familiar with the codebase, but was able to narrow down the file at `gui/qt/utxo_list.py` as the potential modification target.

SomberNight:
done in https://github.com/spesmilo/electrum/pull/5879

PR #5879, opened 2020-01-08
Restoring old behavior of the outpoint copy to clipboard feature

bilthon:
This fixes the issue described at #5878

### DIFF
--- a/electrum/gui/qt/util.py
+++ b/electrum/gui/qt/util.py
@@ -480,6 +480,7 @@ class ElectrumItemDelegate(QStyledItemDelegate):
         return super().createEditor(parent, option, idx)
 
 class MyTreeView(QTreeView):
+    ROLE_CLIPBOARD_DATA = Qt.UserRole + 100
 
     def __init__(self, parent: 'ElectrumWindow', create_menu, *,
                  stretch_column=None, editable_columns=None):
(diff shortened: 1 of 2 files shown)


electrum/gui/qt/util.py
```diff
@@ -653,7 +653,11 @@ def add_copy_menu(self, menu, idx):
         for column in self.Columns:
             column_title = self.model().horizontalHeaderItem(column).text()
             item_col = self.model().itemFromIndex(idx.sibling(idx.row(), column))
-            column_data = item_col.text().strip()
+            data = item_col.data()
```
SomberNight: The name `data` is not very descriptive, I suggest renaming to `clipboard_data`.
I would rather not use `QStandardItem.setData()` and `QStandardItem.data()` without specifying the `role` argument.
At the top of this class, `MyTreeView`, let's define an attribute (of the class), e.g.:
`ROLE_CLIPBOARD_DATA = Qt.UserRole + 100`
Then, we could have:
```suggestion
            clipboard_data = item_col.data(self.ROLE_CLIPBOARD_DATA)
```

electrum/gui/qt/utxo_list.py
```diff
@@ -96,6 +96,8 @@ def insert_utxo(self, idx, utxo: PartialTxInput):
         amount = self.parent.format_amount(utxo.value_sats(), whitespaces=True)
         labels = [name_short, address, label, amount, '%d'%height]
         utxo_item = [QStandardItem(x) for x in labels]
+        # Storing the full outpoint as the data payload of this item
+        utxo_item[0].setData(name)
```
SomberNight: There are already a few similar lines below, in particular `utxo_item[self.Columns.ADDRESS].setData(name, Qt.UserRole)`.
Let's move this below that one.
Also, like those lines below, please do not use a magic number here (`0`).
Along with the suggestion for the other file, we could have:
```suggestion
        utxo_item[self.Columns.OUTPOINT].setData(name, self.ROLE_CLIPBOARD_DATA)
```

electrum/gui/qt/util.py
bilthon: Makes sense, done.

electrum/gui/qt/utxo_list.py
bilthon: Also done.

SomberNight:
Thanks

PR #5890, opened 2020-01-13
Electrum Kivy implementation

radha-cis:
> Electrum Kivy implementation

1. Enable Hardware device option for the Android app.
2. Get Hardware wallet device detection and move to the select a device screen.
3. By selecting the hardware device, move forward for the type of addresses screen for which we want to create a wallet.


ecdsa:
can you describe how this relates/differs from #5880 ?

PeterSurda:
@ecdsa sorry, as I said on IRC, they are different stages and a WIP (not all functionality is implemented yet, need to be cleaned up, only tested with Ledger Nano S, etc). They probably can be both closed after giving feedback.

PR #5898, opened 2020-01-21
Add command line option to forget config on exit.

leo-lb:
By default, Electrum saves the last opened wallet's path as well as
recently opened wallets.

This can be damaging to plausible deniability.

Now it's possible to run Electrum with `--forgetconfig` to not
write to the config at all, which includes the wallet paths.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1171,6 +1171,7 @@ def add_global_options(parser):
 
 def add_wallet_option(parser):
     parser.add_argument("-w", "--wallet", dest="wallet_path", help="wallet path")
+    parser.add_argument("--forgetconfig", action="store_true", dest="forget_config", default=False, help="Forget config on exit")
 
 def get_parser():
     # create main parser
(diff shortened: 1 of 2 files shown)


electrum/commands.py
```diff
@@ -1171,6 +1171,7 @@ def add_global_options(parser):
 
 def add_wallet_option(parser):
     parser.add_argument("-w", "--wallet", dest="wallet_path", help="wallet path")
+    parser.add_argument("-f", "--forgetconfig", action="store_true", dest="forget_config", default=False, help="Forget config on exit")
```
SomberNight: `-f` is already taken to specify the `fee` arg for relevant commands
Especially for "niche" options, there does not need to be a single char shortcut at all.

leo-lb: Right. Removed and pushed.

ecdsa:
We already have another way to not write options to config: passing them on the command line.
The drawback of the existing method is not universal, ie it does not allow to set any option, only a few of them. However, before adding a new method, one should try to not create redundant things.

leo-lb:
@ecdsa

No, this is not what this change helps with. Electrum, by default, and you cannot disable that, writes the last opened wallet path and recent wallet paths to the `config` file. This helps by telling it not to do that at all, as well as not writing anything to the `config` file at all, for plausible deniability purposes.

Issue #5924, opened 2020-01-31
offline usage: PARTIAL_TXN_HEADER_MAGIC causes older versions of electrum to vomit 


electrum-user-account:
https://www.reddit.com/r/Electrum/comments/ewij87/cold_storage_compatibility_issues/

Hello, 

As I wrote in my reddit post, I'm trying to use a new version of electrum as an online watch-only wallet with an old version of electrum as an offline signatory. 

Is this a supported configuration?  It's looking like its not really supported, because it doesn't work.

Basically, when the user generates a transaction from the watch-only  wallet in the new version of electrum 3.3.8, this code executes:  https://github.com/spesmilo/electrum/blob/e81f4bdcd11a072e7c4f38fb1c7eec19c2f7e1a8/electrum/transaction.py#L1009

```
    def serialize(self, estimate_size=False, witness=True):
        network_ser = self.serialize_to_network(estimate_size, witness)
        if estimate_size:
            return network_ser
        if self.is_partial_originally and not self.is_complete():
            partial_format_version = '00'
            return bh2u(PARTIAL_TXN_HEADER_MAGIC) + partial_format_version + network_ser
        else:
            return network_ser
```

And it falls through to the line  `return bh2u(PARTIAL_TXN_HEADER_MAGIC) + partial_format_version + network_ser`

On the old version of electrum, when the transaction file is loaded, this code executes:

https://github.com/spesmilo/electrum/blob/0d4de870a5159d491b604d29780f6bfb539c49f3/lib/transaction.py#L447

```
def deserialize(raw):
    vds = BCDataStream()
    vds.write(raw.decode('hex'))
    d = {}
    start = vds.read_cursor
    d['version'] = vds.read_int32()
    n_vin = vds.read_compact_size()
    d['inputs'] = list(parse_input(vds) for i in xrange(n_vin))
    n_vout = vds.read_compact_size()
    d['outputs'] = list(parse_output(vds,i) for i in xrange(n_vout))
    d['lockTime'] = vds.read_uint32()
    return d
```

The value of `n_vin` will be read from somewhere inside that `PARTIAL_TXN_HEADER_MAGIC` string and result in `n_vin = int(staggeringly large trillzzillion)`, and the parsing just gets worse and blows up from there, resulting in buffer overflow. 

I was just poking around today and I found out that if I modify the electrum code on my watching-only wallet to return `return network_ser` instead of the prefixed string, it works and I can open the transaction file in the old electrum. 

[...]

PR #5926, opened 2020-01-31
Display exported txn compatibility information to the user

electrum-user-account:
txdialog

The serialized transaction format was changed here #4405

The change occurred in Electrum 3.2.0.

(
  before: https://github.com/spesmilo/electrum/blob/624fa4769d9f02fa71696b3d69030ae114ccf573/lib/transaction.py#L891
  after: https://github.com/spesmilo/electrum/blob/0119ab9ee1c4b07255442c7c91a7119ca989b64c/lib/transaction.py#L1013
)

I ran into this myself, here: #5924

Personally I believe that users should be notified of breaking changes.  Users should not have to make educated guesses, ask online, or be python programmers in order to be able to tell whats going on / why doesn't it work.

If you don't like this verbiage (for example, if you think that the user should be instructed to always use the exact same electrum version when importing / exporting)  

Then an alternative verbiage would be appropriate, such as "Always use the exact same Electrum version to import any exported transactions."

If that doesn't fit there the label can be moved. Or it could be  `'Electrum ' + current_electrum_version + ' required to import.'` Or Something. Point is that there should be some indication of this fact of life to the user.

### DIFF
--- a/electrum/gui/qt/transaction_dialog.py
+++ b/electrum/gui/qt/transaction_dialog.py
@@ -183,6 +183,7 @@ class BaseTxDialog(QDialog, MessageBoxMixin):
             vbox.addWidget(self.feecontrol_fields)
         self.hbox = hbox = QHBoxLayout()
         hbox.addLayout(Buttons(*self.sharing_buttons))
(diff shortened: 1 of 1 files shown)


ecdsa:
these are all valid points, but I do not think that your pull request is useful

PR #5951, opened 2020-02-12
Ln backups

ecdsa:


electrum/gui/kivy/main_window.py
```diff
@@ -1200,6 +1200,21 @@ def on_success(old_password, new_password):
...
         self._password_dialog.open()
 
+    def save_backup(self):
+        from .uix.dialogs.password_dialog import PasswordDialog
+        from electrum.util import get_backup_dir
+        if self._password_dialog is None:
+            self._password_dialog = PasswordDialog()
+        message = _("Create backup.") + '\n' + _("Enter your current PIN:")
+        def on_success(old_password, new_password):
+            new_path = os.path.join(get_backup_dir(self.electrum_config), self.wallet.basename() + '.backup')
```
SomberNight: Maybe we could put a timestamp in the name?
To avoid backups overwriting each other. Or is it a feature that the new one overwrites the old one?

electrum/gui/qt/channels_list.py
```diff
@@ -84,10 +84,14 @@ def task():
         WaitingDialog(self, 'please wait..', task, self.on_success, self.on_failure)
 
     def force_close(self, channel_id):
-        def task():
-            coro = self.lnworker.force_close_channel(channel_id)
-            return self.network.run_from_another_thread(coro)
-        if self.parent.question('Force-close channel?\nReclaimed funds will not be immediately available.'):
+        if self.lnworker.wallet.is_lightning_backup():
+            msg = _('<b>WARNING<b>: force-closing from an old state might result in fund loss.\nAre you sure?')
```
SomberNight: </b> to close the tag
Also, `self.parent.question` needs to be passed `rich_text=True` I guess?
and also `\n` might not work in rich text; not sure.

electrum/wallet.py
```diff
@@ -263,6 +263,17 @@ def save_db(self):
         if self.storage:
             self.db.write(self.storage)
 
+    def save_backup(self, path, *, old_password=None, new_password=None):
+        new_db = WalletDB(self.storage.read(), manual_upgrades=False)
```
SomberNight: could `self.storage.read()` return an older state of the db than `db` (which is current)?
does it not simply return the state when the storage object was created (which might be when the wallet was opened)?

SomberNight:
Also note: we should dynamically ask for the WRITE_EXTERNAL_STORAGE permission.

electrum/wallet.py
ecdsa: right, thats a very good point.

electrum/gui/kivy/main_window.py
ecdsa: better overwrite it. in fact, I think we should automate this, as in setup a backup path only once, and save it on each new channel creation.

electrum/wallet.py
```diff
@@ -263,6 +263,28 @@ def save_db(self):
...
+            return
+        new_storage = WalletStorage(new_path)
+        if pin_code:
+            backup_pubkey = self.config.get('backup_pubkey')
+            if backup_pubkey is None:
+                return
+            w2 = Wallet(new_db, None, config=self.config)
+            w2.update_password(pin_code, None)
+            new_storage._encryption_version = StorageEncryptionVersion.USER_PASSWORD
+            new_storage.pubkey = backup_pubkey
```
SomberNight: As discussed out of band, the wallets created this way will have storage-encryption but not keystore-encryption. `update_password` for these will not work

ecdsa:
per today's discussion:
 - use a strong password for each wallet. the password is requested to open the wallet, and retained in memory
 - use a shared PIN (6 digits) to unlock the wallet. the PIN is saved in config

PR #5991, opened 2020-02-25
fix BCDataStream.read_bytes

zebra-lucky:
- fix BCDataStream.read_bytes

### DIFF
--- a/electrum/tests/test_transaction.py
+++ b/electrum/tests/test_transaction.py
@@ -59,7 +59,7 @@ class TestBCDataStream(ElectrumTestCase):
         self.assertEqual(s.read_bytes(3), b'foo')
         self.assertEqual(s.read_bytes(2), b'ba')
         with self.assertRaises(transaction.SerializationError):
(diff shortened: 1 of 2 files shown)


electrum/tests/test_transaction.py
```diff
@@ -58,8 +58,12 @@ def test_bytes(self):
         s.write(b'foobar')
         self.assertEqual(s.read_bytes(3), b'foo')
         self.assertEqual(s.read_bytes(2), b'ba')
-        self.assertEqual(s.read_bytes(4), b'r')
-        self.assertEqual(s.read_bytes(1), b'')
+        with self.assertRaises(transaction.SerializationError):
+            self.assertEqual(s.read_bytes(4), b'r')
```
SomberNight: The `assertEqual` does not really make sense here.
```suggestion
            s.read_bytes(4)
```

zebra-lucky: fixed in 3cc5756

PR #5992, opened 2020-02-25
fix SPV.hash_merkle_root, add tests/test_verifier.py

zebra-lucky:
- fix SPV.hash_merkle_root, add tests/test_verifier.py


SomberNight:
It's unfortunate that this has not been noticed until now.
Thanks for the fix.

related https://github.com/spesmilo/electrum/pull/4436

Issue #5561, opened 2019-08-14
Password for watching wallets


KenKundert:
I currently use Electrum as a front-end for my hardware wallet. Normally my hardware is locked up safe at home, which prevents me from sending my currency when I am out and about, but I can still monitor my wallet contents and even receive currency while on the road.  That is great; just what I want.

However, when using Electrum in this way the wallet is not encrypted. Anybody that gained access to my machine could open the wallet and see how much currency I owned. That seems bad. One of the cardinal rules of bitcoin is 'never tell anybody how much you have'.

I looked into encrypting the wallet, but for hardware wallets it seems that after encrypting the wallet I would only be able to open it if I had the hardware wallet in my possession. That breaks the use-model.

I would like to request that you add the option of having a password that encrypts the whole account, but in such a way that the account could be decrypted and viewed without requiring access to the hardware wallet itself.  Actually, it would be nice to have the option of having a password on any watching wallets (perhaps that is already possible, but I could not find evidence of that in a quick check of the documentation).

-Ken

ecdsa:
you can already create a watching-only wallet that matches your hardware wallet, and password-protect it. adding a password to a wallet file tied to a hardware wallet cannot be achieved unless we also ask the user to enter their password when they have the hardware wallet.

KenKundert:
Thanks for your help. For what is it worth, that it the user experience I desire. I rarely connect my hardware wallet and so the extra burden is mimimal, and it provides a boost in security.

So now it seems like my choices are:

1. move to a watching wallet and do not use Electrum with my hardware wallet, or
2. use wallets, one the existing hardware wallet with a password added, and the second a password protected watching wallet.

gdfy:
I ran into the same problem when setting up Electrum with a Coldcard in air gap mode and trying to encrypt the wallet.

As far as I understand it is possible to use a watch-only wallet to create transactions on one computer and sign them with an offline wallet on another cold computer:
https://electrum.readthedocs.io/en/latest/coldstorage.html

I was trying to use the same setup with the Coldcard. Unfortunately, it is not possible to save PSBTs from a watch-only wallet. Would allowing watch-only wallets to save and load PSBTs be a possible solution to allow the same setup with a hardware wallet? Or are there technical restrictions that I am not aware of?

SomberNight:
[...]

PR #5993, opened 2020-02-25
BitBox02 Electrum plugin support

TheCharlatan:
This commit adds support for the BitBox02 hardware wallet. It supports both single and multisig for the electrum gui wallet.

To use the plugin a local installation of the BitBox02 python library is required. It can be found on PiPy under the name 'bitbox02' and can be installed from the bitbox02-firmware repository in the py/bitbox02 directory.

All communication to and from the BitBox02 is noise encrypted, the keys required for this are stored in the wallet config file under the bitbox02 key.

The BitBox02 registers a multisig configuration before allowing transaction signing. This multisig configuration includes the threshold, cosigner xpubs, keypath, a variable to indicate for mainnet and testnet, and a name that the user can choose during configuration registration.
The user is asked to register the multisig configuration either during address verification or during transaction signing.

To check the xpub of the BitBox02 for other hardware wallets, a button is added in the wallet info dialog.

Since the BitBox02 does some things differently compared to other hardware wallets, some special cases had to be added. These should be removed/refactored after some review, or a later point in time. These include:

1. The BitBox02 does not accept arbitrary keypaths to retrieve a xpub. Thus the password derivation was changed to be the: `hmac(xpub at m/84'/1'/0' || b"electrum_password_lulz")`. To facilitate this a special case had to be added to base_wizard.py, where the BitBox02 is distinguished by its label. This is also a bad comparison, because its stringy and the field is not necessarily meant for this purpose.

2. The BitBox02 has a second interface for u2f. This is filtered in a special case in plugin.py

3. The BitBox02's hid id contains its version, not a unique id for each device. To still be able to uniquely identify different BitBox02's when more than one is plugged in, the id is set to the BitBox02's hid path in plugin.py.


electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -0,0 +1,628 @@
...
+            )
+
+        if not self.bitbox02_device.device_info()["initialized"]:
+            raise BaseException(
+                "Please initialize the BitBox02 using the BitBox app first before using the BitBox02 in electrum"
+            )
+
+    def check_device_firmware_version(self) -> bool:
+        if self.bitbox02_device is None:
+            raise BaseException(
```
SomberNight: Try never to raise BaseException. Just raise Exception instead.

electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -0,0 +1,628 @@
...
+                "Need to setup communication first before attempting any BitBox02 calls"
+            )
+        return self.bitbox02_device.check_firmware_version()
+
+    def coin_network_from_bip32_list(self, keypath: List[int]) -> int:
+        if len(keypath) > 2:
+            if keypath[1] == 1 + HARDENED:
+                return bitbox02.btc.TBTC
+        return bitbox02.btc.BTC
```
SomberNight: This is a bit weird... Is it a conscious decision that you are choosing testnet magic bytes based on the bip32 path, and not e.g. `constants.net.TESTNET` (which is set for the process if we were started in testnet mode)?

electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -0,0 +1,628 @@
...
+    def is_pairable(self) -> bool:
+        if self.bitbox_hid_info is None:
+            return False
+        return True
+
+    def btc_multisig_config(
+        self, coin, bip32_path: List[int], wallet: Multisig_Wallet
+    ):
+        """
+        Get a mock multisig 1-of-2 multisig with the current device and some other arbitrary xpub.
```
SomberNight: Is this comment out of date?

electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -0,0 +1,628 @@
...
+                }
+            )
+
+            if tx_script_type == None:
+                tx_script_type = txin.script_type
+            elif tx_script_type != txin.script_type:
+                raise Exception("Cannot mix different input script types")
+        if full_path is None:
+            raise BaseException(
+                "A wallet owned pubkey was not found in the transaction input to be signed"
```
SomberNight: is it okay that this test is outside the loop and hence will only test the last txin?

electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -0,0 +1,628 @@
...
+    plugin: "BitBox02Plugin"
+
+    def __init__(self, d: StoredDict):
+        super().__init__(d)
+        self.force_watching_only = False
+        self.ux_busy = False
+
+    def dump(self):
+        d = Hardware_KeyStore.dump(self)
+        return d
```
SomberNight: override seems pointless?

electrum/plugins/bitbox02/qt.py
```diff
@@ -0,0 +1,185 @@
...
+                def show_address(keystore=keystore):
+                    keystore.thread.add(
+                        partial(self.show_address, wallet, addrs[0], keystore=keystore)
+                    )
+
+                device_name = "{} ({})".format(self.device, keystore.label)
+                menu.addAction(_("Show on {}").format(device_name), show_address)
+
+    @hook
+    def transaction_dialog(self, dia: TxDialog):
```
SomberNight: this hook looks empty. remove?

SomberNight:
> All communication to and from the BitBox02 is noise encrypted, the keys required for this are stored in the ~~wallet~~ config file under the bitbox02 key.

To be clear, there are many wallet files (one for each "wallet"), and there is a single global config file.
The wallet files can optionally be encrypted, the config is always plaintext.
You are currently saving these keys into the global config file. Is this intentional?

Could you give a high level summary of these keys? I see there is one privkey and many "remote pubkey"s. Is the idea that the host PC has an identity (privkey), and each hw device you have also has one (remote pubkey)?

electrum/plugins/bitbox02/bitbox02.py
TheCharlatan: Come to think of it, is there a scenario where inputs that we cannot sign are combined with inputs that we own in the electrum wallet?

electrum/bitcoin.py
```diff
@@ -423,6 +423,21 @@ def address_to_script(addr: str, *, net=None) -> str:
         raise BitcoinException(f'unknown address type: {addrtype}')
     return script
 
+def address_to_hash(addr: str, *, net=None) -> [int, bytes]:
```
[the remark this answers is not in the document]
TheCharlatan: Done.

electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -0,0 +1,628 @@
...
+from electrum.wallet import Standard_Wallet, Multisig_Wallet, Deterministic_Wallet
+from electrum.util import bh2u, UserFacingException
+from electrum.base_wizard import ScriptTypeNotSupported
+from electrum.logging import get_logger
+from electrum.crypto import hmac_oneshot
+from electrum.plugin import Device, DeviceInfo
+from electrum.simple_config import SimpleConfig
+from electrum.json_db import StoredDict
+
+from electrum.gui.qt.installwizard import InstallWizard
```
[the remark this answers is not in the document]
TheCharlatan: Done.

electrum/plugins/bitbox02/bitbox02.py
TheCharlatan: Done.

electrum/plugins/bitbox02/bitbox02.py
TheCharlatan: It's a hack to get the testnet xpub for the wallet encryption. Once the encryption is fixed, I'd move to the constant as well.

electrum/plugins/bitbox02/bitbox02.py
TheCharlatan: Yes, Updated it now.

electrum/plugins/bitbox02/bitbox02.py
TheCharlatan: Removed.

electrum/plugins/bitbox02/qt.py
TheCharlatan: Removed.

electrum/plugin.py
```diff
@@ -605,6 +606,11 @@ def _scan_devices_with_hid(self) -> List['Device']:
                 if len(id_) == 0:
                     id_ = str(d['path'])
                 id_ += str(interface_number) + str(usage_page)
+                # BitBox02 provides two hid interfaces, the second interface is used for u2f and not relevant here
+                if d["product_id"] == 0x2403 and d["interface_number"] == 1:
+                    continue
```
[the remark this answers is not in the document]
TheCharlatan: I found a nice work around for this, not sure how I missed this initially :P . The interface is now filtered in client constructor, the `hash_usable_connection_with_device` and `is_pairable`.

TheCharlatan:
Thanks for the review @SomberNight.

I don't think the noise keys can be stored in the wallet, since the wallet unlocks after the first encrypted call to the device (get_xpub), which is only available after starting the noise encrypted communication (I might be missing something though - let me know).

> I see there is one privkey and many "remote pubkey"s. Is the idea that the host PC has an identity   (privkey), and each hw device you have also has one (remote pubkey)?

You have the right idea, the privkey is stored once for each app and the remote pubkeys are stored for each BitBox02 device. 

To provide a clear use case for this, imagine a system where the electrum application runs in isolation. On systems where the USB stack is shared between applications, this gives additional privacy to the user.

Maybe also take a look at the noise protocol specification: http://noiseprotocol.org/noise.html#handshake-patterns . We are using the 'XX' handshake pattern and cache the static keys on both the application (electrum) and device (BitBox02) side.

electrum/plugins/bitbox02/bitbox02.py
SomberNight: Not until recently. On master there is a "Combine" button in the advanced tx preview dialog, which allows constructing manual coinjoins. It hasn't really been tested with hw wallets though. Still, the answer is yes.

TheCharlatan: Thanks for the clarification, I moved the check into the if clause now.

electrum/base_wizard.py
```diff
@@ -368,9 +371,14 @@ def f(derivation, script_type):
...
-            xpub = self.plugin.get_xpub(device_info.device.id_, derivation, 'standard', self)
-            password = keystore.Xpub.get_pubkey_from_xpub(xpub, ()).hex()
+            # The BitBox02 does not support non-standard xpubs
+            if device_info.label == "BitBox02":
+                xpub = self.plugin.get_xpub(device_info.device.id_, "m/84'/1'/0'", "p2wpkh", self)
+                password = hmac_oneshot(b"electrum_password_lulz", bitcoin.DecodeBase58Check(xpub), hashlib.sha512).hex()
+            else:
+                derivation = get_derivation_used_for_hw_device_encryption(device_info.label)
+                xpub = self.plugin.get_xpub(device_info.device.id_, derivation, 'standard', self)
+                password = keystore.Xpub.get_pubkey_from_xpub(xpub, ())
```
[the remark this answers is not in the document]
TheCharlatan: Could we put the derivation path into the config and key each derivation path with the corresponding plugin name? Then we can get and set with the plugin name as argument and give the plugins the ability to override the default path.

The reason for the hmac is that I'm not feeling entirely comfortable with using a "normal" account level xpub for this operation. Maybe we could additionally store an argument to indicate that the derived xpub should be hashed.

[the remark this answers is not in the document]
SomberNight: > Could we put the derivation path into the config 

The wallet files should be portable and work standalone. (e.g. user should be able to copy just a wallet file from one machine to another) (btw does this assumption break with the noise keys being in the config?)
Why would putting it into the config be better than just hardcoding it in the source?

> The reason for the hmac is that I'm not feeling entirely comfortable with using a "normal" account level xpub for this operation.

We have been using a custom hardened path exactly to avoid having to think about this stuff :) And also because the custom path is more likely to be "secret" (user might expose account xpub to e.g. accountant or maybe cosigner, but no reason to expose this custom path xpub).

I think the right way to go would be to introduce a new encrypted wallet file format, with new magic bytes, and a proper header (--we were already planning this with @ecdsa to be done later), and put this stuff into the header. At the same time we could do arbitrary changes to how the encryption works.
If we went this route, what should be in the header? for now, assume we keep hw-based file encryption restricted to non-multisig wallets. e.g. we could always do an hmac, and then put a salt and the bip32 path in the header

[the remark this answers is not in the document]
TheCharlatan: > (btw does this assumption break with the noise keys being in the config?)

No, they will just be re-generated in the new wallet and the user will have to re-pair. This is expected behaviour, since every app should get its own static key.

> Why would putting it into the config be better than just hardcoding it in the source?

Special rules could be abstracted to the get and set functions instead of having them all over the place.

> (the) user should be able to copy just a wallet file from one machine to another

Ah, I see. Yeah, that breaks my config proposal.

> If we went this route, what should be in the header? 

This sounds very promising to me. From the top of my mind, how about adding as you said the salt and bip32 path, and additionally a version byte. Maybe @benma has some ideas for this as well.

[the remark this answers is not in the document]
SomberNight: I've opened an issue for the storage file header: https://github.com/spesmilo/electrum/issues/5999

However, we've discussed this with @ecdsa, and that change (if done properly) is non-trivial, and it's not really a priority for us. We don't plan to do it for the next release (4.0) as we don't want to delay that any more.

Wallet files should be standalone as above, so putting it into the config file does not really work; and I would not like to maintain forever a hack that tries to multiple derivation paths.

Would it be possible to whitelist our custom hw encryption path in your firmware? It seems like small change. :)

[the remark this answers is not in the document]
benma: How about either

1) just moving the hack into the get_xpub() function of the BitBox02 plugin? 

```Python
def get_xpub(...):
    if keypath == get_derivation_used_for_hw_device_encryption():
        # return custom bitbox02 encryption key 
```

or b) add a hw plugin base method `get_encryption_key()` which every hw plugin can override?

In both cases, the workaround would be contained only to the bitbox02 plugin, not leaking into the base_wizard and elsewhere.

> Would it be possible to whitelist our custom hw encryption path in your firmware? It seems like small change. :)

We could do it as kind of a "last resort", but I don't think it is a nice solution for a few reasons:

1. it's mixing concerns - getting xpubs should be treated for what it is, allowing a client to generate receive addresses.
1. A special case for Electrum means technical debt that has to be treated differently. E.g. we have plans to have the user confirm exporting accounts (xpubs) on the device by name, so a special keypath like this would need special treatment. more special casing means more testing, more room for unforeseen issues, etc.
1. unclear if it can ever be removed? once wallet files are encrypted with this, it should be supported forever I assume.

Option `a)` above is a similar solution, but implemented in BitBox02 Electrum plugin instead of the actual firmware, so I'd go with that. 

Thoughts?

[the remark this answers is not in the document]
ecdsa: Option a) would indeed be cleaner that your current proposal, and I could accept that. 
However, a separate derivation path for encryption would be cleaner than a hmac salted with "electrum_password_lulz".

In response to your point 2: Seriously? A special case for BitBox means technical debt for Electrum developers. We have no financial interest in accepting your plugin, the only thing we care about is long-term safety of our users. Please read https://github.com/spesmilo/electrum/blob/master/electrum/plugins/README

About point 3:  Once wallet are encrypted with a special derivation path, that special path must be supported forever. This is exactly the definition of technical debt. :-)

benma:
Apologies, I wasn't aware that it could be problematic if the encryption
key was defined by the plugin. Of course I didn't mean to offload any
technical debt.

In this case, we'll add an api call to the firmware to export the xpub at
this derivation path to be used as the encryption key, and call it as
discussed in option a). If that's acceptable we will publish the updated
firmare next week.

electrum/base_wizard.py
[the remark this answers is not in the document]
SomberNight: (g1) Another option is that we could completely disable the storage (file) encryption option for bitbox02 wallets for now. This should not be much work. (note that in this case lightning keys would be plaintext on disk, when enabled)

(g2) With a bit more work, we could allow user-password based encryption for bitbox02 wallets. (This would not be wasted work as it would provide the bases for letting the user choose between userpw and xpubpw (https://github.com/spesmilo/electrum/issues/5561). I think it might not be a lot of work if we don't implement changing between userpw and xpubpw now.) However there might be unforeseen complications, so while I would take a look at this if it is liked, I cannot guarantee it would get implemented right now.

I would really rather not have bitbox02 use a different path for xpubpw. Currently we have the property that decrypting files only depends on your bip32 seed. So e.g. if you have a trezor that you used to encrypt a file, and then you lose the device and obtain a ledger initialised with the same seed, you can decrypt the file. Remember, the file might contain non-seed based stuff, such as tx history metadata and now even lightning state. This would still work with the header-based solution but not with the hacky workarounds.

(g3) Further note, re having an endpoint in the firmware to ask for our custom file enc password, that the current method does not need the full xpub, only the pubkey (not the chaincode, in particular). So you could conceivably create an endpoint that returns just the pubkey at exactly that bip32 path. This would then not interfere with your plans about prompting the user when an xpub is requested as without the chaincode this cannot be used as an account.

SomberNight:
> In this case, we'll add an api call to the firmware to export the xpub at
this derivation path to be used as the encryption key, and call it as
discussed in option a). If that's acceptable we will publish the updated
firmare next week.

Ahh I had not seen your comment before, as it got posted (due to email I guess) out of thread.
Please see in particular this part of my comment in the review-thread:

> (g3) Further note, re having an endpoint in the firmware to ask for our custom file enc password, that the current method does not need the full xpub, only the pubkey (not the chaincode, in particular). So you could conceivably create an endpoint that returns just the pubkey at exactly that bip32 path. This would then not interfere with your plans about prompting the user when an xpub is requested as without the chaincode this cannot be used as an account.

It would be fine to only expose the pubkey and not the full xpub if that better aligns with your future plans.

benma:
>  as it got posted (due to email I guess) out of thread.

Yes exactly, I was wondering where my response went as well :P 

Okay, then we'll add the api call. Thanks!

electrum/base_wizard.py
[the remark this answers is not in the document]
benma: Thanks all. My response landed here due to replying by email: https://github.com/spesmilo/electrum/pull/5993#issuecomment-592471118

SomberNight:
I've cleaned the hw-wallet storage-encryption code up a tiny bit in https://github.com/spesmilo/electrum/commit/f8ba66058308afc73109d4b59b587a38733e1d3a, so you can override `HardwareClientBase.get_password_for_storage_encryption` if needed.

TheCharlatan:
We released a new firmware and library version that allows getting the wallet encryption xpub from the device now. The last commit I pushed removes the special case code for the password. The plugin will now produce a password the same way now as the other devices. As per @SomberNight 's suggestion the network enumeration is now done with the constants.net instead of getting it from the derivation path. 
I did not change anything in the reproducible build system for this plugin yet. Should I include this as well, or do you want to do it yourselves for trust purposes?

SomberNight:
Hmmm. Your python lib has quite a few dependencies :/
Could you at least loosen the version requirements on them a bit? E.g. on `ecdsa` and `protobuf`, the version constraints seem overly restrictive. The constraints could easily become unsatisfiable if just one other hww lib we pull in started specifying so narrow intervals (if there is no overlap).

```
$ pipdeptree
bitbox02==2.0.0
  - base58 [required: >=2.0.0,<3.0.0, installed: 2.0.0]
  - ecdsa [required: >=0.13,<0.14, installed: 0.13.3]
  - hidapi [required: >=0.7.99.post21,<0.8, installed: 0.7.99.post21]
    - setuptools [required: >=19.0, installed: 46.1.3]
  - noiseprotocol [required: >=0.3,<0.4, installed: 0.3.1]
    - cryptography [required: >=2.8, installed: 2.9]
      - cffi [required: >=1.8,!=1.11.3, installed: 1.14.0]
        - pycparser [required: Any, installed: 2.20]
      - six [required: >=1.4.1, installed: 1.14.0]
  - protobuf [required: >=3.7,<3.8, installed: 3.7.1]
    - setuptools [required: Any, installed: 46.1.3]
    - six [required: >=1.9, installed: 1.14.0]
  - semver [required: >=2.8.1,<3.0, installed: 2.9.1]
  - typing [required: >=3.7.4,<3.8, installed: 3.7.4.1]
  - typing-extensions [required: >=3.7.4,<3.8, installed: 3.7.4.2]
btchip-python==0.1.28
  - hidapi [required: >=0.7.99, installed: 0.7.99.post21]
    - setuptools [required: >=19.0, installed: 46.1.3]
ckcc-protocol==1.0.1
  - ecdsa [required: >=0.13, installed: 0.13.3]
  - hidapi [required: >=0.7.99.post21, installed: 0.7.99.post21]
    - setuptools [required: >=19.0, installed: 46.1.3]
  - pyaes [required: Any, installed: 1.6.1]
Cython==0.29.16
keepkey==6.3.1
  - ecdsa [required: >=0.9, installed: 0.13.3]
  - hidapi [required: >=0.7.99.post15, installed: 0.7.99.post21]
    - setuptools [required: >=19.0, installed: 46.1.3]
  - libusb1 [required: >=1.6, installed: 1.7.1]
  - mnemonic [required: >=0.8, installed: 0.19]
  - protobuf [required: >=3.0.0, installed: 3.7.1]
    - setuptools [required: Any, installed: 46.1.3]
    - six [required: >=1.9, installed: 1.14.0]
safet==0.1.5
  - click [required: >=6.2, installed: 7.1.1]
  - ecdsa [required: >=0.9, installed: 0.13.3]
  - hidapi [required: >=0.7.99.post20, installed: 0.7.99.post21]
    - setuptools [required: >=19.0, installed: 46.1.3]
  - libusb1 [required: >=1.6.4, installed: 1.7.1]
  - mnemonic [required: >=0.17, installed: 0.19]
  - requests [required: >=2.4.0, installed: 2.23.0]
    - certifi [required: >=2017.4.17, installed: 2020.4.5.1]
    - chardet [required: >=3.0.2,<4, installed: 3.0.4]
    - idna [required: >=2.5,<3, installed: 2.9]
    - urllib3 [required: >=1.21.1,<1.26,!=1.25.1,!=1.25.0, installed: 1.25.8]
  - setuptools [required: >=19.0, installed: 46.1.3]
trezor==0.12.0
  - click [required: >=7,<8, installed: 7.1.1]
  - construct [required: >=2.9, installed: 2.10.56]
  - ecdsa [required: >=0.9, installed: 0.13.3]
  - libusb1 [required: >=1.6.4, installed: 1.7.1]
  - mnemonic [required: >=0.17, installed: 0.19]
  - requests [required: >=2.4.0, installed: 2.23.0]
    - certifi [required: >=2017.4.17, installed: 2020.4.5.1]
    - chardet [required: >=3.0.2,<4, installed: 3.0.4]
    - idna [required: >=2.5,<3, installed: 2.9]
    - urllib3 [required: >=1.21.1,<1.26,!=1.25.1,!=1.25.0, installed: 1.25.8]
  - setuptools [required: >=19.0, installed: 46.1.3]
  - typing-extensions [required: >=3.7.4, installed: 3.7.4.2]
wheel==0.34.2
```

SomberNight:
Oh, also the latest `hidapi` has some nice fixes, which is too new according to the restrictions above...

SomberNight:
Ok, so there is actually an issue with newer protobuf :P
You seem to have a collision with keepkey.

```
I/p | plugin.Plugins | loaded bitbox02
I/p | plugin.Plugins | loaded coldcard
I/p | plugin.Plugins | loaded digitalbitbox
E/p | plugin.Plugins | cannot load plugin for: keepkey
  File "...\electrum\electrum\plugin.py", line 119, in load_plugin
    plugin = module.Plugin(self, self.config, name)
  File "...\electrum\electrum\plugins\keepkey\keepkey.py", line 80, in __init__
    from . import client
  File "...\electrum\electrum\plugins\keepkey\client.py", line 1, in <module>
    from keepkeylib.client import proto, BaseClient, ProtocolMixin
  File "...\python-keepkey\keepkeylib\client.py", line 37, in <module>
    from . import mapping
  File "...\python-keepkey\keepkeylib\mapping.py", line 1, in <module>
    from . import messages_pb2 as proto
  File "...\python-keepkey\keepkeylib\messages_pb2.py", line 3862, in <module>
    _sym_db.RegisterMessage(Success)
  File "...\python38\site-packages\google\protobuf\symbol_database.py", line 83, in RegisterMessage
    self.RegisterMessageDescriptor(desc)
  File "...\python38\site-packages\google\protobuf\symbol_database.py", line 94, in RegisterMessageDescriptor
    self.pool._AddDescriptor(message_descriptor)
  File "...\python38\site-packages\google\protobuf\descriptor_pool.py", line 238, in _AddDescriptor
    self._CheckConflictRegister(desc, desc.full_name, desc.file.name)
  File "...\python38\site-packages\google\protobuf\descriptor_pool.py", line 191, in _CheckConflictRegister
    raise TypeError(error_msg)
TypeError: Conflict register for file "messages.proto": Success is already defined in file "hww.proto". Please fix the conflict by adding package name on the proto file, or use different name for the duplication.

The above exception was the direct cause of the following exception:

  File "...\electrum\electrum\plugin.py", line 169, in get_hardware_support
    p = self.get_plugin(name)
  File "...\electrum\electrum\plugin.py", line 203, in get_plugin
    self.load_plugin(name)
  File "...\electrum\electrum\plugin.py", line 121, in load_plugin
    raise Exception(f"Error loading {name} plugin: {repr(e)}") from e
Exception: Error loading keepkey plugin: TypeError('Conflict register for file "messages.proto": Success is already defined in file "hww.proto". Please fix the conflict by adding package name on the proto file, or use different name for the duplication.')
I/p | plugin.Plugins | loaded ledger
I/p | plugin.Plugins | loaded safe_t
I/p | plugin.Plugins | loaded trezor
```

I have tested with different versions of protobuf:

protobuf 3.7.1 - works
protobuf 3.8.0 - works
protobuf 3.9.2 - fails
protobuf 3.10.0 - fails
protobuf 3.11.3 - fails

I guess these protos collide:
https://github.com/keepkey/device-protocol/blob/4ef20fbba4a0e2eeb1ddd110b07f818a3cfde94b/messages.proto#L247
https://github.com/digitalbitbox/bitbox02-firmware/blob/4e273ffd60ded62022f35ded9368aafc0fd446fe/messages/hww.proto#L36

I did not really investigate why the version of protobuf matters for a collision though.

benma:
@SomberNight thanks for testing, and very puzzling that different versions behave differently here.

Protobuf seems to have a global symbol table :S I had a similar issue in Go's protobuf implementation, where it would also panic on duplicate messages.

Reading up on how to [add a namespace](https://developers.google.com/protocol-buffers/docs/proto3#packages):

> In Python, the package directive is ignored, since Python modules are organized according to their location in the file system. 

Which is what I would expect. 

A quick fix could be to prefix our protobuf types with `BB02` in the Python package, but we'll investigate the reason behind it and if there is a cleaner fix.

x1ddos:
Not sure this will help but anyway, @SomberNight would you be able to try with a patch to bitbox02 Python package? 
https://gist.github.com/x1ddos/09d75b35e689ffc96dd5ffd09479b113

I simply added

```proto
package shiftcrypto.bitbox02;
```
to https://github.com/digitalbitbox/bitbox02-firmware/blob/c2c6a0e6d4b05b5028fccc7a8d856069cf8bef59/messages/hww.proto and regenerated .pb2.py files.

It's just the conflict check seems to be using `full_name` in https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/python/google/protobuf/descriptor_pool.py#L238 which comprises the package name.

This would be very helpful to identify whether protobuf-python indeed ignores the package directive and to figure out the best fix for this.

SomberNight:
@x1ddos Your patch seems to work. I no longer get the exception with it.

x1ddos:
Cool. So, we'll package-prefix all proto in bitbox02. It's probably best if keepkey and everyone else did as well to avoid similar collisions in the future.

x1ddos:
Created an issue at https://github.com/protocolbuffers/protobuf/issues/7369 but I wouldn't hold my breath. There seem to be a similar issue back from 2016: https://github.com/protocolbuffers/protobuf/issues/2273.

SomberNight:
Btw, I don't think it makes sense to pull in the `typing` module from PyPI.
https://github.com/digitalbitbox/bitbox02-firmware/blob/974c22de5cadb14a4c2396b76923bb3e44087e17/py/bitbox02/setup.py#L60

https://pypi.org/project/typing/
> NOTE: in Python 3.5 and later, the typing module lives in the stdlib, and installing this package has NO EFFECT. 

One fewer dependency :P

SomberNight:
Ah in fact `typing` is causing issues with pyinstaller (for Win/Mac binaries).
(`collect_submodules` fails, and hence parts of `bitbox02` will not get picked up)

```
AttributeError: type object 'Callable' has no attribute '_abc_registry'
  File "<string>", line 41, in <module>
  File "<string>", line 20, in walk_packages
  File "c:\python3\lib\site-packages\bitbox02\bitbox02\__init__.py", line 43, in <module>
    from .bitbox02 import (
  File "c:\python3\lib\site-packages\bitbox02\bitbox02\bitbox02.py", line 21, in <module>
    from typing import Optional, List, Dict, Tuple, Any, Generator, Union
  File "c:\python3\lib\site-packages\typing.py", line 1357, in <module>
    class Callable(extra=collections_abc.Callable, metaclass=CallableMeta):
  File "c:\python3\lib\site-packages\typing.py", line 1005, in __new__
    self._abc_registry = extra._abc_registry
AttributeError: type object 'Callable' has no attribute '_abc_registry'
```

(If I manually make sure the `typing` 3rd party module does get installed during the build; I have tested the built Windows binary and it works)

benma:
> Btw, I don't think it makes sense to pull in the `typing` module from PyPI.

Great catch, thanks! I published bitbox02 2.0.3 that removes this dep.

SomberNight:
Thanks everyone for your work/help.

benma:
Thanks a lot @SomberNight for reviewing and merging this PR!

Do you know a rough ETA for the next electrum release?

PR #6001, opened 2020-02-28
mnemonic: make sure newly generated seeds are not valid as bip39

SomberNight:
Currently there is an approx 1/16 chance that newly generated seeds have a valid bip39 checksum (and hence are both valid electrum seeds and bip39 seeds at the same time).

This patch would make sure that newly generated seeds are not also valid bip39 seeds by accident.

As a result, seed generation would become a bit slower, taking approx. 20% more time.

```
import time
import electrum

mn = electrum.mnemonic.Mnemonic()

t0 = time.time()
for i in range(1000):
    seed = mn.make_seed("standard")

t1 = time.time()
print(f"time taken: {t1-t0} sec")
```


andronoob:
This change makes some sense, however it cannot make any change to all those existing seed phrases generated with unwanted "BIP39-Electrum2.0 duality".

Both BIP39, Electrum2.0, aezeed, ... and all the other alleged "brilliant new design of mnemonic phrase", consist the chaotic situation altogether.

Even if BIP39 got abandoned from now on, by all wallet devs, there would still inevitably be users who rely on BIP39 tightly.

And, ditto for all those "opponent to BIP39", including Electrum2.0, that each of them also has its own users who rely on it tightly.

andronoob:
Currently, there's also a probability of (2^4 + 1 + 1) / 2^12 = 9/2048 (0.439%) for a randomly generated BIP39 mnemonic to be a valid Electrum2.0 seed at same time. This probability could be higher as more and more newly defined version numbers are expected in the future.

Electrum can only make sure that it won't generate a Electrum2.0 seed with such "duality", but it can't stop other wallets generating BIP39 mnemonics which have valid Electrum version number.

PR #6003, opened 2020-03-02
Htlc switch

ecdsa:
First implementation of htlc switch.

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -408,7 +408,7 @@ class Channel(Logger):
         self.logger.info("add_htlc")
         return htlc
 
-    def receive_htlc(self, htlc: UpdateAddHtlc) -> UpdateAddHtlc:
+    def receive_htlc(self, htlc: UpdateAddHtlc, onion_packet:bytes = None) -> UpdateAddHtlc:
         """
         ReceiveHTLC adds an HTLC to the state machine's remote update log. This
         method should be called in response to receiving a new HTLC from the remote
@@ -427,6 +427,11 @@ class Channel(Logger):
                     f' HTLC amount: {htlc.amount_msat}')
         with self.db_lock:
             self.hm.recv_htlc(htlc)
+            local_ctn = self.get_latest_ctn(LOCAL)
+            remote_ctn = self.get_latest_ctn(REMOTE)
+            if onion_packet:
+                self.hm.log['unfulfilled_htlcs'][htlc.htlc_id] = local_ctn, remote_ctn, onion_packet.hex()
+
         self.logger.info("receive_htlc")
         return htlc
 
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -430,7 +430,7 @@ class Channel(Logger):
             local_ctn = self.get_latest_ctn(LOCAL)
             remote_ctn = self.get_latest_ctn(REMOTE)
             if onion_packet:
-                self.hm.log['unfulfilled_htlcs'][htlc.htlc_id] = local_ctn, remote_ctn, onion_packet.hex()
+                self.hm.log['unfulfilled_htlcs'][htlc.htlc_id] = local_ctn, remote_ctn, onion_packet.hex(), False
 
         self.logger.info("receive_htlc")
         return htlc
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -1268,10 +1268,7 @@ class Peer(Logger):
 
     async def _fulfill_htlc(self, chan: Channel, htlc_id: int, preimage: bytes):
         self.logger.info(f"_fulfill_htlc. chan {chan.short_channel_id}. htlc_id {htlc_id}")
-        if not chan.can_send_ctx_updates():
-            self.logger.info(f"dropping chan update (fulfill htlc {htlc_id}) for {chan.short_channel_id}. "
-                             f"cannot send updates")
-            return
+        assert chan.can_send_ctx_updates(), f"cannot send updates: {chan.short_channel_id}"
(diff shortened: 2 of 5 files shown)


electrum/lnworker.py
```diff
@@ -1325,3 +1326,55 @@ def current_feerate_per_kw(self):
         if feerate_per_kvbyte is None:
             feerate_per_kvbyte = FEERATE_FALLBACK_STATIC_FEE
         return max(253, feerate_per_kvbyte // 4)
+
+    @log_exceptions
+    async def htlc_switch(self):
```
SomberNight: note that
- an await re a single peer can ~arbitrarily delay processing for another peer
- an exception raised by any peer that propagates out to here, will kill the task (and hence lnwallet)

ecdsa: thanks for the review. your first point should be addressed by the latest commit.

ecdsa: Regarding your second point, I realize that the htlc_switch task could be distributed to peers, and moved to lnpeer. When I started working on this, I was under the impression that this task would need to be centralized, but that does not seem to be the case anymore.

electrum/lnpeer.py
```diff
@@ -1425,3 +1426,53 @@ def verify_signature(tx, sig):
         # broadcast
         await self.network.try_broadcasting(closing_tx, 'closing')
         return closing_tx.txid()
+
+    @log_exceptions
```
SomberNight: this decorator is unnecessary I think -- the taskgroup should already take care of it

electrum/lnpeer.py
```diff
@@ -1484,3 +1426,53 @@ def verify_signature(tx, sig):
...
+                done = set()
+                unfulfilled = chan.hm.log.get('unfulfilled_htlcs', {})
+                for htlc_id, (local_ctn, remote_ctn, onion_packet_hex, forwarded) in unfulfilled.items():
+                    if chan.get_oldest_unrevoked_ctn(LOCAL) <= local_ctn:
+                        continue
+                    if chan.get_oldest_unrevoked_ctn(REMOTE) <= remote_ctn:
+                        continue
+                    chan.logger.info(f'found unfulfilled htlc: {htlc_id}')
+                    onion_packet = OnionPacket.from_bytes(bytes.fromhex(onion_packet_hex))
+                    htlc = chan.hm.log[REMOTE]['adds'][htlc_id]
```
SomberNight: same as above re "deep access into `chan.hm.log`"

electrum/lnpeer.py
```diff
@@ -1484,3 +1426,53 @@ def verify_signature(tx, sig):
...
+    @log_exceptions
+    async def htlc_switch(self):
+        while True:
+            await asyncio.sleep(0.1)
+            for chan_id, chan in self.channels.items():
+                if not chan.can_send_ctx_updates():
+                    continue
+                self.maybe_send_commitment(chan)
+                done = set()
+                unfulfilled = chan.hm.log.get('unfulfilled_htlcs', {})
```
SomberNight: I think this kind of deep access into `chan.hm.log` should be abstracted away. We should add methods to `HTLCManager` for these specific cases if they don't yet exist.

electrum/lnchannel.py
```diff
@@ -427,6 +427,11 @@ def receive_htlc(self, htlc: UpdateAddHtlc) -> UpdateAddHtlc:
                     f' HTLC amount: {htlc.amount_msat}')
         with self.db_lock:
             self.hm.recv_htlc(htlc)
+            local_ctn = self.get_latest_ctn(LOCAL)
+            remote_ctn = self.get_latest_ctn(REMOTE)
+            if onion_packet:
+                self.hm.log['unfulfilled_htlcs'][htlc.htlc_id] = local_ctn, remote_ctn, onion_packet.hex(), False
```
SomberNight: same as elsewhere re "deep access into `chan.hm.log`"

electrum/lnpeer.py
```diff
@@ -1131,90 +1132,60 @@ def on_update_add_htlc(self, payload):
...
+        htlc = UpdateAddHtlc(
+            amount_msat=amount_msat_htlc,
+            payment_hash=payment_hash,
+            cltv_expiry=cltv_expiry,
+            timestamp=int(time.time()),
+            htlc_id=htlc_id)
+        chan.receive_htlc(htlc, onion_packet)
+
+    def maybe_forward_htlc(self, chan: Channel, htlc: UpdateAddHtlc, *,
+                           onion_packet: OnionPacket, processed_onion: ProcessedOnionPacket):
```
SomberNight: Maybe add a comment about the need to think carefully what happens if an exception is raised.
E.g. if next_peer causes us to raise we disconnect from self ("prev_peer" I guess) -- well depending on the callee.

electrum/lnpeer.py
```diff
@@ -1484,3 +1426,53 @@ def verify_signature(tx, sig):
...
+                    processed_onion = process_onion_packet(onion_packet, associated_data=payment_hash, our_onion_private_key=self.privkey)
+                    preimage, error = None, None
+                    if processed_onion.are_we_final:
+                        preimage, error = self.maybe_fulfill_htlc(
+                            chan=chan,
+                            htlc=htlc,
+                            onion_packet=onion_packet,
+                            processed_onion=processed_onion)
+                    elif not forwarded:
+                        next_chan, next_peer, error = self.maybe_forward_htlc(
```
SomberNight: `next_chan` and `next_peer` are unused -- maybe no need to even return them from method?

electrum/lnpeer.py
ecdsa: good point. I will catch exceptions raised by next_peer

electrum/lnpeer.py

PR #6020, opened 2020-03-07
add dockerization

zoedberg:
Hi,

this PR introduces electrum dockerization.

All related files are placed in `contrib/docker` (let me know if you prefer another location).

The image can be run in daemon or GUI mode.

Other details are in the provided README.

Please double check `.dockerignore`, at the moment is just a copy of `.gitignore`.


SomberNight:
> Please double check .dockerignore, at the moment is just a copy of .gitignore.

We already use dockerfiles for several build scripts. I think a single global dockerignore would interfere with those; wouldn't it?
For some reason the Travis tests did not execute... - it would be easy to tell then :)

zoedberg:
> We already use dockerfiles for several build scripts. I think a single global dockerignore would interfere with those; wouldn't it?
> For some reason the Travis tests did not execute... - it would be easy to tell then :)

Yes, it would apply to all dockerfiles. Its meaning is to avoid copy of unnecessary files into the docker image (protecting from `COPY . .`). I think it can be kept and be useful also to other builds, but we need to be sure we're not excluding files that are necessary.

zoedberg:
If you prefer we can just drop the `.dockerignore` file and that should work fine

x-jokay:
@zoedberg did you already try getting this one work using [alpine](https://hub.docker.com/_/alpine) instead of [debian](https://hub.docker.com/_/debian)?

SomberNight:
note: Even forgetting about dockerignore issues, we haven't merged this because we are not sure how much interest there would be for it from users, and we don't really want to take on the maintenance burden if not much

zoedberg:
Sorry for the late answer.

> @zoedberg did you already try getting this one work using alpine instead of debian?

I generally prefer using debian but if you prefer alpine (I guess for image size) I can change the Dockerfile.

> note: Even forgetting about dockerignore issues, we haven't merged this because we are not sure how much interest there would be for it from users, and we don't really want to take on the maintenance burden if not much

I don't see a lot of burden in keeping the Dockerfile updated, it should require changes only when new project dependencies are added, that I guess doesn't happen often.
But, if you think there will be no interest from users, feel free to reject this PR.

PR #6028, opened 2020-03-09
lnrouter: speed up path-finding by caching edge cost

SomberNight:
Turns out most of the time (~70%) for `find_path_for_payment` is spent in `_edge_cost`.
`_edge_cost` is not doing anything particularly expensive, just a whole bunch
of cheap things, and it is called many times (around once for every channel policy).
By caching `_edge_cost` like this, the runtime (wall clock) for `find_path_for_payment`
goes from ~1.2 sec to ~0.35 sec on my machine.

Not sure about all implications: underlying graph might change while path finding runs, cache can become stale, cache is using buckets for payment amount... but I think nothing horrible could happen, so maybe worth the speedup.

However, the cache is quite large (in terms of mem usage). :/ Maybe around 2.6 MB per payment amount bucket for the current graph (~55k channel policies).
So for say 10 buckets, it's already 26 MB, which is a lot for an Android phone, just for this.

Still, failed research should also be published; hence this PR.

### DIFF
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -26,6 +26,7 @@
 import queue
 from collections import defaultdict
 from typing import Sequence, List, Tuple, Optional, Dict, NamedTuple, TYPE_CHECKING, Set
+import time
 
 from .util import bh2u, profiler
 from .logging import Logger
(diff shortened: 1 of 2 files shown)


SomberNight:
Note: I've tried substituting the stdlib `heapq` (PriorityQueue) for https://pypi.org/project/HeapDict/ -- as `heapq` does not implement "decrease-key". The performance difference is negligible.

PR #6039, opened 2020-03-17
Document how to disable proxy

interrupt00:
`deserialize_proxy()` from `network.py` supports `none` value for `--proxy`.
It is worth documenting since it is a useful feature.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1224,7 +1224,7 @@ argparse._SubParsersAction.__call__ = subparser_call
 def add_network_options(parser):
     parser.add_argument("-1", "--oneserver", action="store_true", dest="oneserver", default=None, help="connect to one server only")
     parser.add_argument("-s", "--server", dest="server", default=None, help="set server host:port:protocol, where protocol is either t (tcp) or s (ssl)")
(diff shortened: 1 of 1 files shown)


electrum/commands.py
```diff
@@ -1224,7 +1224,7 @@ def subparser_call(self, parser, namespace, values, option_string=None):
 def add_network_options(parser):
     parser.add_argument("-1", "--oneserver", action="store_true", dest="oneserver", default=None, help="connect to one server only")
     parser.add_argument("-s", "--server", dest="server", default=None, help="set server host:port:protocol, where protocol is either t (tcp) or s (ssl)")
-    parser.add_argument("-p", "--proxy", dest="proxy", default=None, help="set proxy [type:]host[:port], where type is socks4,socks5 or http")
+    parser.add_argument("-p", "--proxy", dest="proxy", default=None, help="set proxy [type:]host[:port] (or none to disable proxy), where type is socks4,socks5 or http")
```
SomberNight: We should put it into quotes as otherwise it's not obvious what is meant here.
```suggestion
    parser.add_argument("-p", "--proxy", dest="proxy", default=None, help="set proxy [type:]host[:port] (or 'none' to disable proxy), where type is socks4,socks5 or http")
```

interrupt00: Good point, thank you.

SomberNight:
thanks

PR #6054, opened 2020-03-29
Readme: Clarify dependencies of make_libsecp256k1.sh

JeremyRand:
`README.rst` lists the `apt-get` requirements for `pull_locale` (which is good) but neglects to do the same for `make_libsecp256k1.sh`.  This PR adds the relevant info for the latter.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -52,6 +52,7 @@ For elliptic curve operations, `libsecp256k1`_ is a required dependency::
 Alternatively, when running from a cloned repository, a script is provided to build
 libsecp256k1 yourself::
 
+    sudo apt-get install automake libtool
     ./contrib/make_libsecp256k1.sh
 
 Due to the need for fast symmetric ciphers, either one of `pycryptodomex`_


SomberNight:
Maybe also `autotools-dev` and `autoconf`? Not sure tbh.
Also `mingw-w64` if cross compiling to windows.

On that note, would it not be better to list them inside the script itself? (to try to keep the README short -- list of deps with conditionals :/ )
Maybe we could test in the script whether they are available and fail with a nice error msg if not.

SomberNight:
Alternatively to my previous comment, we could just do as the current PR - list the non-trivial shared dependencies;
and re the conditional dep for cross-compiling we can add a comment into the script itself.

SomberNight:
> Maybe also `autotools-dev` and `autoconf`? Not sure tbh.

Just realised that `automake` pulls those in :P

JeremyRand:
@SomberNight the script already displays a hint when it thinks libtool and automake aren't present, but the hint isn't particularly explicit (i.e. it doesn't suggest using `apt-get` and doesn't make it clear what the package names are).  If you prefer to make that hint more explicit, that would work for me.  It also has a hint for the case where a compiler is missing, but that hint is even less explicit.  For consistency I think the same policy should apply to `pull_locale`, whatever we end up doing.

SomberNight:
Yes okay, I see. It's not worth the time pondering on this further.

I will add a comment into the file itself about cross-compiling.

PR #6064, opened 2020-03-31
trezor: bump lib version, implement new passphrase-on-device UI

matejcik:
In upcoming Trezor T firmware, we moved the decision where the passphrase is entered from the device to the host.

IOW, in firmware < 2.3.0, when using a passphrase, the Trezor screen would ask: "enter passphrase on host / device?"
Depending on the choice, either Trezor would ask for a passphrase, or indicate to Electrum that the passphrase dialog should be displayed.

In firmware >= 2.3.0, the passphrase dialog should be shown by default, with an additional option to "Enter Passphrase on Device", which would indicate to Trezor that it should prompt for passphrase.

See [this doc](https://github.com/trezor/trezor-firmware/blob/master/docs/common/communication/passphrase-redesign-migration.md) for details.

Practical effect is that users with new firmware and old trezorlib/Electrum will only be able to enter passphrase on the host.

This PR introduces the necessary new UI element to allow entering passphrase on device. With trezorlib 0.12 this should transparently work both with older firmwares (with choice on device screen) and newer ones.

---

In testing, I've run into a problem when creating a wallet: the pop-up saying "Please enter your passphrase" is not rendered correctly. The pop-up window is displayed, but it is empty.
No idea what this is about. When it is triggered from the wallet after the locking timeout, it is shown correctly.

### DIFF
--- a/contrib/requirements/requirements-hw.txt
+++ b/contrib/requirements/requirements-hw.txt
@@ -8,7 +8,7 @@
 # see https://github.com/spesmilo/electrum/issues/5859
 Cython>=0.27
 
-trezor[hidapi]>=0.11.5
+trezor[hidapi]>=0.12.0
 safet>=0.1.5
 keepkey>=6.3.1
 btchip-python>=0.1.26
--- a/electrum/plugins/trezor/__init__.py
+++ b/electrum/plugins/trezor/__init__.py
@@ -2,7 +2,7 @@ from electrum.i18n import _
 
 fullname = 'Trezor Wallet'
 description = _('Provides support for Trezor hardware wallet')
-requires = [('trezorlib','github.com/trezor/python-trezor')]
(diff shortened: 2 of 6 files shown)


matejcik:
trezorlib 0.12 will be released this week

matejcik:
trezorlib 0.12.0 is out now

SomberNight:
> In testing, I've run into a problem when creating a wallet: the pop-up saying "Please enter your passphrase" is not rendered correctly. The pop-up window is displayed, but it is empty.
No idea what this is about. When it is triggered from the wallet after the locking timeout, it is shown correctly.

"fixed" this in https://github.com/spesmilo/electrum/commit/371f55a0f922cf57196ae3d8d7c2fabc7a9f0f08
(a proper fix would probably involve a lot of refactoring :/ )

PR #6070, opened 2020-04-04
Save channel seed in localconfig

ecdsa:


ecdsa:
@SomberNight this new version preserves old channels.
please add hardened derivations and versioning here

PR #6076, opened 2020-04-08
Use specific Exception when chain isn't synced

JeremyRand:
Makes it easier for calling code to know what error happened.


SomberNight:
OK, but please rename to `WalletNotSynchronized` instead

SomberNight:
Ah I see you are using it in a broader sense in your fork. Well okay.

Issue #6085, opened 2020-04-14
Don't fall back to a random public server if the connection string fails to parse


lukechilds:
If Electrum cannot parse the connection string it falls back to a random public server:

```
W/n | network | failed to parse server-string; falling back to random.
```

I don't think this is desirable behaviour. This means if a user incorrectly types a connection string to their own server which makes it unparsable, Electrum will fall back to using a random public server and potentially dox all of their UTXOs and their IP.

Consider the following:

1. User has a local Electrum server they want to connect to for improved privacy.
2. User starts Electrum with `electrum --server localhost:50001:t:`.
3. The typo (trailing `:`) means Electrum will fail to parse and fall back to a random public server.
4. Electrum proceeds to send all of the users addresses and their IP to a random public server.
5. The user is not made aware they are not connected to their own server and are broadcasting everything to a 3rd party if they don't manually check console logs or the network settings in the GUI.

Instead, if the connection string cannot be parsed Electrum, should just fail with an error so the user can fix it and avoid unexpected behaviour and potential major privacy leaks.

lukechilds:
Related comment: https://github.com/spesmilo/electrum/issues/6084#issuecomment-613220016

h2o10:
how it behaves if user species `--oneserver`?

ecdsa:
indeed that is not desirable.
iirc the purpose of that fallback is to deal with corrupt config files.
I think we can remove it.

PR #6086, opened 2020-04-14
Hard fail on bad server-string

lukechilds:
Resolves https://github.com/spesmilo/electrum/issues/6085

I just copied the same print then exit fail style from here:

https://github.com/spesmilo/electrum/blob/da8b24d61af00109d480fe4ee5b73d130a7c1f25/electrum/gui/stdio.py#L26-L27

However this doesn't seem to actually exit the process for me, in the terminal I see the "Failed to parse server-string." text but the python process just hangs. Am I doing something wrong?

### DIFF
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -268,8 +268,8 @@ class Network(Logger):
             try:
                 deserialize_server(self.default_server)
             except:
-                print('Failed to parse server-string.')
-                exit()
+                self.logger.warning('failed to parse server-string; falling back to localhost.')
+                self.default_server = "localhost:50002:s"
         if not self.default_server:
             self.default_server = pick_random_server()
 
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -268,8 +268,8 @@ class Network(Logger):
             try:
                 deserialize_server(self.default_server)
             except:
-                self.logger.warning('failed to parse server-string; falling back to random.')
-                self.default_server = None
+                print('Failed to parse server-string.')
+                exit()
         if not self.default_server:
             self.default_server = pick_random_server()
 


ecdsa:
you should not call exit() from there.
it would be better to let the user fix the server in their setting dialog, in case the error does not come from command line, but from their configurattion file

SomberNight:
For a GUI user, this would just be interpreted as a mysterious crash.

electrum/network.py
```diff
@@ -268,8 +268,8 @@ def __init__(self, config: SimpleConfig, *, daemon: 'Daemon' = None):
             try:
                 deserialize_server(self.default_server)
             except:
-                self.logger.warning('failed to parse server-string; falling back to random.')
-                self.default_server = None
+                print('Failed to parse server-string.')
+                exit()
```
SomberNight, line 272:
```suggestion
                self.logger.warning('failed to parse server-string; falling back to localhost.')
                self.default_server = "localhost:50002:s"
```

lukechilds:
Good point re not crashing. I've added @SomberNight's suggestion of falling back to localhost.

Do you think it would be worth changing the localhost fallback to a non standard port?

Just incase the tiny chance that the user actually have an SSL Electrum server running on localhost:50002. It would probably be better to not connect to anything than connect to something they didn't intend.

lukechilds:
Obviously this is pretty unlikely but there is still a potential privacy leak here:

1. I have a local electrum server running which I normally connect to.
1. I want to make a private transaction that can't be linked to my IP so I'm gonna connect to a different server over Tor to just to broadcast this TX.
1. I enter the server-string incorrectly which fails to parse and falls back to localhost.
1. Electrum then connects to my local electrum server.
1. I broadcast the sensitive TX thinking it will be sent over Tor and then broadcast, instead it's broadcast from my IP.

Not a particularly likely scenario but still not great that it could happen. Thoughts @SomberNight @ecdsa?

SomberNight:
> Obviously this is pretty unlikely but there is still a potential privacy leak here: [...]

Well... yes okay. A bit contrived but fine. :)
I don't see a two-line fix like this one though. I'll play with the code a bit.

lukechilds:
@SomberNight just changing the port should make this much more unlikely.

Someone running a local Electrum server over SSL on localhost:50002 and entering an invalid connection string, while not very likely, is not outside the realms of possibility.

Someone running their local SSL server on localhost:50123 is almost definitely not gonna happen. Maybe that's good enough?

PR #6087, opened 2020-04-14
Use non-standard localhost port for server-string fallback

lukechilds:
Context: https://github.com/spesmilo/electrum/pull/6086#issuecomment-613447577


SomberNight:
How about port=1 instead?

lukechilds:
Even better.

PR #6094, opened 2020-04-16
Add --serverfingerprint option

### LINKED ISSUE #6084: Allow specifying SSL server fingerprint
It would be great if we could ensure an SSL connection is not being MITM'd on initial connection by specifying a known fingerprint.

E.g:

```
$ electrum --server bitcoin.lukechilds.co:50002:s:ED1FE28B21A5943A8810747B2A6DDFC0E3089115D6205125E5C4D3EE8E126F80
```

The above connection would fail if bitcoin.lukechilds.co:50002 doesn't respond with a certificate with a matching fingerprint.

I'm aware electrum pins certificates in `~/.electrum/certs`, however this is only ensures we can't be MITM'd after the initial connection has been made and manually verified. It doesn't guarantee a secure initial connection.

Using a fingerprint as part of the server URI guarantees a secure initial connection and is much more convenient for configuration and safely sharing between users.

For example, asking my friends and family to securely connect to my server is currently quite awkward and error prone:

1. Find your electrum config dir.
2. Create a file inside it called `./certs/bitcoin.lukechilds.co`.
3. Copy the contents of my certificate file into it:
```
-----BEGIN CERTIFICATE-----
MIIDCTCCAfGgAwIBAgIUL0ou+8rSEE17I9yGq7NeesxhgdkwDQYJKoZIhvcNAQEL
BQAwFDESMBAGA1UECgwJRWxlY3RydW1YMB4XDTE5MTExMTA0MzgxOFoXDTIwMTEx
MDA0MzgxOFowFDESMBAGA1UECgwJRWxlY3RydW1YMIIBIjANBgkqhkiG9w0BAQEF
AAOCAQ8AMIIBCgKCAQEAvJfIOHkhW7NTGYg4cFx7g2OL7TjxKq3NshZ1PpzEmHcg
AbCyrrOEfPp02CL43Q1MUsZZrKkJo47zo3vm93XPfENS3v/j96x5uaL3Gt71rp00
TVKkocjpz5SwIPNrm00kbBtJzGnIw9mhr89Iaw2sDeH7pSZQNDOUo+ENs6jkwA28
kOxFdt6tyGaIic19mnvnHdVM0uHja/DUSA/6lEG/eKxL5skcXw9R3Hwmw/7yyW1R
V0ZDVe09juer95it7R1UNWApKXbg8XYI+4F520tOsLf0Xjvp5K5DNms4lTXy2yHr
xGVLhdqUboy97vPNT/SW31fntwny0KpOwDnEMPzEbQIDAQABo1MwUTAdBgNVHQ4E
FgQUDlsfflhbRtoK4JN/af0hbQZ7HO0wHwYDVR0jBBgwFoAUDlsfflhbRtoK4JN/
af0hbQZ7HO0wDwYDVR0TAQH/BAUwAwEB/zANBgkqhkiG9w0BAQsFAAOCAQEAWwUo
OJA+aw0C9IqCTRWTeqjs4QDuC5oyBui/+nrSlCtYvDFWsjYqtvBWMkxRFUU9SQlR
8NOm7U/ly0+vLsO6onlye05KaSpX1a/2XjF3uwGBNUInoPxfUEeYJmvKZu9ch5pT
8XJb4cRAtKKkhWVdh618QEMPOfhaDc1QuU+AO9vfTQg9liIvVIYF7PDm8pwYhtXp
Skq5V4YPiM8ESplDfOwUVcf0uu1MhBVzmGA/Zwp7ReVtSe1FdUMNldTbYISJIc3s
Nsf62cmm6yh4twg5c0VYiW/HGPsR8/u/PdxjFLx0LeMdVMNGKrIgX74NaT4MolOU
88EAgVz/eYlJYKKIMw==
-----END CERTIFICATE-----
```
4. Run `electrum --server bitcoin.lukechilds.co:50002:s`.

I need to distribute both my connection string and my SSL certificate, the user also needs to understand how to correctly locate their Electrum config dir and create the files. Not trivial for people who aren't too tech savvy.

Supporting fingerprints would give the same security via just:

1. Run `electrum --server bitcoin.lukechilds.co:50002:s:ED1FE28B21A5943A8810747B2A6DDFC0E3089115D6205125E5C4D3EE8E126F80`.

lukechilds:
Resolves #6084 

Needs a bit more work but it's fully functional and testable now.

If you want to test against my server you can run:

```
$ ./run_electrum --oneserver --server bitcoin.lukechilds.co:50002:s --serverfingerprint ED1FE28B21A5943A8810747B2A6DDFC0E3089115D6205125E5C4D3EE8E126F80
```

This should connect as normal.

If you run:

```
$ ./run_electrum --oneserver --server bitcoin.lukechilds.co:50002:s --serverfingerprint DEADBEEF
```

Then an `ErrorSSLCertFingerprintMismatch` error should be thrown.

Two questions:

1. `ErrorSSLCertFingerprintMismatch` is being caught somewhere so it doesn't kill the app, the connection keeps being retried. It's not entirely clear to me where this is getting handled. Instead I want the error handler to exit the application if it encounters an `ErrorSSLCertFingerprintMismatch`. Where should I add this logic?

2. I need to display an error dialog to GUI users before exit, informing them of the fingerprint mismatch. Is it ok to do Qt stuff directly inside `interface.py`? It seems like maybe that's not the right place. What's the best way to structure that?

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1263,7 +1263,8 @@ argparse._SubParsersAction.__call__ = subparser_call
 
 
 def add_network_options(parser):
-    parser.add_argument("-f", "--serverfingerprint", dest="serverfingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint")
+    parser.add_argument("-f", "--serverfingerprint", dest="serverfingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint." + " " +
+                                                                                                  "To calculate this yourself: '$ openssl x509 -noout -fingerprint -sha256 -inform pem -in mycertfile.crt'. Enter as 64 hex chars.")
     parser.add_argument("-1", "--oneserver", action="store_true", dest="oneserver", default=None, help="connect to one server only")
     parser.add_argument("-s", "--server", dest="server", default=None, help="set server host:port:protocol, where protocol is either t (tcp) or s (ssl)")
     parser.add_argument("-p", "--proxy", dest="proxy", default=None, help="set proxy [type:]host[:port] (or 'none' to disable proxy), where type is socks4,socks5 or http")
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1263,7 +1263,7 @@ argparse._SubParsersAction.__call__ = subparser_call
 
 
 def add_network_options(parser):
-    parser.add_argument("-f", "--fingerprint", dest="fingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint")
+    parser.add_argument("-f", "--serverfingerprint", dest="serverfingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint")
     parser.add_argument("-1", "--oneserver", action="store_true", dest="oneserver", default=None, help="connect to one server only")
     parser.add_argument("-s", "--server", dest="server", default=None, help="set server host:port:protocol, where protocol is either t (tcp) or s (ssl)")
     parser.add_argument("-p", "--proxy", dest="proxy", default=None, help="set proxy [type:]host[:port] (or 'none' to disable proxy), where type is socks4,socks5 or http")
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1263,6 +1263,7 @@ argparse._SubParsersAction.__call__ = subparser_call
 
 
 def add_network_options(parser):
+    parser.add_argument("-f", "--fingerprint", dest="fingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint")
     parser.add_argument("-1", "--oneserver", action="store_true", dest="oneserver", default=None, help="connect to one server only")
     parser.add_argument("-s", "--server", dest="server", default=None, help="set server host:port:protocol, where protocol is either t (tcp) or s (ssl)")
     parser.add_argument("-p", "--proxy", dest="proxy", default=None, help="set proxy [type:]host[:port] (or 'none' to disable proxy), where type is socks4,socks5 or http")
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -183,6 +183,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
         self.checking_accounts = False
         self.qr_window = None
         self.pluginsdialog = None
+        self.showing_cert_mismatch_error = False
         self.tl_windows = []
         Logger.__init__(self)
 
@@ -267,7 +268,8 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
                          'banner', 'verified', 'fee', 'fee_histogram', 'on_quotes',
                          'on_history', 'channel', 'channels_updated',
                          'payment_failed', 'payment_succeeded',
-                         'invoice_status', 'request_status', 'ln_gossip_sync_progress']
+                         'invoice_status', 'request_status', 'ln_gossip_sync_progress',
+                         'cert_mismatch']
             # To avoid leaking references to "self" that prevent the
             # window from being GC-ed when closed, callbacks should be
             # methods of this class only, and specifically not be
@@ -442,6 +444,8 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
             self.history_model.on_fee_histogram()
         elif event == 'ln_gossip_sync_progress':
             self.update_lightning_icon()
+        elif event == 'cert_mismatch':
+            self.show_cert_mismatch_error()
         else:
             self.logger.info(f"unexpected network event: {event} {args}")
 
@@ -3118,3 +3122,14 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
                      "to see it, you need to broadcast it."))
             win.msg_box(QPixmap(icon_path("offline_tx.png")), None, _('Success'), msg)
             return True
+
+    def show_cert_mismatch_error(self):
+        if self.showing_cert_mismatch_error:
+            return
+        self.showing_cert_mismatch_error = True
+        self.show_critical(title=_("Certificate mismatch"),
+                           msg="\n\n" +
+                               _("The SSL certificate provided by the main server did not match the fingerprint passed in with the --serverfingerprint option.") + "\n\n" +
+                               _("Electrum will now exit."))
+        self.showing_cert_mismatch_error = False
+        self.close()
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -3128,8 +3128,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
             return
         self.showing_cert_mismatch_error = True
         self.show_critical(title=_("Certificate mismatch"),
-                           msg="\n\n" +
-                               _("The SSL certificate provided by the main server did not match the fingerprint passed in with the --serverfingerprint option.") + "\n\n" +
+                           msg=_("The SSL certificate provided by the main server did not match the fingerprint passed in with the --serverfingerprint option.") + "\n\n" +
                                _("Electrum will now exit."))
         self.showing_cert_mismatch_error = False
         self.close()
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -508,6 +508,7 @@ class Interface(Logger):
         fingerprint = hashlib.sha256(certificate).hexdigest()
         fingerprints_match = fingerprint.lower() == expected_fingerprint.lower()
         if not fingerprints_match:
+            util.trigger_callback('cert_mismatch')
             raise ErrorSSLCertFingerprintMismatch('Refusing to connect to server due to cert fingerprint mismatch')
         self.logger.info("cert fingerprint verification passed")
 
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -429,7 +429,9 @@ class Interface(Logger):
     @handle_disconnect
     async def run(self):
         expected_fingerprint = self.network.config.get("serverfingerprint")
-        if expected_fingerprint and not await self.verify_server_certificate(expected_fingerprint):
+        if (expected_fingerprint
+                and self.is_main_server()
+                and not await self.verify_server_certificate(expected_fingerprint)):
             raise ErrorSSLCertFingerprintMismatch('Refusing to connect to server due to SSL certificate fingerprint mismatch')
         try:
             ssl_context = await self._get_ssl_context()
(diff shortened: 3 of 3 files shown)


ROSAISS81:
rosaiselaperez@gmail.com rosa Isela perez martinez
te  soy sincera no se lo  puedes hacer   tu te lo agradecería tanto y una
pregunta  paraqué es o que es esto no lo se en verdad lo siento pero es la
verdad

electrum/commands.py
```diff
@@ -1258,6 +1258,7 @@ def subparser_call(self, parser, namespace, values, option_string=None):
 
 
 def add_network_options(parser):
+    parser.add_argument("-f", "--fingerprint", dest="fingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint")
```
SomberNight, line 1261: `fingerprint` sounds a bit too generic, it might become ambiguous in the future.
maybe `sslfingerprint` or `serverfingerprint` or `sslcerthash`? well not sure.

electrum/interface.py
```diff
@@ -404,6 +406,9 @@ async def wrapper_func(self: 'Interface', *args, **kwargs):
     @log_exceptions
     @handle_disconnect
     async def run(self):
+        expected_fingerprint = self.network.config.get("fingerprint")
+        if expected_fingerprint and not await self.verify_server_certificate(expected_fingerprint):
+            raise ErrorSSLCertFingerprintMismatch('Refusing to connect to server due to SSL certificate fingerprint mismatch')
```
SomberNight, line 411: if `self.protocol` is not "s", you should not run ssl-related code, e.g. `self.get_certificate()`.
Instead, if protocol is not "s" but fingerprint is provided, you could just fail with a distinct error.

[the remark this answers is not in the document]
lukechilds: It wasn't deliberate, this should be limited to just the main server.

electrum/interface.py
```diff
@@ -471,6 +476,12 @@ async def get_certificate(self):
         except ValueError:
             return None
 
+    async def verify_server_certificate(self, expected_fingerprint):
+        certificate = await self.get_certificate()
```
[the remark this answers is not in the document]
lukechilds: What would you suggest here?

Add a simple cache to `get_certificate()` so it only fetches the cert the first time and returns it in future calls?

electrum/commands.py
lukechilds: I agree, I like `serverfingerprint`.

electrum/interface.py
lukechilds: Implemented the check in https://github.com/spesmilo/electrum/pull/6094/commits/6d7698ec0b5d9078fb8c2e0c9dd5c334925e7533 but having the same issue I mentioned in the original comment with the error being caught somewhere instead of causing the application to exit.

electrum/interface.py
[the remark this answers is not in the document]
lukechilds: Oh sorry, I think misunderstood what you meant. This problem is mitigated by  https://github.com/spesmilo/electrum/pull/6094/commits/e2a8b347dc760693ced2ff0c6923eb4fa06b198d, right?

electrum/interface.py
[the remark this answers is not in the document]
lukechilds: Resolved by https://github.com/spesmilo/electrum/pull/6094/commits/e2a8b347dc760693ced2ff0c6923eb4fa06b198d

electrum/commands.py
lukechilds: Resolved by https://github.com/spesmilo/electrum/pull/6094/commits/0fcde1306197927de3793326aba840df2ad7c80a

SomberNight:
> I need to display an error dialog to GUI users before exit, informing them of the fingerprint mismatch. Is it ok to do Qt stuff directly inside interface.py? It seems like maybe that's not the right place. What's the best way to structure that?

Qt GUI elements must only be created/modified from the GUI thread.
This code in interface all runs on the asyncio event loop, which runs on a separate thread.
Search for e.g. `util.trigger_callback('blockchain_updated')`, and see
https://github.com/spesmilo/electrum/blob/cd199390e2d6376cdf9458c4a14906c564beea1c/electrum/gui/qt/main_window.py#L265
https://github.com/spesmilo/electrum/blob/cd199390e2d6376cdf9458c4a14906c564beea1c/electrum/gui/qt/main_window.py#L407

> ErrorSSLCertFingerprintMismatch is being caught somewhere so it doesn't kill the app, the connection keeps being retried. It's not entirely clear to me where this is getting handled. Instead I want the error handler to exit the application if it encounters an ErrorSSLCertFingerprintMismatch. Where should I add this logic?

I am not sure if we should exit the application.
The exception is getting caught btw in the decorators of `interface.run`.

Maybe instead of exiting, it's good enough to notify the user and keep retrying.
Anyone who reads logs will see the issue; so we should just make sure GUI users get some feedback too.
Also note that if you display some feedback in the GUI and you also want to kill the application, you should make sure the user has time to read the feedback.
Another point is how to show feedback to GUI users: if we show a popup, we should probably kill the application, as spamming popups would be horrible (if we keep retrying). Alternatively if we don't kill the application, we could maybe show OS notifications (although even spamming those is kinda bad).

So, maybe what we could do is:
- for Qt GUI users, show popup, wait for user to click OK, and then quit app
- for Kivy GUI users, nothing atm, as they are not able to use this feature for now
- others, we assume they read logs, and we just keep retrying

electrum/interface.py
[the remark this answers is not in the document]
SomberNight: Well it's somewhat mitigated, yes. However it's still a mildly annoying that we do extra network round trips for the main server if the feature is enabled. Still, as it only happens if a fingerprint is provided, it's fine.
--- But, what's not fine is that due to the implementation, and this extra round trip, these checks do not actually guarantee any extra security. :P
The evil ISP can just look at the code, and realise it's enough to pass through the real certificate on the first round-trip, which we check, and then present their own evil cert (which will get pinned) on the second round-trip.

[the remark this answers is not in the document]
SomberNight: So, my suggestion is this:
- for simplicity, disallow CA-signed certs when a fingerprint is provided (raise distinct error)
- change the API of `verify_server_certificate` to take the cert itself as input
- call `verify_server_certificate` from `interface._is_saved_ssl_cert_available`, and also from `interface.save_certificate` (make sure the latter does not save the cert if the fingerprint mismatches, ofc)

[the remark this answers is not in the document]
lukechilds: Ah yes, my intention was to validate the actual certificate used in the current connection, not make a specific request to get one. Doh!

To allow this to work with CA certs could I not just use [`getpeercert`](https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.getpeercert) on the actual socket that is being used for the main connection, and then check the fingerprint against that?

That way it'll just be one check in one place for all scenarios. CA/self signed/initial connection/future pinned connection, it'll all be checked the same way.

lukechilds:
Thanks for clarifying that, that makes a lot of sense!

It's late here now but I'll crack on with this tomorrow.

electrum/interface.py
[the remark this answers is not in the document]
SomberNight: You could do that, but with how the code is structured now (on master), by the time you have the actual socket, some form of cert pinning has already happened. (i.e. you might end up having pinned the wrong cert)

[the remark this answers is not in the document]
lukechilds: This should be resolved in https://github.com/spesmilo/electrum/pull/6094/commits/54a26d6d595c117c121094c1083641bae581e410 and https://github.com/spesmilo/electrum/pull/6094/commits/416a41e9bf420239944a5882d40ff8830be52409

electrum/interface.py
lukechilds: This is now resolved after refactoring the fingerprint check to be next to the other certificate related logic.

lukechilds:
I think I've addressed all the points in your feedback.

This is what a GUI user sees if they try to connect to a server with an invalid fingerprint:

Screenshot 2020-04-19 at 14 11 35

Let me know if the wording is ok or if you want me to change it.

electrum/gui/qt/main_window.py
```diff
@@ -3118,3 +3122,14 @@ def save_transaction_into_wallet(self, tx: Transaction):
                      "to see it, you need to broadcast it."))
             win.msg_box(QPixmap(icon_path("offline_tx.png")), None, _('Success'), msg)
             return True
+
+    def show_cert_mismatch_error(self):
+        if self.showing_cert_mismatch_error:
+            return
+        self.showing_cert_mismatch_error = True
+        self.show_critical(title=_("Certificate mismatch"),
+                           msg="\n\n" +
```
SomberNight, line 3131: Why the leading newlines? It looks a bit weird on both Ubuntu and Windows.
I see from your picture that on macOS it does not show the title for the popup -- weird. In any case I think we should rm these newlines, at least for uniformity with other popups.

electrum/interface.py
```diff
@@ -350,6 +353,8 @@ async def is_server_ca_signed(self, ca_ssl_context):
     async def _try_saving_ssl_cert_for_first_time(self, ca_ssl_context):
         ca_signed = await self.is_server_ca_signed(ca_ssl_context)
         if ca_signed:
+            if self.network.config.get("serverfingerprint"):
+                raise InvalidOptionCombination("cannot use --serverfingerprint with CA signed servers")
```
SomberNight, line 357: This should also test for `is_main_server`.

electrum/interface.py
```diff
@@ -362,6 +367,8 @@ def _is_saved_ssl_cert_available(self):
         with open(self.cert_path, 'r') as f:
             contents = f.read()
         if contents == '':  # CA signed
+            if self.network.config.get("serverfingerprint"):
+                raise InvalidOptionCombination("cannot use --serverfingerprint with CA signed servers")
```
SomberNight, line 371: As above, this should also test for `is_main_server`.

How about introducing a new method, `interface._get_expected_fingerprint` or similar, which returns
`self.network.config.get("serverfingerprint")` for the main server, and None otherwise? That way, the `is_main_server` tests don't have to be duplicated.

electrum/commands.py
```diff
@@ -1263,6 +1263,7 @@ def subparser_call(self, parser, namespace, values, option_string=None):
 
 
 def add_network_options(parser):
+    parser.add_argument("-f", "--serverfingerprint", dest="serverfingerprint", default=None, help="only allow connecting to servers with a matching SSL certificate SHA256 fingerprint")
```
SomberNight, line 1266: Maybe we could also say here how to calc the fingerprint for yourself.
`To calculate this yourself: '$ openssl x509 -noout -fingerprint -sha256 -inform pem -in mycertfile.crt'. Enter as 64 hex chars.`

electrum/gui/qt/main_window.py
lukechilds: Ah, I was wondering why that didn't show!

The leading newlines were original because I had "Certificate mismatch" at the beginning of `msg` because it wasn't showing in `title`. When I removed the duplicate "Certificate mismatch" I must've forgotten to remove the leading newlines along with it.

lukechilds:
Latest feedback addressed, let me know if it's ok.

lukechilds:
@SomberNight just pinging you to make sure you've seen this. If you have and you just haven't gotten around to it yet, no worries!

PR #6106, opened 2020-04-21
set more restrictive file permissions for exported private keys

gch1p:
When you export private keys to a file (electrum-private-keys.csv or .json), it's created with default file permissions (644 on my system), which is not very secure for private keys I think. Let's use 600 instead.


SomberNight:
thanks

CyberCreator:
Good idea

Issue #6127, opened 2020-04-29
ln funding tx should not be allowed to be fee-bumped or coinjoined


SomberNight:
LN funding tx should not be allowed to be coinjoined (note: also "bump fee")
(e.g. xpub-only wallet)

  - maybe this is already solved via "psbt_only_widgets" ?
    - or maybe not, as for xpub-only wallet, funding tx will be visible while partial
  - note that from "Combine" dropdown, "merge sigs from" is useful... (but not "join")

-----

All changes to a funding tx between sending its txid to the remote node, and broadcasting it (well, getting it mined), that affects the txid are dangerous and should not be allowed.

ecdsa:
what is the scenario where "bump fee" would become relevant?

SomberNight:
I guess "bump fee" is already not reachable for a funding tx, though I think it's sort of accidental.

E.g. if we allowed fee-bumping non-RBF "local" transactions, which to me sounds like a change we might decide to make for whatever reason, we might forgot to consider this case (funding tx).
https://github.com/spesmilo/electrum/blob/6f7a4ab048bbdb4215ae47f206a878aa6215bf31/electrum/wallet.py#L528-L531

But re adding new inputs/outputs, one possible flow to trigger that:
- use watch-only wallet, with LN enabled
- in that watch-only wallet:
  - start opening a channel
  - tx now appears as "local" tx in history
  - in tx details, Combine>"Join inputs/outputs"

ecdsa:
One solution would be to disable Coinjoin for transactions that are not RBF. 
That would be consistent with the assumption that 'final' transaction are not malleable.

PR #6134, opened 2020-04-30
wallet: disallow fee-bumping/coinjoining ln funding tx

SomberNight:
related: #6127

This PR makes it explicit that funding txns cannot be fee-bumped.
It also closes the add inputs/outputs flow described in https://github.com/spesmilo/electrum/issues/6127#issuecomment-622070937. i.e. we disallow joining a partial tx that we know is a funding tx of ours

It does not seem possible to completely close down all footguns re changing the txid of a funding tx via adding inputs/outputs, as e.g. the user could
- create tx1 in wallet1, which is a funding tx for a channel open
- create tx2 in wallet2
- join them together in wallet2, through e.g. the "Load transaction" menu
- and then broadcast; resulting in most likely losing the coins intended for the channel open

### DIFF
--- a/electrum/gui/qt/transaction_dialog.py
+++ b/electrum/gui/qt/transaction_dialog.py
@@ -170,9 +170,9 @@ class BaseTxDialog(QDialog, MessageBoxMixin):
         ptx_merge_sigs_action = QAction(_("Merge signatures from"), self)
         ptx_merge_sigs_action.triggered.connect(self.merge_sigs)
         partial_tx_actions_menu.addAction(ptx_merge_sigs_action)
-        ptx_join_txs_action = QAction(_("Join inputs/outputs"), self)
-        ptx_join_txs_action.triggered.connect(self.join_tx_with_another)
-        partial_tx_actions_menu.addAction(ptx_join_txs_action)
+        self._ptx_join_txs_action = QAction(_("Join inputs/outputs"), self)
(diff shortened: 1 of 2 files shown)


ecdsa:
see my other comment: instead of adding a new special case, why not disable coinjoins on 'final' transactions?

SomberNight:
> instead of adding a new special case, why not disable coinjoins on 'final' transactions?

- In my view that would also add a new special case, except it would be implicit instead of explicit.
We would still need to remember the case of "LN funding tx" (in case of future changes to related logic) whether it's explicitly listed there or not, so I think it might be better to explicitly list it.
- a user might want to create a non-RBF coinjoined tx
- "opt-in RBF" is not even a proper flag: e.g. if you want OP_CSV, you need RBF enabled (because of how consensus interprets nSequence)
- unsure if "full RBF" ever becomes a thing in Bitcoin but it keeps coming up every ~year (this is the idea that all mempool txs should become replaceable)

Because of the first point, if we did what you suggest, I think at the least we should add a comment in the code about this assumption, to make it more explicit.

ecdsa:
The only assumption I made is that a transaction that is marked "final" should actually be final. I do not think that it is unreasonable, nor that it is highly implicit. 

If you insist that we need to allow users to perform coinjoins on transactions are marked as "final (which is questionable), at least the added boolean should have a more generic name than "is_lightning_funding_tx".  ("is_final" would make sense, but it is alrredy taken...)

Issue #6145, opened 2020-05-06
Install Wizard Behavior: Back not possible in step "Master Public Key" of Multi-signature wallet


relativisticelectron:
Kivy Wallet creation (current master):

1. Multi-signature wallet
2. Next
3. I already have a seed
4. Next

On this screen showing the Master public screen again, the back button is greyed out.
![Screenshot](https://user-images.githubusercontent.com/60378539/81171411-a25e9880-8f9c-11ea-959d-dcea2401e9be.png)

The same behavior is also in qt.  In qt there is only the cancel button, but not the back button.

relativisticelectron:
I found out that this is caused by resetting the stack: https://github.com/spesmilo/electrum/blob/master/electrum/base_wizard.py#L552

relativisticelectron:
I have a first version to fix this issue in: https://github.com/relativisticelectron/electrum/commit/18a9d7101ba2d480c9cced2aef1f5ee50a564b6b

PR #6148, opened 2020-05-06
Allow going "back" in the installwizard when in show_xpub_dialog

relativisticelectron:
Problem was described in https://github.com/spesmilo/electrum/issues/6145

The order of execution (before) was:
1. self.keystores.append(k)  (Because this was done first it prevented going back)
2. show_xpub_and_add_cosigners
3. choose_keystore  (called from inside show_xpub_and_add_cosigners)

Now it is:
1. show_xpub_and_add_cosigners 
2. self.keystores.append(k)
3. choose_keystore 

The neat thing is that the code of on_keystore is now actually more simple, because it removes calling choose_keystore from inside show_xpub_and_add_cosigners.

Additionally it also fixes a small bug:
- Previously: self.reset_stack() was only called if   len(self.keystores) == 1 . Therefore allows going back for all len(self.keystores) > 2. 
- Now: self.reset_stack()  is called regardless of  len(self.keystores), and therefore ensures one cannot go back after self.keystores.append(k)  is executed.

Tested on qt testnet (linux).

### DIFF
--- a/electrum/base_wizard.py
+++ b/electrum/base_wizard.py
@@ -549,7 +549,7 @@ class BaseWizard(Logger):
             if len(self.keystores) == 0:
                 xpub = k.get_master_public_key()
                 self.run('show_xpub_and_add_cosigners', xpub)
-                # self.reset_stack()
+                self.reset_stack()
             self.keystores.append(k)
             if len(self.keystores) < self.n:
                 self.run('choose_keystore')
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks, I merged it manually

relativisticelectron:
Thanks for the fast review and merge.

PR #6197, opened 2020-06-03
Fix win32com module error

wakiyamap:
[This commit](https://github.com/spesmilo/electrum/commit/e5b1596b69a1c77b3d8d2b235b65e624a2383921) seems to be producing the following errors.
https://travis-ci.org/github/spesmilo/electrum/jobs/693594640#L5285
```
34572 INFO: Processing pre-safe import module hook   win32com
  File "<string>", line 2, in <module>
ModuleNotFoundError: No module named 'win32com'
34810 INFO: Processing pre-safe import module hook   win32com
  File "<string>", line 2, in <module>
ModuleNotFoundError: No module named 'win32com'
```

This error is gone.
But, I couldn't determine if this is necessary for the operation of the electrum itself.


SomberNight:
This `ModuleNotFoundError: No module named 'win32com'` error seems to also be present in builds prior to that commit, e.g. see https://travis-ci.org/github/spesmilo/electrum/jobs/671727874#L5457 (for https://github.com/spesmilo/electrum/commit/55d0a9587ec1dbadbecd6b97662bda0a84cee565).

Note that if we want to install `pywin32`, it should be included in `contrib/requirements/requirements-wine-build.txt` (from which `contrib/deterministic-build/requirements-wine-build.txt` is built).

SomberNight:
related https://github.com/pyinstaller/pyinstaller/issues/3733

It is unclear to me whether the "error" can be ignored.

wakiyamap:
Was there an older commit? I seem to have missed it. Sorry.

I think it's better to write it(win32com) in `contrib/deterministic-build/requirements-wine-build.txt`.
Because it seems to generate the "error" if it write only `contrib/requirements/requirements-wine-build.txt` .

SomberNight:
Yes, we would need it in both files.

The files in `contrib/deterministic-build/*`
are generated from `contrib/requirements/*` via the
`contrib/freeze_packages.sh` script :)

wakiyamap:
I can't install `pywin32` on linux, because `pywin32` doesn't have a package for linux.
So.....I was building Windows for test. Sorry for late.

The following error occurs when running it on Linux, so [this](https://github.com/spesmilo/electrum/pull/6197/commits/69cd1662755c5f6e584c2ea90b00206c4a9b5d04#diff-11605b2f5f6020371d4acdcc0f8c4bdcR8) using windows only.
```
ERROR: Could not find a version that satisfies the requirement pywin32 (from versions: none)
ERROR: No matching distribution found for pywin32
```

PR #6217, opened 2020-06-07
tx dialog: show various warnings if input amounts cannot be verified

### LINKED ISSUE #5749: Warning on signing non-segwit input

SomberNight:
closes #5749

These warning texts are really hard to write...

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/tx_dialog.py
+++ b/electrum/gui/kivy/uix/dialogs/tx_dialog.py
@@ -132,9 +132,10 @@ class TxDialog(Factory.Popup):
         self.tx = tx  # type: Transaction
         self._action_button_fn = lambda btn: None
 
-        # if the wallet can populate the inputs with more info, do it now.
-        # as a result, e.g. we might learn an imported address tx is segwit,
-        # or that a beyond-gap-limit address is is_mine
+        # If the wallet can populate the inputs with more info, do it now.
+        # As a result, e.g. we might learn an imported address tx is segwit,
+        # or that a beyond-gap-limit address is is_mine.
+        # note: this might fetch prev txs over the network.
         tx.add_info_from_wallet(self.wallet)
 
     def on_open(self):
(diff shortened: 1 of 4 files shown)


andronoob:
>These warning texts are really hard to write...

I'm afraid to say that... the user would probably have a headache as well.

I still think such compromise is just a temporary expedient, until real solution (standardized multi-part/fountain code QR code for transferring large PSBT, which allows feeding/requiring full previus tx data of all inputs) comes out.

ecdsa:
> These warning texts are really hard to write...

Maybe the third message could be improved:

> "Warning: The fee could not be verified. All inputs are segwit, so as long as you only sign this transaction once, you are safe."

This is a bit misleading. First, the issue is not that the fee could not be verified. As a matter of fact, the fee was verified; the issue is that this verification will break apart if we sign it more than once. Second, it does not make things clearer to state that "all inputs are segwit", because segwit is already the best input type we have. (I would mention it only when we have taproot, to encourage switching to taproot).

May I suggest this instead:
> "Warning: If you received this transaction from an untrusted device, do not accept to sign it more than once."

SomberNight:
> "Warning: If you received this transaction from an untrusted device, do not accept to sign it more than once."

It's not clear from that what could happen though. It's the fee where the user could end up overpaying.
So maybe append "Otherwise you could up paying a different fee."

ecdsa:
sorry, abac4a4 has a wrong commit message

Issue #4700, opened 2018-09-09
Feature request: Wallet->Information should show HD path


Talkless:
Currently (3.1.3 in Debian Sid), Wallet->Information does not display BIP32 HD derivation path. 

User might have multiple Electrum wallets (HD accounts) using same master seed generated in hardware device (like Ledger Nano S), named `BIP84_short_term_spending` for `m/84'/0'/0'`, or `legacy_hodl` for `m/44'/0'/3`, etc, etc, but there's now way to know which path was used for previously created Electrum wallet, so re-creating that wallet in another computer (or application) is difficult (if HW wallet device is available, but Electrum wallet files are not), as user has to guess, manually document, or remember these paths used for creating that HW-based wallet.

If BIP32 derivation path would be available in Electrum GUI, and if user would know in advance that it will need to "recreate" wallet in another computer (or another application), he could just check HD path in `Wallet->Information` before proceeding.

Talkless:
I guess I could have named wallet files accordingly to paths, but, well, now it's too late :) . Again, this is manual documentation which maybe could be avoided.

ecdsa:
this is a design choice; electrum does not know about your wallet derivation path, it does not keep that information.

Talkless:
I see. I guess I'll try to recall paths and rename wallet files accordingly.

Talkless:
@ecdsa maybe possibility to add arbitrary note to a wallet file would be reasonable?

SomberNight:
It's stored in the keystore, and persisted in the wallet file for hw wallets.
https://github.com/spesmilo/electrum/blob/55b582511e8d1b6d229836a120572ff9375b891b/electrum/base_wizard.py#L353-L360
We might as well do it for generic bip39 seeds too.

Talkless:
> and persisted in the wallet file

I am lost now. @ecdsa said Electrum does not keep that.

SomberNight:
Yes, and I just said it does for hardware wallets. I had to check myself to be sure :P
It's hard to keep track..
For software bip39 seeds, it is not kept. We might as well keep it for those too I think.

Talkless:
OK so my feature requests still holds, at least for HW wallets :) .

SomberNight:
Note: in Qt console:
for non-multisig, or first cosigner in case of multisig:
```
wallet.keystore.derivation
```
multisig:
```
[k.derivation for k in wallet.get_keystores()]
```
[...]

Issue #1300, opened 2015-06-17
making seeds non-BIP39 compatible 


simcity4242:
@ecdsa I propose that all Electrum 2.x mnemonic seeds are generated such that the 12 word phrases generated are NOT BIP39 compatible. It would clearly delineate Electrum 2.x from BIP39.
@gurnec provides the [loss in entropy as <0.05 bits](http://www.wolframalpha.com/input/?i=solve++%282%5En+-+1%29+%2F+2%5En+%3D+2047%2F32768+%C3%97+1%2F16++for+n) at the [SE discussion](http://bitcoin.stackexchange.com/a/38027/9382)

ecdsa:
I believe 12-word seeds are more user friendly than 13-word seeds.
We currently have 13 words seeds becaus of the size of the wordlist.
I plan to increase the size of the wordlist in order to get back to 12 words seeds without loss of entropy.

gurnec:
In order to guarantee that current Electrum 2.x seeds are not BIP-39 valid, Electrum would have to (now and forever more) ship with the BIP-39 word lists. Avoiding being tied to / having to maintain any particular word list was I think one of the reasons behind Electrum 2.x's choices....

As an alternative, if a new longer (at least 2581 words) list is created for 12-word seeds, it could be created with no shared words with BIP-39. Of course, generating a decent such list is non-trivial....

ecdsa:
A better solution is for other wallets to stop using BIP39.
BIP39 is a failed design, and this is not an Electrum issue.

ecdsa:
OTOH if you insist on using BIP39, it is trivial to ensure that your seeds are non-electrum compatible, you don't need a wordlist for that; you just need a hash.

gurnec:
> A better solution is for other wallets to stop using BIP39. [...] this is not an Electrum issue.

FWIW I agree.

I think the only bullet-proof solution is for @simcity4242 to "bite the bullet" and do a `is_electrum2_mnemonic()`, `is_bip39_mnemonic()`, etc. and force the user to select a type if an ambiguity is found (and likewise for ambiguities between BIP-39 word lists).

> BIP39 is a failed design,

I don't want to start an argument... I'd just say "inferior" IMHO.

simcity4242:
>A better solution is for other wallets to stop using BIP39.

Agreed, to a certain extent. I'm just trying to be pragmatic about standards. From an outsider perspective, a mnemonic phrase should work without having to know: 

[...]

PR #6219, opened 2020-06-08
Automated BIP39 Recovery

### LINKED ISSUE #6155: Feature/Plugin request: Mnemonic recovery UI with derivation path "scanning"
I see many newbies having trouble with recovery of BIP39 mnemonics when they don't know the mnemonic path, especially if created by a wallet that used a non-standard path. I am motivated to help them, while steering them away from dangerous alternative choices such as web-based seed derivation tools or giving their seeds to strangers in return for promises of help (both of which are often proposed on various forums). While this can be solved with a full node and a rescan, such a solution is not useful for newbies as it is too complex. A GUI-based self-hosted multi-platform solution like Electrum, that already has access to indexed UTXO scripts seems to fit perfectly with the use case and be more secure than alternatives. 

I would like to fund a plugin that allows Electrum to import a BIP39 mnemonic such that it "scans" several possible derivation paths to find ones that have UTXO history on at least one address. The plugin would then offer a drop down of "discovered" active derivation paths and allow the user to select the one to import into a wallet file, using that derivation path to follow the normal BIP39 derivation procedure. See https://walletsrecovery.org/ for a good list of how varied the derivation paths in different wallet implementations are. 

I am considering funding this with a gitcoin grant, to encourage someone with experience to implement this, as well as attract more funding from others and more developers to Electrum. 

At this stage, I'd like to gauge interest in the idea and see if there are any principled objections to doing this. I don't want to start any drama accidentally if this is an undesirable use-case or violates some kind of principle held by Electrum developers. I would like to fund this only if it is likely to succeed and be incorporated into a future release.

lukechilds:
Resolves #6155 

Implements automated scanning of all known BIP32 chains that can be imported into Electrum.

I'm opening this PR for review, it's not ready to be merged yet. Account discovery is functional but still needs to be integrated in to the GUI.

Currently it's implemented as a script for easy testing, you can quickly test it out by running the `electrum/scripts/bip39_recovery.py` script with a `mnemonic` parameter and also an optional second `passphrase` parameter. It will return a list of any previously used accounts it can find. You can use this mnemonic that I've created a few test accounts in:

```
much bottom such hurt hunt welcome cushion erosion pulse admit name deer
```

e.g:

```
$ ./bip39_recovery.py "much bottom such hurt hunt welcome cushion erosion pulse admit name deer"
[
    {
        "derivation_path": "m/84'/0'/0'",
        "description": "Standard BIP84 native segwit (Account 0)",
        "script_type": "p2wpkh"
    },
    {
        "derivation_path": "m/84'/0'/1'",
        "description": "Standard BIP84 native segwit (Account 1)",
        "script_type": "p2wpkh"
    },
    {
        "derivation_path": "m/0'",
        "description": "Non-standard legacy (Account 0)",
        "script_type": "p2pkh"
    },
    {
        "derivation_path": "m/84'/0'/2147483646'",
        "description": "Samourai Whirlpool post-mix",
        "script_type": "p2wpkh"
    }
]
```

**Diversions from BIP44**

Due to the fact that at discovery we only need to check which accounts has been used, not the final balance of each account, we take a few shortcuts:
- Don't scan the change chain, if a BIP44 compliant account is used it will have received a TX in one of the first 20 (BIP44 gap limit) addresses of the external chain.
- Move on to the next account as soon as we find a single address with funds. No need to keep scanning until we exhaust the gap limit.

If one of the discovered accounts is selected by the user for recovery, then Electrum will proceed to properly scan both internal external chains and exhaust the gap limit when it creates the wallet.

**Limitations**

Electrum takes an account level import path like `m/84'/0'/0'` and derives the wallet with the non hardened `is_change/address_index` sub trees. Due to this we cannot import Bitcoin Core accounts as they use the format [`m/0'/0'/address_index'`](https://github.com/bitcoin/bitcoin/issues/13302#issuecomment-391227927), notice all sub trees use hardened derivation.

All other wallets listed on walletsrecovery.org can be successfully recovered.

---

Please let me know if this code looks ok so far. Next steps will be to migrate the account discovery code from the script to `electrum.bip39_recovery` and then integrate it in to the recovery wizard GUI.

CC @ecdsa @SomberNight @aantonop

### DIFF
--- a/electrum/base_wizard.py
+++ b/electrum/base_wizard.py
@@ -34,7 +34,7 @@ from . import bitcoin
 from . import keystore
 from . import mnemonic
 from .bip32 import is_bip32_derivation, xpub_type, normalize_bip32_derivation, BIP32Node
-from .keystore import bip44_derivation, purpose48_derivation, Hardware_KeyStore, KeyStore
+from .keystore import bip44_derivation, purpose48_derivation, Hardware_KeyStore, KeyStore, bip39_to_seed
 from .wallet import (Imported_Wallet, Standard_Wallet, Multisig_Wallet,
                      wallet_types, Wallet, Abstract_Wallet)
 from .storage import WalletStorage, StorageEncryptionVersion
@@ -404,7 +404,7 @@ class BaseWizard(Logger):
         else:
             raise Exception('unknown purpose: %s' % purpose)
 
-    def derivation_and_script_type_dialog(self, f):
+    def derivation_and_script_type_dialog(self, f, get_account_xpub=None):
         message1 = _('Choose the type of addresses in your wallet.')
         message2 = ' '.join([
             _('You can override the suggested derivation path.'),
@@ -429,10 +429,10 @@ class BaseWizard(Logger):
             ]
         while True:
             try:
-                self.choice_and_line_dialog(
+                self.derivation_and_script_type_gui_specific_dialog(
                     run_next=f, title=_('Script type and Derivation path'), message1=message1,
                     message2=message2, choices=choices, test_text=is_bip32_derivation,
-                    default_choice_idx=default_choice_idx)
+                    default_choice_idx=default_choice_idx, get_account_xpub=get_account_xpub)
                 return
             except ScriptTypeNotSupported as e:
                 self.show_error(e)
@@ -492,7 +492,8 @@ class BaseWizard(Logger):
     def on_restore_seed(self, seed, is_bip39, is_ext):
         self.seed_type = 'bip39' if is_bip39 else mnemonic.seed_type(seed)
         if self.seed_type == 'bip39':
-            f = lambda passphrase: self.on_restore_bip39(seed, passphrase)
+            def f(passphrase):
+                self.on_restore_bip39(seed, passphrase)
             self.passphrase_dialog(run_next=f, is_restoring=True) if is_ext else f('')
         elif self.seed_type in ['standard', 'segwit']:
             f = lambda passphrase: self.run('create_keystore', seed, passphrase)
@@ -509,7 +510,13 @@ class BaseWizard(Logger):
         def f(derivation, script_type):
             derivation = normalize_bip32_derivation(derivation)
             self.run('on_bip43', seed, passphrase, derivation, script_type)
-        self.derivation_and_script_type_dialog(f)
+        def get_account_xpub(account_path):
+            root_seed = bip39_to_seed(seed, passphrase)
+            root_node = BIP32Node.from_rootseed(root_seed, xtype="standard")
+            account_node = root_node.subkey_at_private_derivation(account_path)
+            account_xpub = account_node.to_xpub()
+            return account_xpub
+        self.derivation_and_script_type_dialog(f, get_account_xpub)
 
     def create_keystore(self, seed, passphrase):
         k = keystore.from_seed(seed, passphrase, self.wallet_type == 'multisig')
--- a/electrum/base_wizard.py
+++ b/electrum/base_wizard.py
@@ -404,7 +404,7 @@ class BaseWizard(Logger):
         else:
             raise Exception('unknown purpose: %s' % purpose)
 
-    def derivation_and_script_type_dialog(self, f, get_account_xpub=None):
+    def derivation_and_script_type_dialog(self, f, *, get_account_xpub=None):
         message1 = _('Choose the type of addresses in your wallet.')
         message2 = ' '.join([
             _('You can override the suggested derivation path.'),
@@ -516,7 +516,7 @@ class BaseWizard(Logger):
             account_node = root_node.subkey_at_private_derivation(account_path)
             account_xpub = account_node.to_xpub()
             return account_xpub
-        self.derivation_and_script_type_dialog(f, get_account_xpub)
+        self.derivation_and_script_type_dialog(f, get_account_xpub=get_account_xpub)
 
     def create_keystore(self, seed, passphrase):
         k = keystore.from_seed(seed, passphrase, self.wallet_type == 'multisig')
--- a/electrum/bip39_recovery.py
+++ b/electrum/bip39_recovery.py
@@ -2,6 +2,8 @@
 # Distributed under the MIT software license, see the accompanying
 # file LICENCE or http://www.opensource.org/licenses/mit-license.php
 
+from typing import TYPE_CHECKING
+
 from aiorpcx import TaskGroup
 
 from . import bitcoin
@@ -10,7 +12,11 @@ from .bip32 import BIP32_PRIME, BIP32Node
 from .bip32 import convert_bip32_path_to_list_of_uint32 as bip32_str_to_ints
 from .bip32 import convert_bip32_intpath_to_strpath as bip32_ints_to_str
 
-async def account_discovery(network, get_account_xpub):
+if TYPE_CHECKING:
+    from .network import Network
+
+
+async def account_discovery(network: 'Network', get_account_xpub):
     async with TaskGroup() as group:
         account_scan_tasks = []
         for wallet_format in BIP39_WALLET_FORMATS:
@@ -21,13 +27,14 @@ async def account_discovery(network, get_account_xpub):
         active_accounts.extend(task.result())
     return active_accounts
 
-async def scan_for_active_accounts(network, get_account_xpub, wallet_format):
+
+async def scan_for_active_accounts(network: 'Network', get_account_xpub, wallet_format):
     active_accounts = []
     account_path = bip32_str_to_ints(wallet_format["derivation_path"])
     while True:
         account_xpub = get_account_xpub(account_path)
         account_node = BIP32Node.from_xkey(account_xpub)
-        has_history = await account_has_history(network, account_node, wallet_format["script_type"]);
+        has_history = await account_has_history(network, account_node, wallet_format["script_type"])
         if has_history:
             account = format_account(wallet_format, account_path)
             active_accounts.append(account)
@@ -36,7 +43,8 @@ async def scan_for_active_accounts(network, get_account_xpub, wallet_format):
         account_path[-1] = account_path[-1] + 1
     return active_accounts
 
-async def account_has_history(network, account_node, script_type):
+
+async def account_has_history(network: 'Network', account_node: BIP32Node, script_type: str) -> bool:
     gap_limit = 20
     async with TaskGroup() as group:
         get_history_tasks = []
@@ -53,6 +61,7 @@ async def account_has_history(network, account_node, script_type):
         if len(history) > 0:
             return True
     return False
+
 
 def format_account(wallet_format, account_path):
     description = wallet_format["description"]
--- a/electrum/bip39_recovery.py
+++ b/electrum/bip39_recovery.py
@@ -1,0 +1,66 @@
+# Copyright (C) 2020 The Electrum developers
+# Distributed under the MIT software license, see the accompanying
+# file LICENCE or http://www.opensource.org/licenses/mit-license.php
+
+from aiorpcx import TaskGroup
+
+from . import bitcoin
+from .constants import BIP39_WALLET_FORMATS
+from .bip32 import BIP32_PRIME, BIP32Node
+from .bip32 import convert_bip32_path_to_list_of_uint32 as bip32_str_to_ints
+from .bip32 import convert_bip32_intpath_to_strpath as bip32_ints_to_str
+
+async def account_discovery(network, get_account_xpub):
+    async with TaskGroup() as group:
+        account_scan_tasks = []
+        for wallet_format in BIP39_WALLET_FORMATS:
+            account_scan = scan_for_active_accounts(network, get_account_xpub, wallet_format)
+            account_scan_tasks.append(await group.spawn(account_scan))
+    active_accounts = []
+    for task in account_scan_tasks:
+        active_accounts.extend(task.result())
+    return active_accounts
+
+async def scan_for_active_accounts(network, get_account_xpub, wallet_format):
+    active_accounts = []
+    account_path = bip32_str_to_ints(wallet_format["derivation_path"])
+    while True:
+        account_xpub = get_account_xpub(account_path)
+        account_node = BIP32Node.from_xkey(account_xpub)
+        has_history = await account_has_history(network, account_node, wallet_format["script_type"]);
+        if has_history:
+            account = format_account(wallet_format, account_path)
+            active_accounts.append(account)
+        if not has_history or not wallet_format["iterate_accounts"]:
+            break
+        account_path[-1] = account_path[-1] + 1
+    return active_accounts
+
+async def account_has_history(network, account_node, script_type):
+    gap_limit = 20
+    async with TaskGroup() as group:
+        get_history_tasks = []
+        for address_index in range(gap_limit):
+            address_node = account_node.subkey_at_public_derivation("0/" + str(address_index))
+            pubkey = address_node.eckey.get_public_key_hex()
+            address = bitcoin.pubkey_to_address(script_type, pubkey)
+            script = bitcoin.address_to_script(address)
+            scripthash = bitcoin.script_to_scripthash(script)
+            get_history = network.get_history_for_scripthash(scripthash)
+            get_history_tasks.append(await group.spawn(get_history))
+    for task in get_history_tasks:
+        history = task.result()
+        if len(history) > 0:
+            return True
+    return False
+
+def format_account(wallet_format, account_path):
+    description = wallet_format["description"]
+    if wallet_format["iterate_accounts"]:
+        account_index = account_path[-1] % BIP32_PRIME
+        description = f'{description} (Account {account_index})'
+    return {
+        "description": description,
+        "derivation_path": bip32_ints_to_str(account_path),
+        "script_type": wallet_format["script_type"],
+    }
--- a/electrum/scripts/bip39_recovery.py
+++ b/electrum/scripts/bip39_recovery.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+import sys
+import asyncio
+
+from electrum.util import json_encode, print_msg, create_and_start_event_loop, log_exceptions
+from electrum.simple_config import SimpleConfig
+from electrum.network import Network
+from electrum.keystore import bip39_to_seed
+from electrum.bip32 import BIP32Node
+from electrum.bip39_recovery import account_discovery
+
+try:
+    mnemonic = sys.argv[1]
+    passphrase = sys.argv[2] if len(sys.argv) > 2 else ""
+except Exception:
+    print("usage: bip39_recovery <mnemonic> [<passphrase>]")
+    sys.exit(1)
+
+loop, stopping_fut, loop_thread = create_and_start_event_loop()
+
+config = SimpleConfig()
+network = Network(config)
+network.start()
+
+@log_exceptions
+async def f():
+    try:
+        def get_account_xpub(account_path):
+            root_seed = bip39_to_seed(mnemonic, passphrase)
+            root_node = BIP32Node.from_rootseed(root_seed, xtype="standard")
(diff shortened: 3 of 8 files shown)


electrum/scripts/bip39_recovery.py
```diff
@@ -0,0 +1,127 @@
...
+        scripthash = bitcoin.script_to_scripthash(script)
+        history = await network.get_history_for_scripthash(scripthash)
+        if len(history) > 0:
+            return True
+    return False
+
+def format_account(wallet_format, account_path):
+    description = wallet_format["description"]
+    if wallet_format["iterate_accounts"]:
+        account_index = account_path.split("/")[-1].replace("'", "")
```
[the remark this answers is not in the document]
lukechilds: Good point, will do.

lukechilds:
Thanks for the review @SomberNight, I have two questions re GUI integration.

**1. Whats the best way to hook into the derivation path view?**

I want to add a button in this view that says "Detect Existing Accounts" which when clicked, opens a popup with a loading icon, calls the account discovery function, displays a list of results to the user, allows them to select one, and populate the path/script type inputs with the result.

Screenshot 2020-06-08 at 21 08 46

For example the previous view to enter the seed has it's own file in `electrum/gui/qt/seed_dialog.py`, so I could just edit that. However the derivation path is implemented in a GUI diagnostic way in `electrum/base_wizard.py` which sets up some params and then inits the UI with `self.choice_and_line_dialog()` where `choice_and_line_dialog()` is a generic view provided by the GUI implementation.

https://github.com/spesmilo/electrum/blob/d0ab0039784d1d0faade548182f72aa6d47b7e82/electrum/base_wizard.py#L407-L439

I can see the QT implementation here:

https://github.com/spesmilo/electrum/blob/d0ab0039784d1d0faade548182f72aa6d47b7e82/electrum/gui/qt/installwizard.py#L605-L633

I could just edit that method and wrap my code with. `if title == "Script type and Derivation path"` but that seems like a pretty nasty solution and would also break when i18n is active.

What's the best way to hook into this?

**2. How should I provide network access to my method?**

Currently in the script I'm creating a network instance with:

```python
from electrum.simple_config import SimpleConfig
from electrum.network import Network
config = SimpleConfig()
network = Network(config)
network.start()
```

When I add this to the GUI I'll obviously want to use the existing Network instance. I see `Network` is a singleton, and I can get the instance with `Network.get_instance()`, however I'd also like to keep the script functional, by just updating it to import `electrum.bip39_recovery` and running it. It's useful to quickly test during dev and seems like a pretty handy script to expose to users too.

What's the best way to go about this? Create a helper that returns the global instance if it exists or creates one if it doesn't? Or maybe pass a `network` instance in to the function, and pass in the one I create for the script or pass in the global instance for the GUI?

Not quite sure the best way to go about this.

SomberNight:
> 2. How should I provide network access to my method?
> Or maybe pass a network instance in to the function, and pass in the one I create for the script or pass in the global instance for the GUI?

Yes, that sounds good.

> GUI I'll obviously want to use the existing network instance (if it exists during account creation?)

Yes, typically it exists, unless the user used the `--offline` CLI flag.
When you get a reference to network, you should test if it's None, and if it is, just display an error message that they are offline.

> 1. Whats the best way to hook into the derivation path view?
> the derivation path is implemented in a GUI agnostic way in electrum/base_wizard.py which sets up some params and then inits the UI with self.choice_and_line_dialog() where choice_and_line_dialog() is a generic view provided by the GUI implementation.

`choice_and_line_dialog` is already only used for this specific usecase.
I think you can just rename it to something "derivation"-specific, and modify it directly.
I imagine you only want to implement all this for the Qt GUI? (which is fine)
Make sure that the change does not break the kivy GUI.
i.e. rename the method, pass it whatever arguments you need, make sure both GUIs accept all the arguments even if they ignore some, make sure the return types are the same
Then you can just implement extra logic in the Qt-specific implementation of this method -- well you probably want to implement it somewhere else as I imagine it would be many lines of code, and just call it from there.

lukechilds:
@SomberNight thanks, makes sense!

Hadn't realised `choice_and_line_dialog` was only used for this one dialog.

And yes, I was only planning on implementing it in the QT GUI. CLI users can use the script, and I don't think the Kivy GUI allows setting the BIP39 option on a seed so not really applicable there.

electrum/scripts/bip39_recovery.py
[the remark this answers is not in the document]
lukechilds: Resolved in https://github.com/spesmilo/electrum/pull/6219/commits/9855fe2115eb341cf359fb128cc84bb242f27f32.

lukechilds:
@SomberNight when calling an async function from a QT `WindowModalDialog` I can do:

```python
accounts = asyncio.run_coroutine_threadsafe(account_discovery(), network.asyncio_loop).result()
```

Which works but it blocks the QT thread so the UI hangs and there's no way for the user click the cancel button while they're waiting.

I'd like to:

- Show the modal
- Show some text like "Scanning for existing accounts..." or maybe an animated spinner
- Fire off the account discovery to run asynchronously
- Allow the user to interact with the UI, cancel the modal if they want
- Once account discovery has completed, update the modal with the results

Are you able to point me in the right direction for how to achieve this?

**Disclaimer:** Apologies if this question seems obvious, I'm not really a Python developer, in fact the only other time I've ever written Python was for my SSL fingerprint PR. I'm not familiar with Python's async API works.

SomberNight:
> if this question seems obvious

No worries, it's not obvious at all. Qt does not really support python asyncio.

Take a look at https://github.com/spesmilo/electrum/blob/b0230f6a4bd7498aa0dd1d48ec61326d35a40386/electrum/gui/qt/main_window.py#L1658

It does not behave exactly as you described but similar.

lukechilds:
I've integrated the functionality into the GUI.

Demo:

electrum_bip39

rdymac:
The resulting file name could use another label (as the [standard] one), so the user knows which one he selected after the scan.

jlopp:
I'd highly encourage using a gap limit that far exceeds the standard. Consider the fact that there are a variety of applications such as payment processing that can result in huge gaps. If I recall correctly, btcpayserver set their gap limit to either 1,000 or 10,000. In my opinion the additional computation required is well worth the pain it will save for folks who have funds stranded beyond the standard gap limit.

Enegnei:
> I'd highly encourage using a gap limit that far exceeds the standard. Consider the fact that there are a variety of applications such as payment processing that can result in huge gaps. If I recall correctly, btcpayserver set their gap limit to either 1,000 or 10,000. In my opinion the additional computation required is well worth the pain it will save for folks who have funds stranded beyond the standard gap limit.

For [reference](https://github.com/btcpayserver/btcpayserver/blob/3b91b380146182d48d0e41262075e8f141ac9249/BTCPayServer/Models/WalletViewModels/RescanWalletModel.cs), yes, their gap limit appears to be 10,000.

SomberNight:
> I'd highly encourage using a gap limit that far exceeds the standard.

Do you mean for the discovery, or after the wallet is created for normal operations?

If you mean for the discovery, then ok, perhaps. It seems a bit unlucky to have a wallet where none of the first few addresses are used but later ones are, but I guess it can happen if a merchant was griefed as soon as they started using a wallet.

But to be clear, regardless of what insane gap limits are some services are using, we are not able to use insane high gap limits simply due to architecture limitations. IMO what these services do when they use a gap limit of 10000 by default is anti-social behaviour: they are pulling the whole ecosystem in the same direction and slowly forcing everyone to use larger and larger gap limits (what if poor user used insane service no.17, we should use a higher gap limit then...). By using those high gap limits they introduce the risk for the user that they might not be able to restore easily using other services. That's just how it goes, we are not participating in this game.

lukechilds:
Some back-of-the-envelope calculations re gap limit:

It currently takes around 2 seconds per account for discovery to complete with a gap limit of 20.

That means in the best case scenario where there are 0 used accounts, we will only scan 13 paths so it'll take ~26 seconds.

Setting the gap limit to 10,000 will mean the **best case** recovery process increases from ~26 seconds to ~3.6 hours.

SomberNight:
> Setting the gap limit to 10,000 will mean the best case recovery process increases from ~26 seconds to ~3.6 hours.

Default ElectrumX DOS limits will get hit a lot sooner than that, slowing you down to oblivion.

lukechilds:
One solution could be to read the gap limit from the Electrum config if it exists instead of hardcoding to 20. Or maybe add a new `BIP39_recovery_gap_limit` param specifically for this.

Then if you connect to your own local Electrum instance, disable DoS limits, and manually set the config param to 10,000 it would work.

But if you're capable of all this, you're probably also capable of just manually restoring your wallet.

aantonop:
The audience for this "wizard" is newbies who don't know their derivation
path. The standard gap limit on import is enough. This is not a tool for
merchants or advanced users.

Let's keep it as simple and efficient as possible.

lukechilds:
>The resulting file name could use another label (as the [standard] one), so the user knows which one he selected after the scan.

@rdymac that would indeed be nice but unfortunately that isn't really possible without making quite large changes to the way the Electrum import wizard works.

Currently this PR is just adding a modal to one window and then auto-filling some fields in that window on completion.

Keeping this change simple and not too intrusive into the way Electrum works was a specific goal for this PR: https://github.com/spesmilo/electrum/issues/6155#issuecomment-629783062

SomberNight:
>> The resulting file name could use another label (as the [standard] one), so the user knows which one he selected after the scan.
>
> @rdymac that would indeed be nice but unfortunately that isn't really possible without making quite large changes to the way the Electrum import wizard works.

Yes, let's not change `[standard]`, it is related to the wallet type.
Note that the chosen script type is visible in `Wallet`>`Information`.
The derivation path is not visible in the GUI atm, but we could add it to the same dialog. If you want this, please see https://github.com/spesmilo/electrum/issues/4700 (and move that discussion there).

ecdsa:
> IMO what these services do when they use a gap limit of 10000 by default is anti-social behaviour

that's my opinion too. a high limit might work if you are scanning blocks yourself, but we should not engage in such behaviour with public servers.

electrum/bip39_recovery.py
```diff
@@ -0,0 +1,51 @@
...
+    gap_limit = 20
+    for address_index in range(gap_limit):
+        address_node = account_node.subkey_at_public_derivation("0/" + str(address_index))
+        pubkey = address_node.eckey.get_public_key_hex()
+        address = bitcoin.pubkey_to_address(script_type, pubkey)
+        script = bitcoin.address_to_script(address)
+        scripthash = bitcoin.script_to_scripthash(script)
+        history = await network.get_history_for_scripthash(scripthash)
+        if len(history) > 0:
+            return True
```
SomberNight, line 39: This `for` [...] `await task` construct here will mean that we send a request to the server, wait for it to answer, and then send the next request. This is why it is taking as long as you said (several seconds or similar).
That is what the `await` does, it actually stops the coroutine, and asks to only be rescheduled after that task has returned (mid-loop).

Instead, I suggest you use a TaskGroup. A TaskGroup lets you create a context (a `with` block) which will only exit once all the tasks inside returned; or if any task raises, it will cancel the remaining tasks and just re-raise the exception (so that it can propagate out).
That way, you can send all the requests "concurrently", i.e. send them all, and handle the responses as they arrive.
See these examples:
https://github.com/spesmilo/electrum/blob/fcbc1c9a459d79d54d6c102912f56b922d58fb19/electrum/interface.py#L567
https://github.com/spesmilo/electrum/blob/fcbc1c9a459d79d54d6c102912f56b922d58fb19/electrum/interface.py#L594
https://github.com/spesmilo/electrum/blob/fcbc1c9a459d79d54d6c102912f56b922d58fb19/electrum/tests/test_lnpeer.py#L504

Note that atm these nested loops you have will send all N*20 requests in series, so we will wait for over 100 network RTTs. By putting a TaskGroup here, you can reduce that to around N RTTs (N being the number of derivation path prefixes).
As further optimisation, you could use a TaskGroup at the outer loop too, so that it becomes just a single RTT (disregarding the account stuff for the moment).

lukechilds: Ahh, thanks!

I actually attempted to get all requests going concurrently by first saving executing the async functoins and saving the return values in a variable and then awaiting them later.

Something like:

```python
requests = []
for url in urls:
  requests.append(fetch(url))
results = []
for request in requests:
  results.append(await request)
```

This works in JavaScript since the async function is triggered by the function execution, `await` is just used to wait for the response.

Seems like in Python the async function execution just returns some "thing" that represents an async function call and doesn't actually execute anything until it is awaited, is that correct?

Since the above code still seemed to execute the requests serially.

e.g, equiv JavaScript code is:

```js
equests = []
for (url of urls) {
    requests.push(fetch(url))
}
results = []
for (request of requests) {
    results.push(await request)
}
```

All `fetch()` requests actually happen concurrently here and once the slowest request has completed you and up with `results` being an array of all the resolved values.

I'll have a play with `TaskGroup` and try and get it working concurrently.

Also, is there not a risk we hit ElectrumX rate limits if we do all requests concurrently?

lukechilds: I implemented concurrency, both at the address lookup level and the derivation scheme level.

The entire discovery process now takes ~2s for my test seed. 🎉

electrum/base_wizard.py
```diff
@@ -492,7 +492,11 @@ def restore_from_seed(self):
     def on_restore_seed(self, seed, is_bip39, is_ext):
         self.seed_type = 'bip39' if is_bip39 else mnemonic.seed_type(seed)
         if self.seed_type == 'bip39':
-            f = lambda passphrase: self.on_restore_bip39(seed, passphrase)
+            def f(passphrase):
+                # TODO: Check with @SomberNight if this is ok
+                self.seed = seed
```
lukechilds, line 497: @SomberNight is it ok to persist sensitive data on the `BaseWizard` instance like this?

I'm attaching `seed` and `passphrase` to the instance so I can access them later in the derivation dialog.

I'm not exactly sure how long the `BaseWizard` instance lives for, but I assume it's destroyed once the wizard has completed so `seed`/`mnemonic` would be garbage collected.

lukechilds:
I've addresses the feedback and tested/fixed when setting up with a HWW wallet.

Unless storing seed data on `BaseWizard` is an issue https://github.com/spesmilo/electrum/pull/6219#discussion_r443206598 I think this is ready to be merged.

Let me know if you require any other changes or you think there's any other scenarios I should test.

electrum/base_wizard.py
SomberNight: As on IRC, I think (seed, passphrase) could be passed around as parameters. Probably optional keyword-only args, as the hardware wallet codepath will not have them.

SomberNight:
To expand on https://github.com/spesmilo/electrum/pull/6219/files#r444897453 ,
I think what we could pass around instead is a function `(derivation_prefix) -> (account_xpub)` (or similar), as then this could be (later?) implemented for hardware wallets as well.

andronoob:
@aantonop

>This is not a tool for merchants or advanced users.

Why it can't? In my opinion, it can bring convenience for them as well. I think this may be implemented in a progressive favour, with some rate limit to avoid unintentionally DoSing the server.

andronoob:
I also wonder whether the "BIP39-Electrum2.0 duality" issue (#1300) is properly handled here?

Electrum 2.0 seed phrase is not compatible with BIP39, but it still shares exactly the same wordlist with BIP39. Therefore, an Electrum 2.0 seed can be both a valid Electrum 2.0 seed and a valid BIP39 mnemonic, with a probability of one in sixteen. (12-word BIP39 mnemonic phrase has 4-bit checksum. 2^4=16)

It's then possible (although not very possible, because the BIP39 wallet would show empty transaction history and zero balance, so that the user is highly unlikely to continue) for a user to import an Electrum 2.0 seed into another BIP39-compliant wallet.

andronoob:
I dreamed that the new derivation path step of the wizard may look like this:


### Specify your derivation path
----
#### Standard derivation paths
- Legacy (P2PKH)
  Example: 1CuC3b...
- SegWit, compatible (P2SH-P2WPKH)
  Example: 3Dq4Y9...
- SegWit, native Bech32 (P2WPKH)
  Example: bc1qff6a...

  Account number: _____
#### Non-standard derivation paths
- `Click here to select from the list...`
#### Not sure?
- `Let me guess!`

**(empty, if no detection is ever executed) / No fund detected / Recoverable funds detected, congratulations! You may click "next" now.** `[Wait for 30s] Detect funds / (Button hidden if funds detected) / Click to probe the next XXXX addresses`

----
  *Don't touch things below if you have no idea what they mean.*

  BIP32 derivation path: _________  Address type: `P2WPKH (dropdown menu)`

  Gap limit: ____
- [ ] Manually specify the starting child key index: ____

ecdsa:
related: #6001

lukechilds:
@SomberNight @ecdsa are you able to re-review when you have time?

I've resolved the issue of attaching the mnemonic/passphrase to the instance and instead now pass an `(account_path) -> (account_xpub)` function around.

I haven't implemented this function for when a HWW is being restored but I have tested the wizard and it correctly leaves out the restore option when in HWW mode.

lukechilds:
@SomberNight @ecdsa sorry to keep pinging you guys, it's been a month now since the last time I checked in with you.

AFAIK I've implemented all required changes and there aren't any conflicts, is there anything else you want from me for this or can we get it merged in?

SomberNight:
Starting with https://github.com/spesmilo/electrum/commit/41827cff1943819c221acf2e2be675d049b4b7c4 (copy here: https://github.com/SomberNight/electrum/tree/pr/6219_bak20200820),
I've squashed all your commits, https://github.com/spesmilo/electrum/commit/cc7a6e8e64585d241c6ad2fd85a38c420a81bf3f,
and rebased on master, https://github.com/spesmilo/electrum/commit/7b122d26792af7e856c8e841dff8f4c5e83c3566.

electrum/bip39_recovery.py
```diff
@@ -0,0 +1,75 @@
...
+def format_account(wallet_format, account_path):
+    description = wallet_format["description"]
+    if wallet_format["iterate_accounts"]:
+        account_index = account_path[-1] % BIP32_PRIME
+        description = f'{description} (Account {account_index})'
+    return {
+        "description": description,
+        "derivation_path": bip32_ints_to_str(account_path),
+        "script_type": wallet_format["script_type"],
+    }
```
SomberNight, line 75: Would be better to return a NamedTuple or similar; e.g. so that fields can be typed and can be enumerated statically (by IDE).

SomberNight (approved):
Looks good. Reviewed code and tested.

I have some minor nits (in addition to what I pushed) but that's fine.
Noticed that the Qt dialog has a sizing bug on dark theme (too small by default) - might be platform dependent, and probably related to QDarkStyle, so ignoring for now (my naive attempts to fix it failed).

SomberNight:
Thank you for your work.

AlistairTB:
I would like to do exactly this with a multisig wallet using Zpub keys (for a read-only wallet). In case it's not obvious, the Zpubs were derived from BIP39 words and extended passphrase. Is this possible?

> I've integrated the functionality into the GUI.
> 
> Demo:
> 
> electrum_bip39
> 
> Let me know if it looks ok or if you have any suggestions, thanks!

SomberNight:
@AlistairTB No, this is only implemented for singlesig wallets.

Also, it is not clear exactly how you would want to use this.
The functionality here is for bruteforcing the derivation path and script type.
For a multisig wallet given Zpubs, you already know the script type, and you would typically have the Zpubs already at the "account level" of the derivation, so you already know the derivation path.
Further note that it is not possible to do hardened derivation starting from a master public key.

AlistairTB:
Thank you for the reality check. I was using it the way it was intended, however, I had the wallet type wrong (3/3 vs 2/3) and was looking to solve the wrong problem.

SomberNight:
> I'd say this feature would be very well appreciated also when creating a watch-only wallet from a hardware device. Shall I open a dedicated issue?

Sure, go ahead.

PR #6220, opened 2020-06-09
Remove dependencies: jsonrpcserver, jsonrpcclient

ecdsa:
we don't need to pull those dependencies, aiohttp makes it easy to write from scratch

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1368,3 +1368,30 @@ class MySocksProxy(aiorpcx.SOCKSProxy):
         else:
             raise NotImplementedError  # http proxy not available with aiorpcx
         return ret
+
+
+class myAiohttpClient:
+
+    def __init__(self, session, url):
+        self.session = session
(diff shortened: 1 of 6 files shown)


SomberNight:
Oh wow, it would be great to remove those dependencies!

electrum/util.py
```diff
@@ -1368,3 +1368,30 @@ def from_proxy_dict(cls, proxy: dict = None) -> Optional['MySocksProxy']:
...
+
+
+class myAiohttpClient:
+
+    def __init__(self, session, url):
+        self.session = session
+        self.url = url
+
+    async def request(self, endpoint, *args):
+        data = '{"jsonrpc": "2.0", "id":"0", "method":"%s", "params": %s }' %(endpoint, json.dumps(args))
```
SomberNight, line 1380: I guess `id` could be based on a counter in `self`.

PR #6252, opened 2020-06-22
LN invoices: support msat precision

### LINKED ISSUE #6250: cannot pay LN invoices that have msat resolution
As in title, Electrum cannot pay LN invoices that have msat resolution,
as the invoice-handling code will store the invoice with a rounded down satoshi integer amount.
The remote node will see we are trying to underpay and fail the HTLC.

See e.g. https://github.com/spesmilo/electrum/blob/f91674992f6982ddbc0a7bd1a1a57fb43165c9e0/electrum/invoices.py#L130

SomberNight:
fixes #6250

@ecdsa please check if the storage upgrade looks ok


SomberNight:
regtest tests are failing :/ maybe re `jq`?

Issue #6254, opened 2020-06-22
Number of confirmations uses local height (and ignores server height), potentially resulting in SPV security confirmation numbers even with own server


SomberNight:
With default settings, e.g. `auto_connect` enabled, it is expected that confirmation counts for txns have SPV-security.
It is also expected that with manual server selection (`auto_connect` disabled) and `oneserver`, assuming  you selected a trusted server, confirmation counts have full-node security.

What is perhaps unexpected is what happens if you do manual server selection (`auto_connect` disabled) but no `oneserver`.
In this case, the first confirmation has full-node security, as it is the main interface that sends the SPV proof (so the main server tells us that the tx has been mined).
However any further confirmations only have SPV-security:
if we receive POW-valid block headers that link up to the block that mined a tx, even if those were not received from the main server, those will be counted as confirmations.

https://github.com/spesmilo/electrum/blob/85841ba20de7cb6459066538474d3f13e010031e/electrum/address_synchronizer.py#L605-L606
https://github.com/spesmilo/electrum/blob/85841ba20de7cb6459066538474d3f13e010031e/electrum/address_synchronizer.py#L589
https://github.com/spesmilo/electrum/blob/85841ba20de7cb6459066538474d3f13e010031e/electrum/network.py#L1124-L1129

SomberNight:
Sort of related:
For transactions we make, we set the locktime to the current block height,
to imitate bitcoind and for anti-fee-sniping reasons.
https://github.com/spesmilo/electrum/blob/85841ba20de7cb6459066538474d3f13e010031e/electrum/wallet.py#L194

The locktime is (effectively) being set to `network.get_local_height()`, which then has the same problem: if any interface (not just the main server) sends us valid POW headers that are for invalid blocks, we will set the locktime into the "future", and will not be able to broadcast/propagate our transactions (until the real chain overcomes the fake headers).

Note that for any transaction that is the result of user interaction, the user would notice that something is going on, and they might be able to remedy the issue. Atm we only set the locktime to the chaintip for user-created transactions, however I was thinking we could do it for e.g. submarine swap claim txns and lightning sweeps (lnsweep.py).

drdran:
Hi we are facig off to the same problem, we can support u for testing if and debugging the problem, if u have some ideas how to fixit.

PR #6255, opened 2020-06-22
anti-fee-sniping for lnsweep and swap claim

SomberNight:
We set the locktime to the chaintip for wallet transactions.
We do this for anti-fee-sniping reasons (and to imitate Bitcoin Core):
https://github.com/bitcoin/bitcoin/blob/c27330897d84144ad128994dc51efde6084e796f/src/wallet/wallet.cpp#L2617

Currently this is only done for wallet transactions, so e.g. not lightning sweep stuff and submarine swap claim txs.

Note however that wallet transactions are made with user interaction, while ln sweeps and swap claims are automated.
Related:
https://github.com/spesmilo/electrum/issues/6254#issuecomment-647233004

### DIFF
--- a/electrum/bitcoin.py
+++ b/electrum/bitcoin.py
@@ -27,6 +27,7 @@ import hashlib
 from typing import List, Tuple, TYPE_CHECKING, Optional, Union
 import enum
 from enum import IntEnum, Enum
+import random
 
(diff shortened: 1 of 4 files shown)


ecdsa:
Fixed by #9298 :  `TxBatcher` calls wallet.make_unsigned_transaction, which will set a locktime if the passed locktime is None.
Related: b64109ea8868345a1ebbdf851daddc1f211220fb

Issue #6299, opened 2020-06-28
qt dark theme: status bar buttons and help buttons are too wide with qdarkstyle 2.8.1


SomberNight:
With latest qdarkstyle (2.8.1), the status bar buttons (bottom right of main window), and the help buttons ("?") became too wide.
(This only affects the dark theme)

![qt_push_buttons_too_wide](https://user-images.githubusercontent.com/29142493/85953282-cae08e80-b95e-11ea-8409-53833cdc3087.PNG)

Compare with default (light) theme:
![qt_push_buttons_too_wide2](https://user-images.githubusercontent.com/29142493/85953354-3e829b80-b95f-11ea-8b3f-c179ed88ad44.PNG)

SomberNight:
This was introduced by https://github.com/ColinDuquesnoy/QDarkStyleSheet/issues/194

See
https://github.com/ColinDuquesnoy/QDarkStyleSheet/issues/217
https://github.com/ColinDuquesnoy/QDarkStyleSheet/issues/230

SomberNight:
this was fixed by https://github.com/spesmilo/electrum/pull/6300

PR #6300, opened 2020-06-28
qt StatusBarButton: use QToolButton instead of QPushButton

SomberNight:
see: #6299

I think in general it would look better with this change;
except on macOS with the light theme :( There the status bar buttons have an outline/background, which makes it ugly...

-----

Pics before:
<details>
 <summary>Ubuntu GNOME</summary>

  ubuntu_light
  ubuntu_dark
</details>

<details>
 <summary>Windows</summary>

  win_light
  win_dark
</details>

<details>
 <summary>macOS</summary>

  mac_light
  mac_dark
</details>

-----

Pics after:
<details>
 <summary>Ubuntu GNOME</summary>

  ubuntu_light
  ubuntu_dark
</details>

<details>
 <summary>Windows</summary>

  win_light
  win_dark
</details>

<details>
 <summary>macOS</summary>

  mac_light
  mac_dark
</details>

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -46,7 +46,7 @@ from PyQt5.QtWidgets import (QMessageBox, QComboBox, QSystemTrayIcon, QTabWidget
                              QHBoxLayout, QPushButton, QScrollArea, QTextEdit,
                              QShortcut, QMainWindow, QCompleter, QInputDialog,
                              QWidget, QSizePolicy, QStatusBar, QToolTip, QDialog,
-                             QMenu, QAction, QStackedWidget)
+                             QMenu, QAction, QStackedWidget, QToolButton)
 
 import electrum
 from electrum import (keystore, ecc, constants, util, bitcoin, commands,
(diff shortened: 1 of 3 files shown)


danuker:
The autoRaise property is ignored by the Mac OS implementation:
https://doc.qt.io/qt-5/qtoolbutton.html#autoRaise-prop

My $0.02 is that this change would be a step forward, and the inconsistency between OSs is a bug in QT. I have reported it here: https://bugreports.qt.io/browse/QTBUG-85843

If you want to, a StackOverflow workaround is to use a custom style to hide the background rectangles:
https://stackoverflow.com/questions/11803260/qtoolbutton-autoraise-not-working-for-mac-osx

SomberNight:
@danuker Thank you for reporting it to Qt! While the autoRaise behaviour not working on macOS is documented in the Qt docs, I agree that it is a bug. Also thanks for pointing out that it could be worked around with a custom stylesheet.

I added a custom stylesheet just for the macOS+default_theme case, and it looks good now.

<details>
 <summary>macOS light, with custom stylesheet</summary>

  macos_light_2
</details>

danuker:
@SomberNight Cheers! I don't even have a mac, but I love Electrum. Thanks for your work!

PR #6304, opened 2020-06-29
fix raise UserCacnelled calls

zebra-lucky:
These `raise UserCacnelled` calls looks like typos


SomberNight:
note that both
```
raise Exception
```
and
```
raise Exception()
```
are valid python

SomberNight:
but yeah, ok, I guess we use the latter more frequently.

zebra-lucky:
> note that both
> 
> ```
> raise Exception
> ```
> 
> and
> 
> ```
> raise Exception()
> ```
> 
> are valid python

Sure, but mixed use in one module somewhat confusing. Thanks!

PR #6309, opened 2020-06-29
plugins/bitbox02: fix compatibility with bitbox02-4.0.0

benma:
It would be great if this made it into the next beta release (or real release :crossed_fingers:).

Edit: since the bb02 plugin is now released with Electrum, we will be careful to preserve backwards compatibility as best as we can going forward. Thanks and congrats on Electrum 4! :rocket:

### DIFF
--- a/contrib/requirements/requirements-hw.txt
+++ b/contrib/requirements/requirements-hw.txt
@@ -13,5 +13,5 @@ safet>=0.1.5
 keepkey>=6.3.1
 btchip-python>=0.1.26
 ckcc-protocol>=0.7.7
-bitbox02>=3.0.0
+bitbox02>=4.0.0
 hidapi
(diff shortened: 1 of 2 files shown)


SomberNight:
Just to clarify, this needs fw 9.0.0 but that has not been released yet?
I could not test this (old firmware) but I trust that you tested it.

Please update `BitBox02Plugin.minimum_library` around line 571 then I will merge.

benma:
Good catch, fixed the min library version.

Exactly, we tested it and will basically release as soon as you release if this week, or latest in around 1 week.

If you want to test it anyway, I can send you a pre-release app with the new firmware.

Thanks a lot.

Stadicus:
Tested successfully for both singlesig and multisig with BitBox02 firmware v9.0.0

SomberNight:
> Exactly, we tested it and will basically release as soon as you release if this week, or latest in around 1 week.

fyi we tagged 4.0.1

PR #6315, opened 2020-07-02
interface: check server response for some methods

SomberNight:
some basic sanity checks

Previously if the server sent back a malformed response, it could partially corrupt a wallet file.
(as sometimes the response would get persisted, and issues would only arise later when the values were used)

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -29,7 +29,7 @@ import sys
 import traceback
 import asyncio
 import socket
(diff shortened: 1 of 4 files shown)


SomberNight:
I did not end up using `is_real_number` in the end as turns out [the protocol doc](https://electrumx-spesmilo.readthedocs.io/en/latest/protocol-methods.html#blockchain-scripthash-get-balance) was [incorrect](https://github.com/spesmilo/electrumx/pull/31), so I could rm that.

Issue #6321, opened 2020-07-03
qt-specific dependencies are listed in shared requirements.txt, so e.g. the android apk will include them


SomberNight:
`contrib/requirements/requirements.txt` lists shared pure-python dependencies that we declare are always needed for Electrum.
These will be installed even if only using CLI, or Qt, or kivy.

Note that atm `qdarkstyle` is listed there, which is only needed for the Qt GUI.
As a consequence, it will be included even e.g. in the Android apk (along with its dependencies!).

https://github.com/spesmilo/electrum/blob/0f6898ed90be7777254b6868e1e1b0b5af79372f/contrib/requirements/requirements.txt#L6

This is not that important though as it's just a relatively small source package, but still a nit.

SomberNight:
reported by @akshayaurora

PR #6333, opened 2020-07-04
remove QDarkStyle from packages before building for android

akshayaurora:
For https://github.com/spesmilo/electrum/issues/6321


ecdsa:
this would better be done with exclude_dir in buildozer.spec

PR #6375, opened 2020-07-11
Change default invoice to never expires

### LINKED ISSUE #6364: Set default expiration time for payment requests 'Never"
Why a fresh wallet will default payment requests expiration to some other value than "Never"?
After upgrading, I had to configure an existing wallet to "Never" on requests tab. I think this should be the default, having 10 mins or something like this will increase the gap of users too much if many requests get burned  / don't end up paid. 

If someone wants a limited lifetime for the validity of a payment request for sure they will set it, but the average users almost never need or want to have an expiring payment request, so I think we should default to "Never" here on both Android and desktop.

hoganri:
Resolves #6364 

I think this a valid point - an invoice with a time limit will definitely be set by the user but most payment times are flexible. I'd rather have it not expire by default and manually change the expiry during the few situations where it is time-sensitive.

### DIFF
--- a/electrum/invoices.py
+++ b/electrum/invoices.py
@@ -50,7 +50,7 @@ pr_tooltips = {
     PR_ROUTING: _('Computing route...'),
 }
 
-PR_DEFAULT_EXPIRATION_WHEN_CREATING = 24*60*60  # 1 day
+PR_DEFAULT_EXPIRATION_WHEN_CREATING = 0 # Never
 pr_expiration_values = {
     0: _('Never'),
     10*60: _('10 minutes'),


danuker:
This poses a problem when restoring a wallet from a seed.

A Denial of Service attack could request tons of invoices, using up lots of addresses, then not paying anything.

The seed restoration process would have to pointlessly generate past those empty addresses, in order to find the ones that are actual received payments.

I disagree with having to remember invoices indefinitely; set it to a finite but longer interval if you wish. I think a sensible value would be the mempool expiry of 2 weeks: https://github.com/bitcoin/bitcoin/blob/e9fc8f6e7f572224d2bd4577eee6d538331ef663/test/functional/mempool_expiry.py#L22

danuker:
An invoice with no time limit will force the server to remember it forever. I disagree with this.

hoganri:
Won't do

PR #6386, opened 2020-07-14
bitbox02: implement get_soft_device_id so multisig runs more smoothly

benma:
Without it, if you have say a 1-of-2 multisig with two BitBox02s, you
would run into trouble if the first keystore would try to match to the
wrong inserted BitBox02 (wrong order, or the first one is not
inserted, etc. ).

With the soft device id, the device manager can figure it on its own
which keystore belongs to which connected bb02.

### DIFF
--- a/electrum/plugins/bitbox02/bitbox02.py
+++ b/electrum/plugins/bitbox02/bitbox02.py
@@ -251,7 +251,12 @@ class BitBox02Client(HardwareClientBase):
             return super().label()
         if self.bitbox02_device is None:
             self.pairing_dialog()
-        return self.bitbox02_device.device_info()["name"]
+        # We add the fingerprint to the label, as if there are two devices with the same label, the
+        # device manager can mistake one for another and fail.
+        return "%s (%s)" % (
+            self.bitbox02_device.device_info()["name"],
+            self.bitbox02_device.root_fingerprint().hex(),
+        )
 
     def request_root_fingerprint_from_device(self) -> str:
         if self.bitbox02_device is None:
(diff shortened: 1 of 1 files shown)


electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -85,7 +85,16 @@ def has_usable_connection_with_device(self) -> bool:
             return False
         return True
 
-    def pairing_dialog(self, wizard: bool = True):
+    def get_soft_device_id(self) -> Optional[str]:
+        if self.handler is None:
+            # Can't do the pairing without the handler. This happens at wallet creation time.
+            return None
```
benma, line 91: @SomberNight 

When setting up the wallet in the wizard, this function is called while `handler` is still None. Not sure how to handle this, but just not returning any device id at this stage seems to work fine. Is this okay or should this always return an ID?

If you want, you can test the multisig with one device by enabling the optional passphrase and just choosing two different passphrases.

benma: Looks like the handler is None only when the devices are listed in the wizard. In our testing, this PR works as expected.

benma:
Bump (just in case it dropped off your backlog :smile: ). Would be nice to have this in the next release.

Stadicus:
Tested successfully yesterday with three BitBox02's in parallel. Works as expected, including display of device labels.

SomberNight:
I've only tested the basics as it's really cumbersome with only one device :)

but the code and logic looks good

benma:
@SomberNight would you consider a point release for this? It would be very helpful, then our users could start using multiple bitbox02's in a multisig setup without any issues. :pray:

ecdsa:
we'll have a release before the end of the month

benma:
Awesome!

Issue #2070, opened 2016-12-08
Could you use xor to add custom entropy without letting users shoot themselves in the foot?


n8n3k:
As I understand you don't let users add custom entropy because a user might just type 12345 and think they're clever.

Could you let people automatically add custom entropy, using the xor function to combine the custom entropy with the generated entropy?

This wouldn't reduce the entropy even if the user typed in "1111" and thus couldn't hurt.

AbdussamadA:
you can add entropy using the command line option "make_seed". type `electrum help make_seed` to learn more. the entropy has to be provided as a decimal number

There is also the option to extend the seed with your own custom words. You can do that in the gui wallet creation wizard.

ecdsa:
indeed, see @AbdussamadA answer.

Issue #1574, opened 2015-12-07
Support custom_entropy from command line


the9ull:
The method [Mnemonic().make_seed()](https://github.com/spesmilo/electrum/blob/master/lib/mnemonic.py#L150) supports entropy provided by the user.

Would be useful add to the command line create command the options --entropy-{dice,permutation,...} to manage the custom entropy parameter.

Entropy could be provided in different formats:

* a single big integer
* a list of dice rolls, different dice type could be supported (base 6, 8, 12, 20, 30)
* permutation of n elements. The id of the elements are chosen by the user.

ecdsa:
I think this is out of scope; you have more flexibility by passing entropy as a parameter

the9ull:
How is it possible now pass the entropy as parameter? `electrum create` doen't have a way, am I wrong?

ecdsa:
you can use make_seed and restore from it

Issue #523, opened 2014-01-03
Enter custom entropy


TheMastor:
With the [NSA's shenanigans](http://wccftech.com/intel-possibly-amd-chips-permanent-backdoors-planted-nsa-updated-1/) and also [this fatal weakness](http://android-developers.blogspot.com.au/2013/08/some-securerandom-thoughts.html) in android there is cause to doubt the CSPRNG on many systems.

To satisfy this concern it would be important to allow users to input their entropy much like how [TrueCrypt allows](http://i.imgur.com/4LvPFjW.jpg).

farzher:
hello. i want to use my own entropy bits that i got by flipping coins or whatever instead of electrum's randomness. but how? i made a BIP39 seed but you don't accept it and i don't know how to generate a valid electrum seed.

farzher:
also you won't accept an xpub, ypub, or zpub for a multisig segwit wallet.
you want a Zpub. but that also seems to be a non standard electrum thing and i can't figure out how to generate one.

i just want to use my own entropy :(

SomberNight:
> i made a BIP39 seed but you don't accept it and i don't know how to generate a valid electrum seed.

You can use a BIP39 seed if you tick the checkbox.

> you want a Zpub. but that also seems to be a non standard electrum thing and i can't figure out how to generate one.

See https://github.com/spesmilo/electrum-docs/blob/master/xpub_version_bytes.rst
x/y/z/Y/Z version bytes come from Electrum. Others started using the same thing. The only reason Zpubs are not really used by other wallets is because no one else supports segwit multisig.
The reason why multisig uses different version bytes is explained at the above link.
You can use the `convert_xkey` CLI command to convert between different xpub version bytes.

farzher:
i have to use electrum mobile version for offline key generation. the BIP39 option dosen't exist there :(

but thank you, i think i finally understand the xpub insanity. `convert_xkey` isn't too useful, because i don't have an offline computer. but at least i know exactly what source code to read to make a javascript implementation or something that i could run on my phone

SomberNight:
> allow generating mnemonics that are verifiably immune to false randomness from a backdoored CPU

What do you mean by "verifiably"?
If the code naively mixes in custom entropy, you won't be able to tell whether the provided entropy was really used or the backdoored CPU discarded it.
[...]

PR #6411, opened 2020-07-19
Add option to use deterministic entropy in make_seed

hodlwave:
I noticed that electrum does not currently have the option to generate a seed with user-provided entropy. Also, the current source of entropy comes from `os.urandom()` via the python-ecdsa dependency. That project's [README](https://github.com/warner/python-ecdsa) explicitly states:

```
This library depends upon a strong source of random numbers. Do not use it on a system where os.urandom() does not provide cryptographically secure random numbers.
```

While practically speaking this will rarely be an issue, I think it's a useful to be able to supply entropy deterministically so users can _eliminate_ this risk categorically. 

My main concern with adding this feature is novices mistakenly using it and providing a very low entropy input without clearly understanding what it does. Is it correct to assume the CLI is generally only used by power-users? In any case, I made the option's description cautionary to guard against this risk.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -320,10 +320,10 @@ class Commands:
         return self.config.get_ssl_domain()
 
     @command('')
-    async def make_seed(self, nbits=132, language=None, seed_type=None):
+    async def make_seed(self, nbits=132, language=None, seed_type=None, entropy=None):
         """Create a seed"""
         from .mnemonic import Mnemonic
-        s = Mnemonic(language).make_seed(seed_type, num_bits=nbits)
+        s = Mnemonic(language).make_seed(seed_type, num_bits=nbits, user_entropy=entropy)
         return s
 
     @command('n')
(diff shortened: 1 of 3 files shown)


SomberNight:
There had already been a custom entropy option, I think for several years, but it was removed in https://github.com/spesmilo/electrum/commit/5e5134b76ffec8ed9702f5661a1d16e9e52a24c8
Though note that that mixed in user-provided entropy, unlike your proposal which uses it exclusively.

I think a feature that relies entirely on user-provided entropy is dangerously easy to misuse: it's a footgun.

SomberNight:
Related:
https://github.com/spesmilo/electrum/issues/523
https://github.com/spesmilo/electrum/issues/1574
https://github.com/spesmilo/electrum/issues/1964
https://github.com/spesmilo/electrum/issues/2070

hodlwave:
Good points. Come to think of it, adding user entropy with a strong passphrase should be sufficient for this purpose anyway.

PR #6435, opened 2020-07-23
Save wallet DB after adding or removing payment requests

johnzweng:
## Description

When using the `add_request` command I realized that the returned address will be reused in the next payment request after I restart my electrum daemon (if no other operation which saves the walletDB has been performed in the meantime).

It seems that the `add_request`, `add_lightning_request` and `rmrequest` commands change the wallet DB (_by adding or removing payment requests_), but do not flush the changes down to the disk.

If any other operation later on changes the wallet DB, then the changes of the 3 commands will also be flushed. If not, then the changes (payment requests) might get lost.

(`clear_requests` is not affected as `self.save_db()` is called in `wallet.py` in `clear_requests(self)`.)

### This PR:
- adds `wallet.save_db()` to RPC commands which change the Wallet DB but didn't save them

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -846,12 +846,14 @@ class Commands:
         expiration = int(expiration) if expiration else None
         req = wallet.make_payment_request(addr, amount, memo, expiration)
         wallet.add_payment_request(req)
+        wallet.save_db()
         return wallet.export_request(req)
 
     @command('wn')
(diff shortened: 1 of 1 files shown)


SomberNight:
This looks okay. I am just wondering whether this is the right place to call `wallet.save_db()` in general; maybe it should be done by the callee not the caller. That would involve a larger change though probably.

PR #6455, opened 2020-07-31
add electrum plugin for cobo vault

evallozhe:
1 Cobo Vault interacts via QRcode not USB
2 Cobo Vault have no python dependent libraries


ecdsa:
Not sure if this PR is still active. Please note that the upcoming release of Electrum will let users install third-party plugins.
Hardware wallet manufacturers will have the possibility to distribute their own plugins.

PR #6467, opened 2020-08-03
Remove empty startup message from console

verretor:
- Change >> to >>> as in a normal Python console.
- Avoid printing an empty string which is why the console looked like:
```
>>
Network banner
>>
```

Instead of:
```
Network banner
>>>
```

- Remove **welcome_message** from console.py
It is not used anymore.
The console prints **network.banner** instead.

### DIFF
--- a/electrum/gui/qt/console.py
+++ b/electrum/gui/qt/console.py
@@ -44,7 +44,7 @@ class OverlayLabel(QtWidgets.QLabel):
 
 
 class Console(QtWidgets.QPlainTextEdit):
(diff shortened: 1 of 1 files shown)


SomberNight:
I am reverting https://github.com/spesmilo/electrum/commit/822083d16848915f54c1bf4d7328536a5d861efc.
It breakes using the console with --offline, and presumably also when trying to use it before establishing a connection to the first server (consider a very slow network).
see https://github.com/spesmilo/electrum/issues/6731

PR #6471, opened 2020-08-04
trezor: use `init_device` instead of `ping` to check connection

### LINKED ISSUE #6457: Trezor with autlock feature error when creating new or accessing old wallet
**Enviroment**
Latest Windows 10
Electrum 4.0.2
FW Trezor 2.3.2-731e9b7c

**Behavior**
When you enable autlock feature on Trezor T and device is locked and you will try to create new wallet or loading existing one.
You see error on screenshot and not proper unlock device dialogue. And of course device freezes for some time
Electrum probably check state of device and it gets confused ;)

Send, receive, sign and verify works fine ans it show unlock dialogue. 

@matejcik please check :1st_place_medal: 

**Screenshot**
photo_2020-07-31_13-46-55

matejcik:
fixes #6457

Trezor T does not allow `Ping` message without unlocking the device. There's a matching bug in `trezorctl` which makes this worse: `client.ping()` should cleanly handle the situation but it doesn't.

The easiest and most user-friendly solution in Electrum is to use `init_device()` call instead; it's guaranteed to not require user interaction, which is what we want in `has_usable_connection`.

### DIFF
--- a/electrum/plugins/trezor/clientbase.py
+++ b/electrum/plugins/trezor/clientbase.py
@@ -112,8 +112,7 @@ class TrezorClientBase(HardwareClientBase, Logger):
             return True
 
         try:
-            res = self.client.ping("electrum pinging device")
-            assert res == "electrum pinging device"
+            self.client.init_device()
         except BaseException:
             return False
         return True


SomberNight:
Thanks

PR #6475, opened 2020-08-05
Allow 'serialize' command to be run from cli

brianddk:
CLI arguments are always strings and `serialize` expects a dict.  Added simple call to `json_decode` to allow either a string or a dict to be passed as the jsontx parameter.  The `json_decode` command is non-destructive so if it is passed a dict, it simply returns the passed in dict.

This PR allows the following type commands to work

```shell
python run_electrum serialize --testnet '{"inputs": ... }'
```

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -360,6 +360,7 @@ class Commands:
         Inputs must have a redeemPubkey.
         Outputs must be a list of {'address':address, 'value':satoshi_amount}.
         """
+        jsontx = json_decode(jsontx)
         keypairs = {}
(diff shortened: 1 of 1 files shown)


brianddk:
The `arg_types` dict should have done the JSON parse, but didn't.  I'll need to figure out what went wrong on [line 1243](https://github.com/spesmilo/electrum/blob/b5cfaebd27fd6db64e231c0940e524c57bec1686/electrum/commands.py#L1243)

'''
    'jsontx': json_loads,
'''

PR #6476, opened 2020-08-05
Correct errors in command 'serialize(deserialize(pbst))'

brianddk:
Ideally the commands `deserialize` and `serialize` should reverse each other.  This will provide the minimum amount of correction to allow `serialize(deserialize(pbst))` to complete without error

* Fixed 'locktime' typo that had it incorrectly camel-cased
* Made 'value' and 'value_sats' synonyms so either can be used

It is assumed that the field "value" is a already in SATs.  If "value" is assumed to be in BTC or mBTC, a correction will be needed.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -362,7 +362,7 @@ class Commands:
         """
         keypairs = {}
         inputs = []  # type: List[PartialTxInput]
-        locktime = jsontx.get('lockTime', 0)
+        locktime = jsontx.get('locktime', 0)
         for txin_dict in jsontx.get('inputs'):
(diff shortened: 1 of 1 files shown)


ecdsa:
I think we should drop 'value' in favor of 'value_sats'.
it might break existing scripts, but upgrading them should be easy.
I think it's better than adding complexity.

SomberNight:
Indeed serialize and deserialize should be inverses of each other.
I think we should do a breaking change to change the field names; instead of carrying compat code, however only merge it into a minor release, not a patch release. So e.g. 4.1

PR #6490, opened 2020-08-08
fix Transaction.ouputs method on non-standard transactions

zebra-lucky:
Possible fix for #5513


zebra-lucky:
Slightly modify changes in second try.

SomberNight:
I appreciate the PR but do you understand the underlying cause? How exactly does the issue get triggered?
I don't see how atm. :/
You wrote "non-standard transactions" in the title; do you have a tx by any chance that triggers it?

zebra-lucky:
Yep, it is not standard tx and current core will result with `bad-txns-vout-empty`.
Seems block with such tx will be rejected also.

Need to read more, maybe I'll found some cause of empty outputs,
or it is some non standard blockchain.

Issue #6453, opened 2020-07-30
Quick start example script error


breakfixrepeat:
I'm playing with some of the example scripts provided and am not having much luck with the quick_start.py example.

Running the script results in the following error, attributed to the `daemon = Daemon(config, listen_jsonrpc=False)` line:

`assert self.asyncio_loop.is_running(), "event loop not running"`

Any hints how to resolve this?

PR #6492, opened 2020-08-09
fixed initial network assertion error

frogeyedpeas:
Added a create_network so daemon assertions don't fail upon running.


SomberNight:
Thank you for the pull request.
Apologies but I have not noticed it in time and already pushed another fix for this in https://github.com/spesmilo/electrum/commit/42da407ee16784eb6e2d1f1c644f77d97d7c3007, while looking at https://github.com/spesmilo/electrum/issues/6453.

Issue #6511, opened 2020-08-15
Ledger: New BIP44 checking in ledger firmware 1.6.1+ breaks electrum


smogm:
Since this new BIP44 enforce checking (https://github.com/LedgerHQ/app-bitcoin/commit/bae78c044ace0480945f84db84bf67a8790dfacb) in the app-bitcoin ledger app, the electrum ledger plugin seems not working anymore.

https://support.ledger.com/hc/en-us/articles/360015739499-Sign-or-derivation-path-is-unusual

During startup of the electrum client the Ledger shows "The derivation path is unusual" a few times which you have to "approve" each time.
Maybe the startup and initialization sequence of the electrum ledger plugin is not correct anymore?

Also see https://github.com/LedgerHQ/app-bitcoin/issues/161

smogm:
I also supplied a fix for this: https://github.com/spesmilo/electrum/pull/6512

SomberNight:
note: more discussion in https://github.com/spesmilo/electrum/pull/6512

PR #6512, opened 2020-08-15
fix ledger issues with bip44 path restriction checking

smogm:
Fixes the issues described here: https://github.com/spesmilo/electrum/issues/6511

### DIFF
--- a/electrum/plugins/ledger/ledger.py
+++ b/electrum/plugins/ledger/ledger.py
@@ -4,13 +4,13 @@ import sys
 import traceback
 from typing import Optional, Tuple
 
-from electrum import ecc
+from electrum import ecc, constants
 from electrum import bip32
 from electrum.crypto import hash_160
 from electrum.bitcoin import int_to_hex, var_int, is_segwit_script_type
 from electrum.bip32 import BIP32Node, convert_bip32_intpath_to_strpath
 from electrum.i18n import _
-from electrum.keystore import Hardware_KeyStore
+from electrum.keystore import Hardware_KeyStore, Xpub
 from electrum.transaction import Transaction, PartialTransaction, PartialTxInput, PartialTxOutput
 from electrum.wallet import Standard_Wallet
 from electrum.util import bfh, bh2u, versiontuple, UserFacingException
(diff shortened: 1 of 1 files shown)


SomberNight:
I don't think this is going to work.  Have you tested, e.g. opening an existing wallet?

The Ledger firmware (bitcoin app) might need changes instead.
We had similar issues with bitbox02 (see comments in https://github.com/spesmilo/electrum/pull/5993 and https://github.com/spesmilo/electrum/pull/5993#discussion_r384212257) and we concluded this is best fixed in the firmware.

There are two mostly independent things for which we ask xpubs for:

1. We want a password to encrypt the wallet file. For this we use a hardcoded path atm.
https://github.com/spesmilo/electrum/blob/9b416b577edfa2e4f4a62956b64262a1fcd52afa/electrum/storage.py#L40-L43
This current PR just changes that hardcoded path in case of Ledger. This breaks existing wallets.
Ideally, our hardcoded path should get whitelisted in the Ledger firmware.

2. We need a root fingerprint, e.g. due to how PSBT works. The current PR is broken about this: it's not getting the fingerprint at the correct path. Ideally, Ledger should either provide an API to get the root fingerprint directly, or whitelist a path that makes it possible to calculate it.

smogm:
@SomberNight 
About 1.: That is true, but as the key derivation path is still the same, the generated keys of the hw device are the same and thus creating a new wallet file with this "new" password path will lead to the same keys.
So creating a new wallet is enough.

SomberNight:
@btchip Yes; re (1), the wallet files that are encrypted "with a hardware device" have their password set to the pubkey at that hardcoded hardened derivation path `m/4541509h/1112098098h` (chaincode, etc is not used). The idea is that even if users expose arbitrary xpubs at paths they use for coins e.g. for accounting purposes, they would likely not leak this pubkey.

btchip:
ok, I'll do that over the week end and test a bit

kodxana:
@btchip Any updates?

btchip:
A new Bitcoin application version has been prepared (https://github.com/LedgerHQ/app-bitcoin/commit/08997a31d840c313aa2335d46c989d815335ba3e) and will be reviewed by the team shortly

btchip:
Bitcoin application 1.4.8 has been deployed and should fix this issue

SomberNight:
@btchip Great, thanks. I've just tested and it looks ok.

I've found another instance not mentioned above, during wallet creation / opening of encrypted wallets, that still triggers with 1.4.8. Worked around that in https://github.com/spesmilo/electrum/commit/e60aede77ed4cccede0c8416c039ea7ad204c40d
So no warnings any more :)

PR #5533, opened 2019-07-27
Add support for Satochip hardware wallet

Toporin:
This is a pull request to add support for the Satochip hardware wallet.
(this pull request supersedes #5523 )

The Satochip hardware wallet is based on a jacavard smartcard and is fully open-source.
The wallet is composed of a javacard applet (https://github.com/Toporin/SatochipApplet) that is to be loaded on the smartcard, and an Electrum client plugin (https://github.com/Toporin/electrum-satochip) that acts as the interface between the card and the network.
The (optional) Satochip-2FA plugin allows to confirm transaction requests on a second android device.

More info:
https://github.com/Toporin/ (official repository)
https://prezi.com/p/mpq-xhh3mxjl/satochip-gent-meetup/ (Slides from previous meetups in Belgium)

PR #6520, opened 2020-08-18
Add support for Satochip hardware wallet

Toporin:
This is a pull request to add support for the Satochip hardware wallet.

The Satochip hardware wallet is based on a jacavard smartcard and is fully open-source.
The wallet is composed of a javacard applet (https://github.com/Toporin/SatochipApplet) that is to be loaded on the smartcard, and an Electrum client plugin (https://github.com/Toporin/electrum-satochip) that acts as the interface between the card and the network. The (optional) Satochip-2FA plugin allows to confirm transaction requests on a second android device before signing them.

This pull request supersedes #5533. Changes compared to previous PR:
* support for Electrum v4.0.2 instead of v3.3.8
* use pysatochip v0.11.2 with secure channel support

More info:
https://github.com/Toporin/ (official repository)
https://pypi.org/project/pysatochip/ (pysatochip library)
https://prezi.com/p/mpq-xhh3mxjl/satochip-gent-meetup/ (Slides from previous meetups in Belgium)
https://t.me/Satochip (Telegram support)


contrib/build-wine/prepare-wine.sh
```diff
@@ -17,6 +17,14 @@ PYINSTALLER_REPO="https://github.com/SomberNight/pyinstaller.git"
 PYINSTALLER_COMMIT="e934539374e30d1500fcdbe8e4eb0860413935b2"
 # ^ tag 3.6, plus a custom commit that fixes cross-compilation with MinGW
 
+# pyscard prebuilt binaries for Satochip 
+# PYSCARD_FILENAME=pyscard-1.9.9-cp36-cp36m-win32.whl  # python 3.6, 32-bit
+# PYSCARD_URL=https://github.com/cculianu/Electron-Cash-Build-Tools/releases/download/v1.0/pyscard-1.9.9-cp36-cp36m-win32.whl
+# PYSCARD_SHA256=99d2b450f322f9ed9682fd2a99d95ce781527e371006cded38327efca8158fe7
+PYSCARD_FILENAME=pyscard-1.9.9-cp37-cp37m-win32.whl # python 3.7, 32-bit
+PYSCARD_URL=https://ci.appveyor.com/api/buildjobs/f9cmce4j8hkau9n4/artifacts/dist/pyscard-1.9.9-cp37-cp37m-win32.whl
+PYSCARD_SHA256=3f7d52dd6694dd369b02e797fe1a3e39b63cf1d1c4b5fc0e1341aafa24f87e7a
```
SomberNight, line 26: self-note:
see previous discussion at https://github.com/spesmilo/electrum/pull/5523#discussion_r306469215
hash `3f7d52dd6694dd369b02e797fe1a3e39b63cf1d1c4b5fc0e1341aafa24f87e7a` matches corresponding file from "official" appveyor build:
https://ci.appveyor.com/project/LudovicRousseau/pyscard/build/job/f9cmce4j8hkau9n4/artifacts

SomberNight:
I have been testing this PR today.
(self-note: trivial rebase [branch here](https://github.com/SomberNight/electrum/tree/pr/6520))

-----

On Ubuntu, I have tested basic functionality, and it works.
(self-note re deps: `sudo apt-get install pcscd libpcsclite-dev swig`)
I haven't tested with 2FA; but wallet creation/restoration and tx signing works, for a standard wallet.

-----

On Windows, I could not get it to work. It seems it can communicate with the device but it seems to fail quite early in the flow.
This is using Python 3.8.3 AMD64 on Win10-1909, with `pyscard 1.9.9` and `pysatochip 0.11.3`.
```
D | pysatochip.SecureChannel | In initiate_secure_channel()
I | plugins.satochip.satochip | [SatochipClient] client request: update_status
I | plugins.satochip.satochip | [SatochipClient] self.handler is None! 
I | pysatochip.CardConnector | +Inserted: 3B FD 13 00 00 81 31 FE 15 80 73 C0 21 C0 57 59 75 62 69 4B 65 79 40
  File "...\Python38\site-packages\smartcard\CardMonitoring.py", line 179, in run
    self.observable.notifyObservers(
  File "...\Python38\site-packages\smartcard\Observer.py", line 62, in notifyObservers
    observer.update(self, arg)
  File "...\Python38\site-packages\pysatochip\CardConnector.py", line 82, in update
    self.cc.cardservice.connection.connect()
  File "...\Python38\site-packages\smartcard\CardConnectionDecorator.py", line 54, in connect
    self.component.connect(protocol, mode, disposition)
  File "...\Python38\site-packages\smartcard\pcsc\PCSCCardConnection.py", line 123, in connect
    raise CardConnectionException(
smartcard.Exceptions.CardConnectionException: Unable to connect with protocol: T0 or T1. The smart card cannot be accessed because of other connections outstanding. 

I | plugins.satochip.satochip | [SatochipClient] client request: show_error
I | plugins.satochip.satochip | [SatochipClient] self.handler is None! 
D | pysatochip.CardDataParser | In parse_initiate_secure_channel
E | gui.qt.installwizard.InstallWizard | 
  File "...\electrum\electrum\base_wizard.py", line 366, in on_device
    client = self.plugin.setup_device(device_info, self, purpose)
  File "...\electrum\electrum\plugins\satochip\satochip.py", line 516, in setup_device
    client = devmgr.client_by_id(device_id)
  File "...\electrum\electrum\plugin.py", line 486, in client_by_id
    self.scan_devices()
  File "...\electrum\electrum\plugin.py", line 402, in func_wrapper
  File "...\electrum\electrum\plugin.py", line 709, in scan_devices
    if pair in pairs and client.has_usable_connection_with_device():
  File "...\electrum\electrum\plugins\satochip\satochip.py", line 113, in has_usable_connection_with_device
    (response, sw1, sw2)=self.cc.card_select() #TODO: something else? get ATR?
  File "...\Python38\site-packages\pysatochip\CardConnector.py", line 221, in card_select
    self.card_initiate_secure_channel()
  File "...\Python38\site-packages\pysatochip\CardConnector.py", line 955, in card_initiate_secure_channel
    peer_pubkey = self.parser.parse_initiate_secure_channel(response)
  File "...\Python38\site-packages\pysatochip\CardDataParser.py", line 131, in parse_initiate_secure_channel
    data_size = ((response[0] & 0xff)<<8) + (response[1] & 0xff)
IndexError: list index out of range
```

SomberNight (changes requested):
You are abusing Electrum seeds. Electrum seeds are [versioned by design](https://electrum.readthedocs.io/en/latest/seedphrase.html): based on the seed itself the wallet structure including derivation paths and script types are determined.

This is contrary to how BIP39 works where the seed itself only determines some root entropy.
BIP39 in itself is insufficient to define a wallet, it is typically used in conjunction with BIP44 (and similar).

In current PR, when restoring from an Electrum seed, you allow selecting an arbitrary derivation path and script type. It is treated ~as if it was BIP39 (except the root entropy is derived as for Electrum seeds).

Also, when creating a new seed (device init), you generate an Electrum seed (of "standard" type), however on the next screen you allow the user to select an arbitrary derivation and script type.
Imagine the user lost/broke their Satochip hardware device. If they tried to restore from seed in software, they would see a different wallet. Electrum will not ask the user for derivation path and script type, it will already know these based on the version number encoded in the seed; except the plugin disregards that versioning.

Long term this could only lead to headaches and loss of user funds. It cannot stay.

If you want arbitrary derivation paths and script types, I suggest disallowing using Electrum seeds for Satochip completely.
You can still support both BIP39 and Electrum seeds if you want, but it is crucial that when using Electrum seeds always the expected wallet is derived: i.e. creating a software wallet from the same seed must result in the same wallet as if the seed is used with Satochip.

SomberNight:
Also note that macOS support is missing here.
(some related commits here: https://github.com/Electron-Cash/Electron-Cash/pull/1668/commits/3584cc8edce8b52db6f0f42eeb2017662b7c3117, https://github.com/Electron-Cash/Electron-Cash/commit/a93169b2cf1359f7521f4bccae03540092f3e239, https://github.com/Electron-Cash/Electron-Cash/commit/3cda78a40278cbb7433ea10469f6f5ed3e74dc36)

Toporin:
Thanks for your feedback!

Regarding the issue with card communication on Windows, we will try to reproduce the error. It may be caused by the application trying to send commands to the card before low-level communication is established with the card.

Regarding Electrum seeds, you are right, we completely overlooked that point. 

Digging deeper into the issue, here are a few observations (please correct if I am wrong):
* Most (all?) hardware wallets use BIP43-style derivation paths (like m / purpose' / coin_type' / account' / change / address_index), in combination with BIP39 seeds
* Electrum uses BIP43 derivation paths for all hardware wallets, AND for BIP39 seeds imported in software wallets
* Electrum seeds (not BIP39) use seed versioning instead of purpose derivation path. 
* Software wallets with Electrum seeds use "m / change / address_index" derivation paths

Consequently:
* Addresses derived from the same Electrum seed will differ when used in a software wallet versus a Satochip (as the derivation paths are different)
* On the other hand with BIP39 seeds, the same addresses are derived both for software and hardware wallets (both use BIP43-style derivation path)
* If Satochip uses Electrum-seed compatible derivation paths ("m / change / address_index"), it will end up reusing the same private keys for different Electrum forks such as Electrum-LTC 
* If Satochip uses Electrum-seed but with BIP43-style derivation paths, it will be non-standard and not compatible with other hardware and software wallets

**This leaves us with mainly one option for Satochip: to support only BIP39 seeds, which would ensure compatibility with hardware and software wallets. In this case however, we would need to be able to initialize a new Satochip with Electrum, thus generating BIP39 seeds on Electrum specifically for Satochip. Would that be acceptable for you?**

SomberNight:
> Software wallets with Electrum seeds use "m / change / address_index" derivation paths

That's the path "standard" Electrum seeds; "segwit" ones have [another field distingushing multisig](https://github.com/spesmilo/electrum/blob/5337331fa00ab2ca6767f1a07fcda34099a01c38/electrum/keystore.py#L989-L1000).

> If Satochip uses Electrum-seed compatible derivation paths ("m / change / address_index"), it will end up reusing the same private keys for different Electrum forks such as Electrum-LTC

Yes, Electrum seeds were not designed with altcoins in mind.

> This leaves us with mainly one option for Satochip: to support only BIP39 seeds, which would ensure compatibility with hardware and software wallets.

That's fine.

> In this case however, we would need to be able to initialize a new Satochip with Electrum, thus generating BIP39 seeds on Electrum specifically for Satochip. Would that be acceptable for you?

Yes. It should be made clear to the user, when the seed is displayed, that it is a BIP39 seed and not an Electrum seed.

Toporin:
I have updated the installation wizard for satochip, so that BIP39 seeds are now the default for Satochip. The user can now choose between creating and importing a a BIP39 seed. When creating a new seed, a BIP39 seed is generated. When importing an existing seed, Electrum seeds are also accepted but a warning is shown to the user to recommend using BIP39 for compatibility with other hardware wallet. I hope this new wizard is sufficiently explicit for the user. 

I have also tried to run Electrum with Satochip from sources on windows, but I could not reproduce your issue (Python 3.8.3, using pyscard 1.9.9 and Pyscard 2.0.0). I usually run from the .exe build on Windows rather than from the python source as it is more reproducible.

Toporin:
This pull request is superseded by #7518.

Issue #6529, opened 2020-08-23
Calling payto too frequently results in same transactions broadcasted multiple times


MrNaif2018:
When using payto+broadcast command multiple times without delays (sometimes up to 5 seconds), payto command seems to build the same raw transaction.
If rbf is enabled (by default it is), when broadcasting, transaction is replacing previous one. But as it has same settings, same fee calculated, etc., it is rejected. Error "insufficient fee" is returned, which is confusing.
In electrum debug logs it says actual message from bitcoind:
"Transaction not passing network rules. new fee <= old fee"
When rbf is off, usually tx conflict error is returned.
In some cases it is possible that broadcast function can broadcast same transaction multiple times, returning same txid, and then failing.
This error is easy to reproduce, assuming there is a function pay_to:

```
def pay_to(address, amount):
    tx_data = payto(address, amount)
    return broadcast(tx_data)
```

To reproduce, pay_to needs to be called 2 times in a row or a bit more, i.e.:
`pay_to("address", 0.1)
pay_to("address", 0.1)`
I have found this error on regtest initially, but I was able to reproduce it on testnet and mainnet too.
Because of that issue it is hard to use payto/paytomany in concurrent environment (for example, from multiple threads).

SomberNight:
The tx is only added to the wallet history when the wallet learns of it from the server, after it broadcasts the tx.
So there is at the very least a full network round trip's worth of delay.

Workaround:
call `addtransaction` between `payto` and `broadcast`

Though this will clearly NOT provide atomicity, so there could still be a race, but it would be orders of magnitudes less likely.

MrNaif2018:
Thanks, adding addtransaction during broadcast function helped. It failed after 25 txes (too-long-mempool-chain), but it is expected I think.
Should I open a PR to fix broadcast function, or is there a better fix?

SomberNight:
`broadcast` should be independent of any wallet; it should be possible to broadcast any transaction you obtained in any way

perhaps `payto` could take a flag whether to have a side effect on the wallet and add the tx

PR #6532, opened 2020-08-25
Added for_broadcast argument to payto/paytomany

MrNaif2018:
This PR is a fix discussed in #6529
`for_broadcast` argument added is to determine whether to also call `addtransaction` or not.
@SomberNight

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -615,7 +615,7 @@ class Commands:
             locktime=locktime)
         result = tx.serialize()
         if addtransaction:
-            await self.addtransaction(result)
+            await self.addtransaction(result, wallet=wallet)
         return result
 
     @command('wp')
@@ -646,7 +646,7 @@ class Commands:
             locktime=locktime)
         result = tx.serialize()
         if addtransaction:
-            await self.addtransaction(result)
+            await self.addtransaction(result, wallet=wallet)
         return result
 
     @command('w')
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -592,7 +592,7 @@ class Commands:
 
     @command('wp')
     async def payto(self, destination, amount, fee=None, feerate=None, from_addr=None, from_coins=None, change_addr=None,
-                    nocheck=False, unsigned=False, rbf=None, password=None, locktime=None, addtransaction=True, wallet: Abstract_Wallet = None):
+                    nocheck=False, unsigned=False, rbf=None, password=None, locktime=None, addtransaction=False, wallet: Abstract_Wallet = None):
         """Create a transaction. """
         self.nocheck = nocheck
         tx_fee = satoshis(fee)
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks for the PR. We believe that `for_broadcast` is not a good name for that.
could you rename that flag to be the same as the command, ie `--addtransaction`, and remove the -B short name?

MrNaif2018:
Renamed.

electrum/commands.py
```diff
@@ -592,7 +592,7 @@ async def verifymessage(self, address, signature, message):
 
     @command('wp')
     async def payto(self, destination, amount, fee=None, feerate=None, from_addr=None, from_coins=None, change_addr=None,
-                    nocheck=False, unsigned=False, rbf=None, password=None, locktime=None, wallet: Abstract_Wallet = None):
+                    nocheck=False, unsigned=False, rbf=None, password=None, locktime=None, addtransaction=True, wallet: Abstract_Wallet = None):
```
SomberNight, line 595: The default should be "False", as "True" would be a significant change in behaviour.

MrNaif2018: Just a question: why exactly is transaction added to wallet after a full network round trip? I can see it not only from CLI, but from GUI. When sending a transaction, it tells tx hash, and only after some time (not significant of course) it is showing "synchronizing (3/3)" and then showing that in GUI + notification.
As to review comment, will change it now

SomberNight: It works the way it works due to historical reasons.
There used to be no "local" transaction support; the tx history in the wallet used to always mirror the last connected server.

It could be changed now but the implications should be considered. For example, what if the tx is rejected by the server? (e.g. imagine the user creating a non-standard tx, rejected by bitcoind policy, e.g. creating dust outputs - I suspect that sort of scenario might result in increased support load).

MrNaif2018: Hmm, I am unsure but if broadcast fails then no synchronizing step is done. This fix seems to be more of a workaround than a real solution for me, but it works.

electrum/commands.py
```diff
@@ -641,7 +644,10 @@ async def paytomany(self, outputs, fee=None, feerate=None, from_addr=None, from_
             rbf=rbf,
             password=password,
             locktime=locktime)
-        return tx.serialize()
+        result = tx.serialize()
+        if addtransaction:
+            await self.addtransaction(result)
```
SomberNight, line 649:
```suggestion
            await self.addtransaction(result, wallet=wallet)
```

electrum/commands.py
```diff
@@ -613,11 +613,14 @@ async def payto(self, destination, amount, fee=None, feerate=None, from_addr=Non
...
```
SomberNight, line 618:
```suggestion
            await self.addtransaction(result, wallet=wallet)
```

SomberNight (changes requested):
I have tested this now. It does not work because `addtransaction` needs `wallet` passed to it.

After fixing that, we can merge.

electrum/commands.py
MrNaif2018: Hmm, I tested that on CLI with regtest and it worked, approximately like so:
`./run_electrum daemon start`
`./run_electrum load_wallet`
and then I just used
`./run_electrum payto x y`
and checked that it was added via `./run_electrum onchain_history`.
Didn't give errors. I think it should work because of daemon passing arguments (command decorator), or am I missing something?

SomberNight: I had tested on master and it did not work.
The command decorator cannot set the wallet arg correctly in this case as it mostly does its magic based on kwargs which will be empty in this case.

MrNaif2018: Ok, will apply the fixes.

PR #6546, opened 2020-09-01
Fix for qt local timezone

relativisticelectron:
Problem: 
- the Date field in the Locktime is prefilled with the local time.  However the time is then actually used as a UTC time.

Solution: 
- Replace  QDateTime.currentDateTime  by QDateTime.currentDateTimeUTC.

### DIFF
--- a/electrum/gui/qt/locktimeedit.py
+++ b/electrum/gui/qt/locktimeedit.py
@@ -33,7 +33,7 @@ class LockTimeEdit(QWidget):
         self.editors = [self.locktime_raw_e, self.locktime_height_e, self.locktime_date_e]
 
         self.combo = QComboBox()
-        options = [_("Raw"), _("Block height"), _("Date")]
+        options = [_("Raw"), _("Block height"), _("Date (UTC)")]
         option_index_to_editor_map = {
             0: self.locktime_raw_e,
             1: self.locktime_height_e,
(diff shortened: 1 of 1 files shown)


SomberNight:
>  However the time is then actually used as a UTC time.

Where?

The code on master looks okay to me. The dates in `locktimeedit.py` and `transaction_dialog.py` are all local times.

I think current PR as-is actually breaks things. `LockTimeDateEdit.get_locktime()` is supposed to return a unix timestamp.
https://github.com/spesmilo/electrum/blob/c9bf1d4c800d8a07ae04de97f27cee51707f10de/electrum/gui/qt/locktimeedit.py#L170-L173

Before PR:
```
import time
from PyQt5.QtCore import QDateTime
dt = QDateTime.currentDateTime().toPyDateTime()
unix_time = int(time.mktime(dt.timetuple()))
```

After PR:
```
import time
from PyQt5.QtCore import QDateTime
dt = QDateTime.currentDateTimeUtc().toPyDateTime()
unix_time = int(time.mktime(dt.timetuple()))
```
Check the values for `unix_time`, "Before PR" looks correct to me.

relativisticelectron:
Thanks for the super fast feedback. You are right and this should not be merged. 
I will investigate further, why there was the UTC timeshift before the transaction was able to be accepted by the bitcoin node.

relativisticelectron:
On testnet I get the same problem:
image
It seems to be not related to electrum at all.  And it's also not related to my bitcoin node (because I used others testnet electrum servers).

SomberNight:
> before the transaction was able to be accepted by the bitcoin node.

oh you did not mention consensus/policy. In that case I would guess what you saw was because of "median time past"
see [BIP113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki)

relativisticelectron:
> > before the transaction was able to be accepted by the bitcoin node.
> 
> oh you did not mention consensus/policy. In that case I would guess what you saw was because of "median time past"
> see [BIP113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki)

Thanks for the link. That explains it. I am sorry for the wasted time.

Issue #6554, opened 2020-09-04
crash when closing 1 of 2 running instances (hidapi, libusb, threading, segfault)


jlopp:
Electrum 4.0.2 appimage on debian

Create an 8 of 8 multisig with 8 hardware devices: trezor one, trezor T, Nano S, Nano X, BitBox01, BitBox02, Coldcard Mk1, Coldcard Mk2

Create a 1 of 8 multisig with same 8 hardware devices

Open the 8 of 8 wallet and unlock all 8 plugged in devices

Then open the 1 of 8 wallet.

The close the 8 of 8 wallet and wait for a few seconds until both instances crash. I've repeated this process several times and get different errors.

```
$ ~/Downloads/apps/electrum-4.0.2-x86_64.AppImage --testnet
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
E | plugin.DeviceMgr | failed to create client for bitbox02 at b'0001:0009:01': Exception('No BitBox02 detected')
E | plugins.coldcard.coldcard.Coldcard_KeyStore | 
Traceback (most recent call last):
  File "/tmp/.mount_electrcFBIsQ/usr/lib/python3.7/site-packages/electrum/plugins/coldcard/coldcard.py", line 383, in sign_transaction
    resp = client.sign_transaction_poll()
  File "/tmp/.mount_electrcFBIsQ/usr/lib/python3.7/site-packages/electrum/plugins/coldcard/coldcard.py", line 233, in sign_transaction_poll
    return self.dev.send_recv(CCProtocolPacker.get_signed_txn(), timeout=None)
  File "/tmp/.mount_electrcFBIsQ/usr/lib/python3.7/site-packages/ckcc/client.py", line 139, in send_recv
    buf = self.dev.read(64, timeout_ms=(timeout or 0))
  File "hid.pyx", line 123, in hid.device.read
OSError: read error
double free or corruption (fasttop)
Aborted
```

```
$ ~/Downloads/apps/electrum-4.0.2-x86_64.AppImage --testnet
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
E | plugin.DeviceMgr | failed to create client for bitbox02 at b'0001:0009:01': Exception('No BitBox02 detected')
[...]

PR #6561, opened 2020-09-06
hardware devices: run all device communication on dedicated thread

SomberNight:
hidapi/libusb etc are not thread-safe.

related: #6554


jlopp:
I'm unable to reproduce the crashes when running this code; seem to be a fix!

PR #6567, opened 2020-09-09
gui: update stdio and text after db and storage separation

Grimler91:
Hi,

e1ce3aace7e3143da24115890d9bae78a9f5bca updated the qt and kivy guis, but not stdio or text one, so trying to use the text gui gives an error `missing 1 required positional argument: 'storage'`, coming from the line
```
self.wallet = Wallet(storage, config=config)
```
and `electrum --gui stdio` just closes without any output.

With this patch electrum starts as it should for both guis again.


SomberNight:
Thanks

Issue #6018, opened 2020-03-05
Windows Install: Camera Function not working (QR codes, zbar issue, scan, scanner)


BittBurger:
I apologize in advance, as I have zero experience with GitHub but was directed here.  So any rules I break or content I inadvertantly omit, please forgive.  I'm just going to write this in "longhand" and plain english:

I have two laptops, both with Windows 10 installed.  On both laptops, when clicking the "Send" tab, and then the black camera icon in the top-right corner, two popup windows appear, and one contains error text.  I am given the option to retry, abort, or ignore.  No matter which option I select, the program closes fully.  Screen cap attached.
![photo_2020-03-04_14-41-19](https://user-images.githubusercontent.com/9684377/76016832-ec22f800-5ef3-11ea-8622-8c8dd558bdd0.jpg)

As this has been checked on two unrelated machines I will make the assumption that it's fairly easy to duplicate on any machine, but, version is 3.3.8 and windows camera driver information:
Intel AV Stream Camera 2500.   Driver:  Intel  / Version 30.15063.6.8611 built in camera into top of laptop screen.

I don't know why this would be relevant but its weirdly relevant in several other app bugs I've seen:  I have Norton Internet Security Version 22.20.1.69 and Malwarebytes 4.0.4.49 / Update Package 1.0.20268 / Component Package 1.0.823 installed.

EDIT: text on image:
```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Assertion failed!

Program: ...
File: zbar/video/dshow.c
Line: 518

Expression: img->datalen == bufferlen

For information on how your program can cause an assertion
failure, see the Visual C++ documentation on asserts

(Press Retry to debug the application - JIT must be enabled)
---------------------------
Abort   Retry   Ignore   
---------------------------
```

MrNaif2018:
From investigating it looks like it might be zbar problem, I didn't find that issue in electrum's issues list(I was helping BittBurger), but I found ElectronCash had that issue before, it looks like they solved that by using native QtMultimedia functions instead of zbar(which is from 2012, which looks like might not support newer cameras).
Although there is a newer zbar version as from discussion in related issue:
https://github.com/mchehab/zbar/
Related issues and PRs:
https://github.com/Electron-Cash/Electron-Cash/issues/1282 (that same error)
https://github.com/Electron-Cash/Electron-Cash/pull/1376 (PR which probably fixed it)

SomberNight:
Thanks for the report.
[...]

PR #6593, opened 2020-09-14
win binary: build zbar ourselves

SomberNight:
This allows bundling much newer zbar that includes many fixes.
related: #6018

This is largely based on
https://github.com/Electron-Cash/Electron-Cash/pull/1362
https://github.com/Electron-Cash/Electron-Cash/pull/1363
https://github.com/Electron-Cash/Electron-Cash/pull/1365
https://github.com/Electron-Cash/Electron-Cash/commit/eda015908e9d6ea9a0adfbda9db55b929c0926ba

(note: tested that reproducible build still works: got same hashes for win binaries on an Ubuntu and a Manjaro host)

### DIFF
--- a/contrib/build-wine/Dockerfile
+++ b/contrib/build-wine/Dockerfile
@@ -17,8 +17,11 @@ RUN apt-get update -q && \
         p7zip-full=16.02+dfsg-6 \
         make=4.1-9.1ubuntu1 \
         mingw-w64=5.0.3-1 \
+        mingw-w64-tools=5.0.3-1 \
+        win-iconv-mingw-w64-dev=0.0.8-2 \
         autotools-dev=20180224.1 \
         autoconf=2.69-11 \
+        autopoint=0.19.8.1-6ubuntu0.3 \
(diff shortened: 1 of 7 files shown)


SomberNight:
Note: I wonder if we should build zbar for the AppImage too, instead of using the old version in apt.
Well, if someone complains that QR scanning does not work for them using the AppImage, then we can try that.

SomberNight:
Maybe we could use it for macOS as well...
https://github.com/Electron-Cash/Electron-Cash/commit/b2b737001c8cc41a38fa580ea252a6d24e08f5d5
https://github.com/Electron-Cash/Electron-Cash/commit/163224cf1fad3af63f2d3cbe68a34fb8ff279af6
https://github.com/Electron-Cash/Electron-Cash/commit/34b1ccd0adf0e2b096157f4c3838e1da29971ef7
Although I am unsure if we would need https://github.com/Electron-Cash/Electron-Cash/pull/1376 for that, which is a much larger change.

PR #6595, opened 2020-09-15
Pass wallet to invoice_status/request_status

MrNaif2018:
This PR returns passing wallet to `request_status` (and `invoice_status` for symmetry) events.
It is useful to track which wallet changed request/invoice refers to in event handler (useful for payment processing, complex event forwarding, etc.)
One question: should I change `self.wallet` to `wallet` in some event handlers?

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -318,7 +318,7 @@ class PayServer(Logger):
         # FIXME specify wallet somehow?
         return list(self.daemon.get_wallets().values())[0]
 
-    async def on_payment(self, evt, key, status):
+    async def on_payment(self, evt, wallet, key, status):
         if status == PR_PAID:
             self.pending[key].set()
 
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -249,7 +249,7 @@ class ElectrumWindow(App):
     def on_fee_histogram(self, *args):
         self._trigger_update_history()
 
-    def on_request_status(self, event, key, status):
+    def on_request_status(self, event, wallet, key, status):
         if key not in self.wallet.receive_requests:
             return
         self.update_tab('receive')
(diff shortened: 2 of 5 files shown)


SomberNight (commented):
> should I change self.wallet to wallet in some event handlers?

No. Instead it should be tested, at least in the Qt GUI, that wallet == self.wallet; otherwise return early.

SomberNight:
> This PR returns passing wallet to request_status

Was there a time wallet was passed with these events?

I can only find a brief back-and-forth; the commit history for this is a bit of a mess.

https://github.com/spesmilo/electrum/commit/13317c2f5139f1f1873c7fae1861b5d3031b690d
https://github.com/spesmilo/electrum/commit/0070e5036c1c54c2bd29911de218e92709e9795a
https://github.com/spesmilo/electrum/commit/e2c2c8998870ac48230890c3d9adc56838d67eb5

@ecdsa do you remember what was going on here? :P shall we add wallet as an arg?

MrNaif2018:
Yes, it was, at least in 3.3.8 via payment_received event
https://github.com/spesmilo/electrum/blob/1773bd6cd6b9aaa693c6c1cb865f88fc3597e09f/electrum/wallet.py#L1393
It was removed (maybe accidentally) by this commit: e2c2c8998870ac48230890c3d9adc56838d67eb5

SomberNight:
I don't think `payment_received` was in 3.3.8 but ok.

MrNaif2018:
Tarball build failure seems to be unrelated to PR

SomberNight:
> @ecdsa do you remember what was going on here? :P shall we add wallet as an arg?

I guess it was removed because we did not have a direct need for it; but if someone else finds it useful, we can have it.

Also, actually, re 'invoice_status' I think if a user has multiple LN wallets open, and tries to pay the same invoice from more than one; there might be interference without this change.

MrNaif2018:
Thank you for merging!

PR #6607, opened 2020-09-18
Keep console input when switching server

verretor:
Console input was being replaced by an empty line every time the server used changed.

Before:
```
>>> for i in range(0, 3):
...     print(i)
Welcome to electrum.blockstream.info - no logs and Tor support - Please learn about running your own server!
... 
```

After:
```
>>> for i in range(0, 3):
...     print(i)
Welcome to electrum.blockstream.info - no logs and Tor support - Please learn about running your own server!
...     print(i)
... 
0
1
2
>>> 
```


SomberNight:
Looks good; thank you.

PR #6615, opened 2020-09-24
Fix dumpgraph command: give out json encoded nodes and channels

bitromortac:
The `dumpgraph` command is broken at the moment, which is fixed by this PR.
This PR adds channels to the output to have a complete view of the whole graph
(excluding one own's node) in terms of a dict, which is json encodeable.

Tested with the Qr-console and with the command line interface.

The `ShortChannelID` is currently encoded to a hex string, as it came from [bytes](https://github.com/spesmilo/electrum/blob/52bd0eb1a656bdd0eb32340350fb4bf82af5c690/electrum/util.py#L258). Should we
print out a more human readable form, we could use the x notation or the int version that lnd uses?

As a side note, I observe that channels seem have no capacity assigned (tested on LN testnet).

### DIFF
--- a/electrum/channel_db.py
+++ b/electrum/channel_db.py
@@ -37,7 +37,7 @@ from enum import IntEnum
 
 from .sql_db import SqlDB, sql
 from . import constants, util
-from .util import bh2u, profiler, get_headers_dir, bfh, is_ip_address, list_enabled_bits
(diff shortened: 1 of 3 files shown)


bitromortac:
Updated Address(NamedTuple) to NodeAddress to be less ambiguous.

PR #6617, opened 2020-09-25
Save indented lines in console history

verretor:
Mimic the behavior of the standard Python console by storing indented lines in history.


ecdsa:
why? the point of thif `if` statement is to be able to enter commands that will not be saved in your history.
(`bash` has the same feature)

verretor:
Example:
```
>>> for i in range(0, 3):
...     print(i)
```

The line `print(i)` is not saved otherwise.

I will try to make it so that it saves ```print(i)``` but not ```>>> importprivkey()```

verretor:
Done.

PR #6624, opened 2020-09-29
network dialog: include protocol in server address field

### LINKED ISSUE #6576: Electrum reverts to server type with SSL
If I specify a server manually in the Network->Overview page, by filling in the "Server:" field, with "Select server automatically" unchecked, then it always configures that server as SSL even if I set port 50001 (non SSL).

In the .electrum/config file the `"server": "127.0.0.1:50001:s",` line is set with `s`. I wonder if it wouldn't be better to set have a checkbox to select SSL type on/off? I have to close Electrum, and go manually change this in the config file. But it seems to revert periodically as well.

bitromortac:
Fixes #6576.

When a server address is entered without the protocol (`t`/`s`), we guess the preferred protocol to use (`s`) and configure it.
It is also possible to give the string `host:port:protocol` as a connection string as the server. On update of the server field,
only the server address and port are written back into the server field.

In #6576, a server is entered to use the plain protocol. Focus loss of the field triggers a config update (working correctly) and the field is updated, removing the protocol part. Because the protocol was missing due to autofill, the protocol guessing gets active and wrongly selects again the `s` protocol and configures it. Thanks to @SomberNight for the hints.

In this PR, the full connection string is written back to the server field, which fixes the described behavior above. This enables a user again to change from the `s` to `t` protocol manually (for experienced users), while having also the possibility of protocol guessing, when only the host and port are entered (for more unexperienced users).

Possible alternatives:
* introduce `s`/`t` checkbox

### DIFF
--- a/electrum/gui/qt/network_dialog.py
+++ b/electrum/gui/qt/network_dialog.py
@@ -327,7 +327,7 @@ class NetworkChoiceLayout(object):
         server = net_params.server
         auto_connect = net_params.auto_connect
         if not self.server_e.hasFocus():
-            self.server_e.setText(str(server))
+            self.server_e.setText(server.to_friendly_name())
         self.autoconnect_cb.setChecked(auto_connect)
 
         height_str = "%d "%(self.network.get_local_height()) + _('blocks')
--- a/electrum/gui/qt/network_dialog.py
+++ b/electrum/gui/qt/network_dialog.py
@@ -327,7 +327,7 @@ class NetworkChoiceLayout(object):
         server = net_params.server
         auto_connect = net_params.auto_connect
         if not self.server_e.hasFocus():
-            self.server_e.setText(server.net_addr_str())
+            self.server_e.setText(str(server))
         self.autoconnect_cb.setChecked(auto_connect)
 
         height_str = "%d "%(self.network.get_local_height()) + _('blocks')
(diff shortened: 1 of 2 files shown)


ecdsa:
I am not sure if #6576 should be considered as a bug.
The choice of the protocol was removed from the GUI a long time ago, because users were randomly disabling SSL.
The idea is that SSL should be disabled if only if a server string is passed from the command line

bitromortac:
Ok, I see that's well understandable, now it's just the question if we want the user to be able to configure a plain server (which is not recommended). From the perspective of disabling SSL randomly by the user, I think it is less likely to occur because there's no gui element the user can click on, no binary choice, but the user has to particularly target the protocol string to mess up.

SomberNight:
> I am not sure if #6576 should be considered as a bug.

The issue described there is definitely a bug: simply opening the network dialog can change the network protocol used.

We should aim to hide complexity from inexperienced users, so instead of current PR why don't we implement a new method `ServerAddr.to_friendly_name()` (or similar) which hides `:s` but shows `:t` and call that here.
That is, if the protocol is `s`, hide it; otherwise show it.
This way, the change here would be invisible to almost everyone, except people who have manually configured using plaintext TCP.

ROSAISS81:
Tengo que mandar esto

ROSAISELAPEREZMARTINEZ

hendeyk:
Kell merge gofthub

PR #6633, opened 2020-10-05
tx: add deserialize_if_need decorator, fix locktime/version properties

zebra-lucky:
- add `Transaction.deserialize_if_need`
- fix `locktime`, `version` (and as side effect `to_json` methods), when default values returned if no deserialization if done before


SomberNight:
Hmm. I see, so the issue is that an API user might create a `Transaction` object but not call `deserialize()` on it before accessing the `version`/`locktime` fields. This is already handled for the `inputs()` and `outputs()` methods.

I suggest we just copy the two respective lines from `inputs()`/`outputs()`, e.g.:
```
    @property
    def locktime(self):
        if self._inputs is None:
            self.deserialize()
        return self._locktime
```

zebra-lucky:
Ok, I'll redo with:
```
        if self._inputs is None:
            self.deserialize()
```

> Hmm. I see, so the issue is that an API user might create a Transaction object but not call deserialize()

Found during testing (local fork tests) when creating tx from raw and then testing `to_json`.

zebra-lucky:
Maybe with?
```
        if self._inputs is None and self._cached_network_ser is not None:
            self.deserialize()
```

SomberNight:
> `if self._inputs is None and self._cached_network_ser is not None:`

Actually `deserialize()` itself is already testing these, so we could simply call `deserialize()` without branching/if.

zebra-lucky:
remake commit with `deserialize`, renaming PR subject

SomberNight:
Thanks

PR #6641, opened 2020-10-08
wallet: implement cancelling tx by double-spending to self ("dscancel")

SomberNight:
Adds option in Qt GUI to "cancel" an unconfirmed RBF tx by double-spending its inputs to self.


SomberNight:
note: especially the kivy "dscancel" dialog is mostly a copy-paste of the "bump fee" dialog
I wonder if it would be better to deduplicate the code somehow, e.g. using a shared base class.

shivaenigma:
I request to remove this. I know this is currently possible by anyone just constructing a RBF using raw tx. However, explicitly giving an option to cancel transaction will result in very bad payment experience due to higher rate of cancelled transactions. 

Another case if when I spend unconfirmed coins, which are then cancelled. In this case my payment itself gets cancelled (which I didn't intend to do) without any warning and announcement to me. This is a big UI nightmare for users. Majority of people aren't technical enough to use coin control selection or force only to spend confirmed input

SomberNight:
> Another case is when I spend unconfirmed coins, which are then cancelled. In this case my payment itself gets cancelled (which I didn't intend to do) without any warning and announcement to me. This is a big UI nightmare for users. Majority of people aren't technical enough to use coin control selection or force only to spend confirmed input

The coinchooser always prefers to spend confirmed coins -- it only selects unconfirmed coins if it cannot be avoided.

> explicitly giving an option to cancel transaction will result in very bad payment experience due to higher rate of cancelled transactions.

Merchants already do not accept zero-conf RBF txs. Without an option to cancel, the user would get into the unfortunate situation of there being a transitional period where they are both out of their money and don't yet have a fulfilled invoice.

I made this PR because I personally have needed to cancel txs using console commands multiple times because of shitty merchant software and invoices expiring after 30-60 minutes while I had unconfirmed txs in the mempool.

Further, the mempool is unpredictable and low fee transactions sometimes linger around for days. At some point bumping the fee is not the solution anymore as the invoice / payment window might have expired by then. During mempool congestions we often get users asking for support to cancel such txs.

Also, Lightning works today. High value payments can use onchain txs and should wait for confirmations; low value payments should just use LN - this way they can even be instant.

When BIP-125 RBF was introduced into bitcoind years ago, AFAIU there was sort of a gentlemen's agreement between wallets not to add this kind of feature to the UI. Some part of the community was already outraged by the bitcoind policy change at the time, so no one wanted to put oil on the fire. However, we think that most of those people have probably left the community during the big block hard fork; and that Lightning changes the landscape. Also, a wallet that did not exist during the gentlemen's agreement, bluewallet, have actually unwittingly already broken it and have this "cancel tx" feature in its UI.

All things considering, I think this is a useful feature for users.

shivaenigma:
>The coinchooser always prefers to spend confirmed coins -- it only selects unconfirmed coins if it cannot be avoided.

You are talking about coinchooser of electrum. If the user is using another wallet like blockchain.info, we don't know what their algorithm does. Take example a novice user using blockchain wallet. He will see his balance go up and then suddenly go down after sometime. If he spends those funds in between its another nightmare for users and support teams 

>Merchants already do not accept zero-conf RBF txs
Yes merchants don't accept zero confirm txs. But some predictably in payments interface is required. If user is paying on checkout ,the order is confirmed on zero confirms and the user leaves the payment screen.  Payment can then confirm in hours/days after which the goods are delivered. One would expect the payment to confirm 99% of the time and it does 

I am not sure how widely this will be used. However, if payments  cancellation rate is high it will be a problem for all merchants and payment systems. Nobody like balances going up and down randomly and  waiting for payment to confirm to hours/days

At best this should be an advanced option in electrum console if you really require this to clear stuck transactions

shivaenigma:
Feel free to see the [discussion here](https://twitter.com/blockonomics_co/status/1318540752360648704)

PR #6643, opened 2020-10-10
Handle KeyboardInterrupt in Python Console

verretor:
Use Ctrl+C to raise a KeyboardInterrupt.
This is especially useful to escape constructs.

Example:
```
>>> for i in range(0, 3):
...
KeyboardInterrupt
>>>
```


SomberNight:
Thanks

PR #6645, opened 2020-10-12
android build: update list of apt deps for buildozer

SomberNight:
compare https://github.com/kivy/buildozer/blob/0.39/docs/source/installation.rst
and https://github.com/kivy/buildozer/blob/1.2.0/docs/source/installation.rst


SomberNight:
also see https://github.com/kivy/python-for-android/blob/5a94d0740e3a0623407b49b2cbb4ef106eea959e/Dockerfile#L51
which is still seemingly outdated

PR #6651, opened 2020-10-15
Use default sys.ps1 and ps2 as console prompts

verretor:
sys.ps1 and sys.ps2 define the strings used as primary and secondary prompts in the Python interpreter.

Also fix a rare bug introduced by myself in 7772af6 (#6607) where spaces at the end of the current line would not be remembered when switching server.

### DIFF
--- a/electrum/gui/qt/console.py
+++ b/electrum/gui/qt/console.py
@@ -15,6 +15,10 @@ from electrum.i18n import _
 
 from .util import MONOSPACE_FONT
 
+# sys.ps1 and sys.ps2 are only declared if an interpreter is in interactive mode.
+sys.ps1 = '>>> '
+sys.ps2 = '... '
+
 
 class OverlayLabel(QtWidgets.QLabel):
     STYLESHEET = '''
@@ -44,10 +48,9 @@ class OverlayLabel(QtWidgets.QLabel):
 
 
 class Console(QtWidgets.QPlainTextEdit):
-    def __init__(self, prompt='>>> ', parent=None):
+    def __init__(self, parent=None):
         QtWidgets.QPlainTextEdit.__init__(self, parent)
 
-        self.prompt = prompt
         self.history = []
         self.namespace = {}
         self.construct = []
@@ -86,7 +89,7 @@ class Console(QtWidgets.QPlainTextEdit):
         self.namespace.update(namespace)
 
     def showMessage(self, message):
-        curr_line = self.getCommand()
+        curr_line = self.getCommand(strip=False)
         self.appendPlainText(message)
         self.newPrompt(curr_line)
 
@@ -102,9 +105,9 @@ class Console(QtWidgets.QPlainTextEdit):
 
     def newPrompt(self, curr_line):
         if self.construct:
-            prompt = '... ' + curr_line
+            prompt = sys.ps2 + curr_line
         else:
-            prompt = self.prompt + curr_line
+            prompt = sys.ps1 + curr_line
 
         self.completions_pos = self.textCursor().position()
         self.completions_visible = False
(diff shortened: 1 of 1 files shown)


electrum/gui/qt/console.py
```diff
@@ -86,7 +89,9 @@ def updateNamespace(self, namespace):
         self.namespace.update(namespace)
 
     def showMessage(self, message):
-        curr_line = self.getCommand()
+        doc = self.document()
+        curr_line = doc.findBlockByLineNumber(doc.lineCount() - 1).text()
+        curr_line = curr_line[len(sys.ps1):]
```
SomberNight, line 94: I would prefer parameterising `getCommand()` to optionally not do the rstrip instead of duplicating it here.
I suggest `def getCommand(self, *, strip=True)`

verretor: I was thinking of doing something like that at first but I also wonder if rstrip() is really useful. I will do it like you ask.

electrum/gui/qt/console.py
```diff
@@ -15,6 +15,10 @@
 
 from .util import MONOSPACE_FONT
 
+# sys.ps1 and sys.ps2 are only declared if an interpreter is in interactive mode.
+sys.ps1 = '>>> '
+sys.ps2 = '... '
```
SomberNight, line 20: In practice these values would never be defined when using the Qt python console, as presumably Qt would not get run from an interactive interpreter. What are you trying to do by setting these -- is the main purpose just factoring out magic values into named variables/constants?

electrum/gui/qt/console.py
```diff
@@ -86,7 +89,7 @@ def updateNamespace(self, namespace):
         self.namespace.update(namespace)
 
     def showMessage(self, message):
-        curr_line = self.getCommand()
+        curr_line = self.getCommand(strip = False)
```
SomberNight, line 92: nit: the python convention is to leave out whitespaces here: `getCommand(strip=False)`

electrum/gui/qt/console.py
```diff
@@ -102,21 +105,22 @@ def keyboard_interrupt(self):
...
         self.appendPlainText(prompt)
         self.moveCursor(QtGui.QTextCursor.End)
 
-    def getCommand(self):
+    def getCommand(self, *, strip=True):
         doc = self.document()
         curr_line = doc.findBlockByLineNumber(doc.lineCount() - 1).text()
-        curr_line = curr_line.rstrip()
-        curr_line = curr_line[len(self.prompt):]
+        if strip == True:
```
SomberNight, line 121: why not simply `if strip:`? :)

electrum/gui/qt/console.py
verretor: I wasn't aware of that.

electrum/gui/qt/console.py
verretor: Shame on me haha.

electrum/gui/qt/console.py
verretor: I don't really like the way the prompts were before. The primary prompt '>>> ' was an argument of def __init__() and then was assigned to a property of Console. The other one '... ' wasn't defined anywhere.

Although not very useful, you could now change ps1 and ps2 if you like. I just find it cleaner to use the same variables as in the Python interpreter.

electrum/gui/qt/console.py
verretor: There's a few other things that I'd like to change because they're not conventional such as methods named:
`updateNamespace`, `showMessage`, `newPrompt`, etc which I believe should be called:
`update_namespace`, `show_message`, `new_prompt`.

I'm a bit shy to do these changes but I realize console.py comes mostly from Stack Overflow that's why I've been focusing on it so much.

electrum/gui/qt/console.py
verretor: `sys.ps1` and `sys.ps2` are defined when using the code module.

Example file `interact.py`:
```
#!/usr/bin/env python3
import code

code.interact()
```

```
$ ./interact.py 
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
(InteractiveConsole)
>>> import sys
>>> sys.ps1
'>>> '
>>> sys.ps2
'... '
>>> sys.ps1 = '+++ '
+++ sys.ps2 = '!!! '
+++ def f(x):
!!! 
```

electrum/gui/qt/console.py
SomberNight: You are correct, the method names are not following convention either.

Issue #6656, opened 2020-10-17
New lightning channel not saved to wallet (funds lost!)


mwnx:
Hi,

I tried opening a lightning channel (using the GUI) and got a segfault very soon after the channel was created and the funding transaction was broadcast. The channel appeared in the GUI's channel list before the crash. This bug report isn't about the segfault however (which I wasn't able to look into since my system did not produce a core dump) but about the fact that my wallet was not saved to disk before the crash, hence why my funds (thankfully not a large amount) are lost.

The good news is I think I've identified the reason why the wallet was not saved to disk. In `lnworker.py`:
```
    def add_new_channel(self, chan):
        self.add_channel(chan)
        channels_db = self.db.get_dict('channels')
        channels_db[chan.channel_id.hex()] = chan.storage
        for addr in chan.get_wallet_addresses_channel_might_want_reserved():
            self.wallet.set_reserved_state_of_address(addr, reserved=True)
        self.wallet.save_backup() # !!!
```
I'm pretty sure `save_backup` should be `save_db`, like everywhere else in `lnworker.py`! In my case, `save_backup` did nothing since by default the config file does not specify a backup location and I did not set one manually.

By the way, I also wonder if `save_db` shouldn't also be called before calling `try_broadcasting` in `_open_channel_coroutine` to account for the possibility of a crash after the transaction has been broadcast. I'm guessing this would not lead to a loss of funds however and so would not be as much of an issue.

Also, just to satisfy my curiousity, in the same function, this condition has me a bit perplexed:
```
        if funding_tx.is_complete():
            await self.network.try_broadcasting(funding_tx, 'open_channel')
```
In what legitimate case would the funding transaction be incomplete, and how could the channel be properly opened if the transaction were to not be broadcast?

Thanks for your great work on Electrum by the way.

SomberNight:
ack, I could reproduce :/
Thank you for the report!

> The good news is I think I've identified the reason why the wallet was not saved to disk. In lnworker.py:

Yes you are correct, the wallet db should be flushed to disk at that point.
I have changed this now in https://github.com/spesmilo/electrum/commit/c3fb79d412e0dc86b65f6ac8f117d28e0be08b4a

`chan.set_state` already results in a save to disk near the end of `channel_establishment_flow`, however a bit confusingly it is called just before actually adding the channel to the db, so it will not save the channel itself...
[...]

PR #6663, opened 2020-10-18
Fix lightning save_db issues (issue #6656)

mwnx:


ecdsa:
why is this PR still open? isn't the issue fixed?

SomberNight:
It is fixed. Just forgot to close.

PR #6736, opened 2020-11-16
appimage build: build most of our python dependencies from source

SomberNight:
In the AppImage build, we now compile most of our non-pure python dependencies from source, instead of using pre-built binary wheels from PyPI.


SomberNight:
force-pushed to remove a workaround for https://github.com/plizonczyk/noiseprotocol/issues/42

SomberNight:
new failure: https://github.com/vpelletier/python-libusb1/issues/64

Issue #6738, opened 2020-11-16
Multiple FX Currency Converters Need To Be Removed


scramblr:
Recently, 3 out of the 4 currency converters contained within Electrum have stopped allowing Tor & VPN connections, and therefore I'm proposing they should be removed as choices. 

Here are the logs showing unsuccessful attempts to change configurations. 

```
CoinGecko: NO LONGER ALLOWS TOR CONNECTIONS:

20201116T233657.155142Z |     INFO | exchange_rate.CoinGecko | failed fx history: ClientResponseError(RequestInfo(url=URL('https://api.coingecko.com/api/v3/coins/bitcoin/market_chart?vs_currency=NZD&days=max'), method='GET', headers=<CIMultiDictProxy('Host': 'api.coingecko.com', 'User-Agent': 'Electrum', 'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate')>, real_url=URL('https://api.coingecko.com/api/v3/coins/bitcoin/market_chart?vs_currency=NZD&days=max')), (), status=403, message='Forbidden', headers=<CIMultiDictProxy('Date': 'Mon, 16 Nov 2020 23:36:56 GMT', 'Content-Type': 'text/html; charset=UTF-8', 'Transfer-Encoding': 'chunked', 'Connection': 'close', 'CF-Chl-Bypass': '1', 'Set-Cookie': '__cfduid=dd51654833fe1e827e79ea235e6b144c61605569816; expires=Wed, 16-Dec-20 23:36:56 GMT; path=/; domain=.coingecko.com; HttpOnly; SameSite=Lax', 'Cache-Control': 'private, max-age=0, no-store, no-cache, must-revalidate, post-check=0, pre-check=0', 'Expires': 'Thu, 01 Jan 1970 00:00:01 GMT', 'X-Frame-Options': 'SAMEORIGIN', 'cf-request-id': '06750441830000178a6bbc2000000001', 'Expect-CT': 'max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"', 'Vary': 'Accept-Encoding', 'Server': 'cloudflare', 'CF-RAY': '5f35097c0f8e178a-FRA', 'Content-Encoding': 'gzip')>)


```
```
BitcoinVenezuela: NO LONGER ALLOWS TOR CONNECTIONS:
[...]

PR #6739, opened 2020-11-17
Removed CoinGecko, BitcoinVenezuela, CoinCap as they no longer accept…

scramblr:
… Tor connections.

These Cryptocurrency price services no longer accept connections from Tor IP addresses in a stable manner and should be removed from the code until the issue is fixed or alternatives are found.

### DIFF
--- a/electrum/currencies.json
+++ b/electrum/currencies.json
@@ -341,15 +341,6 @@
         "ZMW",
         "ZWL"
     ],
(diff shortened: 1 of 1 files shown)


SomberNight:
Closing this. See justification in https://github.com/spesmilo/electrum/issues/6738#issuecomment-728502280

PR #6740, opened 2020-11-17
lnrater: follow-up for save channel db attribute accesses

bitromortac:
Add comment on node score and return copies of channel db dicts as per comment in #6705 @SomberNight.


SomberNight:
Thanks :)

Issue #6697, opened 2020-10-28
Enable Electrum to handle big wallets and bad connections with incremental sync


ln2max:
As per #5489 Electrum has trouble with syncing large wallets, or syncing over bad connections. Extending the timeout would help, but ultimately the only way to address this in a scaleable way is by having wallet address synchronization done incrementally (with resume after a timeout/disconnect):

If the connection drops after syncing addresses 1 to 50000 in a wallet with 100000 addresses, then when reconnecting, Electrum should start synchronizing at address 50001 and go until address 100000.

Electrum should not start at the beginning of the wallet until the synchronization of the remaining addresses is complete. (Unfortunately, this is what it currently does, and it effectively limits how large a wallet can get before you have to discard the wallet and start over)

SomberNight:
Note that it's a lot more nuanced than that.

During sychronisation what actually happens is the client subscribing to its addresses and at the same time getting back a status hash for each address, using which it can tell whether it has up-to-date state for that address. Subscribing to an address means that while the connection (session) is alive, the server will send notifications if the status of the address changes.

So if the client does not resubscribe to some addresses after a disconnect then the server will not send a notification for those addresses. If and address status changes, e.g. there is a new tx, but the client does not get a notification, so does not get to know about it, that would clearly result in bad user experience.
 
If we made some assumptions re user behaviour, such as that already used empty addresses will not be re-used, then the client would not have to subscribe to those, and we could speed synchronisation that way.

ln2max:
On Wed, Oct 28, 2020 at 05:09:31PM -0700, ghost43 wrote:
>So if the client does not resubscribe to some addresses after a disconnect then the server will not send a notification for those addresses. If and address status changes, e.g. there is a new tx, but the client does not get a notification, so does not get to know about it, that would clearly result in bad user experience.
>
>If we made some assumptions re user behaviour, such as that already used empty addresses will not be re-used, then the client would not have to subscribe to those, and we could speed synchronisation that way.

That makes sense, but there are 2 ways to address this, one of which is identical to my original suggestion:

1. We may (as you suggest) assume a uniform distribution of new transactions across the wallet: we assume an old address is equally likely to get a new transaction as a new address.

In this case, my original suggestion is the correct answer. This is a product of the practical realities of overloaded servers and bad connections...

[...]

Issue #6625, opened 2020-10-01
[US$ 1500 bounty] Fix Electrum Crash & Lag resulting from TX history


kennedyiowa:
With big wallets containing a lot of transactions, Electrum will become unbearably slow, lag, freeze and eventually crash. https://github.com/spesmilo/electrum/issues/2540 is possibly related. My wallet size is 50MB by now, but it started way earlier, probably 10MB.

Version 3.3.8 is the last Electrum version where I can still open my wallet. Starting with Electrum 4 (I tried 4.0.3), Electrum won't even open anymore, the process is running but nothing happens. The experience below is from 3.3.8:

Doing some work with the API, def update_wallet(self): is running very very slow as well.

Electrum runs fine for a minute, freezes completely for 10-20 seconds, then works fine again. After running Electrum for about 1 day in the background, eventually it will synchronize endlessly (it finishes, then starts to synchronize again immediately), resulting in a crash after doing that for a certain amount of time.

If I had to take a guess, the lags and freezes probably result from Electrum loading and showing all transactions at once. It is possible that having these loaded and drawn as needed could fix the issue.

Electrum-LTC has the same issue.

The only active command I have:
wallet.change_gap_limit(1000)

My PC is fast and uses an NVME M.2 SSD.

Since this kind of load would only burden public servers, I'm using my own Electrumx server and node.

While these are bugs and if sufficiently documented, you could probably fix it; the bounty is primarily here because its an edge case and so I don't have to bother with handing you over a wallet that reproduces this. Maybe you can generate a big wallet via Testnet or import a lot of addresses.

The bounty will be valid for 2 months from now. If you think you can tackle this issue and the 2 months are over or need more time, just reply and I can renew it, depending on the situation.

If you are unsure whether this bounty is legitimate or not, just reply below posting status updates to this issue and additionally shortly before its ready to be PR'd. I can make the transaction before your work is delivered. US$ 1500 in XMR or BTC.

SomberNight:
How many addresses and transactions are there in your wallet?
In the `Console` tab, run these (the results are just examples):
```
>>> len(wallet.db.transactions)
104
>>> len(wallet.get_addresses())
145
```

kennedyiowa:
Electrum BTC:

```
>> len(wallet.db.transactions)
10000
>> len(wallet.get_addresses())
85000
```

Electrum LTC:

```
>> len(wallet.db.transactions)
2000
>> len(wallet.get_addresses())
45000
```

[...]

Issue #5489, opened 2019-07-05
Wallet sync issues


shikita:
Thank you.

Using latest electrum and X.

Wallet having around 20k addresses  and its unable to sync.  While synching up to around 5k then it again syncs from starting number.  it resets again  to **"Synchronizing... (343/20100)"**.  What might be the issue.

SomberNight:
Did it work before?

Check the client [logs](https://github.com/spesmilo/electrum-docs/blob/master/faq.rst#how-to-enable-debug-logging).

> Using latest electrum and X.

Do you run the server yourself? ElectrumX?
By default, there are [resource limits for sessions](https://electrumx.readthedocs.io/en/latest/environment.html#envvar-COST_SOFT_LIMIT).
You can disable them:
```
COST_SOFT_LIMIT=0
COST_HARD_LIMIT=0
```

shikita:
I enabled the log and retried it.

`201907 |     INFO | interface.[shikita_rasp:50001] | disconnecting due to RequestTimedOut("request timed out: ('blockchain.scripthash.subscribe', ['redacted_script_hash']) (id: 6189)")`

shikita:
i tried ` REQUEST_TIMEOUT=9999`  but no use.

SomberNight:
Is your server public or private? (so e.g. can it be found via server peer discovery?)
Are you connecting to it on a LAN, or even localhost? Or e.g. over internet, via Tor / VPN / etc?
Did you try disabling the resource limits like I said?

Can you sync a smaller wallet successfully?

It would be useful if you were less terse.

shikita:
The server is private. Its on my localhost on raspberrypi.
Yes smaller wallets sync was successfull.
Config file of electrumx.

```
export COIN=BitcoinSegwit
export PEER_DISCOVERY=off
export PEER_ANNOUNCE=''
export DB_ENGINE=leveldb
export DB_DIRECTORY=/media/shikita/sonyssd/xdb/
export ALLOW_ROOT=allow

export DAEMON_URL=http://shikita:shikita123@localhost:8332

export SERVICES=tcp://0.0.0.0:50001,rpc://localhost:8000
export COST_SOFT_LIMIT=0
export COST_HARD_LIMIT=0
export REQUEST_TIMEOUT=9999
```

SomberNight:
Try starting the client with `--oneserver`

shikita:
Tried `--oneserver`. Got the same error `RequestTimedOut`

[...]

PR #6741, opened 2020-11-17
Improve performance with large wallets and bad connections

ln2max:
Implement a number of fixes as per issues #6697 #6625 and #5489 to handle situations where electrum is currently choking on large wallets and/or failing to complete the initial request-missing-tx run.

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -80,7 +80,8 @@ class NetworkTimeout:
     class Generic:
         NORMAL = 30
         RELAXED = 45
-        MOST_RELAXED = 180
+        MOST_RELAXED = 600
+
     class Urgent(Generic):
         NORMAL = 10
         RELAXED = 20
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -731,7 +731,7 @@ class Network(Logger, NetworkRetryManager[ServerAddr]):
         util.trigger_callback('network_updated')
 
     def get_network_timeout_seconds(self, request_type=NetworkTimeout.Generic) -> int:
-        if self.oneserver and not self.auto_connect:
+        if self.oneserver:
             return request_type.MOST_RELAXED
         if self.proxy:
             return request_type.RELAXED
--- a/electrum/synchronizer.py
+++ b/electrum/synchronizer.py
@@ -232,7 +232,7 @@ class Synchronizer(SynchronizerBase):
     async def main(self):
         self.wallet.set_up_to_date(False)
         # request missing txns, if any
-        for addr in self.wallet.db.get_history():
+        for addr in random_shuffled_copy(self.wallet.db.get_history()):
             history = self.wallet.db.get_addr_history(addr)
             # Old electrum servers returned ['*'] when all history for the address
             # was pruned. This no longer happens but may remain in old wallets.


SomberNight:
The "sync addresses in random order" commit is good. Though I am not sure what is stated in the commit message is correct but the change does not hurt and in fact might help privacy; it can get merged in any case.

Re "always use MOST_RELAXED if oneserver is set" I think there might be a misunderstanding there.
`oneserver` does not imply `not auto_connect`.
Specifying `--server` on the command-line, or disabling `auto_connect` in the GUI, those imply `not auto_connect`.
While it is a bit silly, `not auto_connect` and `oneserver` can happen together, in which case I am not sure we want the extra long timeout. Probably not.

Re "increase MOST_RELAXED default timeout to 10 minutes", are there really scenarios where 3 minutes is not enough?

ln2max:
On Tue, Nov 17, 2020 at 12:12:15PM -0800, ghost43 wrote:
>The "sync addresses in random order" commit is good. Though I am not 
>sure what is stated in the commit message is correct but the change 
>does not hurt and in fact might help privacy; it can get merged in any 
>case.

I've experimented a lot and under certain circumstances the prior approach (non-random ordering that was always the same) resulted in wallets never being able to get new transactions (e.g when timeouts were too short).

>
>Re "always use MOST_RELAXED if oneserver is set" I think there might be a misunderstanding there.
>`oneserver` does not imply `not auto_connect`.
>Specifying `--server` on the command-line, or disabling `auto_connect` in the GUI, those imply `not auto_connect`.
>While it is a bit silly, `not auto_connect` and `oneserver` can happen together, in which case I am not sure we want the extra long timeout. Probably not.

I'm not able to find it at the moment, but the how-to that I originally used for setting up Electrum to always, only use a specific private server left my Electrum config with oneserver:true and auto_connect:true.

This results in the client *not* using the extra long timeout when it actually should, and I had to manually set auto_connect:false to use a more appropriate timeout. Since this was the result of published documentation, it seems safe to assume it is also the case for other users.

More generally, the user setting oneserver:true is an affirmative indication that the user wants to use a specific (perhaps less performant) server. auto_connect may be `false` even if there is a special server config (that warrants a longer timeout).

If both auto_connect and oneserver are `false` then the longer timeout is obviously warranted. I'm not sure what additional information `auto_connect` gives us that `oneserver` does not.

>Re "increase MOST_RELAXED default timeout to 10 minutes", are there really scenarios where 3 minutes is not enough?

My testing indicates the answer is "yes," but you're welcome to do your own experimentation.

SomberNight:
> I'm not able to find it at the moment, but the how-to that I originally used for setting up Electrum to always, only use a specific private server left my Electrum config with oneserver:true and auto_connect:true.

If you use a CLI flag to set the server, `auto_connect` will get disabled.
I would be curious to see which guide recommends setting `oneserver` but not setting which server to connect to.

If you end up in a situation where `oneserver is True` and `auto_connect is True`, Electrum will happily switch servers for you if the current one errors or times out. That's what `auto_connect` means. `oneserver` just means "only maintain a single server connection at any given time".

SomberNight:
As above, I would rather not merge the oneserver/auto_connect change.
I have cherry-picked the other two commits from this PR:
https://github.com/spesmilo/electrum/commit/42366ba57dea5f2b08c7c95882bcd0b59c7ae422
https://github.com/spesmilo/electrum/commit/7fdedd5c4071e687bb5b3fa2b17a851e45947a97

Thanks.

PR #6765, opened 2020-11-22
Fix get_running_loop (python<3.7)

wakiyamap:
Fix https://github.com/spesmilo/electrum/issues/6764

### DIFF
--- a/electrum/lnrater.py
+++ b/electrum/lnrater.py
@@ -11,7 +11,13 @@ from pprint import pformat
 from random import choices
 from statistics import mean, median, stdev
 from typing import TYPE_CHECKING, Dict, NamedTuple, Tuple, List
+import sys
 import time
+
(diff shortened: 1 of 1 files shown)


ecdsa:
can you explain what is going on here?

wakiyamap:
`get_running_loop` is from python3.7.
But.....The same thing exists in python3.6.
`_get_running_loop` is.

ref.
https://github.com/python/cpython/blob/3.7/Lib/asyncio/events.py
https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.get_running_loop
https://github.com/dask/distributed/pull/3383

SomberNight:
Alternatively we could just use [asyncio.get_event_loop](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop); but anyway, this is fine too.

PR #6771, opened 2020-11-24
Fix daemon being treated as a function

verretor:
`daemon` was improperly turned into a function in the Python console.
Point `daemon` to `window.gui_object.daemon` instead.

```
Before:
>>> window.gui_object.daemon
<electrum.daemon.Daemon object at 0xffff8d116970>
>>> daemon
'daemon' is a function. Type 'daemon()' to use it in the Python console.
>>> daemon()
  File "***/electrum/electrum/gui/qt/main_window.py", line 2141, in <lambda>
    return lambda *args, **kwargs: f(method,
  File "***/electrum/electrum/commands.py", line 161, in _run
    cmd = known_commands[method]
KeyError: 'daemon'

After:
>>> window.gui_object.daemon
<electrum.daemon.Daemon object at 0xffffb1eff730>
>>> daemon
<electrum.daemon.Daemon object at 0xffffb1eff730>
```

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -2143,7 +2143,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
                                              self.password_dialog,
                                              **{**kwargs, 'wallet': self.wallet})
         for m in dir(c):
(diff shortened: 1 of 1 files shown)


SomberNight:
thanks

PR #6779, opened 2020-11-25
kivy: unfocus passwd field input in on_dismiss

zebra-lucky:
After password is entered history screen is shown.
Back button is not work correctly if focus leaved on passwd input.
This commit seems fix the problem.

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/password_dialog.py
+++ b/electrum/gui/kivy/uix/dialogs/password_dialog.py
@@ -223,6 +223,8 @@ class AbstractPasswordDialog(Factory.Popup):
                 return True
         else:
             if self.on_success:
+                if 'textinput_generic_password' in self.ids:
+                    self.ids.textinput_generic_password.focus = False
                 args = (self.pw, self.new_password) if self.is_change else (self.pw,)
                 Clock.schedule_once(lambda dt: self.on_success(*args), 0.1)
 


ecdsa:
I cannot reproduce the issue. Can you clarify what you mean?

zebra-lucky:
> I cannot reproduce the issue. Can you clarify what you mean?

When starting wallet (linux/kivy or android build), after finished password dialog,
and when history screen is shown try to exit (Esc or back button). It's not work as early.
Need more button press. And `ElectrumWindow.on_keyboard` binding is not called first time.

ecdsa:
> Need more button press
such as?

zebra-lucky:
> > Need more button press
> 
> such as?

Seems 3 in kivy/linux, 4 (not sure) when back button in android pressed.

Users claims exit is misbehave (on Dash fork), and I've found this problem.

Maybe on different versions of kivy it works differently.

ecdsa:
so, apparently you are not the user who has this issue?
are you sure this is not the old "user needs to press home" bug?
it would be nice if you could reproduce the issue yourself, before you submit a pull request.

zebra-lucky:
> so, apparently you are not the user who has this issue?

I'm repeated the problem. For clarity I'll test on 4.0.5 arm64 build in a hour.

zebra-lucky:
Firstly on linux/kivy:

```pip freeze | grep Kivy
Kivy==1.11.1
Kivy-Garden==0.1.4```

other requirements from `contrib/deterministic-build/*`

after showing history and pressing `Esc` nothing happens

zebra-lucky:
Testing 4.0.5 arm64 build:

Ffter showing history and pressing back button nothing happens (on second press also).

zebra-lucky:
Fixed bug in prev commit (fail on Pincode setging).
The probmem still exists. Simplest way to reproduse is to change password from settings.
And then change pincode from settings and compare count of esc/back button presses is need to exit from settings.

ecdsa:
the kivy gui was removed in #8591

PR #6786, opened 2020-11-26
Fix travis build(MacOS)

wakiyamap:
Copy of the android description.


ecdsa:
travis failed to build MacOS on your pull request...

wakiyamap:
I guess I was stupid......
I think travis is still better for builds, because trying to do it with google action requires nearly twice the cost.

PR #6795, opened 2020-11-30
[wip] swaps: fix infinite recursion for max button

### LINKED ISSUE #6770: Setting Submarine Swap to 'Max' crashes Electrum
REPRO:
I am running Electrum 4.0.5 on Ubuntu 20.04.1
I have one Lightning channel open.
Lightning is synced
Under Channel I select Swap

ACTUAL:
- Upon clicking Max
- The fields 'you send' and 'you receive' fill up with my available balance
- Electrum crashes without an error notice

Additional information:
When first clicking on the Lightning symbol (you send on-chain, you receive lightning), then clicking Max, Electrum does not crash. When clicking on the blue Bitcoin symbol again (you send lightning, you receive on-chain), Electrum crashes without error notice.

bitromortac:
Fixes #6770 a crash happening when clicking the max button in the swap dialog.

### DIFF
--- a/electrum/gui/qt/swap_dialog.py
+++ b/electrum/gui/qt/swap_dialog.py
@@ -39,20 +39,25 @@ class SwapDialog(WindowModalDialog):
         vbox = QVBoxLayout(self)
         self.description_label = WWLabel(self.get_description())
         self.send_amount_e = BTCAmountEdit(self.window.get_decimal_point)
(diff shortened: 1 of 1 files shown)


ecdsa:
like I wrote in the issue, this is not sufficient to fix it, at least on my machine.
if I press Max it only fills one of the fields

ecdsa:
looks good, thanks

PR #6805, opened 2020-12-03
Kivy: use the same password for all wallets

ecdsa:
When a wallet password is updated, check the password for all
wallets in directory. If that check passes, update password for
all wallets. If it fails, update only the current wallet.

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -12,6 +12,8 @@ from typing import TYPE_CHECKING, Optional, Union, Callable, Sequence
 from electrum.storage import WalletStorage, StorageReadWriteError
 from electrum.wallet_db import WalletDB
 from electrum.wallet import Wallet, InternalAddressCorruption, Abstract_Wallet
+from electrum.wallet import check_password_for_directory, update_password_for_directory
+
 from electrum.plugin import run_hook
 from electrum import util
 from electrum.util import (profiler, InvalidPassword, send_exception_to_crash_reporter,
@@ -367,6 +369,7 @@ class ElectrumWindow(App, Logger):
         self.pause_time = 0
         self.asyncio_loop = asyncio.get_event_loop()
         self.password = None
+        self._use_single_password = False
 
         App.__init__(self)#, **kwargs)
         Logger.__init__(self)
@@ -634,6 +637,9 @@ class ElectrumWindow(App, Logger):
 
     def on_wizard_success(self, storage, db, password):
         self.password = password
+        if self.electrum_config.get('single_password'):
+            self._use_single_password = check_password_for_directory(self.electrum_config, password)
+        self.logger.info(f'use single password: {self._use_single_password}')
         wallet = Wallet(db, storage, config=self.electrum_config)
         wallet.start_network(self.daemon.network)
         self.daemon.add_wallet(wallet)
@@ -648,6 +654,12 @@ class ElectrumWindow(App, Logger):
         if not path:
             return
         if self.wallet and self.wallet.storage.path == path:
+            return
+        if self.password and self._use_single_password:
+            storage = WalletStorage(path)
+            # call check_password to decrypt
+            storage.check_password(self.password)
+            self.on_open_wallet(self.password, storage)
             return
         d = OpenWalletDialog(self, path, self.on_open_wallet)
         d.open()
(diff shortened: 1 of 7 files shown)


ecdsa:
note: once this is enabled, we should let users reset their PIN with their password

electrum/gui/kivy/main_window.py
```diff
@@ -649,6 +649,10 @@ def load_wallet_by_name(self, path):
             return
         if self.wallet and self.wallet.storage.path == path:
             return
+        if self.password and self.electrum_config.get('use_single_password'):
+            storage = WalletStorage(path)
+            storage.check_password(self.password)
```
SomberNight, line 654: what if wallet is only keystore-encrypted?

electrum/gui/kivy/main_window.py
```diff
@@ -1213,11 +1218,60 @@ def check_pin_code(self, pin):
...
+    def check_password_for_directory(self, old_password, new_password, update=False):
+        dirname = os.path.dirname(self.electrum_config.get_wallet_path())
+        self.logger.info(f'update pw {dirname}')
+        failed = []
+        success = []
+        for filename in os.listdir(dirname):
+            path = os.path.join(dirname, filename)
+            basename = os.path.basename(path)
+            storage = WalletStorage(path)
+            if not storage.is_encrypted():
```
SomberNight, line 1230: what is wallet is already keystore-encrypted (but not storage-encrypted)?

electrum/gui/kivy/main_window.py
```diff
@@ -1213,11 +1218,60 @@ def check_pin_code(self, pin):
...
+                # let's encrypt it
+                db = WalletDB(storage.read(), manual_upgrades=False)
+                wallet = Wallet(db, storage, config=self.electrum_config)
```
SomberNight, line 1233: maybe these lines too should be guarded with `if update`?

electrum/gui/kivy/main_window.py
```diff
@@ -1213,11 +1218,60 @@ def check_pin_code(self, pin):
...
+                success.append(basename)
+                if update:
+                    wallet.update_password(None, new_password)
```
SomberNight, line 1236: I guess `InvalidPassword` is not handled as you assume it cannot be raised due to `if not storage.is_encrypted()` - but again, wallet might be keystore-encrypted

electrum/gui/kivy/main_window.py
```diff
@@ -1213,11 +1218,60 @@ def check_pin_code(self, pin):
         if pin != self.electrum_config.get('pin_code'):
             raise InvalidPassword
 
+    def check_password_for_directory(self, old_password, new_password, update=False):
```
SomberNight, line 1221: Could you either add a docstring or change the API?
E.g. 
```
"""Checks password against all wallets and returns (list of matches, list of mismatches).
If update is True, it also changes the wallet passwords to new_password.
"""
```
Alternatively, the API might be clearer if "update" and "new_password" are merged together. Simplest would be to set `new_password` to `None` to correspond to `update==False`, however `None` is a valid value for `new_password` I guess, so we could introduce a sentinel e.g.
```
SENTINEL=object()
def check_password_for_directory(self, old_password, new_password=SENTINEL): ...
```

electrum/gui/kivy/main_window.py
```diff
@@ -1213,11 +1218,60 @@ def check_pin_code(self, pin):
...
         def on_success(old_password, new_password):
-            self.wallet.update_password(old_password, new_password)
+            # called if old_password works on self.wallet
             self.password = new_password
-            self.show_info(_("Your password was updated"))
+            success, failed = self.check_password_for_directory(old_password, new_password)
+            if not failed:
+                self.check_password_for_directory(old_password, new_password, update=True)
+                self.electrum_config.set_key('use_single_password', True)
+                msg = _(f"Password updated successfully for {len(success)} wallets")
```
SomberNight, line 1269: this does not work, see https://github.com/spesmilo/electrum/blob/78513affe536ba7d8150dc2ee044e3258097317f/electrum/i18n.py#L33-L35

electrum/gui/kivy/main_window.py
ecdsa: I agree, it is better to change the API. `None` is not a valid password for Kivy, so I guess it is fine to use it as long as this method stays in the GUI

electrum/gui/kivy/main_window.py
ecdsa: ack

ecdsa:
note: if a user does not know the unique password, they can still create a new wallet, and this will result in a situation where the unique password is disabled. This could be annoying in case of an unauthorized third party accessing the phone.

electrum/gui/kivy/main_window.py
ecdsa: This case is handled in `check_password_for_directory`. `self._use_single_password` is set only if the password was checked against all wallets and keystores. Here check_password is called for its side effect, to decrypt the storage. I will add a comment to clarify.

electrum/gui/kivy/main_window.py
ecdsa: this is now handled.

electrum/gui/kivy/main_window.py
ecdsa: same as above

ecdsa:
Rebased and squashed to have a single commit. Wallet creation is now disabled if the user does not know their password, so that password unification is irreversible.

An edge case remains: if the first wallet to be open is unencrypted and watching-only, no password is asked and `use_single_password` will be set to False

PR #6806, opened 2020-12-03
Wizard: do not ask seed type, default to segwit

ecdsa:
segwit addresses are widely supported today.


SomberNight:
Note that we will need this screen again with taproot, as witness v1 looks to be breaking compat for the address format.
Also note that while much less prevalent now, there are still one or two well known services not supporting sending to witness v0 - although maybe we can now ignore them at this point.

How about putting a checkbox/radios/dropdown into the "Options" dialog?
Capture

ecdsa:
>  How about putting a checkbox/radios/dropdown into the "Options" dialog?

We will see later if we need that screen again for taproot, and if it can be replaced with more choices in the options dialog. 
For the moment I think we should not bother.

PR #6824, opened 2020-12-07
auto-remove paid invoices from GUI

ecdsa:
- delay 3 seconds in GUI
 - rename invoice lists to "send queue", "receive queue"
 - kivy remove 'delete' buttons from send/receive screens

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -242,9 +242,12 @@ class ElectrumWindow(App, Logger):
         self._trigger_update_history()
 
     def on_request_status(self, event, wallet, key, status):
-        if key not in self.wallet.receive_requests:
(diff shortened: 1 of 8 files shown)


ecdsa:
note: `refresh_status` was needed to check for expiration of invoices

SomberNight:
why is the 3 sec delay needed?

ecdsa:
> why is the 3 sec delay needed?

the delay is there so that the transition to "paid" is visible for the user.
note that `update()` is currently called from from `update_tabs()`, this also needs to be fixed

ecdsa:
Another issue is that we currently decide that an onchain invoice has been "paid", based on blockchain history. This obviously needs to be changed. I guess a popup question could be shown to the user in case of address reuse.

Issue #6538, opened 2020-08-27
dnspython: compatibility with 2.0.0


prusnak:
Electrum requires `dnspython<2.0` in `requirements.txt`.

However, it would be nice it Electrum was able to use recently released dnspython 2.0.0 as well.

Here is the error when using dnspython 2.0.0:

```
=================================== FAILURES ===================================
_________________ TestDnsSec.test_python_validate_rrsig_ecdsa __________________

self = <electrum.tests.test_dnssec.TestDnsSec testMethod=test_python_validate_rrsig_ecdsa>

    def test_python_validate_rrsig_ecdsa(self):
        rrset = dns.rrset.from_text("getmonero.org.", 3599, 1, 48,
                                    "257 3 13 mdsswUyr3DPW132mOi8V9xESWE8jTo0d xCjjnopKl+GqJxpVXckHAeF+KkxLbxIL fDLUT0rAK9iUzy1L53eKGQ==",
                                    "256 3 13 koPbw9wmYZ7ggcjnQ6ayHyhHaDNMYELK TqT+qRGrZpWSccr/lBcrm10Z1PuQHB3A zhii+sb0PYFkH1ruxLhe5g==")
        rrsig = dns.rdtypes.ANY.RRSIG.RRSIG.from_text(1, 46,
            dns.tokenizer.Tokenizer("DNSKEY 13 2 3600 20180612115508 20180413115508 2371 getmonero.org. SSjtP2jCtXPukps7E3kum709xq2TH6Lt Ur32UhE7WKwSUfLTZ4EAoD5g22mi1fpB GDGb30kCMndDVjnHAEBDWw=="))
        keys = {dns.name.Name([b'getmonero', b'org', b'']): rrset}
        origin = None
        now = 1527185178.7842247

        # 'None' means it is valid
>       self.assertEqual(None, dnssec.python_validate_rrsig(rrset, rrsig, keys, origin, now))

electrum/tests/test_dnssec.py:21:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

[...]

PR #6828, opened 2020-12-08
dependencies: require dnspython 2.0, require cryptography 2.6, drop python-ecdsa

SomberNight:
resolves https://github.com/spesmilo/electrum/issues/6538


SomberNight:
note: we require dnspython 2.0 and not support older versions so that the dependency on python-ecdsa can be dropped

PR #6832, opened 2020-12-09
qt swap dialog: fix enabling OK button

### LINKED ISSUE #6831: problem with swaps on 4.0.6
i noticed that when creating a swap, the OK button remains always not working even if all parameters are correctly in place. I doing same exactly operation with version 4.0.5 all is going fine.

SomberNight:
fixes #6831

related: https://github.com/spesmilo/electrum/pull/6795

should we maybe rename `self.tx`?
It is evidently confusing for everyone that it is only applicable for the forward swaps.


ecdsa:
> should we maybe rename self.tx?

I think the usual terminology is that the first tx is a "lock tx" and then the other party does a "claim tx". 
not sure if using that terminology would make things clearer :-)

PR #6837, opened 2020-12-10
[wip] Swap slider

bitromortac:
This is an attempt to simplify the user interface for swaps.

Previously we had the amount fields and buttons to trigger the swap direction or spending/receiving maximal amounts. As the user doesn't know what the maximally allowed values for the swap are, it's a bit more complicated to enter the amounts manually and a trial and error process.

Here instead, all user interaction is accomplished with a slider element, which slided to the left represents a reverse swap and slided to the right represents a normal (submarine or forward) swap. The user is presented with an exact breakdown of onchain/offchain balance changes as well as the fees and type of fees paid. The slider range is initialized with maximal swap values. When opening the dialog we suggest a swap for a 50:50 balanced sending/receiving capacity on Lightning. A downside I can see with the slider is that the user is not able to enter exact amounts anymore, but I think that tradeoff may be acceptable.

We disable the swap button, when no channels are present. 

There are still some minor things left to do:
- [x] Enable swap button if channel is opened (Could separate disabling/enabling into own PR)
- [x] Figure out exact fees for normal swap case (right now they are off by some sats, need to go through the protocol details)
- [x] Add an explainer for swaps as a '?' button
- [x] Improve swap slider range and forbidden regions

### DIFF
--- a/electrum/gui/qt/swap_dialog.py
+++ b/electrum/gui/qt/swap_dialog.py
@@ -1,15 +1,16 @@
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, List
 
-from PyQt5.QtWidgets import QLabel, QVBoxLayout, QGridLayout, QPushButton
+from PyQt5.QtWidgets import QLabel, QVBoxLayout, QGridLayout, QSlider, QHBoxLayout, QSpacerItem
+from PyQt5.QtCore import Qt, QSize
+from PyQt5.QtGui import QFont, QPainter, QPen
 
+from electrum.gui.qt.util import HelpButton
 from electrum.i18n import _
 from electrum.util import NotEnoughFunds, NoDynamicFeeEstimates
 from electrum.lnutil import ln_dummy_address
 from electrum.transaction import PartialTxOutput, PartialTransaction
 
-from .util import (WindowModalDialog, Buttons, OkButton, CancelButton,
-                   EnterButton, ColorScheme, WWLabel, read_QIcon)
-from .amountedit import BTCAmountEdit
+from .util import (WindowModalDialog, Buttons, OkButton, CancelButton, read_QIcon)
 from .fee_slider import FeeSlider, FeeComboBox
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks! I think the zero should always be in the center of the slider, (so, one of the directions would max out before the end of the visible range)

ecdsa:
Also, when the dialog is open, I think the slider should be set to zero (no swap).
There is no real advantage to having 50:50 capability; what you want really depends on the context, and starting with an initial position of zero makes sure the user needs to activate the slider in order to enable the OK button (it prevents shooting yourself in the foot)

bitromortac:
Thanks for the review @ecdsa!
> Also, when the dialog is open, I think the slider should be set to zero (no swap).
> There is no real advantage to having 50:50 capability; what you want really depends on the context, and starting with an initial position of zero makes sure the user needs to activate the slider in order to enable the OK button (it prevents shooting yourself in the foot)

Yes, that was just an idea to suggest a swap, I agree that this could be a footgun. I'm going to remove it.

>I think the zero should always be in the center of the slider, (so, one of the directions would max out before the end of the visible range)

That's a point I'm still thinking about. It's confusing to the user which part of the interval belongs to which type of swap. Although, in the way it is presented right now we have the most fine granularity in the slider. At least we should indicate where the zero is located and which regions of the slider are forbidden (below min swap size). I'm thinking about adding a zero tick and painting the forbidden region in red.

ecdsa:
The slider is also huge in the current. I am not sure if that is really intended (maybe for granularity?). 
The issue with a huge slider is that it is difficult to recognize it as a slider, it looks like a separator.

There are also lots of redundant elements in the window: "swap action", "swap type", "increase in ...", "decresase in ..."
 
I think it would be better to keep the existing formulation and icons: "you send/you receive" (just without buttons)

bitromortac:
I updated the dialog a little bit and tried to incorporate your suggestions @ecdsa. The slider is reduced and displays now forbidden regions and I removed duplicate information. I didn't put the full [-max_swap, max_swap] swap range yet, as I think it makes the slider more complicated with two additional forbidden regions (but we can certainly try), however, I think it is intuitive enough like that. The previous behavior of showing the "you send" and "you receive" values is restored. The swap suggestion is removed and the swap is initialized at a zero value.
I also found a possible mistake in the calculation of the amounts in the forward swap case related to the service fee (see commit message). When debugging through the server response, Electrum now finds the exact same onchain amount as is expected in the server response. Previously a wrong calculation led to a slight overpayment in the send amount (only some ~tens of sats at max).

The help button displays still a draft for swap explanation. I'm also not sure whether we should display the warning messages after the swap.

bitromortac:
Closing in favor of the current approach.

I'm going to open a new PR for the amount calculation and disabling of the swap button when no channels are present.

For future reference, this is how a swap slider could look like:
swap-dialog

Issue #6664, opened 2020-10-18
Provide estimate of fees in fiat, and warn if they are high


micah:
When you send coins, the fee estimator is extremely useful. However, it presents things only in btc, and it would be really convenient if it would use the Fiat currency setting to show approximately how much the fee translates to.

The other day I created a transaction, it was pretty small, but it had a lot of inputs. The fee looked reasonable to me, but I misread how many zeros there were after the decimal point, and I nearly sent a transaction with a 220$ fee. Fortunately, I took the fee number and checked it, but if electrum had shown me the fiat conversion, it would have been a lot easier.

Also, it would be *really* nice if Electrum issued a warning when you were about to send a fee over a certain "reasonable" threshold, to make sure you want to do that. It does ask you to confirm your transaction, but if it says "WARNING: Your fee is abnormally high: 220$! Are you sure you want to send with this fee?"

SomberNight:
> the fee estimator [...] presents things only in btc, and it would be really convenient if it would use the Fiat currency setting to show approximately how much the fee translates to.
> if electrum had shown me the fiat conversion, it would have been a lot easier

Isn't the fee already shown in fiat too?
<details>
 <summary>pic1</summary>

![Capture](https://user-images.githubusercontent.com/29142493/96385081-11156400-1181-11eb-8df1-2dfa44e9683d.PNG)
</details>


> Also, it would be really nice if Electrum issued a warning when you were about to send a fee over a certain "reasonable" threshold, to make sure you want to do that. It does ask you to confirm your transaction, but if it says "WARNING: Your fee is abnormally high: 220$! Are you sure you want to send with this fee?"

There are already two warning conditions:
- if the fee exceeds 5% of the amount Electrum thinks you are sending
    <details>
     <summary>pic2</summary>
    
    ![Capture2](https://user-images.githubusercontent.com/29142493/96385084-14105480-1181-11eb-8f07-16745f4e73fb.PNG)
    </details>
- if the feerate is over 600 sat/byte
    <details>
     <summary>pic3</summary>
    
    ![Capture3](https://user-images.githubusercontent.com/29142493/96385136-76695500-1181-11eb-9f3b-656a29055140.PNG)
    </details>

What new warning condition would you like to see added specifically?

micah:
Hum, you are right!

However, I was working in the 'advanced' section, which does not show
the fiat conversion.

I no longer have any coins to generate screenshots, I'm sorry.

micah:
[...]

PR #6851, opened 2020-12-15
added fiat fee estimate for advanced mode

mnaei:


ecdsa:
related: #6664

ecdsa:
this does not work. 
you are passing a transaction size (in bytes) to a function that is expecting a transaction value (satoshis).
please fix.

mnaei:
I see, thank you very much for pointing this out. I will start working on a fix right away.

PR #6865, opened 2020-12-18
lnchannel: reflect frozen amounts and disconnected peers

### LINKED ISSUE #6855: Freezing of funds could be reflected in the can receive/can send info
Freezing of channel funds is not refleced in the amounts on the top of the channels list, as it uses `lnworker.num_sats_can_receive/send`. I think we could update these methods, as they are also used in swaps. Could give it a try if you are ok with it.

bitromortac:
in the num_sats_can_send/receive methods of the lnwallet. Fixes #6855.

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -530,6 +530,9 @@ class Channel(AbstractChannel):
     def is_initiator(self):
         return self.constraints.is_initiator
 
+    def is_active(self):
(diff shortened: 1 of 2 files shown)


ecdsa:
Note that this affects the semantics of `can_send_invoice` and `can_receive_invoice`, but I think that is fine.

PR #6871, opened 2020-12-19
Contacts import validation

### LINKED ISSUE #6841: AttributeError: 'list' object has no attribute 'items'
Crash Report
============

Traceback
--------------

```Python traceback
  File "electrum\gui\qt\contact_list.py", line 76, in <lambda>
  File "electrum\gui\qt\main_window.py", line 2880, in import_contacts
  File "electrum\gui\qt\util.py", line 934, in import_meta_gui
  File "electrum\contacts.py", line 56, in import_file
  File "electrum\contacts.py", line 125, in _validate

AttributeError: 'list' object has no attribute 'items'
```

Reporter
------------

| Electrum Version  | Python Version | Operating System  | Wallet Type  | Locale | Stack |
|---|---|---|---|---|---|
| 4.0.7  | 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:01:55) [MSC v.1900 32 bit (Intel)] | Windows-10-10.0.19041-SP0 | imported | en_ZA | [ℹ](https://crashhub.electrum.org//crash/16240) | 

Additional Information
------------------------

stickies-v:
The validation of imported contacts data has some assumptions about the data types and shape that may not always hold. I've added exception handling for those assumptions. 

fixes #6841

### DIFF
--- a/electrum/contacts.py
+++ b/electrum/contacts.py
@@ -131,8 +131,11 @@ class Contacts(dict, Logger):
             if not bitcoin.is_address(k):
                 data.pop(k)
             else:
-                _type, _ = v
-                if _type != 'address':
+                try:
+                    _type, _ = v
+                    if _type != 'address':
+                        data.pop(k)
+                except TypeError:
                     data.pop(k)
         return data
 
--- a/electrum/contacts.py
+++ b/electrum/contacts.py
@@ -25,6 +25,7 @@ import re
 import dns
 from dns.exception import DNSException
 
+from electrum.util import FileImportFailed
 from . import bitcoin
 from . import dnssec
 from .util import read_json_file, write_json_file, to_string
@@ -122,6 +123,8 @@ class Contacts(dict, Logger):
             return None
             
     def _validate(self, data):
+        if not isinstance(data, dict):
+            raise FileImportFailed(f'Invalid import data, JSON object expected')
         for k, v in list(data.items()):
             if k == 'contacts':
                 return self._validate(v)
--- a/electrum/contacts.py
+++ b/electrum/contacts.py
@@ -25,10 +25,9 @@ import re
 import dns
 from dns.exception import DNSException
 
-from electrum.util import FileImportFailed
 from . import bitcoin
 from . import dnssec
-from .util import read_json_file, write_json_file, to_string
+from .util import read_json_file, write_json_file, to_string, FileImportFailed
 from .logging import Logger
 
 


ecdsa:
if a single entry fails the validation, I think it might be better to fail the entire file, because the user might be trying to import something completely unrelated. but really, all the json files we import and export should have a type.

stickies-v:
I agree on the validation failing, that was my first intuition as well but I didn't want to just change behaviour if not really necessary. I suppose we probably want similar behaviour across different import functions (contacts, labels, invoices, requests)? If so, I'll have a think and come up with a proposal in a new PR.

> all the json files we import and export should have a type.
What do you mean with "have a type"? Give it an extension like .electrum_contacts, .electrum_invoices, ... etc? Alternatively, we could introduce a "type" record in the JSON object, such as:
```
{
    "type": "electrum/contacts/v1"
}
```
This would keep things nice and simple for GUI users, and give us a bit more control (and room for upgrades with the versioning parameter) over how files are parsed.

ecdsa:
Indeed, it would be good to have a generic function for contacts/labels/invoices/requests
I think the type should be "electrum/contacts" and the version number could be another field.

SomberNight:
> What do you mean with "have a type"? Give it an extension like .electrum_contacts, .electrum_invoices, ... etc? Alternatively, we could introduce a "type" record in the JSON object, such as:

An extension in itself is definitely not a sufficient solution. I agree that some kind of type information should be encoded as part of the file itself and checked.

An alternative to putting the type into the payload json would be to introduce a header (some magic bytes).

For example, a label export currently looks like this:
```
{
    "04d252dc4a571f44b8439d77536bf529ce229870c94fb9f2264e65ebc3949bbe": "Miner fee for Reverse Swap to BTC address",
    "06b84ada51e6866d5182ef6a609932931f740351309609e6a87301d626065af7": "Open channel"
}
```

We could instead change it to something like this:

```
electrum/labels/v1{
    "04d252dc4a571f44b8439d77536bf529ce229870c94fb9f2264e65ebc3949bbe": "Miner fee for Reverse Swap to BTC address",
    "06b84ada51e6866d5182ef6a609932931f740351309609e6a87301d626065af7": "Open channel"
}
```

i.e. we could add a header consisting of entirely some magic bytes that we use to identify the file type, which then can be discarded by our decoder internally.
This would result in a file that is not valid json as a whole; but I don't think it matters much. It is trivial to add/remove the magic bytes prefix.

stickies-v:
Sorry for never finalizing this, not sure I'll get around to it anytime soon so I'll close this for now.

PR #6872, opened 2020-12-19
Qt GUI: add dotted underline to HelpLabels

SomberNight:
I am fairly certain the majority of users never noticed many of the
labels in the UI are clickable.

Case in point, lots of information hidden behind clicking is out-of-date.
We should update them; ideally this could reduce support load.

pic

-----

In particular the expiration time of "bitcoin addresses" is a recurring question on the forums.
I am hoping that after making there being a help text more visible (and perhaps updating the text itself) fewer people would ask.

### DIFF
--- a/electrum/gui/qt/util.py
+++ b/electrum/gui/qt/util.py
@@ -101,11 +101,22 @@ class WWLabel(QLabel):
 
 class HelpLabel(QLabel):
 
+    STYLESHEET = """
+    HelpLabel {
+        border-bottom-width: 1px;
+        border-bottom-style: dotted;
+        border-radius: 0px;
+    }
+    HelpLabel::hover {
+        border-bottom-style: solid;
+    }
+    """
+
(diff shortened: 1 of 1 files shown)


ecdsa:
This adds clutter, it is visually not very pleasant. The current underline on hover was introduced as a replacement for tooltips, because tooltips need the mouse to stay for a long time.

ecdsa:
see also the question mark button in the network dialog

SomberNight:
I agree dotted underlines (this PR) add some clutter.
I wonder though how we could make the help info more visible in a less intrusive way -- e.g. the question mark approach as-is is not a nice fit for e.g. the `Send` tab :)

SomberNight:
I will close this as I agree it does not look too good; no point leaving it open.

but still think the premise stands:
> I am fairly certain the majority of users never noticed many of the labels in the UI are clickable.

PR #6875, opened 2020-12-20
lightning: upfront shutdown script

### LINKED ISSUE #6589: implement LN feature upfront shutdown script (commit to scriptPubKey for cooperative close)
We should implement the [`option_upfront_shutdown_script` feature](https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md) for lightning. (https://github.com/lightningnetwork/lightning-rfc/pull/266)

bitromortac:
Still a work in progress.

Upfront shutdown script is a Lightning feature that allows specifying a shutdown script on channel open where in the cooperative close funds are locked to that script (https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md). If say from our side the channel is closed to another address, the peer will disconnect and refuse to close cooperatively, which is a security feature in case we get compromised. Fixes #6589.

I wanted to gather feedback on the storage and wallet migration, if it's ok this way.
I tested already against lnd, regtests and unit tests pass, but I would still like to implement a unit test, probably in lnpeer?
Should we make upfront shutdown script optional or required? I opt for required, as this is around already a long time.

Todo:
- [x] Optional/required: do statistics for the graph
- [x] Don't activate upfront_shutdown from our side
- [x] Get rid of is_upfront_shutdown_script_enabled?
- [x] Unit test cooperative close to different shutdown script


ecdsa:
note: it would be nice to have statistics that tell us whether the features announced in gossip are now reliable. We had that issue last year with `option_static_remotepubkey`, and that is why we have an early return in `is_good_peer`

bitromortac:
I have checked the graph features on mainnet. About 50% of the nodes support `upfront_shutdown_script_opt` (25% on testnet) and none of the nodes require it, as it is implemented in LND right now. Concerning the comparison between signaled and actual support of the feature, I tried to connect to about a thousand nodes and the ones I could connect to all set the the feature flag as advertized. I can share the script, but it is a bit rough. So then let's set `upfront_shutdown_script_opt`.

Another open question is whether we really want to set an upfront shutdown script at all by default. The reason for the existence of this feature is that one is able to enforce cooperative closes for the peer to his desired address. One reason for us to implement this feature is to not lose potential peers in a future when it is required.
The feature only really has a use if we close out to a different wallet, but if we enforce payout to the same wallet, then we don't gain anything, right? It would make more sense to provide a close address to another wallet on channel opening (maybe an advanced setting in the channel open dialog?). If we enforce upfront shutdown script we also lose the flexibility to spontaneously set another address when closing the channel (such a field could be added to the channel close dialog), for instance closing to a coinjoin wallet.

By the way, `option_static_remotepubkey_opt` has support of 60% of the network (`option_static_remotepubkey_req` has 3%) and about 5% wrongly advertize themselves as supporting `option_static_remotepubkey_opt`.

ecdsa:
Not sure if we want to set an upfront shutdown script on our side. I believe @SomberNight wanted to have this feature so that we do not lose potential peers. If nobody currently requires it, it was indeed not very urgent. 

Regarding `option_static_remotepubkey_opt`, did you observe a discrepancy between the percentage of support in node announcement and the init messages?

bitromortac:
> Not sure if we want to set an upfront shutdown script on our side. I believe @SomberNight wanted to have this feature so that we do not lose potential peers. If nobody currently requires it, it was indeed not very urgent.

I removed setting the upfront script from our side and also got rid of the boolean in the channel storage for whether upfront_shutdown_script is used. This can be deduced from the upfront_shutdown_script fields in the local and remote configs.

> Regarding `option_static_remotepubkey_opt`, did you observe a discrepancy between the percentage of support in node announcement and the init messages?

Yes I saw minor discrepancy. I attached the script I used to exctract this info (very rough and will proably remove it once this PR is ready). For `option_static_remotepubkey` we have 4859 (60.21811872598835%) nodes in the graph with `opt` and 242 (2.999132482339819%) nodes with `req` of that feature. When really connecting to the nodes and checking their init message, out of the 1102 nodes I could connect to 61 nodes have falsely advertized to support `option_static_remotepubkey_opt`.

I'm still working on the unit test.

SomberNight:
> The feature only really has a use if we close out to a different wallet, but if we enforce payout to the same wallet, then we don't gain anything, right? 
> Not sure if we want to set an upfront shutdown script on our side. I believe @SomberNight wanted to have this feature so that we do not lose potential peers. If nobody currently requires it, it was indeed not very urgent.

It is not urgent for us; it is just a nice to have.
Indeed I think the main upside is not having to exclude peers that require it.

----

If I really wanted to, I could come up with contrived scenarios where it is ~useful for us directly... :D
E.g. consider wallet with master public key, p2wpkh addresses, lightning enabled;
if wallet file gets stolen (and password too :P) then
- if feature was enabled, thief would need to spend channel balances via LN
- without feature, thief can simply close channels and get paid on-chain to an external address of their choice

I know, far fetched. Also, even with feature enabled, they could e.g. do submarine swaps to extract funds. :)

----

On a more serious note, I think we can set it to `optional` (as opposed to `required`), however we definitely need a unit test to make sure both on/off states keep working.

electrum/lnpeer.py
```diff
@@ -1416,6 +1527,13 @@ async def close_channel(self, chan_id: bytes):
     @log_exceptions
     async def on_shutdown(self, chan: Channel, payload):
         their_scriptpubkey = payload['scriptpubkey']
+        their_upfront_scriptpubkey = chan.config[REMOTE].upfront_shutdown_script
+
+        # BOLT-02 check if they use the upfront shutdown script they advertized
+        if their_upfront_scriptpubkey:
+            if not (their_scriptpubkey == their_upfront_scriptpubkey):
+                raise Exception("remote didn't use upfront shutdown script it commited to in channel opening")
```
SomberNight, line 1535: As you have rightly pointed out on IRC, this exception will not be handled anywhere, it will just get logged and discarded. If you want to test for it, I suggest changing `process_message` to run the coroutine in the Peer's TaskGroup instead.

See patch:

```patch
index 5a7109eea..6550b9c30 100644
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -193,7 +193,7 @@ class Peer(Logger):
                 payload['raw'] = message
             execution_result = f(*args)
             if asyncio.iscoroutinefunction(f):
-                asyncio.ensure_future(execution_result)
+                asyncio.ensure_future(self.taskgroup.spawn(execution_result))
 
     def on_error(self, payload):
         self.logger.info(f"remote peer sent error [DO NOT TRUST THIS MESSAGE]: {payload['data'].decode('ascii')}")
```

This should ensure that the exception gets re-raised in `peer.main_loop()`, around line 305, and will result in us disconnecting from the peer.

In the tests, I think `peer.main_loop` is not actually used atm, so the taskgroup is not started; you would need to change this of course. To be able to test for the exception being raised, there are multiple ways you could go - for example, (1) instead of calling `peer.main_loop`, you could start the taskgroup manually in the unit test;
(2) or you could modify `peer.main_loop` by adding a try-except that handles the exception in the unit-test case and then re-raises it.
Re (2), I mean
```
try:
        ...
except Exception as e:
    self.main_loop_exiting_with_exc(e)
    raise
```
and `self.main_loop_exiting_with_exc` could be a no-op normally, but e.g. `Peer` could be subclassed in the unit tests and  `main_loop_exiting_with_exc` overridden.
There are probably even more ways to do it though.

-----

Regardless of implementation details, I recommend changing `process_message` to use the peer's taskgroup, as it's not nice not to await these coroutines anywhere. If you do that change, please consider other consequences, for example, the `@log_exceptions` decorator for `on_shutdown` becomes useless and annoying; and take a look at other affected message handlers whose exceptions will now not be ignored anymore; i.e. grep for `async def on_`.

ecdsa: nice patch.

bitromortac:
This should be ready for review. Thanks for the help @SomberNight! I ended up implementing the main_loop in the test, as we are not running all peer tasks in the unit test as compared to the ones in main_loop. I hope the scope in the unit test is broad enough. I removed the log_exceptions decorator from on_shutdown but left it at the other message handlers, I still need to think more about the consequences of the propagated message exceptions. We could also define more specific exceptions for protocol violations to catch them in unit tests.

electrum/lnchannel.py
```diff
@@ -649,6 +649,9 @@ def construct_channel_announcement_without_sigs(self) -> bytes:
     def is_static_remotekey_enabled(self) -> bool:
         return bool(self.storage.get('static_remotekey_enabled'))
 
+    def is_upfront_shutdown_script_enabled(self) -> bool:
+        return bool(self.storage.get('upfront_shutdown_script_enabled'))
+
```
ecdsa, line 654: this method is not used

ecdsa:
`test_close_upfront_shutdown_script` is failing for me:

```
I/P | lnpeer.Peer.[MockLNWallet, alice] | channel_reestablish (6945641x15186387x12973): sent channel_reestablish with (next_local_ctn=1, oldest_unrevoked_remote_ctn=0)
D/P | lnpeer.Peer.[MockLNWallet, bob] | Sending CHANNEL_REESTABLISH
I/P | lnpeer.Peer.[MockLNWallet, bob] | channel_reestablish (6945641x15186387x12973): sent channel_reestablish with (next_local_ctn=3, oldest_unrevoked_remote_ctn=2)
I/P | lnpeer.Peer.[MockLNWallet, alice] | channel_reestablish (6945641x15186387x12973): received channel_reestablish with (their_next_local_ctn=3, their_oldest_unrevoked_remote_ctn=2)
I/P | lnpeer.Peer.[MockLNWallet, alice] | channel_reestablish (6945641x15186387x12973): replayed 0 unacked messages
W/P | lnpeer.Peer.[MockLNWallet, alice] | channel_reestablish (6945641x15186387x12973): expected remote ctn 1, got 3
W/P | lnpeer.Peer.[MockLNWallet, alice] | channel_reestablish (6945641x15186387x12973): expected local ctn 0, got 2
W/P | lnpeer.Peer.[MockLNWallet, alice] | channel_reestablish (6945641x15186387x12973): remote is ahead of us! They should force-close. Remote PCP: 03cb799cc35f45d89b1f6e7e6c14b7bea87cbd4ad0512a3910d899e946efe1f2fb
Ignoring channel save
I/P | lnpeer.Peer.[MockLNWallet, bob] | channel_reestablish (6945641x15186387x12973): received channel_reestablish with (their_next_local_ctn=1, their_oldest_unrevoked_remote_ctn=0)
I/P | lnpeer.Peer.[MockLNWallet, bob] | channel_reestablish (6945641x15186387x12973): replayed 0 unacked messages
W/P | lnpeer.Peer.[MockLNWallet, bob] | channel_reestablish (6945641x15186387x12973): expected remote ctn 3, got 1
W/P | lnpeer.Peer.[MockLNWallet, bob] | channel_reestablish (6945641x15186387x12973): expected local ctn 2, got 0
W/P | lnpeer.Peer.[MockLNWallet, bob] | channel_reestablish (6945641x15186387x12973): we are ahead of remote! trying to force-close.

```

ecdsa:
the test is failing for me, it needs another instance of CancelledError

```diff
@@ -674,7 +674,7 @@ class TestPeer(ElectrumTestCase):
             coros = [close(), main_loop(p1), main_loop(p2)]
             gath = asyncio.gather(*coros)
             await gath
-        with self.assertRaises(asyncio.CancelledError):
+        with self.assertRaises(concurrent.futures.CancelledError):
             run(test())
 
     def test_channel_usage_after_closing(self):

```

ecdsa:
It looks good, besides the minor issues.
I merged it manually.

PR #6916, opened 2021-01-07
Kivy: add Lightning swaps

### LINKED ISSUE #6486: [Android-REQ]: Add "swap" button next to "new" on channel dialog for submarine swaps
In Android there is no option (that I see) to get to the Submarine-Swap feature.  Please consider adding a "Swap" button next to the "New" button in the channel dialog.  This would provide parity with the QT interface which has the "Swap" button on the channel tab.

bitromortac:
Fixes #6486. Adds a swap button to the channel maintenance dialog (renamed and reordered the buttons a bit). I ported the swap slider concept #6837 to kivy, which may be more useful in a mobile setup.

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -325,7 +325,7 @@ class SimpleConfig(Logger):
         slider_pos = max(slider_pos, 0)
         slider_pos = min(slider_pos, len(FEE_ETA_TARGETS))
         if slider_pos < len(FEE_ETA_TARGETS):
-            num_blocks = FEE_ETA_TARGETS[slider_pos]
+            num_blocks = FEE_ETA_TARGETS[int(slider_pos)]
             fee = self.eta_target_to_fee(num_blocks)
         else:
             fee = self.eta_target_to_fee(1)
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -84,7 +84,7 @@ from electrum.util import (NoDynamicFeeEstimates, NotEnoughFunds,
                            BITCOIN_BIP21_URI_SCHEME, LIGHTNING_URI_SCHEME)
 
 from .uix.dialogs.lightning_open_channel import LightningOpenChannelDialog
-from .uix.dialogs.lightning_channels import LightningChannelsDialog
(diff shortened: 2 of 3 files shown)


ecdsa:
thanks!

ecdsa:
I have not tried on mobile yet, but having the fee slider on the same screen might be too much

ecdsa:
Note that the "fee slider problem" predates this PR.
The only way we have to set the fee in kivy is through the send screen, which does not apply to "open channel" transactions

bitromortac:
Ok, that's interesting. I only tested on the desktop and tried to reduce the window resolution to somewhat match the appearance on the phone. The fee slider really takes up some space. On the other hand, for swaps it is quite important to select the right onchain fees. Should we maybe suggest a default fee based on ETA? To me it is unintuitive to set a global fee in the sending screen.

ecdsa:
Please remove the fee slider from the swap popup. I agree that it is un-intuitive to set on-chain fee in the sending screen. but that is a separate issue. The correct way to address such an issue is not to add more widgets to other dialogs. Adding more stuff is always at the expense of what is already being displayed. 

I think the on-chain fee dialog should be accessible in the the settings. It might also be good to access it from the popup that is displayed when paying on-chain (that popup displays the fee, so the label that displays the fee could be a button that opens the dialog). Similarly, we could access it from the swap dialog, with the same button.

bitromortac:
That's a good solution, working on it.

bitromortac:
Removed the fee slider and added a fee button. A short warning is displayed should the onchain fee be too low.

electrum/simple_config.py
```diff
@@ -325,7 +325,7 @@ def eta_to_fee(self, slider_pos) -> Optional[int]:
         slider_pos = max(slider_pos, 0)
         slider_pos = min(slider_pos, len(FEE_ETA_TARGETS))
         if slider_pos < len(FEE_ETA_TARGETS):
-            num_blocks = FEE_ETA_TARGETS[slider_pos]
+            num_blocks = FEE_ETA_TARGETS[int(slider_pos)]
```
ecdsa, line 328: how does that happen?

ecdsa:
merged manually

PR #6922, opened 2021-01-08
Export full derivation paths for Coldcard v3.2.1

peter-conalgo:
Latest version of Coldcard firmware (3.2.1) requires derivation paths to be known for all co-signers during multisig signing. This change adds that information, which was previously shown as comment in the same spot. With a multisig wallet open, go to _Wallet > Information > Export for Coldcard_ to exercise this change.

Also updates plugin README, removing historical junk.

[v3.2.1 firmware release notes](https://blog.coinkite.com/version-3.2.1-released/)

### DIFF
--- a/electrum/plugins/coldcard/README.md
+++ b/electrum/plugins/coldcard/README.md
@@ -1,4 +1,3 @@
-
 # Coldcard Hardware Wallet Plugin
 
 ## Just the glue please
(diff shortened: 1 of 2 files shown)


SomberNight:
Looks ok.
I've tested the "show address" functionality with new firmware and this PR, and that worked.

electrum/plugins/coldcard/README.md
```diff
@@ -14,22 +13,18 @@ this reason, all PIN code entry is done directly on the device.
 Coldcard does not appear on the USB bus until unlocked with appropriate
 PIN. Initial setup, and seed generation must be done offline.
 
-Coldcard uses an emerging standard for unsigned tranasctions:
+Coldcard uses the standard for unsigned tranasctions:
```
gruve-p, line 16: tranasctions > transactions

Issue #6960, opened 2021-01-21
Electrum lacks a clear way to handle dust attacks


CodeForcer:
Recently our hot wallet was dusted with an output from an advertisement transaction. The advertisement transaction remains in the mempool unconfirmed, assumably for a couple days before being dropped.

The hot wallet continues processing withdrawals, which are all unconfirmed on the blockchain, until finally crashing with the mempool too long error.

At this point we are in an "all hands on deck situation".

A developer loaded the hot-wallet private key into the Electrum GUI and attempted a cancellation for the transaction immediately after the dust input arrived (spending the dust input) with a very high transaction fee, almost $200 USD.

This cancelled the 20+ unconfirmed tx's, leaving just the RBF transaction. However, because the dust attack never confirms the RBF never confirms, despite the huge fee. Another RBF is attempted with a $400 USD transaction fee, but this likewise never confirms.

Unfortunately at this point it doesn't seem like Electrum has any effective way to fix the situation. Freezing inputs is only allowed for inputs which are not currently part of a transaction, so the coins tab doesn't not show the dust inputs or otherwise allow you to freeze them.

Also, modifying with RBF doesn't allow you to drop an input, nor does Electrum GUI seem to allow a custom transaction while selecting inputs from transactions you want to replace.

What can we do in these situations to free coins that have been dusted? The ability to freeze inputs only helps you if a dust input has arrived in your wallet, not in a situation where your wallet has already attempted to spend the dust input.

Lastly, is it possible to implement functionality in Electrum to easily stop this kind of attack? For example, automatically disabling the use of dust inputs in new transactions would prevent this from happening.

Thanks

SomberNight:
Was this dust sent to an address that has had other larger UTXOs at that time?
I don't see why the coinchooser would spend the dust, except if it wanted to spend another larger UTXO and due to the privacy heuristics pulled in all other UTXOs from that address.

-----

workaround.
- remove the unwanted unconfirmed txs from your wallet:
  (could also use `wallet.remove_transaction(txid)` but that assumes you never delete txs that have children in the wallet, whereas this handles that too)
```
window.history_list.remove_local_tx("fea1476847268cca3f9b6ecfdbf4422601a657c84660262e9696e805872d878f")
```
- freeze dust UTXO in Coins tab using GUI

-----

I am trying to come up with an automatic solution.

CodeForcer:
[...]

PR #6962, opened 2021-01-22
wallet: auto-freeze small unconfirmed UTXOs

SomberNight:
see #6960

this implements option (1) from https://github.com/spesmilo/electrum/issues/6960#issuecomment-765407023

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -287,7 +287,7 @@ class Abstract_Wallet(AddressSynchronizer, ABC):
         self.multiple_change       = db.get('multiple_change', False)
         self._labels                = db.get_dict('labels')
         self._frozen_addresses      = set(db.get('frozen_addresses', []))
-        self._frozen_coins          = set(db.get('frozen_coins', []))  # set of txid:vout strings
+        self._frozen_coins          = db.get_dict('frozen_coins')  # type: Dict[str, bool]
         self.fiat_value            = db.get_dict('fiat_value')
         self.receive_requests      = db.get_dict('payment_requests')  # type: Dict[str, Invoice]
         self.invoices              = db.get_dict('invoices')  # type: Dict[str, Invoice]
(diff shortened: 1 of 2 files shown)


electrum/wallet.py
```diff
@@ -1323,7 +1326,46 @@ def is_frozen_address(self, addr: str) -> bool:
...
+        """
+        # confirmed UTXOs are fine; check this first for performance:
+        block_height = utxo.block_height
+        assert block_height is not None
+        if block_height > 0:
+            return False
+        # exempt large value UTXOs
+        value_sats = utxo.value_sats()
+        assert value_sats is not None
+        threshold = self.config.get('auto_freeze_unconf_utxo_value_threshold_sats', 5_000)
```
ecdsa, line 1356: That is quite a long variable name. Also, the "u" in utxo already means unconfirmed.
How about 'utxo_freeze_threshold'?

electrum/wallet.py
```diff
@@ -1323,7 +1326,46 @@ def is_frozen_address(self, addr: str) -> bool:
...
-        return prevout_str in self._frozen_coins
+        frozen = self._frozen_coins.get(prevout_str, None)
+        # note: there are three possible states for 'frozen':
+        #       True/False if the user explicitly set it,
+        #       None otherwise
+        if frozen is None:
+            return self._is_coin_small_and_untrusted(utxo)
+        return bool(frozen)
+
+    def _is_coin_small_and_untrusted(self, utxo: PartialTxInput) -> bool:
```
ecdsa, line 1337: I guess you meant small_and_unconfirmed

ecdsa:
merged manually

electrum/wallet.py
SomberNight: sure that's fine

electrum/wallet.py
SomberNight: I simply was not sure how to call it; either is fine.

PR #6964, opened 2021-01-24
Changed color of coin selected in amount dialog in kivy gui.

aldrinpscastro:


SomberNight:
Could you upload pictures before/after, to compare?

aldrinpscastro:
Captura de tela_2021-01-24_03-10-23v2

ecdsa:
that looks worse to me, not better

aldrinpscastro:
i always get confused if i'm putting value on fiat or btc. I think that is why the previous color stands out a lot. I always think that it is she who is selected.

ecdsa:
the change in font size makes the display unstable.
you do not want a widget to change position when you tap on it.

PR #6966, opened 2021-01-24
Trampoline

ecdsa:
- trampoline is enabled by default in config, to prevent download of `gossip_db`.
   (if disabled, `gossip_db` will be downloaded, regardless of the existence of channels)
 - if trampoline is enabled:
    - the wallet can only open channels with trampoline nodes
    - already-existing channels with non-trampoline nodes are frozen for sending.
 - there are two types of trampoline payments: legacy and end-to-end (e2e).
 - we decide to perform legacy or e2e based on the invoice:
    - we use `trampoline_routing_opt` in features to detect Eclair and Phoenix invoices
    - we use `trampoline_routing_hints` to detect Electrum invoices
 - when trying a legacy payment, we add a second trampoline to the path to preserve privacy.
   (we fall back to a single trampoline if the payment fails for all trampolines)
 - the trampoline list is hardcoded, it will remain so until `trampoline_routing_opt` feature flag is in INIT.
 - there are currently only two nodes in the hardcoded list, it would be nice to have more.
 - similar to Phoenix, we find the fee/cltv by trial-and-error.
   if there is a second trampoline in the path, we use the same fee for both.
   the final spec should add fee info in error messages, so we will be able to fine-tune fees

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -58,6 +58,7 @@ from .lnutil import (Outpoint, LNPeerAddr,
                      UpdateAddHtlc, Direction, LnFeatures, ShortChannelID,
                      HtlcLog, derive_payment_secret_from_payment_preimage)
 from .lnutil import ln_dummy_address, ln_compare_features, IncompatibleLightningFeatures
+from .lnrouter import TrampolineEdge
 from .transaction import PartialTxOutput, PartialTransaction, PartialTxInput
 from .lnonion import OnionFailureCode, process_onion_packet, OnionPacket, OnionRoutingFailure
 from .lnmsg import decode_msg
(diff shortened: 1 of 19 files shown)


ecdsa:
Note: This was rebased and squashed into a single commit.

electrum/lnworker.py
```diff
@@ -1275,6 +1278,24 @@ def save_payment_info(self, info: PaymentInfo) -> None:
...
+    def add_received_htlc(self, htlc, expected_msat):
+        status = self.get_payment_status(htlc.payment_hash)
+        if status == PR_PAID:
+            return True, None
+        s = self.mpp_htlcs[htlc.payment_hash]
+        if htlc not in s:
+            s.add(htlc)
+        total = sum([htlc.amount_msat for htlc in s])
+        first_timestamp = min([htlc.timestamp for htlc in s])
+        expired = time.time() - first_timestamp > MPP_EXPIRY
```
bitromortac, line 1499: If I understand correctly, only a new incoming HTLC would trigger the expiry of the total payment, but we would hold on to the already accepted HTLCs indefinitely if the last part of the payment never arrives. A possible solution to this could be adding a coroutine that watches for all the incoming parts and cancels once a time limit is reached.

ecdsa: indeed, good point.

PR #6968, opened 2021-01-25
Added fiat fee estimate to Advanced Preview

HardCorePawn:
Related: #6664 
Fixes: #6851

### DIFF
--- a/electrum/gui/qt/transaction_dialog.py
+++ b/electrum/gui/qt/transaction_dialog.py
@@ -68,6 +68,9 @@ class TxSizeLabel(QLabel):
     def setAmount(self, byte_size):
         self.setText(('x   %s bytes   =' % byte_size) if byte_size else '')
 
+class TxFiatLabel(QLabel):
+    def setAmount(self, fiat_fee):
+        self.setText(('≈  %s' % fiat_fee) if fiat_fee else '')
 
 class QTextEditWithDefaultSize(QTextEdit):
     def sizeHint(self):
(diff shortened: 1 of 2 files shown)


electrum/gui/qt/transaction_dialog.py
```diff
@@ -848,6 +860,8 @@ def update_fee_fields(self):
         fee = tx.get_fee()
 
         self.size_e.setAmount(size)
+        fiat_size = self.main_window.format_fiat_and_units(fee)
+        self.size_f.setAmount(fiat_size)
```
SomberNight, line 864: please do not call this `size`... that word is used to mean the tx size in bytes everywhere else in this file.
it can be be `fiat_fee` or similar.

electrum/gui/qt/transaction_dialog.py
```diff
@@ -476,8 +480,10 @@ def update(self):
             amount_str = ''
         elif amount > 0:
             amount_str = _("Amount received:") + ' %s'% format_amount(amount) + ' ' + base_unit
+            amount_str += ' %s'%  format_fiat_and_units(amount)
         else:
             amount_str = _("Amount sent:") + ' %s'% format_amount(-amount) + ' ' + base_unit
+            amount_str += ' %s'%  format_fiat_and_units(-amount)
```
SomberNight, line 486: This field is visible not only in the "Create Transaction" dialog (advanced preview), but also in the "Transaction Details" dialog (when inspecting old txs in history). In such cases, this would use the incorrect FX conversion rate (should use historical rate instead of current rate).

SomberNight (changes requested):

HardCorePawn:
Modified the variable name as requested.

Also modified the code to use historical fiat values when viewing an existing transaction's details. Otherwise, "current rate" is used for "Create Transaction" dialog. 

Additionally, fiat values are only displayed if the user has actually selected a currency in the fiat settings.

SomberNight:
Is it intentional that the commits are getting attributed to a different github user?

HardCorePawn:
> Is it intentional that the commits are getting attributed to a different github user?

No. Something screwy with the way the commits were being created when using WSL??!? I have no idea why it was showing up as "jonhen"...

SomberNight:
ah I see the email in the commits is `foo@foo.com`, I guess that explains it :P

SomberNight:
Would you like to change the email address in the commits to something else? I can change it for you too.
but it's fine as-is if you don't care.

SomberNight:
I've rebased and squashed your commits, and added a follow-up commit.

HardCorePawn:
Thanks, I appreciate it.

PR #6977, opened 2021-01-28
fix htlc switch fulfilling htlc too soon (in case of race)

SomberNight:
This PR adds a unit test which showcases that on current master we might fulfill an HTLC before it is irrecovably committed;
and fixes that behaviour.

Note that this makes two of the four fields stored in `hm.log['unfulfilled_htlcs']` useless, so we could rm those. This would need a wallet upgrade. Shall we do that?


ecdsa:
Thanks. Can we try to keep variable and function names to a reasonable length?
I don*t see how `is_add_htlc_irrevocably_committed_yet` is better than `is_htlc_committed`
Adding unneeded text makes reading difficult, especially now that we have named parameters almost everywhere...

Issue #6971, opened 2021-01-26
Failed to build Electrum for windows


Toporin:
I tried to build Electrum (master commit @ d6e293052a858cb55427bac2fdcb4b742d91135e) for Windows using the deterministic build scripts and it failed silently with the following trace:

<details>
 <summary>log</summary>

```
💬 INFO:  Installing Python.
gpg: directory '/root/.gnupg' created
gpg: keybox '/root/.gnupg/keyring-electrum-build-python-dev.gpg' created
gpg: key FC624643487034E5: 5 signatures not checked due to missing keys
gpg: /root/.gnupg/trustdb.gpg: trustdb created
gpg: key FC624643487034E5: public key "Steve Dower (Python Release Signing) <steve.dower@microsoft.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg: no ultimately trusted keys found
Installing core...
--2021-01-26 14:52:03--  https://www.python.org/ftp/python/3.8.7/win32/core.msi
Resolving www.python.org (www.python.org)... 151.101.36.223, 2a04:4e42:9::223
Connecting to www.python.org (www.python.org)|151.101.36.223|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1425408 (1.4M) [application/octet-stream]
Saving to: ‘/opt/wine64/drive_c/electrum/contrib/build-wine/.cache/python3.8.7-win32/core.msi’

/opt/wine64/drive_c 100%[===================>]   1.36M  4.45MB/s    in 0.3s    

2021-01-26 14:52:04 (4.45 MB/s) - ‘/opt/wine64/drive_c/electrum/contrib/build-wine/.cache/python3.8.7-win32/core.msi’ saved [1425408/1425408]

--2021-01-26 14:52:04--  https://www.python.org/ftp/python/3.8.7/win32/core.msi.asc
Resolving www.python.org (www.python.org)... 151.101.36.223, 2a04:4e42:9::223
Connecting to www.python.org (www.python.org)|151.101.36.223|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 836 [application/octet-stream]
Saving to: ‘/opt/wine64/drive_c/electrum/contrib/build-wine/.cache/python3.8.7-win32/core.msi.asc’

/opt/wine64/drive_c 100%[===================>]     836  --.-KB/s    in 0s      

[...]

PR #6981, opened 2021-01-30
Fix windows build(travis)

wakiyamap:
Using Docker-ce(18.x) will cause windows build to fail.
I change to use the latest stable.

build log https://gist.github.com/wakiyamap/dcb638a6df1aeeffd3a61d94cb1a9488

ref. https://www.docker.com/blog/multi-arch-build-what-about-travis/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,13 @@ jobs:
         - TARGET_OS=Windows
       services:
         - docker
+      before_install:
+        - sudo rm -rf /var/lib/apt/lists/*
+        - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks. we should remember that this is probably temporary.

wakiyamap:
I think so, too.
If true, it is an unnecessary code.

Maybe it's related to this.
https://github.com/spesmilo/electrum/issues/6971

wakiyamap:
I tried a different version(19.03.8) of docker-ce, but it didn't work with the same error.

SomberNight:
I've [run Travis with this](https://travis-ci.org/github/SomberNight/electrum/jobs/757071217) and it passed.
Thanks!

PR #6804, opened 2020-12-02
WIP: payjoin as sender

rage-proof:
Hi, this PR is the try to implement the payjoin functionality as defined in BIP-78. It only covers the sender part in the interactive-tx and not the receiver stuff.
I tested it successful on Testnet with BTCPay Server using a hot wallet. 
I'm not sure if its a good approach of even intended, but I added the payjoin-link in some functions and classes to forward it to the transaction window. It's a draft and you can do with it whatever you want to.

ToDo's:
These points are the obvious(for me) that must be finished.
1. Added the 'requests' package, but probably it's not necessary and can be replaced by aiohttp package that is often present in electrum.
2. The validation that checks if the newly added input has the same script-type(p2pkh, p2wpkh) as our own input is still missing. I don't know if there exist already a proper function in electrum for determine the script-type from a witness-utxo in a psbt.
3. The check if the order of inputs and outputs was not changed. I didn't touch this, because I don't understand the BIP at this specific point.

Potential problem:
I saw this: https://github.com/btcpayserver/btcpayserver/issues/1631 & https://blog.trezor.io/details-of-firmware-updates-for-trezor-one-version-1-9-1-and-trezor-model-t-version-2-3-1-1eba8f60f2dd .
Therefore probably a payjoin between HW-wallets and a merchant using BTCPay-Server is not feasible, but I didn't test it yet.

PR #6984, opened 2021-01-31
add function to determine script-type for all inputs in a psbt

rage-proof:
Currently it's only possible to set the script-type for the inputs in a psbt for wallet owned inputs. For other imported inputs(payjoin) they stay 'unknown'. 
This PR adds a function to determine the inputs-types by checking if a script(in scriptpubkey or redeem-script) matches a scripts pattern.

I wanna use this function in this PR: https://github.com/spesmilo/electrum/pull/6804 to validate equality of input-types in the payjoin.

---
general question:
Why is the naming of the script-types different than in most other application/BIPs? If I'm right, in Electrum composed names are in the form: [inner/encapsulated-type]-outer (p2wpkh-p2sh), but in the BIP's the naming is p2sh-p2wpkh. Do I understand it correctly?

### DIFF
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -449,7 +449,6 @@ def match_script_against_template(script, template) -> bool:
             return False
     return True
 
-
 def get_script_type_from_output_script(_bytes: bytes) -> Optional[str]:
     if _bytes is None:
         return None
@@ -466,9 +465,6 @@ def get_script_type_from_output_script(_bytes: bytes) -> Optional[str]:
     if match_script_against_template(decoded, SCRIPTPUBKEY_TEMPLATE_P2WSH):
         return 'p2wsh'
     return None
-
-
-
 
 def get_address_from_output_script(_bytes: bytes, *, net=None) -> Optional[str]:
     try:
@@ -1144,6 +1140,7 @@ class PartialTxInput(TxInput, PSBTSection):
         self.witness_script = None  # type: Optional[bytes]
         self._unknown = {}  # type: Dict[bytes, bytes]
 
+        self.script_type = 'unknown'
         self.num_sig = 0  # type: int  # num req sigs for multisig
         self.pubkeys = []  # type: List[bytes]  # note: order matters
         self._trusted_value_sats = None  # type: Optional[int]
@@ -1352,19 +1349,18 @@ class PartialTxInput(TxInput, PSBTSection):
             return self.witness_utxo.scriptpubkey
         return None
 
-    @property
-    def script_type(self) -> Optional[str]:
+    def set_script_type(self) -> None:
         if self.scriptpubkey is None:
-            return 'unknown'
+            return
         type = get_script_type_from_output_script(self.scriptpubkey)
-        if type:
+        if type is not None:
             if type in ('p2sh', 'p2wsh'):
                 if self.redeem_script is not None:
                     inner_type = get_script_type_from_output_script(self.redeem_script)
-                    if inner_type:
+                    if inner_type is not None:
                         type = inner_type + '-' + type
-            return type
-        return 'unknown'
+            self.script_type = type
+        return
 
     def is_complete(self) -> bool:
         if self.script_sig is not None and self.witness is not None:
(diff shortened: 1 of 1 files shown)


SomberNight:
>  Why is the naming of the script-types different than in most other application/BIPs? If I'm right, in Electrum composed names are in the form: [inner/encapsulated-type]-outer (p2wpkh-p2sh), but in the BIP's the naming is p2sh-p2wpkh. Do I understand it correctly?

Specifically which BIPs use either naming convention?

rage-proof:
> Specifically which BIPs use either naming convention?

BIP-174(psbt) for example has Test Vectors and their description uses the later form.

SomberNight:
Well our nomenclature predates BIP174; I think it predates any BIP that has one :)
There isn't really a fundamental reason to prefer one over the other; we use what we use for historical reasons: these are the names we arbitrarily came up with years ago.

rage-proof:
Thanks for the explanation.

electrum/transaction.py
```diff
@@ -1331,6 +1349,19 @@ def scriptpubkey(self) -> Optional[bytes]:
             return self.witness_utxo.scriptpubkey
         return None
 
+    def set_script_type(self) -> None:
+        if self.scriptpubkey is None:
+            return
+        type = get_script_type_from_output_script(self.scriptpubkey)
+        if type is not None:
+            if type in ('p2sh', 'p2wsh'):
+                if self.redeem_script is not None:
```
SomberNight, line 1358: do you expect `type == 'p2wsh'`?
note that if `type == 'p2wsh'`, it is `self.witness_script` that should be checked
(if `type == 'p2sh'` then it is `self.redeem_script`)

electrum/transaction.py
```diff
@@ -447,6 +449,22 @@ def match_script_against_template(script, template) -> bool:
...
+    except MalformedBitcoinScript:
+        return None
+    if match_script_against_template(decoded, SCRIPTPUBKEY_TEMPLATE_P2PKH):
+        return 'p2pkh'
+    if match_script_against_template(decoded, SCRIPTPUBKEY_TEMPLATE_P2SH):
+        return 'p2sh'
+    if match_script_against_template(decoded, SCRIPTPUBKEY_TEMPLATE_P2WPKH):
+        return 'p2wpkh'
+    if match_script_against_template(decoded, SCRIPTPUBKEY_TEMPLATE_P2WSH):
+        return 'p2wsh'
```
SomberNight, line 466: Note that in many other places in `transaction.py`, a `script_type` of "p2sh" means p2sh-multisig, while a "p2wsh" means p2wsh-multisig.
I guess something will have to be done about that sooner or later anyway... Maybe it's ok to ignore that for now.
However, this is something to keep in mind if you are planning on comparing these script_type strings to determine if certain inputs use similar scripts.

electrum/transaction.py
rage-proof: changed

SomberNight:
> Note that in many other places in transaction.py, a script_type of "p2sh" means p2sh-multisig, while a "p2wsh" means p2wsh-multisig.
> I guess something will have to be done about that sooner or later anyway...

Ah, actually this might already become a problem here.

It seems to me that `is_complete` will consider non-ismine inputs that this PR marks as e.g. "p2sh" to be complete, and the code will try to finalise them. Even (non-ismine) inputs that are multisig can get incorrectly finalised because `num_sig` is not set for them correctly.
https://github.com/spesmilo/electrum/blob/d34b8d69f63e957e775cc8f2f24a431cf2926c7c/electrum/transaction.py#L1334-L1352
https://github.com/spesmilo/electrum/blob/d34b8d69f63e957e775cc8f2f24a431cf2926c7c/electrum/transaction.py#L1354-L1370
https://github.com/spesmilo/electrum/blob/d34b8d69f63e957e775cc8f2f24a431cf2926c7c/electrum/transaction.py#L1126

I think the easiest fix would be to limit the detection added in this PR to only
`p2pkh`, `p2wpkh-p2sh`, `p2wpkh`.

The longer term fix would be
- to split up `p2sh` et al to `p2sh-ms` and `p2sh-unknown` (or "p2sh-generic"? or just "p2sh"?) -- probabably also change them to an Enum or similar instead of string literals
- and then adapt the detection to specifically match for multisig templates, and also set `num_sig`

rage-proof:
Oh, interesting case. I agree this is a good solution for now.

> I think the easiest fix would be to limit the detection added in this PR to only
> p2pkh, p2wpkh-p2sh, p2wpkh.

This are the only types defined for Payjoin anyway.

SomberNight:
Thanks.

PR #7001, opened 2021-02-05
lnpeer: fix timed out mpp

bitromortac:
Fix to correcly instantiate OnionRoutingFailureMessage, otherwise this can lead to an endless reconnection when a multipart payment times out.


ecdsa:
thanks!

PR #7010, opened 2021-02-08
GUI: fix send Payment msg (BIP70)

zebra-lucky:
In qt/kivy there is no Payment msg (bip70) sent. Try to fix it (possibly not very clean way).

- qt: fix sending Payment msg on tx broadcast
- kivy: fix sending Payment msg on tx broadcast


zebra-lucky:
Found more correct way to fix the problem, will add changes in a couple of hours.

zebra-lucky:
> Found more correct way to fix the problem, will add changes in a couple of hours.
No, sorry, seem wrong, no changes.

ecdsa:
the kivy gui was removed in #8591

Issue #6783, opened 2020-11-26
Feature, Subtract fee from amount + RBF, Important for privacy. Offering 1.5m sat. ($350)


Transisto:
![image](https://user-images.githubusercontent.com/388716/100335911-84419f80-2fa3-11eb-8c99-ecd1802ee8a0.png)

Somewhere on this tab could be a "Subtract fees from amount" checkbox.

ecdsa:
@Transisto what do you mean?

Transisto:
![image](https://user-images.githubusercontent.com/388716/102684055-56ffa000-41a3-11eb-9a86-c52fa2b9d0b6.png)
BTCPAY

It would be extremely useful for accounting when a customer is asking to increase his transaction priority via RBF and has already paid with cash and left.

Having the option would also increase privacy chain wide by removing the general assumption that the extra fee for RBF is taken from the change.

Transisto:
Sorry I meant 1m sat, bumping it to 1.5m sat (~350USD at of now)

SomberNight:
To be clear, do you mean there could be an option for this in the "bump fee dialog"? So only when increasing the fee of an existing tx? -- as opposed to during initial construction of a tx

Transisto:
I mean it'd be great if fee could be taken from amount when creating the first transaction, like the two wallets above, but for us that can already be calculated easily, What's more useful for us is to make the customer pay after he left but call to request a speed up. 

But yeah, doesn't need to be part of the fee increase dialog too if the wallet just remember where to take the fee when it was first created.

SomberNight:
> BTW Bitcoin core take into account pay to many and will take the fee evenly from all outputs.

What does "evenly" mean?
Is the amount taken away linearly scaled according to the output values,
or are all outputs decreased by the same value?

Transisto:
Good point, it should be based on data weight of the utxo but I think
disregarding the address type is good enough, although I think we should
consider doing the same thing as Bitcoin Core to avoid wallet fingerprints.
`subtractfeefrom'

https://chainquery.com/bitcoin-cli/sendmany

SomberNight:
I think adding this "subtract fee from payment amount" option to the `Send` tab for normal tx creation is overkill. The `Send` tab is already versatile enough that you can achieve the same effect manually.

However, you can not achieve this currently with `bump_fee`. I think we could add such an option to the `bump_fee` dialog.
[...]

PR #7026, opened 2021-02-15
wallet: refactor bump_fee; add new strategy; change Qt dialog to have "advanced" button

SomberNight:
- wallet: refactor `bump_fee` strategies so that `bump_fee` can use any subset of them in any permutation
- wallet: add a new `bump_fee` strategy which decreases the payment outputs (instead of change)
  (this would close https://github.com/spesmilo/electrum/issues/6783)
- qt: move RBF dialog out of `main_window.py` into its own file
- qt: add "Advanced" button to RBF dialogs; expose "strategy" selection to bump_fee dialog
- qt: rename "Final" checkbox to "Keep Replace-By-Fee enabled"

-----

Re UI changes:

<details>
 <summary>Before PR</summary>

before
</details>

<details>
 <summary>After PR</summary>

after1
after2
</details>

### DIFF
--- a/electrum/coinchooser.py
+++ b/electrum/coinchooser.py
@@ -238,7 +238,7 @@ class CoinChooserBase(Logger):
             assert is_address(change_addrs[0])
 
         # This takes a count of change outputs and returns a tx fee
-        output_weight = 4 * Transaction.estimated_output_size(change_addrs[0])
+        output_weight = 4 * Transaction.estimated_output_size_for_address(change_addrs[0])
         fee_estimator_numchange = lambda count: fee_estimator_w(tx_weight + count * output_weight)
         change = self._change_outputs(tx, change_addrs, fee_estimator_numchange, dust_threshold)
         tx.add_outputs(change)
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -100,6 +100,7 @@ from .update_checker import UpdateCheck, UpdateCheckThread
 from .channels_list import ChannelsList
 from .confirm_tx_dialog import ConfirmTxDialog
 from .transaction_dialog import PreviewTxDialog
+from .rbf_dialog import BumpFeeDialog, DSCancelDialog
 
 if TYPE_CHECKING:
     from . import ElectrumGui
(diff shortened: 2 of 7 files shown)


ecdsa:
Thanks. Can we move the "show advanced settings" button to the bottom left (in the same line as ok, cancel), ant not make it disappear?  Disappearing widgets may trigger layout resizes, which are not nice.

Also, "Automatic" and "Coinchooser" is are not self-descriptive name for strategies. How is anyone except the developer supposed to know what they are doing?

SomberNight:
> Can we move the "show advanced settings" button to the bottom left (in the same line as ok, cancel), ant not make it disappear? Disappearing widgets may trigger layout resizes, which are not nice.

Sure, I've changed it as below:

<details>
 <summary>pics</summary>

pic1
pic2
pic3
</details>

> Also, "Automatic" and "Coinchooser" is are not self-descriptive name for strategies. How is anyone except the developer supposed to know what they are doing?

For now I've changed the text for `COINCHOOSER` to "Allow adding new inputs".
The "automatic"/default option is difficult to explain concisely.

It's quite hard to describe most of the strategies in user-friendly ways IMO.
- "automatic"/default is `COINCHOOSER` with fallback to `DECREASE_CHANGE` --> this is the existing behaviour on master / previous releases
- `COINCHOOSER` is "decrease change first, fallback to adding new inputs"
- `DECREASE_CHANGE` is "decrease any outputs but prioritise change"
- `DECREASE_PAYMENT` is "decrease all payment outputs"

Maybe the "Strategies" text next to the combobox could be made a `HelpLabel` with longer explanations.

ecdsa:
Regarding the dialog, it would be better to avoid having to refer to an extra help text. 
I think that it would it be better to display the description in the combo box, e.g.:

"try to decrease change, fallback to adding new inputs"
"try to decrease change, fallback to decreasing another output"
"decrease all outputs"

Also, when the dialog is displayed, the transaction to which it applies is already known, therefore it should be possible to predict exactly what each strategy is going to do, and we would not have to display "try A, fallback to B". We could display exactly what we are going to do.

SomberNight:
> Also, when the dialog is displayed, the transaction to which it applies is already known, therefore it should be possible to predict exactly what each strategy is going to do, and we would not have to display "try A, fallback to B". That would result in less choices and shorter descriptions.

Do you mean re the "automatic" choice which tries multiple strategies; or do you mean on the sub-strategy level, e.g. re `COINCHOOSER`, whether it would decrease change or add new inputs?
Maybe we could do the former, so on the strategy-level, but the latter sounds more complicated than it is worth.

Regardless, note that this would mean that moving the slider can change both the list of available strategies and what each strategy does. So the contents of the combobox could fluctuate.

> Regarding the "strategy", it would be better to avoid having to refer to an extra help text.
I think that it would it be better to display the description in the combo box, e.g.: [...]

Sure, ok, we can try to make the descriptions short and put those in the combo box.

What should the text be for the default option though? (which just tries two of the other strategies in order...)
so it's basically "try to decrease change, fallback to adding new inputs, fallback to decreasing any outputs"
This option should be kept because there should be an option corresponding to what happens by default -- and this is a very good default imho as it probably matches what many casual users would want. (also note that the default option is what happens if the advanced widget is hidden)

One thing we could do if we wanted fewer options is remove `[COINCHOOSER]`, i.e. leave three options:
- default: `[COINCHOOSER, DECREASE_CHANGE]`
- `[DECREASE_CHANGE]`
- `[DECREASE_PAYMENT]`

I guess there is no utility of a `[COINCHOOSER]` option if `[COINCHOOSER, DECREASE_CHANGE]` is there.

ecdsa:
> Do you mean re the "automatic" choice which tries multiple strategies; or do you mean on the sub-strategy level, e.g. re COINCHOOSER, whether it would decrease change or add new inputs?
> Maybe we could do the former, so on the strategy-level, but the latter sounds more complicated than it is worth.

I am saying that there is no need to force users to learn about "strategies". What you call a strategy is an "if then else" statement, or a combination of those statements. In your current proposal, you are asking users to learn that they have the choice between various "if then else" strategies. Secondly, you are asking them to learn a set of symbols that refer to those strategies. Thirdly, you are introducing two levels: sub-strategies, and a top-level automatic strategy, and you want users to learn that too. I am sorry, but I think this is wrong. You are not measuring how much cognitive load you are introducing.

An "if then else" statement makes only sense if it is executed by a machine. If the decision has to be taken by a human, it is easier to display a choice between the final outcomes that are available, so that humans do not have to run the "if then else" in their head (they might not even know how many inputs the transaction has, at this point), and so that they do not need to learn new symbols. Thus, the list should not contain logical statements such as "decrease change, or decrease any outputs", of "try A and fallback to B". It should only contain the final branches of the decision tree:

 - "decrease change"
 - "add new inputs"
 - "decrease a single output"
 - "decrease all outputs"

The first option will not be displayed at all if there is no change in the transaction, and the last 2 can be merged if there is only one output (that is the usual case). If users select one of the last 2, display a warning that the payment might be incomplete.

> Regardless, note that this would mean that moving the slider can change both the list of available strategies and what each strategy does. So the contents of the combobox could fluctuate.

Indeed, moving the slider would change the list of available options (again, those are final outcomes, not "strategies"). If a combo box does not work well with that, we can display those options as radio buttons, and disable them dynamically.

SomberNight:
>> Regardless, note that this would mean that moving the slider can change both the list of available strategies and what each strategy does. So the contents of the combobox could fluctuate.
>
> Indeed, moving the slider would change the list of available options (again, those are final outcomes, not "strategies"). If a combo box does not work well with that, we can display those options as radio buttons, and disable them dynamically.
> It should only contain the final branches of the decision tree

Lets say the user clicked "Advanced" to have the options shown but did not tweak the advanced options yet.
The user now starts moving the fee slider.
If the UI displays only final branches of strategies, at this point the selection will necessarily fluctuate.
Having a single option correspond to the default behaviour is not compatible with (displaying only final branches AND no fluctuation of GUI selection while slider is moving). This follows from the default option being a complex decision tree (and it is true regardless of what kind of GUI elements are used to represent the choice).

So I guess part of the problem is that I would like one of the options to correspond to "the user did not tweak this at all, i.e. use default behaviour", while what you propose makes that impossible because the default behaviour itself is a complex decision tree.

> I am saying that there is no need to force users to learn about "strategies". What you call a strategy is an "if then else" statement, or a combination of those statements. In your current proposal, you are asking users to learn that they have the choice between various "if then else" strategies. Secondly, you are asking them to learn a set of symbols that refer to those strategies.

If by set of symbols you mean the `BumpFeeStrategy` enum, that is not exposed to users, only developers need to learn it.

> Thirdly, you are introducing two levels: sub-strategies, and a top-level automatic strategy, and you want users to learn that too. 

Yes, the top-level automatic strategy is a combination of multiple "strategies" but users don't need to learn this either.
In my previous comment, I suggested we could have these options in the GUI:

- default: `[COINCHOOSER, DECREASE_CHANGE]`
    - label: "decrease change, or add new inputs, or decrease any outputs"
- `[DECREASE_CHANGE]`
    - label: "decrease change, or decrease any outputs"
- `[DECREASE_PAYMENT]`
    - label: "decrease all payment outputs"

Users can be completely oblivious to how these options are implemented and whether they involve multiple "strategies" (from the PoV of `wallet.bump_fee`).

> I think you might not be realizing how much cognitive load you are introducing.

Note that we are talking about an "advanced" option here which is not visible by default.


> It should only contain the final branches of the decision tree, such as:
> - "decrease change"
> - "add new inputs"
> [...]

From an implementation point of view, I imagine we would do this by splitting up the strategies (i.e. methods in wallet.py) to have a "try this or fail" approach instead of being complex decision trees. However, note that it would be weird to split some of the strategies up.
Consider the `COINCHOOSER` strategy, which I've described as "decrease change, or add new inputs". The description is just that, a description. There is actually no decision tree in the code at all: the code just takes all the non-ismine outputs and all the inputs, passes it to the coinchooser and lets it do whatever it wants. In practice this will result in "decrease change, or add new inputs" but there is no branching anywhere. What that strategy truly does is "recreate tx with higher fee" :D

Anyway, my main point is that I think most of the complexity comes from trying to have one of the options correspond to the default behaviour, and the default behaviour being a complex decision tree.

ecdsa:
> So I guess part of the problem is that I would like one of the options to correspond to "the user did not tweak this at all, i.e. use default behaviour", while what you propose makes that impossible because the default behaviour itself is a complex decision tree.

> Anyway, my main point is that I think most of the complexity comes from trying to have one of the options correspond to the default behaviour, and the default behaviour being a complex decision tree.

I think that you are indeed pointing at the problem. But what is the point of having one of the options representing the default behaviour? 

I think it would be more transparent to select by the outcome of the default policy (using radio buttons), and to update which item is selected when the slider is moved. The choices will be easy to understand, and the user will know instantly what is going to happen, instead of guessing.

ecdsa:
> Consider the COINCHOOSER strategy, which I've described as "decrease change, or add new inputs". The description is just that, a description. There is actually no decision tree in the code at all: the code just takes all the non-ismine outputs and all the inputs, passes it to the coinchooser and lets it do whatever it wants. In practice this will result in "decrease change, or add new inputs" but there is no branching anywhere. What that strategy truly does is "recreate tx with higher fee" :D

OK, but that is an implementation choice, and it is probably questionable. If the user decides to visit the advanced options, I can imagine that it is because they want more control. To tell them that the coinchooser is going to do what it want is not really interesting in that context. It is not empowering, and it is not transparent. It is fine to use that as a default policy, but I do not think that it should be in the list of choices.

SomberNight:
I've force-pushed this change re UI labels:
> - default: `[COINCHOOSER, DECREASE_CHANGE]`
>     - label: "decrease change, or add new inputs, or decrease any outputs"
> - `[DECREASE_CHANGE]`
>     - label: "decrease change, or decrease any outputs"
> - `[DECREASE_PAYMENT]`
>     - label: "decrease all payment outputs"

---

According to our out-of-band conversation, we agree that the PR as-is is an improvement over current master.
The UI is not great; definitely could use more work -- but the suggested changes as above are non-trivial to implement (and they need `wallet.py` changes too, not just UI).

So I am going to merge this now; it can always be improved upon later.

PR #7043, opened 2021-02-19
Added back the 'to_height' and 'from_height' to onchain_history command

mariobyn:


ecdsa:
thanks

ecdsa:
you are not handling the special case of unconfirmed transactions, for which `height` is zero.
is that a conscious choice? I do not remember if the previous code was handling that either..

mariobyn:
It didn't from my research but I can do that. question, is there an option to filter only the unconfirmed transactions? If not, should we add also here a parameter for that? do you think it's necessary?

ecdsa:
My point is, if you want transaction above a given height, then it feels intuitive to include unconfirmed transactions in the list. Conversely, if you want transactions below a given height, it feels intuitive to exclude unconfirmed transactions. The current code does the opposite.

mariobyn:
Updated!

ecdsa:
thanks. I forgot to mention that there are more cases than height==0, so you want to test for height <= 0
see https://github.com/spesmilo/electrum/blob/master/electrum/address_synchronizer.py#L46

mariobyn:
This raises another question:
should we include 
```
TX_HEIGHT_FUTURE 
TX_HEIGHT_LOCAL
TX_HEIGHT_UNCONF_PARENT
```
in from_height? not only unconfirmed transactions? or maybe just 
```
TX_HEIGHT_UNCONF_PARENT,
TX_HEIGHT_FUTURE
```
?

ecdsa:
I think we can include them all. introducing distinctions would complicate things even more.

mariobyn:
Updated

ecdsa:
thanks

ecdsa:
merged manually, after squashing.

Issue #6586, opened 2020-09-13
implement LN multi part payments (basic MPP)


SomberNight:
We should implement the [`basic_mpp` feature](https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md) for lightning. (https://github.com/lightningnetwork/lightning-rfc/pull/643)
Receive support is easier so probably that first, but then sending as well.

(also called `base AMP`)

ecdsa:
note: receive support is required for trampoline.
see https://github.com/spesmilo/electrum/commits/basic_mpp

ecdsa:
mpp receive is already implemented, and @bitromortac is currently working on mpp send.

ecdsa:
merged.

PR #7050, opened 2021-02-20
Complete multipart payment sending support

bitromortac:
Implements a strategy to split payments and adds some last ingredients to complete multipart payment support. Related to #6586.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -76,6 +76,7 @@ from .channel_db import UpdateStatus
 from .channel_db import get_mychannel_info, get_mychannel_policy
 from .submarine_swaps import SwapManager
 from .channel_db import ChannelInfo, Policy
+from .mpp_split import suggest_splits
 
 if TYPE_CHECKING:
     from .network import Network
@@ -1019,7 +1020,7 @@ class LNWallet(LNWorker):
         key = payment_hash.hex()
         payment_secret = lnaddr.payment_secret
         invoice_pubkey = lnaddr.pubkey.serialize()
-        invoice_features = lnaddr.get_tag('9') or 0
+        invoice_features = LnFeatures(lnaddr.get_tag('9') or 0)
         r_tags = lnaddr.get_routing_info('r')
         t_tags = lnaddr.get_routing_info('t')
         amount_to_pay = lnaddr.get_amount_msat()
@@ -1377,6 +1378,16 @@ class LNWallet(LNWorker):
                 node_features=trampoline_features))
         return route
 
+    def channels_with_funds(self) -> Dict[bytes, int]:
+        """Determines a dict of channels (keyed by channel id in bytes) that
+        maps to their spendable amounts."""
+        with self.lock:
+            channels = {}
+            for cid, chan in self._channels.items():
+                spend_amount = int(chan.available_to_spend(HTLCOwner.LOCAL))
+                channels[cid] = spend_amount
+            return channels
+
     @profiler
     def create_routes_for_payment(
             self,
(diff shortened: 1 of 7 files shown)


ecdsa:
The fee for each edge is added to the amount in `lnrouter` `get_distances`, starting from the payee back to the payer.
If the private channel we use cannot pay the amount with the fees, find_route will reject that route.
Therefore I do not think that `fee_reserve_factor` is useful.

Did you actually observe a case where it was useful?

electrum/lnworker.py
```diff
@@ -1387,11 +1399,100 @@ def create_routes_for_payment(
...
+                full_path=full_path
+            )
+            return [(route, amount_msat)]
+
+        channels_with_funds = self.channels_with_funds()
+
+        # check if payment is supported by a single channel
+        use_single_channels = False
+        if len([v for v in channels_with_funds.values() if v > amount_msat]) > 1:
+            use_single_channels = True
```
ecdsa, line 1426: At this point it is too early to know whether we can pay wit a single channel. We might have a large capacity channel with a node that will be unable to forward the payment because its other channels are small. We need to run `find_route` in order to detect that. 

I think it would be simpler if single-part payments were returned in the list returned by `suggest_splits`, before multi-part payments. That way `find_route` could detect if a single part payment works, and try multi-part payments if it fails.

Also, `mpp_supported` could be passed as a parameter to `suggest_splits`, so that we don't need the if statement that  duplicates the call to `create_route_for_payment`.

bitromortac:
Removed the `fee_reserve_factor`. Whether a single channel is used is determined by a try-except statement. We will fall back to payment splitting if the amount cannot be sent via a single channel. During comparison of the tests we noticed that the tests (of a pseudorandom algorithm) depend on the python version used, here pinned to 3.8.

ecdsa:
thanks. I will remove the method `channels_with_funds`, because it fits in 1 or 2 lines, and is only called once.

ecdsa:
oh, the tests are failing because of the part that depends on the python version.
is it possible make it robust to that?

ecdsa:
Regarding `fee_reserve_factor`:  There are cases where `create_route_for_payment` returns a route where the amount exceeds what we can pay. I think this is a pre-existing bug. I guess you might have encountered that, and added `fee_reserve_factor` for that reason.

Normally, you would expect line 235 of `lnrouter.py` to make sure that the channel can pay the amount plus all the forwarding fees. The problem is that this does not include the forwarding fees of the channels that are passed in the invoice routing hints.

SomberNight:
> the tests are failing because of the part that depends on the python version.

What exactly depends on the python version?

e.g. this works for me on py 3.6, 3.7, 3.8:
```
>>> import random
>>> random.seed(22)
>>> random.randint(1, 1000)
982
```

bitromortac:
@SomberNight: I couldn't really figure it out, but #7060 could give a hint.

electrum/tests/test_mpp_split.py
```diff
@@ -0,0 +1,75 @@
...
+PART_PENALTY = mpp_split.PART_PENALTY
+
+
+class TestMppSplit(ElectrumTestCase):
+    def setUp(self):
+        super().setUp()
+        random.seed(0)  # split should only weakly depend on the seed
+        # test is dependent on the python version used, here 3.8
+        # undo side effect
+        mpp_split.PART_PENALTY = PART_PENALTY
```
SomberNight, line 17: Would it not make more sense to do this in `tearDown`, after (instead of before) each test?
It probably does not matter right now but note that other tests (in other files) might now see a non-default `mpp_split.PART_PENALTY`; which would not be the case if this was moved to `tearDown`.

PR #7060, opened 2021-02-23
mpp_split: fix tests for python versions < 3.8

bitromortac:
Tests should now pass also for versions 3.6 and 3.7.

I removed the specific numbers from the test as we are interested in the general splitting behavior. I couldn't figure out what is the exact difference between 3.8 and <3.8, but one thing I observe is that the tests have equal outcomes to the 3.8 version once the seeding is moved to the global scope instead of seeding during the test set up :fearful:.


ecdsa:
test_mpp_split still randomly fails, approximately 20% of the time

PR #7062, opened 2021-02-23
mpp_split: try to fix non-determinicity of tests

SomberNight:
related #7050
related #7060
(this conflicts with master due to latter; but I would like Travis to run)

-----

`mpp_split.unique_hierarchy()` is returning different results depending on python version

see:

```
> python3.7
Python 3.7.7 (tags/v3.7.7:d7c567b08f, Mar 10 2020, 10:41:24) [MSC v.1900 64 bit (AMD64)] on win32
>>> unique_configs={((0, 798000000), (1, 0), (2, 302000000), (3, 0)), ((0, 1000000000), (1, 0), (2, 100000000), (3, 0))}
>>> item = ((0, 600000000), (1, 500000000), (2, 0), (3, 0))
>>> unique_configs.add(item)
>>> unique_configs
{((0, 600000000), (1, 500000000), (2, 0), (3, 0)), ((0, 798000000), (1, 0), (2, 302000000), (3, 0)), ((0, 1000000000), (1, 0), (2, 100000000), (3, 0))}
```
```
> python3.8
Python 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)] on win32
>>> unique_configs={((0, 798000000), (1, 0), (2, 302000000), (3, 0)), ((0, 1000000000), (1, 0), (2, 100000000), (3, 0))}
>>> item = ((0, 600000000), (1, 500000000), (2, 0), (3, 0))
>>> unique_configs.add(item)
>>> unique_configs
{((0, 798000000), (1, 0), (2, 302000000), (3, 0)), ((0, 1000000000), (1, 0), (2, 100000000), (3, 0)), ((0, 600000000), (1, 500000000), (2, 0), (3, 0))}
```

### DIFF
--- a/electrum/mpp_split.py
+++ b/electrum/mpp_split.py
@@ -24,7 +24,7 @@ def unique_hierarchy(hierarchy: Dict[int, List[Dict[bytes, int]]]) -> Dict[int, 
         for config in configs:
             # config dict can be out of order, so sort, otherwise not unique
             unique_configs.add(tuple((c, config[c]) for c in sorted(config.keys())))
-        for unique_config in unique_configs:
+        for unique_config in sorted(unique_configs):
             new_hierarchy[number_parts].append(
                 {t[0]: t[1] for t in unique_config})
     return new_hierarchy
--- a/electrum/tests/test_mpp_split.py
+++ b/electrum/tests/test_mpp_split.py
@@ -29,8 +29,8 @@ class TestMppSplit(ElectrumTestCase):
 
         with self.subTest(msg="do a payment with a larger amount than what is supported by a single channel"):
             splits = mpp_split.suggest_splits(1_100_000_000, self.channels_with_funds, exclude_single_parts=True)
-            self.assertEqual({0: 798_000_000, 1: 0, 2: 302_000_000, 3: 0}, splits[0][0])
-            self.assertEqual({0: 908_000_000, 1: 0, 2: 192_000_000, 3: 0}, splits[1][0])
+            self.assertEqual({0: 600_000_000, 1: 500_000_000, 2: 0, 3: 0}, splits[0][0])
+            self.assertEqual({0: 710_000_000, 1: 390_000_000, 2: 0, 3: 0}, splits[1][0])
 
         with self.subTest(msg="do a payment with the maximal amount spendable over all channels"):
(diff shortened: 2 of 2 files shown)


SomberNight:
travis is ignoring this PR :(
but it ran on my branch now and passed: https://travis-ci.org/github/SomberNight/electrum/builds/760174080

SomberNight:
@bitromortac so re https://github.com/spesmilo/electrum/pull/7060#issuecomment-784043562 and the original version-dependent failure,
maybe we could call `random.seed()` as originally, and sort the elements in the set as here;
that might be enough

bitromortac:
Nice, I have not tested, but yes we can do that (just revert #7060). How did you find that out about the sets :grin:? I saw the comment of the still failing tests and wanted to get back to it tomorrow morning. These test failures hint that the algo constants need to be fine tuned (probably increasing the number of starting configurations should make it more stable). I'll take care of that.

SomberNight:
> How did you find that out about the sets 😁?

I have patched `_randbelow` in the standard library `random` module to print `n` and `r`, see:
https://github.com/python/cpython/blob/9dd018e35cce30bc2545290b6083dbf6e50d7b61/Lib/random.py#L258
Ran the unit test a few dozen times, going back and forth between python 3.7 and 3.8.
Saw that `_randbelow` is returning deterministic randomness (due to using `random.seed()`), as expected, however after around 50 calls to it, the `n` argument it gets diverges between py37 and py38.

I then patched `_randbelow` to store the sequence of `n` values it is called with, and hardcoded the sequence of `n` values I observed preceding the divergence into `_randbelow`. I patched `_randbelow` so that it would print `traceback.print_stack()` at the point of divergence (for the first differing `n` value), to see exactly where it is called from. That landed me in `mpp_split.propose_new_configuration.swap`.

I observed that at that point the states have already diverged (`config` were different between py37 and py38), so I have started adding print statements to the code execution path preceding that point; kind of bisecting until pinpointing where the divergence starts.

So, in short, I did it the boring way. :)

bitromortac:
Not boring at all :). Thanks for the writeup! I tested it also with python versions 3.6/3.7/3.8.

SomberNight:
I've committed a minimal change just to eliminate the flakyness; but feel free to still tweak the tests if you want to.

PR #7065, opened 2021-02-24
lnpeer: MPP recv: only fulfill htlc if amt sum exact-matches total_msat

SomberNight:
when receiving htlcs as part of MPP
- it is not the invoice_amount the htlc set sum should be tested against but `total_msat`,
  i.e. htlc set sum and `total_msat` should be compared
- specifically, BOLT-04 seems to suggest testing equality of htlc set sum and `total_msat` (instead of `>=`)
  > if the total `amount_msat` of this HTLC set equals `total_msat`
- re zero amount invoices, they too might be MPP, so the same tests still need to be done
- (the FIXME in `lnchannel.py` is pre-existing brokenness)

https://github.com/lightningnetwork/lightning-rfc/blob/90468030d5c374ced568c91df5d37699fe020010/04-onion-routing.md#basic-multi-part-payments

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1051,7 +1051,9 @@ class Channel(AbstractChannel):
             if is_sent:
                 self.lnworker.htlc_fulfilled(self, payment_hash, htlc.htlc_id, htlc.amount_msat)
             else:
-                self.lnworker.htlc_received(self, payment_hash)
+                # FIXME
+                #self.lnworker.htlc_received(self, payment_hash)
+                pass
 
     def balance(self, whose: HTLCOwner, *, ctx_owner=HTLCOwner.LOCAL, ctn: int = None) -> int:
         assert type(whose) is HTLCOwner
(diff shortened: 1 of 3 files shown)


electrum/lnchannel.py
```diff
@@ -1051,7 +1051,9 @@ def extract_preimage_from_htlc_txin(self, txin: TxInput) -> None:
             if is_sent:
                 self.lnworker.htlc_fulfilled(self, payment_hash, htlc.htlc_id, htlc.amount_msat)
             else:
-                self.lnworker.htlc_received(self, payment_hash)
+                # FIXME
+                #self.lnworker.htlc_received(self, payment_hash)
+                pass
```
ecdsa, line 1056: why is that commented out?

SomberNight: > (the FIXME in lnchannel.py is pre-existing brokenness)

I just wanted to highlight that it is broken

ecdsa: can you clarify?

SomberNight: yes. `htlc_received` takes three required arguments but that call only provides one

ecdsa: ah righht

PR #7090, opened 2021-03-08
installwizard: show hw xpub after adding cosigner

zebra-lucky:
Show hardware wallet xpub when hw wallet is used as cosigner.

It could be useful, for example, when setting up 2of2 on desktop and mobile.


ecdsa:
not merging; the install wizard code has been updated.

Issue #6456, opened 2020-07-31
Multisig wallet setup needs better UX


fresheneesz:
Going through a multisig creation process is confusing with Electrum. For example, if you have 3 different devices you want to use to setup a multisig wallet, you may not be able to plug in all the wallets at once, so there should be some way to asynchronously set up a multisig wallet. However, electrum seems to assume that all the information is available right away all at once. It asks you to create a seed, gives the master public key, and then immediately asks for all the other master public keys. Well what if I don't have them yet? 

It seems like Electrum is assuming that you'll have single-seed wallets that you can use the master public key for. However, if a user just wants to create a new multisig wallet, they're likely to just hit "create multisig wallet" and then get stuck one it asks for the master public keys for other devices. 

There needs to be a better flow like the following:

1. User begins creation of a new wallet,
2. chooses multisig wallet and m of n,
3. chooses seed location (eg connected hardware wallet, create seed on device, etc),
4. chooses segwit vs legacy
5. user writes down the seed

At this point the wallet should be created.  You should be able to exit the app and return to it and select the multisig wallet you created. The interface should tell you the multisig wallet still needs master public keys to be set up and should provide a way for the user to enter them.

As it stands, if you don't have access to ALL the master public keys on creation of your multisig wallet, you can't even cancel the process or go back. Killing the app loses all your progress and you then have to go through the restoring process if you want to use the seed you created (and hopefully wrote down). This is pretty confusing for a user, and is incredibly inconvenient to the point of being unusable if there are multiple parties trying to create the multisig wallet and aren't doing it all at the same time (eg because they have a more secure process for backing up their seed than writing it on a piece of paper in full view of other people). 

Another option, that might even be better than what I described above is to treat multisig wallets as extensions of single-seed wallets, where to create a multisig wallet you would start with a single-seed wallet, choose an option to create a multisig wallet which would ask for master public keys, then would open it as a separate wallet at that point. Similarly to access a multisig wallet, you would go into your single-seed wallet, and choose the multisig wallet within it. A single-seed wallet could even have multiple multisig wallets it is part of. The best way to create multisig wallets is to already have single-seed wallets you can get the master public keys of. Why not make this explicit and move multisig wallet creation as a subwallet of standard wallets?

fresheneesz:
So continuing to actually go through this, its even worse.

1. Adding a Trezor wallet as a cosigner doesn't actually display any Zpub key when adding it as a multisig cosigner (like it does for Coldcard). 
[...]

PR #7095, opened 2021-03-10
Add support to save uncompleted multisig wallets

zebra-lucky:
- Add support to save uncompleted multisig wallets and then
continue setup when cosigners xpub is available.

Maybe related to #6456.

Some screens:

1. "Save and exit" button is added when saving is possible:

2. Continue multisig setup, when already stored xpub can be viewed:

### DIFF
--- a/electrum/base_wizard.py
+++ b/electrum/base_wizard.py
@@ -24,6 +24,7 @@
 # SOFTWARE.
 
 import os
+import json
 import sys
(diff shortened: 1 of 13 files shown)


accumulator:
Thanks for the PR, but it is so outdated now it will require a full rewrite..

closing this, let's use #6456 to track this issue.

PR #7098, opened 2021-03-11
mpp_split: add penalty for exhaustion of channels

bitromortac:
A penalty is added for split configurations which saturate a channel.
Saturation of channels is discouraged as we don't know the fees
beforehand. The penalty is accomplished via an exponential function that
kicks in when the subamount reaches about the total funds available
(this amount is controlled by the parameter EXHAUST_DECAY_FRACTION).


ecdsa:
thanks, I merged it manually

PR #7099, opened 2021-03-11
fail pending htlcs on shutdown

SomberNight:
We now try to fail pending HTLCs when shutting down.

This is most useful when receiving MPP where there is a non-trivial chance
that we have received some HTLCs for a payment but not all, and the user
closes the program. We try to fail them and wait for the fails to get
ACKed, with a timeout of course.


ecdsa:
it would be nice if we could encapsulate `received_htlc_removed_event` and `received_htlcs_pending_removal` in the same object. 
I looked into asyncio docs but didnt find anything that does that.

Issue #7087, opened 2021-03-06
if trampoline is used, prioritize mpp_splits that use the same trampoline


ecdsa:
If trampoline is used, mpp_split should prioritize solutions that send HTLCs to the same trampoline.
This will reduce fees, and also reduce the probability of failure.

ecdsa:
@bitromortac

bitromortac:
The issue with payment indexing was fixed in eclair, see https://github.com/ACINQ/eclair/issues/1723.

https://github.com/spesmilo/electrum/blob/9c1a51547a301e765b9b0f9935c6d940bb9d658e/electrum/lnworker.py#L1526-L1530
This should be no longer relevant and we could test sending via different trampoline nodes.

PR #7107, opened 2021-03-15
mpp_split: use single nodes for mpp payments over trampoline

bitromortac:
Adds the node id as an input to mpp_split (extending the `short_channel_id` key), to be able to filter out split configurations that pay via several nodes. We need to do this in case of trampoline multipart payments, see https://github.com/ACINQ/eclair/issues/1723. Related: #7087.

TestPeer now inherits from `TestCaseForTestnet` instead of `ElectrumTestCase` to disable filtering of single nodes in tests, to not break `test_multipart_payment_with_trampoline`, which we want to keep in its current form (as intended for the future without the above mentioned bug). Let me know if there's a better way to do this.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1166,7 +1166,7 @@ class LNWallet(LNWorker):
             if code == OnionFailureCode.MPP_TIMEOUT:
                 raise PaymentFailure(failure_msg.code_name())
             # trampoline
-            if self.channel_db is None:
+            if not self.channel_db:
                 if code == OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT:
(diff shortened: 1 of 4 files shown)


ecdsa:
It looks good. Note that we use `self.channel_db is None` as a way to detect trampoline all over the place. It may be clearer to use an explicit variable name or method, but in that case it should be done everywhere, not just in a single method. I think introducing heterogeneous notations is worse. Also, if a variable is a boolean, the name should reflect it, eg `is_trampoline` instead of `trampoline_payment`.

bitromortac:
I reverted to `not self.channel_db` which is also used in other places, to unify. Also a good point concerning naming of boolean variables.

PR #7114, opened 2021-03-18
bitbox02: bump dependency to 5.2.0

benma:


benma:
Contains support for [anti-klepto](https://shiftcrypto.ch/blog/anti-klepto-explained-protection-against-leaking-private-keys/).

PR #7123, opened 2021-03-20
Update device IDS to correct ones

porokka:
I tried to connect to wallet many times but never successfully but then I googled and found out that ledger.py had wrong IDS for my Nano x and changed that and it started working. I added IDS that I found and have tested for nano x

### DIFF
--- a/electrum/plugins/ledger/ledger.py
+++ b/electrum/plugins/ledger/ledger.py
@@ -585,9 +585,9 @@ class LedgerPlugin(HW_PluginBase):
                    (0x2581, 0x2b7c), # HW.1 transitional production
                    (0x2581, 0x3b7c), # HW.1 ledger production
                    (0x2581, 0x4b7c), # HW.1 ledger test
-                   (0x2c97, 0x0000), # Blue
-                   (0x2c97, 0x0001), # Nano-S
-                   (0x2c97, 0x0004), # Nano-X
+                   (0x2c97, 0x0015), # Blue
+                   (0x2c97, 0x1015), # Nano-S
+                   (0x2c97, 0x4015), # Nano-X
                    (0x2c97, 0x0005), # RFU
                    (0x2c97, 0x0006), # RFU
                    (0x2c97, 0x0007), # RFU


SomberNight:
Note the code here:
https://github.com/spesmilo/electrum/blob/9609a60ab640bd86e37ebc54870fea1c04f9b522/electrum/plugins/ledger/ledger.py#L643-L649
https://github.com/spesmilo/electrum/blob/9609a60ab640bd86e37ebc54870fea1c04f9b522/electrum/plugins/ledger/ledger.py#L599-L601

So you say that `(0x2c97, 0x4015), # Nano-X` works for you, but
```
>>> 0x4015 >> 8 == 0x40
True
```
So `(0x2c97, 0x4015)`  should be already covered by that.

SomberNight:
What version of Electrum are you using? What operating system?

wakiyamap:
It might be helpful.
https://www.ledger.com/windows-10-update-sunsetting-u2f-tunnel-transport-for-ledger-devices

SomberNight:
@wakiyamap yes, that is exactly why the code snippets from my previous comment were introduced in https://github.com/spesmilo/electrum/commit/b78cbcffd11364ed4651af1886d507b36bc1b013

I don't have a Ledger Nano X, but I do have a Nano S, and the current code works for me.

porokka:
I use version 3.3.8-rvn  for windows




On 21 Mar 2021 00:07, Jin Eguchi ***@***.***> wrote:

It might be helpful.
https://www.ledger.com/windows-10-update-sunsetting-u2f-tunnel-transport-for-ledger-devices

SomberNight:
> I use version 3.3.8-rvn  for windows

Right. This got fixed in the above referenced commit, which your version does not have.
Please don't create patches based on 1 year+ old code without testing on master first.

porokka:
Ok sorry.

On 21 Mar 2021 00:13, ghost43 ***@***.***> wrote:

I use version 3.3.8-rvn for windows

Right. This got fixed in the above referenced commit, which your version does not have.
Please don't create patches based on 1 year+ old code without testing on master first.

PR #7133, opened 2021-03-23
qt channels list: add "features" column with icons (e.g. trampoline)

SomberNight:
This PR adds a new column in the Qt `Channels` tab, "Features", which can contain a list of icons.
Currently two icons are defined:
- whether the channel peer can route trampoline payments
- whether the channel has stored a static backup with op_return on-chain (or rather, the inverse of this)

Capture


ecdsa:
note: `chan.lnworker.has_recoverable_channels()` is not telling us whether the channel has an onchan backup

PR #7135, opened 2021-03-24
wallet: implement get_utxos at specific block height

SomberNight:
This allows looking up what UTXOs the wallet had at a specific time in the past.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -835,27 +835,47 @@ class AddressSynchronizer(Logger):
         return result
 
     @with_local_height_cached
-    def get_utxos(self, domain=None, *, excluded_addresses=None,
-                  mature_only: bool = False, confirmed_only: bool = False,
-                  nonlocal_only: bool = False) -> Sequence[PartialTxInput]:
+    def get_utxos(
+            self,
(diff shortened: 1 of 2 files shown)


electrum/address_synchronizer.py
```diff
@@ -835,27 +835,47 @@ def get_addr_balance(self, address, *, excluded_coins: Set[str] = None) -> Tuple
...
+        else:
+            block_height = self.get_local_height()
         coins = []
         if domain is None:
             domain = self.get_addresses()
         domain = set(domain)
         if excluded_addresses:
             domain = set(domain) - set(excluded_addresses)
-        mempool_height = self.get_local_height() + 1  # height of next block
+        mempool_height = block_height + 1  # height of next block
```
ecdsa, line 862: this variable should not be called like that anymore

SomberNight: I think it's fine. There is a comment that it is the height of the next block; and that value is what the "mempool height" was in the past assuming given `block_height`.
It is currently only used for UTXO maturity; which raises a related question btw: should we include coins that have not been mature back then but are mature now, if `mature_only` is set. Well it's quite an edge case anyway.

PR #7143, opened 2021-03-26
Issue #7141 RE #fixing poor_design_fix_address_to_scripthash

### LINKED ISSUE #7141: Poor design of address_to_scripthash and address_to_script
I consider the following to be an example of the poor design:
- electrum/bitcoin.py looks like a library of useful utility functions, while some of its functions, which could be reused in many places, are tight-coupled with the currently loaded network type (eg. Mainnet)
- I mean the functions address_to_script and address_to_scripthash
- address_to_script allows a programmer to provide both address ("addr" parameter) and network ("net" parameter). When "net" is not provided, the global variable "net" from another module (electrum.constants) is being used.  The variable is being set using the functions defined in this module (set_mainnet, set_testnet etc). This design is poor because utility functions should not rely this way on the parameters defined somewhere else, they should take arguments and return results, without changing any state in the application ("stateless function"), however such behaviour might yet be eventually tolerated (though it should not).
- The bigger problem is the structure of "address_to_scripthash" function. It makes use of address_to_script while allowing to provide the address only, assuming that the network type has been already set into electrum.constants. Therefore, the function is useless as utility function because you can't call it the stateless way, without changing the current settings of the application. Here's the code:
```
def address_to_scripthash(addr: str) -> str:
    script = address_to_script(addr)
    return script_to_scripthash(script)
```

- As a solution, I'd suggest that address_to_scripthash had the "net" parameter as well and passed it into address_to_script. The behaviour of getting the default network type may be kept, though I'd suggest moving it outside the utility function into some other, possibly more central place of the application. I'd like to fix this, but I don't have permission to push new branches into the repo. 
- Also, utilty modules including electrum.bitcoin might be moved into some directory suggesting it's a set of libraries or utility functions, eg. electrum/util or electrum/lib

krzysobo:
Fixing the poor design of address_to_scripthash.
closes #7141

### DIFF
--- a/electrum/bitcoin.py
+++ b/electrum/bitcoin.py
@@ -506,9 +506,10 @@ def address_to_hash(addr: str, *, net=None) -> Tuple[OnchainOutputType, bytes]:
     raise BitcoinException(f"unknown address type: {addrtype}")
 
 
-def address_to_scripthash(addr: str) -> str:
-    script = address_to_script(addr)
+def address_to_scripthash(addr: str, *, net=None) -> str:
+    script = address_to_script(addr, net=net)
     return script_to_scripthash(script)
+
 
 def script_to_scripthash(script: str) -> str:
     h = sha256(bfh(script))[0:32]


SomberNight:
Thanks

PR #7147, opened 2021-03-28
cmd: add sign with privkey and wallet

n1rna:
I saw a new `TODO` was introduced in https://github.com/spesmilo/electrum/commit/638d4f747238f1255b58d40035366903a832dfb4 on the `signtransaction` command and tried to address the `TODO` by separating the command to two new commands: `signtransaction_with_privkey` and `signtransaction_with_wallet`.

Also did a tweak to the `signtransaction_with_privkey` logic to accept list of private keys and sign with many private keys instead of only one.

### DIFF
--- a/electrum/tests/test_commands.py
+++ b/electrum/tests/test_commands.py
@@ -11,6 +11,7 @@ from electrum.simple_config import SimpleConfig
 from electrum.transaction import Transaction, TxOutput, tx_from_any
 
 from . import TestCaseForTestnet, ElectrumTestCase
+from .test_wallet_vertical import WalletIntegrityHelper
 
 
 class TestCommands(ElectrumTestCase):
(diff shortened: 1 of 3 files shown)


ecdsa:
The reason why this command should be split is that one could use it without having an open wallet.
This should be achieved by using the decorator `@command('p')` instead of  `@command('wp')`

I suggest keeping the name `signtransaction` for the one that uses a wallet, in order to minimize the number of scripts that are broken by this change.

ecdsa:
also, there is no reason why `privkey` should remain an optional argument

n1rna:
>  This should be achieved by using the decorator @command('p') instead of @command('wp')
First I wrote the code using the `p` argument only, but some test cases were broken. That's why I reverted. Will fix the tests and change to `p`.

I agree with your suggestion. Can you let me know if there are some places on the docs where this change needs be mentioned?

n1rna:
I guess when the command doesn't require the wallet, it wouldn't require the password also. So the correct decorator would be `@commands('')`. I fixed it in https://github.com/spesmilo/electrum/pull/7147/commits/47b240de821c3c94665c6f4c749a5f56ed3e5781

ecdsa:
> I guess when the command doesn't require the wallet, it wouldn't require the password also.

indeed. you can also remove the wallet from the list of arguments.

ecdsa:
also please note that the 3 lines you touched in `init_cmdline` are here only to override the decorator.
if we split this command they are not needed anymore.

n1rna:
any updates on this one? please let me know so I can update the branch and fix conflicts so it can be merged :)

electrum/tests/test_commands.py
```diff
@@ -218,11 +222,36 @@ def test_getprivatekeyforpath(self, mock_save_db):
...
         privkey = "cVtE728tULSA4gut4QWxo218q6PRsXHQAv84SXix83cuvScvGd1H"
         self.assertEqual("020000000221d3645ba44f33fff6fe2666dc080279bc34b531c66888729712a80b204a32a1010000006a47304402205b30e188e30c846f98dacc714c16b7cd3a58a3fa24973d289683c9d32813e24c0220153855a29e96fb083084417ba3e3873ccaeb08435dad93773ab60716f94a36160121033f6737e40a3a6087bc58bc5b82b427f9ed26d710b8fe2f70bfdd3d62abebcf74fdffffffdd7f90d51acf98dc45ad7489316a983868c75e16bf14ffeb9eae01603a7b4da4010000006a473044022010daa3dadf53bdcb071c6eff6b8787e3f675ed61feb4fef72d0bf9d99c0162f802200e73abd880b6f2ee5fe8c0abab731f1dddeb0f60df5e050a79c365bd718da1c80121033f6737e40a3a6087bc58bc5b82b427f9ed26d710b8fe2f70bfdd3d62abebcf74fdffffff02e8030000000000001976a9149a9ec2b35a7660c80dae38dd806fdf9b0fde68fd88ac74c11000000000001976a914f0dc093f7fb1b76cfd06610d5359d6595676cc2b88aca79b1d00",
-                         cmds._run('signtransaction', (), tx=unsigned_tx, privkey=privkey, wallet=dummy_wallet))
+                         cmds._run('signtransaction_with_privkey', (), tx=unsigned_tx, privkey=privkey, wallet=dummy_wallet))
```
SomberNight, line 232: I suggest removing the wallet argument here. `dummy_wallet` is not needed in this test at all anymore (so rm first line of method body too).

electrum/tests/test_commands.py
```diff
@@ -218,11 +222,36 @@ def test_getprivatekeyforpath(self, mock_save_db):
...
+
+    @mock.patch.object(wallet.Abstract_Wallet, 'save_db')
+    def test_signtransaction_with_wallet(self, mock_save_db):
+        wallet = self.create_standard_wallet_from_seed('bitter grass shiver impose acquire brush forget axis eager alone wine silver',
```
SomberNight, line 236:
```suggestion
        wallet = restore_wallet_from_text(
            'bitter grass shiver impose acquire brush forget axis eager alone wine silver',
            gap_limit=2, path='if_this_exists_mocking_failed_648151893', config=self.config)['wallet']
```

and with this change, the `create_standard_wallet_from_seed` method is not needed anymore.

electrum/tests/test_commands.py
```diff
@@ -218,11 +222,36 @@ def test_getprivatekeyforpath(self, mock_save_db):
...
+
+        unsigned_tx = "cHNidP8BAHECAAAAAQOSwZQOLsnyNykZyjiHMn/luYuGYCLMebq1y+1aU9KtAAAAAAD+////AigjAAAAAAAAFgAUaQtZqBQGAvsjzCkE7OnMTa82EFIwGw8AAAAAABYAFKwOLSKSAL/7IWftb9GWrvnWh9i7AAAAAAABAN8BAAAAAUV22sziZMJNgYh2Qrcm9dZKp4JbIbNQx7daV/M32mhFAQAAAGtIMEUCIQCj+LYVXHGpitmYbt1hYbINJPrZm2RjwjtGOFbA7lSCbQIgD2BgF/2YdpbrvlIA2u3eki7uJkMloYTVu9qWW6UWCCEBIQLlxHPAUdrjEEPDNSZtDvicHaqy802IXMdwayZ/MmnGCf////8CQEIPAAAAAAAWABSKKL3bf2GGS9z1iyrRPVrrOrw8QqLduQ4AAAAAGXapFMOElQNCy2+N9VF1tIWGg4sDEw+tiKwAAAAAIgYDD67ptKJbfbggI8qYkZJxLN1MtT09kzhZHHkJ5YGuHAwQsuNafQAAAIAAAAAAAAAAAAAiAgKFhOeJ459BORsvJ4UsoYq+wGpUEcIb41D+1h7scSDeUxCy41p9AAAAgAEAAAAAAAAAAAA="
+
+        self.assertEqual("020000000001010392c1940e2ec9f2372919ca3887327fe5b98b866022cc79bab5cbed5a53d2ad0000000000feffffff022823000000000000160014690b59a8140602fb23cc2904ece9cc4daf361052301b0f0000000000160014ac0e2d229200bffb2167ed6fd196aef9d687d8bb02473044022027e1e37172e52b2d84106663cff5bcf6e447dcb41f6483f99584cfb4de2785f4022005c72f6324ad130c78fca43fe5fc565526d1723f2c9dc3efea78f66d7ae9d4360121030faee9b4a25b7db82023ca989192712cdd4cb53d3d9338591c7909e581ae1c0c00000000",
+                         cmds._run('signtransaction', (), tx=unsigned_tx))
```
SomberNight, line 251:
```suggestion
                         cmds._run('signtransaction', (), tx=unsigned_tx, wallet=wallet))
```

electrum/tests/test_commands.py
n1rna: Nice. I should've used the `restore_wallet_from_text` in the first place.

SomberNight (approved):
Looks good. Thanks!

PR #7150, opened 2021-03-29
allow 15 cosigners on android (typofix in GUI)

chuhi4Ai:
looks like a "1" got missing. electrum desktop allows 15, android just 5 cosigners with multisig wallet.

for reference maximum number of cosigners in qt version is here: https://github.com/spesmilo/electrum/blob/1c52203346f9049f6c3fe201f672960339d8273b/electrum/gui/qt/installwizard.py#L757

i tested it its just kivy ui everything else works. thank you very much


SomberNight:
Please search existing tickets before opening new ones.
It is not a typo. See https://github.com/spesmilo/electrum/pull/5133

PR #7151, opened 2021-03-29
detect conflicting channel backups, add warning before channel creation

ecdsa:


electrum/lnworker.py
```diff
@@ -2121,6 +2121,12 @@ def import_channel_backup(self, data):
         util.trigger_callback('channels_updated', self.wallet)
         self.lnwatcher.add_channel(cb.funding_outpoint.to_str(), cb.get_funding_address())
 
+    def has_conflicting_backup_with(self, pk: bytes):
```
SomberNight, line 2124: This method could use a docstring; it is not quite clear what it is doing. For example:
```
Returns whether we have a channel with this node on another device (using same local node id).
```


Also, I suggest renaming `pk` to `remote_node_id`.

ecdsa: done

PR #7152, opened 2021-03-30
Liquidity hints for pathfinding

bitromortac:
Implements liquidity hints for pathfinding, i.e., take information of successful and failed routing attempts to improve pathfinding, similar to the strategy LND employs.

We interpret temporary channel failures such that a node couldn't forward a payment as the channel (or node pair) didn't have enough balance on its side of the channel. Based on this assumption, we can keep a record of the previous part of the route (which was successful) and the failing channel. For each channel we record the `can_send` and cannot_send values upon which we build a new penalty for edge weights in pathfinding. Likewise, when a payment succeeds, we record `can_send` amounts for the channels involved (the assumption is that we can send the same amount again, even if we changed the balances by sending).

Model:
The liquidity hint for a channel keeps four pieces of information, the amounts it can and cannot send for each direction. The direction dependence is kept, as we don't know the capacity of the channel. Furthermore, there can be [non strict forwarding](https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#non-strict-forwarding) and shadow (private) channels present. Therefore we cannot relate `local_balance = capacity - remote_balance`.

Penalty:
The penalty depends on the `can_send` and `cannot_send` values that were possibly recorded in previous payment attempts. A channel that can send an amount is assigned a penalty of zero, a channel that cannot send an amount is assigned an infinite penalty. If the sending amount lies between `can_send` and `cannot_send`, there's uncertainty and we give a default penalty. The default penalty serves the function of giving a positive offset, from which we can discount from. There is a competition between low-fee channels and channels where we know with some certainty that they can support a payment. The penalty ultimately boils down to: how much more fees do we want to pay for certainty of payment success? This can be tuned via `DEFAULT_PENALTY_BASE_MSAT` and `DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH`. A base _and_ relative penalty was chosen such that the penalty will be able to compete with the regular base and relative fees. `DEFAULT_PENALTY_BASE_MSAT` is set to 500 msat (same as before) and `DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH` is 100 (i.e. 0.000100 relative factor), which seems fair but is open for discussion.

Solutions / Benefits:
* Bundles all the transient routing information we get about the graph (blacklist, hints) in one data structure
* Should make payments more reliable in general
* Once a payment has found its path to the destination, a repetition of the payment is much quicker
* Doesn't blacklist channels any more, but avoids sending over channels based on the amount, which is beneficial for multipart payments
* Introduces the LiquidityHints data structure which in the future can be used to keep track of in-flight HTLCs for mpp (to avoid sending multiple HTLCs over the same channel)

Downsides:
* Can lead to increased latency in pathfinding (although liquidity hints are only checked in the end of `_edge_cost`)
* Increases path re-usage (could randomize penalty)

Open questions:
* Right now an amount-independent default penalty is applied. We could make it dependent on the amount, based on how far we are away from known facts (e.g. the `can_send` amount)
* It may be useful to persist the information we gain for the next usage of the wallet
* If the wallet runs long enough, or liquidity information is loaded, we may disregard information after some time

### DIFF
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -27,9 +27,11 @@ import queue
 from collections import defaultdict
 from typing import Sequence, List, Tuple, Optional, Dict, NamedTuple, TYPE_CHECKING, Set
 import time
+from threading import RLock
 import attr
+from math import inf
 
-from .util import bh2u, profiler
+from .util import bh2u, profiler, with_lock
 from .logging import Logger
 from .lnutil import (NUM_MAX_EDGES_IN_PAYMENT_PATH, ShortChannelID, LnFeatures,
                      NBLOCK_CLTV_EXPIRY_TOO_FAR_INTO_FUTURE)
@@ -37,6 +39,10 @@ from .channel_db import ChannelDB, Policy, NodeInfo
 
 if TYPE_CHECKING:
     from .lnchannel import Channel
+
+DEFAULT_PENALTY_BASE_MSAT = 500  # how much base fee we apply for unknown sending capability of a channel
+DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH = 100  # how much relative fee we apply for unknown sending capability of a channel
+BLACKLIST_DURATION = 3600  # how long (in seconds) a channel remains blacklisted
 
 
 class NoChannelPolicy(Exception):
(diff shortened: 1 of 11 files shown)


electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+        else:
+            self.cannot_send_backward = amount
+
+    def __repr__(self):
+        return f"forward: can send: {self._can_send_forward}, cannot send: {self._cannot_send_forward}, \n" \
+               f"backward: can send: {self._can_send_backward} cannot send: {self._cannot_send_backward}, \n" \
+               f"blacklisted: {self.blacklisted}"
+
+
+class LiquidityHints:
```
SomberNight, line 275: Please no names where one name is X and the other is X in plural.
Maybe `LiquidityHintMgr` instead.

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+    A LiquidityHint is the value of a dict, which is keyed to node ids and the
+    channel.
+    """
+    def __init__(self):
+        # use "can_send_forward + can_send_backward < cannot_send_forward + cannot_send_backward" as a sanity check?
+        self._can_send_forward = None
+        self._cannot_send_forward = None
+        self._can_send_backward = None
+        self._cannot_send_backward = None
+        self.blacklisted = False
```
SomberNight, line 183: I suggest using `is_blacklisted` as name instead

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+    graph. Implements a penalty function for edge weighting in the pathfinding
+    algorithm that favors channels which can route payments and penalizes
+    channels that cannot.
+    """
+    # TODO: incorporate in-flight htlcs
+    # TODO: use timestamps for can/not_send to make them None after some time?
+    # TODO: hints based on node pairs only (shadow channels, non-strict forwarding)?
+    def __init__(self):
+        self.lock = RLock()
+        self._liquidity_hints: Dict[Tuple[bytes, bytes, ShortChannelID], LiquidityHint] = {}
```
SomberNight, line 288: Why does the key contain both node ids and scid?
I first thought you only want the two nodes to determine direction but then I saw the nodes are ordered.

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+            if amount <= can_send:
+                return 0
+            elif amount < cannot_send:
+                return fee_for_edge_msat(amount, DEFAULT_PENALTY_BASE_MSAT, DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH)
+            else:
+                return inf
+        return 0
+
+    @with_lock
+    def blacklist(self, node_from: bytes, node_to: bytes, channel_id: ShortChannelID):
```
SomberNight, line 365: Suggested name: `add_to_blacklist`.

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+        self._liquidity_hints: Dict[Tuple[bytes, bytes, ShortChannelID], LiquidityHint] = {}
+
+    @staticmethod
+    def _hint_key(node1: bytes, node2: bytes, channel_id: ShortChannelID):
+        return (node1, node2, channel_id) if node1 < node2 else (node2, node1, channel_id)
+
+    @with_lock
+    def get_hint(self, node1: bytes, node2: bytes, channel_id: ShortChannelID):
+        key = self._hint_key(node1, node2, channel_id)
+        return self._liquidity_hints.get(key, LiquidityHint())
```
SomberNight, line 297: There is a potential API issue here: maybe the default value of the get should be saved in the map before returning it.
If this method creates a new `LiquidityHint()` and the caller starts mutating it, the caller might expect the mutations not to get lost.
If you saved the newly created `LiquidityHint()` in the map before returning it, you could use `get_hint()` in `update_can_send`, `update_cannot_send`, and `blacklist`.

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+    def update_liquidity_hints(
+            self,
+            route: LNPaymentRoute,
+            amount_msat: int,
+            failing_channel: ShortChannelID=None
+    ):
+        # go through the route and record successes until the failing channel,
+        # for the failing channel, add a cannot_send liquidity hint
+        for r in route:
+            if not r.short_channel_id == failing_channel:
```
SomberNight, line 408: instead of `not A == B`, just write `A != B`

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+        self.liquidity_hints = LiquidityHints()
+
+    def update_liquidity_hints(
+            self,
+            route: LNPaymentRoute,
+            amount_msat: int,
+            failing_channel: ShortChannelID=None
+    ):
+        # go through the route and record successes until the failing channel,
+        # for the failing channel, add a cannot_send liquidity hint
```
SomberNight, line 406: maybe add a short fixme that the actual forwarded amounts are a bit different as the routing fees are ignored atm

electrum/lnrouter.py
```diff
@@ -161,12 +167,251 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+                return fee_for_edge_msat(amount, DEFAULT_PENALTY_BASE_MSAT, DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH)
+        # we know how much we can/cannot send
+        elif can_send and cannot_send:
+            if amount <= can_send:
+                return 0
+            elif amount < cannot_send:
+                return fee_for_edge_msat(amount, DEFAULT_PENALTY_BASE_MSAT, DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH)
+            else:
+                return inf
+        return 0
```
SomberNight, line 362:
```suggestion
        if cannot_send is not None and amount >= cannot_send:
            return inf
        if can_send is not None and amount <= can_send:
            return 0
        return fee_for_edge_msat(amount, DEFAULT_PENALTY_BASE_MSAT, DEFAULT_PENALTY_PROPORTIONAL_MILLIONTH)
```

SomberNight:
> Penalty: [...]
> If the sending amount lies between `can_send` and `cannot_send`, there's uncertainty and we give a default penalty. 

In that case (and more generally when there is uncertainty, e.g. because both values are unknown), maybe we could take into consideration the channel capacity or `htlc_maximum_msat`. E.g. if the htlc is large compared to the channel capacity, it is likely to fail.

electrum/lnrouter.py
```diff
@@ -261,8 +507,6 @@ def get_distances(
         # main loop of search
         while nodes_to_explore.qsize() > 0:
             dist_to_edge_endnode, amount_msat, edge_endnode = nodes_to_explore.get()
-            if edge_endnode == nodeA:
-                break
```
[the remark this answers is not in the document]
bitromortac: I removed it because it led to unpredictable test outcomes in a previous state of the code. I put it back now, as it's not breaking anything.

electrum/lnrouter.py
bitromortac: Done

electrum/lnrouter.py
bitromortac: Done

electrum/lnrouter.py
bitromortac: Yes, I had it that way in the beginning and I'm still not 100% sure whether we should do hints per channel or per node pairs. I changed it now to key to the channel id only.

electrum/lnrouter.py
bitromortac: Done

electrum/lnrouter.py
bitromortac: Great point, I applied your suggestions. Only in the penalty function I'm using the dict get, as otherwise many insignificant hints would be created.

electrum/lnrouter.py
bitromortac: Done

electrum/lnrouter.py
bitromortac: Thank you for this, a nice simplification!

bitromortac:
> > Penalty: [...]
> > If the sending amount lies between `can_send` and `cannot_send`, there's uncertainty and we give a default penalty.
> 
> In that case (and more generally when there is uncertainty, e.g. because both values are unknown), maybe we could take into consideration the channel capacity or `htlc_maximum_msat`. E.g. if the htlc is large compared to the channel capacity, it is likely to fail.

Thank you for the review!
I think that could make sense, although I tried to avoid to resort to channel specific information, because of the non-strict forwarding issue (we would need additional lookups of the capacity, or save it into the hints -- I wanted to reduce data access for better performance too). We could in general refine the interpolation in the unknown territory. I'll think more about that.

I think in LND they use hints based on node pairs and I should look up their reason for that (I guess info on a per channel basis is too specific, we would retry over a different parallel channel between the same nodes, even though the routing node already tried its best to forward over the other channel and therefore we waste a payment attempt).
Do you think we should forget about the hints after some time? We could update the existing `blacklist_timestamp`, or introduce a separate `hint_timestamp` and return `None`s inside the penalty if the timestamp is too far in the past.

electrum/lnrouter.py
```diff
@@ -161,12 +167,229 @@ def is_fee_sane(fee_msat: int, *, payment_amount_msat: int) -> bool:
...
+    def add_to_blacklist(self, node_from: bytes, node_to: bytes, channel_id: ShortChannelID):
+        hint = self.get_hint(channel_id)
+        hint.is_blacklisted = True
+        now = int(time.time())
+        hint.timestamp = now
+
+    @with_lock
+    def get_blacklist(self) -> Set[ShortChannelID]:
+        now = int(time.time())
+        return set(k for k, v in self._liquidity_hints.items() if now - v.timestamp < BLACKLIST_DURATION)
```
SomberNight, line 353: This is a bit confusing. So `hint.is_blacklisted` is not used?
Could we not merge `.is_blacklisted` and `.timestamp` to a single thing somehow?
Maybe just have a `.blacklist_timestamp` field, initialised to 0, and deduce whether a the edge is blacklisted from that and the `BLACKLIST_DURATION` global.

SomberNight:
> I think that could make sense, although I tried to avoid to resort to channel specific information, because of the non-strict forwarding issue (we would need additional lookups of the capacity, or save it into the hints -- I wanted to reduce data access for better performance too). We could in general refine the interpolation in the unknown territory. I'll think more about that.
> 
> I think in LND they use hints based on node pairs and I should look up their reason for that (I guess info on a per channel basis is too specific, we would retry over a different parallel channel between the same nodes, even though the routing node already tried its best to forward over the other channel and therefore we waste a payment attempt).

Ok, we can think about it more and do it later, separately.

In general let's try to keep the code either simple or smart; I think it's best not to write complex code in anticipation of something we might implement in the future - as quite often those follow-ups never get implemented or they do but only much later by which time the code have changed too much. So by "simple" I mean implement something easy to understand and maintain; and by "smart" I mean something that is more complex but for a good reason.

> Do you think we should forget about the hints after some time? We could update the existing blacklist_timestamp, or introduce a separate hint_timestamp and return Nones inside the penalty if the timestamp is too far in the past.

Yes, I think that might be helpful if some people have a long-running process, and it should be easy to do.
Not sure what timeout to use; I guess it should be somewhere between 1 hour and 1 day -- we can use 1 hour for now to make it match the blacklist duration.

electrum/lnrouter.py
bitromortac: Oh yes, somehow this slipped through. I changed to use only the timestamp.

bitromortac:
I agree, I think it is helpful as it is already. The only pressing point for me was the invalidation of hints after some time, which I added in another commit. I guess the timescale is a bit short, but somebody should probe on which timescales the balances of the network are drastically changing, then we could adjust that number.

PR #7154, opened 2021-03-31
commands: extend "importprivkey" to allow importing a list of keys

SomberNight:
This is most useful if the user wants to import a significant number of keys,
as every invocation of `importprivkey` results in rewriting the wallet file to disk.

The API of "importprivkey" is left unchanged for the single privkey case.


SomberNight:
I guess we could merge this? :D

I don't remember what was the issue with it, I guess there might have been a discussion to just break backwards compat and have a cleaner API. Nevertheless, it is useful as-is.

PR #7159, opened 2021-03-31
Rename RELEASE-NOTES to RELEASE-NOTES.md

taxmeifyoucan:
Not sure why `RELEASE-NOTES` is not treated as `.md` document since it uses markdown formating. This allows better readability, at least for most users on github web.


SomberNight:
I think it is just a happy coincidence that it looks good as markdown. :P

taxmeifyoucan:
@SomberNight :D lol. So I suggest coming out as markdown. Nothing really changes, doc can be updated in the same manner as before, it will just be easier to handle by GUI apps supporting md.

ecdsa:
Like @SomberNight said, it is a happy coincidence. Renaming the file would break a lot of links...

PR #7187, opened 2021-04-07
android: update app icon; use "adaptive icons" on API 26+

SomberNight:
see https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive

p4a and buildozer had to be patched for this; I will try to upstream the patches, see:
https://github.com/kivy/python-for-android/pull/2446
https://github.com/kivy/buildozer/pull/1312

-----
Here is how the launcher looks on two different phones (both API 26+, so using the adaptive icon):
launcher1
launcher2


ecdsa:
with this PR, the legacy icon gets used on my phone (android 10), and it is used "as is" (full rectangle)

SomberNight:
have you rebuilt the docker image + buildozer?

ecdsa:
only buildozer; I did not see that the Docker image had changes :-)

ecdsa:
I guess it is working for me now, after rebuilding.
A couple of issues remain:
 - the new legacy icon looks worse than the one that was replaced (Electrum_512.png). why did you change it?
 - with the adaptive icons, my android crops part of the image (ellipses)

SomberNight:
> the new legacy icon looks worse than the one that was replaced (Electrum_512.png). why did you change it?

Mainly to make the adaptive/non-adaptive icons look similar; but I guess it's not important. Also, while I was testing this I could never be sure *why* it was working or not-working; so I replaced all icons with ones that have identical resolution to what the default project created by Android Studio has: 192x192 for legacy; 432x432 for adaptive (well actually Android Studio created them in 5 different resolutions... these being the largest).

> with the adaptive icons, my android crops part of the image (ellipses)

Well, looking at the images I've uploaded in the opening comment I guess it's not surprising. The adaptive foreground needs even larger margins...

SomberNight:
>> with the adaptive icons, my android crops part of the image (ellipses)
>
> Well, looking at the images I've uploaded in the opening comment I guess it's not surprising. The adaptive foreground needs even larger margins...

(from https://medium.com/google-design/designing-adaptive-icons-515af294c783)
see


I will make the margins larger.

PR #7191, opened 2021-04-08
Prevent spaces from causing issues in make_osx

alexrecuenco:
In some places in the script if some directory name had spaces it could cause issues. 

I also added quotes around a few other directory names that currently are hardcoded with names without spaces as a way to prevent any future mishap


SomberNight:
Looks good. Thanks.

Issue #6232, opened 2020-06-13
Considering (partial?) removal of openalias support. Comment here if you use it.


SomberNight:
With recent refactoring of invoices/payment requests, OpenAlias support is now partially broken.
Looking up an OpenAlias (name -> addr) works, however
- signing a payment request with an openalias is broken
  https://github.com/spesmilo/electrum/blob/eb39aa143b8112549a5f8156d8e3ee41034eac6a/electrum/wallet.py#L1772
- trying to pay a bip21 URI with embedded openalias name and signature is broken
  https://github.com/spesmilo/electrum/blob/eb39aa143b8112549a5f8156d8e3ee41034eac6a/electrum/util.py#L854
  https://github.com/spesmilo/electrum/blob/eb39aa143b8112549a5f8156d8e3ee41034eac6a/electrum/paymentrequest.py#L449

I am wondering whether this functionality is worth restoring/fixing.
We could remove the broken functionality;
or maybe we could strip OpenAlias support completely.

Do people actually use OpenAlias?
Do you use OpenAlias? What do you use it for?

If you want the project to keep supporting OpenAlias, please comment here.

fluffypony:
I know of a fair number of people that use OpenAlias within Electrum, and we're actually busy with an OpenAlias v2 spec. There's also some activity around OpenAlias with Lightning, so I'd definitely push for it to continue.

A bip21 URI with OA embedded could get confusing, as the OA record itself could include an amount, so I'd veer away from that particular edge-case - unless we start to see adoption by payment processors that dictates revisiting this.

relativisticelectron:
Without paynyms is this not a huge privacy leak and should not be used?

fluffypony:
@relativisticelectron yes, but that doesn't mean it's not useful. OpenAlias is a very broad standard, and works with Monero (for instance) where privacy isn't an issue. It already allows you to specify a Lightning invoice (via ```oa1:lightning```), which may be somewhat useless given the nature of DNS, but there will be long-lived LN payment mechanisms in future that OA will support.

ysangkok:
@fluffypony keysend is already supported in LND and c-lightning: https://lightning.readthedocs.io/lightning-keysend.7.html

fluffypony:
@ysangkok awesome that's great news!

JeremyRand:
[...]

PR #7192, opened 2021-04-08
dependencies: document again choice between pycryptodomex/cryptography

SomberNight:
Before upgrading `dnspython` to 2.0 in #6828, we have supported both
`pycryptodomex` and `cryptography` as crypto backends. We use `dnspython`
for DNSSEC, and also some other things. The DNSSEC part of `dnspython` is
an optional extra which requires `cryptography`. This is why we removed
the choice of `pycryptodomex` as a backend: we would need `cryptography`
for DNSSEC anyway.

Note that DNSSEC is only used for OpenAlias, which itself is probably
only used by a handful of people. We have considered removing OpenAlias
support as then we could also remove DNSSEC, see #6232.

Recently `cryptography` (since version 3.4) [started using Rust code](https://github.com/pyca/cryptography/blob/58f0ad5b6b928677931c7ad44deee839a29ee9d3/CHANGELOG.rst#34---2021-02-07), which
necessitates having a Rust compiler if building from source.
Some people complained this is problematic for them.

With this PR, we restore the choice between `pycryptodomex` and `cryptography`.
(Though the code that can utilise either backend has not even been removed,
it is in `crypto.py`)
We support having either dependency; with the caveat that DNSSEC is only
available if `cryptography` is available.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,8 +55,9 @@ libsecp256k1 yourself::
     sudo apt-get install automake libtool
     ./contrib/make_libsecp256k1.sh
 
-Due to the need for fast symmetric ciphers, `cryptography`_ is required.
-Install from your package manager (or from pip)::
+Due to the need for fast symmetric ciphers, either one of `pycryptodomex`_
+or `cryptography`_ is required. Install from your package manager
+(or from pip)::
 
     sudo apt-get install python3-cryptography
 
--- a/electrum/dnssec.py
+++ b/electrum/dnssec.py
@@ -53,9 +53,23 @@ import dns.rdtypes.IN.A
 import dns.rdtypes.IN.AAAA
 
 from .logging import get_logger
+from .util import versiontuple
 
 
 _logger = get_logger(__name__)
+
+IS_DNSSEC_AVAILABLE = False
+MIN_DNSPYTHON_VERSION = "2.0"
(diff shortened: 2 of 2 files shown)


SomberNight:
reminder to self; maybe change text here:
https://github.com/spesmilo/electrum/blob/a99e0d9e17c66229572786b19c92bdb049793262/contrib/requirements/requirements.txt#L11-L13

SomberNight:
> Note that DNSSEC is only used for OpenAlias, which itself is probably
only used by a handful of people. We have considered removing OpenAlias
support as then we could also remove DNSSEC, see https://github.com/spesmilo/electrum/issues/6232.

Annoyingly, [bip-0353](https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki) (if we wanted to support that) also requires DNSSEC.

ecdsa:
> > Note that DNSSEC is only used for OpenAlias, which itself is probably
> > only used by a handful of people. We have considered removing OpenAlias
> > support as then we could also remove DNSSEC, see #6232.
> 
> Annoyingly, [bip-0353](https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki) (if we wanted to support that) also requires DNSSEC.

yes, we want to keep dnssec for BIP353

ecdsa:
apparently dnspython has migrated from pycryptodome to cryptography in 2018.
before that they supported both. see https://github.com/rthalley/dnspython/pull/319

PR #7202, opened 2021-04-12
MPP splitting algorithm: redesign and split within channels

### LINKED ISSUE #7180: mpp splitting improvements
Currently, when we have a single channel and want to send a large payment (yet to be defined), we only try to send over the channel with the large amount until we have to give up. Instead, we could detect the large payment amount and split it. If the payment amount is intermediate and we can't succeed to send it, we should be able to send it in smaller chunks over the single channel.

When having more than one channel and multiple payment attempts already happened, we should increase the number of splits dynamically, which is why a feedback into the splitting algorithm is needed (via split penalty).

To solve these issues, I suggest to extend the split algorithm to also allow for sub-splittings within a channel. When we quantify an amount as large, we should already allow for splits from the get go. Then, one should track the attempts keyed by the sending amount and make the split penalty dependent on that number. When a new amount is encountered, we should reset the split penalty.

I would start to work on this now and be happy for feedback.

bitromortac:
Changes the splitting algorithm to an easier approach and allows for split amounts in individual channels.

Fixes #7180. I redesigned the splitting algorithm, which is in my view easier to understand and maintain. The main change is how the split configurations are created. Here, I'm using a more top-down probabilistic approach to fill up the channels. I think it doesn't really matter how the configurations are created as long as it samples the solution space relatively well. Rating of the solutions stays the same (except for adapting to individual amounts in a channel) and one can also see from the tests that the algo doesn't significantly change things.
I think this in combination with #7152 will significantly improve mpp sending. In the edge case of a single channel, it was previously not possible to send out multiple HTLCs. With splitting into smaller amounts this enables to probe the balances on the receiving end. Combined with #7152, we probably don't need to implement a feedback mechanism based on the number of failed payment attempts, as after splitting, the pathfinder automatically sorts out paths that use a lot of bandwidth (large chunks are more likely for small number of parts).

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -8,7 +8,7 @@ from decimal import Decimal
 import random
 import time
 from typing import (Optional, Sequence, Tuple, List, Set, Dict, TYPE_CHECKING,
-                    NamedTuple, Union, Mapping, Any, Iterable, AsyncGenerator)
+                    NamedTuple, Union, Mapping, Any, Iterable, AsyncGenerator, DefaultDict)
 import threading
 import socket
 import aiohttp
@@ -1148,7 +1148,7 @@ class LNWallet(LNWorker):
                 raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON, data=b'')
 
         self.logs[payment_hash.hex()] = log = []
-        trampoline_fee_level = self.INITIAL_TRAMPOLINE_FEE_LEVEL
+        trampoline_fee_levels = defaultdict(lambda: self.INITIAL_TRAMPOLINE_FEE_LEVEL)  # type: DefaultDict[bytes, int]
         use_two_trampolines = True # only used for pay to legacy
 
         amount_inflight = 0  # what we sent in htlcs (that receiver gets, without fees)
@@ -1168,7 +1168,7 @@ class LNWallet(LNWorker):
                     full_path=full_path,
                     payment_hash=payment_hash,
                     payment_secret=payment_secret,
-                    trampoline_fee_level=trampoline_fee_level,
+                    trampoline_fee_levels=trampoline_fee_levels,
                     use_two_trampolines=use_two_trampolines,
                     fwd_trampoline_onion=fwd_trampoline_onion
                 )
@@ -1211,11 +1211,12 @@ class LNWallet(LNWorker):
             # if we get a channel update, we might retry the same route and amount
             route = htlc_log.route
             sender_idx = htlc_log.sender_idx
+            erring_node_id = route[sender_idx].node_id
             failure_msg = htlc_log.failure_msg
             code, data = failure_msg.code, failure_msg.data
             self.logger.info(f"UPDATE_FAIL_HTLC. code={repr(code)}. "
                              f"decoded_data={failure_msg.decode_data()}. data={data.hex()!r}")
-            self.logger.info(f"error reported by {bh2u(route[sender_idx].node_id)}")
+            self.logger.info(f"error reported by {bh2u(erring_node_id)}")
             if code == OnionFailureCode.MPP_TIMEOUT:
                 raise PaymentFailure(failure_msg.code_name())
             # trampoline
@@ -1227,7 +1228,7 @@ class LNWallet(LNWorker):
                 if code in (OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT,
                             OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON):
                     # todo: parse the node parameters here (not returned by eclair yet)
-                    trampoline_fee_level += 1
+                    trampoline_fee_levels[erring_node_id] += 1
                     continue
                 elif use_two_trampolines:
                     use_two_trampolines = False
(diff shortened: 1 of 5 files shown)


electrum/lnworker.py
```diff
@@ -1563,18 +1564,19 @@ def create_routes_for_payment(
...
-                                    invoice_pubkey=invoice_pubkey,
-                                    min_cltv_expiry=min_cltv_expiry,
-                                    r_tags=r_tags,
-                                    invoice_features=invoice_features,
-                                    channels=[channel],
-                                    full_path=None)
-                                routes.append((route, part_amount_msat, final_total_msat, part_amount_msat, min_cltv_expiry, payment_secret, fwd_trampoline_onion))
+                        for (chan_id, _), part_amounts_msat in s[0].items():
+                            if part_amounts_msat:
+                                for part_amount_msat in part_amounts_msat:
```
ecdsa, line 1569: isn't the if statement redundant ?

bitromortac: It's because the returned dict is a defaultdict that can have empty lists, see:
https://github.com/spesmilo/electrum/blob/49470a34f63b2bd68256bb671471882ef3257fe1/electrum/tests/test_mpp_split.py#L29-L38
I will remove those empty entries.

bitromortac:
I'm still testing this and discovered a minor issue. Consider sending via a single channel and it does not succeed with a singlepart payment. We will then split into about two equal parts and send both HTLCs out, but it happens to be with very high probability over the same paths as fees do not distinguish the paths. With this we lose half of the probing attempts as both routes will fail at the same channel. I would suggest to first implement penalties for inflight HTLCs via liquidity hints, then we would avoid sending two splits over the same path. I'm currently experimenting with this.

bitromortac:
This needs more testing with #7292 merged.

bitromortac:
The splitting itself is probably ok by now, but I still need to think about the asynchronous payment looping over the different parts.

electrum/lnworker.py
bitromortac: Yes, it was redundant :)

bitromortac:
I made some further changes concerning `create_routes_for_payment`, see commit message.

The tests pass, but in `test_payment_multipart_trampoline` there is some weird behavior of the forwarding nodes (they don't seem to find routes), which is why I also included a test for single-part trampoline payments, which is flaky. This test is present here only for discussion purposes and should be removed (and reintroduced in another PR that fixes the forwarding behavior). In general, I think it is necessary to include more tests for trampoline (end-to-end as well as legacy payments).
I tested the current state also on testnet:
- [x] single-part payment (spp) (Electrum-Electrum)
- [x] multi-part payment (mpp) (Electrum-Electrum)
- [x] spp trampoline (Electrum-Electrum)
- [x] mpp trampoline (Electrum-Electrum)
- [x] mpp (Electrum-LND)
- [x] mpp trampoline (Electrum-LND)

The redesign of the splitting algorithm shouldn't be that much of a concern, as it behaves qualitatively the same as before (see the tests).

bitromortac:
The issues with trampoline forwarding seem to be connected with 3a7f5373ac859adffe4ab0ddae07b3d96fb33217. Am investigating and trying to provide a fix. Will also remove the single part trampoline test.

SomberNight:
> The issues with trampoline forwarding seem to be connected with 3a7f5373ac859adffe4ab0ddae07b3d96fb33217. Am investigating and trying to provide a fix. Will also remove the single part trampoline test.

Ah, right... that commit made the trampoline behaviour for tests flaky indeed.
See the last item in the commit msg:
> - also, previously if both the sender and the recipient used trampoline and
  they shared a trampoline forwarder (that they were both connected to), the
  private channels the recipient had (with nodes other than the shared TF)
  would never be attempted.

also note how I have increased attempts from 3 to 30 in https://github.com/spesmilo/electrum/commit/3a7f5373ac859adffe4ab0ddae07b3d96fb33217#diff-73d55b4269c7275dbc700a6d2e91fe31240c7748053517aa319e4f2f7d32c22eR950

The flakyness has to do with the square graph we use in the tests being unrealistic:
```
A -- B
|    |
C -- D
```
A is trying to pay D. B and C are trampoline forwarders.
The problem is that there is no path between B and C; however A, when constructing the route consisting of trampoline forwarders assumes that all TFs should be able to route to each other.
So if A chooses a path T_B->T_C->D; or T_C->T_B->D the attempt fails; while if A chooses T_B->D or T_C->D, it succeeds;
and A chooses this randomly.

bitromortac:
Thank you for the explanation, that makes sense. I removed the test.

SomberNight (approved):
Changes look good.
I also did some test payments manually with both local gossip and trampoline.

SomberNight:
I think we should try sending with MPP from the start (at least for the local gossip case, maybe not for trampoline (?)).
That is, if the invoice signals MPP, we should allow mpp_split.py to split the payment right away.

This could improve privacy somewhat, and potentially improve payment success a tiny bit for larger amounts.
I think it would also simplify the code in `create_routes_for_payment` -- I am hoping the huge try-except can be removed that way :P
https://github.com/spesmilo/electrum/blob/ce44a03c24a71edc3514090ce9e19322c6fb5746/electrum/lnworker.py#L1537

PR #7203, opened 2021-04-12
Allow sending to scripts from payto and paytomany RPCs

metalicjames:
This PR:

- Refactors `parse_script` and `parse_output` from the `gui` module into `bitcoin`
- Allows `payto` to accept scripts just like the GUI
- Allows `paytomany` to accept scripts like the GUI

The use-case of this PR is for creating OP_RETURN outputs from the command line.

Before this change:
```
electrum --testnet payto -F tb1q2s7kk7qp0urvz46wpjg2jgp7vfxhj7gy63p2sh -c tb1q2s7kk7qp0urvz46wpjg2jgp7vfxhj7gy63p2sh "OP_RETURN 94cab8cbeb53d63dba67e015450eb83c0682b5aa2d232fbb06ddf260d64a65b1" 0
invalid bitcoin address: OP_RETURN 94cab8cbeb53d63dba67e015450eb83c0682b5aa2d232fbb06ddf260d64a65b1
``` 

After:
```
electrum --testnet payto -F tb1q2s7kk7qp0urvz46wpjg2jgp7vfxhj7gy63p2sh -c tb1q2s7kk7qp0urvz46wpjg2jgp7vfxhj7gy63p2sh "OP_RETURN 94cab8cbeb53d63dba67e015450eb83c0682b5aa2d232fbb06ddf260d64a65b1" 0
02000000000101d7e6deda8ad91c1fe7ab3d0cff3d2d95b0e9efc3b774a74d995ff69736f547f80100000000fdffffff020000000000000000226a2094cab8cbeb53d63dba67e015450eb83c0682b5aa2d232fbb06ddf260d64a65b18641100000000000160014543d6b78017f06c1574e0c90a9203e624d79790402473044022002d20f5298dfa29fd106903c62b93138dcada10745547d45edbcd6922f52cea602204f69dd8bba9908821fa8004fd1c758122aada6643b1753527e6eed89e0c695460121022dab9a6611baa4d42d2ec0cc8741c1f369f8265b1eeaedcbd971533885d9eebf28141e00
```

Here's the TX: https://blockstream.info/testnet/tx/c6c4ba712ceb1f948dc66067228529ae1d21733ffccba681927c8fb8fda61d55

### DIFF
--- a/electrum/bitcoin.py
+++ b/electrum/bitcoin.py
@@ -752,6 +752,26 @@ def is_private_key(key: str, *, raise_on_error=False) -> bool:
             raise
         return False
 
+def parse_script(x):
+    script = ''
+    for word in x.split():
+        if word[0:3] == 'OP_':
+            opcode_int = opcodes[word]
+            script += construct_script([opcode_int])
+        else:
+            bfh(word)  # to test it is hex data
+            script += construct_script([word])
+    return script
+
+def parse_output(x) -> bytes:
+    if is_address(x):
+        return bfh(address_to_script(address))
+    try:
+        script = parse_script(x)
+        return bfh(script)
+    except Exception:
+        pass
+    raise Exception("Invalid address or script.")
 
 ########### end pywallet functions #######################
 
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -622,7 +622,7 @@ class Commands:
         change_addr = self._resolver(change_addr, wallet)
         domain_addr = None if domain_addr is None else map(self._resolver, domain_addr, repeat(wallet))
         amount_sat = satoshis_or_max(amount)
-        outputs = [PartialTxOutput.from_address_and_value(destination, amount_sat)]
+        outputs = [PartialTxOutput(scriptpubkey=bitcoin.parse_output(destination), value=amount_sat)]
         tx = wallet.create_transaction(
             outputs,
             fee=tx_fee,
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -653,7 +653,7 @@ class Commands:
         for address, amount in outputs:
             address = self._resolver(address, wallet)
             amount_sat = satoshis_or_max(amount)
-            final_outputs.append(PartialTxOutput.from_address_and_value(address, amount_sat))
+            final_outputs.append(PartialTxOutput(scriptpubkey=bitcoin.parse_output(address), value=amount_sat))
         tx = wallet.create_transaction(
             final_outputs,
             fee=tx_fee,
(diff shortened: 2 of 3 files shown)


electrum/bitcoin.py
```diff
@@ -752,6 +753,36 @@ def is_private_key(key: str, *, raise_on_error=False) -> bool:
...
+            opcode_int = opcodes[word]
+            script += construct_script([opcode_int])
+        else:
+            bfh(word)  # to test it is hex data
+            script += construct_script([word])
+    return script
+
+def parse_address(line):
+    r = line.strip()
+    m = re.match('^'+RE_ALIAS+'$', r)
```
SomberNight, line 769: `RE_ALIAS` is not defined here.
(but see other comment first: don't move `parse_address` to bitcoin.py at all)

electrum/bitcoin.py
```diff
@@ -752,6 +753,36 @@ def is_private_key(key: str, *, raise_on_error=False) -> bool:
...
+    address = str(m.group(2) if m else r)
+    assert is_address(address)
+    return address
+
+def parse_output(x) -> bytes:
+    try:
+        address = parse_address(x)
+        return bfh(address_to_script(address))
+    except Exception:
+        pass
```
SomberNight, line 779: I would strongly prefer not exposing the openalias regex stuff to the CLI, and would keep that in the Qt GUI.
First, I suggest doing this here instead:

```suggestion
    if is_address(x):
        return return bfh(address_to_script(address))
```

Second, let's keep `parse_address` in the GUI code where it was.
Then, to maintain the current behaviour for the GUI code, also keep a method `parse_output` in the GUI, which first calls `parse_address` and then falls back to this methiod here (`bitcoin.parse_output`).

SomberNight (changes requested):
I think we can expose the script-parsing to the payto CLI command, but please do not expose the openalias-parsing.

electrum/bitcoin.py
```diff
@@ -752,6 +752,26 @@ def is_private_key(key: str, *, raise_on_error=False) -> bool:
...
+            opcode_int = opcodes[word]
+            script += construct_script([opcode_int])
+        else:
+            bfh(word)  # to test it is hex data
+            script += construct_script([word])
+    return script
+
+def parse_output(x) -> bytes:
+    if is_address(x):
+        return bfh(address_to_script(address))
```
SomberNight, line 768: `address` is not defined.

SomberNight (changes requested):
I've added a unit test in https://github.com/spesmilo/electrum/commit/af44c846cc53d31c32a25826af596340053448e7 that tests using the `payto` command to pay a bitcoin address. This PR would currently make that test fail. Please fix that, otherwise it looks ok.
It would also be nice if you could add a similar test that pays to an OP_RETURN output or some other script that uses the new functionality.

accumulator:
stale

PR #7231, opened 2021-04-22
swaps: fix swap sanity check

bitromortac:
In #7186  it is reported that the inverse `send_amount` is not the `send_amount` that we gave.
I tried to run through some explicit sample values and I see that
`recv_amount == SwapManager._get_recv_amount(SwapManager._get_send_amount(recv_amount))`
but
`send_amount != SwapManager._get_send_amount_get(SwapManager._recv_amount(send_amount))`
for some `send_amount` values, see the test cases. The probability of the occurrence of these errors is the same as the swap fee percentage, but it shows up when the swap slider in kivy is moved many times or continuously.

Both methods internally use `math.floor` and `math.ceil` which are non-invertible, which probably leads to these off-by-one errors. I suggest to relax the cross check regarding off-by-ones. I don't think those methods can be made exactly the inverse of each other if we follow the specification of the bolt backend. I also renamed the variable names of `inverted_XXX`, which is now more intuitive to me. I would remove the tests once you agree with the change.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -519,10 +519,11 @@ class SwapManager(Logger):
         recv_amount = self._get_recv_amount(send_amount, is_reverse=is_reverse)
         # sanity check calculation can be inverted
         if recv_amount is not None:
-            inverted_recv_amount = self._get_send_amount(recv_amount, is_reverse=is_reverse)
-            if send_amount != inverted_recv_amount:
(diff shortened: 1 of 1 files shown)


SomberNight:
Ok. Looks good to me.

bitromortac:
Ok thanks, I removed the tests and added more documentation.

SomberNight:
Thanks.

PR #7247, opened 2021-04-26
lnworker: fix path_finder access

### LINKED ISSUE #7242: AttributeError("'Network' object has no attribute 'path_finder'")
Hello! I'm playing with testnet lightning payments and have some troubles. When I pay invoice, I've got an error AttributeError("'Network' object has no attribute 'path_finder'"). On remote node invoice is settled, but in electrum its stuck on Status "In progress"

What I do:
1. Open channel with trampoline "endurance" 
2. Receive some LN payments 
3. Make LN payment 

[electrum_log_20210424T163303Z_197398.log](https://github.com/spesmilo/electrum/files/6370306/electrum_log_20210424T163303Z_197398.log)

bitromortac:
Fixes #7242. I went through all occurrences of `network.path_finder` and added checks for `path_finder` initialization, in scopes that do not apply the `if not self.channeldb` check previously. Right now I'm unsure whether we should check explicitly for `network.path_finder` or to use `lnworker.channel_db` as a check, i.e. check on the per wallet or per network basis.

I also think we could make the code more readable by converting `if not self.channeldb` to something more explicit, `network.uses_gossip`/`network.uses_trampoline`, perhaps.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1092,11 +1092,13 @@ class Commands:
 
     @command('n')
     async def clear_ln_blacklist(self):
-        self.network.path_finder.liquidity_hints.clear_blacklist()
+        if self.network.path_finder:
+            self.network.path_finder.liquidity_hints.clear_blacklist()
 
     @command('n')
     async def reset_liquidity_hints(self):
-        self.network.path_finder.liquidity_hints.reset_liquidity_hints()
+        if self.network.path_finder:
+            self.network.path_finder.liquidity_hints.reset_liquidity_hints()
 
     @command('w')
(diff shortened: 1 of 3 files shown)


SomberNight:
> Right now I'm unsure whether we should check explicitly for network.path_finder or to use lnworker.channel_db as a check, i.e. check on the per wallet or per network basis.

Note that lnworker.channel_db is a property which just proxies network.channel_db.
The channel_db / path_finder objects are ~singletons, managed by the network.

SomberNight:
> I also think we could make the code more readable by converting if not self.channeldb to something more explicit, network.uses_gossip/network.uses_trampoline, perhaps.

Good point, we could definitely try to improve this.

PR #7254, opened 2021-04-29
plugins README - fewer double negatives

mbarkhau:


gasull:
NB: English isn't my first language.

Shouldn't it say "if we feel that users who rely on it would not be threatened if we remove it", without the "any"?

With the "any" it sounds to me as if as long as one user doesn't feel threatened, then it's fine to remove the plugin.

PR #7267, opened 2021-05-01
add signet support

wakiyamap:
From https://github.com/spesmilo/electrumx/pull/122
ref? https://github.com/spesmilo/electrum/issues/7162

lseed.wakiyamap.dev
・9735 ports open for IPv4
・9736 ports open for IPv6
So far, I've only captured my node(signet-electrumx.wakiyamap.dev).
https://signet-lightning.wakiyamap.dev/

If you don't trust me, I or you will delete it.

PR #7282, opened 2021-05-06
add signet support

wakiyamap:
rebase https://github.com/spesmilo/electrum/pull/7267

Please let me know if you have any problems.

### DIFF
--- a/electrum/lnaddr.py
+++ b/electrum/lnaddr.py
@@ -178,6 +178,9 @@ def pull_tagged(stream):
     return (CHARSET[tag], stream.read(length * 5), stream)
 
 def lnencode(addr: 'LnAddr', privkey) -> str:
+    # see https://github.com/lightningnetwork/lightning-rfc/pull/844
+    if constants.net.NET_NAME == "signet":
+        addr.currency += "s"
     if addr.amount:
         amount = addr.currency + shorten_amount(addr.amount)
     else:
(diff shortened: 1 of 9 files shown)


electrum/lnaddr.py
```diff
@@ -178,6 +178,9 @@ def pull_tagged(stream):
     return (CHARSET[tag], stream.read(length * 5), stream)
 
 def lnencode(addr: 'LnAddr', privkey) -> str:
+    # see https://github.com/lightningnetwork/lightning-rfc/pull/844
+    if constants.net.NET_NAME == "signet":
+        addr.currency += "s"
```
SomberNight, line 183: I don't like this. I imagine if you call `lnencode` multiple times with the same LnAddr, it only works on the first call? As it mutates the LnAddr. I haven't tested though.
The root problem here is that `net.SEGWIT_HRP` is used for both the bech32 address prefix, and the bolt11 invoice prefix; and this duplication is not true anymore for signet. Let us simply have two constants in `net` then:
`net.SEGWIT_HRP` and `net.BOLT11_HRP`.
For most networks, they would be equal, but for signet they would be different.

wakiyamap: As far as I've tested it was fine, but it's certainly true. I fixed it.

johnzweng:
Is there any update on Signet support? 

Now that we have Taproot locked in for activation in November, I'd like starting to play around with it also in Electrum. 🙂  
(Taproot is already enabled on Signet, but not on Testnet yet).

SomberNight:
Thanks for the pull request.
I've pushed a follow-up commit (https://github.com/spesmilo/electrum/commit/57e52da77fd25300f6035a87c330086caf1caffc), which instead removes the `currency` field altogether.

carnhofdaki:
Wonderful! Thanks for adding Signet! Works well.

PR #7292, opened 2021-05-11
lnrouter: add inflight htlcs to liquidity hints

bitromortac:
When making a multipart payment, the payment in principle can have overlapping paths for individual parts.
The [spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#basic-multi-part-payments) states that we `SHOULD try to use diverse paths to the recipient for each HTLC`.

Here, the inflight HTLCs are tracked in the liquidity hints system, just by counting the active HTLCs per channel. The penalty for channels gets additional penalties proportional to the number of HTLCs in the channel.

I needed to make `create_routes_for_payment` an async generator, such that inflight HTLCs can be asynchronously added and removed.

This PR also solves an issue with #7202, where when sending an MPP over our single channel with equal split sizes, we avoid probing the same route.

The downside of only tracking the number of HTLCs is that we don't use the full capacity of a routing channel anymore as we don't track the amounts.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1196,12 +1196,14 @@ class LNWallet(LNWorker):
                 raise Exception(f"amount_inflight={amount_inflight} < 0")
             log.append(htlc_log)
             if htlc_log.success:
-                # TODO: report every route to liquidity hints for mpp
-                # even in the case of success, we report channels of the
-                # route as being able to send the same amount in the future,
-                # as we assume to not know the capacity
                 if self.network.path_finder:
+                    # TODO: report every route to liquidity hints for mpp
+                    # in the case of success, we report channels of the
+                    # route as being able to send the same amount in the future,
+                    # as we assume to not know the capacity
                     self.network.path_finder.update_liquidity_hints(htlc_log.route, htlc_log.amount_msat)
+                    # remove inflight htlcs from liquidity hints
+                    self.network.path_finder.update_htlcs_liquidity_hints(htlc_log.route, add_htlcs=False)
                 return
             # htlc failed
             if len(log) >= attempts:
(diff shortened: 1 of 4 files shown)


SomberNight:
> The downside of only tracking the number of HTLCs is that we don't use the full capacity of a routing channel anymore as we don't track the amounts.

What is the rationale for tracking the number of pending HTLCs instead of the value sum (locked up capacity)?
Is it simplicity? (I guess due to the forwarding fees, it is not straightforward to add/deduct the htlc values from each channel)

electrum/lnworker.py
```diff
@@ -1347,6 +1337,8 @@ def handle_error_code_from_failed_htlc(
         # for errors that do not include a channel update
         else:
             self.network.path_finder.liquidity_hints.add_to_blacklist(fallback_channel)
+        # remove inflight htlcs from liquidity hints
+        self.network.path_finder.update_htlcs_liquidity_hints(route, add_htlcs=False)
```
SomberNight, line 1341: I think this needs to be done earlier in the method as otherwise we might have already raised before getting here.

SomberNight:
I've fixed some nits in a follow-up commit.

bitromortac:
> What is the rationale for tracking the number of pending HTLCs instead of the value sum (locked up capacity)?
> Is it simplicity? (I guess due to the forwarding fees, it is not straightforward to add/deduct the htlc values from each channel)

Yes, I wanted to keep it simple. Also, we don't know channel capacities. Is it actually possible to lock up more capital than capacity given the HTLCs have different directions and cancel each other? The way it is now, we have less diversity in possible paths.

> I've fixed some nits in a follow-up commit.

Thank you!

SomberNight:
> Is it actually possible to lock up more capital than capacity given the HTLCs have different directions and cancel each other? 

No, pending HTLCs in opposite directions do not cancel each other. Just consider the commitment transaction, and for simplicity ignore the dust limit and trimming: all HTLCs show up as tx outputs and are allocated some part of the total capacity. They do not cancel each other until they are fulfilled.

bitromortac:
Yes, of course, sorry for my lazy thinking. Was a bit confused from the accounting point of view.

PR #7294, opened 2021-05-13
Not connecting to network

jjdewit:
Hello,

The electrumsys`wallet is not connecting to the network.
Is it possible that you update the wallet please? I use it with a Trezor hardware key and there is no solution to send them to another wallet. Nor can I export the seed from the Trezor to import to a QT wallet.

Awaiting your response.


SomberNight:
Altcoins are off-topic.
Also this should not be a pull request...

PR #7296, opened 2021-05-14
commands load_wallet: add set current for commands option

zebra-lucky:
Add `--set_current` command option for load_wallet command,
which allow to run commands where wallet required without additional `wallet_path` option.


SomberNight:
If I understand correctly, the use case is not having to pass the `--wallet some_path` argument to every CLI command; purely for convenience reasons?
If so, then I am not sure the additional complexity is worth it.

ecdsa:
concept ack

ecdsa:
replaced by  #9885

PR #7300, opened 2021-05-16
Android: Fix wallet backup 

thohemp:
This change fixes the storage access behavior for the wallet backup for android. Ref: https://github.com/spesmilo/electrum/issues/6774

The permission for accessing storage has to be granted in runtime. This is already implemented, but in the wrong order when it comes to creating the back up dir.

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -1335,7 +1335,6 @@ class ElectrumWindow(App, Logger):
                 self.show_error(_("Backup NOT saved. Backup directory not configured."))
             return
 
-        backup_dir = util.android_backup_dir()
         from android.permissions import request_permissions, Permission
         def cb(permissions, grant_results: Sequence[bool]):
             if not grant_results or not grant_results[0]:
@@ -1343,6 +1342,7 @@ class ElectrumWindow(App, Logger):
                 return
             # note: Clock.schedule_once is a hack so that we get called on a non-daemon thread
             #       (needed for WalletDB.write)
+            backup_dir = util.android_backup_dir()
             Clock.schedule_once(lambda dt: self._save_backup(backup_dir))
         request_permissions([Permission.WRITE_EXTERNAL_STORAGE], cb)
 


SomberNight:
This does not seem to work for me, on Android 11.
The backup file is not getting saved to disk.
I am still getting the same exception, except in a different thread now. The only difference is that the crash reporter no longer appears.

```
06-07 13:15:19.593 10424 10424 I python  :  Traceback (most recent call last):
06-07 13:15:19.593 10424 10424 I python  :    File "jnius/jnius_proxy.pxi", line 50, in jnius.jnius.PythonJavaClass.invoke
06-07 13:15:19.593 10424 10424 I python  :    File "jnius/jnius_proxy.pxi", line 76, in jnius.jnius.PythonJavaClass._invoke
06-07 13:15:19.593 10424 10424 I python  :    File "/home/user/wspace/electrum/.buildozer/android/platform/build-arm64-v8a/build/python-installs/Electrum/android/permissions.py", line 447, in onRequestPermissionsResult
06-07 13:15:19.593 10424 10424 I python  :    File "/home/user/wspace/electrum/.buildozer/android/platform/build-arm64-v8a/build/python-installs/Electrum/android/permissions.py", line 530, in python_callback
06-07 13:15:19.593 10424 10424 I python  :    File "/home/user/wspace/electrum/.buildozer/android/app/electrum/gui/kivy/main_window.py", line 1345, in cb
06-07 13:15:19.594 10424 10424 I python  :    File "/home/user/wspace/electrum/.buildozer/android/app/electrum/util.py", line 422, in android_backup_dir
06-07 13:15:19.594 10424 10424 I python  :  PermissionError: [Errno 1] Operation not permitted: '/storage/emulated/0/org.electrum.electrum'

```
Have you tested this? is it different for you?

SomberNight:
I believe it is failing due to ["Scoped Storage"](https://developer.android.com/training/data-storage#scoped-storage) introduced in Android 10, which has locked down access to external storage.

see https://developer.android.com/training/data-storage/app-specific
note specifically:
> The files stored in these directories are removed when your app is uninstalled.

So I guess we should use ["shared storage"](https://developer.android.com/training/data-storage/shared).

Looks like we could "temporarily opt-out" of Scoped Storage, and then the issue might go away for Android 10; but the opt-out is ignored on Android 11.
I think the problem is that we are accessing the file system directly (external storage), and Android essentially no longer allows that...
maybe we need to use e.g. https://developer.android.com/reference/android/content/Intent#ACTION_CREATE_DOCUMENT

thohemp:
I can confirm it is working for Android 9, but I did not test it on a version above that. I will take a look at the new introduced regulations.

thohemp:
That's right.

SomberNight:
Ok, so in your case the issue is simply that `util.android_backup_dir()` is creating the directory as a side-effect before we have the permission; which I guess you have already realised :P unfortunately apparently there is another issue for newer Android.
Still, I will merge this then.

SomberNight:
Thanks

Issue #7253, opened 2021-04-28
feature: Lightning payment requests via BIP21/lightning prefix


bitromortac:
We could support paying to lightning invoices via BIP21 URIs using the `&lightning=` key or by using the `lightning:` prefix, see
https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#encoding-overview

SomberNight:
I think we already support the `lightning:` prefix for bolt11 invoices. If it does not work then that's a bug.
https://github.com/spesmilo/electrum/blob/1dfa81e7b6faf2e33dac253a5520a8db0d4c4a9b/electrum/util.py#L985-L992

> support paying to lightning invoices via BIP21 URIs using the `&lightning=` key 

What exactly does the BOLT suggest there?
Is it about a practically empty bip21 URI that only has this single key, as in
`bitcoin:?lightning=lnbc1asdasd`
? (as in no bitcoin address given - AFAIR this is a valid bip21 URI)

If it allows mixing a bitcoin address and a lightning invoice, and the bip21 URI contains both, how should that be handled?

bitromortac:
> I think we already support the `lightning:` prefix for bolt11 invoices. If it does not work then that's a bug.

I tested `./run_electrum bitcoin:bip21url`, which works and `./run_electrum lightning:lninvoice` which errors with `unknown_command`.

> https://github.com/spesmilo/electrum/blob/1dfa81e7b6faf2e33dac253a5520a8db0d4c4a9b/electrum/util.py#L985-L992
> 
> > support paying to lightning invoices via BIP21 URIs using the `&lightning=` key
> 
> What exactly does the BOLT suggest there?
> Is it about a practically empty bip21 URI that only has this single key, as in
> `bitcoin:?lightning=lnbc1asdasd`
> ? (as in no bitcoin address given - AFAIR this is a valid bip21 URI)
> 
> If it allows mixing a bitcoin address and a lightning invoice, and the bip21 URI contains both, how should that be handled?
> 
> Also, do you know of any project/service that creates bip21 URIs with a lightning key?

Exactly, we could give precedence for Lighting invoices as done here: https://github.com/LN-Zap/zap-desktop/pull/3131

SomberNight:
> I tested `./run_electrum bitcoin:bip21url`, which works and `./run_electrum lightning:lninvoice` which errors with `unknown_command`.

Ah, right, for that we need to change:
[...]

PR #7301, opened 2021-05-17
qt+android: add lightning URI support

bitromortac:
Adds `lightning:` URI support for Android and Qt. I hope I found all the places where the mimetype is added to the system, but unsure about macOS. Related: #7253.

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -79,7 +79,7 @@ from electrum.exchange_rate import FxThread
 from electrum.simple_config import SimpleConfig
 from electrum.logging import Logger
 from electrum.lnutil import ln_dummy_address, extract_nodeid, ConnStringFormatError
-from electrum.lnaddr import lndecode, LnDecodeException
+from electrum.lnaddr import lndecode, LnInvoiceException
 
 from .exception_window import Exception_Hook
 from .amountedit import AmountEdit, BTCAmountEdit, FreezableLineEdit, FeerateEdit, SizedFreezableLineEdit
@@ -1962,12 +1962,15 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
         else:
             self.payment_request_error_signal.emit()
 
-    def parse_lightning_invoice(self, invoice):
+    def set_ln_invoice(self, invoice: str):
         """Parse ln invoice, and prepare the send tab for it."""
         try:
             lnaddr = lndecode(invoice)
-        except Exception as e:
-            raise LnDecodeException(e) from e
+        except LnInvoiceException as e:
+            self.show_error(_("Error parsing Lightning invoice") + f":\n{e}")
+            return
+
+        self.payto_e.lightning_invoice = invoice
         pubkey = bh2u(lnaddr.pubkey.serialize())
         for k,v in lnaddr.tags:
             if k == 'd':
@@ -1980,22 +1983,18 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
         self.message_e.setText(description)
         if lnaddr.get_amount_sat() is not None:
             self.amount_e.setAmount(lnaddr.get_amount_sat())
-        #self.amount_e.textEdited.emit("")
         self.set_onchain(False)
 
     def set_onchain(self, b):
         self._is_onchain = b
         self.max_button.setEnabled(b)
 
-    def pay_to_URI(self, URI):
-        if not URI:
-            return
+    def set_bip21(self, text: str):
         try:
-            out = util.parse_URI(URI, self.on_pr)
+            out = util.parse_URI(text, self.on_pr)
         except InvalidBitcoinURI as e:
             self.show_error(_("Error parsing URI") + f":\n{e}")
             return
-        self.show_send_tab()
         self.payto_URI = out
         r = out.get('r')
         sig = out.get('sig')
@@ -2016,8 +2015,19 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
             self.message_e.setText(message)
         if amount:
             self.amount_e.setAmount(amount)
-            self.amount_e.textEdited.emit("")
 
+    def pay_to_URI(self, text: str):
+        if not text:
+            return
+        # first interpret as lightning invoice
+        bolt11_invoice = maybe_extract_bolt11_invoice(text)
+        if bolt11_invoice:
+            self.set_ln_invoice(bolt11_invoice)
+        else:
+            self.set_bip21(text)
+        # update fiat amount
+        self.amount_e.textEdited.emit("")
+        self.show_send_tab()
 
     def do_clear(self):
         self.max_button.setChecked(False)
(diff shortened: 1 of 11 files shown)


electrum/gui/qt/main_window.py
```diff
@@ -1964,12 +1964,15 @@ def on_pr(self, request: 'paymentrequest.PaymentRequest'):
...
-    def parse_lightning_invoice(self, invoice):
+    def set_ln_invoice(self, invoice: LNInvoice):
         """Parse ln invoice, and prepare the send tab for it."""
         try:
             lnaddr = lndecode(invoice, expected_hrp=constants.net.SEGWIT_HRP)
-        except Exception as e:
-            raise LnDecodeException(e) from e
+        except (LnDecodeException, LnAddressError) as e:
+            self.show_error(_("Error parsing Lightning invoice") + f":\n{e}")
+            return
```
SomberNight, line 1973: This looks suspect.
Note that `lndecode` has a bit unfortunate interface: it can raise any Exception. This is why the old code is catching Exception and raising instead a more specific exception (LnDecodeException) so that callers can distinguish what happened.
`lndecode` itself is not raising `LnDecodeException`. We could change that potentially, but that is simply not how it works atm.
My point is, with your suggested change, this block would "leak" e.g. ValueErrors and so on.

bitromortac:
I pushed a fix for very large invoice amount conversions (not saying this will eventually happen :)).
I added specific `LnInvoiceExceptions` (not sure about the name) to the `lnaddr` module. I would like to not propagate the expected errors to the crash reporter, but to the user, as it's bad user input and the bad request is not Electrum's fault. Other unexpected errors should be propagated to the crash reporter.

electrum/gui/qt/main_window.py
```diff
@@ -1967,12 +1967,17 @@ def on_pr(self, request: 'paymentrequest.PaymentRequest'):
...
+    def set_ln_invoice(self, invoice: str):
         """Parse ln invoice, and prepare the send tab for it."""
         try:
             lnaddr = lndecode(invoice)
+        except LnInvoiceException as e:
+            self.show_error(_("Error parsing Lightning invoice") + f":\n{e}")
+            return
         except Exception as e:
-            raise LnDecodeException(e) from e
+            raise LnInvoiceException(e) from e
```
SomberNight, line 1978: Is the intention here to propagate this exception to the crash reporter?
Note that even without catching and re-raising it with a different type, i.e. without these two lines, this exception would get to the crash reporter anyway.

So what is the intention of these two lines; specifically re re-raising the original exception? One side-effect it will have is that the crash reporter might group the reports together (irrespective of the original exception) as it only looks at the last exception. ( Reports are grouped by `id`, see: https://github.com/spesmilo/electrum/blob/d0f0669e8f7519018b6c79958184dabf467cf4ff/electrum/base_crash_reporter.py#L85-L89 )

bitromortac: Ah, ok then this doesn't make so much sense. I just repeated the pattern that was existent before. Removed the re-raising.

electrum/gui/qt/main_window.py
```diff
@@ -1967,12 +1967,15 @@ def on_pr(self, request: 'paymentrequest.PaymentRequest'):
...
         """Parse ln invoice, and prepare the send tab for it."""
         try:
             lnaddr = lndecode(invoice)
-        except Exception as e:
-            raise LnDecodeException(e) from e
+        except LnInvoiceException as e:
+            self.show_error(_("Error parsing Lightning invoice") + f":\n{e}")
+            return
+
+        self.payto_e.lightning_invoice = invoice
```
SomberNight, line 1978: This line is also done by the caller in paytoedit.py, around line 174. One of them is redundant I guess? Is this one needed?

bitromortac: Oh right, don't recall why I added it here. Where do you think it is cleaner to add? I would rather expect it in `set_ln_invoice` in analogy to the android `set_ln_invoice`.

SomberNight: Well I suppose if `lndecode` raises, `payto_e.lightning_invoice` should not be set.
That's what the old code is doing at least:
from paytoedit, `parse_lightning_invoice()` is called, and only if it does not raise, `payto_e.lightning_invoice` is set.

The PR is changing the API of `set_ln_invoice`: previously if all was fine it returned None, and if there was a "typical" error, it raised an exception. Now, it returns None in either case.
Look at the caller in paytoedit.py: that code should also be changed probably: it is still expecting the exception for typical errors; and right now with the PR that code will set `payto_e.lightning_invoice` even if it raised (as the caller won't get the exception).

So maybe around L169 in paytoedit.py, this
```
                try:
                    self.win.set_ln_invoice(bolt11_invoice)
                except LnDecodeException as e:
                    self.errors.append(PayToLineError(line_content=data, exc=e))
                else:
                    self.lightning_invoice = bolt11_invoice
```
could be replaced with
```
                self.win.set_ln_invoice(bolt11_invoice)
```

bitromortac: Yes, totally makes sense, thank you.

SomberNight:
> I hope I found all the places where the mimetype is added to the system, but unsure about macOS

I think this is where the Android app registers its hooks:
https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/contrib/android/buildozer.spec#L143
https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/contrib/android/bitcoin_intent.xml

and macOS is here:
https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/contrib/osx/make_osx#L170

bitromortac:
Thanks for the hints. Hope that's the correct syntax (I tried to look up other examples). I'm currently testing android, but will not be able to test OSX.

SomberNight:
I've tested macOS at commit https://github.com/spesmilo/electrum/pull/7301/commits/b20b9d2c869caf0ce701c46a781a0d6301a6ec3a
Looks ok; it worked.

(had to build a dmg, copy app into /Applications/, and then clicked a lightning URI on htlc.me (for testnet) --- also modified run_electrum a bit so that it launches in testnet mode by default)

bitromortac:
Thank you for the test!

> Edit: on android `lightning:` is registered, but something still goes wrong (`bech32_checksum` fails). Will investigate.

This is fixed now with e9a1bd1f1e64e375ecfa3bdd438696ad1f52480b, input was not correctly sanitized. I think it should be good to go now.

SomberNight:
Thanks for the PR.

Note for future reference: I was pondering if I should squash this or maybe merge manually,
I think it would have been best to have two commits: one for the invoice amount fix, another for the URI support; potentially even as separate PRs. I did not want to invalidate commit sigs though.
So anyway, my point is that it is best to squash fixup commits within branches.

samthelion:
So if i I'm using the AppImage version how do I receive a `lightning:LNURLLDJGKLDNFGIRNGDFK...` payment?

PR #7308, opened 2021-05-19
swaps: fix satoshi conversion bug

### LINKED ISSUE #7305: Reverse swap from command line doesn't work (compares wrong amount)
I tried to do a reverse_swap from the command line. In the UI it works fine, but the command line gives the following problem.

Here is what I did. First, I did a dry run to get the expected receive amount:

```
# electrum reverse_swap 0.001 dryrun
{
    "lightning_amount": "0.001",
    "onchain_amount": "0.00081123",
    "success": null
}
```

So I want to send 0.001 (1 mBTC), and I'm getting back that I can expect 0.00081123 (0.8mBTC). Seems reasonable so far. However, when I actually initiate the swap:

```
# electrum reverse_swap 0.001 0.00081123
rswap check failed: onchain_amount is less than what we expected: 89555 < 843200081123
```

I'm getting the error that the swap server doesn't give me enough. It wants to give 89555 satoshi, which is the correct amount, but Electrum expects 843200811023 satoshi (8432 BTC). That's almost as much as a 10000 BTC pizza, and while I would be very happy with that swap, I don't expect it. :) I also tried with expecting 0 back, and got a similar error:

```
# electrum reverse_swap 0.001 0
rswap check failed: onchain_amount is less than what we expected: 89555 < 843200000000
```

I also think I located the error in `electrum/commands.py`. In the function `reverse_swap` here: https://github.com/spesmilo/electrum/blob/master/electrum/commands.py#L1202. The line reads:

```
onchain_amount_sat = satoshis(onchain_amount + claim_fee)
```

I thnk the claim_fee is already in satoshi, so the line should be:

```
onchain_amount_sat = satoshis(onchain_amount) + claim_fee
```

This code is specific to the command line, which would be consistent with the symptom (I.e. it works in the UI, but not command line). If this is indeed the problem, I would be happy to test it and submit a PR. But since the fix is so small and since I'm relatively new to the code base, perhaps someone else would prefer to make the fix? Please let me know!

bitromortac:
Fixes #7305.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1199,7 +1199,7 @@ class Commands:
         else:
             lightning_amount_sat = satoshis(lightning_amount)
             claim_fee = sm.get_claim_fee()
-            onchain_amount_sat = satoshis(onchain_amount + claim_fee)
+            onchain_amount_sat = satoshis(onchain_amount) + claim_fee
             success = await wallet.lnworker.swap_manager.reverse_swap(
                 lightning_amount_sat=lightning_amount_sat,
                 expected_onchain_amount_sat=onchain_amount_sat,


SomberNight:
Looks like that was me... thanks for fixing it :)

PR #7323, opened 2021-05-29
transaction: cache address determination from output script

theStack:
In order to avoid repeatedly calling `get_addr_from_output_script()` on every read of the `TxOutput.address` property, determine and cache it only whenever the output script is created/changed.

### DIFF
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -134,8 +134,17 @@ class TxOutput:
         raise Exception(f"unexptected legacy address type: {_type}")
 
     @property
+    def scriptpubkey(self) -> bytes:
+        return self._scriptpubkey
+
+    @scriptpubkey.setter
+    def scriptpubkey(self, scriptpubkey: bytes):
+        self._scriptpubkey = scriptpubkey
+        self._address = get_address_from_output_script(scriptpubkey)
+
+    @property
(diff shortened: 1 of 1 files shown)


SomberNight:
I am just curious:
Have you noticed/measured a performance difference in some use case, or did you simply notice the "TODO" and decided to fix this?

theStack:
> I am just curious:
> Have you noticed/measured a performance difference in some use case, or did you simply notice the "TODO" and decided to fix this?

No, I didn't benchmark anything, it was really just the TODO that caught my attention. I'm quite new to the codebase, but happy to investigate further to find out whether the change is worth it. Suggestions welcome :)

SomberNight:
Cherry-picked manually in https://github.com/spesmilo/electrum/commit/ede9b2b37256992c83e35949c09f12b3ee795efa as part of https://github.com/spesmilo/electrum/commit/3ce2986c1fb96863395b887891e3dbcf1d4afdb9.
I did not want to merge this without good justification, but I have now found a scenario where it really makes a significant difference.
Thanks for the PR.

SomberNight:
note: in certain cases this caused a significant performance regression. fixed in https://github.com/spesmilo/electrum/commit/910832c153d736d042c429a58ab54075e973000c

Issue #6463, opened 2020-08-02
not to use change address in rpc


sanada08:
Let me know how not to use change address in rpc. Whenever i am trying to transfer fund. Fund moved to change address. I dont want to use change address.

SomberNight:
note: this is a per-wallet setting persisted in the wallet file:
https://github.com/spesmilo/electrum/blob/ec6baa12f81065ee5b110d6c277a24d752acb11d/electrum/wallet.py#L292

It is currently not configurable via CLI/RPC.

sanada08:
Thanks. Let me try

Issue #6016, opened 2020-03-04
implement advanced wallet option: multiple xpubs, mixed script types, and similar (maybe with output script descriptors)


SomberNight:
Currently HD wallets created by Electrum are severely limited:
- there is a single bip32 xpub/xprv and from that a specific simple tree structure is used to derive keys:
    - keys used to derive receive addresses are at path `m/0/i`, and
    - keys used to derive change addresses are at path `m/1/j`.
- no mixed script types; all the addresses use the same (as it is encoded in the xpub/xprv)

-----

I think it would be really nice if there was an option to create an advanced wallet; a power user feature.
This wallet could not be restored from seed.

It would allow importing xprvs/xpubs: so the user could have multiple of these. Importing could even be allowed for existing wallets (not just in wizard at creation time). These could have differing script types associated with them.

I think even single privkeys/addresses could be allowed to be imported.

Perhaps the user could mark certain paths to be preferred to be given out as receive addresses, and others as preferred change.

Maybe this would be best implemented after support for output script descriptors (https://github.com/spesmilo/electrum/issues/5694), as those seem useful for creating this advanced wallet; and actually the other direction too: output script descriptors seem a lot more useful if such an advanced wallet can be created.

jonjonez:
Certainly, if this release occurs with all these features mentioned, it would be great. I hope you take into consideration what I mentioned in #8879 , and adding:

> For a newbie, the easiest and safest thing to do is to sweep the private key.
As an advanced user, you can create an imported wallet, and use pay-to-multi (batched payment) where you directly pay whatever amount you want to whomever and also send the rest to a change address in your other (HD) wallet.

Sweeping a private key is not viable in times of congested mempool, in addition, the user may have several reasons for just wanting to import instead of sweep. Please take this into consideration.

I know about pay-to-multi, but doing this manually is not viable, it would be nice if this was automatic in the wallet (considering if the private key was imported into the existing keypair).

Currently I can import private keys by creating a new file just for imported keys, there is a change address option to select, then the wallet sends the remaining balance to the next address added as change, but this is not satisfactory, because in an imported wallet I I can only add private keys, that is, if I want to add only addresses to this file without revealing the private key to use it as change I can't, I have to import the private key, and this address of mine can be used as cold storage, but then it will no longer be cold storage because I'd have to reveal the private key, this is insecure and exposes me to risk without any need.

[...]

Issue #5353, opened 2019-05-16
Imported wallets send change back to "from address". address-reuse


SomberNight:
`Dean_Guss` reported in `#electrum` on freenode IRC:

When I import 2 private keys to create a wallet, and 1 address has coins and the other is unused,
then I create a transaction which spends the funds in the funded address, by default change goes back to the same address (address reuse), whereas if there is an unused address in the wallet, change *should* go to the unused address.
Imported addresses which are empty don't seem to be among the available change addresses.

It would be lovely if electrum preferred sending change to wallet addresses in this order:
- Unused, not frozen 	(Ideal)
- Unused, frozen		(prioritizes not re-using addresses)
- Used, not frozen
- Used, frozen		(last resort!)

SomberNight:
> Imported wallets send change back to "from address"

ACK, this is indeed the current behaviour.

gits7r:
I don't think this is a bug or a feature we'd want. I think the current behavior is correct. Wallets with imported private keys can't possible use the same logic as Electrum generated deterministic wallets, which has a purpose for a given address when it's generated, it uses different paths for receive and change and also nests them under separate buckets. This is not technically possible with imported private keys.

How can we know that if an user imported a private key, he wants Electrum to use it for change?

More important, if I freeze an address, this means maybe I have some plans with it, or I froze it with a reason. Would it be OK for Electrum to gracefully use it for change, just to force me not to re-use addresses? Won't the feature "Freeze Address" be confusing in this case? I don't think prioritizing not re-using addresses is more important than what an user explicitly does (freeze), because address re-use is not optimal and not recommended but if would be critical it would have to be addressed at protocol level. Only critical or at least severe practices should be overwritten by the application without caring of user option.

I mean, of course address re-use is bad and not recommended, but this is something to address at protocol level, not at wallet application level. It has been discussed over and over. While the enforcement at application level is totally OK and ideal for Electrum deterministic generated wallets, I don't think I can say the same for imported private keys.

[...]

PR #7330, opened 2021-06-01
Attempt to use empty imported addresses as change

chris-belcher:
Previously when spending coins on imported-key wallets, the change address
would always be an address from one of the inputs. This address reuse is
pretty bad for privacy because it leaks which output is change.

This commit attempts to find unused imported addresses, and if there are any
then use them as change addresses.

The practical use-case I considered was the situation when using the Electrum
android app to do a cash-in-person trade. Some people might want to bring only
one of their UTXOs to the meetup for safety. This commit allows them to import
that one UTXO plus an unused change address, allowing them to create a
transaction without address reuse.

The diff is a single line which shouldn't cause any serious maintenance burden.

I haven't actually tested it on the Android app but by my reading it uses the classes in `electrum/wallet.py` all the same.

Here are screenshots showing before and after the edit.

Before:

imported-wallet-no-reuse-change-addr-before

After:

imported-wallet-no-reuse-change-addr-after

As you can see, the transaction created now uses the unused imported address as change.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -2639,6 +2639,8 @@ class Imported_Wallet(Simple_Wallet):
 
     def __init__(self, db, storage, *, config):
         Abstract_Wallet.__init__(self, db, storage, config=config)
+        #Keep the old default behaviour of not using change address in imported wallets
+        self.use_change = db.get('use_change', False)
 
     def is_watching_only(self):
         return self.keystore is None
@@ -2681,7 +2683,7 @@ class Imported_Wallet(Simple_Wallet):
         return self.get_addresses()
 
     def get_change_addresses(self, **kwargs):
-        return []
+        return [addr for addr in self.get_addresses() if len(self.db.get_addr_history(addr)) == 0]
 
     def import_addresses(self, addresses: List[str], *,
                          write_to_disk=True) -> Tuple[List[str], List[Tuple[str, str]]]:


SomberNight:
See previous discussion at https://github.com/spesmilo/electrum/issues/5353

I agree that your use case makes sense.
However, I think there are scenarios, e.g. if a user imported addresses with different policies (e.g. (1) multisig and singlesig -- though this example is only possible currently for a watch-only wallet; or (2) p2pkh and p2wpkh: should change go to a legacy address to avoid reuse?), where it is difficult to tell where change should go.
Also note that there are users who do not want to use change addresses (see e.g. https://github.com/spesmilo/electrum/issues/6463).

Maybe there could be an option (e.g. in config) to toggle this (send change back to same address, or select arbitrary other address).
Note that there is already an option (not a config option, but rather a per-wallet setting) that toggles this kind of behaviour, except it is currently a no-op for imported wallets:
https://github.com/spesmilo/electrum/blob/a425ab030194c135b1f3b853073be1a7ec6ffc7c/electrum/wallet.py#L291
https://github.com/spesmilo/electrum/blob/a425ab030194c135b1f3b853073be1a7ec6ffc7c/electrum/wallet.py#L1268
In fact, it is a no-op precisely because `get_change_addresses` returns an empty list for imported wallets.
So that option would allow toggling this behaviour.

There is also the question of what the default value for the option should be.
`use_change` defaults to True but currently for an imported wallet it's as if it was set to False; so with your change we are effectively changing the default, affecting both existing and new users.

-----

Note that longer term I plan to rewrite wallet.py/keystore.py to allow importing output script descriptors (or similar) (https://github.com/spesmilo/electrum/issues/6016).
I imagine the UI will have to expose some functionality to let the user mark descriptors as "to be used for change", which would solve all this.

chris-belcher:
Thanks for the reply.

I disagree with @gits7r's reasoning from the other thread that this would be harmful for privacy, and I've written a comment in that thread explaining why.

If users dont want to use change addresses then an easier way of doing that is to create another imported wallet with just their one single private key. Imported wallets are only really for advanced users, regular users should definitely just create seed phrase wallets. So these advances users shouldn't have a problem creating more imported wallet with just one private key if that's what they want, that brings the old behaviour back if that's what they really want.

I understand that this would change the default behaviour, though I'd argue address reuse is so bad for privacy that it's worth it to slightly break backward compatible behaviour.

On the desktop app this PR isn't even needed, because users can use "!" as an amount in the pay-to-many feature, and that address can act as change. This PR is only needed for mobile wallets.

I'm not sure where to go from here. I was thinking this PR might be something simple to easily merge but it looks like there's one or two unresolved issues.

gits7r:
I do understand your use-case, and agree mostly with the arguments but the fix is not acceptable.

Basically, Electrum will force the user to use another imported address for change. What if that user imported that address for a particular reason, and the change sent to it will link it to the other one and create problems for the user? This is a decision the user has to take, when creating the transaction. not Electrum. That is why it's an imported wallet.

Either Electrum uses a NEW/FRESH address for the change from an imported private key, either re-use the address. Deciding "imported private key X is good to receive the change from imported private key Y" is a terrible idea that can have bad complications for the users, and it makes Electrum take a decision on user's behalf that it wouldn't normally want to take.

gits7r:
And don't forget about the wallet back-up problem. Imagine I have just 1 private key saved. I want to restore it every day using an amnesic OS like Tails. I don't have no seed or anything. How do I recover my change? Playing with this behavior is not something we'd like and while of course address re-use is terrible for privacy, imported private keys is not the place where we can make guesses on what the user wants and how they want or don't want to link each imported private key in a wallet between each other.

SomberNight:
Let's keep the high-level discussion in the other thread (https://github.com/spesmilo/electrum/issues/5353).

chris-belcher:
My concern with `use_change` is that it is pretty hard to edit on the desktop app, and basically impossible to edit on the android app. And the android app was my main use-case in mind for this PR.

SomberNight:
> My concern with `use_change` is that it seems pretty hard to edit on the desktop app, and basically impossible to edit on the android app. And the android app was my main use-case in mind for this PR.

Qt: menubar>`Tools`>`Preferences`>`Transactions`>`Use change addresses`
kivy: top-right-corner>`Settings`>`Use change addresses`

chris-belcher:
Ah! Glad to have misunderstood.

I checked and `use_change` already has the effect you describe, where it can disable this feature. I simply added a line changing to default option to False for newly created wallets.

gits7r:
I am not against this feature as long as the default option is "off" but I'd just like to state again with apologies for the redundant comment that: given we think imported private key(s) wallets are for advanced users  -- there's no argument about this we all agree on this, how is it not easier for these advanced users to add one more address in a new pay-to line with ! as amount or whatever, if they'd like the change to go there? Instead of creating a new wallet with only used private keys or freeze addresses, etc. I think it's easier for the user with the current behavior and more rational for Electrum. Electrum can't know if private key X (regardless it is fresh/unused) is OK to receive the change from private key Y. Maybe it was imported for a different reason, maybe it's waiting for an incoming payment not to be linked with the other imported keys, etc. - lot of scenarios. 

My opinion is still that current status quo wrt this topic is OK, while I also heavily agree on the fact that addresses should not be reused, this is particular scenario (imported private key(s) wallets) is simply non-enforceable, except with a lot of tradeoffs and complications for Electrum.

chris-belcher:
> how is it not easier for these advanced users to add one more address in a new pay-to line with ! as amount or whatever, if they'd like the change to go there?

This is not possible on the android app (I think, I wasn't able to do it when I tried a few weeks ago, but I could be wrong, since I was wrong about `Tools>Preferences>Transactions>Use change addresses` just now)

You are right though, if this actually is possible on the android app then my PR isn't really necessary.

SomberNight:
@chris-belcher 
I did not like that `get_change_addresses` depended on the history of the wallet.
Though I appreciate you were most likely trying to minimised the diff :)
I've implemented the same functionality a bit differently in https://github.com/spesmilo/electrum/commit/fbd8c5f7b039fcabc30bcaa020246efb4e50b50c

Thanks for the discussion and your work.

The default behaviour remains the same: imported wallets send change back to the "from address".
However, if the `use_change` option is set, the change is sent to a random unused imported address.

Issue #7322, opened 2021-05-28
Fee increment adds Input and Change address without advertising


donob4n:
Hi,

When raising fee of a 1input x 1output transaction is relative easy to end with a 2input x 2output transaction without noticing which is terrible. First because if it is a 1x1 transaction it is very likely that the user wants to avoid change address and second because the fee raise will be also applied to a double sized transaction (I went from ~2$ to ~8$ when just wanted to go from 40sat to 60sat).

Considering that a fee raise scenario is pretty likely to be under some stress it would be nice to have a clear warn about you are doing and the final result.

SomberNight:
Note the there are advanced options for bumping the fee. In your case the option you seem to want is "decrease payment".

Also note that that is not the default specifically because in a significant chunk of cases that is not what the user wants. E.g. if the user is paying a merchant, decreasing the payment output is highly likely going to invalidate the payment.

With the above in mind, what do you specifically suggest; or can the issue be closed?

donob4n:
I think that some warn dialog like ("Your transaction will have more inputs!") should be nice. I got some testnet coins for try to improve this but the max button is not working properly.

Let me dig a little on it.

donob4n:
How do you see something like this?


![raisefee](https://user-images.githubusercontent.com/35405566/120898274-7774c700-c62a-11eb-948f-f08dd61cda3b.png)

SomberNight:
I don't like it. It is fairly common that we end up adding new inputs, and it is not dangerous or anything. Sure, some people might not want this, but keep in mind that many users probably don't even know what tx inputs/outputs are. For those people this warning is just scary, and without benefits.

Instead, try changing the text to: `(X new inputs will be added)`
Only display it if `X > 0`, and change the font colour from red to the default.
No need to mention the "advanced settings" option that is directly 20 pixels below.
Basically make it a short informational note, instead of scary warning.

donob4n:
Ok, is not dangerous but you will end with pretty higher fee than expected. If you want to raise a 2$ fee transaction you expect something like 2,5$, 3$.... 4$... . A sat/byte increment of 50% is not expected to end in a 400% growth.

But I'm agree, just information message is nice and cleaner.

donob4n:
Ey @SomberNight , could you take a look here?

ecdsa:
Since the refactoring of the transaction creation flow (#8197), it now possible to view tx inputs and outputs interactively, when choosing/bumping the fee. This fixes this issue.

PR #7337, opened 2021-06-06
Added "({0} new inputs will be added)"

donob4n:
Fixes https://github.com/spesmilo/electrum/issues/7322

It needs explicit call to 'on_feerate()' for showing the message on first display.

### DIFF
--- a/electrum/gui/qt/rbf_dialog.py
+++ b/electrum/gui/qt/rbf_dialog.py
@@ -48,19 +48,31 @@ class _BaseRBFDialog(WindowModalDialog):
         ok_button = OkButton(self)
         self.adv_button = QPushButton(_("Show advanced settings"))
         warning_label = WWLabel('\n')
-        warning_label.setStyleSheet(ColorScheme.RED.as_stylesheet())
         self.feerate_e = FeerateEdit(lambda: 0)
         self.feerate_e.setAmount(max(old_fee_rate * 1.5, old_fee_rate + 1))
 
         def on_feerate():
             fee_rate = self.feerate_e.get_amount()
             warning_text = '\n'
+            warning_label.setStyleSheet(ColorScheme.DEFAULT.as_stylesheet())
             if fee_rate is not None:
                 try:
                     new_tx = self.rbf_func(fee_rate)
                 except Exception as e:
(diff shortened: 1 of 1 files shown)


electrum/gui/qt/rbf_dialog.py
```diff
@@ -48,19 +48,23 @@ def __init__(
...
         def on_feerate():
             fee_rate = self.feerate_e.get_amount()
             warning_text = '\n'
+            warning_label.setStyleSheet(ColorScheme.DEFAULT.as_stylesheet())
             if fee_rate is not None:
                 try:
                     new_tx = self.rbf_func(fee_rate)
+                    new_inputs = len(new_tx.inputs()) - len(tx.inputs())
+                    if new_inputs > 0:
+                        warning_text = ("({0} new inputs will be added)".format(new_inputs))
```
SomberNight, line 63: I think it would be better not to guard these three lines by the try-except.
I suggest creating an else branch (try-except-else), and moving these there;
that way, f they raise for whatever reason, the exception can propagate out.

donob4n: UhM, testing this change before committing I noticed this behavior, base tx is 1sat/byte:

- raising to 1 says -> "Cannot bump fee: The new fee rate ...."
- raising to 2 says -> (1 new inputs will be added)
- raising to 3 says -> (2 new inputs will be added)
- raising to 4 it says nothing...

I investigated and yes, starting at 4sat/byte it doesn't add any input but reduces the output. Don't you think that would be nice to add the appropriate message? (output amount will be reduced)

SomberNight: Sure, we could do that, feel free to have a look.
Be warned though that that might require a lot more work to do without ugly hacks as it is mostly abstracted away from the GUI code.

donob4n: I tried to do it in the wallet.py code, but since it needs to add another field to transaction class and also is not clean to implement because the transaction is created at the end of the bump functions... I finally created a comparison function which determines if the change, the payment or both have been decreased.

ecdsa:
no longer needed

PR #7350, opened 2021-06-17
convert contacts at wallet upgrade

proof-ofWork:
issue #7174

### DIFF
--- a/electrum/contacts.py
+++ b/electrum/contacts.py
@@ -41,12 +41,6 @@ class Contacts(dict, Logger):
             self.update(d)
         except:
             return
-        # backward compatibility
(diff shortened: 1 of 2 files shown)


SomberNight (changes requested):
You can't just move this block of code without changing it...

First of all, `self` refers to a different object (with a different type) here;
you would need to find the sub-dictionary where the contacts are stored in the db and modify that (probably `self.data.get('contacts', {})` or similar).

Second, existing db upgrade methods, such as `_convert_version_40` should not be extended after they have been committed to master. The modified upgrade will not run again for wallets the old upgrade had already run for.

PR #7369, opened 2021-06-27
Export daemon stopping code to public API

MrNaif2018:
This PR allows stopping daemon gracefully when using it from async code already (aiohttp app wrapping Daemon object, for example).
It should cause no issues as it is just taking `stop_async` from outside `on_stop` function.
Otherwise, without this fix, calling `on_stop` directly (i.e. from aiohttp `on_shutdown` function) makes the code hang forever because of `run_coroutine_threadsafe` called from async code.
See similar fix there: 66f68d6b1f1a4758059ae8ca377cc0b515784b8d
Ran it locally, all tests still pass

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -454,7 +454,6 @@ class Daemon(Logger):
                 self.network.start_gossip()
 
         self.stopping_soon = threading.Event()
-        self.stopped_event = asyncio.Event()
         self.taskgroup = TaskGroup()
         asyncio.run_coroutine_threadsafe(self._run(jobs=daemon_jobs), self.asyncio_loop)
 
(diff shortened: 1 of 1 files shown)


MrNaif2018:
@SomberNight  I have moved `stop_async` function to `async def stop` function, that's what I meant. From searching references I saw that `run_daemon` is only called by `run_electrum` script, and `async def stop` is called from commands. Instead of waiting for event, it awaits the finalization code directly. All cases you mentioned in IRC seems to work as expected

MrNaif2018:
Closing this. There is no clear way to refactor it without making it more complex, and in my async code I can use `await loop.run_in_executor(None, daemon.on_stop())`

SomberNight:
done in https://github.com/spesmilo/electrum/commit/3643b9f0560dcdc06e3f656af38e535deee99678 and https://github.com/spesmilo/electrum/commit/e0246b30b9e4c5702f6fd48b4139e0a51d745af7

PR #7373, opened 2021-06-28
Fix kivy RefLabel on_touched; status.kv: update balances only when shown

zebra-lucky:
Two small fixes on Kivy GUI:

In `uix/ui_screens/status.kv`, when touched seed and then cancel Pin/Question
popup, second try to show seed is not working. Same for privk in `uix/dialogs/addresses.py`.
This commit fixes this problem:

- kivy: fix RefLabel on_touched processing

Second problem can be reproduced only in Kivy+Linux. If `status.kv` popup is shown
and window is closed by window manager means, exception is raised (parent is set to None,
but `app.wallet` is unloaded already). Possibly it should be fixed by:

- kivy status.kv: update balance only when shown

### DIFF
--- a/electrum/gui/kivy/main.kv
+++ b/electrum/gui/kivy/main.kv
@@ -183,7 +183,10 @@
         touch = args[1]
         touched = bool(self.collide_point(*touch.pos))
         if touched: self.touch_callback()
-        if touched: self.touched = True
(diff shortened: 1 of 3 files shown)


ecdsa:
the kivy GUI was removed.

PR #7385, opened 2021-07-02
Friendlier minimum sizing for small screens for air-gapped signing

BamaHodl:
The current window minimum sizing makes electrum a bit cumbersome on smaller screens (640x480 in my case).
Simply making the central widget of the main window a QScrollArea fixes the main window.
Reducing the minimum size on the BaseTxDialog and setting the width explicitly both allows for it to be resized smaller and still retains the original size.  They both work perfectly on 640x480 after this modification.

### DIFF
--- a/electrum/gui/qt/amountedit.py
+++ b/electrum/gui/qt/amountedit.py
@@ -3,9 +3,9 @@
 from decimal import Decimal
 from typing import Union
 
-from PyQt5.QtCore import pyqtSignal, Qt
+from PyQt5.QtCore import pyqtSignal, Qt, QSize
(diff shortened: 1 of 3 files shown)


SomberNight:
There are some issues with small window sizes, when running your commit.
Some widgets are overlapping, or extending beyond the window; see images below.

<details>
 <summary>pics</summary>

Capture1
Capture2
</details>


----

I've pushed a follow-up, and now I cannot see further obvious issues.

SomberNight:
Thanks for the PR.

BamaHodl:
Thanks for fixing the gui issues created by my PR and merging it!

PR #7395, opened 2021-07-07
Add lightning tag for commands

MrNaif2018:
This pull request adds a new tag for commands decorator to mark command as requiring lightning. It allows for better error messages and identification of lightning commands in cases where it is unsupported. For example, instead of errors about `wallet.lnworker` being `None`, it would say `Lightning not supported in this wallet type`

Few questions:
1) Should commands not requiring wallet be marked as lightning (i.e. `decode_invoice`, `clear_ln_blacklist`)? They seem to not use lightning keys
2) Is it safe to assume that if `wallet.lnworker is None`, then lightning is not supported in a wallet, or a better check should be used?
3) (for understanding 4.1 changes) when restoring a wallet, init_lightning() would always be called for supported wallets, so in my code there is no need to manually call it?

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -93,6 +93,7 @@ class Command:
         self.requires_network = 'n' in s
         self.requires_wallet = 'w' in s
         self.requires_password = 'p' in s
+        self.requires_lightning = 'l' in s
         self.description = func.__doc__
         self.help = self.description.split('.')[0] if self.description else None
         varnames = func.__code__.co_varnames[1:func.__code__.co_argcount]
@@ -145,6 +146,8 @@ def command(s):
                 raise Exception('wallet not loaded')
             if cmd.requires_password and password is None and wallet.has_password():
                 raise Exception('Password required')
+            if cmd.requires_lightning and (not wallet or not wallet.has_lightning()):
+                raise Exception('Lightning not enabled in this wallet')
             return await func(*args, **kwargs)
         return func_wrapper
     return decorator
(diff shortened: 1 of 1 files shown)


SomberNight:
> For example, instead of errors about `wallet.lnworker` being `None`, it would say `Lightning not supported in this wallet type`
> Is it safe to assume that if `wallet.lnworker is None`, then lightning is not supported in a wallet, or a better check should be used?

Let's check `wallet.has_lightning()` (which is equivalent).
Also, in some cases the wallet might support lightning but not have it enabled.

> Should commands not requiring wallet be marked as lightning (i.e. decode_invoice, clear_ln_blacklist)? They seem to not use lightning keys

No, let's not require it for those commands.

> (for understanding 4.1 changes) when restoring a wallet, init_lightning() would always be called for supported wallets, so in my code there is no need to manually call it?

Yes, something like it is done automatically now when restoring from seed:
https://github.com/spesmilo/electrum/blob/8945dcda7ad928e56b9aabf2b46ee5a51551bad7/electrum/wallet.py#L3302-L3305

MrNaif2018:
I have applied your suggestions, anything else to change? Also thanks for clarifying about new 4.1 behaviour

SomberNight:
It looks ok now, thanks.

Please squash the changes into a single commit, and then I will merge that.
Your commits are signed so I cannot squash them for you without losing the signature.

Also, small nit, if you need to touch this again anyway:
re the error message, instead of `Lightning not supported in this wallet type`,
I suggest `Lightning not enabled in this wallet`.

MrNaif2018:
Ok, I hope I've done it right (:

PR #7396, opened 2021-07-07
Don't cleanup lockfile if listen_jsonrpc is False

MrNaif2018:
If `listen_jsonrpc` is `False`, lockfile is not used, so if doing cleanup gracefully, error occurs:
`FileNotFoundError: [Errno 2] No such file or directory: '/home/alex/.electrum/testnet/daemon`


SomberNight:
Thanks

PR #7403, opened 2021-07-12
Improve _mythread checks

MrNaif2018:
This PR removes `_mythread` checks and uses native asyncio APIs instead.
Why it works:
If there is no event loop in current thread, `_get_running_loop()` returns None, which is not equal to self.asyncio_loop. Network thread has self.asyncio_loop running. Otherwise loop objects won't be equal
I have tested scripts/quick_start, works as expected, also tried using `run_from_another_thread` in `async def _start` method running in network thread, assert is working.
It creates `get_running_loop` utility
This can be changed of course
All tests pass (tested locally, remotely travis hates me)

### DIFF
--- a/electrum/lnrater.py
+++ b/electrum/lnrater.py
@@ -14,13 +14,8 @@ from typing import TYPE_CHECKING, Dict, NamedTuple, Tuple, List, Optional
 import sys
 import time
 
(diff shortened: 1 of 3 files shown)


SomberNight:
thanks

Issue #7266, opened 2021-05-01
Reproducible build for macOS/OSX (.dmg)


SomberNight:
Currently [the macOS build](https://github.com/spesmilo/electrum/tree/ef3293ab6ac64970bb6fa24be33dbe606bbd344e/contrib/osx) is not deterministic/reproducible.

It is the last holdout as the Windows exes, Linux AppImage, the Android APKs, and the source dist are all reproducible.

I suspect this might be more convenient if we managed to [cross-compile from Linux to macOS](https://github.com/spesmilo/electrum/issues/7265) first, and then tackle this.

Around early 2018, @bauerj spent some time on trying to make the macOS build reproducible, without success; the approach tried then did not involve cross-compiling from Linux to OSX, but rather to use a Parallels VM with macOS guest running on macOS host, along with small adaptations to the build script (but still using pyinstaller). IIRC the main issue seemed to be compiling the Python interpreter reproducibly. I think bauerj managed to get reproducibility on the same machine, but not across different machines.

@achow101 also [tried for HWI](https://github.com/bitcoin-core/HWI/pull/121#issuecomment-469533748), and got similar results.

Note that it would be okay to not build the python interpreter ourselves but e.g. somehow use the "official" build from `python.org`. This is already what we do for the Windows binaries.

I tried to look several times over the years, but could never find a python project that managed to build reproducible binaries for macOS.

PR #7415, opened 2021-07-19
mac build: attempt at "reproducible" codesigned builds

SomberNight:
- added notes about reproducibility requirements
- adapted build scripts from Bitcoin Core that can
    - extract signatures from a signed .app
    - apply previously extracted signatures to an unsigned .app

closes https://github.com/spesmilo/electrum/issues/7266

### DIFF
--- a/contrib/osx/README.md
+++ b/contrib/osx/README.md
@@ -1,9 +1,8 @@
 Building macOS binaries
 =======================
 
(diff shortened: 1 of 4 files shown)


ecdsa:
note: this assumes that the signed app has been notarized; `Contents/CodeResources` is created by notarization

Issue #6423, opened 2020-07-20
Feature Suggestion: Ability to display balance in satoshis with thousands separators


wingsuit:
To improve readability when displaying balance in satoshis, displaying the balance with spaces, commas or periods would dramatically improve readability.

Allowing the user to click on the balance to cycle through the various options would be useful (no separator -> commas -> dots -> spaces)

Another readable way could be using SI prefixes, such that a balance of 584849211 satoshis (5.8 btc) would be displayed as 584M 849k 211 

While the debate around satoshis is ongoing, the ability to at least allow proponents of this unit to display them in meaningful ways would increase inclusivity.

SomberNight:
I agree we should try to do something like this.
To be clear it's not just about the wallet "balance" but most values displayed all over the GUI, so also e.g. the "Amount" and "Balance" columns in the Qt GUI.

Re the thousands separator character itself, I think we should use whitespaces " ", as the dots and commas are used by many countries with varying meaning (in some place dot is decimal point, comma is thousands separator; in other places it is the other way around).

So instead of `86940861 sat`, we could show `86 940 861 sat`.
I think we can also extend the thousands separator to the fractional part:
so instead of `0.86940861 BTC`, we could show `0.869 408 61 BTC`.

Maybe there could be a config option to enable/disable this behaviour (the thousands separator).
Although I think it is a net improvement, so maybe there is no need for an option.

dhruv-joshi-7:
I have implemented some changes to my local system regarding the readability of the amount. Attaching the images with commas and with whitespaces. Do tell me which one you find better.

![commas](https://user-images.githubusercontent.com/73574557/125932756-77d030bf-323e-4039-8a2b-0b8b35380cc0.png)
![spaces](https://user-images.githubusercontent.com/73574557/125932771-bb09cc79-0214-42b4-9459-eec4e9968b4c.jpeg)
![spaces_decimal](https://user-images.githubusercontent.com/73574557/125932780-d8bc8f1c-353f-4abc-8d46-ffc9936a88ce.jpeg)

SomberNight:
Done in https://github.com/spesmilo/electrum/pull/7427
(atm only whitespaces can be used as thousands separators)

paintballpanda:
[...]

PR #7416, opened 2021-07-19
Added whitespaces for better readability for balance in satoshis

dhruv-joshi-7:
Addressed #6423 about adding whitespaces for the thousands separator.

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -681,7 +681,27 @@ def format_satoshis(
     if whitespaces:
         result += " " * (decimal_point - len(fract_part))
         result = " " * (15 - len(result)) + result
-    return result
+    dot = result.find('.')
+    if len(result) == dot + 1:
+        if dot != -1:
+            satoshi_integer = result[0:dot]
+            satoshi_num = int(satoshi_integer)
+            result = "{:,}".format(satoshi_num).replace(',', ' ') + result[dot:]
+            return result
+        else:
+            return result
+    else:
+        if dot != -1:
+            satoshi_integer = result[0:dot]
+            satoshi_num = int(satoshi_integer)
+            satoshi_decimal = result[dot+1:]
+            s_new = ''
+            for i in range(0,len(satoshi_decimal),3):
+                s_new += satoshi_decimal[i:i+3] + ' '
+            result = "{:,}".format(satoshi_num).replace(',', ' ') + '.' + s_new
+            return result
+        else:
+            return result
 
 
 FEERATE_PRECISION = 1  # num fractional decimal places for sat/byte fee rates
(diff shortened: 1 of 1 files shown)


electrum/util.py
```diff
@@ -681,7 +681,27 @@ def format_satoshis(
     if whitespaces:
         result += " " * (decimal_point - len(fract_part))
         result = " " * (15 - len(result)) + result
-    return result
+    dot = result.find('.')
```
SomberNight, line 684: Note that, at this point, it might not be the `.` character that is used to separate the integer part and the fractional part:
it was replaced above with the locale-specific `DECIMAL_POINT`.

SomberNight: Also note that the number was already split based on the decimal point into `integer_part` and `fract_part`; and then recombined into a single string.
Instead of splitting the number string and combining it again, you should move your logic further up and add the thousands separators there: perhaps right after where the extra zeros are added to the fractional part (line 678).

SomberNight (changes requested):
I've read and tested your changes. The code works and it's relatively readable; however it is too verbose; and I think has unnecessary branching (lots of if-else cases that could be handled together).

In general, instead of treating what has been done in the method so far as a black box, you should try to understand what the existing code does, to see how that interacts with what you want and see if you can or need to adapt that.
Note that code is only written once but read hundreds of times: you should make it succint and understandable, not too verbose, to help the people who will maintain the code for years (which might include yourself). Keep it simple.

Also, note that there unit tests for this method in tests/test_util.py, which are now failing.
Please adapt the tests so that they pass, and consider if it would make sense to write new ones.
See https://github.com/spesmilo/electrum/blob/173225ae92e1fa1a1926d485f3d61f3c498fab9b/electrum/tests/test_util.py#L14
(we are having issues with Travis CI atm, so it did not run automatically)

SomberNight:
While looking at this, I've found two bugs with `format_satoshis` and fixed them in https://github.com/spesmilo/electrum/commit/5c80293696681c811610a0caaa84a1a2854b2697 and https://github.com/spesmilo/electrum/commit/46badd128e642d5f5ee3e91d8c70263eb53dbb32.
I've also added more unit tests.

While not directly related to this feature, please have a look at those commits.
Also, make sure to git pull latest master so you have them, and rebase this pull request.

Re adding thousands separators, I think `format_satoshis` should take a new parameter, named e.g. `add_thousands_sep`, which defaults to False, and only add the thousands separators when set to True. This would help regarding the unit tests as well: existing tests would keep working -- but a few new ones should be added that exercise the new code.

Also see this third commit I made that exposes a config option to the user interface to modify how `format_satoshis` behaves:
https://github.com/spesmilo/electrum/commit/5891e039b1db74c0eed14fbf9b9f4e61d9f3546c
Note how most code paths go through `config.format_amount` -- you could set `add_thousands_sep=True` there, when calling `format_satoshis`.

SomberNight:
See e.g. this re how to rebase a pull request: https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request
(although maybe only for next time, as you now deleted your branch - I think that means you need to open a new PR)

Issue #7189, opened 2021-04-07
Installwizard.py crashes if you try to go back to step 1


gits7r:
Got a crash. To reproduce:
Electrum -> New/Restore wallet -> Standard -> I already have a seed -> enter seed + tick Options/BIP39 seed -> go to the derivation path menu, then click Back , Back, Back to go to phase 1 (step before I already have a seed). It'll crash with this:

![bugtrace](https://user-images.githubusercontent.com/6549887/113936816-de6f2e80-97e7-11eb-86f9-e28f83da6b58.png)

Also our automated crash report does not work:

![crashof crash](https://user-images.githubusercontent.com/6549887/113936844-e9c25a00-97e7-11eb-91d1-21738eaa1dba.png)

ecdsa:
thanks, this explains #6301

SomberNight:
No longer relevant. The qt wizard was rewritten mostly from scratch.

PR #7424, opened 2021-07-20
redraw window before raising GoBack, fix Back button handling on some windows

Imaniac230:
A simple change. The `GoBack` exception forces execution before an installwizard window updates. If there are longer calculations from a previous window that have to be redone, the current screen acts like it's frozen. This was more visible on older HW where I found myself clicking the "Back" button several times thinking the app had frozen. I think the user could be notified here that the GUI acknowledged his Back button press by displaying the "Please wait" screen the same way as when pressing Next.

//Deleted the question relating to `ReRunDialog` and `go_back`, the changes are now applied in https://github.com/spesmilo/electrum/pull/7424/commits/014075462fea304f365d44c8e6eed1a4514e2161.


Imaniac230:
When using the optional passphrase, the actions of the dialogs do not get added to the stack. The displayed windows then got disassociated from the stack and reacted incorrectly to the `Back` button. (ex. `New wallet > Standard wallet > Create new seed > extend with passphrase >> then go back all the way from seed confirmation dialog`). Commit https://github.com/spesmilo/electrum/pull/7424/commits/efcd1fdda05953baf5f3907259aa9fcd01560774 changes function calls so that the corresponding tasks get added to the stack and it is always even with displayed windows and chosen options.

Commit https://github.com/spesmilo/electrum/pull/7424/commits/4dfd49439424dfe0e142d2369d463b6e5269f340 refactors the `func_wrapper()` in `wizard_dialog()` to use the `rerun_previous` argument of `go_back()` instead of raising the `ReRunDialog` exception.

These might relate to https://github.com/spesmilo/electrum/issues/7189 and https://github.com/spesmilo/electrum/issues/6301 maybe.

Tested all of this with various combinations of options, `Back` and `Next` choices and wallet types and all seems to work as expected.

SomberNight:
Thanks for looking at this stuff.
The wizard "go back" behaviour in general is more complex than I would like it to be, and often when I look into a related bug I have to rediscover and re-test how it works and why :/
So any simplification is welcome!

> When using the optional passphrase, the actions of the dialogs do not get added to the stack

Indeed, this is a bug.

Re the other GoBack changes, I've tested this a bit with a hardware wallet.

Please test these scenarios:
- scenario A, on master:
  1. (create new) standard wallet
  2. (keystore) use a hw device
  3. on "select device" screen, select first device from list (assuming there is one connected)
  4. now I am on "script type and derivation" screen
  5. pressing back goes to "select device" screen
  
  repeat (3-4-5) and note that the stack is getting eaten

- scenario A, on your branch:
  1. (create new) standard wallet
  2. (keystore) use a hw device
  3. on "select device" screen, select first device from list (assuming there is one connected)
  4. now I am on "script type and derivation" screen
  5. pressing back goes to "keystore" screen << it would make more sense if it went to "select device" scren :/

- scenario B, on master:
  1. (create new) standard wallet
  2. (keystore) use a hw device
  3. on "select device" screen, select first device from list (assuming there is one connected)
  4. now I am on "script type and derivation" screen, press next
  5. on "password" screen, press back
  6. note that on master you **can go back** (although if you press back a few times, I think the stack is missing the first item again, so it might be getting smaller) -- on your branch it is behaving weird
  
  Also note that for the non-hw-device case, the password screen already resets the stack and forbids going back, so if needed we could do that in the hw-device case too. see https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L638-L639

SomberNight:
btw if it turns out that `ReRunDialog` is indeed not needed at all, then you should also just delete the class definition and all references

Imaniac230:
Unfortunately, I don't have a HW wallet so I wasn't able to test that path. But I'll see if I can set up an emulator which might hopefully work for the window tests and I can play with that too.

> The wizard "go back" behaviour in general is more complex than I would like it to be, and often when I look into a related bug I have to rediscover and re-test how it works and why :/

The one thing I found a little confusing is figuring out which functions should be called with `run()` (which appends to the stack) and which directly (which doesn't) so that the stack is maintained correctly. That I think leads to these kinds of mistakes. Maybe this is something that could be more streamlined.

Also found similar behavior with 2fa wallet creation. After seed confirmation from the Terms of Service dialog onward, the stack doesn't get updated.

SomberNight:
> Unfortunately, I don't have a HW wallet so I wasn't able to test that path. But I'll see if I can set up an emulator which might hopefully work for the window tests and I can play with that too.

Ah sorry, for some reason reading "older HW" got me thinking hardware signer instead of "slow PC". :P
You could try with the Trezor T emulator: https://github.com/trezor/trezor-firmware/blob/master/docs/core/emulator/index.md

Imaniac230:
Ok, so for some reason I wasn't able to compile the latest Trezor T emulator but the [deprecated repo](https://github.com/trezor/trezor-core) worked. This should be fine since we're only testing the screens.

Going through the whole wallet creation, here's what the stack looked like when running on master:
1. `len: 1. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={})]`
2. `len: 2. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`
3. `len: 3. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='choose_hw_device', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`
4. `len: 3. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='choose_hw_device', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`
5. `len: 5. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='choose_hw_device', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='on_hw_derivation', args=('trezor', DeviceInfo(device=Device(path='bridge:2', interface_number=-1, id_='bridge:2', product_key='Trezor', usage_page=0, transport_ui_string='bridge:2'), label='', initialized=True, exception=None, plugin_name='trezor', soft_device_id='0DEBCAE30811A99BC700962A', model_name='Trezor T'), "m/84'/0'/0'", 'p2wpkh'), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='create_wallet', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`

There are 5 screens total from _Create new wallet_ to the "password" screen. However at the "script type and derivation" screen the stack just repeats (doesn't get updated). That's where it was getting eaten, popping one off and repeating. Also at the last step more items get added to the stack offsetting it further. After https://github.com/spesmilo/electrum/pull/7424/commits/de39d17d24fa17641286720ce64a692786ae8433 the stack looks like this and the `Back` and `Next` clicks should work properly:
1. `len: 1. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={})]`
2. `len: 2. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`
3. `len: 3. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='choose_hw_device', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`
4. `len: 4. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='choose_hw_device', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='on_device', args=('trezor', DeviceInfo(device=Device(path='bridge:2', interface_number=-1, id_='bridge:2', product_key='Trezor', usage_page=0, transport_ui_string='bridge:2'), label='', initialized=True, exception=None, plugin_name='trezor', soft_device_id='0DEBCAE30811A99BC700962A', model_name='Trezor T')), kwargs={'purpose': 0, 'storage': None}, db_data={'wallet_type': 'standard'})]`
5. `len: 5. stack: [WizardStackItem(action='new', args=(), kwargs={}, db_data={}), WizardStackItem(action='choose_keystore', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='choose_hw_device', args=(), kwargs={}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='on_device', args=('trezor', DeviceInfo(device=Device(path='bridge:2', interface_number=-1, id_='bridge:2', product_key='Trezor', usage_page=0, transport_ui_string='bridge:2'), label='', initialized=True, exception=None, plugin_name='trezor', soft_device_id='0DEBCAE30811A99BC700962A', model_name='Trezor T')), kwargs={'purpose': 0, 'storage': None}, db_data={'wallet_type': 'standard'}), WizardStackItem(action='create_wallet', args=(), kwargs={}, db_data={'wallet_type': 'standard'})]`

When testing with a wiped HW signer, the additional setup interactions I believe are out of scope of `base_wizard` and don't pass through `wizard_dialog()`. So I think it makes sense if it currently just returns to "select device" if not completely canceled.

> * Also note that for the non-hw-device case, the password screen already resets the stack and forbids going back, so if needed we could do that in the hw-device case too.

Referring to scenario B, I'm not sure if it's genuine HW behavior or just another bug but after I get to the "password" screen and return several steps back it's not able to detect the device anymore as if it's already busy with an opened wallet file. So maybe it also makes sense to reset the stack at this point?

I think for now it should work in most scenarios.

Imaniac230:
In https://github.com/spesmilo/electrum/pull/7424/commits/225e68ec9d6a0c5140caf7c431951bef5a4a5a46 I reformatted some more function calls so that the stack should now be even with all screen dialogs up to the final "password" screen. The non-HW device path (`Create new standard wallet > Create new seed > Confirm seed (with optional passphrase if used) > Encrypt with password`) can now return all the way from the last screen, create a new seed (or passphrase) and create the wallet with the updated seed (passphrase).

I assumed the `on_keystore()` function will always create a single keystore for a standard (non-multisig) wallet and clear the list before appending here. https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L576-L582 Then it creates the correct wallet even after going back from the "password" screen.

I tested the updated branch with various scenarios with standard, multisig and HW wallets and it seems to work correctly. However, this still remains 
> Referring to scenario B, I'm not sure if it's genuine HW behavior or just another bug but after I get to the "password" screen and return several steps back it's not able to detect the device anymore as if it's already busy with an opened wallet file. So maybe it also makes sense to reset the stack at this point?

Imaniac230:
Commit https://github.com/spesmilo/electrum/pull/7424/commits/0f2e552143daf0575960b1629409e5f825d9977a fixed the last problem and both standard wallet paths (HW and non-HW) can return back from the last screen correctly. Multisig shouldn't be influenced by this. Though I think other inconsistencies might still resurface with different untested combinations in the future before some significant simplifications are made.

SomberNight:
Re side-effects of wizard.keystores, I would prefer not allowing going back after a keystore is added, I think that might be simpler. This is already the behaviour in the majority of cases.
In the multisig case, there is an explicit wizard.reset_stack() right before wizard.keystores.append(): https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L599-L606
in the standard wallet non-hw case, there is this: https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L638-L639
So instead of your changes where you do `self.keystores.clear()` (added in two places), we could always do like the multisig case, and call wizard.reset_stack() before wizard.keystores.append(). If you like, these two lines could even be moved to a separate method, e.g. `_append_keystore` or similar, to make it clear they should always go together.

Otherwise the changes look good.

Imaniac230:
Yes, that would the clearest solution. The `self.keystores.clear()` and the other additions in https://github.com/spesmilo/electrum/pull/7424/commits/0f2e552143daf0575960b1629409e5f825d9977a would not be necessary since we wouldn't need to manage going back from the final screen (after a keystore is created).

The problems I had were:

1. When a keystore gets created from an initial (1st) seed, but I return back to generate a new (2nd) seed and append the new keystore, the `create_wallet()` (or `on_password()`) function always uses the initial (1st) keystore to create the wallet.
2. Similarly when going all the way back to the first screen and attempting to choose a different wallet type, the original keystore would get incorrectly reused.
3. Additionally in the HW case, the signer was left paired with the created key so going back and re-selecting the device was not possible.

These, and possibly more with other different combinations of scenarios, would indeed not need to be accounted for since we wouldn't let the user go back after the keystore gets created.

> So instead of your changes where you do `self.keystores.clear()` (added in two places), we could always do like the multisig case, and call wizard.reset_stack() before wizard.keystores.append(). If you like, these two lines could even be moved to a separate method, e.g. `_append_keystore` or similar, to make it clear they should always go together.

I think this would be good. Regarding the naming convention, does the _ at the beginning relate to any specific method type (like a private method) or is it just to differentiate from the native `append()` method?

SomberNight:
>> e.g. `_append_keystore` or similar, to make it clear they should always go together.
>
> I think this would be good. Regarding the naming convention, does the _ at the beginning relate to any specific method type (like a private method) or is it just to differentiate from the native append() method?

I meant that to signify that it is a private/internal method.
I don't think it needs much differentiating from e.g. `list.append`, as this would be a method on the wizard.

Imaniac230:
Commit https://github.com/spesmilo/electrum/pull/7424/commits/d36230aa35f6d16bed3d9a7b5aaaff2ea65fccad reflects the discussed default behavior when appending a keystore.

Additionally, I found that the stack could get misaligned at the `ChooseHwDeviceAgain` exception since it can be raised when the stack is at different states (lengths), so I added some additional management. Also if it gets raised from the `create_wallet()` function, the old appended keystore would get incorrectly reused in the next cycle (similar to the previously described problems) so I added the `self.keystores.pop()` for those cases.

Imaniac230:
Also in `on_keystore()` I think these returns https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L597-L598 could chain the stack with multiple `choose_keystore` actions, so maybe this should also get a fix. Maybe creating an exception and reverting the stack to the last `choose_keystore` as in was added in `ChooseHwDeviceAgain`?

And I'm not really sure how this https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L234-L253 should behave regarding the `keystores.append` and going back.

Imaniac230:
> Also in `on_keystore()` I think these returns
> 
> https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/base_wizard.py#L597-L598
> 
> could chain the stack with multiple `choose_keystore` actions, so maybe this should also get a fix. Maybe creating an exception and reverting the stack to the last `choose_keystore` as in was added in `ChooseHwDeviceAgain`?

Or maybe simply just reset the stack before calling `self.run('choose_keystore')`.

Imaniac230:
@SomberNight please see the latest changes. Do they look OK?

ecdsa:
this will not be merged; the wizard code has been changed.

Issue #7410, opened 2021-07-15
kivy: show fiat amount when paying lightning invoice


SomberNight:
(Assuming fiat exchange rates are enabled,) when paying a lightning invoice, in the kivy GUI, it is very hard to see how much we are going to send in terms of fiat.

The Qt GUI shows a confirmation prompt to the user with both the bitcoin and the fiat amount displayed.
The kivy GUI includes a similar confirmation prompt, however it currently does not show either amount.

We could change the kivy GUI to also show the bitcoin and the fiat amount when paying a lightning invoice.
(if fiat exchange rates are disabled, only show the bitcoin amount)

Here are some pics for how the confirmation prompt looks currently when trying to pay a LN invoice:
<details>
 <summary>qt gui</summary>

![Capture](https://user-images.githubusercontent.com/29142493/125791846-99b97336-7388-470f-94a3-dc14ef04510c.PNG)
</details>
<details>
 <summary>kivy gui, pin code enabled</summary>

![Capture2](https://user-images.githubusercontent.com/29142493/125791909-9ae0e733-f3e2-4c5f-a30a-6c9551d39edf.PNG)
</details>
<details>
 <summary>kivy gui, pin code disabled</summary>

![Capture3](https://user-images.githubusercontent.com/29142493/125791931-6221cea8-7b9c-481a-b513-f58b20e46629.PNG)
</details>

PR #7425, opened 2021-07-21
Added BTC and Fiat amount on the confirmation screen for Lightning Invoice

siddhantchawla:
Added BTC and Fiat amount(as enabled by the user) on the confirmation screen for a Lightning Invoice in the Kivy GUI #7410

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -1202,6 +1202,11 @@ class ElectrumWindow(App, Logger):
         self.set_fee_status()
 
     def protected(self, msg, f, args):
+        tx_btc = args[0].amount_msat/1000
+        msg += "\n\n" + _("This will send " + str(self.format_amount(tx_btc).strip()) + ' ' + self.base_unit )
+        if self.fx.is_enabled():
+            msg += _(' (' + str(self.fx.format_amount(tx_btc).strip()) + ' ' + self.fx.ccy + ')')
+        msg += '?'
         if self.electrum_config.get('pin_code'):
             msg += "\n" + _("Enter your PIN code to proceed")
             on_success = lambda pw: f(*args, self.password)
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -973,6 +973,13 @@ class ElectrumWindow(App, Logger):
             return 'max'
         return format_satoshis_plain(x, decimal_point=self.decimal_point()) + ' ' + self.base_unit
 
+    def format_amount_and_units_with_fiat(self, x) -> str:
+        text = self.format_amount_and_units(x)
+        fiat = self.fx.format_amount_and_units(x) if self.fx else None
+        if text and fiat:
+            text += f' ({fiat})'
+        return text
+
     def format_fee_rate(self, fee_rate):
         # fee_rate is in sat/kB
         return format_fee_satoshis(fee_rate/1000) + ' sat/byte'
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -1202,11 +1202,6 @@ class ElectrumWindow(App, Logger):
         self.set_fee_status()
 
     def protected(self, msg, f, args):
-        tx_btc = args[0].amount_msat/1000
-        msg += "\n\n" + _("This will send " + str(self.format_amount(tx_btc).strip()) + ' ' + self.base_unit )
-        if self.fx.is_enabled():
-            msg += _(' (' + str(self.fx.format_amount(tx_btc).strip()) + ' ' + self.fx.ccy + ')')
-        msg += '?'
         if self.electrum_config.get('pin_code'):
             msg += "\n" + _("Enter your PIN code to proceed")
             on_success = lambda pw: f(*args, self.password)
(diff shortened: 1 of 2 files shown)


electrum/gui/kivy/main_window.py
```diff
@@ -1202,6 +1202,11 @@ def on_fee(self, event, *arg):
         self.set_fee_status()
 
     def protected(self, msg, f, args):
+        tx_btc = args[0].amount_msat/1000
+        msg += "\n\n" + _("This will send " + str(self.format_amount(tx_btc).strip()) + ' ' + self.base_unit )
```
SomberNight, line 1206: `_()` is a function that is used to translate strings. It can only translate static strings.
So if we want translation to work, instead of
`_("This will send " + str(self.format_amount(tx_btc).strip()) + ' ' + self.base_unit )`
you could do
`_("This will send {}").format(str(self.format_amount(tx_btc).strip()))  + ' ' + self.base_unit`

I haven't checked but why do we need `str(self.format_amount(tx_btc).strip())`, instead of just `self.format_amount(tx_btc)`? If the latter works, just do that.

See another example of string formatting with translation here:
https://github.com/spesmilo/electrum/blob/5891e039b1db74c0eed14fbf9b9f4e61d9f3546c/electrum/gui/qt/main_window.py#L1557

electrum/gui/kivy/main_window.py
```diff
@@ -1202,6 +1202,11 @@ def on_fee(self, event, *arg):
         self.set_fee_status()
 
     def protected(self, msg, f, args):
+        tx_btc = args[0].amount_msat/1000
```
SomberNight, line 1205: This function, `protected`, is used as an indirection to call other methods `f` when the user should be prompted for their PIN code. It is not only used when paying an invoice, but also when e.g. displaying their seed or exporint a private key or deleting the wallet.
With the app in kivy open, e.g. try tapping the wallet name in the top left corner, and then tap your seed near the middle of the screen ("Tap to show"). That call is going through `protected`.

I suggest you instead modify the msg argument being passed to `protected` here:
https://github.com/spesmilo/electrum/blob/5891e039b1db74c0eed14fbf9b9f4e61d9f3546c/electrum/gui/kivy/uix/screens.py#L344

electrum/gui/kivy/main_window.py
SomberNight: Re the line breaks, see pic how the screen looks currently:
<details>
 <summary>pic</summary>

Capture
</details>

It's just a nit, but I suggest one fewer `\n` after `Pay lightning invoice?`,
and one more before `Enter your PIN code to proceed`.
i.e. there should be two paragraphs: first two sentences in first, `Enter your PIN code to proceed` in second.

SomberNight (changes requested):
I've looked at the code and tested it.
Please see comments.

The happy path works but there are other code paths that break (such as displaying your seed).
In general, please try to search the code to see where a method is called from to understand if a change should be made there.

electrum/gui/kivy/main_window.py
siddhantchawla: Hi, I have made the changes, please have a look again. 
Re the line breaks, I have kept 2 between all these lines to give a cleaner look, have a look and tell me if it's fine!

siddhantchawla: image

electrum/gui/kivy/uix/screens.py
```diff
@@ -341,7 +341,9 @@ def do_pay(self):
     def do_pay_invoice(self, invoice):
         if invoice.is_lightning():
             if self.app.wallet.lnworker:
-                self.app.protected(_('Pay lightning invoice?'), self._do_pay_lightning, (invoice,))
+                amount_sat = invoice.amount_msat/1000
```
SomberNight, line 344: I suggest simply calling `invoice.get_amount_sat()`.
https://github.com/spesmilo/electrum/blob/2881d1e62c8894205d27f565fb3808ec27889eb8/electrum/invoices.py#L195

electrum/gui/kivy/uix/screens.py
```diff
@@ -341,7 +341,9 @@ def do_pay(self):
...
+                msg = _("Pay lightning invoice?") + '\n\n' + _("This will send {}?\n").format(self.app.format_amount_and_units(amount_sat))
```
SomberNight, line 345: Note that in kivy (unlike the similarly named method in qt), `format_amount_and_units` does not append the fiat value, so e.g. this will only return "0.05 BTC".
I suggest introducing a new method in main_window.py, named e.g. `format_amount_and_units_with_fiat` that behaves like `format_amount_and_units` does in qt: it returns a string with both the formatted btc value and the fiat value. (later, but not in this PR, all these `format_amount_*` method should be refactored amd cleaned up...)

SomberNight: Please move the line break outside the translation, i.e.:
```suggestion
                msg = _("Pay lightning invoice?") + '\n\n' + _("This will send {}?").format(self.app.format_amount_and_units(amount_sat)) + '\n'
```

SomberNight (changes requested):
Looking better, but still some changes needed :)
see comments

SomberNight (approved):
Looks good. Thanks!

PR #7426, opened 2021-07-21
flexible lnpeer test graphs and spp trampoline tests

bitromortac:
~~Based on #7202 (first three commits).~~

- Implements a way to define arbitrary test graphs
- Adds tests for legacy and end-to-end single-part trampoline payments
- When a (trampoline) payer computes a trampoline route based on route hints of a trampoline receiver, we need to use typical trampoline forwarder fees instead of the normal fee policy (they would be too low)
- Fixes a bug, where an Electrum trampoline forwarder doesn't include the correct payment secret when forwarding to a legacy receiver

### DIFF
--- a/electrum/tests/test_lnpeer.py
+++ b/electrum/tests/test_lnpeer.py
@@ -932,6 +932,51 @@ class TestPeer(TestCaseForTestnet):
         self._run_mpp(graph, {'mpp_invoice': False}, {'mpp_invoice': True})
 
     @needs_test_with_all_chacha20_implementations
+    def test_payment_trampoline(self):
+        async def turn_on_trampoline_alice():
+            if graph.workers['alice'].network.channel_db:
+                graph.workers['alice'].network.channel_db.stop()
+                await graph.workers['alice'].network.channel_db.stopped_event.wait()
+                graph.workers['alice'].network.channel_db = None
+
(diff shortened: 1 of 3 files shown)


SomberNight:
needs rebase, now that base PR is merged

electrum/tests/test_lnpeer.py
```diff
@@ -289,38 +289,62 @@ class PeerInTests(Peer):
...
+            },
+        },
+        'bob': {
+            'channels': {
+                'dave': high_fee_channel.copy(),
+            },
+            'config': {
+                'lightning_forward_payments': True,
+                'lightning_forward_trampoline_payments': True,
+            },
```
SomberNight, line 327: So channels only need to be defined for one of the peers and they will get added to both peers,
but configs are per-peer-specific. Maybe this could be briefly mentioned in a comment.
Looks reasonable.
Note that this construction does not permit asymmetric forwarding fees for a given channel -- but the current tests do not need that so it's fine.

SomberNight (approved):
Looks good. Thanks!

electrum/tests/test_lnpeer.py
bitromortac: True, here I tried to opt for a more compact representation.

PR #7427, opened 2021-07-21
Fixed issue with thousands separator for better readability

dhruv-joshi-7:
Note: this is a continuation of https://github.com/spesmilo/electrum/pull/7416


dhruv-joshi-7:
@SomberNight I opened this new PR please check.

electrum/util.py
```diff
@@ -677,6 +678,15 @@ def format_satoshis(
...
         fract_part += "0" * (num_zeros - len(fract_part))
     result = integer_part + DECIMAL_POINT + fract_part
+    # add whitespaces as thousands' separator for better readability of numbers
+    if add_thousands_sep:
+        integer_part = "{:,}".format(int(integer_part)).replace(',', ' ')
+        if len(fract_part) > 3:
+            fractional_part = fract_part
+            fract_part = ''
+            for i in range(0,len(fractional_part),3):
+                fract_part += fractional_part[i:i+3] + ' '
```
SomberNight, line 688: I think this whole block can just be replaced like this :)
```suggestion
        fract_part = " ".join(fract_part[i:i+3] for i in range(0, len(fract_part), 3))
```

electrum/util.py
```diff
@@ -677,6 +678,15 @@ def format_satoshis(
...
+        result = integer_part + DECIMAL_POINT + fract_part
```
SomberNight, line 689: No need to duplicate this line from line 680.
Just move this out to after the outer if block, and remove line 680.

SomberNight (changes requested):
Looks a lot better than before! See comments.
However, please add unit tests. I think the `is_diff=True` case is broken.

electrum/util.py
```diff
@@ -677,6 +678,15 @@ def format_satoshis(
     if len(fract_part) < num_zeros:
         fract_part += "0" * (num_zeros - len(fract_part))
     result = integer_part + DECIMAL_POINT + fract_part
+    # add whitespaces as thousands' separator for better readability of numbers
+    if add_thousands_sep:
+        integer_part = "{:,}".format(int(integer_part)).replace(',', ' ')
```
SomberNight, line 683: Note: with `is_diff=True`, this line is swallowing the plus sign, as e.g. `int("+7")==7`,
and also the minus sign for values less than 1 BTC (100 million sat):
`int("-0")==0` but `int("-5")==-5`

dhruv-joshi-7: The main point of `is_diff` is to show the `+/-` sign in the Amount column right?

dhruv-joshi-7:
Please review the new changes!

electrum/simple_config.py
```diff
@@ -659,14 +659,15 @@ def get_netaddress(self, key: str) -> Optional[NetAddress]:
             except:
                 pass
 
-    def format_amount(self, x, is_diff=False, whitespaces=False):
+    def format_amount(self, x, is_diff = True, whitespaces = False, add_thousands_sep = True):
```
SomberNight, line 662: Why change the default value of `is_diff`?

SomberNight: also, nit, but please remove the whitespaces around the "=" sign in function arguments (unless the argument is type-annotated) - the rule of thumb is to try to make newly added code look similar to existing code;
but this specific recommendation is also part of [PEP-8](https://www.python.org/dev/peps/pep-0008/#other-recommendations)

electrum/simple_config.py
```diff
@@ -659,14 +659,15 @@ def get_netaddress(self, key: str) -> Optional[NetAddress]:
...
         return format_satoshis(
             x,
             num_zeros=self.num_zeros,
             decimal_point=self.decimal_point,
             is_diff=is_diff,
             whitespaces=whitespaces,
             precision=self.amt_precision_post_satoshi,
+            add_thousands_sep = add_thousands_sep,
```
SomberNight, line 670: nit: remove whitespaces around "="

electrum/util.py
```diff
@@ -681,7 +682,14 @@ def format_satoshis(
     integer_part, fract_part = result.split(".")
     if len(fract_part) < num_zeros:
         fract_part += "0" * (num_zeros - len(fract_part))
-    result = integer_part + DECIMAL_POINT + fract_part
+    # add whitespaces as thousands' separator for better readability of numbers
+    if add_thousands_sep:
+        sign = integer_part[0]
```
SomberNight, line 687: what if `is_diff==False` and `x>=0`? then this will be the first digit, not a plus sign.
```
>>> from electrum import util
>>> util.format_satoshis(5, add_thousands_sep=True)
'00.000 000 05'
```

electrum/util.py
SomberNight: yes, IIRC that is the main use case

SomberNight (changes requested):
Again, please add unit tests.
They would have helped you find a bug during this iteration too.
See comments.

dhruv-joshi-7:
Made the changes just wanted to know how do I test the `is_diff` with a `'-'` sign.

SomberNight:
> just wanted to know how do I test the `is_diff` with a `'-'` sign.

What do you mean?
Do you mean how to write a unit test for it, or how to test it manually?

Re manual, you can fire up a python interpreter and call it.
See the example I shared above:
```
>>> from electrum import util
>>> util.format_satoshis(5, add_thousands_sep=True)
'00.000 000 05'
```

Re unit tests, just see how the existing tests are written:
https://github.com/spesmilo/electrum/blob/a8ecc68833b6a64bac614a61736f621dc55aeb7b/electrum/tests/test_util.py#L70-L72

dhruv-joshi-7:
```
sign = integer_part[0] if integer_part[0] == "+" or integer_part[0] == "-" else ""
if sign == "-":
     integer_part = integer_part[1:]
integer_part = "{:,}".format(int(integer_part)).replace(',', " ")
integer_part = sign + integer_part
```


I have made the above changes in the code( in the if `add_thousands_sep` block), all tests are passing with this code. And I also made a new unit test which is also passing with the above code. 
Should I change anything in this, and also should I commit the new unittest?

SomberNight:
> `sign = integer_part[0] if integer_part[0] == "+" or integer_part[0] == "-" else ""`
> Should I change anything in this [...]?

I find it a bit unsatisfying to have to add four new code lines just to handle the sign in the `is_diff` & `add_thousands_sep` case;
however I can't really see a nicer solution atm either.
At least simplify the first line to:
`sign = integer_part[0] if integer_part[0] in ("+", "-") else ""`


> should I commit the new unittest?

Yes, please commit your new changes and the unit tests as well.
It is also significantly easier to give feedback that way because reviewers can then use the standard interface.

dhruv-joshi-7:
Please review the changes!

electrum/tests/test_readability.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Test_Satoshi_Readability(ElectrumTestCase):
+
+    def test_format_satoshis_readability(self):
+        self.assertEqual("178 890 000.", format_satoshis(Decimal(178890000), decimal_point=0, add_thousands_sep=True))
+        self.assertEqual("458 312.757 48", format_satoshis(Decimal("45831275.748"), decimal_point=2, add_thousands_sep=True, precision=5))
+        self.assertEqual("+4 583 127.574 8", format_satoshis(Decimal("45831275.748"), decimal_point=1, is_diff=True, add_thousands_sep=True, precision=4))
+        self.assertEqual("+456 789 112.004 56", format_satoshis(Decimal("456789112.00456"), decimal_point=0, is_diff=True, add_thousands_sep=True, precision=5))
+        self.assertEqual("-0.00001234", format_satoshis(-1234, is_diff=True)) 
+        self.assertEqual("-456789.00001234", format_satoshis(-45678900001234, is_diff=True))
```
SomberNight, line 20: These two asserts already appear in another unit test:
https://github.com/spesmilo/electrum/blob/07219cc4c6435011afacda546f652cea91d98cd4/electrum/tests/test_util.py#L71-L72

Please do not duplicate them as that is pointless.

electrum/simple_config.py
SomberNight: Again, please remove these whitespaces as well.

electrum/tests/test_readability.py
```diff
@@ -0,0 +1,22 @@
...
+                           format_satoshis_plain, is_private_netaddress, is_hex_str,
+                           is_integer, is_non_negative_integer, is_int_or_float,
+                           is_non_negative_int_or_float)
+
+from . import ElectrumTestCase
+
+
+class Test_Satoshi_Readability(ElectrumTestCase):
+
+    def test_format_satoshis_readability(self):
```
SomberNight, line 14: Please move this test to `test_util.py/TestUtil`, next to the existing unit tests that exercise `format_satoshis`.
No need to create a new file just for this.

SomberNight: For the name of the function, I suggest: `test_format_satoshis_add_thousands_sep` instead.

SomberNight (changes requested):

SomberNight:
Btw, if you think you've dealt with feedback in a specific comment thread, mark it as resolved using the "Resolve conversation" button: this will collapse it and make it easier to see the outstanding feedback.

electrum/tests/test_util.py
```diff
@@ -79,7 +87,7 @@ def test_format_satoshis_plain_decimal(self):
 
     def test_format_satoshis_plain_to_mbtc(self):
         self.assertEqual("0.01234", format_satoshis_plain(1234, decimal_point=5))
-
+
```
SomberNight, line 90: please remove whitespace change

SomberNight (changes requested):
Looks good now! :)

electrum/tests/test_util.py
dhruv-joshi-7: I think it showing that because I hit backspace accidentally while making those changes and hit enter to keep the same spacing. The spacing is the same, there are no changes.

SomberNight: This line was previously empty but now it contains some spaces. It is a minor thing but still annoying.

dhruv-joshi-7: Okay got it, thanks :)

SomberNight (approved):
Looks good. Thank you.

I've found a small issue about how `add_thousands_sep` and `whitespaces` interact, but I will fix that in a follow-up.

SomberNight:
So I've added a follow-up that fixed the `whitespaces` issue (amounts were not correctly aligned in columns sometimes): https://github.com/spesmilo/electrum/commit/9f5370f9c622c89894daedf3af87272f51823495

Also, I've turned this into an option (so that it can be turned on/off from Preferences) in https://github.com/spesmilo/electrum/commit/e7ccf1584ea4ff07f9de2c699e6a360023d058bb

dhruv-joshi-7:
> Also, I've turned this into an option (so that it can be turned on/off from Preferences) in [e7ccf15](https://github.com/spesmilo/electrum/commit/e7ccf1584ea4ff07f9de2c699e6a360023d058bb)

Looks neat :)

PR #7431, opened 2021-07-25
Add Cirrus CI

JeremyRand:
As discussed on IRC, this PR adds Cirrus CI.

To see what the Cirrus results look like, see https://github.com/JeremyRand/electrum-nmc/runs/3155036786

Some random notes:

1. A few of the Lightning unit tests fail nondeterministically.  I haven't looked at this too closely, but I don't think this is caused by the Cirrus environment.
2. The functional regtests fail to break out of the "wait until Alice sees channel open" loop.  Again, I didn't look at this very closely.
3. I had to refactor the scripts that launch Bitcoin Core and ElectrumX for the functional tests.  One benefit of the refactor is that Cirrus shows the logs for each in a separate log section, which is helpful for debugging when the functional tests fail.
4. I had to slightly tweak the `docker build` command for the Android build script, so that it can access a file in a parent directory of the Dockerfile.   Note that the Cirrus docs indicate that the contents of `COPY` files are not taken into account when keying the Docker image cache.  I *think* that just means that you shouldn't configure Cirrus to allow the public to write to the cache (i.e. leave at default settings).  If a malicious actor did write to the cache, e.g. by submitting a malicious PR, they could change the contents of the "requirements" file that's cached in that Docker image.
5. Cirrus is configured to use the Dockerfiles directly, rather than running the `build.sh` scripts in a hardware VM.  This makes things much simpler and much faster, but does mean that if there's a Bash error in those scripts, Cirrus won't catch it.
6. Cirrus uploads artifacts of all the binary builds.  In many cases, these may be bit-for-bit identical to what manual builds produce, in which case you could hypothetically use Cirrus as a co-signer for reproducible builds.  The artifacts may also be useful for QA testers who want to try a nightly binary without building from source (obviously this should not be done with mainnet wallets with significant amounts of money, unless the QA testers trust Cirrus).  The Cirrus API can give you a URL that will always point to the latest build artifact that passed all tests.
7. I didn't test the submodule tests, since they only run for tags.
8. Cirrus runs the unit tests with PyPy too, but doesn't fail the build if they fail.  This is mainly to make it easier for people who want to improve PyPy compatibility to see what the status is.
9. Similarly, Cirrus runs additional flake8 tests, but doesn't fail the build if they fail.  The reason is the same: it makes it easier for people who want to improve that situation to see the status.
10. This PR also includes two trivial regtest fixes relating to output buffering.  Without these fixes, the regtests appear to hang until the regtests finish.

Let me know if anything looks obviously wrong.

### DIFF
--- a/contrib/android/build.sh
+++ b/contrib/android/build.sh
@@ -22,13 +22,11 @@ if [ ! -z "$ELECBUILD_NOCACHE" ] ; then
 fi
 
 info "building docker image."
-cp "$CONTRIB/deterministic-build/requirements-build-android.txt" "$CONTRIB_ANDROID/requirements-build-android.txt"
 sudo docker build \
     $DOCKER_BUILD_FLAGS \
     -t electrum-android-builder-img \
     --file "$CONTRIB_ANDROID/Dockerfile" \
     "$PROJECT_ROOT"
-rm "$CONTRIB_ANDROID/requirements-build-android.txt"
 
 
 # maybe do fresh clone
--- a/contrib/android/build.sh
+++ b/contrib/android/build.sh
@@ -26,7 +26,8 @@ cp "$CONTRIB/deterministic-build/requirements-build-android.txt" "$CONTRIB_ANDRO
 sudo docker build \
     $DOCKER_BUILD_FLAGS \
     -t electrum-android-builder-img \
-    "$CONTRIB_ANDROID"
+    --file "$CONTRIB_ANDROID/Dockerfile" \
+    "$PROJECT_ROOT"
 rm "$CONTRIB_ANDROID/requirements-build-android.txt"
 
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -158,3 +158,14 @@ task:
     path: "contrib/build-wine/dist/*"
   env:
     CIRRUS_WORKING_DIR: /opt/wine64/drive_c/electrum
+
+task:
+  name: Android build
+  container:
+    dockerfile: contrib/android/Dockerfile
+    cpu: 2
+    memory: 2G
+  build_script:
+    - ./contrib/android/make_apk
+  binaries_artifacts:
+    path: "dist/*"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -76,3 +76,31 @@ task:
   depends_on:
     - Tox Python 3.9
   only_if: $CIRRUS_BRANCH == 'master'
+
+task:
+  container:
+    image: $ELECTRUM_IMAGE
+    cpu: 1
+    memory: 1G
+  pip_cache:
+    folder: ~/.cache/pip
+    fingerprint_script: echo Flake8 && echo $ELECTRUM_IMAGE && cat $ELECTRUM_REQUIREMENTS
+    populate_script: mkdir -p ~/.cache/pip
+  electrum_cache:
+    folder: /tmp/electrum-build
+    populate_script: mkdir -p /tmp/electrum-build
+  install_script:
+    - pip install flake8
+  flake8_script:
+    - flake8 . --count --select=$ELECTRUM_LINTERS --show-source --statistics
+  env:
+    ELECTRUM_IMAGE: python:3.7
+    ELECTRUM_REQUIREMENTS: contrib/requirements/requirements-travis.txt
+  matrix:
+    - name: Flake8 Mandatory
+      env:
+        ELECTRUM_LINTERS: E9,F63,F7,F82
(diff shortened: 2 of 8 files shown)


contrib/android/build.sh
```diff
@@ -26,7 +26,8 @@ cp "$CONTRIB/deterministic-build/requirements-build-android.txt" "$CONTRIB_ANDRO
 sudo docker build \
     $DOCKER_BUILD_FLAGS \
     -t electrum-android-builder-img \
-    "$CONTRIB_ANDROID"
+    --file "$CONTRIB_ANDROID/Dockerfile" \
+    "$PROJECT_ROOT"
 rm "$CONTRIB_ANDROID/requirements-build-android.txt"
```
SomberNight, line 31: I guess with this change the `cp` and `rm` lines are not needed anymore?

JeremyRand: > I guess with this change the `cp` and `rm` lines are not needed anymore?

@SomberNight I think that's correct.



> Note that a side-effect of this change is sending a potentially much bigger build context to docker. E.g. consider having build caches for all the different builds on your local machine; all that would now be included.
> Couldn't we use instead the same `cp` approach in the cirrus yml?

The issue here is that Cirrus's automatic Dockerfile-to-container-image generator only supports using the repo root as the `PATH` argument to `docker build`.  Or at least, if Cirrus supports custom `PATH` arguments, it isn't documented.  I'll check with Cirrus support and see if there's a good workaround for this.

JeremyRand: Okay, so, Cirrus implemented support for custom Docker contexts, and it works.  However, there is another hitch: Cirrus does not support running custom scripts before it builds the Docker container from the Dockerfile.  So, we can't run `cp` to copy `requirements-build-android.txt` from `contrib/deterministic-build` to `contrib/android`.  Symlinks do not work as a workaround.  I see a few options:

1. Use `contrib/deterministic-build` as the build context, since the only file being copied into the container is in that directory.  Might constrain future changes to the Dockerfile.
2. Use `contrib` as the build context, since that includes both `android` and `deterministic-build`.  Bigger context to cache than `contrib/android`, but better than the repo root.
3. Manually check the results of `cp` into Git every time `requirements-build-android.txt` is updated.  Seems error-prone and a pain to audit.
4. Ask Cirrus to support custom hook scripts in their automatic docker_builder VM's.

My preference is to do (1) as a short-term solution, and pursue (4) as a longer-term solution that doesn't block this PR.  @SomberNight does that approach work for you?

SomberNight: > 1. Use `contrib/deterministic-build` as the build context, since the only file being copied into the container is in that directory.  Might constrain future changes to the Dockerfile.

Does this work? My understanding was that the Dockerfile must be located somewhere inside the build context.
This is also stated [in the docs](https://docs.docker.com/engine/reference/commandline/build/#text-files):
> By default the docker build command will look for a Dockerfile at the root of the build context. The -f, --file, option lets you specify the path to an alternative file to use instead. This is useful in cases where the same set of files are used for multiple builds. The path must be to a file within the build context.

SomberNight: Ok... how about we do `ln -s .gitignore .dockerignore`, commit `.dockerignore` into the repo, and use `$PROJECT_ROOT` as the build context for the android build?

JeremyRand: > Does this work? My understanding was that the Dockerfile must be located somewhere inside the build context.

It definitely looks like it worked, see https://github.com/JeremyRand/electrum-nmc/commit/9f224ae1cc3acf41e94f226e28a8ba3540a31406 (the Android Docker build on Cirrus succeeded for that commit).  I don't know if maybe that's relying on undefined behavior and I just got lucky though.  :/

> Ok... how about we do `ln -s .gitignore .dockerignore`, commit `.dockerignore` into the repo, and use `$PROJECT_ROOT` as the build context for the android build?
> (related: [126afeb](https://github.com/spesmilo/electrum/commit/126afebea1cd774d6f610dbefe84348ed9f278a3))

Ah, that's a clever trick.  I have no objection to this if you're okay with it.  I will make that change and verify that it still works, and then update this PR if all looks good.

JeremyRand:
@SomberNight PR updated.  Should merge cleanly as-is, but let me know if you want me to rebase against master to make it easier to test the `.dockerignore` trick.

SomberNight:
The code looks good.  I've tested some of the changes locally.
I have tried to configure Cirrus CI for this repository. Let's see if it runs.

Thank you for your work!

PR #7433, opened 2021-07-26
lnrouter: fix self-payments

bitromortac:
Previously pathfinding returned `NoPathFound` exceptions for circular pathfinding, mainly because `nodeA` and `nodeB` were not considered to potentially be equal.

Note that I removed the `is_mine` checks and added an alternative code block, as in the meantime the meaning of `my_channels` has changed within the `create_routes_for_payment` scope. Right now it is interpreted like "these are the channels over which I want payments to go out because they are active and can send something", so they will not include receiving-only channels.

The effect of the removal of `is_mine` checks is less reliability on the receiving side, but it decouples concerns (we have liquidity hints which take care of this), but alternatively we could also restore the old `my_channels` meaning, which is an open question on my end. After this is sorted out, I would also rename `my_channels` to `my_sending_channels`.

### DIFF
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -25,13 +25,13 @@
 
 import queue
 from collections import defaultdict
-from typing import Sequence, List, Tuple, Optional, Dict, NamedTuple, TYPE_CHECKING, Set
+from typing import Sequence, Tuple, Optional, Dict, TYPE_CHECKING, Set
 import time
 from threading import RLock
 import attr
 from math import inf
 
-from .util import bh2u, profiler, with_lock
+from .util import profiler, with_lock, bh2u
 from .logging import Logger
 from .lnutil import (NUM_MAX_EDGES_IN_PAYMENT_PATH, ShortChannelID, LnFeatures,
                      NBLOCK_CLTV_EXPIRY_TOO_FAR_INTO_FUTURE)
@@ -511,7 +511,7 @@ class LNPathFinder(Logger):
         overall_cost = fee_msat + cltv_cost + liquidity_penalty
         return overall_cost, fee_msat
 
-    def get_distances(
+    def get_shortest_path_hops(
             self,
             *,
             nodeA: bytes,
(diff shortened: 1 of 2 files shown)


electrum/lnrouter.py
```diff
@@ -554,14 +566,6 @@ def get_distances(
                     continue
                 edge_startnode = channel_info.node2_id if channel_info.node1_id == edge_endnode else channel_info.node1_id
                 is_mine = edge_channel_id in my_channels
-                if is_mine:
-                    if edge_startnode == nodeA:  # payment outgoing, on our channel
-                        if not my_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
-                            continue
```
SomberNight, line 560: Are you sure it is a good idea to remove this block?
Removing the stuff directly below is only about self-payments so I am not that concerned about reducing reliability there.
However this block highlighted affects almost all payments and it being there might make attempts more reliable.

electrum/lnrouter.py
```diff
@@ -529,22 +529,34 @@ def get_distances(
         blacklist = self.liquidity_hints.get_blacklist()
         distance_from_start = defaultdict(lambda: float('inf'))
         distance_from_start[nodeB] = 0
-        prev_node = {}  # type: Dict[bytes, PathEdge]
+        previous_hops = {}  # type: Dict[bytes, PathEdge]
```
SomberNight, line 532: why the renaming?
are you trying to reflect that the dict values are edges ("hops"?) instead of nodes?

electrum/lnrouter.py
bitromortac: Yes, you're right, I was too eager there for the outgoing channels. I was a bit mislead, as `active_channels` in `create_routes_for_payment` only selects active sending channels, but does not take into account sending amounts (they would be taken into account with mpp-presplitting). What do you think about what `my_channels` should contain, all our channels or only the sending ones?

electrum/lnrouter.py
bitromortac: Yes, when first looking at this part of the code the naming was very confusing to me, especially the singular `node`, but I'm not 100% sure how to name it. What would be your interpretation of `prev_node`? Related is also the naming of the method itself.

SomberNight: This dict is the analogue of the `prev` map in the [wikipedia Dijkstra pseudocode](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm#Pseudocode). `prev[v] == u` means that the cheapest path from the start_node to `v` has the `u->v` edge as its last edge; and using the `prev` map you can construct a path from the start_node to any `v` via simple backtracking through `prev`. So in the original pseudocode the map stores the previous node on the cheapest path to a given node. In our code the value is the `u->v` edge instead of the `u` node.

electrum/lnrouter.py
SomberNight: Either is fine. If there is a benefit in passing only a subset of our channels as `my_channels`, we can do that.
If we do, either we should rename the parameter, or add a docstring/comment to explain what is being passed.

electrum/lnrouter.py
SomberNight: > also the word prev could be confusing.

Hmm, what other meaning do you see? :)
Maybe I have tunnel vision here; especially as it was probably I who named this variable...

bitromortac: Ok, thanks for the explainer. I think the confusion just came from the directionality in connection with the word previous, as we search from the end node and go back to the start node. But it's difficult to explain why oneselve's confused :).

SomberNight: well ok, I don't mind either naming; so we can rename it if you think it's clearer now

SomberNight:
Thanks!

Issue #7408, opened 2021-07-14
Implement more SIGHASH types besides SIGHASH_ALL


SomberNight:
When spending a UTXO, the input spending it usually contains signatures (e.g. to satisfy an OP_CHECKSIG opcode) using a pubkey that the UTXO committed to.
The specific message the signature covers (the UTXO being spent, the rest of the inputs/outputs of the spending txs) depends on the sighash flags (which are chosen by the party creating the signature). Most transactions on Bitcoin mainnet use the SIGHASH_ALL flag, in which case the signature covers ~all of the spending transaction.

Other sighash flags e.g. allow not-invalidating the tx when adding more inputs, as a signature created for a specific input could simply *not* commit to what any of the other tx inputs are.

We currently only implement signing with SIGHASH_ALL.
Just grep [`transaction.py`](https://github.com/spesmilo/electrum/blob/ea24ab593f6f8fa5069e7fcb880b4c11cee4f5c5/electrum/transaction.py) for `SIGHASH_ALL`.

However, we should implement signing/verifying with the other sighash flags besides SIGHASH_ALL.

- Note that combinations of flags is also possible, such as `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY`.
- Note that the sighash flags and the construction of the preimage (what is being signed) differs when spending legacy inputs and segwit inputs (and also when spending taproot inputs).
  I think legacy inputs are not worth caring about.
  The segwit sighash behaviour is documented in [bip-0143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki).
  The taproot sighash behaviour is also out of scope for now (until we implement spending from taproot utxos at least).
- Our current code supports SIGHASH_ALL for the legacy input case, and SIGHASH_ALL for the segwit input case; nothing else atm.
- We will need at least to be able to sign/verify signatures with the `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` flags to be able to implement [anchor outputs for LN](https://github.com/spesmilo/electrum/issues/6588).

So what I suggest, for now, is implementing support for all sighash flags for the segwit input case.

also see https://en.bitcoin.it/wiki/OP_CHECKSIG

MorenoProg:
[...]

PR #7453, opened 2021-08-12
Added different Sighash_Types

dhruv-joshi-7:
Implements https://github.com/spesmilo/electrum/issues/7408


dhruv-joshi-7:
Please note that the tests for `SIGHASH_ALL|ANYONECANPAY` and `SIGHASH_NONE|ANYONECANPAY` are failing. 
I have verified that the preimage is correct, probably there might be a change in the serialization method for these flags.

dhruv-joshi-7:
Note that the original tests were working after I tested them while commenting the grinding `r` `while loop` in `ecc.py`

electrum/tests/test_transaction.py
```diff
@@ -917,3 +917,145 @@ def test_spending_op_cltv_p2wsh(self):
                          tx.txid())
         self.assertEqual('020000000001019ad573c69e60c209e0ff36f281ae4f700a8d59f846e7ff5c020352fd1e97808600000000000000000001fa840100000000001600145a209b202bc19b3d345a75cf8ab51cb471913a790247304402207b191c1e3ff1a2d3541770b496c9f871406114746b3aa7347ec4ef0423d3a975022043d3a746fa7a794d97e95d74b6d17d618dfc4cd7644476813e08006f271e51bd012a046c4f855fb1752102aec53aa5f347219a7378b13006eb16ce48125f9cf14f04a5509a565ad5e51507ac6c4f855f',
                          tx.serialize())
+
+class Test_Sighash_Types(ElectrumTestCase):
+    #These tests are taken from bip143, https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki
+    def check_sighash_types_sighash_all(self):
+        #Input of transaction
+        locktime=0
```
bitromortac, line 925: There's lots of code repetition, it would be cool if you could define the transaction in one place and then only change the sighash types.

bitromortac (commented):
Great work! Just wanted to give some feedback on minor things. Is there a reason you didn't include the other test vectors? I think those would be good to have. So if I understand you well, you replaced the signatures from the test vectors in the `SIGHASH_ALL|ANYONECANPAY` and `SIGHASH_NONE|ANYONECANPAY` cases with the ones from non-grinded r values. I can confirm that the preimages are correct.

dhruv-joshi-7:
> Great work! Just wanted to give some feedback on minor things. Is there a reason you didn't include the other test vectors? I think those would be good to have. So if I understand you well, you replaced the signatures from the test vectors in the `SIGHASH_ALL|ANYONECANPAY` and `SIGHASH_NONE|ANYONECANPAY` cases with the ones from non-grinded r values. I can confirm that the preimages are correct.

The other test vectors contained the `OP_CODESEPARATOR` in the `witness script`. Hence they were not included.

electrum/tests/test_transaction.py
```diff
@@ -917,3 +917,76 @@ def test_spending_op_cltv_p2wsh(self):
...
+    #Input of transaction
+    locktime=0
+    prevout = TxOutpoint(txid=bfh('6eb98797a21c6c10aa74edf29d618be109f48a8e94c694f3701e08ca69186436'), out_idx=1)
+    txin = PartialTxInput(prevout=prevout)
+    txin.nsequence=0xffffffff
+    txin.script_type='p2sh-p2wsh'
+    witness_script = bfh('56210307b8ae49ac90a048e9b53357a2354b3334e9c8bee813ecb98e99a7e07e8c3ba32103b28f0c28bfab54554ae8c658ac5c3e0ce6e79ad336331f78c428dd43eea8449b21034b8113d703413d57761b8b9781957b8c0ac1dfe69f492580ca4195f50376ba4a21033400f6afecb833092a9a21cfdf1ed1376e58c5d1f47de74683123987e967a8f42103a6d48b1131e94ba04d9737d61acdaa1322008af9602b3b14862c07a1789aac162102d8b661b0b3302ee2f162b09e07a55ad5dfbe673a9f01d9f0c19617681024306b56ae')
+    txin.witness_script = witness_script
+    redeem_script = bfh('0020a16b5755f7f6f96dbd65f5f0d6ab9418b89af4b1f14a1bb8a09062c35f0dcb54')
+    txin.redeem_script = redeem_script
```
SomberNight, line 932: No need for the temp var `witness_script`. Just assign `txin.witness_script` directly and save one line.
Same for `redeem_script`.

electrum/tests/test_transaction.py
```diff
@@ -917,3 +917,76 @@ def test_spending_op_cltv_p2wsh(self):
...
+        self.assertEqual('3044022059ebf56d98010a932cf8ecfec54c48e6139ed6adb0728c09cbe1e4fa0915302e022007cd986c8fa870ff5d2b3a89139c9fe7e499259875357e20fcbb15571c76795403',
+                        sig)
+
+    def test_check_sighash_types_sighash_all_anyonecanpay(self):
+        self.txin.sighash=Sighash.SIGHASH_ALL|Sighash.SIGHASH_ANYONECANPAY
+        self.txin.pubkeys = [bfh('033400f6afecb833092a9a21cfdf1ed1376e58c5d1f47de74683123987e967a8f4')]
+        privkey = bfh('14af36970f5025ea3e8b5542c0f8ebe7763e674838d08808896b63c3351ffe49')
+        tx = PartialTransaction.from_io(inputs=[self.txin], outputs=[self.txout1,self.txout2], locktime=self.locktime, version=1, BIP69_sort=False)
+        sig = tx.sign_txin(0,privkey)
+        #Note that this test is modified to the new signature rules which were introduced after bip143.
```
SomberNight, line 971: This is a bit unclear...
I suggest:
```suggestion
        # Note that the expected sig differs from the one in BIP-0143, due to us grinding the R value. see https://github.com/bitcoin/bitcoin/pull/13666
```
(remember to modify other identical comment accordingly)

electrum/transaction.py
```diff
@@ -1844,24 +1854,42 @@ def get_fee(self) -> Optional[int]:
         except MissingTxInputAmount:
             return None
 
+    def is_flag_valid(self, sighash) -> bool:
```
SomberNight, line 1857: This should be a method of `Sighash`, with prototype/signature e.g.: 
`def is_valid(self) -> bool:`

electrum/transaction.py
```diff
@@ -88,7 +88,16 @@ class MissingTxInputAmount(Exception):
     pass
 
 
-SIGHASH_ALL = 1
+class Sighash(IntFlag):
+    SIGHASH_ALL = 1
+    SIGHASH_NONE = 2
+    SIGHASH_SINGLE = 3
+    SIGHASH_ANYONECANPAY = 0X80
```
SomberNight, line 95: I think we should just remove the `SIGHASH_` prefixes, as `Sighash.ALL` is clear enough (no need for `Sighash.SIGHASH_ALL`).

electrum/transaction.py
```diff
@@ -1844,24 +1854,42 @@ def get_fee(self) -> Optional[int]:
         except MissingTxInputAmount:
             return None
 
+    def is_flag_valid(self, sighash) -> bool:
+        for flag in Sighash:
+            for base_flag in BaseSighash:
```
SomberNight, line 1859: For this single usage, I don't think it's worth it to define `BaseSighash`.
You could just list the types directly here: `[Sighash.SIGHASH_ALL, Sighash.SIGHASH_NONE, Sighash.SIGHASH_SINGLE]`.

electrum/transaction.py
SomberNight: Please use lowercase `x` for hex literals: `0x80`.

electrum/transaction.py
```diff
@@ -88,7 +88,16 @@ class MissingTxInputAmount(Exception):
     pass
 
 
-SIGHASH_ALL = 1
+class Sighash(IntFlag):
```
SomberNight, line 91: I recommend `IntEnum` instead of `IntFlag` here.
Flag is very confusing here as only ANYONECANPAY is a true flag, there other values are not...
E.g. `ALL | NONE == SINGLE` <- but this does not make sense semantically.

electrum/transaction.py
```diff
@@ -1844,24 +1854,42 @@ def get_fee(self) -> Optional[int]:
...
         if txin.is_segwit():
             if bip143_shared_txdigest_fields is None:
                 bip143_shared_txdigest_fields = self._calc_bip143_shared_txdigest_fields()
-            hashPrevouts = bip143_shared_txdigest_fields.hashPrevouts
-            hashSequence = bip143_shared_txdigest_fields.hashSequence
-            hashOutputs = bip143_shared_txdigest_fields.hashOutputs
+            if not(sighash & Sighash.SIGHASH_ANYONECANPAY):
+                hashPrevouts = bip143_shared_txdigest_fields.hashPrevouts
+            else:
+                hashPrevouts = ('0'*64)
```
SomberNight, line 1882:
```suggestion
                hashPrevouts = '00' * 32
```
and same a few lines below.

electrum/transaction.py
```diff
@@ -1895,11 +1923,13 @@ def sign(self, keypairs) -> None:
     def sign_txin(self, txin_index, privkey_bytes, *, bip143_shared_txdigest_fields=None) -> str:
         txin = self.inputs()[txin_index]
         txin.validate_data(for_signing=True)
+        sighash = txin.sighash if txin.sighash is not None else Sighash.SIGHASH_ALL
+        sighash_type = '{0:02x}'.format(sighash)
```
SomberNight, line 1927: Maybe it's just personal preference, but I find this more clear:
```suggestion
        sighash_type = sighash.to_bytes(length=1, byteorder="big").hex()
```

electrum/transaction.py
dhruv-joshi-7: I did not understand how to implement it as a method with prototype/signature.  Right now, I have implemented it as a `classmethod`

SomberNight: By prototype/signature, I meant what was on the following line; the function name, input types, output types, etc.
See https://en.wikipedia.org/wiki/Function_prototype

So I would like something like this:
```python
class Sighash(IntEnum):
    ALL = 1
    NONE = 2
    SINGLE = 3
    ANYONECANPAY = 0x80

    def is_valid(cls, sighash) -> bool:
        ...
```
That is, `is_valid` does not need to be a classmethod; and instead of taking `sighash` as a param, just use `self`.

SomberNight: Ah, nevermind, this does not actually work with `Sighash(IntEnum)`, only with `Sighash(IntFlag)`.
(As, with IntEnum, `type(Sighash.ALL|Sighash.ANYONECANPAY) == int`).
So let's leave it as-is, as
```python
    @classmethod
    def is_valid(cls, sighash) -> bool:
```

electrum/transaction.py
```diff
@@ -88,7 +88,19 @@ class MissingTxInputAmount(Exception):
...
+    ALL = 1
+    NONE = 2
+    SINGLE = 3
+    ANYONECANPAY = 0x80
+
+    @classmethod
+    def is_valid(cls, sighash) -> bool:
+        for flag in Sighash:
+            for base_flag in [Sighash.ALL, Sighash.NONE, Sighash.SINGLE]:
+                if (flag & ~0x1f | base_flag) == sighash:
```
SomberNight, line 101: Where is this code from? Took me some time to understand what it is doing...

```python
>>> byt = int.to_bytes(~0x1f, byteorder="big", length=1, signed=True)
>>> int.from_bytes(byt, byteorder="big", signed=False)
224
>>> bin(224)
'0b11100000'
```

electrum/transaction.py
```diff
@@ -1844,24 +1857,42 @@ def get_fee(self) -> Optional[int]:
...
+            else:
+                hashPrevouts = '00' * 32
+            if (not(sighash & Sighash.ANYONECANPAY) and (sighash & 0x1f) != Sighash.SINGLE and (sighash & 0x1f) != Sighash.NONE):
+                hashSequence = bip143_shared_txdigest_fields.hashSequence
+            else:
+                hashSequence = '00' * 32
+            if ((sighash & 0x1f) != Sighash.SINGLE and (sighash & 0x1f) != Sighash.NONE):
+                hashOutputs = bip143_shared_txdigest_fields.hashOutputs
+            elif ((sighash & 0x1f) == Sighash.SINGLE and txin_index < len(outputs)):
+                hashOutputs = bh2u(sha256d(bfh(outputs[txin_index].serialize_to_network().hex())))
```
SomberNight, line 1893: This
`bfh(outputs[txin_index].serialize_to_network().hex())`
can be simplified to
`outputs[txin_index].serialize_to_network()`

electrum/transaction.py
dhruv-joshi-7: https://github.com/bitcoin-sv/bitcoin-sv/blob/master/src/script/sighashtype.h
I implemented a similar logic.

SomberNight: That link is not really reassuring though as it is for an altcoin's repository...
Does upstream bitcoin/bitcoin have the same logic?

bitromortac:
I managed to reproduce signatures for HTLC transaction test vectors for anchor outputs with ANYONECANSPEND|SINGLE, however only without r-value grinding. It would be great if we could reproduce the original test signatures here.

electrum/transaction.py
```diff
@@ -1844,24 +1857,42 @@ def get_fee(self) -> Optional[int]:
         except MissingTxInputAmount:
             return None
 
+    def is_flag_valid(self, sighash) -> bool:
+        for flag in Sighash:
+            for base_flag in BaseSighash:
+                if (flag & ~0x1f | base_flag) == sighash:
+                    return True
+        return False
```
SomberNight, line 1865: With `Sighash.is_valid`, this method can just be deleted.

SomberNight:
@bitromortac 

> I managed to reproduce signatures for HTLC transaction test vectors for anchor outputs with ANYONECANSPEND|SINGLE, however only without r-value grinding. It would be great if we could reproduce the original test signatures here.

So to be clear, you have the same issue with the anchor outputs test vectors as this PR has with bip143 test vectors: that the original test vectors do not use R value grinding but our codebase always does it now, right?

SomberNight:
@dhruv-joshi-7 @bitromortac re R value grinding, I've added a function decorator in https://github.com/spesmilo/electrum/commit/a39bfba2d9a2e7509bcc1638afa67b042c1b0868 that you can use in your tests.

@dhruv-joshi-7 Please use the new decorator in this PR instead of changing the test vectors. You should rebase this PR on master so that you have the referenced commit, and then in `test_transaction.py`, you can do `from .test_bitcoin import disable_ecdsa_r_value_grinding`, and then decorate the two respective test vectors with `@disable_ecdsa_r_value_grinding`.

bitromortac:
> @bitromortac
> 
> > I managed to reproduce signatures for HTLC transaction test vectors for anchor outputs with ANYONECANSPEND|SINGLE, however only without r-value grinding. It would be great if we could reproduce the original test signatures here.
> 
> So to be clear, you have the same issue with the anchor outputs test vectors as this PR has with bip143 test vectors: that the original test vectors do not use R value grinding but our codebase always does it now, right?

Yes, right.

> @dhruv-joshi-7 @bitromortac re R value grinding, I've added a function decorator in a39bfba that you can use in your tests.

Thank you!

SomberNight:
Looks good now, thanks!

PR #7464, opened 2021-08-19
Add network setup option when Kivy GUI starts for the First time. 

### LINKED ISSUE #5512: [Feature request] for Android app to have Networking options as the first form in first start (kivy, select server, network setup, before wallet creation)
Suppose you want to change the server and enable one server option, currently you have to create a temp wallet and go to settings and change the networking options. it would be very handy to have the network settings at the first view on the first run.
also when adding custom server's address it would be very helpful to have the scan QR code feature in case the address is rather long.

siddhantchawla:
This PR is an attempt to resolve #5512


SomberNight (changes requested):
Some issues/remarks:
1. if the user presses the back button (or `Esc` on keyboard) during `first_screen` or `initial_network_setup`, the app should exit. Currently it opens the main window with "no wallet".
    - one potential way to fix:
        ```
        was_cancelled: True
        on_dismiss: app.stop() if nd.was_cancelled else None
        ```
        add something like this to both Popups, and set `was_cancelled` just before calling `nd.dismiss()` - this way you can distinguish whether you explicitly called `.dismiss()` or if it got called due to the user aborting the dialog
        
2. the app should remember what the user selects re auto-connect. Specifically, on `first_screen`, if the user selects `Auto connect`, and then closes the app, we should not ask it again. Also, on `first_screen`, if the user selects `Select server manually`, and then on `initial_network_setup` they click `Next`, we should again make sure that we won't show the `first_screen` again on next app start. These scenarios currently are not handled and we end up showing `first_screen` on each start.
3. I know I had previously said using buttons on `first_screen` is fine (for `Auto connect` and `Select server manually`) but now looking at the dialog I think it is potentially too confusing for newbies and we should instead use radio buttons, with `Auto connect` pre-selected, and with a `Next` button they need to press.
    - See e.g. [the kivy docs re CheckBoxes](https://kivy.org/doc/stable/api-kivy.uix.checkbox.html) or [this tutorial](https://kivycoder.com/radio-buttons-for-kivy-python-kivy-gui-tutorial-29/)

SomberNight:
> if the user presses the back button (or `Esc` on keyboard) during `first_screen` or `initial_network_setup`, the app should exit. Currently it opens the main window with "no wallet".

This is still not fixed on `initial_network_setup` AFAICT.

> the app should remember what the user selects re auto-connect. Specifically, on `first_screen`, if the user selects `Auto connect`, and then closes the app, we should not ask it again. Also, on `first_screen`, if the user selects `Select server manually`, and then on `initial_network_setup` they click `Next`, we should again make sure that we won't show the `first_screen` again on next app start. 

You've changed the behaviour so that whatever the user selects on `first_screen` will get saved.
I am not sure if this is ideal behaviour.

Please read the quoted text again: I think if the user selects `Select server manually` on `first_screen` but then closes the app (either by accident or intentionally) before properly tweaking their settings on `initial_network_setup`, we should not persist their choice and let `first_screen` appear again when they start the app the next time.
The idea is that if the user wants to customise the network settings, but closes the app before they can finish doing so, we should offer them a chance to customise the settings again on next start.

That is, on `first_screen`, only persist `auto_connect` if the user selects `Auto connect`;
and, on `initial_network_setup`, persist `auto_connect` (regardless of value) when the user clicks `Next`.

SomberNight (approved):
Thanks. Looks good now.

I have some minor nits but will fix them myself in a follow-up commit to avoid holding this up.
(EDIT: done in https://github.com/spesmilo/electrum/commit/67f14cdfa899970ad69fa27b88662df498ef4d73)

PR #7477, opened 2021-08-29
Added pip requirements

backupbrain:
This `requirements.txt` allows users to use the following standard procedure to install dependencies

`$ pip3 install -r requirements`

This is what Python developers will expect when installing dependencies


SomberNight:
I don't quite understand what you are trying to do.

We have a set of list of dependencies with loose version bounds in `contrib/requirements/`,
and we have another set of lists with specific pinned versions used by our build scripts in `contrib/deterministic-build/`.

See e.g. https://github.com/spesmilo/electrum/blob/f62b23b210ab0ad7b91bd5a8a154040c5a008336/contrib/requirements/requirements.txt
and https://github.com/spesmilo/electrum/blob/f62b23b210ab0ad7b91bd5a8a154040c5a008336/contrib/deterministic-build/requirements.txt

Depending on whether you want pinned dependencies or not, and whether you need e.g. support for hardware wallets etc,
you can just do e.g.
`$ python3 -m pip install -r contrib/requirements/requirements.txt`
but note that that's already what setup.py specifies, so you might as well just do `$ pip install .`
(though I would suggest e.g. `$ python3 -m pip install --user -e .` if running from git clone).

Also, I don't know how you have arrived at your list (the one in this PR) but e.g. we stopped using `jsonrpclib-pelix` more than a year ago.

PR #7485, opened 2021-09-07
Fix tab order in Qt setup wizard

nyanpasu64:
Currently, in the multisig page of the wizard, the tab order is "bottom slider, top slider, back/next, and nothing (the entire scroll area)". This makes the scroll area non-focusable (on all pages) so the focus rectangle won't disappear once per Tab-key cycle, and (on the multisig page) constructs the top slider before the bottom slider so it comes first in the tab order.


SomberNight:
Looks reasonable. Thanks.

Issue #5507, opened 2019-07-12
Feature request: add Child Pays for Parent (CPFP) button to Kivy / Android


gits7r:
We can use replace by fee (RBF) for outgoing transactions in Kivy / Android. Then, we will have on that particular tx a "Bump fee" button displayed after we click "Details" on the transaction in cause to learn more about it. This is great and very useful.

I think it's useful and rational to add a Child pays for the Parent button on incoming transactions in Kivy / Adnroid, displayed the same after we click "Details" on an incoming transaction.

Currently there is no way to use child pay for the parent on Kivy / Android and it's annoying when you receive a tx with low fee because you cannot control to spend it if you have other UTXOs in your wallet, and the only way is to send max to some new address but this is ugly.

SomberNight:
done in https://github.com/spesmilo/electrum/pull/7487

PR #7487, opened 2021-09-08
Added CPFP Feature for Kivy GUI

dhruv-joshi-7:
implements https://github.com/spesmilo/electrum/issues/5507

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
+++ b/electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
@@ -1,0 +1,161 @@
+from typing import TYPE_CHECKING, Optional
+
+from kivy.app import App
+from kivy.factory import Factory
+from kivy.properties import ObjectProperty
+from kivy.lang import Builder
+
+from electrum.gui.kivy.i18n import _
+
+if TYPE_CHECKING:
+    from ...main_window import ElectrumWindow
+
+from .fee_dialog import FeeSliderDialog
+from electrum.gui.qt.amountedit import BTCAmountEdit
+from electrum.util import format_satoshis_plain
+
+Builder.load_string('''
+<CPFPDialog@Popup>
+    title: _('CPFP')
+    size_hint: 0.8, 0.8
+    pos_hint: {'top':0.9}
+    method: 0
+    BoxLayout:
+        orientation: 'vertical'
+        padding: '10dp'
+        GridLayout:
+            height: self.minimum_height
+            size_hint_y: None
+            cols: 1
+            spacing: '10dp'
+            BoxLabel:
+                id: msg1
+                text: _('CPFP stands for Child Pays for Parent')
+            BoxLabel:
+                id: total_size
+                text: _('Total Size')
+                value: ''
+            BoxLabel:
+                id: input_amount
+                text: _('Input amount')
+                value: ''
+            BoxLabel:
+                id: output_amount
+                text: _('Output amount')
+                value: ''
+            BoxLabel:
+                id: fee_for_child
+                text: _('Fee for child')
+                value: ''
+        BoxLayout:
+            orientation: 'horizontal'
+            size_hint: 1, 0.5
+            Label:
+                text: _('Target') + ' (%s):' % (_('mempool') if root.method == 2 else _('ETA') if root.method == 1 else _('static'))
+            Button:
+                id: fee_target
+                text: ''
+                background_color: (0,0,0,0)
+                bold: True
+                on_release:
+                    root.method  = (root.method + 1) % 3
+                    root.update_slider()
+                    root.on_slider(root.slider.value)
+
+        Slider:
+            id: slider
+            range: 0, 4
+            step: 1
+            on_value: root.on_slider(self.value)
+        BoxLayout:
+            orientation: 'horizontal'
+            size_hint: 1, 0.2
+            Label:
+                text: _('Final')
+            CheckBox:
+                id: final_cb
+        GridLayout:
+            height: self.minimum_height
+            size_hint_y: None
+            cols: 1
+            spacing: '10dp'
+            BoxLabel:
+                id: total_fee
+                text: _('Total fee')
+                value: ''
+            BoxLabel:
+                id: total_feerate
+                text: _('Total feerate')
+                value: ''
+        Widget:
+            size_hint: 1, 1
+        BoxLayout:
+            orientation: 'horizontal'
+            size_hint: 1, 0.5
+            Button:
+                text: 'Cancel'
+                size_hint: 0.5, None
+                height: '48dp'
+                on_release: root.dismiss()
+            Button:
+                text: 'OK'
+                size_hint: 0.5, None
+                height: '48dp'
+                on_release:
+                    root.dismiss()
+                    root.on_ok()
+''')
+
+class CPFPDialog(FeeSliderDialog, Factory.Popup):
+
+    def __init__(self, app: 'ElectrumWindow', parent_fee, total_size, new_tx, callback):
+        self.parent_fee = parent_fee
+        self.total_size = total_size
+        self.new_tx = new_tx
+        self.max_fee = self.new_tx.output_value()
+        Factory.Popup.__init__(self)
+        FeeSliderDialog.__init__(self, app.electrum_config, self.ids.slider)
+        self.app = app
+        self.callback = callback
+        self.config = app.electrum_config
+        self.ids.total_size.value = ('%d bytes'% self.total_size)
+        self.ids.input_amount.value = self.app.format_amount(self.max_fee) + ' ' + self.app._get_bu()
+        suggested_feerate = self.config.fee_per_kb()
+        self.fee = self.get_child_fee_from_total_feerate(suggested_feerate)
+        self.fee_e = self.app.format_amount_and_units(self.fee)
+        self.ids.fee_for_child.value = self.fee_e
+        self.update_slider()
+        self.update_text()
+
+    def get_child_fee_from_total_feerate(self, fee_per_kb: Optional[int]) -> Optional[int]:
+        if fee_per_kb is None:
+            return None
+        fee = fee_per_kb * self.total_size / 1000 - self.parent_fee
+        fee = round(fee)
+        fee = min(self.max_fee, fee)
+        fee = max(self.total_size, fee)  # pay at least 1 sat/byte for combined size
+        return fee
+
+    def update_text(self):
+        target, tooltip, dyn = self.config.get_fee_target()
+        self.ids.fee_target.text = target
+        fee_per_kb = self.config.fee_per_kb()
+        self.fee = self.get_child_fee_from_total_feerate(fee_per_kb)
+        if fee_per_kb is None:
+            self.ids.fee_for_child.value = "unknown"
+        else:
+            comb_fee = self.fee + self.parent_fee
+            comb_feerate = comb_fee / self.total_size 
+            self.ids.fee_for_child.value = f'{self.fee:.1f} sat'
+            self.ids.output_amount.value = f'{self.max_fee-self.fee:.1f} sat'
+            self.ids.total_fee.value = f'{self.fee+self.parent_fee:.1f} sat'
+            self.ids.total_feerate.value = f'{comb_feerate:.1f} sat/B'
+
+    def on_ok(self):
+        fee = self.fee
+        is_final = self.ids.final_cb.active
+        self.callback(fee, self.max_fee, is_final)
+
+    def on_slider(self, value):
+        self.save_config()
+        self.update_text()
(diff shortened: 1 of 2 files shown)


electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+            range: 0, 4
+            step: 1
+            on_value: root.on_slider(self.value)
+        BoxLayout:
+            orientation: 'horizontal'
+            size_hint: 1, 0.2
+            Label:
+                text: _('Final')
+            CheckBox:
+                id: final_cb
```
SomberNight, line 76: Let's just remove the "final" checkbox. You are not even using the value of the checkbox atm. :P
(The Qt GUI does not have a checkbox either. It's ok to always opt-in to RBF for CPFP txs, and this is what happens already.)

Make sure to also remove related code, e.g. its value is passed to the callback currently.

```suggestion
```

electrum/gui/kivy/uix/dialogs/tx_dialog.py
```diff
@@ -192,6 +194,7 @@ def update_action_dropdown(self):
             ActionButtonOption(text=_('Sign'), func=lambda btn: self.do_sign(), enabled=self.can_sign),
             ActionButtonOption(text=_('Broadcast'), func=lambda btn: self.do_broadcast(), enabled=self.can_broadcast),
             ActionButtonOption(text=_('Bump fee'), func=lambda btn: self.do_rbf(), enabled=self.can_rbf),
+            ActionButtonOption(text=_('CPFP'), func=lambda btn: self.do_cpfp(), enabled=self.can_cpfp),
```
SomberNight, line 197:
```suggestion
            ActionButtonOption(text=_('Child pays\nfor parent'), func=lambda btn: self.do_cpfp(), enabled=self.can_cpfp),
```

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+        orientation: 'vertical'
+        padding: '10dp'
+        GridLayout:
+            height: self.minimum_height
+            size_hint_y: None
+            cols: 1
+            spacing: '10dp'
+            BoxLabel:
+                id: msg1
+                text: _('CPFP stands for Child Pays for Parent')
```
SomberNight, line 33: You should use e.g. `TopLabel` instead of `BoxLabel`. `BoxLabel` was made for text-value pairs, so it ~halves the available horizontal space if you use it for just text.
Also, note that `id` is unused.

```suggestion
            TopLabel:
                text:
                    _(\
                    "A CPFP is a transaction that sends an unconfirmed output back to "\
                    "yourself, with a high fee. The goal is to have miners confirm "\
                    "the parent transaction in order to get the fee attached to the "\
                    "child transaction.")
```

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+if TYPE_CHECKING:
+    from ...main_window import ElectrumWindow
+
+from .fee_dialog import FeeSliderDialog
+from electrum.gui.qt.amountedit import BTCAmountEdit
+from electrum.util import format_satoshis_plain
+
+Builder.load_string('''
+<CPFPDialog@Popup>
+    title: _('CPFP')
```
SomberNight, line 19: Why not write out the full name? There is enough space for it AFAICT.

```suggestion
    title: _('Child Pays for Parent')
```

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+        fee = min(self.max_fee, fee)
+        fee = max(self.total_size, fee)  # pay at least 1 sat/byte for combined size
+        return fee
+
+    def update_text(self):
+        target, tooltip, dyn = self.config.get_fee_target()
+        self.ids.fee_target.text = target
+        fee_per_kb = self.config.fee_per_kb()
+        self.fee = self.get_child_fee_from_total_feerate(fee_per_kb)
+        if fee_per_kb is None:
```
SomberNight, line 144: It would make sense to test `self.fee` here instead as that is what gets used in the branches.

```suggestion
        if self.fee is None:
```

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+            self.ids.fee_for_child.value = "unknown"
+        else:
+            comb_fee = self.fee + self.parent_fee
+            comb_feerate = comb_fee / self.total_size 
+            self.ids.fee_for_child.value = f'{self.fee:.1f} sat'
+            self.ids.output_amount.value = f'{self.max_fee-self.fee:.1f} sat'
+            self.ids.total_fee.value = f'{self.fee+self.parent_fee:.1f} sat'
+            self.ids.total_feerate.value = f'{comb_feerate:.1f} sat/B'
```
SomberNight, line 152: Please don't format amounts/feerates by hand but use the existing helpers.

```suggestion
            comb_feerate = 1000 * comb_fee / self.total_size
            self.ids.fee_for_child.value = self.app.format_amount_and_units(self.fee)
            self.ids.output_amount.value = self.app.format_amount_and_units(self.max_fee-self.fee)
            self.ids.total_fee.value = self.app.format_amount_and_units(self.fee+self.parent_fee)
            self.ids.total_feerate.value = self.app.format_fee_rate(comb_feerate)
```
Note that in order to be able to reference `self.app` here; in `__init__`, you might have to move its assignment (`self.app = app`) to a few lines earlier.

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def on_ok(self):
+        fee = self.fee
+        is_final = self.ids.final_cb.active
+        self.callback(fee, self.max_fee, is_final)
+
+    def on_slider(self, value):
+        self.save_config()
+        self.update_text()
```
SomberNight, line 161: The file should end with a newline.
(i.e. add an empty last line here)

electrum/gui/kivy/uix/dialogs/tx_dialog.py
```diff
@@ -250,6 +253,42 @@ def _do_rbf(
...
+            return
+        cb = partial(self._do_cpfp, parent_tx=parent_tx)
+        d = CPFPDialog(self.app, parent_fee, total_size, new_tx=new_tx, callback=cb)
+        d.open()
+
+    def _do_cpfp(
+            self,
+            fee,
+            max_fee,
+            is_final,
```
SomberNight, line 275: Note that `is_final` is not used.
As said in another comment, let's just remove the checkbox from the UI and the related code.

```suggestion
```

electrum/gui/kivy/uix/dialogs/tx_dialog.py
```diff
@@ -250,6 +253,42 @@ def _do_rbf(
...
+            *,
+            parent_tx: Transaction
```
SomberNight, line 277: (nit) In case a function definition has parameters listed one per line as is done here, it's best to add a comma even after the last parameter. This ensures that if a new parameter is added later, this line will not have to be touched, and so it minimises the diff and potential merge conflicts.

```suggestion
            parent_tx: Transaction,
```

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -0,0 +1,161 @@
...
+        FeeSliderDialog.__init__(self, app.electrum_config, self.ids.slider)
+        self.app = app
+        self.callback = callback
+        self.config = app.electrum_config
+        self.ids.total_size.value = ('%d bytes'% self.total_size)
+        self.ids.input_amount.value = self.app.format_amount(self.max_fee) + ' ' + self.app._get_bu()
+        suggested_feerate = self.config.fee_per_kb()
+        self.fee = self.get_child_fee_from_total_feerate(suggested_feerate)
+        self.fee_e = self.app.format_amount_and_units(self.fee)
+        self.ids.fee_for_child.value = self.fee_e
```
SomberNight, line 126: `self.fee_e` is not used anywhere else...
and note that `self.ids.fee_for_child.value` also gets set in `self.update_text()` which is gets called directly below.
Similarly, `self.fee` gets set in `self.update_text()` anyway.

```suggestion
```

electrum/gui/kivy/uix/dialogs/tx_dialog.py
```diff
@@ -250,6 +253,42 @@ def _do_rbf(
...
+        if fee > max_fee:
+            self.show_error(_('Max fee exceeded'))
+            return
+        try:
+            new_tx = self.wallet.cpfp(parent_tx, fee)
+        except CannotCPFP as e:
+            self.app.show_error(str(e))
+            return
+        self.app.tx_dialog(new_tx)
+        self.do_sign()
```
SomberNight, line 290: `do_sign` does not make sense here as that would be for signing the current tx but that is the parent in this case.
The user can sign the child tx manually using the next tx dialog which the line above opens correctly.

```suggestion
```

SomberNight (changes requested):
I've tested this and reviewed the code. It works and is largely ok, but there are many small changes I would like you to make. Some UI changes, some code simplification/clean-ups, and some code style changes. Details in comments.

dhruv-joshi-7:
I noticed that, after performing CPFP, it still shows `Unconfirmed[rbf, feerate]` in the main window, below the transaction.
Should it display `rbf` as it doesn't do so in the Qt GUI?

electrum/gui/kivy/uix/dialogs/cpfp_dialog.py
```diff
@@ -141,21 +134,20 @@ def update_text(self):
...
         else:
             comb_fee = self.fee + self.parent_fee
-            comb_feerate = comb_fee / self.total_size 
-            self.ids.fee_for_child.value = f'{self.fee:.1f} sat'
-            self.ids.output_amount.value = f'{self.max_fee-self.fee:.1f} sat'
-            self.ids.total_fee.value = f'{self.fee+self.parent_fee:.1f} sat'
-            self.ids.total_feerate.value = f'{comb_feerate:.1f} sat/B'
+            comb_feerate = 1000 * comb_fee / self.total_size
+            self.ids.fee_for_child.value = self.app.format_amount_and_units(self.fee)
+            self.ids.output_amount.value = self.app.format_amount_and_units(self.max_fee-self.fee)
```
SomberNight, line 143: One difference between the Qt GUI and here is, that in Qt if the user sets the fee to too high (over max_fee) the displayed text for "Output amount" is "", which I think is better than e.g. "0 BTC" which gets shown here. (In either case the user gets an error that the fee is too high if they try to proceed)
While neither is particularly good UX, I find it somewhat better to do what Qt does and set "Output amount" to "", so let's do that.

See pics:

<details>
 <summary>pic1</summary>

t1
</details>
<details>
 <summary>pic2</summary>

t2
</details>

https://github.com/spesmilo/electrum/blob/bf6400ae30453306c4e8889647d92ddd2bc1d090/electrum/gui/qt/main_window.py#L3334

SomberNight (commented):
Looks good. Just one small thing.

SomberNight:
> I noticed that, after performing CPFP, it still shows `Unconfirmed[rbf, feerate]` in the main window, below the transaction.
Should it display `rbf` as it doesn't do so in the Qt GUI?

I am not sure what you mean. Could you maybe upload screenshots?

Do you mean that for the same transaction,
in kivy, in History tab, it shows `Unconfirmed[rbf, feerate]`, while
in Qt, in History tab, it shows `Unconfirmed[feerate]` (without mentioning `rbf`)?
If so, that sounds weird.

dhruv-joshi-7:
@SomberNight I am not sure how to include only one Button (CPFP or Bump Fee) as both are getting set as True when `update` method is called in `tx_dialog.py`. Either both are appearing or both are not appearing.

SomberNight:
> I am not sure how to include only one Button (CPFP or Bump Fee) 

The logic we want is similar to what is in Qt:
https://github.com/spesmilo/electrum/blob/8ef2052a622a02e293fbc7808faba7dab94eedd4/electrum/gui/qt/history_list.py#L742-L746
That is, only show the CPFP button if cpfp is possible and the bump fee button is not shown, i.e.
re `ActionButtonOption(text=_('Child pays\nfor parent'), func=lambda btn: self.do_cpfp(), enabled=...),`
try e.g. `enabled=(not self.can_rbf and self.can_cpfp)`

SomberNight (approved):
Looks good! Thanks.

PR #7492, opened 2021-09-09
Multiple max spends functionality added

### LINKED ISSUE #7054: paytomany to split max among multiple outputs (allow specifying weights for max amounts, send max, exclamation mark, sharing revenue)
Hey hey,

Quick question: I want to be able to split all funds in the wallet to multiple addresses on a JSONRPC call. Is this possible already?

Else, is there a possibility for a paymaxtomany()-method? It would do exactly the same as paytomany() but instead, it calculates the maximum according to a provided feerate, and instead of hard-coded amounts, it takes percentages.

**Example JSONRPC array for this:**
`{
   "jsonrpc":"2.0",
   "id":"curltext",
   "method":"paymaxtomany",
   "params":{
      "outputs":[
         [
            "tb1qq0g39r3n38y37t6yptru564m9vme6szjjy0nks", 75
         ],
         [
            "tb1qqdkj9feahthysuwdypxznwu94w4hl0uqyj3x5p", 12.5
         ],
         [
            "tb1qsl5sn4frf4tu0f3zajgv3ggqgx6wjqk8sdejjv", 12.5
         ]
      ],
      "feerate":1
   }
}`

Where **75 = 75% of the maximum** that we can send and **12.5 = 12.5% of maximum** that we can send.

If anyone has a solution for this, do let me know. It'd be great!
I don't know if this even possible, but it would be a blessing to us (and I think many others 👍).

siddhantchawla:
Attempt to resolve #7054

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -1709,11 +1709,12 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
             fee=fee_est,
             is_sweep=is_sweep)
         output_values = [x.value for x in outputs]
-        if output_values.count('!') > 1:
-            self.show_error(_("More than one output set to spend max"))
-            return
-
-        output_value = '!' if '!' in output_values else sum(output_values)
+        for value in output_values:
+            if value[-1] == '!':
+                output_value = '!'
+                break
+            else:
+                output_value += value
         conf_dlg = ConfirmTxDialog(window=self, make_tx=make_tx, output_value=output_value, is_sweep=is_sweep)
         if conf_dlg.not_enough_funds:
             # Check if we had enough funds excluding fees,
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -64,7 +64,7 @@ from electrum.util import (format_time,
                            InvalidBitcoinURI, maybe_extract_bolt11_invoice, NotEnoughFunds,
                            NoDynamicFeeEstimates, MultipleSpendMaxTxOutputs,
                            AddTransactionException, BITCOIN_BIP21_URI_SCHEME,
-                           InvoiceError)
+                           InvoiceError, parse_max_spend)
 from electrum.invoices import PR_TYPE_ONCHAIN, PR_TYPE_LN, PR_DEFAULT_EXPIRATION_WHEN_CREATING, Invoice
 from electrum.invoices import PR_PAID, PR_FAILED, pr_expiration_values, LNInvoice, OnchainInvoice
 from electrum.transaction import (Transaction, PartialTxInput,
@@ -1709,12 +1709,10 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
             fee=fee_est,
             is_sweep=is_sweep)
         output_values = [x.value for x in outputs]
-        for value in output_values:
-            if value[-1] == '!':
-                output_value = '!'
-                break
-            else:
-                output_value += value
+        if any(parse_max_spend(outval) for outval in output_values):
+            output_value = '!'
+        else:
+            output_value = sum(output_values)
         conf_dlg = ConfirmTxDialog(window=self, make_tx=make_tx, output_value=output_value, is_sweep=is_sweep)
         if conf_dlg.not_enough_funds:
             # Check if we had enough funds excluding fees,
--- a/electrum/gui/qt/paytoedit.py
+++ b/electrum/gui/qt/paytoedit.py
@@ -31,7 +31,7 @@ from typing import NamedTuple, Sequence, Optional, List, TYPE_CHECKING
 from PyQt5.QtGui import QFontMetrics, QFont
 
 from electrum import bitcoin
-from electrum.util import bfh, maybe_extract_bolt11_invoice, BITCOIN_BIP21_URI_SCHEME
+from electrum.util import bfh, maybe_extract_bolt11_invoice, BITCOIN_BIP21_URI_SCHEME, check_max_spend, parse_max_spend
 from electrum.transaction import PartialTxOutput
 from electrum.bitcoin import opcodes, construct_script
 from electrum.logging import Logger
@@ -135,6 +135,8 @@ class PayToEdit(CompletionTextEdit, ScanQRTextEdit, Logger):
             return '!'
         p = pow(10, self.amount_edit.decimal_point())
         try:
+            if check_max_spend(x): #check for max spend
+                return str(parse_max_spend(x)) + '!'
             return int(p * Decimal(x))
         except decimal.InvalidOperation:
             raise Exception("Invalid amount")
@@ -203,7 +205,7 @@ class PayToEdit(CompletionTextEdit, ScanQRTextEdit, Logger):
                         idx=i, line_content=line.strip(), exc=e, is_multiline=True))
                     continue
             outputs.append(output)
-            if output.value == '!':
+            if check_max_spend(output.value):
                 is_max = True
             else:
                 total += output.value
--- a/electrum/gui/qt/paytoedit.py
+++ b/electrum/gui/qt/paytoedit.py
@@ -31,7 +31,7 @@ from typing import NamedTuple, Sequence, Optional, List, TYPE_CHECKING
 from PyQt5.QtGui import QFontMetrics, QFont
 
 from electrum import bitcoin
-from electrum.util import bfh, maybe_extract_bolt11_invoice, BITCOIN_BIP21_URI_SCHEME, check_max_spend, parse_max_spend
+from electrum.util import bfh, maybe_extract_bolt11_invoice, BITCOIN_BIP21_URI_SCHEME, parse_max_spend
 from electrum.transaction import PartialTxOutput
 from electrum.bitcoin import opcodes, construct_script
 from electrum.logging import Logger
@@ -131,12 +131,10 @@ class PayToEdit(CompletionTextEdit, ScanQRTextEdit, Logger):
         x = x.strip()
         if not x:
             raise Exception("Amount is empty")
-        if x == '!':
-            return '!'
+        if parse_max_spend(x):
+            return x
         p = pow(10, self.amount_edit.decimal_point())
         try:
-            if check_max_spend(x): #check for max spend
-                return str(parse_max_spend(x)) + '!'
             return int(p * Decimal(x))
         except decimal.InvalidOperation:
             raise Exception("Invalid amount")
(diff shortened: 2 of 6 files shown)


electrum/gui/qt/main_window.py
```diff
@@ -1709,11 +1709,12 @@ def pay_onchain_dialog(
...
-            self.show_error(_("More than one output set to spend max"))
-            return
-
-        output_value = '!' if '!' in output_values else sum(output_values)
+        for value in output_values:
+            if value[-1] == '!':
+                output_value = '!'
+                break
+            else:
+                output_value += value
```
SomberNight, line 1717: bug1: This last line `+=` does not work as `output_value` is not yet defined.
bug2: the `value[-1] == '!'` is going to raise if value is an int, which is actually the most common case

-----

Some examples:
```
tb1qk3g0t9pw5wctkzz7gh6k3ljfuukn729s67y54e, 0.002
````
^ is broken.

```
tb1qr7evucrllljtryam6y2k3ntmlptq208pghql2h, !
tb1qar4ye43tdfj6y5n3yndp9adhs2wuz2v0wgqn5l, 3!
````
^ works.

As discussed, we would like even mixed cases such as below to work:
```
tb1qk3g0t9pw5wctkzz7gh6k3ljfuukn729s67y54e, 0.002
tb1qr7evucrllljtryam6y2k3ntmlptq208pghql2h, !
tb1qs3msqp0n0qade2haanjw2dkaa5lm77vwvce00h, 0.003
tb1qar4ye43tdfj6y5n3yndp9adhs2wuz2v0wgqn5l, 3!
```
^ also broken atm.

-----

You could just do (with the modified `parse_max_spend` I describe in another comment):

```suggestion
        if any(parse_max_spend(outval) for outval in output_values):
            output_value = '!'
        else:
            output_value = sum(output_values)
```

electrum/util.py
```diff
@@ -108,6 +108,16 @@ def base_unit_name_to_decimal_point(unit_name: str) -> int:
...
+def check_max_spend(input) -> bool:
+    return isinstance(input, str) and input[-1] == '!'
+
+def parse_max_spend(input: str) -> Decimal:
+    x = ''
+    if input == '!':
+        x = '1'
+    else:
+        x = input[:-1]
+    return Decimal(x)
```
SomberNight, line 120: We should restrict the weights to be positive integers only; Decimal is overkill.
Also, `parse_max_spend` currently assumes `check_max_spend(input) is True`; might as well just add an assert for it near the beginning of `parse_max_spend`.
Also, `input` is not a good name for the function parameter here...

However looking at how you are using these two functions, I think it might make more sense to merge them into a single one.
Why don't we do something like this?

```suggestion
def parse_max_spend(amt: Any) -> Optional[int]:
    """Checks if given amount is "spend-max"-like.
    Returns None or the positive integer weight for "max". Never raises.

    When creating invoices and on-chain txs, the user can specify to send "max".
    This is done by setting the amount to '!'. Splitting max between multiple
    tx outputs is also possible, and custom weights (positive ints) can also be used.
    For example, to send 40% of all coins to address1, and 60% to address2:
    ```
    address1, 2!
    address2, 3!
    ```
    """
    if not (isinstance(amt, str) and amt and amt[-1] == '!'):
        return None
    if amt == '!':
        return 1
    x = amt[:-1]
    try:
        x = int(x)
    except ValueError:
        return None
    if x > 0:
        return x
    return None
```
This of course means you would also have to modify the call sites (to adapt to the new "API").

electrum/wallet.py
```diff
@@ -754,10 +754,13 @@ def create_invoice(self, *, outputs: List[PartialTxOutput], message, pr, URI) ->
         height=self.get_local_height()
         if pr:
             return OnchainInvoice.from_bip70_payreq(pr, height)
-        if '!' in (x.value for x in outputs):
-            amount = '!'
-        else:
-            amount = sum(x.value for x in outputs)
+        amount = 0
+        for x in outputs:
+            if check_max_spend(x.value):
```
SomberNight, line 759: If using the changed `parse_max_spend` I suggest in another comment,
you could instead do `if parse_max_spend(x.value):` here

electrum/wallet.py
```diff
@@ -1333,12 +1336,13 @@ def make_unsigned_transaction(
...
+        i_max_sum = 0
         for i, o in enumerate(outputs):
-            if o.value == '!':
-                if i_max is not None:
-                    raise MultipleSpendMaxTxOutputs()
-                i_max = i
+            if check_max_spend(o.value):
+                x = parse_max_spend(o.value)
+                i_max_sum += x
+                i_max.append((x,i))
```
SomberNight, line 1345: If using the changed `parse_max_spend` I suggest in another comment,
you could instead do
```suggestion
            weight = parse_max_spend(o.value)
            if weight:
                i_max_sum += weight
                i_max.append((weight, i))
```

electrum/gui/qt/paytoedit.py
```diff
@@ -135,6 +135,8 @@ def parse_amount(self, x):
             return '!'
         p = pow(10, self.amount_edit.decimal_point())
         try:
+            if check_max_spend(x): #check for max spend
+                return str(parse_max_spend(x)) + '!'
```
SomberNight, line 139: If using the changed `parse_max_spend` I suggest in another comment,
you could remove these two lines,
and simply change the `if x == '!': return '!'` test above
to `if parse_max_spend(x): return x`

SomberNight (changes requested):
Needs more work.

Looks like you have only tested this for the simplest happy path (where all the amounts are spend-max-like), and only in Qt.
i.e. even the most typical "enter address, enter some satoshi amount, send tx" use case is broken with this.

Please also test saving an invoice (in Qt), i.e. fill "pay to" field and then click "Save".

Also, after having saved an invoice in Qt, close the application, launch it with kivy, and try to pay the invoice. Looking at the kivy code (grep `'!'`), that might not work atm.

The commandline should also support the new syntax (the `paytomany` command in particular, as pointed out in the original issue), which should be easy but not done yet. Just grep `commands.py` for `'!'`.

As discussed originally, this task is not too difficult but it involves fiddling with many code paths and 2-3 line changes over potentially many files.

Lastly, having a unit test would have helped you catch at least some of the above; so that would be nice. Although if you don't have enough time for it then I guess it's not strictly needed.

SomberNight (commented):
Largely ok now!

I have two traces:
- create an invoice in Qt Send tab (fill payto line, and click "Save" button). Try with this (slightly modified from previous example):
  ```
  tb1qk3g0t9pw5wctkzz7gh6k3ljfuukn729s67y54e, 0.002
  tb1qr7evucrllljtryam6y2k3ntmlptq208pghql2h, 2!
  tb1qs3msqp0n0qade2haanjw2dkaa5lm77vwvce00h, 0.003
  tb1qar4ye43tdfj6y5n3yndp9adhs2wuz2v0wgqn5l, 3!
  ```
- right-click>`Details` on it at the bottom (in "Send queue")
<details>
 <summary>trace1</summary>

```
  File "/home/user/wspace/electrum/electrum/gui/qt/invoice_list.py", line 161, in <lambda>
    menu.addAction(_("Details"), lambda: self.parent.show_onchain_invoice(invoice))
  File "/home/user/wspace/electrum/electrum/gui/qt/main_window.py", line 2150, in show_onchain_invoice
    outputs_str = '\n'.join(map(lambda x: x.address + ' : ' + self.format_amount(x.value)+ self.base_unit(), invoice.outputs))
  File "/home/user/wspace/electrum/electrum/gui/qt/main_window.py", line 2150, in <lambda>
    outputs_str = '\n'.join(map(lambda x: x.address + ' : ' + self.format_amount(x.value)+ self.base_unit(), invoice.outputs))
  File "/home/user/wspace/electrum/electrum/gui/qt/main_window.py", line 885, in format_amount
    return self.config.format_amount(amount_sat, is_diff=is_diff, whitespaces=whitespaces)
  File "/home/user/wspace/electrum/electrum/simple_config.py", line 664, in format_amount
    return format_satoshis(
  File "/home/user/wspace/electrum/electrum/util.py", line 692, in format_satoshis
    assert isinstance(x, (int, float, Decimal)), f"{x!r} should be a number"
AssertionError: '2!' should be a number
```
</details>
- then, close Qt (but don't delete the invoice!), start kivy and open the wallet
- go to Send tab, click the invoice at the bottom, and click "Pay"
<details>
 <summary>trace2</summary>

```
E | gui.kivy.uix.dialogs.crash_reporter.ExceptionHook | exception caught by crash reporter
  File "/home/user/.local/lib/python3.8/site-packages/kivy/base.py", line 347, in mainloop
    self.idle()
  File "/home/user/.local/lib/python3.8/site-packages/kivy/base.py", line 391, in idle
    self.dispatch_input()
  File "/home/user/.local/lib/python3.8/site-packages/kivy/base.py", line 342, in dispatch_input
    post_dispatch_input(*pop(0))
  File "/home/user/.local/lib/python3.8/site-packages/kivy/base.py", line 308, in post_dispatch_input
    wid.dispatch('on_touch_up', me)
  File "kivy/_event.pyx", line 709, in kivy._event.EventDispatcher.dispatch
  File "/home/user/.local/lib/python3.8/site-packages/kivy/uix/behaviors/button.py", line 179, in on_touch_up
    self.dispatch('on_release')
  File "kivy/_event.pyx", line 705, in kivy._event.EventDispatcher.dispatch
  File "kivy/_event.pyx", line 1248, in kivy._event.EventObservers.dispatch
  File "kivy/_event.pyx", line 1132, in kivy._event.EventObservers._dispatch
  File "/home/user/.local/lib/python3.8/site-packages/kivy/lang/builder.py", line 57, in custom_callback
    exec(__kvlang__.co_value, idmap)
  File "<string>", line 70, in <module>
  File "/home/user/wspace/electrum/electrum/gui/kivy/uix/dialogs/invoice_dialog.py", line 131, in do_pay
    self.app.send_screen.do_pay_invoice(invoice)
  File "/home/user/wspace/electrum/electrum/gui/kivy/uix/screens.py", line 359, in do_pay_invoice
    self._do_pay_onchain(invoice)
  File "/home/user/wspace/electrum/electrum/gui/kivy/uix/screens.py", line 374, in _do_pay_onchain
    amount = sum(map(lambda x: x.value, outputs)) if '!' not in [x.value for x in outputs] else '!'
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```
</details>

SomberNight (approved):
Looks good now. Thanks!

SomberNight:
I have done some minor follow-up cleanups in https://github.com/spesmilo/electrum/commit/acbb3632409cfcd8cb02b2effbeda8f1ab0faac7
and added a unit test in https://github.com/spesmilo/electrum/commit/85780eea1a32429fe513f5bdc03ac8cca207d1fd

Issue #7475, opened 2021-08-27
change password from terminal. gets prompted for new password even if already provided as parameter


paintmeyellow:
### Why is there a prompt?
![image](https://user-images.githubusercontent.com/24860764/131099918-b26bfbe7-308c-4d49-8ddd-323402f3ae4e.png)
![image](https://user-images.githubusercontent.com/24860764/131099990-396bf018-e03e-4d46-a89e-957ca64dcd27.png)

SomberNight:
ACK, indeed. it's because we unconditionally prompt for "new password",
see https://github.com/spesmilo/electrum/blob/c338cebe556ddd0a9e2e5e3579a56c807b992c6e/run_electrum#L169-L171

PR #7495, opened 2021-09-10
Fix bug with new password prompt

bradley-ray:
No longer prompts for new password if `--new_password` parameter is used

fixes https://github.com/spesmilo/electrum/issues/7475


SomberNight:
thanks

PR #7502, opened 2021-09-20
No convert key

ecdsa:
Do not convert StoredDict keys to str.
Instead, convert json keys to int when storage is read.


SomberNight:
Concept ACK.
Haven't tested but largely looks ok.

PR #7505, opened 2021-09-22
watchtower: continuously check for added channels

bitromortac:
From my testing I saw that the watchtower isn't actually tested for publication of the punishment transaction. That test fails, which is fixed by the watchtower polling the sweepstore. Maybe there's a more efficient way using events?

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -306,7 +306,7 @@ class WatchTowerServer(AuthenticatedServer):
         await site.start()
 
     async def get_ctn(self, *args):
-        return await self.lnwatcher.sweepstore.get_ctn(*args)
+        return await self.lnwatcher.get_ctn(*args)
 
     async def add_sweep_tx(self, *args):
         return await self.lnwatcher.sweepstore.add_sweep_tx(*args)
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -313,10 +313,11 @@ class WatchTower(LNWatcher):
                 await self.tx_progress[funding_outpoint].tx_queue.put(tx)
             return txid
 
-    def get_ctn(self, outpoint, addr):
-        async def f():
-            return await self.sweepstore.get_ctn(outpoint, addr)
-        return self.network.run_from_another_thread(f())
+    async def get_ctn(self, outpoint, addr):
+        if addr not in self.callbacks.keys():
+            self.logger.info(f'watching new channel: {outpoint} {addr}')
+            self.add_channel(outpoint, addr)
+        return await self.sweepstore.get_ctn(outpoint, addr)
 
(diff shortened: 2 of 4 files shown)


SomberNight:
Hmm, good catch...

First off, notice that `start_watching` is called via `asyncio.ensure_future`, which means, if it raises, the exception will not get logged and debugging also becomes more difficult.
https://github.com/spesmilo/electrum/blob/a555eb3106f15c1177c2a9e7c1d351c657c4af7d/electrum/network.py#L352
Especially for long running tasks, such as what `start_watching` would become with this PR, at least it should be decorated with `@log_exceptions`. (in other places of the code, often TaskGroups are used instead, such as in LNWorker.taskgroup and Peer.taskgroup)

> Maybe there's a more efficient way using events?

Right... I think we should try to fix the issue that way - not for efficiency reasons, but to make it easier to understand what is going on.
Note that the watchtower methods exposed via RPC are defined here:
https://github.com/spesmilo/electrum/blob/a555eb3106f15c1177c2a9e7c1d351c657c4af7d/electrum/daemon.py#L308-L312

It is this `add_sweep_tx` method that gets called with a "new channel" which is the root cause of your bug.
So we should simply make sure the callback gets added then.

What I suggest is this:
`WatchTowerServer.add_sweep_tx` should not access `lnwatcher.sweepstore` directly. Instead, there could be a proxy method in `WatchTower`.
So `WatchTowerServer.add_sweep_tx` could call `WatchTower.add_sweep_tx`, which makes sure the channel callback exists and calls `SweepStore.add_sweep_tx`.

bitromortac:
I modified `WatchTower.get_ctn` to be async and added the channels to be watched there, as in https://github.com/spesmilo/electrum/blob/baff4fa6259372b667042dae808123bfcb0f0964/electrum/lnworker.py#L712-L720 `get_ctn` is called before `add_sweep_tx` and in this way the watchtower should already watch for that channel. The `add_sweep_tx` method doesn't know about the address the callbacks are keyed by, so I can't check for existence of the callback nor add a new channel.

SomberNight:
>  `get_ctn` is called before `add_sweep_tx` and in this way the watchtower should already watch for that channel. `The add_sweep_tx` method doesn't know about the address the callbacks are keyed by, so I can't check for existence of the callback nor add a new channel.

Right, ok. It looks ok. Thanks.

I think it is bad design that a `get_*` method has such a non-intuitive side-effect that we rely on. Ideally getters should not have side-effects. That is, it might be better to just change to a cleaner API here.

But re this PR specifically, I see `SweepStore.get_ctn` already had a semantically similar side-effect, and this is just extended to `WatchTower.get_ctn` also having it, so it's fine.

SomberNight (approved):

PR #7545, opened 2021-10-25
Add support for Unix domain sockets

### LINKED ISSUE #4955: Unix domain sockets for RPC
It would be useful if Electrum supported optionally making the RPC server listen on a Unix domain socket instead of a TCP socket, for the same reasons that this has been [proposed](https://github.com/bitcoin/bitcoin/issues/5029) for Bitcoin Core.

yanmaani:
This patch adds support for Unix domain sockets to Electrum. To use, spawn the daemon with `-u` or `--unixdomain`, e.g.:
`electrum daemon -du`

You can also specify a path for the domain socket manually with the `--sockpath` argument, e.g.:
`electrum daemon -du --sockpath=$(mktemp --dry-run)`

An integration test is included under `electrum/tests/test_unixsocket.sh`. I am not at all sure if this would meet your standards for tests, because it's a shell script. I would thus be happy to receive feedback on this point.

Closes #4955.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1475,6 +1475,8 @@ def get_parser():
     # daemon
     parser_daemon = subparsers.add_parser('daemon', help="Run Daemon")
     parser_daemon.add_argument("-d", "--detached", action="store_true", dest="detach", default=False, help="run daemon in detached mode")
+    parser_daemon.add_argument("--rpcsock", dest="rpcsock", default=None, help="what socket type to which to bind RPC daemon", choices=['unix', 'tcp', 'auto'])
+    parser_daemon.add_argument("--rpcsockpath", dest="rpcsockpath", help="where to place RPC file socket")
     add_network_options(parser_daemon)
     add_global_options(parser_daemon)
     # commands
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -61,9 +61,14 @@ _logger = get_logger(__name__)
 class DaemonNotRunning(Exception):
     pass
 
+def get_rpcsock_defaultpath(config: SimpleConfig):
+    return os.path.join(config.path, 'daemon_rpc_socket')
+
+def get_rpcsock_default_type(osname):
+    return 'tcp'
+
 def get_lockfile(config: SimpleConfig):
     return os.path.join(config.path, 'daemon')
-
 
 def remove_lockfile(lockfile):
     os.unlink(lockfile)
@@ -96,7 +101,15 @@ def request(config: SimpleConfig, endpoint, args=(), timeout=60):
         create_time = None
         try:
             with open(lockfile) as f:
-                (host, port), create_time = ast.literal_eval(f.read())
+                socktype, address, create_time = ast.literal_eval(f.read())
+                if socktype == 'unix':
+                    path = address
+                    (host, port) = "127.0.0.1", 0
+                    # We still need a host and port for e.g. HTTP Host header
+                elif socktype == 'tcp':
+                    (host, port) = address
+                else:
+                    raise Exception(f"corrupt lockfile; socktype={socktype!r}")
         except Exception:
             raise DaemonNotRunning()
         rpc_user, rpc_password = get_rpc_credentials(config)
@@ -104,7 +117,13 @@ def request(config: SimpleConfig, endpoint, args=(), timeout=60):
         auth = aiohttp.BasicAuth(login=rpc_user, password=rpc_password)
         loop = asyncio.get_event_loop()
         async def request_coroutine():
-            async with aiohttp.ClientSession(auth=auth) as session:
+            if socktype == 'unix':
+                connector = aiohttp.UnixConnector(path=path)
+            elif socktype == 'tcp':
+                connector = None # This will transform into TCP.
+            else:
+                raise Exception(f"impossible socktype ({socktype!r})")
+            async with aiohttp.ClientSession(auth=auth, connector=connector) as session:
                 c = util.JsonRPCClient(session, server_url)
                 return await c.request(endpoint, *args)
         try:
@@ -225,6 +244,9 @@ class CommandsServer(AuthenticatedServer):
         self.daemon = daemon
         self.fd = fd
         self.config = daemon.config
+        sockettype = self.config.get('rpcsock', 'auto')
+        self.socktype = sockettype if sockettype != 'auto' else get_rpcsock_default_type(os.name)
+        self.sockpath = self.config.get('rpcsockpath', get_rpcsock_defaultpath(self.config))
         self.host = self.config.get('rpchost', '127.0.0.1')
         self.port = self.config.get('rpcport', 0)
         self.app = web.Application()
@@ -239,10 +261,21 @@ class CommandsServer(AuthenticatedServer):
     async def run(self):
         self.runner = web.AppRunner(self.app)
         await self.runner.setup()
-        site = web.TCPSite(self.runner, self.host, self.port)
+        if self.socktype == 'unix':
+            site = web.UnixSite(self.runner, self.sockpath)
+        elif self.socktype == 'tcp':
+            site = web.TCPSite(self.runner, self.host, self.port)
+        else:
+            raise Exception(f"unknown socktype '{self.socktype!r}'")
         await site.start()
         socket = site._server.sockets[0]
-        os.write(self.fd, bytes(repr((socket.getsockname(), time.time())), 'utf8'))
+        if self.socktype == 'unix':
+            addr = self.sockpath
+        elif self.socktype == 'tcp':
+            addr = socket.getsockname()
+        else:
+            raise Exception(f"impossible socktype ({self.socktype!r})")
+        os.write(self.fd, bytes(repr((self.socktype, addr, time.time())), 'utf8'))
         os.close(self.fd)
 
     async def ping(self):
--- a/electrum/tests/regtest/regtest.sh
+++ b/electrum/tests/regtest/regtest.sh
@@ -356,3 +356,19 @@ if [[ $1 == "watchtower" ]]; then
     echo "watchtower publishes justice transaction"
     wait_until_spent $ctx_id 1  # alice's to_local gets punished immediately
 fi
+
+if [[ $1 == "unixsockets" ]]; then
+    # This looks different because it has to run the entire daemon
+    # Test domain socket behavior
+    ./run_electrum --regtest daemon -d --rpcsock=unix # Start daemon with unix domain socket
+    ./run_electrum --regtest stop # Errors if it can't connect
+    # Test custom socket path
+    f=$(mktemp --dry-run)
+    ./run_electrum --regtest daemon -d --rpcsock=unix --rpcsockpath=$f
+    [ -S $f ] # filename exists and is socket
+    ./run_electrum --regtest stop
+    rm $f # clean up
+    # Test for regressions in the ordinary TCP functionality.
+    ./run_electrum --regtest daemon -d --rpcsock=tcp
+    ./run_electrum --regtest stop
+fi
--- a/electrum/tests/regtest.py
+++ b/electrum/tests/regtest.py
@@ -34,6 +34,13 @@ class TestLightning(unittest.TestCase):
             self.run_shell(['stop', agent])
 
 
+class TestUnixSockets(TestLightning):
+    agents = []
+
+    def test_unixsockets(self):
+        self.run_shell(['unixsockets'])
+
+
 class TestLightningAB(TestLightning):
     agents = ['alice', 'bob']
 


electrum/tests/regtest/regtest.sh
```diff
@@ -356,3 +356,19 @@ if [[ $1 == "watchtower" ]]; then
...
+    ./run_electrum --regtest daemon -d --unixdomain # Start daemon with unix domain socket
+    ./run_electrum --regtest stop # Errors if it can't connect
+    # Test custom socket path
+    f=$(mktemp --dry-run)
+    ./run_electrum --regtest daemon -d --unixdomain --unixsockpath=$f
+    [ -S $f ] # filename exists and is socket
+    ./run_electrum --regtest stop
+    rm $f # clean up
+    # Test for regressions in the ordinary TCP functionality.
+    ./run_electrum --regtest daemon -d --unixdomain # Start daemon with Internet socket
```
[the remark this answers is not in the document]
yanmaani: > Isn't this using a Unix socket?

Oops!

>  Also, can we explicitly test for whether the Unix socket file is gone here like it should be?

Why should it be gone? I think that might create some problems of its own. Just spitballing here, but if an attacker can cause me to use a certain socket, that would also enable them to make me to delete that socket. That seems like not desirable behavior at all, potentially.

Under "standard" conditions, it will always use the same socket file, so it doesn't really have to be cleaned up. And if a caller designates a new socket file to use, it's the caller's task to clean it up. 

At the end of the day, I don't want to add complexity in detecting whether we created that socket file or it existed before and some extra config about that, and I don't want the undesired behavior of going around and deleting random socket files, so the current behaviour seems fine.

yanmaani:
>Only concern I have is whether those exception classes can be triggered by other errors too, e.g. it would be bad if an attacker could induce a downgrade to TCP sockets by causing, say, a permission error of some kind.

It'd still be better than the old state, where it always uses TCP. If you've managed to botch something about Unix domain sockets, which I'm frankly not even sure is possible, that seems like a very user-hostile thing for Electrum to start going off about.

Just spitballing here again, but what if you're on some ultra-strange ReactOS config where all the systems are totally broken? Should that also cause Electrum to fail, just because it uses some obscure corner of the OS that probably isn't that well tested?

Seems to me like, if you absolutely want domain sockets, you'd specify that explicitly in the config and then it should respect that, but if you don't, then you clearly don't have a view as to which should be used. In that case, either of TCP or file would be in accordance with the user's wishes, no?

electrum/commands.py
```diff
@@ -1475,6 +1475,8 @@ def get_parser():
     # daemon
     parser_daemon = subparsers.add_parser('daemon', help="Run Daemon")
     parser_daemon.add_argument("-d", "--detached", action="store_true", dest="detach", default=False, help="run daemon in detached mode")
+    parser_daemon.add_argument("--unixdomain", action="store_true", dest="unixdomain", default=False, help="bind daemon to unix socket")
```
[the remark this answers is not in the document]
yanmaani: Okay, sure.

yanmaani:
@JeremyRand Ready for review.

electrum/commands.py
```diff
@@ -1475,6 +1475,8 @@ def get_parser():
     # daemon
     parser_daemon = subparsers.add_parser('daemon', help="Run Daemon")
     parser_daemon.add_argument("-d", "--detached", action="store_true", dest="detach", default=False, help="run daemon in detached mode")
+    parser_daemon.add_argument("--unixsock", action="store_true", dest="unixsocket", default=False, help="bind daemon to unix socket")
+    parser_daemon.add_argument("--unixsockpath", dest="unixsocketpath", help="where to place unix socket")
```
SomberNight, line 1479: If we made unix sockets the default on Linux, which to me sounds like a good default, `--unixsock` (noarg) couldn't really be adopted as-is.
We could instead have `--unixsock` that takes a boolean...
Except, sounds like we might (later) add support for Windows named pipes, in which case we might as well do it completely differently now:
how about this:
- `--rpcsock`, which takes one arg, valid arg values for now being `tcp` and `unixsock` (and later also `winpipe`)? and
- `--rpcsockpath`, which also takes one arg, a file system path, used in case of `rpcsock==unixsock` (and later also when `rpcsock==winpipe`)
- the respective config keys can be named identically to the cli/rpc options: `rpcsock` and `rpcsockpath`

SomberNight:
>> if they are better for our use case than TCP sockets, could we just make these the default on platforms where they are available?
>
> I think making Unix domain sockets the default on supported platforms is reasonable. Doing so will require users to change a setting if they want to access the RPC port via a TCP socket, but AFAIK we already require users to change a setting (to set a constant port) in such situations, so that seems harmless as long as it's mentioned in the release notes.

So to be clear, to CLI and GUI users (but not RPC users), using unix sockets by default (when available) should be an invisible change, right?
In that case we should just make this the default (it can be in a follow-up PR but changes here should already take it into account that we might change defaults soon).

electrum/commands.py
yanmaani: In Python 3.9, we'll get `argparse.BooleanOptionalAction` in `argparse`:
```
>>> import argparse
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--foo', action=argparse.BooleanOptionalAction)
>>> parser.parse_args(['--no-foo'])
Namespace(foo=False)
```

Maybe you could do a poor man's version of that, by making `--unixsock` a no-op and then making `--no-unixsock` the active one, and then just slot in `argparse.BooleanOptionalAction` once Python 3.9 is old enough. It would be a bit ungainly, but only temporary.

As for `--rpcsock`, there is, to my knowledge, not any system where you can have both Windows pipes and Unix sockets. So I don't understand why you would need the triple-barreled option - to me, a boolean (use file/tcp) seems more suitable.

You could however rename the parameter to `--rpcsock` now already, because that way you wouldn't have to deprecate `--unixsock` because it never existed.

SomberNight: > As for `--rpcsock`, there is, to my knowledge, not any system where you can have both Windows pipes and Unix sockets. So I don't understand why you would need the triple-barreled option - to me, a boolean (use file/tcp) seems more suitable.

There might already be [unix socket support on Windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/), or maybe there will be in the future. Pretty sure they cannot be used for our purposes as of right now though. Nevertheless this was not really my motivation.
Simply, the alternative where the user runs with `--unixsock` on Windows, and it uses something similar (named pipes) but not quite the same, seemed a bit weird.

> making `--unixsock` a no-op and then making `--no-unixsock` the active one, and then just slot in `argparse.BooleanOptionalAction` once Python 3.9 is old enough.

Right... that too could work.

-----

In any case, the main benefit of `--unixsock` (noarg) is ~ease of typing on the CLI. If we make a sane default, and also make it configurable via config options (so that users don't need to type it for every command), then it does not matter if it is convenient to type.

- Let's just make the option a bit more generic now, before introducing it, so that it does not have to be changed later.
- Also, I like `rpcsock` more than `unixsock` as all(?) the exisiting local rpc options are named `rpc*` (`rpchost`, `rpcport`, `rpcuser`, `rpcpassword`). Relatedly, the `unixsock` name is somewhat ambiguous re what it refers to: is it the local rpc or is it the client-server communication?
- I guess you have a decent point that there might be no point in distinguishing between unix sockets and windows named pipes.

So, how about:

- `--rpcsock`, which takes one arg, valid arg values for now being `tcp` and `file`
    `parser.add_argument('rpcsock', choices=['tcp', 'file'])`
    later, if needed, we can split "file" up into separate options and default "file" to one of them
- `--rpcsockpath`, which also takes one arg, a file system path, used in case of `rpcsock==file`
- the respective config keys can be named identically to the cli/rpc options: `rpcsock` and `rpcsockpath`

JeremyRand:
> It'd still be better than the old state, where it always uses TCP. If you've managed to botch something about Unix domain sockets, which I'm frankly not even sure is possible, that seems like a very user-hostile thing for Electrum to start going off about.

It would be extremely bad for a secure default to silently downgrade to an insecure configuration, because users cannot be trusted to explicitly choose a secure configuration when it's already the default.

The aiohttp documentation does not guarantee any particular exception when creating a Unix server when Unix sockets are unsupported.  The source code seems to simply call out to the Python asyncio event loop's `create_unix_server` without checking.  The Python docs for that function do not guarantee any exception either, nor do they explicitly guarantee OS availability, but they do indicate that it's available when `socket.AF_UNIX` is available.  Python socket docs indicate that `socket.AF_UNIX` is defined if and only if Unix sockets are available.  So I think we should check for `socket.AF_UNIX` existing instead of relying on undocumented exception behavior.

Windows named pipe support isn't in aiohttp stable yet (it was added in v4.0.0a1), so Electrum can't use it yet, unless Electrum wants to use an unstable aiohttp.  The docs for aiohttp don't guarantee any exception behavior for named pipes either; the source code does check for an event loop implementation that's guaranteed by Python docs to only be available on Windows, but the Python docs do not guarantee that it's the only event loop implementation available on Windows.  So I think we should explicitly check         whether the OS is Windows rather than relying on aiohttp to produce an exception when named pipes aren't available.  Anyway we can discuss that in a follow-up issue since named pipes aren't part of this PR.

> As for --rpcsock, there is, to my knowledge, not any system where you can have both Windows pipes and Unix sockets. So I don't understand why you would need the triple-barreled option - to me, a boolean (use file/tcp) seems more suitable.

They are both supported on current Windows 10, although Python doesn't yet support Unix sockets on Windows, see Python issue 33408.

> Also, I like rpcsock more than unixsock as all(?) the exisiting local rpc options are named rpc* (rpchost, rpcport, rpcuser, rpcpassword). Relatedly, the unixsock name is somewhat ambiguous re what it refers to: is it the local rpc or is it the client-server communication?

There are at least 3 things that would benefit from Unix socket support: the local RPC socket, the client/server socket, and the SOCKS proxy socket (Tor supports listening on a Unix socket, and they recommend this as preferable to using a TCP socket).  So yes, we should make sure that our flag names are unambiguous between these 3 things.

> I guess you have a decent point that there might be no point in distinguishing between unix sockets and windows named pipes.

Disagree; Windows software needs to care because Unix sockets are only on Win10+ (AFAIK Electrum still supports earlier Windows versions), and different RPC client software might only support one or the other.  So I'd recommend the name "unix" rather than "file".

> So to be clear, to CLI and GUI users (but not RPC users), using unix sockets by default (when available) should be an invisible change, right?

That's correct.

> In that case we should just make this the default (it can be in a follow-up PR but changes here should already take it into account that we might change defaults soon).

Agree.

yanmaani:
Okay, so I'm getting mixed signals here. But seems like everyone will be happy if I do `--rpcsock=[namedpipe|unix|tcp|auto]`. If some niche use case calls for it, you could make a new setting that allows you to change what OS Electrum thinks it's on in a future PR. 

Since changing the default was for a future PR, and I'm not doing Windows named pipes, I'll implement `unix`, `tcp`, and `auto` (as an unconditional alias of `tcp`). This will enable people in the future to change it without keeping ugly legacy code.

yanmaani:
@JeremyRand @SomberNight Ready for review.

yanmaani:
Oops, I made a typo. Ready for review for real this time.

electrum/commands.py
```diff
@@ -1475,6 +1475,8 @@ def get_parser():
     # daemon
     parser_daemon = subparsers.add_parser('daemon', help="Run Daemon")
     parser_daemon.add_argument("-d", "--detached", action="store_true", dest="detach", default=False, help="run daemon in detached mode")
+    parser_daemon.add_argument("--rpcsock", dest="rpcsocket", default='auto', help="what socket type to which to bind RPC daemon", choices=['unix', 'tcp', 'auto'])
+    parser_daemon.add_argument("--rpcsockpath", dest="rpcsocketpath", help="where to place RPC file socket")
```
SomberNight, line 1479: Let's make the config keys match the cli options.

```suggestion
    parser_daemon.add_argument("--rpcsock", dest="rpcsock", default='auto', help="what socket type to which to bind RPC daemon", choices=['unix', 'tcp', 'auto'])
    parser_daemon.add_argument("--rpcsockpath", dest="rpcsockpath", help="where to place RPC file socket")
```

electrum/daemon.py
```diff
@@ -96,15 +101,29 @@ def request(config: SimpleConfig, endpoint, args=(), timeout=60):
...
-                (host, port), create_time = ast.literal_eval(f.read())
+                socktype, address, create_time = ast.literal_eval(f.read())
+                if socktype == 'unix':
+                    path = address
+                    (host, port) = "127.0.0.1", 0
+                    # We still need a host and port for e.g. HTTP Host header
+                elif socktype == 'tcp':
+                    (host, port) = address
+                else:
+                    assert False
```
SomberNight, line 112: Please only use asserts for sanity checks or to clearly state your assumptions for readers, not for important logic.
Raise exceptions instead (guarded with if branches).
Asserts are not guaranteed to be there at runtime: https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE

```suggestion
                    raise Exception(f"unknown socktype={socktype!r}")
```

yanmaani: This is a sanity check. It's not supposed to be possible, I'm not supposed to have random non-existent socktypes in my lockfile. As for the command-line option:
```
$ ./run_electrum --regtest daemon --rpcsock=x    
usage: run_electrum daemon [-h] [-d] [--rpcsock {unix,tcp,auto}]
                           [--rpcsockpath RPCSOCKETPATH]
                           [-f SERVERFINGERPRINT] [-1] [-s SERVER] [-p PROXY]
                           [--noonion] [--skipmerklecheck] [-v VERBOSITY]
                           [-V VERBOSITY_SHORTCUTS] [-D ELECTRUM_PATH] [-P]
                           [--testnet] [--regtest] [--simnet] [--signet] [-o]
run_electrum daemon: error: argument --rpcsock: invalid choice: 'x' (choose from 'unix', 'tcp', 'auto')
```

electrum/commands.py
yanmaani: Disagree. For external options, being shorter is more important. For internal code, being descriptive is more important.

If you insist, I'm not strictly opposed to changing it though.

electrum/daemon.py
SomberNight: The issue is that the user might set the config var directly, not via CLI option, in which case the possible values are unrestricted.
The same applies (and it's more apparent) to the other lines where you added an `assert False`.
`daemon.socktype` can have any value currently

electrum/commands.py
SomberNight: > For external options, being shorter is more important. For internal code, being descriptive is more important.

Yes, I agree. However having the same name in both user-facing interfaces is another tradeoff (eases documentation, more intuitive, etc). I would argue the difference here is so small that it's worth it to unify the names.

I am fine with either the longer name or the shorter name, but I would prefer if they matched. We can use the longer names if you like that more.

yanmaani: That is a fair point. No, in that case I prefer the shorter ones everywhere.

electrum/daemon.py
yanmaani: Right, I missed that. I changed it everywhere for consistency/simplicity, although it would arguably be more correct in a sense to keep the asserts for the lockfile.

yanmaani:
The requested changes have been made; ready for review.

electrum/commands.py
```diff
@@ -1475,6 +1475,8 @@ def get_parser():
     # daemon
     parser_daemon = subparsers.add_parser('daemon', help="Run Daemon")
     parser_daemon.add_argument("-d", "--detached", action="store_true", dest="detach", default=False, help="run daemon in detached mode")
+    parser_daemon.add_argument("--rpcsock", dest="rpcsock", default='auto', help="what socket type to which to bind RPC daemon", choices=['unix', 'tcp', 'auto'])
```
SomberNight, line 1478:
```suggestion
    parser_daemon.add_argument("--rpcsock", dest="rpcsock", default=None, help="what socket type to which to bind RPC daemon", choices=['unix', 'tcp', 'auto'])
```
This is to fix the following bug:
```
$ ./run_electrum --testnet -o setconfig rpcsock unix
true
$ ./run_electrum --testnet daemon -d
starting daemon (PID 207559)
$ cat ~/.electrum/testnet/daemon
('tcp', ('127.0.0.1', 7777), 1635959223.0871708)
$ ./run_electrum --testnet -o getconfig rpcsock
unix
```

yanmaani: Thanks, didn't catch that. The correct thing is indeed to apply the default in `CommandsServer`.

yanmaani:
Ready for review.

electrum/daemon.py
```diff
@@ -239,10 +261,21 @@ def __init__(self, daemon, fd):
     async def run(self):
         self.runner = web.AppRunner(self.app)
         await self.runner.setup()
-        site = web.TCPSite(self.runner, self.host, self.port)
+        if self.socktype == 'unix':
+            site = web.UnixSite(self.runner, self.sockpath)
+        elif self.socktype == 'tcp':
+            site = web.TCPSite(self.runner, self.host, self.port)
+        else:
+            raise Exception(f"unknown socktype '{socktype!r}'")
```
SomberNight, line 269: flake8 test is failing
```
./electrum/daemon.py:269:30: F821 undefined name 'socktype'
            raise Exception(f"unknown socktype '{socktype!r}'")
                             ^
```
```suggestion
            raise Exception(f"unknown socktype '{self.socktype!r}'")
```

same thing a few lines below

yanmaani: Thanks, I missed that.

SomberNight (approved):
Looks good to me. Thanks.

electrum/tests/regtest/regtest.sh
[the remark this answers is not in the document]
SomberNight: Oh, I forgot about this thread, sorry.
@JeremyRand, could you clarify if the socket file should be cleaned up by the daemon upon stopping - and if so, what is the purpose of doing that? (note that if the daemon got killed for whatever reason it couldn't delete the socket file in any case)

SomberNight:
Are you interested in making another PR to make `rpcsock=unix` the default?

>  Python socket docs indicate that `socket.AF_UNIX` is defined if and only if Unix sockets are available. So I think we should check for `socket.AF_UNIX` existing instead of relying on undocumented exception behavior.

Sounds like we should branch on `hasattr(socket, "AF_UNIX")`.

yanmaani:
Sure, if you want. All you'd have to do is to change one line, as seen in this diff:
https://github.com/spesmilo/electrum/compare/9ad67dc6db03f34a875b7b3b08ffd3159fd7d196..710ae4aad5c7c85451654a8b20f43ccf6a43affc

I'm not sure if it would be good to wait for the functionality to be tested first, though. Let's see what @JeremyRand says.

PR #7577, opened 2021-11-25
integration tests: all Electrum components inside an interpreter

bitromortac:
Currently, the tests have the same prerequisites as the regtest tests and can only be executed in isolation (one test at a time per pytest/unittest call). These tests are intended as an a supplement for the regtests.

Advantages:
* launches several Electrum components to interact on a regtest network in a single interpreter
* eases development, as one can debug into all clients within the same IDE (helped me a lot)
* makes tests more flexible as one can inspect/patch methods and attributes while testing

Downsides:
* tests do not reproduce fully realistic scenarios, where each component is an executable (to an extent this is also true for the regtest tests where only a single bitcoind is used)

- [ ] some Electrum imports are singletons (network, logging, ...), further work is needed to let tests run sequentially
- [ ] synchronization of code, get rid of asyncio.sleeps


helpmelorda (commented):
Auto-Correct/fetch LOAD all Bitcoin Unspent outputs into history

PR #7584, opened 2021-12-05
Add with_sync option to restore command

ZenulAbidin:
This commit adds a `restore_sync` command to Electrum that behaves like restore except it waits for the wallet to synchronize with the network, by starting the network module inside the wallet.


ZenulAbidin:
Is there any news about this?

electrum/wallet.py
```diff
@@ -3342,6 +3342,12 @@ def restore_wallet_from_text(text, *, path, config: SimpleConfig,
     assert not storage.file_exists(), "file was created too soon! plaintext keys might have been written to disk"
     wallet.update_password(old_pw=None, new_pw=password, encrypt_storage=encrypt_file)
     wallet.synchronize()
+    if wait_for_sync:
+        if not sync_network:
+            raise Exception('sync_network is required if wait_for_sync is True')
+        wallet.start_network(sync_network)
+        while not wallet.network.is_connected:
+            time.sleep(0.1)
```
SomberNight, line 3350: Note that `async def restore_sync` is an async function, running on the event loop.
Hence this polling here is blocking the event loop (for the daemon, when used via RPC).
I recommend moving this out into the async context, into `restore_sync`. i.e. after `restore_wallet_from_text()` returns, implement the new logic there, outside this method, in `restore_sync`. You can poll there with e.g. `await asyncio.sleep(0.1)`, which would then not block the event loop.

electrum/wallet.py
```diff
@@ -3342,6 +3342,12 @@ def restore_wallet_from_text(text, *, path, config: SimpleConfig,
     assert not storage.file_exists(), "file was created too soon! plaintext keys might have been written to disk"
     wallet.update_password(old_pw=None, new_pw=password, encrypt_storage=encrypt_file)
     wallet.synchronize()
+    if wait_for_sync:
+        if not sync_network:
+            raise Exception('sync_network is required if wait_for_sync is True')
+        wallet.start_network(sync_network)
+        while not wallet.network.is_connected:
```
SomberNight, line 3349: `wallet.network.is_connected` has very little to do with the wallet being synchronized. It is ~sort of a necessary but not sufficient condition. You can check `wallet.is_up_to_date()` instead.

electrum/commands.py
```diff
@@ -276,6 +275,27 @@ async def restore(self, text, passphrase=None, password=None, encrypt_file=True,
             'msg': d['msg'],
         }
 
+    @command('n')
+    async def restore_sync(self, text, passphrase=None, password=None, encrypt_file=True, wallet_path=None):
```
SomberNight, line 279: I see the TODO says "create a separate command", but now thinking about it, adding an option to the existing `restore` command seems nicer. e.g. `sync_history=False`. You can add some description for such an option in `command_options` (around line 1300).

SomberNight (changes requested):

electrum/commands.py
ZenulAbidin: I have modified the commit to make it an option that polls `wallet.is_up_to_date()` using the wallet object returned by `restore_wallet_from_text()`. I have tested it and it works as intended now.

ZenulAbidin:
@SomberNight Commit has been amended with recommended changes.

electrum/commands.py
ZenulAbidin: @SomberNight can you verify if the contents of the new commits are acceptable for merge?

PR #7586, opened 2021-12-06
lnpeer: modern closing negotiation

bitromortac:
Implements a recent spec change concerning channel closing https://github.com/lightning/bolts/pull/847.
The new fee negotiation is based on fee ranges but is also backwards compatible to the incremental closing negotiation.
For testing purposes there's a configuration constant `MODERN_FEE` to trigger the new behavior, which is on by default.
I also added a new regtest test for collaborative channel closing.

Todo:
- [x] Clarify what lower bound we should send.
- [x] Test compatibility with Eclair (new negotiation)
- [x] Test compatibility with LND (old negotiation)

Future:
* Once anchor commitments are in, we should get rid of some constraints (marked with TODOs) like the limiting fees to the current ctx fees, as they can be set lower than the mempool tip.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -51,6 +51,7 @@ from .lnrouter import fee_for_edge_msat
 from .lnutil import ln_dummy_address
 from .json_db import StoredDict
 from .invoices import PR_PAID
+from .simple_config import FEE_LN_ETA_TARGET
 
 if TYPE_CHECKING:
     from .lnworker import LNGossip, LNWallet
@@ -1840,30 +1841,68 @@ class Peer(Logger):
         assert our_scriptpubkey
         # estimate fee of closing tx
         our_sig, closing_tx = chan.make_closing_tx(our_scriptpubkey, their_scriptpubkey, fee_sat=0)
-        fee_rate = self.network.config.fee_per_kb()
-        our_fee = fee_rate * closing_tx.estimated_size() // 1000
+        fee_rate_per_kb = self.network.config.eta_target_to_fee(FEE_LN_ETA_TARGET)
+        if not fee_rate_per_kb:  # fallback
+            fee_rate_per_kb = self.network.config.fee_per_kb()
+        our_fee = fee_rate_per_kb * closing_tx.estimated_size() // 1000
+        # TODO: anchors: remove this, as commitment fee rate can be below chain head fee rate?
         # BOLT2: The sending node MUST set fee less than or equal to the base fee of the final ctx
         max_fee = chan.get_latest_fee(LOCAL if is_local else REMOTE)
         our_fee = min(our_fee, max_fee)
-        drop_to_remote = False
+
+        drop_to_remote = False  # does the peer drop its to_local output or not?
         def send_closing_signed():
+            MODERN_FEE = True
+            if MODERN_FEE:
+                nonlocal fee_range_sent  # we change fee_range_sent in outer scope
+                fee_range_sent = fee_range
+                closing_signed_tlvs = {'fee_range': fee_range}
+            else:
+                closing_signed_tlvs = {}
+
             our_sig, closing_tx = chan.make_closing_tx(our_scriptpubkey, their_scriptpubkey, fee_sat=our_fee, drop_remote=drop_to_remote)
-            self.send_message('closing_signed', channel_id=chan.channel_id, fee_satoshis=our_fee, signature=our_sig)
+            self.logger.info(f"Sending fee range: {closing_signed_tlvs} and fee: {our_fee}")
+            self.send_message(
+                'closing_signed',
+                channel_id=chan.channel_id,
+                fee_satoshis=our_fee,
+                signature=our_sig,
+                closing_signed_tlvs=closing_signed_tlvs,
+            )
         def verify_signature(tx, sig):
             their_pubkey = chan.config[REMOTE].multisig_key.pubkey
             preimage_hex = tx.serialize_preimage(0)
             pre_hash = sha256d(bfh(preimage_hex))
             return ecc.verify_signature(their_pubkey, sig, pre_hash)
+
+        # this is the fee range we initially try to enforce
+        # we aim at a fee between next block inclusion and some lower value
+        fee_range = {'min_fee_satoshis': our_fee // 2, 'max_fee_satoshis': our_fee * 2}
+        their_fee = None
+        fee_range_sent = {}
+        is_initiator = chan.constraints.is_initiator
         # the funder sends the first 'closing_signed' message
-        if chan.constraints.is_initiator:
+        if is_initiator:
             send_closing_signed()
+
         # negotiate fee
         while True:
-            # FIXME: the remote SHOULD send closing_signed, but some don't.
-            cs_payload = await self.wait_for_message('closing_signed', chan.channel_id)
+            try:
+                cs_payload = await self.wait_for_message('closing_signed', chan.channel_id)
+            except asyncio.exceptions.TimeoutError:
+                if not is_initiator and not their_fee:  # we only force close if a peer doesn't reply
+                    self.lnworker.schedule_force_closing(chan.channel_id)
+                    raise Exception("Peer didn't reply with closing signed, force closed.")
+                else:
+                    # situation when we as an initiator send a fee and the recipient
+                    # already agrees with that fee, but doens't tell us
+                    raise Exception("Peer didn't reply, probably already closed.")
+
+            their_previous_fee = their_fee
             their_fee = cs_payload['fee_satoshis']
-            if their_fee > max_fee:
-                raise Exception(f'the proposed fee exceeds the base fee of the latest commitment transaction {is_local, their_fee, max_fee}')
+
+            # 0. integrity checks
+            # determine their closing transaction
             their_sig = cs_payload['signature']
             # verify their sig: they might have dropped their output
             our_sig, closing_tx = chan.make_closing_tx(our_scriptpubkey, their_scriptpubkey, fee_sat=their_fee, drop_remote=False)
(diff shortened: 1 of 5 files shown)


electrum/lnpeer.py
```diff
@@ -1769,29 +1769,69 @@ async def _shutdown(self, chan: Channel, payload, *, is_local: bool):
...
         # negotiate fee
         while True:
-            # FIXME: the remote SHOULD send closing_signed, but some don't.
-            cs_payload = await self.wait_for_message('closing_signed', chan.channel_id)
+            try:
+                cs_payload = await self.wait_for_message('closing_signed', chan.channel_id)
+            except asyncio.exceptions.TimeoutError:
+                if not is_initiator and not their_fee:  # we only force close if a peer doesn't reply
+                    await self.lnworker.force_close_channel(chan.channel_id)
+                    raise Exception("Peer didn't reply with closing signed, force closed.")
```
ecdsa, line 1820: Note: the BOLT does not say that we have to force close. 
I suppose this choice follows from https://github.com/lightning/bolts/pull/847/files#r689977518

ecdsa: Note 2: `_shutdown` is called from two separate places, one of them is is a `Taskgroup`.
In that case, `_shutdown` could be cancelled if the remote closes the connection, and might not have time to force close.

This way of calling a task from non-homogeneous contexts seems dangerous. This comment is not about this specific pull request, but about how we expect code to be executed. A similar issue was encountered in `request_force_close_from_backup`

electrum/lnpeer.py
```diff
@@ -1769,29 +1769,69 @@ async def _shutdown(self, chan: Channel, payload, *, is_local: bool):
         # estimate fee of closing tx
         our_sig, closing_tx = chan.make_closing_tx(our_scriptpubkey, their_scriptpubkey, fee_sat=0)
         fee_rate = self.network.config.fee_per_kb()
-        our_fee = fee_rate * closing_tx.estimated_size() // 1000
+        our_fee = fee_rate * closing_tx.estimated_size() // 1000  # inclusion in one block
```
ecdsa, line 1772: Actually, no, `fee_rate` is not defined as inclusion in one block. It is the fee rate resulting from user preferences.
Note that it differs from the fee rate used in commitment transactions. which uses a dynamic fee with ETA of 2 blocks.

ecdsa:
>  Clarify what lower bound we should send. We send `max_fee // 2` right now.

That is not what this PR is doing, it is sending the interval `[our_fee//2, our_fee]`, where `our_fee` results from user preferences, and is capped by `max_fee`. `max_fee` is the fee from the commitment tx.

ecdsa:
I think that our current use of user preferences in order to set `fee_rate` should have been complemented with displaying a fee slider in the "close channel" dialog. Otherwise, it might be better to simply use the ETA, as we do with commitment transactions.

electrum/lnpeer.py
```diff
@@ -1769,29 +1769,69 @@ async def _shutdown(self, chan: Channel, payload, *, is_local: bool):
...
+                    # already agrees with that fee, but doens't tell us
+                    raise Exception("Peer didn't reply, probably already closed.")
+
+            their_previous_fee = their_fee
             their_fee = cs_payload['fee_satoshis']
+
+            # 0. integrity checks
+            # TODO: anchors, remove this?
             if their_fee > max_fee:
                 raise Exception(f'the proposed fee exceeds the base fee of the latest commitment transaction {is_local, their_fee, max_fee}')
```
ecdsa, line 1832: Actually, the `max_fee` capping can be removed already, according to the BOLT.  It does not matter that our channels are not using anchor outputs. We can keep using this max value for our own interval, but we should not raise if the remote violates it.

ecdsa:
Regarding the choice of values, I think we could use the same fee rate as in the latest commitment. 
`our_fee` would be lower than `max_fee` because the close tx is smaller than the commitment tx. 
For the interval, I guess we could use `[our_fee//2, our_fee*2] `
Later we can propose a fee slider for `our_fee` with that range.

electrum/lnpeer.py
bitromortac: I see, but I also don't recall why I added the force close here, what would you suggest?

bitromortac:
> Does the regtest test bring something more than the already existing unit tests?

I think I was able to catch a bug by adding it, but we could remove it.

> > Clarify what lower bound we should send. We send `max_fee // 2` right now.
> 
> That is not what this PR is doing, it is sending the interval `[our_fee//2, our_fee]`, where `our_fee` results from user preferences, and is capped by `max_fee`. `max_fee` is the fee from the commitment tx.

That may be an old comment for the specific value, it was just referring to the range definition problem.

> Regarding the choice of values, I think we could use the same fee rate as in the latest commitment (that is, until anchor outputs are supported).
> `our_fee` would be lower than `max_fee` because the close tx is smaller than the commitment tx.
> For the interval, I guess we could use `[our_fee//2, our_fee*2] `
> Later we can propose a fee slider for `our_fee` with that range.

I switched to ETA of 2 blocks, the fee interval seems reasonable, but I have to think about this again, as already some time has passed since the PR.

electrum/lnpeer.py
bitromortac: Ok, I removed it.

electrum/lnpeer.py
ecdsa: Sorry, I had not seen the new definition of "fail the channel". Indeed, we have to force close.
Maybe this should be using try_force_closing, as in #7645.

ecdsa: we don't need to wait for #7645. See the new API, `schedule_force_closing` in 556b98736e78eb04bc68fd44e701e3fa174023a9

ecdsa: Also, the BOLT does not make a distinction between funder and non-funder. I think we have to force close in both cases.

ecdsa:
I will try to rewrite the fee negotiation, I think it needs to be broken down into smaller functions

electrum/lnpeer.py
bitromortac: > we don't need to wait for #7645. See the new API, `schedule_force_closing` in [556b987](https://github.com/spesmilo/electrum/commit/556b98736e78eb04bc68fd44e701e3fa174023a9)

I'll use it, but I meant that we introduce a wrapper `lnpeer.schedule_force_closing` that is added and could be used here as well.

> Also, the BOLT does not make a distinction between funder and non-funder. I think we have to force close in both cases.

My motivation was to catch the case where we would have sent an initial `closing_signed` message (as the initiator) and wait for the first response of the peer, if it doesn't, force close. I haven't tackled the case when the peer doesn't reply during a general step of the negotiation, of course we could close there as well.

electrum/lnpeer.py
```diff
@@ -1885,17 +1924,93 @@ def verify_signature(tx, sig):
                 to_remote_amount = closing_tx.outputs()[to_remote_idx].value
                 transaction.check_scriptpubkey_template_and_dust(their_scriptpubkey, to_remote_amount)
 
-            # Agree if difference is lower or equal to one (see below)
-            if abs(our_fee - their_fee) < 2:
+            # 1. check fees
+            # if fee_satoshis is equal to its previously sent fee_satoshis:
+            if our_fee == their_fee:
+                # SHOULD sign and broadcast the final closing transaction.
+                break  # we publish
```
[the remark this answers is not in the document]
bitromortac: I see that there could be a problem, but in this case I interpreted it that way that the peer would know that we have agreement and that we would publish (that's why we have the else clause in the timeout scenario). It seems like there's no instruction in the spec that we should reply again with `closing_signed`. We could send it for safety reasons.

bitromortac:
> I will try to rewrite the fee negotiation, I think it needs to be broken down into smaller functions

Ok sure, the spec is kind of complicated and I tried to follow it as closely as possible. I have even simulated the closing negotiation in a simulated environment with two asynchronous peers.

electrum/lnpeer.py
```diff
@@ -1885,17 +1924,98 @@ def verify_signature(tx, sig):
...
-            # another round
-            send_closing_signed()
-        # the non-funder replies
-        if not chan.constraints.is_initiator:
+            # otherwise:
+            else:
+                # MUST propose a value "strictly between" the received fee_satoshis and its previously-sent fee_satoshis.
+                our_fee_proposed = (our_fee + their_fee) // 2
+                if not (min(our_fee, their_fee) < our_fee_proposed < max(our_fee, their_fee)):
+                    our_fee_proposed += (their_fee - our_fee) // 2
```
ecdsa, line 2009: I do not understand what this code is doing.
First, I suppose that the intent is to set `our_fee`, not `our_fee_proposed`. 
So, maybe you wanted to write something like `our_fee = our_fee_proposed + (their_fee - our_fee)//2`
Second, if `abs(our_fee - their_fee) > 1`, then `(our_fee + their_fee) // 2` will always be strictly between the values, as explained in the comment on line 1892. Therefore, that code seems unnecessary.

electrum/lnpeer.py
```diff
@@ -1885,17 +1924,98 @@ def verify_signature(tx, sig):
...
+            # 3. if fee_satoshis matches its previously sent fee_range:
+            if fee_range_sent and (fee_range_sent['min_fee_satoshis'] <= their_fee <= fee_range_sent['max_fee_satoshis']):
+                # SHOULD reply with a closing_signed with the same fee_satoshis value if it is different from its previously sent fee_satoshis
+                if our_fee != their_fee:
+                    our_fee = their_fee
+                    send_closing_signed()  # peer publishes
+                    break
+                # SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
+                else:
+                    our_fee = their_fee
```
ecdsa, line 1953: this assignment is not needed..

electrum/lnpeer.py
bitromortac: yes, you're right, I think I experienced an off-by-one somewhere here and tried to fix it that way

electrum/lnpeer.py
bitromortac: that's true, I was translating the spec proverbially (`SHOULD use fee_satoshis`)

ecdsa:
closing in favor of #7680

PR #7590, opened 2021-12-09
trezorlib 0.13 compatibility

matejcik:
minor compatibility tweaks, plus Taproot support (looking to the future)

summary of changes:
* trezorlib now recognizes protobuf required fields, and so some values are needed at construction time. So there's a reorganization of how `TxInputType` and `TxOutputType` instances are created.
* `trezorlib.btc.sign_tx` accepts kwargs instead of `details`
* the signature of `button_request` is changed, it now accepts the full `ButtonRequest` object instead of just code

tested for normal and multisig wallets, seems to work fine.

Draft because trezorlib 0.13 is not out yet.

### DIFF
--- a/contrib/requirements/requirements-hw.txt
+++ b/contrib/requirements/requirements-hw.txt
@@ -1,5 +1,5 @@
 hidapi
-trezor[hidapi]>=0.12.0
+trezor[hidapi]>=0.13.0,<0.14
 safet>=0.1.5
 keepkey>=6.3.1
 btchip-python>=0.1.32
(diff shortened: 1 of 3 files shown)


matejcik:
...and it's out

prusnak:
@matejcik Can you please pull in the extra commit from https://github.com/trezor/electrum/tree/trezorlib-0.13 (use the same amount unit (satoshi, etc.) on device) into your PR branch?

matejcik:
modified to reflect configuration changes and pushed as 1e8e289

SomberNight (approved):
Looks good. Thanks.

PR #7592, opened 2021-12-10
Add pow_hash_header abstraction

JeremyRand:
There is no reason why the hash function for identifying a block and the hash function used to instantiate Hashcash must be the same; it's only a coincidence that Bitcoin happens to use the same hash for both use cases.  Reflecting this fact by adding this abstraction makes the code more flexible.

(As discussed on IRC, this is particularly desirable for downstream projects that use different PoW functions, such as Namecoin.)

### DIFF
--- a/electrum/blockchain.py
+++ b/electrum/blockchain.py
@@ -86,6 +86,9 @@ def hash_raw_header(header: str) -> str:
     return hash_encode(sha256d(bfh(header)))
 
 
+pow_hash_header = hash_header
+
+
 # key: blockhash hex at forkpoint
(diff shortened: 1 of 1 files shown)


JeremyRand:
@SomberNight Let me know if there's a better way to do this; I think this is roughly in line with what we talked about on IRC but I am very much open to suggestions.

JeremyRand:
Rebased to fix merge conflict.

@SomberNight Any chance you could take a look at this soon?

SomberNight:
Looks okay. Thanks.

PR #7594, opened 2021-12-10
Produce packager-friendly tarballs

### LINKED ISSUE #7479: Produce packager-friendly tarballs
[Debian policy](https://wiki.debian.org/UpstreamGuide#No_inclusion_of_third_party_code) states the following recommendation for upstream tarballs:

> Please do not include other code (like libraries) or data that are also shipped separately inside your source archive

[Debian policy](https://wiki.debian.org/UpstreamGuide#Pristine_Upstream_Source) also states:

> The source code archive should contain source files only, never any files that are automatically generated from other files.

[Debian policy](https://wiki.debian.org/UpstreamGuide#Pristine_Upstream_Source) also states:

> we need to rebuild all generated files to make sure that they can really be built from source, so please include methods to regenerate these files

In order to comply with Debian policy, we should distribute a packager-friendly tarball with the following changes relative to the current tarballs:

* Do not include the `packages` folder.
* Do not include `paymentrequest_pb2.py`.
* Include a script in `contrib` to generate `paymentrequest_pb2.py` from `paymentrequest.proto`.

I'm happy to implement this if it gets a Concept ACK.

yanmaani:
This PR adds a series of changes that are necessary to produce "clean" tarballs, without generated files, dependencies, etc.

To `make_sdist.sh`, three options (passed by environment variable) are added:
* `NO_INSTALL_PACKAGES` (do not run the scripts for installation of packages)
* `NO_BUILD_LOCALE` (omit the compiled locale files)
* `OMIT_PB2_GENERATED` (omit the file `paymentrequest_pb2.py`)

By default, nothing changes. If the options are unset, `make_sdist.sh` behaves exactly as before. This is all opt-in.

There is also a slight refactor of the same script, where some locale generation code moves into `contrib/build_locale.sh`. This is necessary to be able to invoke this script separately, e.g. after unpacking a tarball.

Finally, a script is added to generate the file `paymentrequest_pb2.py`. This replaces the manual advice in `electrum/paymentrequest.py`.

Closes #7479.

### DIFF
--- a/contrib/build-linux/sdist/make_sdist.sh
+++ b/contrib/build-linux/sdist/make_sdist.sh
@@ -46,7 +46,25 @@ fi
     find -exec touch -h -d '2000-11-11T11:11:11+00:00' {} +
 
     # note: .zip sdists would not be reproducible due to https://bugs.python.org/issue40963
-    TZ=UTC faketime -f '2000-11-11 11:11:11' python3 setup.py --quiet sdist --format=gztar
+    if ([ "$OMIT_UNCLEAN_FILES" = 1 ])
+        then PY_DISTDIR="dist/_sourceonly" # The DISTDIR variable of this script is only used to find where the output is *finally* placed.
+        else PY_DISTDIR="dist"
+    fi
+    TZ=UTC faketime -f '2000-11-11 11:11:11' python3 setup.py --quiet sdist --format=gztar --dist-dir="$PY_DISTDIR"
+    if ([ "$OMIT_UNCLEAN_FILES" = 1 ]); then
+        for fn in "$DISTDIR/_sourceonly/"*; do
+            # Since ELECTRUM_VERSION is not available to us in this script, we have to use a regex.
+            # Expression 1: Electrum-X.Y.Z.tar.gz -> Electrum-sourceonly-X.Y.Z.tar.gz
+            #   Capture group \1 = Electrum
+            #   Capture group \2 = X.Y.Z.tar.gz
+            # Expression 2: dist/_sourceonly/X.tar.gz -> dist/X.tar.gz
+            mv "$fn" $(sed \
+                -e 's/\(.*\)-\([^-]*\)/\1-sourceonly-\2/' \
+                -e 's/\/_sourceonly//' \
+                <<< "$fn")
+        done
+        rmdir "$PY_DISTDIR"
+    fi
 )
 
 
--- a/contrib/build-linux/sdist/make_sdist.sh
+++ b/contrib/build-linux/sdist/make_sdist.sh
@@ -19,18 +19,26 @@ break_legacy_easy_install
 # (make_packages will later install a pinned version of pip in a venv)
 python3 -m pip install --upgrade pip
 
-"$CONTRIB"/make_packages || fail "make_packages failed"
+if ([ "$OMIT_UNCLEAN_FILES" != 1 ]); then
+  "$CONTRIB"/make_packages || fail "make_packages failed"
+fi
 
 git submodule update --init
 
 (
-    # We include both source (.po) and compiled (.mo) locale files in the source dist.
-    # Maybe we should exclude the compiled locale files? see https://askubuntu.com/a/144139
-    # (also see MANIFEST.in)
+    # By default, include both source (.po) and compiled (.mo) locale files in the source dist.
+    # Set option OMIT_UNCLEAN_FILES=1 to exclude the compiled locale files
+    # see https://askubuntu.com/a/144139 (also see MANIFEST.in)
     rm -rf "$LOCALE"
     cp -r "$CONTRIB/deterministic-build/electrum-locale/locale/" "$LOCALE/"
-    "$CONTRIB/build_locale.sh" "$LOCALE"
+    if ([ "$OMIT_UNCLEAN_FILES" != 1 ]); then
+      "$CONTRIB/build_locale.sh" "$LOCALE"
+    fi
 )
+
+if ([ "$OMIT_UNCLEAN_FILES" = 1 ]); then
+  rm "$PROJECT_ROOT/electrum/paymentrequest_pb2.py"
+fi
 
 (
     cd "$PROJECT_ROOT"
--- a/contrib/build-linux/sdist/make_sdist.sh
+++ b/contrib/build-linux/sdist/make_sdist.sh
@@ -6,7 +6,7 @@ PROJECT_ROOT="$(dirname "$(readlink -e "$0")")/../../.."
 CONTRIB="$PROJECT_ROOT/contrib"
 CONTRIB_SDIST="$CONTRIB/build-linux/sdist"
 DISTDIR="$PROJECT_ROOT/dist"
-LOCALE="$PROJECT_ROOT/electrum/locale/"
+LOCALE="$PROJECT_ROOT/electrum/locale"
 
 . "$CONTRIB"/build_tools_util.sh
 
@@ -24,21 +24,12 @@ python3 -m pip install --upgrade pip
 git submodule update --init
 
 (
-    cd "$CONTRIB/deterministic-build/electrum-locale/"
-    if ! which msgfmt > /dev/null 2>&1; then
-        echo "Please install gettext"
-        exit 1
-    fi
     # We include both source (.po) and compiled (.mo) locale files in the source dist.
     # Maybe we should exclude the compiled locale files? see https://askubuntu.com/a/144139
     # (also see MANIFEST.in)
     rm -rf "$LOCALE"
-    for i in ./locale/*; do
-        dir="$PROJECT_ROOT/electrum/$i/LC_MESSAGES"
-        mkdir -p "$dir"
-        msgfmt --output-file="$dir/electrum.mo" "$i/electrum.po" || true
-        cp $i/electrum.po "$PROJECT_ROOT/electrum/$i/electrum.po"
-    done
+    cp -r "$CONTRIB/deterministic-build/electrum-locale/locale/" "$LOCALE/"
+    "$CONTRIB/build_locale.sh" "$LOCALE"
 )
 
 (
--- a/contrib/build_locale.sh
+++ b/contrib/build_locale.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+if [ ! -d "$1" ]; then
+    echo "usage: $0 path/to/locale"
+    exit 1
+fi
+
+if ! which msgfmt > /dev/null 2>&1; then
+    echo "Please install gettext"
+    exit 1
+fi
+
+for i in "$1/"*; do
+  mkdir -p "$i/LC_MESSAGES"
+  (msgfmt --output-file="$i/LC_MESSAGES/electrum.mo" "$i/electrum.po" || true)
+done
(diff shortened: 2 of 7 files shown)


JeremyRand:
I would recommend collapsing the new environment variables into a single one.  I'm not aware of any use case for only using a subset of them, and using a single one will be easier for downstream projects that have additional steps (e.g. Namecoin's Qt form build step).

JeremyRand:
It would be desirable for `sdist/build.sh` to support this.  Ditto for Cirrus.

yanmaani:
What do you mean by adding support to Cirrus? Should it be a default option? The whole idea of Cirrus is it runs unattended, right?

The rest should be fine, give me 5 minutes...

yanmaani:
@JeremyRand Ready for review.

JeremyRand:
> What do you mean by adding support to Cirrus? Should it be a default option? The whole idea of Cirrus is it runs unattended, right?

@yanmaani I'd suggest adding a `matrix` clause to the `tarball build` Cirrus task, similar to what the `Flake8` task does.

JeremyRand:
Also I'd suggest adding this to `contrib/release.sh`, i.e. running the release.sh script should produce both a source-only tarball and a regular tarball.

yanmaani:
Review comments have been addressed.

JeremyRand:
Needs rebase.

contrib/build_locale.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
SomberNight, line 1: Would be nice if the expected arg was checked for being present and validated.

I did not check if this works, but something like this:  
```sh
if [ -z "$1" ] || [ ! -d "$1" ]; then
    echo "usage: $0 locale_path"
    exit 1
fi
```

contrib/release.sh
```diff
@@ -88,6 +88,14 @@ else
    ./contrib/build-linux/sdist/build.sh
 fi
 
+# create source tarball
+srctarball="Electrum-source-$VERSION.tar.gz"
```
SomberNight, line 92: Please see how `$tarball` is used in the rest of the file, and add the analogous code.
In particular, the non-releasemanager codepath needs to compare that they built the same distributable as the releasemanager.

contrib/build-linux/sdist/make_sdist.sh
```diff
@@ -6,45 +6,44 @@ PROJECT_ROOT="$(dirname "$(readlink -e "$0")")/../../.."
...
 
-info "Installing pinned requirements."
-python3 -m pip install --no-dependencies --no-warn-script-location -r "$CONTRIB"/deterministic-build/requirements-build-sdist.txt
+  info "Installing pinned requirements."
+  python3 -m pip install --no-dependencies --no-warn-script-location -r "$CONTRIB"/deterministic-build/requirements-build-sdist.txt
 
 
-"$CONTRIB"/make_packages || fail "make_packages failed"
+  "$CONTRIB"/make_packages || fail "make_packages failed"
+fi
```
SomberNight, line 28: What's going on here, why is this whole block guarded by that envvar? Presumably you only want to guard the `make_packages` call?

If `pip install deterministic-build/requirements-build-sdist.txt` is not run, is the build even reproducible?
On line53, `python3 setup.py sdist` is run, and the generated tar.gz might very well depend on the setuptools version used.

contrib/release.sh
SomberNight, line 92:
```suggestion
# create source-only tarball
srctarball="Electrum-sourceonly-$VERSION.tar.gz"
```
I mean, the other tarball is pretty-much a source tarball too.
This one has removed
- generated code, which is source
- bundled dependencies, which are source-only
- and ok, fair enough, compiled locale

I would rather this had a name that gives at least a hint to users that this is not what they want. Maybe let's call it `sourceonly` then...

SomberNight (commented):
Thanks for the PR. See comments inline.

contrib/release.sh
JeremyRand: No objection to `sourceonly` from my end.

contrib/build-linux/sdist/make_sdist.sh
```diff
@@ -6,45 +6,44 @@ PROJECT_ROOT="$(dirname "$(readlink -e "$0")")/../../.."
...
-    cd "$CONTRIB/deterministic-build/electrum-locale/"
-    if ! which msgfmt > /dev/null 2>&1; then
-        echo "Please install gettext"
-        exit 1
-    fi
-    # We include both source (.po) and compiled (.mo) locale files in the source dist.
-    # Maybe we should exclude the compiled locale files? see https://askubuntu.com/a/144139
-    # (also see MANIFEST.in)
+    # By default, include both source (.po) and compiled (.mo) locale files in the source dist.
+    # Set option NO_BUILD_LOCALE=1 to exclude the compiled locale files
```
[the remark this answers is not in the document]
yanmaani: What do you mean?

contrib/release.sh
yanmaani: Sure.

yanmaani: Adding the analogous code that applies for `$tarball` as well as for `$appimage`.

contrib/build_locale.sh
yanmaani: Sure. I don't see the need of the first check - `''` is never a valid directory name anyway.

contrib/build-linux/sdist/make_sdist.sh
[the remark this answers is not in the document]
yanmaani: Oh yeah, duh.

yanmaani:
Sorry I didn't see this earlier. Anyway the necessary changes have been made and I have rebased on top of master.

contrib/release.sh
SomberNight: It is not sufficient to modify this string, the file should be named as such in the first place :)

The sdist is created here:
https://github.com/spesmilo/electrum/blob/cbe1f3fd807b0bc673ad1b8a2efd86856644a629/contrib/build-linux/sdist/make_sdist.sh#L49

and look at e.g.:
https://cirrus-ci.com/task/6247100135505920?logs=build#L31
```
💬 INFO:  done.
total 8216
drwxr-xr-x 2 user user    4096 Jul  4 12:43 .
drwxr-xr-x 9 user user    4096 Jul  4 12:43 ..
-rw-r--r-- 1 user user 8401137 Jul  4 12:43 Electrum-4.3.0a0.tar.gz
c733079846670b507289b2049125ee76491905b72d87373c4e0fb6bbaa41378e  /tmp/cirrus-ci-build/contrib/build-linux/sdist/../../../dist/Electrum-4.3.0a0.tar.gz
```

I suppose you should branch on `$OMIT_UNCLEAN_FILES` in `make_sdist.sh` to rename the file to e.g. `Electrum-sourceonly-4.3.0a0.tar.gz`.
However, it is not clear to me how exactly to do this. We are calling `python3 setup.py sdist` - I don't think there's an arg we could pass to customise the name. Maybe we can determine what the output name is and simply rename the file. (but even then, as that would not be atomic, note that there can be headaches with the sourceonly tar.gz overwriting the non-sourceonly one...)

yanmaani: Oh yeah, whoops.

I tried doing a gross hack, but this also changed the hash of the file. Do we want for `Electrum-sourceonly-4.3.0a0.tar.gz` to contain the directory `Electrum-4.3.0a0/` or `Electrum-sourceonly-4.3.0a0/`? And is the version important?

If the only thing that's to be changed is the filename, I suppose you could just make it emit the output tarball into a different directory? Like:

```s_o -> a/Electrum-4.3.0a0.tar.gz
normal -> Electrum-4.3.0a0.tar.gz
a/Electrum-4.3.0a0.tar.gz -> Electrum-sourceonly-4.3.0a0.tar.gz.```

This isn't atomic, but it shouldn't have overwriting issues at least.

Comment (SomberNight) [reply contrib/release.sh]: > Do we want for `Electrum-sourceonly-4.3.0a0.tar.gz` to contain the directory `Electrum-4.3.0a0/` or `Electrum-sourceonly-4.3.0a0/`?

I don't mind, either is ok.

> And is the version important?

At first I did not understand what you mean by that but then I looked at your "gross hack" in https://github.com/spesmilo/electrum/pull/7594/commits/9edcc427320cfe7babf285d89ad6c9c53b30ef9a ... :P
I've just tested a build, and while it works, I am getting warnings:
```
/home/user/.local/lib/python3.8/site-packages/setuptools/dist.py:530: UserWarning: The version specified ('sourceonly-4.3.0a0') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  warnings.warn(
/home/user/.local/lib/python3.8/site-packages/pkg_resources/__init__.py:123: PkgResourcesDeprecationWarning: sourceonly-4.3.0a0 is an invalid version and will not be supported in a future release
  warnings.warn(
```
So let's not do this.

> If the only thing that's to be changed is the filename, I suppose you could just make it emit the output tarball into a different directory? [...]
> This isn't atomic, but it shouldn't have overwriting issues at least.

That sounds fine. Better than hacking the `version` arg...
I think we can use the `--dist-dir` arg, e.g.:
```
python3 setup.py --quiet sdist --format=gztar --dist-dir=dist/sourceonly/
```
And then rename and move the file, like you said. I guess also delete the `dist/sourceonly/` folder after.

File: contrib/build-linux/sdist/make_sdist.sh [hunk 5]
```diff
@@ -46,7 +46,8 @@ fi
     find -exec touch -h -d '2000-11-11T11:11:11+00:00' {} +
 
     # note: .zip sdists would not be reproducible due to https://bugs.python.org/issue40963
-    TZ=UTC faketime -f '2000-11-11 11:11:11' python3 setup.py --quiet sdist --format=gztar
+    [ "$OMIT_UNCLEAN_FILES" = 1 ] && SUFFIX='sourceonly-' || SUFFIX=''
```
SomberNight, line 49: please do an `if` - `else` - `fi` instead, I find that much easier to read.

yanmaani:
Right, the necessary changes have been made, should be ready for review. It now produces a distinctly named tarball without the use of any gross hacks.

I hope the use of `sed` when renaming isn't gratuitous, I couldn't think of a better way.

contrib/release.sh
yanmaani: Sure. I used the folder name `_sourceonly` instead, to clarify that it's meant to be ephemeral. (I guess I could've gone harder on putting this inside of a variable and stuff, but IMO it would've harmed readability more than it'd have helped.)

contrib/build-linux/sdist/make_sdist.sh
yanmaani: Done.

SomberNight:
> I hope the use of sed when renaming isn't gratuitous, I couldn't think of a better way.

I've replaced that with a similarly convoluted python script, but I find that easier to understand than the regex.

-----

merged in https://github.com/spesmilo/electrum/commit/d0de44a794be43572fce73cdbe18e60c07a3274d
including follow-up in https://github.com/spesmilo/electrum/commit/7d36f2ba3ab603044556b9548bea9893520747e1

Thank you for the PR.

SomberNight:
The 4.3.1 release now includes source-only tarballs.
We have decided not to link to them directly from the [website (html table)](https://electrum.org/#download), but they are available on the [download server](https://download.electrum.org/4.3.1/).

Issue #7593, opened 2021-12-10
Cirrus Regtest functional tests failing


JeremyRand:
The Regtest functional tests on Cirrus are hanging in an infinite loop at `wait until alice sees channel open`.  Example run: https://github.com/spesmilo/electrum/runs/4241116855

It's not obvious to me why the failure is happening.   Not sure if it's Lightning-specific.

JeremyRand:
It might be desirable to refactor the functional tests so that output from Electrum gets logged to Cirrus, similarly to how ElectrumX output is logged.  That would make it easier to debug this kind of failure.

PR #7609, opened 2021-12-28
Cirrus: Use VM instead of Docker for functional tests

JeremyRand:
Electrum seems to encounter bizarre network connectivity errors (in the form of failed syscalls) when run in a Docker container on Cirrus.  This was causing the functional tests on Cirrus to fail, presumably because some of Electrum's Lightning-related socket syscalls were failing.  Switching that task to run in a VM instead of a Docker container fixes the failure.

I am not sure whether this is a Cirrus-specific issue or if Electrum has these issues in Docker in general.  There is also some nondeterministic behavior in the functional tests that can cause them to sometimes hang; that is out of scope for this PR, but the easy workaround is to restart the affected Cirrus task.

I also fixed the ElectrumX version at v1.15.0, since Electrum master is using the BSV ElectrumX repo (which obviously won't work), and the latest spesmilo/electrumx release has broken wheels.

Fixes https://github.com/spesmilo/electrum/issues/7593

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -108,7 +108,8 @@ task:
     - apt-get update
     - apt-get -y install libsecp256k1-0 curl jq bc
     - pip3 install .[tests]
-    - pip3 install electrumx
+    #- pip3 install e-x # Broken because of https://github.com/spesmilo/electrumx/issues/117 , use older version for now.
+    - pip3 install electrumx==1.15.0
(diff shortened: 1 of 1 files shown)


JeremyRand:
>  There is also some nondeterministic behavior in the functional tests that can cause them to sometimes hang; that is out of scope for this PR, but the easy workaround is to restart the affected Cirrus task.

For the record, the functional tests task in this PR required 1 restart due to this.

SomberNight:
Thank you.

PR #7623, opened 2022-01-10
Multi-trampoline multipart payments

### LINKED ISSUE #7604: Enable multipart trampoline payments via several different trampoline nodes
Since https://github.com/ACINQ/eclair/issues/1723 was merged since a while I think we could enable splitting payments via different trampoline nodes. Currently, this is prevented:
https://github.com/spesmilo/electrum/blob/bdbd59300fbd35b01605e66145458e5f396108e8/electrum/lnworker.py#L1549

On testnet, I have two channels, each with a different trampoline. I try to send an amount (400000 sat) that forces splitting (legacy multipart trampoline payment). On testnet, the above restriction is not active, so it should work to send via different trampoline nodes. The payment reaches the trampoline nodes, but at the receiver side (an LND instance) I get:
```
2021-12-20 16:05:50.486 [DBG] INVC: Invoice(pay_hash=XXX, pay_addr=XXX): failure resolution result outcome: set total too low for invoice, at accept height: 2132478, amt=270211462 mSAT, expiry=2132518, circuit=(Chan ID=XXX, HTLC ID=24), mpp=total=270211462 mSAT, addr=XXX, amp=<nil>
```

This means that the trampoline payments didn't include the correct `total_msat` value for the invoice, which is why it makes sense that the client receives `INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS`. From what I can see, is that Electrum includes the correct `total_msat` value, so it could be a bug in eclair, but I'm unsure.

Anyhow, enabling mpp trampoline payments would further increase robustness and usefulness having channels with different trampolines.

The trampoline payloads look like this for instance:
```
I | trampoline | payload 0 {'amt_to_forward': {'amt_to_forward': 261693790}, 'outgoing_cltv_value': {'outgoing_cltv_value': 2133093}, 'outgoing_node_id': {'outgoing_node_id': pubkeyT1}}
I | trampoline | payload 1 {'amt_to_forward': {'amt_to_forward': 261664624}, 'outgoing_cltv_value': {'outgoing_cltv_value': 2132517}, 'outgoing_node_id': {'outgoing_node_id': pubkeyT2}, 'invoice_features': {'invoice_features': <LnFeatures.BASIC_MPP_OPT|PAYMENT_SECRET_REQ|VAR_ONION_OPT: 147968>}, 'invoice_routing_info': {'invoice_routing_info': b''}, 'payment_data': {'payment_secret': XXX1, 'total_msat': 400000000}}
I | trampoline | payload 2 {'amt_to_forward': {'amt_to_forward': 261664624}, 'outgoing_cltv_value': {'outgoing_cltv_value': 2132517}, 'payment_data': {'payment_secret': XXX1, 'total_msat': 400000000}}
```

bitromortac:
~~Based on #7426 (first four commits) and~~ partially fixes #7604, related to #7648.

* Refactored `create_trampoline_route`.
* Enables end-to-end multi-trampoline multipart payments. Trampoline-to-legacy payments are still not enabled, as this is currently not supported by Eclair.
* Reintroduces a global trampoline fee level.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -80,7 +80,7 @@ from .channel_db import get_mychannel_info, get_mychannel_policy
 from .submarine_swaps import SwapManager
 from .channel_db import ChannelInfo, Policy
 from .mpp_split import suggest_splits
-from .trampoline import create_trampoline_route_and_onion, TRAMPOLINE_FEES
+from .trampoline import create_trampoline_route_and_onion, TRAMPOLINE_FEES, is_legacy_relay
 
 if TYPE_CHECKING:
     from .network import Network
@@ -1157,8 +1157,13 @@ class LNWallet(LNWorker):
                 raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON, data=b'')
 
         self.logs[payment_hash.hex()] = log = []
-        trampoline_fee_levels = defaultdict(lambda: self.INITIAL_TRAMPOLINE_FEE_LEVEL)  # type: DefaultDict[bytes, int]
-        use_two_trampolines = True # only used for pay to legacy
+
+        # when encountering trampoline forwarding difficulties in the legacy case, we
+        # sometimes need to fall back to a single trampoline forwarder, at the expense
+        # of privacy
+        use_two_trampolines = True
+
+        trampoline_fee_level = self.INITIAL_TRAMPOLINE_FEE_LEVEL
 
         amount_inflight = 0  # what we sent in htlcs (that receiver gets, without fees)
         while True:
@@ -1177,7 +1182,7 @@ class LNWallet(LNWorker):
                     full_path=full_path,
                     payment_hash=payment_hash,
                     payment_secret=payment_secret,
-                    trampoline_fee_levels=trampoline_fee_levels,
+                    trampoline_fee_level=trampoline_fee_level,
                     use_two_trampolines=use_two_trampolines,
                     fwd_trampoline_onion=fwd_trampoline_onion
                 )
@@ -1236,8 +1241,10 @@ class LNWallet(LNWorker):
                 #       instead we should give feedback to create_routes_for_payment.
                 if code in (OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT,
                             OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON):
-                    # todo: parse the node parameters here (not returned by eclair yet)
-                    trampoline_fee_levels[erring_node_id] += 1
+                    # TODO: parse the node policy here (not returned by eclair yet)
+                    # TODO: erring node is always the first trampoline even if second
+                    #  trampoline demands more fees, we can't influence this
+                    trampoline_fee_level += 1
                     continue
                 elif use_two_trampolines:
                     use_two_trampolines = False
@@ -1457,9 +1464,9 @@ class LNWallet(LNWorker):
             invoice_features: int,
             payment_hash,
             payment_secret,
-            trampoline_fee_levels: DefaultDict[bytes, int],
+            trampoline_fee_level: int,
             use_two_trampolines: bool,
-            fwd_trampoline_onion = None,
+            fwd_trampoline_onion=None,
             full_path: LNPaymentPath = None) -> AsyncGenerator[Tuple[LNPaymentRoute, int], None]:
 
         """Creates multiple routes for splitting a payment over the available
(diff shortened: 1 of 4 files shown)


bitromortac:
Rebased.

SomberNight:
> * Removes all second-hop trampoline forwarders for the legacy case, as trampoline failures are currently not handled properly, see (#7648), with the downside of removing some privacy (not much, there are only three trampolines), but the upside of improvements in sending reliability.
> * In the e2e case, we sometimes need two trampoline forwarders, e.g., when the two endpoints are connected to different trampolines. We cannot know about which of the two trampoline fee levels should be bumped on insufficient fee error, therefore we revert to a global fee level and bump both (which doubles the fees).

Why remove the second hop TF for the legacy case -- can't we just use the same approach as for e2e and revert to a global fee level?

bitromortac:
> Why remove the second hop TF for the legacy case -- can't we just use the same approach as for e2e and revert to a global fee level?

My impression is that currently it takes many attempts to make larger payments, with quite a few fee bumps involved. 

The is some dynamics between random choice of second forwarder and insufficient fees which gets doubled up if we consider a global policy. I don't know if the next makes fully sense (for say a 1Msat payment) or if there are other edge cases, there could be more fee bumps:

### two TF scenario (kind of worst case):
* Route: TF1 -> TF2 -> Receiver
Fee is insufficient for T1 -> T2, bump fee level
* Route TF1 -> TF3 -> Receiver
Fee is insufficient for T1 -> T3, bump fee level
* Route TF1 -> TF2 -> Receiver
Fee is insufficient for T2 -> Receiver, bump fee level
* Route TF1 -> TF3 -> Receiver
Fee is insufficient for TF3 -> Receiver, bump fee level

fee level: 4 -> 0.001000 * 1_000_000 sat = 1000 sat -> 2000 sat

### single TF scenario:
* Route: TF1 -> Receiver
Fee is insufficient for T1 -> Receiver, bump fee level
* Route: TF1 -> Receiver
Fee is insufficient for T1 -> Receiver, bump fee level

fee level: 2 -> 0.000100 * 1_000_000 sat = 100 sat

Of course there's a tradeoff between reliability, fees, payment time and privacy. I didn't make many large trampoline payments but from my experience they seemed rather unreliable.

ecdsa:
@bitromortac can you clarify in the example above, whether those are MPPs or sequential attempts?

bitromortac:
> @bitromortac can you clarify in the example above, whether those are MPPs or sequential attempts?

Those would be sequential attempts.

bitromortac:
I reverted to a global fee rate and added the second forwarder again for legacy payments.

electrum/lnworker.py
```diff
@@ -1238,11 +1236,11 @@ async def pay_to_node(
...
                             OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON):
-                    # todo: parse the node parameters here (not returned by eclair yet)
-                    trampoline_fee_levels[erring_node_id] += 1
+                    # TODO: parse the node policy here (not returned by eclair yet)
+                    # TODO: erring node is always the first trampoline even if second
+                    #  trampoline demands more fees, we can't influence this
+                    trampoline_fee_level += 1
                     continue
-                elif use_two_trampolines:
-                    use_two_trampolines = False
```
SomberNight, line 1245: I am trying to understand how exactly this fallback from two to one TFs works atm, and whether it is related to https://github.com/spesmilo/electrum/issues/7669.
AFAIU the logic here, if the failures were trampoline-feelevel-related, that is not cause in itself to fallback to one TF, but we would have fallen back if the first TF cannot find a route to the second one or if the second one cannot find a route to the destination, right? One thing to consider is given we only know of 3 TFs atm, it's plausible that some of the time  there aren't two TFs online. Although maybe the current code is falling back too soon -- say TF_Alice is offline and we tried to route through it, instead of immediately falling back to using one TF, we could still try selecting a new route of two TFs that excludes Alice.
Anyway, what is the motivation for removing this?

bitromortac: I forgot to add it back. Right, we could do better than immediately falling back to a single TF, in principle we could track whether a TF was able to deliver a certain amount to a specific target and then remove it from the list of possible second forwarders.

bitromortac: The fallback in the legacy case is added back now, hope it looks better now.

bitromortac:
I will explicitly test again an e2e payment to a wallet with a channel which has not enough receiving capacity with ACINQ and will let you know.

ecdsa:
I tested myself, it looks like the ACINQ node no longer tries `pay-to-open` on Electrum clients.

PR #7636, opened 2022-01-19
lightning: implement channel types

bitromortac:
Spec: https://github.com/lightning/bolts/pull/880/

This PR adds a TLV to `open_channel` and `accept_channel` messages. The initiator checks the peer's node features and selects a channel type that is supported by both peers, which signals the type of the channel to be opened explicitly. This lets us remove the `static_remotekey_enabled` field in the storage, which is now a IntFlag (`ChannelType`).

This should also be applied to #7509.

Tests:
- [x] open channel with legacy node (testnet)
- [x] open channel with node supporting `option_channel_type` (testnet)


ecdsa:
looks good to me.

Issue #6886, opened 2020-12-26
electrum --version


jingxi285:
could --version be added, make it easy to check installed version number

SomberNight:
There is a `version` command. You can run it as:
`$ electrum version`

jingxi285:
But that's non-standard.? It's trivial to at least allow the conventional command line as well.
Why not support regular --version ? you must have used 
gcc --version
cat --version
bash --version

SomberNight:
I guess we could support this.

Many programs indeed support `$ prog --version`, though some use `$ prog -version` and some `$ prog version`.
Still `--version` seems to be the most widespread.

SomberNight:
Feel free to open a PR.

johlar:
Are you still working on this @ritikramuka ? Let me know, because otherwise I have a PR ready :)

ritikramuka:
@johlar you can go ahead, I just started with something else.

PR #7662, opened 2022-02-16
Add support for cli --version command

johlar:
This PR addresses issue #6886 and adds support for the CLI command --version.

Considerations:
* If we were to use an alias instead, the alias (`--version`) is what would actually get stored in the parser instance, rather than the base name which is referred to (`version`), which then would be problematic because the CLI commands are bound to the names of their respective python functions.
* I think this approach is cleaner than adding one more check in `run_electrum/main()` which would substitute the arg `--version` with `version`. This would also have made the command undocumented in the help section.

This is my first contribution to Electrum and I'm pretty new to Python overall, so I very much welcome any feedback or clarifications if I have misunderstood anything!


SomberNight:
> Considerations: [...]

Sounds good.

But then we should instead add the option to the main parser only, instead of the (perhaps confusingly named) "global options". e.g.:
```patch
diff --git a/electrum/commands.py b/electrum/commands.py
index de1313dd67..dd99c9c41c 100644
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1457,6 +1457,7 @@ def get_parser():
     # create main parser
     parser = argparse.ArgumentParser(
         epilog="Run 'electrum help <command>' to see the help for a command")
+    parser.add_argument("--version", dest="cmd", action='store_const', const='version', help="Return the version of Electrum.")
     add_global_options(parser)
     add_wallet_option(parser)
     subparsers = parser.add_subparsers(dest='cmd', metavar='<command>')

```

Try running the following two commands:
```
$ ./run_electrum help
$ ./run_electrum help signmessage
```
The second returns help text specific to `signmessage`. The `--version` option should not be mentioned in that case.

johlar:
Thanks for your feedback, that's a very good point.

However adding the option directly to the main `parser` will (surprisingly) not work, because the default subparser will be the `gui_subparser`. I've amended the PR to address your point by adding the option to `gui_subparser`. This produces the expected result. The only downside I see is that this command will be documented under `help gui`.

Is this solution acceptable? A nice future non-breaking improvement would be to re-work how command functions are mapped to CLI commands in order to allow for proper aliases.

SomberNight:
> However adding the option directly to the main parser will (surprisingly) not work, because the default subparser will be the gui_subparser. 

Oh, right... How about special-casing `--version` here then?
https://github.com/spesmilo/electrum/blob/9bbf92b80b0892d87f04cfde63cc3f576e7e1c4a/electrum/commands.py#L1383

johlar:
Thanks, that was a better solution. It now works as expected :)

Issue #7660, opened 2022-02-15
message signature verification


massmux:
if creating a message signature with bluewallet or sparrow, electrum always returns signature invalid. i've tested both from the program and from the python libs.

massmux:
So at the moment it is not possible to verify bluewallet generated message signatures for p2wpkh addresses on electrum. is it correct? when do you predict this can be included in electrum?

Issue #3861, opened 2018-02-06
Segwit message signing is not compatible with other software


SomberNight:
AFAICT there is no clear consensus how to sign messages with segwit addresses (txin type of `p2wpkh-p2sh` and `p2wpkh`)

We do one thing, and Trezor does something else. Bitcoin Core so far has this disabled.
Maybe we could transition to what Trezor does, in a way that:
- signatures created by new Electrum
    - can be verified by new Electrum software, but not verified by old
- signatures created by old Electrum
    - can be verified by both new and old Electrum

-----

Electrum issue/commit
- https://github.com/spesmilo/electrum/issues/2977#issuecomment-350341771
- https://github.com/spesmilo/electrum/commit/e299df7b82f626f669098ccc0dbe041934d72d86

Bitcoin Core issue
 - https://github.com/bitcoin/bitcoin/issues/10542#issuecomment-316032523

Trezor issue/commit/tests
- https://github.com/trezor/trezor-mcu/issues/169#issuecomment-315850302
- https://github.com/trezor/trezor-mcu/commit/b5f9a5738f558002c3c41d67ff8230483a98047d
- https://github.com/trezor/python-trezor/commit/11b686a9f2b4cb3a1d53e50390c61b9e2c607fb2
- https://github.com/trezor/trezor-firmware/blob/2ce1e6ba7dbe5bbaeeb336fff0a038e59cb40ef8/tests/device_tests/bitcoin/test_signmessage.py

gribble PR doing what Electrum does
- https://github.com/nanotube/supybot-bitcoin-marketmonitor/pull/80

ecdsa:
I don't mind following what trezor does. 
The only drawback is that we might have to change it again if Core introduces a new method

matejcik:
at the moment Trezor-generated signatures will fail to verify in Electrum because how `from_signature65` decodes it. That's probably worth fixing, without touching the other parts.

Giszmo:
Is there any consensus yet on test vectors?

SomberNight:
There have been two related new BIPs submitted in recent months:
- [BIP-0322](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki) which is a very generic scheme for which I don't foresee widespread support anytime soon (although there is a pending [Bitcoin Core PR](https://github.com/bitcoin/bitcoin/pull/16440))
[...]

PR #7668, opened 2022-02-16
signmessage: also accept Trezor-type sigs for p2wpkh and p2wpkh-p2sh addrs

SomberNight:
see https://github.com/spesmilo/electrum/issues/3861:
> there is no clear consensus how to sign messages with segwit addresses (txin type of `p2wpkh-p2sh` and `p2wpkh`)
> We do one thing, and Trezor does something else. Bitcoin Core has this disabled.

This PR only changes the signature verification we do:
- the signatures we create are unchanged,
- but we now also accept signatures created by Trezor and others.

closes https://github.com/spesmilo/electrum/issues/3861
closes https://github.com/spesmilo/electrum/issues/7660

### DIFF
--- a/electrum/ecc.py
+++ b/electrum/ecc.py
@@ -173,19 +173,34 @@ class ECPubkey(object):
         return ECPubkey._from_libsecp256k1_pubkey_ptr(pubkey)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


massmux:
I tested this patched branch and works as a charm. Anxious to see it merged into the main branch

PR #7674, opened 2022-02-18
Separate reestablish_channel and on_channel_reestablish.

ecdsa:
all messages are sent in reestablish_channel
 on_channel_reestablish will not be cancelled if the connection is closed
 revert 57583c05cf0a4e083801e375421ded77f44e113f

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -65,7 +65,7 @@ class Peer(Logger):
     LOGGING_SHORTCUT = 'P'
 
     ORDERED_MESSAGES = (
-        'accept_channel', 'funding_signed', 'funding_created', 'accept_channel', 'channel_reestablish', 'closing_signed')
+        'accept_channel', 'funding_signed', 'funding_created', 'accept_channel', 'closing_signed')
     SPAMMY_MESSAGES = (
         'ping', 'pong', 'channel_announcement', 'node_announcement', 'channel_update',)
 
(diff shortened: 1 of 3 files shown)


electrum/lnpeer.py
```diff
@@ -1050,27 +981,108 @@ def are_datalossprotect_fields_valid() -> bool:
...
+            self.logger.warning(
+                f"channel_reestablish ({chan.get_id_for_log()}): "
+                f"remote is ahead of us! They should force-close. Remote PCP: {bh2u(their_local_pcp)}")
             # data_loss_protect_remote_pcp is used in lnsweep
             chan.set_data_loss_protect_remote_pcp(their_next_local_ctn - 1, their_local_pcp)
             self.lnworker.save_channel(chan)
             chan.peer_state = PeerState.BAD
-            return
-        elif we_are_ahead:
+            raise RemoteMisbehaving("remote ahead of us")
```
SomberNight, line 999: What if we haven't sent `channel_reestablish` yet? If we immediately disconnect, the remote won't know they are ahead.

ecdsa: right. note that I replaced the `Future` with a `Queue`
To address that case, we could revert to a `Future` and use `set_exception` instead of raising

SomberNight: ok, I guess that's a convincing argument for a Future here.
It also justifies breaking the homogeneity between the error cases; we could `set_exception` on this line and 4 lines below, and add a comment that it is to wait until `channel_reestablish` is sent.

ecdsa: fixed in 395911b08651639593c9f90a3401f0c19b6d23a7

PR #7681, opened 2022-02-25
create chmod aware of XDG_RUNTIME_DIR

scgbckbone:
`chmod` is not implemented in `XDG_RUNTIME_DIR` created by pam_systemd (tmpfs)

I use this directory for testing. I do not think people should use this as after logout directory is wiped.

I tried to find a better and more general way to decide whether file is stored in tmpfs only from its path using standard POSIX tools, yet was not able - maybe some of you can help here.

Resources:
http://liupeng0518.github.io/2018/06/28/linux/linux-tmpfs/
https://unix.stackexchange.com/questions/162900/what-is-this-folder-run-user-1000

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1132,6 +1132,7 @@ def read_json_file(path):
         raise FileImportFailed(e)
     return data
 
+
 def write_json_file(path, data):
     try:
         with open(path, 'w+', encoding='utf-8') as f:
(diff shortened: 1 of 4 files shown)


electrum/util.py
```diff
@@ -1141,13 +1142,25 @@ def write_json_file(path, data):
         raise FileExportFailed(e)
 
 
+def os_chmod(path, mode):
+    """chmod aware of tmpfs"""
+    try:
+        os.chmod(path, mode)
+    except OSError as e:
+        xdg_runtime_dir = os.environ.get("XDG_RUNTIME_DIR", None)
+        if xdg_runtime_dir and xdg_runtime_dir in path:
```
SomberNight, line 1151: The string contains/"in" keyword does not look robust.

maybe something like this instead?
`standardize_path(path).startswith(standardize_path(xdg_runtime_dir)+os.sep)`

maybe it might be worth creating a separate helper function for this, e.g. `is_subpath`

electrum/util.py
```diff
@@ -1141,13 +1142,25 @@ def write_json_file(path, data):
...
+            print("Tried to chmod in tmpfs. Skipping...", e)
```
SomberNight, line 1152:
```suggestion
            _logger.info(f"Tried to chmod in tmpfs. Skipping... {e!r}")
```

SomberNight:
Merged manually in https://github.com/spesmilo/electrum/commit/b5d2b3c51297058fd68a920414ca2dae52f8a70a
Thanks for the contribution.

PR #7682, opened 2022-02-25
add ability to import multisig wallet to Coldcard via USB

scgbckbone:
Was not sure how to add new messages to translations - so I have not added them.


electrum/plugins/coldcard/qt.py
```diff
@@ -1,19 +1,15 @@
-import time, os
```
scgbckbone, line 1: removed unused imports

electrum/plugins/coldcard/qt.py
```diff
@@ -47,23 +47,58 @@ def show_address(keystore=keystore):
...
+        if not coldcard_keystores:
             # doesn't involve a Coldcard wallet, hide feature
             return
 
-        btn = QPushButton(_("Export for Coldcard"))
-        btn.clicked.connect(lambda unused: self.export_multisig_setup(main_window, wallet))
+        btn_export = QPushButton(_("Export multisig for Coldcard as file"))
+        btn_export.clicked.connect(lambda unused: self.export_multisig_setup(main_window, wallet))
+        buttons.append(btn_export)
+        btn_import_usb = QPushButton(_("Import multisig to Coldcard via USB"))
```
scgbckbone, line 68: Button texts are kind of long - but imo more comprehensible

scgbckbone:
what do you think @SomberNight ? I'd say it is ready...

electrum/gui/qt/main_window.py
```diff
@@ -2578,10 +2578,13 @@ def label(idx, ks):
             vbox.addWidget(ks_stack)
 
         vbox.addStretch(1)
-        btn_export_info = run_hook('wallet_info_buttons', self, dialog)
```
scgbckbone, line 2581: same as in https://github.com/spesmilo/electrum/pull/7682/files - we need an ability to add multiple wallet info buttons

scgbckbone:
merged current master in and tested import via USB. This is very helpful for devs that do not want to handle sd cards all the time - would be nice if it can go in @SomberNight @doc-hex

PR #7685, opened 2022-02-28
docs-compliant get_balance method

coval3nte:
This value can be also negative because of an unconfirmed spending transaction.

References:
- https://electrumx-spesmilo.readthedocs.io/en/latest/protocol-methods.html#blockchain-scripthash-get-balance
- https://github.com/spesmilo/electrumx/blob/c8d2cc0d5cf9e549a90ca876d85fed9a90b8c4ed/electrumx/server/session.py#L1133
- https://github.com/spesmilo/electrumx/blob/c8d2cc0d5cf9e549a90ca876d85fed9a90b8c4ed/electrumx/server/mempool.py#L371
- https://github.com/spesmilo/electrum/blob/master/electrum/interface.py#L1023

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -1019,7 +1019,7 @@ class Interface(Logger):
         # check response
         assert_dict_contains_field(res, field_name='confirmed')
         assert_dict_contains_field(res, field_name='unconfirmed')
-        assert_integer(res['confirmed'])
+        assert_non_negative_integer(res['confirmed'])
         assert_integer(res['unconfirmed'])
         return res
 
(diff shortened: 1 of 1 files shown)


electrum/interface.py
```diff
@@ -1019,8 +1019,8 @@ async def get_balance_for_scripthash(self, sh: str) -> dict:
         # check response
         assert_dict_contains_field(res, field_name='confirmed')
         assert_dict_contains_field(res, field_name='unconfirmed')
-        assert_non_negative_integer(res['confirmed'])
-        assert_non_negative_integer(res['unconfirmed'])
+        assert_integer(res['confirmed'])
```
cculianu, line 1022: Technically I believe one can assert that confirmed is non-negative, since it must be in "MoneyRange" which is all non-negative integers. Author's choice here.  Of course just asserting integer is not wrong either.. depends on how constraining one wants to be here.

PR #7689, opened 2022-03-02
BUGFIX: invalid client loaded for identical wallet with different HW type

scgbckbone:
This is a BUG which causes electrum to load incorrect client for keystores with identical xpub.

Steps to reproduce:
    1. create multisig/singlesig wallet with some of your hardware devices
    2. create same multisig/singlesig wallet with at least one device different (but with same wallet loaded, in my case I used same seed and derivation path loaded on ledger and coldcard)
    3. open 1. in electrum
    4. open 2. in electrum but with device from one connected
    5. get Attribute error back from specific plugin (depends on which you used, in my case opened wallet with ledger with coldcard connected)

`  AttributeError: 'CKCCClient' object has no attribute 'checkDevice'` --> tried to use LedgerClient methods on CKCCClient

After above error is thrown, your Ledger device, that is not connected will appear as connected (green icon) and its label will be overwritten with label from connected coldcard which is definitely wrong.

I found an easy and imo elegant fix where we type check current plugin with plugin attribute from loaded client. I have only tested with ledger, trezor and coldcard, yet as all client classes have plugin attribute, this should work even for other hw_types
(stating based on this `HardwareClientBase.__init__(self, plugin=plugin)`)

### DIFF
--- a/electrum/plugin.py
+++ b/electrum/plugin.py
@@ -499,6 +499,16 @@ class DeviceMgr(ThreadJob):
                     return client
         return None
 
+    def _type_check_client_plugin(self, client: 'HardwareClientBase', plugin: "HW_PluginBase"):
+        # client must have correct plugin attribute - hasattr unnecessary
+        client_plugin_type = type(client.plugin)
+        plugin_type = type(plugin)
+        if client_plugin_type == plugin_type:
+            return True
+        else:
+            self.logger.info("client.plugin {} does not match plugin {}".format(client_plugin_type, plugin_type))
+            return False
+
     def client_by_id(self, id_, *, scan_now: bool = True) -> Optional['HardwareClientBase']:
         '''Returns a client for the device ID if one is registered.  If
         a device is wiped or in bootloader mode pairing is impossible;
@@ -532,18 +542,9 @@ class DeviceMgr(ThreadJob):
             else:
                 client = self.force_pair_xpub(plugin, handler, info, xpub, derivation)
         if client:
-            client_plugin_type = type(client.plugin)
-            plugin_type = type(plugin)
-            if client_plugin_type == plugin_type:
-                # make sure we do not use wrong client (typecheck)
-                handler.update_status(True)
-                # note: if select_device was called, we might also update label etc here:
-                keystore.opportunistically_fill_in_missing_info_from_device(client)
-            else:
-                self.logger.info("client.plugin {} does not match plugin {}".format(client_plugin_type, plugin_type))
-                # if type of client.plugin is different from chosen plugin
-                # wrong client was initialized - return None
-                client = None
+            handler.update_status(True)
+            # note: if select_device was called, we might also update label etc here:
+            keystore.opportunistically_fill_in_missing_info_from_device(client)
         self.logger.info("end client for keystore")
         return client
 
@@ -551,6 +552,8 @@ class DeviceMgr(ThreadJob):
                        devices: Sequence['Device']) -> Optional['HardwareClientBase']:
         _id = self.xpub_id(xpub)
         client = self._client_by_id(_id)
+        if client and not self._type_check_client_plugin(client, plugin):
+            return
         if client:
             # An unpaired client might have another wallet's handler
             # from a prior scan.  Replace to fix dialog parenting.
@@ -567,7 +570,7 @@ class DeviceMgr(ThreadJob):
         # choose an unpaired device and compare its first address.
         xtype = bip32.xpub_type(xpub)
         client = self._client_by_id(info.device.id_)
-        if client and client.is_pairable():
+        if client and client.is_pairable() and self._type_check_client_plugin(client, plugin):
             # See comment above for same code
             client.handler = handler
             # This will trigger a PIN/passphrase entry request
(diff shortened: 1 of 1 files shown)


SomberNight:
Hi. Thanks for reporting this issue and especially for providing a patch.

Your fix looks okay, however I sort of feel it is treating the symptoms rather than fixing the cause.
How can the code execution possibly reach line 535 with `type(client.plugin) != type(plugin)`? That is the essence of the bug, really. It should not happen. Instead of adding the extra logic there, just add `assert type(client.plugin) == type(plugin)`, and let's try to fix the root cause.

I suspect the issue is with `force_pair_xpub()`.
Could you try adding the check around here instead?
https://github.com/spesmilo/electrum/blob/eb07d5503a07cb0a2080a203caa7f0eb9627800c/electrum/plugin.py#L562

SomberNight:
> solves the case when you have same seed/wallet on different devices which are connected at the same time (opened in two different electrum wallets(windows at the same time))

Does such a setup truly work? The `DeviceMgr` is keying some dicts by xpub so some things are bound to break I think...

electrum/plugin.py
```diff
@@ -543,6 +552,8 @@ def client_by_xpub(self, plugin: 'HW_PluginBase', xpub, handler: 'HardwareHandle
                        devices: Sequence['Device']) -> Optional['HardwareClientBase']:
         _id = self.xpub_id(xpub)
         client = self._client_by_id(_id)
+        if client and not self._type_check_client_plugin(client, plugin):
```
SomberNight, line 555: probably cleaner to move this check one line below, to be at the top of the existing `if client` block

electrum/plugin.py
```diff
@@ -559,7 +570,7 @@ def force_pair_xpub(self, plugin: 'HW_PluginBase', handler: 'HardwareHandlerBase
         # choose an unpaired device and compare its first address.
         xtype = bip32.xpub_type(xpub)
         client = self._client_by_id(info.device.id_)
-        if client and client.is_pairable():
+        if client and client.is_pairable() and self._type_check_client_plugin(client, plugin):
```
SomberNight, line 573: I think we should just do a direct comparison here
`type(client.plugin) == type(plugin)`
then there is no need for a helper function

Or do you think it might be useful to log this? I am not so sure...

scgbckbone:
> > solves the case when you have same seed/wallet on different devices which are connected at the same time (opened in two different electrum wallets(windows at the same time))
> 
> Does such a setup truly work? The `DeviceMgr` is keying some dicts by xpub so some things are bound to break I think...

Depends what you mean under "works". The newly opened wallet will overwrite entry in `DeviceMgr.xpub_ids` so only the last opened wallet is fully working as xpub points to the XX last loaded client

electrum/plugin.py
scgbckbone: I still think that it has to be done in `client_by_xpub` too... then we have it on 2 places and helper functions fit there well

scgbckbone:
it is kind of unfortunate that whole `DeviceMgr` thing expects that only one HWW will have specific xpub (I know it is probably niche and not many people that are not developers will bump into this)

electrum/plugin.py
SomberNight: Ah sorry, I meant like this:
```python
        if client:
            if not self._type_check_client_plugin(client, plugin):
                return
            # An unpaired client might have another wallet's handler
            # from a prior scan.  Replace to fix dialog parenting.
            client.handler = handler
            return client
```
(but see other comment about the helper function)

electrum/plugin.py
SomberNight: Yes, I agree it is probably needed in `client_by_xpub` too.
In this case I am suggesting duplicating the code.
I am not against helper functions in general but don't see the need here.
For one, the direct comparison is fewer characters, so that in itself is a hint there is no point for the helper :P
`self._type_check_client_plugin(client, plugin)`
`type(client.plugin) == type(plugin)`

SomberNight:
Thanks, the code looks good now.
Could you confirm that you can no longer reproduce the original issue with the current diff?

scgbckbone:
yes - the issue is gone

scgbckbone:
> > it is kind of unfortunate that whole `DeviceMgr` thing expects that only one HWW will have specific xpub (I know it is probably niche and not many people that are not developers will bump into this)
> 
> Right... well that's for a later PR. :) but what do you suggest? I suppose it would be an easy change to key by `(xpub, plugin)`, which certainly looks it could handle your use case. It would not handle duplicate xpubs with matching manufacturer but not sure that is something we want/can handle. Think of it this way: if the user loses their device but buys a new one and initialises it with the same seed, should we auto-pair with that? I think we should.

as I said, this is a niche thing - I do not expect that many people having connected a.) different hw devices with same wallet loaded (even less to have b.) same hw device with same wallet)

(xpub, plugin) - good idea imo - solves at least a.) --> will look into it

if he purchases the same device - it will be loaded automatically in current implementation IMO - if he chooses to buy different device than the one he's lost - then not (I'm currently working on plugin to do this kind of conversion for coldcard) - hope I have PR soon

SomberNight (approved):

PR #7696, opened 2022-03-03
timeout improvement

coval3nte:
If custom timeout is defined as config, electrum doesn't honor it.

### DIFF
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -759,6 +759,8 @@ class Network(Logger, NetworkRetryManager[ServerAddr]):
         util.trigger_callback('network_updated')
 
     def get_network_timeout_seconds(self, request_type=NetworkTimeout.Generic) -> int:
+        if self.config.get('timeout'):
+            return self.config.get('timeout')
         if self.oneserver and not self.auto_connect:
             return request_type.MOST_RELAXED
         if self.proxy:


SomberNight:
- What is your usecase?
- `timeout` is a too generic name for the config. At least name it `network_timeout`.

coval3nte:
> * What is your usecase?
> 
>     * `timeout` is a too generic name for the config. At least name it `network_timeout`.

timeout config already exists and its used in daemon.request... can't electrum use same parameter for querying electrumx?

SomberNight:
> My usecase is a private ElectrumX server with !autoconnect and oneserver, so I need electrum to be relaxed as much as it can

In that case, there is already a 600 second timeout in your case. How long do you want to set it to? :)

>  By the way, if the electrum client timeout is lower than electrum daemon timeout it will end up by closing connection in (electrum client -> daemon) and not in (electrum daemon -> electrumx) so it will throw same error in both cases.

Ok, but direct CLI usage is not the only scenario to consider.
When you have a daemon running, I am wondering if it might make sense to have a very long timeout on operations it is doing, but a short one for your own interactions with the daemon.
Also, many of the daemon requests do not involve communicating with electrum servers. Besides taking a long time due to a server not responding, there is also lightning stuff or doing local queries on a large wallet.

> timeout config already exists and its used only in daemon.request for querying electrum daemon... 

Right... I forgot about that.

Anyway, we can make the network request timeout configurable, sure, I am just not sure if we should reuse this config key...
@ecdsa what do you think?

electrum/network.py
```diff
@@ -759,6 +759,8 @@ async def connection_down(self, interface: Interface):
         util.trigger_callback('network_updated')
 
     def get_network_timeout_seconds(self, request_type=NetworkTimeout.Generic) -> int:
+        if self.config.get('timeout'):
+            return self.config.get('timeout')
```
SomberNight, line 763: make sure the type hint is respected: cast the return value to int

coval3nte:
@SomberNight I agree with you.. I think it's better to have two separate timeout variables, so it can fit to every usecase.

ecdsa:
This user contacted me on IRC because `listunspent` was very slow with a very large wallet (20k addresses).
I asked whether the wallet was synchronized before running the command, and he said that it was.
Not sure if his current issue is related, but if it is, it might not be caused by network timeout parameters.

coval3nte:
> This user contacted me on IRC because `listunspent` was very slow with a very large wallet (20k addresses). I asked whether the wallet was synchronized before running the command, and he said that it was. Not sure if his current issue is related, but if it is, it might not be caused by network timeout parameters.

No, It isn't related. Today I was checking logs and I found out this. That's all.

Issue #7602, opened 2021-12-21
Unable to run the build scripts on Jenkins


pedro-at-decenomy:
I'm not able to run the build scripts on a Jenkins build server, due to the fact that there are several `sudo` commands on the scripts, and also because there are also several `docker run -it` on them.

I fixed it on my server, if you have to have these fixes on the repository I can make a PR with them, they are rather easy to do:

1. remove `sudo` from the build commands;
2. change the `docker run -it` commands to `docker run -i`;

The user `jenkins` or any other that runs the script, needs to be on the docker group: `sudo usermod -aG docker ${USER}` + logout/login or restart

Best regards,
  Pedro Simões, DECENOMY IT

SomberNight:
Thanks for the suggestions.
We've adopted your recommendation (1) of dropping sudo in https://github.com/spesmilo/electrum/pull/7697.

-----

Re (2), I want to keep the `-t` (pseudo tty) option because of better UX for the more common use case of a developer running the commands themselves in a terminal.
E.g. if I manually run `$ ./contrib/build-linux/appimage/build.sh`, as-is (with `-t`) pressing `Ctrl+C` works as expected and stops the build. Without `-t`, `Ctrl+C` is ignored.

Also, the logs printed to the terminal are more terminal-friendly with `-t`. For example:
With `-t` (as-is), the text is dynamically updated inline:
```
--2022-11-06 05:50:07--  https://www.python.org/ftp/python/3.9.15/Python-3.9.15.tar.xz
Resolving www.python.org (www.python.org)... 151.101.36.223, 2a04:4e42:9::223
Connecting to www.python.org (www.python.org)|151.101.36.223|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 19712208 (19M) [application/octet-stream]
Saving to: ‘/opt/electrum/contrib/build-linux/appimage/../../../contrib/build-linux/appimage/.cache/appimage/Python-3.9.15.tar.xz’

/opt/electrum/contrib/bui 100%[=====================================>]  18.80M  11.6MB/s    in 1.6s    

2022-11-06 05:50:09 (11.6 MB/s) - ‘/opt/electrum/contrib/build-linux/appimage/../../../contrib/build-linux/appimage/.cache/appimage/Python-3.9.15.tar.xz’ saved [19712208/19712208]
```
Compare that with (no `-t` this time):


<details>
 <summary>log</summary>

```
[...]

PR #7697, opened 2022-03-03
build: rm need for sudo in most places; and do not run as root

SomberNight:
This includes two logically separate changes:
- on the host, try not to require sudo when running the build scripts
    - namely when interacting with the docker daemon, this requires the unix user on the host to be part of the `docker` group
    - this solves part of https://github.com/spesmilo/electrum/issues/7602
- while running inside the docker containers, do not run as root
    - this means that e.g. files created in mounted folders should no longer be owned by root on the host
    - there is some code duplication involved here - not sure how it could be deduped.

### DIFF
--- a/contrib/android/Dockerfile
+++ b/contrib/android/Dockerfile
@@ -94,11 +94,6 @@ RUN curl --location --progress-bar \
     && rm -rf "${APACHE_ANT_ARCHIVE}"
 
 
-ENV USER="user"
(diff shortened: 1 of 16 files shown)


SomberNight:
Note: users with existing git clones and leftover artifacts from previous builds might have to do some manual clean-up (such as `sudo rm`-ing certain dirs or `chown`-ing them).

PR #7698, opened 2022-03-04
Update udev rules for Nano S Plus

gruve-p:


SomberNight (approved):
matches https://github.com/LedgerHQ/udev-rules/blob/2776324af6df36c2af4d2e8e92a1c98c281117c9/20-hw1.rules

PR #7700, opened 2022-03-07
Convert2cc

scgbckbone:
* add new plugin convert2cc

Testing:
1. create multisig/singlesig wallet with non coldcard hardware device/s (in my case trezor, ledger)
2. click on 'Wallet' tab and choose 'Information' - wallet information dialog will open and you should see 'convert2cc' button Screenshot from 2022-03-07 09-04-40
3. if 2. does not hold - go to plugins, activate 'convert2cc' and go back to 2. (needs ckcc-protocol as dependency)
4. if wallet created in 1. is singlesig disconnect hww and load same seed + passphrase (if used) to coldcard and connect it to PC, if multisig, do the same for just one cosigner
   (can be done even without Coldcard connected, it won't give you a hint based on master fingerprint match)
5. click 'convert2cc' button in wallet information modal - convert2cc modal will pop up 
Screenshot from 2022-03-07 09-05-07
6. one of the keystores/cosigner should match with your connected coldcard (icon_paired + [matches connected Coldcard])
   if your coldcard is not connected there will be no match Screenshot from 2022-03-07 09-40-07
8. choose correct keystore and hit 'convert button'
9. select how to name a where to store newly created converted wallet
10. hit yes - confirming that you want to open newly created wallet
11. old wallet is closed, converted wallet is opened and one of your devices is converted to Coldcard (should show paired icon)
12. try that wallet is working as expected


scgbckbone:
1. This only works for standard and multisig wallets - if tried with different one (i.e. imported) convert2cc modal:
Screenshot from 2022-03-07 09-41-01
2. at least one of keystore cosigners must not be coldcard (so there is something to convert):
Screenshot from 2022-03-07 09-41-38

electrum/gui/qt/main_window.py
```diff
@@ -593,15 +594,16 @@ def on_cb(x):
         if cb_checked:
             self.config.set_key('dont_show_testnet_warning', True)
 
-    def open_wallet(self):
+    def open_wallet(self, filename=None):
```
scgbckbone, line 597: add ability to skip `getOpenFileName` dialog if filename is provided

electrum/gui/qt/main_window.py
```diff
@@ -2571,17 +2574,21 @@ def label(idx, ks):
...
 
             select_ks(0)
             vbox.addWidget(ks_stack)
 
         vbox.addStretch(1)
-        btn_export_info = run_hook('wallet_info_buttons', self, dialog)
+        wallet_info_btns = run_hook('wallet_info_buttons', self, dialog)
         btn_close = CloseButton(dialog)
-        btns = Buttons(btn_export_info, btn_close)
-        vbox.addLayout(btns)
```
scgbckbone, line 2584: similar to https://github.com/spesmilo/electrum/pull/7682/files where we need an ability to add multiple buttons to wallet information dialog

electrum/gui/qt/main_window.py
```diff
@@ -2571,17 +2574,21 @@ def label(idx, ks):
...
+        if wallet_info_btns:
+            # now more than one wallet_info_butons hook, needs concat
+            wallet_info_btns = reduce(operator.concat, wallet_info_btns)
```
scgbckbone, line 2587: now we also expect more than one hook to provide data

electrum/plugin.py
```diff
@@ -232,6 +232,8 @@ def run_hook(name, *args):
                 results.append(r)
 
     if results:
+        if name == "wallet_info_buttons":
```
scgbckbone, line 235: not sure If I understand this constraint - why can only one result from hook be accepted? Here I expect more than one hook to provide data

electrum/plugins/convert2cc/qt.py
```diff
@@ -0,0 +1,236 @@
...
+        dialog.setLayout(vbox)
+        dialog.exec_()
+
+    def do_convert2cc(self, main_window, labels_clayout, keystores, dialog):
+        wallet = main_window.wallet
+        selected_index = labels_clayout.selected_index()
+        if selected_index is not None:
+            target_keystore = keystores[selected_index]
+            with self.coldcards_connected() as connected_cc_clients:
+                # dev = self.match_candidate_keystore_to_connected_cc_device(connected_cc_clients, target_keystore)
```
scgbckbone, line 208: for now - not sending device to convert2cc - it is not needed and in current `ckcc-protocol` version it causes huge slowdown, see https://github.com/Coldcard/ckcc-protocol/pull/17

scgbckbone:
added msg of coldcard device which we cannot connect to but is plugged in (other application using it, or even electrum itself for no reason check https://github.com/spesmilo/electrum/pull/7701)

Screenshot from 2022-03-07 15-02-29

scgbckbone:
added warning for encrypted wallets
Screenshot from 2022-03-09 18-02-36

scgbckbone:
instead of warning for encrypted wallets - I found and easy way to preserve encryption settings. User can choose to preserve them or not (checkbox)

hardware device encryption:
Screenshot from 2022-03-10 09-21-25



password encryption:
Screenshot from 2022-03-10 09-25-38

If one chooses to preserve encryption settings, same pwd (pubkey) will be used for encryption. If one chooses NOT to preserve, newly generated wallet is plaintext and can be encrypted later via standard path (wallet -> password ...)

SomberNight:
Do I understand it correctly that the whole purpose of this plugin is to help users who have non-coldcard hw devices and want to migrate to coldcard? -- and this plugin creates a new wallet file which is sort of a copy of the existing one, except one of the keystores is replaced with Coldcard_KeyStore?

accumulator:
A plugin like this is more suitable as an external plugin. See also #9235.

PR #7701, opened 2022-03-07
close unused client connections

scgbckbone:
This is more of a discussion PR, why do we keep clients for unused hww open? 

Scenario:
1. create sigle sig wallet with one of your hww devices and encrypt it
2. connect other hww to PC
3. cloose and re-open wallet from 1.
4. choices dialog from base wizard is opened with hww choices
5. choose one of your hww to decrypt wallet
6. after this wallet 1. is opened and connected to target device, yet all other hww are connected too, even if unused. This will cause that no other application can use them as they are already opened in electrum (but for no reason)

### DIFF
--- a/electrum/base_wizard.py
+++ b/electrum/base_wizard.py
@@ -389,6 +389,11 @@ class BaseWizard(Logger):
             self.show_error(str(e))
             raise ChooseHwDeviceAgain()
 
+        # close clients that are not used
+        for cl, (path, id_) in list(devmgr.clients.items()):  # list func will copy it, therefore we can work with devmgr.clients dict
+            if cl != client:
+                devmgr._close_client(id_)
+
         if purpose == HWD_SETUP_NEW_WALLET:
             def f(derivation, script_type):
                 derivation = normalize_bip32_derivation(derivation)


SomberNight:
Right, I guess when the wizard scans for hw devices, it creates clients for all connected devices, in `BaseWizard._choose_hw_device`:
https://github.com/spesmilo/electrum/blob/a65b97d25c3e37c0e77484f1846e537f9d5be274/electrum/base_wizard.py#L317
https://github.com/spesmilo/electrum/blob/a65b97d25c3e37c0e77484f1846e537f9d5be274/electrum/plugin.py#L601

> why do we keep clients for unused hww open?

No reason I guess; it's not clear where we should close them.
For example, what the current PR is doing looks weird... what if there is another main window open with some device connected? We should not close that client...

Maybe `unpaired_device_infos` could close the clients when it finishes? Not sure... I did not think through all the implications though.

scgbckbone:
if there is already another window using some devices I think you won't even see them in wizard as they are already used, right?

SomberNight:
Yes, I believe that is correct. However here you are iterating `list(devmgr.clients.items())`, which I expect would include them.

scgbckbone:
you're right - I'm accessing it incorrectly - let me dig deeper here

SomberNight:
The wizard was completely refactored in https://github.com/spesmilo/electrum/pull/8560, so this PR is now too stale.

PR #7724, opened 2022-03-19
bump fee of swap claim transactions

ecdsa:
Note: this is somewhat inefficient, because we loop over swaps.
This PR adds a new field to PartialTxOutput

set_label has been commented out because it would create dangling labels.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -300,7 +300,7 @@ class SwapManager(Logger):
             dummy_output = PartialTxOutput.from_address_and_value(ln_dummy_address(), expected_onchain_amount_sat)
             tx.outputs().remove(dummy_output)
             tx.add_outputs([funding_output])
-            tx.set_rbf(True) # rbf must not decrease payment
+            tx.set_rbf(True)  # note: rbf must not decrease payment
             self.wallet.sign_transaction(tx, password)
         # save swap data in wallet in case we need a refund
         swap = SwapData(
@@ -542,6 +542,12 @@ class SwapManager(Logger):
             if txin.prevout.txid.hex() == swap.funding_txid:
                 return swap
         return None
+
+    def is_lockup_address_for_a_swap(self, addr: str) -> bool:
+        for key, swap in self.swaps.items():  # TODO take lock? or add index to avoid looping
+            if addr == swap.lockup_address:
+                return True
+        return False
 
     def sign_tx(self, tx, swap):
         preimage = swap.preimage if swap.is_reverse else 0
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -92,6 +92,7 @@ class SwapData(StoredObject):
     funding_txid = attr.ib(type=Optional[str])
     spending_txid = attr.ib(type=Optional[str])
     is_redeemed = attr.ib(type=bool)
+    _funding_prevout = None # for RBF
 
 
 def create_claim_tx(
@@ -171,6 +172,7 @@ class SwapManager(Logger):
                 self.logger.info('amount too low, we should not reveal the preimage')
                 continue
             swap.funding_txid = txin.prevout.txid.hex()
+            swap._funding_prevout = txin.prevout
             spent_height = txin.spent_height
             if spent_height is not None:
                 swap.spending_txid = txin.spent_txid
@@ -542,7 +544,7 @@ class SwapManager(Logger):
 
     def get_swap_by_claim_txin(self, txin: TxInput) -> Optional[SwapData]:
         for key, swap in self.swaps.items():
-            if txin.prevout.txid.hex() == swap.funding_txid:
+            if txin.prevout == swap._funding_prevout:
                 return swap
         return None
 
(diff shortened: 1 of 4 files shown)


electrum/submarine_swaps.py
```diff
@@ -541,3 +532,32 @@ def get_send_amount(self, recv_amount: Optional[int], *, is_reverse: bool) -> Op
         if is_reverse and send_amount is not None:
             send_amount += self.get_claim_fee()
         return send_amount
+
+    def get_swap_by_tx(self, tx):
+        txin = tx.inputs()[0]
+        # FIXME: not clear when TxInput has address
+        if hasattr(txin, 'address'):
```
SomberNight, line 539: Pure `TxInput` never has `address`; only a `PartialTxInput` might. i.e. a "complete" tx will not have it.
In general, to know the address, we need some information about the output of the parent tx, which we often do not have.

SomberNight:
> set_label has been commented out because it would create dangling labels.

I guess that's a pre-existing issue as it can probably already happen when the user manually RBFs any tx that has a txid-specific label.
We could try to clean-up dangling labels during wallet-open, e.g. in `load_and_cleanup`, however it looks problematic as the label keys are heterogeneous.
https://github.com/spesmilo/electrum/blob/e00c998df346bd81a2de2720c3b01307d4a31eba/electrum/wallet.py#L399
https://github.com/spesmilo/electrum/blob/e00c998df346bd81a2de2720c3b01307d4a31eba/electrum/wallet.py#L1154-L1156

electrum/submarine_swaps.py
SomberNight: It might be interesting to look at `wallet.get_txin_address`; but it's not useful here.
https://github.com/spesmilo/electrum/blob/e00c998df346bd81a2de2720c3b01307d4a31eba/electrum/address_synchronizer.py#L139-L152

ecdsa:
note: in the case of a forward swap, the funding transaction should also be RBF

electrum/submarine_swaps.py
ecdsa: this is not needed anymore, I found how to do it by comparing `funding_txid`

electrum/submarine_swaps.py
```diff
@@ -541,3 +534,50 @@ def get_send_amount(self, recv_amount: Optional[int], *, is_reverse: bool) -> Op
...
+    def get_swap_by_tx(self, tx: Transaction) -> Optional[SwapData]:
+        # determine if tx is spending from a swap
+        txin = tx.inputs()[0]
+        return self.get_swap_by_claim_txin(txin)
+
+    def get_swap_by_claim_txin(self, txin: TxInput) -> Optional[SwapData]:
+        for key, swap in self.swaps.items():
+            if txin.prevout.txid.hex() == swap.funding_txid:
+                return swap
+        return None
```
SomberNight, line 547: It is insufficient to compare just the txid to identify claim txs. The full outpoint needs to be compared (txid+out_idx).

Consider doing two forward swaps in a row.
tx1 funds swap1 and also creates a change output.
Now we start doing swap2, and the previous change output is chosen as an input for funding tx2.
tx2 which is funding swap2 is going to be incorrectly identified as a claim tx for swap1, as
`tx2.intputs()[0].prevout.txid.hex() == swap1.funding_txid`
but note that
`tx2.intputs()[0].prevout.out_idx != swap1.funding_outidx` (or something... the swap object is missing the required field)

ecdsa: Indeed, this is relevant for refund transactions. 

Note that since 1364e7538acf37d1b649d2ca8ef647858540b227, the `funding_txid` field of `SwapData` is populated by `lnwatcher`, and does not need to be persisted anymore. Without a wallet_db upgrade, we could populate that field with the serialized outpoint, or add a new, non-persisted field to `SwapData`, that would have the proper name.

(Not sure if that would still work with  `subscribe_to_outpoint`)

ecdsa: There is a pending db upgrade for `SwapData` in the `seamless_lightning` branch, that persists `receive_address`.
Maybe we should move that part of the db upgrade out of that branch, and bundle it whatever we need here.

ecdsa: This would also work, I guess (without db upgrade):
```python
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -92,6 +92,7 @@ class SwapData(StoredObject):
     funding_txid = attr.ib(type=Optional[str])
     spending_txid = attr.ib(type=Optional[str])
     is_redeemed = attr.ib(type=bool)
+    _funding_prevout = None
 
 
 def create_claim_tx(
@@ -171,6 +172,7 @@ class SwapManager(Logger):
                 self.logger.info('amount too low, we should not reveal the preimage')
                 continue
             swap.funding_txid = txin.prevout.txid.hex()
+            swap._funding_prevout = txin.prevout
             spent_height = txin.spent_height
             if spent_height is not None:
                 swap.spending_txid = txin.spent_txid
@@ -542,7 +544,7 @@ class SwapManager(Logger):
 
     def get_swap_by_claim_txin(self, txin: TxInput) -> Optional[SwapData]:
         for key, swap in self.swaps.items():
-            if txin.prevout.txid.hex() == swap.funding_txid:
+            if txin.prevout == swap._funding_prevout:
                 return swap
         return None
```

Issue #7725, opened 2022-03-20
Opening "Preferences" results in segfault on macOS (crash, QtMultimedia camera handling)


bitkarrot:
See attached file for crash report. Happens when clicking on 'preferences' or while closing open wallets.
[electrum_crash.txt](https://github.com/spesmilo/electrum/files/8310726/electrum_crash.txt)

SomberNight:
> Happens when clicking on 'preferences'

Maybe that's related to `QtMultimedia` and the camera-handling...

Could you test going into the `Console` tab (menubar>`View`>`Show Console`), and then running:
```
from PyQt5.QtMultimedia import QCameraInfo
```

Also, to try to narrow things down, could you test if Electrum 4.1.5 also reproduces the crash?

> crash happens [...] while closing open wallets

if you mean during application shutdown (e.g. closing last open wallet), that might be https://github.com/spesmilo/electrum/issues/6889

bitkarrot:
> Could you test going into the `Console` tab (menubar>`View`>`Show Console`), and then running:
> 
> ```
> from PyQt5.QtMultimedia import QCameraInfo
> ```
<img width="358" alt="Screen Shot 2022-03-20 at 4 57 44 PM" src="https://user-images.githubusercontent.com/73979971/159191562-5dedbc84-3fa0-47bb-b6f0-ecc6130e88e2.png">

this import actually returns nothing. 

tried to repeat the 'preferences' crash from last night. i noticed that just open/closing wallets today its okay. when new wallets are closed and opened, that's when the crash happens, sometimes on preferences, sometimes when closing wallets. 

i'm on bigsur 11.6 if that helps. 

when i get a chance to uninstall i'll try out the earlier version

> 
> if you mean during application shutdown (e.g. closing last open wallet), that might be #6889

yes, its definitely #6889

bitkarrot:
i installed the .dmg, not sure if that had an impact. 

perhaps i should run from source? (need to set up dev environment anyway ;)

danshao:
Also crashes on non M1.

I'm on Mac OS Monterey 12.3 Macbook Air 2019. Crash occurs whenever I hit the Preferences menu button.

SomberNight:
Could you either of specifically test the following steps:
- start Electrum, open a wallet
- go into the `Console` tab (menubar>`View`>`Show Console`), and then run:
  ```
  from PyQt5.QtMultimedia import QCameraInfo
  ```
- open Preferences, see if it crashes

[...]

PR #7727, opened 2022-03-21
qt gui: (maybe fix) opening "Preferences" segfaults on macOS for some

SomberNight:
see https://github.com/spesmilo/electrum/issues/7725
related https://github.com/Electron-Cash/Electron-Cash/pull/1545


bitkarrot:
see https://github.com/spesmilo/electrum/issues/7725#issuecomment-1074817914 for test result

PR #7730, opened 2022-03-23
Seamless lightning

ecdsa:
<s>This is still a draft, and the code gets rebased frequently. 
I am opening a PR to facilitate discussions.</s>

No longer a draft, but the `wallet_db` upgrade needs to be carefully checked.


accumulator:
[copied from code line remark]

IMO there should still be an opt-out for using fallback addresses (e.g set globally in config, or per wallet), because there might be privacy implications, as when an invoice+fallback is paid via lightning, the fallback address eventually gets reused in future requests.

ecdsa:
Yes, we can make fallback addresses opt-out.
Note that the issue of recycling addressed already exists with on-chain invoices. 
There is little we can do about it, because Electrum relies on a gap limit to recover from seed.

ecdsa:
We could also add an option to include the lightning invoice in the bip21 URI, see #7253

accumulator:
We could introduce something like a preferences option for 'lightning: don't use onchain fallback addresses', in which case BIP21 doesn't include a lightning fallback, and a lightning invoice doesn't contain an onchain fallback, which forces the user to choose between presenting an onchain invoice or a lightning invoice?

accumulator:
This is an interesting resource : https://github.com/sbddesign/bip21-site/blob/main/wallet_support.json
Seems to document lightning fallback support in bip-21 for many wallets, including if the wallet generates bip-21 w. lightning fallback

ecdsa:
Note about unconfirmed balance: the current code on master (and also in this branch) sums positive and negative unconfirmed values. This does not make sense.

If we send a payment, we receive change that is, of course unconfirmed. However, there is no reason to consider that this affects our confirmed balance; if the outgoing transaction does not confirm, then we will still own the coins (and even more). Therefore, the ownership of those coins should be considered as confirmed.

Thus, we should only consider incoming transactions in the unconfirmed balance. This also fixes the current issue of negative and positive values canceling out each other.

PR #7747, opened 2022-03-29
Notes for Apple M1 build from source

bitkarrot:
Installation notes for building from source for Apple M1 (ARM) 
@SomberNight

### DIFF
--- a/contrib/osx/notes_for_arm.md
+++ b/contrib/osx/notes_for_arm.md
@@ -1,19 +1,19 @@
-Notes on how to build Electrum from source on Apple M1 OSX 
+# Notes on how to build Electrum on Apple M1 OSX 
 
 
 1. Clone repository and build Electrum as usual:
 
 ```
-git clone https://github.com/spesmilo/electrum.git
-cd electrum
-./contrib/osx/make_osx
+$ git clone https://github.com/spesmilo/electrum.git
+$ cd electrum
+$ ./contrib/osx/make_osx
 ```
 
 2. pycryptodomex and cryptography and will need to be manually installed:
 
 ```
-pip install pycryptodomex
-pip install cryptography
+$ pip install pycryptodomex
+$ pip install cryptography
 ```
 
 4. Manually install libsecp256k1 from source (https://github.com/bitcoin-core/secp256k1#build-steps) but with the following modifications: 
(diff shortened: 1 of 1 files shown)


contrib/osx/notes_for_arm.md
```diff
@@ -0,0 +1,56 @@
+# Notes on building Electrum on Apple M1 OSX (ARM)
```
SomberNight, line 1:
```suggestion
# Notes on running Electrum from source on ARM-based Macs (Apple M1 OSX)
```

Looks like the title is not technically correct - the text is re how to run from source on apple m1.
Building suggests it is about creating binaries (such as the .dmg files we distribute).

contrib/osx/notes_for_arm.md
```diff
@@ -0,0 +1,56 @@
...
+```
+python3 -m pip install --user -e .
+```
+
+2. pycryptodomex and cryptography and will need to be manually installed:
+
+```
+$ pip install pycryptodomex
+$ pip install cryptography
+```
```
SomberNight, line 23: You don't need both, do you?
Either one should be enough; they are alternatives to each other.
Assuming that's so, we prefer `cryptography` (or at least let's be consistent with the top-level README.rst).

contrib/osx/notes_for_arm.md
```diff
@@ -0,0 +1,56 @@
...
+
+3. Manually install libsecp256k1 from source (https://github.com/bitcoin-core/secp256k1#build-steps) but with the following modifications: 
+
+```
+$ ./autogen.sh
+$ ./configure --enable-module-recovery
+$ make
+$ make check  # run the test suite
+$ sudo make install  # optional
+```
```
SomberNight, line 33: Does [`contrib/make_libsecp256k1.sh`](https://github.com/spesmilo/electrum/blob/cb39bbbd943e19b392302a1ac557c374fcd2d36b/contrib/make_libsecp256k1.sh) not work?

contrib/osx/notes_for_arm.md
```diff
@@ -0,0 +1,56 @@
...
+$ echo 'export PATH="/opt/homebrew/opt/qt@5/bin:$PATH"' >> ~/.zshrc
+$ echo 'export PATH="/opt/homebrew/opt/pyqt@5/5.15.4_1/bin:$PATH"' >> ~/.zshrc
+$ source ~/.zshrc
+```
+
+Finally, try it in python to ensure it works: 
+
+```
+$ python3
+>> import PyQt5
```
SomberNight, line 48:
```suggestion
>>> import PyQt5
```

contrib/osx/notes_for_arm.md
bitkarrot: Ok lets just make it `cryptography`

contrib/osx/notes_for_arm.md
bitkarrot: oops didn't see that. lets edit it to be explicit. Sure makes the next person's life a lot easier.

bitkarrot:
let me know if anything else

Thanks @SomberNight

SomberNight:
I think most of this is covered by the [README](https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/README.rst), except for the brew PyQt5 workaround, but for that I will merge it.
Thanks.

PR #7754, opened 2022-04-04
wallet: "up_to_date" to wait for SPV/Verifier

SomberNight:
This PR changes `wallet.is_up_to_date()` to consider both the Synchronizer and the Verifier (SPV).
Without this, currently, `wallet.is_up_to_date()` only considers the Synchronizer.

IIRC historically this choice was made to show the wallet balance as early as possible when restoring from seed. At the time, the Synchronizer did not depend on the Verifier at all, so this was reasonable. (obviously the Verifier does depend on the Synchronizer: that's how it gets txs to verify)
However, this is no longer the case, since https://github.com/spesmilo/electrum/commit/9c31c1f970d9bc371ed2dcba26354c71c6ec8756, the wallet only generates new addresses and rolls the gap limit forward after existing addresses have spv-verified confirmed history. That is, the Synchronizer too depends on the Verifier.

So, this PR makes it a bit slower to get to the state where we display the balance when restoring, however, not by much...
The above commit already made the Synchronizer wait for the Verifier before new addresses are generated, so the only (new) slowdown is re the "last batch of addresses", after which the gap limit won't be rolled forward -- previously the Synchronizer would not wait in that case only. Hmm... how does the Synchronizer even know that there wouldn't be new addresses generated? Well... it is racing the Verifier, and there is a timing issue.

Without this, while syncing a wallet, `wallet.is_up_to_date()` can flip-flop between true and false multiple times, depending on timing. An easy way to reproduce is to sync a wallet of non-trivial size but slow down the polling interval of the Verifier to e.g. 5 seconds:
https://github.com/spesmilo/electrum/blob/250795f137281b8d2c8d29081949ee781a1c0b2a/electrum/verifier.py#L73
If `wallet.is_up_to_date()` flips, the sync state progress indicator is reset, probably confusing the user:
https://github.com/spesmilo/electrum/blob/250795f137281b8d2c8d29081949ee781a1c0b2a/electrum/synchronizer.py#L268-L269

Originally I just wanted to provide some new API to wait until a wallet is fully synced including SPV, but now I think we should just change the behaviour of `wallet.is_up_to_date()` (as in this PR).

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -90,7 +90,7 @@ class AddressSynchronizer(Logger):
         # Txs the server claims are in the mempool:
         self.unconfirmed_tx = defaultdict(int)  # type: Dict[str, int]  # txid -> height. Access with self.lock.
         # true when synchronized
-        self._up_to_date = False
+        self._up_to_date = False  # considers both Synchronizer and Verifier
         # thread local storage for caching stuff
         self.threadlocal_cache = threading.local()
 
@@ -922,5 +922,6 @@ class AddressSynchronizer(Logger):
         c, u, x = self.get_addr_balance(address)
         return c+u+x == 0
 
-    def synchronize(self):
-        pass
+    def synchronize(self) -> int:
+        """Returns the number of new addresses we generated."""
+        return 0
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -672,8 +672,14 @@ class AddressSynchronizer(Logger):
         with self.lock:
             status_changed = self._up_to_date != up_to_date
             self._up_to_date = up_to_date
-        if self.network:
-            self.network.notify('status')
+        # reset sync state progress indicator
+        if up_to_date:
+            if self.synchronizer:
+                self.synchronizer.reset_request_counters()
+            if self.verifier:
+                self.verifier.reset_request_counters()
+        # fire triggers
+        util.trigger_callback('status')
         if status_changed:
             self.logger.info(f'set_up_to_date: {up_to_date}')
 
@@ -681,10 +687,16 @@ class AddressSynchronizer(Logger):
         return self._up_to_date
 
     def get_history_sync_state_details(self) -> Tuple[int, int]:
+        nsent, nans = 0, 0
         if self.synchronizer:
-            return self.synchronizer.num_requests_sent_and_answered()
-        else:
-            return 0, 0
+            n1, n2 = self.synchronizer.num_requests_sent_and_answered()
+            nsent += n1
+            nans += n2
+        if self.verifier:
+            n1, n2 = self.verifier.num_requests_sent_and_answered()
+            nsent += n1
+            nans += n2
+        return nsent, nans
 
     @with_transaction_lock
     def get_tx_delta(self, tx_hash: str, address: str) -> int:
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -85,8 +85,10 @@ class AddressSynchronizer(Logger):
         self.lock = threading.RLock()
         self.transaction_lock = threading.RLock()
         self.future_tx = {}  # type: Dict[str, int]  # txid -> wanted height
-        # Transactions pending verification.  txid -> tx_height. Access with self.lock.
-        self.unverified_tx = defaultdict(int)
+        # Txs the server claims are mined but still pending verification:
+        self.unverified_tx = defaultdict(int)  # type: Dict[str, int]  # txid -> height. Access with self.lock.
+        # Txs the server claims are in the mempool:
+        self.unconfirmed_tx = defaultdict(int)  # type: Dict[str, int]  # txid -> height. Access with self.lock.
         # true when synchronized
         self._up_to_date = False
         # thread local storage for caching stuff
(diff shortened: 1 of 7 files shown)


ecdsa:
note: I agree with the concept that `up_to_date` should wait for verifier.

SomberNight:
> the `up_to_date_changed_event` Event is not awaited anywhere.

Right, this PR does not use that, as we currently don't have code that is polling up_to_date (although we used to).
It could be useful in e.g. https://github.com/spesmilo/electrum/pull/7584/files#diff-c1fda5db5b757c568c29740d4093bf99c4bb2120f1a361f97edfda4863bacdb9R282-R283

```python
while not wallet.is_up_to_date():
    await asyncio.sleep(0.1)
```
The above can be replaced with:
```python
while True:
    await wallet.up_to_date_changed_event.wait()
    if wallet.is_up_to_date():
        break
```

> also, how about renaming it to something shorter, eg `status_changed`

Well how about `up_to_date_changed` then? :P That's more explicit.

SomberNight:
>> also, how about renaming it to something shorter, eg `status_changed`
>
> Well how about `up_to_date_changed` then? :P That's more explicit.

I've renamed it to `up_to_date_changed` in a force-push.

So no, this is not a draft, it is mergeable IMO; but yes it adds that API (intentionally) which is currently unused.

ecdsa:
```python
while True:
    await wallet.up_to_date_changed_event.wait()
    if wallet.is_up_to_date():
        break
```
This seems to be horribly convoluted.

If the use case is to convey the info that the wallet actually is up to date, then why not just create an event named `up_to_date`, that is set when we are up to date, and unset when we are not? it would replace both the boolean `self.up_to_date` and the event `up_to_date_changed`. The method `self.is_up_to_date` would check if the event is set.

SomberNight:
>> ```python
>> while True:
>>     await wallet.up_to_date_changed_event.wait()
>>     if wallet.is_up_to_date():
>>         break
>> ```
> 
> This seems to be horribly convoluted.

To me it looks pretty standard :D  but ok

> If the use case is to convey the info that the wallet actually is up to date, then why not just create an event named `up_to_date`, that is set when we are up to date, and unset when we are not? it would replace both the boolean `self.up_to_date` and the event `up_to_date_changed`. The method `self.is_up_to_date` would check if the event is set.

Well the mentioned use case is to wait until the wallet is synced, so for that of course that would work.
However, there could be other use cases that might want to get notified when the wallet was up to date but starts syncing again -- such an example is the GUI status bar, however that is using the `status` callback atm (and Qt does not work nicely with asyncio so I would not want to touch that any time soon, it's just an example).
Regardless, as there are no concrete use cases now, I am fine with implementing what you suggest, but then I would want to change `up_to_date` to be private/internal. - that would allow easily changing it later as needed

Currently `up_to_date` is being used from the GUI, both Qt and kivy, in multiple places. It is being accessed directly, not via `is_up_to_date()`. Presumably this is to avoid having to take the wallet lock. I don't understand the logic of this...
I mean, there is the `up_to_date` field, the `is_up_to_date()` method, and the `set_up_to_date()` method. The methods take wallet.lock, but the direct access of the field obviously does not. Luckily direct accesses of `up_to_date` are always reads.
My question is, do you see a reason why `is_up_to_date()` needs to take wallet.lock? If it did not take it, it would be pretty straight-forward to make the underlying field private.

SomberNight:
>>> ```python
>>> while True:
>>>     await wallet.up_to_date_changed_event.wait()
>>>     if wallet.is_up_to_date():
>>>         break
>>> ```
>> 
>> This seems to be horribly convoluted.
>
> To me it looks pretty standard :D  but ok

Oh wait, one can just do this instead:

```python
while not wallet.is_up_to_date():
    await wallet.up_to_date_changed_event.wait()
```

I should have thought that through :) in fact there needs to be a test before the first await to be sane.

ecdsa:
> Presumably this is to avoid having to take the wallet lock.

I am not sure. maybe that code/practice is simply older than the lock...

ecdsa:
I do not like adding variables when they are not strictly needed. If we can replace a boolean with an event, then we should do that, instead of adding an event to the existing logic, and then adding extra logic to ensure that the event follows what the boolean is doing.

I we wanted to be notified when the wallet is no longer in sync, then I would add another event, and call `x.set()` and `y.clear()` together. That is easier to read. But we do not need that at the moment, so please do not add it until you actually use it.

SomberNight:
Ok, I removed the controversial commit. It is not needed atm anyway.

SomberNight:
> I would want to change `up_to_date` to be private/internal. - that would allow easily changing it later as needed
>
> Currently `up_to_date` is being used from the GUI, both Qt and kivy, in multiple places. It is being accessed directly, not via `is_up_to_date()`. Presumably this is to avoid having to take the wallet lock. I don't understand the logic of this...
I mean, there is the `up_to_date` field, the `is_up_to_date()` method, and the `set_up_to_date()` method. The methods take wallet.lock, but the direct access of the field obviously does not. Luckily direct accesses of `up_to_date` are always reads.
My question is, do you see a reason why `is_up_to_date()` needs to take wallet.lock? If it did not take it, it would be pretty straight-forward to make the underlying field private.

Done on master in https://github.com/spesmilo/electrum/commit/837fc1606cc83f99f8493c050b384cfcbb6c2835, and PR rebased.

Issue #7750, opened 2022-04-01
Qt GUI freezes on exit if wallet.task is awaiting some future


ecdsa:
reported in #7713 
My proposal: register all those futures, and cancel them when we call `wallet.task.stop()`

SomberNight:
Example: if we are attempting to pay a LN invoice and the user closes the app, the gui freezes until the payment attempts either succeed or fail. (instead, the attempts should simply be cancelled)

PR #7756, opened 2022-04-07
qt taskthread cleaner shutdown

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/7750

`window.run_coroutine_from_thread` no longer uses `wallet.thread`, hence the underlying coroutines are not serialised anymore (they can now run concurrently). Note that this means e.g. the GUI now allows trying to pay multiple LN invoices concurrently.

Also, each task we schedule on `TaskThread` (`wallet.thread`) can provide an optional `cancel` method.
When stopping `TaskThread`, we call this `cancel` method on all tasks in the queue. If the currently running task does not implement `cancel`, `TaskThread.stop` will block until that task finishes.

### DIFF
--- a/electrum/gui/qt/bip39_recovery_dialog.py
+++ b/electrum/gui/qt/bip39_recovery_dialog.py
@@ -2,6 +2,9 @@
 # Distributed under the MIT software license, see the accompanying
 # file LICENCE or http://www.opensource.org/licenses/mit-license.php
 
+import asyncio
+import concurrent.futures
+
 from PyQt5.QtCore import Qt
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QGridLayout, QLabel, QListWidget, QListWidgetItem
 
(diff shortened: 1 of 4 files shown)


SomberNight:
> since `wallet.thread` is a Qt object, maybe it should be a field of `window` rather than `wallet`. That would be consistent with the addition of a `TaskThread` to the bip39 recovery dialog.

I guess `wallet.thread` was added to wallet for hardware wallets, and in some cases they had easier access to the wallet than to the window (and maybe also conceptually it's a bit ~purer):
https://github.com/spesmilo/electrum/commit/06c262d0dc4c90c0478234d51e14df1fb7091bcc
however, since then, hardware wallets now have these `TaskThread` objects per keystore:
https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/electrum/plugins/hw_wallet/qt.py#L227
so yes, it would probably make more sense to move it from `wallet` to `window`.

> is the serialization of tasks in `TaskThread` useful anywhere? 

Not sure... but I think it might not be. It largely seems to be a side-effect.

> why is the `cancel` method optional? are there tasks that should not have one?

What about non-async stuff? How would you want to cancel that?
You could wrap it in a `concurrent.futures.Future`, and cancel the future, but what exactly does that give us? It's a synchronous call which if started, cannot be stopped cleanly.

For example:
https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/electrum/plugins/hw_wallet/qt.py#L270-L274
https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/electrum/gui/qt/main_window.py#L1811-L1817

You are right that there are more things that could be rewritten to be nicely cancellable, mainly network traffic related things, such as
- tx broadcast
  https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/electrum/gui/qt/main_window.py#L1858-L1859
- channel open:
  https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/electrum/gui/qt/main_window.py#L1904
- sweeping:
  https://github.com/spesmilo/electrum/blob/ad41f4aed01a8848ce9139f22f84a433c3b7c3cc/electrum/gui/qt/main_window.py#L3160-L3163
maybe even more, depending on how ambitious we are...
Many of these use `WaitingDialog`, so that too would need a ~rewrite.

So yes, we could make many of the `WaitingDialog` stuff cancellable (so that closing the dialog would cancel the operation).
That seemed much larger scope than the original issue, but it could be done.

SomberNight:
>> since `wallet.thread` is a Qt object, maybe it should be a field of `window` rather than `wallet`. That would be consistent with the addition of a `TaskThread` to the bip39 recovery dialog.
> 
> [...] so yes, it would probably make more sense to move it from `wallet` to `window`.

done

If you think it's worth it, I could look into making the `WaitingDialog`s "usefully cancellable". It's a "nice to have" but I don't think it's urgent - and it's certainly beyond the scope of fixing the original issue.

PR #7759, opened 2022-04-11
android: create relative .buildozer symlink at correct folder, regardless of CWD

accumulator:


SomberNight (approved):
Looks good. Thanks.

PR #7765, opened 2022-04-14
Update git version in Dockerfile

michelvankessel:
Git 2.17.1-1ubuntu0.9 is not available anymore, it is should be git=1:2.17.1-1ubuntu0.10

see https://packages.ubuntu.com/bionic/amd64/git/download


accumulator:
closed by #7764

Issue #7777, opened 2022-04-22
invoice-key mismatch for pre-4.0.8 outgoing onchain invoices


SomberNight:
On master, in the Qt `Send` tab, right-clicking an old onchain invoice results in an error:
```
Traceback (most recent call last):
  File "/home/user/wspace/electrum/electrum/gui/qt/invoice_list.py", line 155, in create_menu
    if invoice.is_lightning():
AttributeError: 'NoneType' object has no attribute 'is_lightning'
```

This is because around version 4.0.8, the key for outgoing invoices was changed: https://github.com/spesmilo/electrum/commit/cdfaaa260942b807f809c2c0414fb242a03e945a

such example invoices I have saved in a wallet:
```
a1. inv=Invoice(amount_msat=100000000, message='', time=1591209612, exp=0, outputs=[<TxOutput script=00140f4dcc26e0ac0ec1890223c4b8c5b48808cc1d67 address=tb1qpaxucfhq4s8vrzgzy0zt33d53qyvc8t8pnqg8n value=100000>], height=0, bip70=None, lightning_invoice=None)
a2. realkey='654b80e92a5032f63c251cad2ce8d109' expected_key='bcc8ea937f'
a1. inv=Invoice(amount_msat=1000000000, message='', time=1592943967, exp=0, outputs=[<TxOutput script=6a0401020304 address=None value=0>, <TxOutput script=ac21020202020202020202020202020202020202020202020202020202020202020202 address=None value=1000000>], height=0, bip70=None, lightning_invoice=None)
a2. realkey='35c7f061df3f00e195af3b205492468e' expected_key='18e079e36e'
a1. inv=Invoice(amount_msat=5500000000000, message='', time=1593687959, exp=0, outputs=[<TxOutput script=0014ec8f27fdb283bfb5f76937e3640c1e4631e9827c address=tb1qaj8j0ldjswlmtamfxl3kgrq7gcc7nqnun4dwn6 value=5500000000>], height=0, bip70=None, lightning_invoice=None)
a2. realkey='849c7f4067a248ad1dfd9a3e8c62b29c' expected_key='8d9fdab586'
```

Until recently (https://github.com/spesmilo/electrum/commit/e392197ab94715815c00503d8fd5cb881d0a26c3), there was an "id" field stored in invoices, but now we are computing the id on demand using the "new algorithm", which does not work for pre-4.0.8 invoices.

SomberNight:
bip70 invoices also do not work.
[...]

PR #7778, opened 2022-04-22
wallet_db upgrade: recalc keys of outgoing on-chain invoices

SomberNight:
I think this should fix https://github.com/spesmilo/electrum/issues/7777

- the key of outgoing on-chain invoices is redefined to avoid using `TxOutput.__repr__()`
- a wallet_db upgrade is included that rekeys the (outgoing) `invoices` dict accordingly
- `PaymentRequest.get_id()` is removed

### DIFF
--- a/electrum/gui/kivy/main_window.py
+++ b/electrum/gui/kivy/main_window.py
@@ -19,7 +19,7 @@ from electrum import util
 from electrum.util import (profiler, InvalidPassword, send_exception_to_crash_reporter,
                            format_satoshis, format_satoshis_plain, format_fee_satoshis,
                            maybe_extract_bolt11_invoice, parse_max_spend)
-from electrum.invoices import PR_PAID, PR_FAILED
+from electrum.invoices import PR_PAID, PR_FAILED, Invoice
 from electrum import blockchain
 from electrum.network import Network, TxBroadcastError, BestEffortRequestFailed
 from electrum.interface import PREFERRED_NETWORK_PROTOCOL, ServerAddr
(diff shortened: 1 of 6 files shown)


ecdsa:
note: `_convert_version_25` is now failing because `PaymentRequest` has no `id` field

ecdsa:
This could fix it:
```diff
         self.data['seed_version'] = 24
 
     def _convert_version_25(self):
+        from .crypto import sha256
         if not self._is_upgrade_method_needed(24, 24):
             return
         # add 'type' field to onchain requests
@@ -578,17 +579,19 @@ class WalletDB(JsonDB):
         # convert bip70 invoices
         invoices = self.data.get('invoices', {})
         for k, r in list(invoices.items()):
-            data = r.get("hex")
+            hex_data = r.get("hex")
+            data = bytes.fromhex(hex_data)
+            pr_id = sha256(data)[0:16].hex()
             if data:
-                pr = PaymentRequest(bytes.fromhex(data))
-                if pr.id != k:
+                pr = PaymentRequest(data)
+                if pr_id != k:
                     continue
                 invoices[k] = {
                     'type': PR_TYPE_ONCHAIN,
                     'amount': pr.get_amount(),
-                    'bip70': data,
+                    'bip70': hex_data,
                     'exp': pr.get_expiration_date() - pr.get_time(),
-                    'id': pr.id,
+                    'id': pr_id,
                     'message': pr.get_memo(),
                     'outputs': [x.to_legacy_tuple() for x in pr.get_outputs()],
                     'time': pr.get_time(),
```

However, the deep reason why this problem occurred is that we instantiate a `PaymentRequest` in the wallet upgrade. 
This is not robust, and should be avoided. I would not mind changing `_convert_version_25` so that it deletes all bip70 requests.

SomberNight:
> note: `_convert_version_25` is now failing because `PaymentRequest` has no `id` field

Ah, right.

> However, the deep reason why this problem occurred is that we instantiate a `PaymentRequest` in the wallet upgrade. 
This is not robust, and should be avoided. I would not mind changing `_convert_version_25` so that it deletes all bip70 requests.

Indeed. This is another example that we need to be careful with imporing other modules of the code inside a wallet upgrade :/
OK, I've pushed a second commit to the branch to change `_convert_version_25` to delete bip70 invoices. That upgrade was released almost 2 years ago, so deleting such old invoices (and only for people who have not opened their wallets since) is relatively safe I think.

Mdk2022 (approved):
Thanks

Mdk2022 (approved):

PR #7792, opened 2022-04-29
asyncio: stop using get_event_loop(). introduce ~singleton loop.

SomberNight:
`asyncio.get_event_loop()` became deprecated in python3.10. (see https://github.com/python/cpython/issues/83710)
```
.../electrum/electrum/daemon.py:470: DeprecationWarning: There is no current event loop
  self.asyncio_loop = asyncio.get_event_loop()
.../electrum/electrum/network.py:276: DeprecationWarning: There is no current event loop
  self.asyncio_loop = asyncio.get_event_loop()
```
Also, according to that thread, "set_event_loop() [... is] not deprecated by oversight".
So, we stop using `get_event_loop()` and `set_event_loop()` in our own code.
Note that libraries we use (such as the stdlib for python <3.10), might call `get_event_loop`,
which then relies on us having called `set_event_loop` e.g. for the GUI thread. To work around
this, a custom event loop policy providing a `get_event_loop` implementation is used.

Previously, we have been using a single asyncio event loop, created with
`util.create_and_start_event_loop`, and code in many places got a reference to this loop
using `asyncio.get_event_loop()`.
Now, we still use a single asyncio event loop, but it is now stored as a global in
`util._asyncio_event_loop` (access with `util.get_asyncio_loop()`).

I believe these changes also fix https://github.com/spesmilo/electrum/issues/5376

### DIFF
--- a/electrum/tests/__init__.py
+++ b/electrum/tests/__init__.py
@@ -6,6 +6,7 @@ import shutil
 import electrum
 import electrum.logging
 from electrum import constants
+from electrum import util
 
 
 # Set this locally to make the test suite run faster.
@@ -37,9 +38,12 @@ class ElectrumTestCase(SequentialTestCase):
 
     def setUp(self):
         super().setUp()
+        self.asyncio_loop, self._stop_loop, self._loop_thread = util.create_and_start_event_loop()
         self.electrum_path = tempfile.mkdtemp()
 
     def tearDown(self):
+        self.asyncio_loop.call_soon_threadsafe(self._stop_loop.set_result, 1)
+        self._loop_thread.join(timeout=1)
         super().tearDown()
         shutil.rmtree(self.electrum_path)
 
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -187,7 +187,7 @@ class Commands:
                 kwargs.pop('wallet')
 
         coro = f(*args, **kwargs)
-        fut = asyncio.run_coroutine_threadsafe(coro, asyncio.get_event_loop())
+        fut = asyncio.run_coroutine_threadsafe(coro, util.get_asyncio_loop())
         result = fut.result()
 
         if self._callback:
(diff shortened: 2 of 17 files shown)


electrum/tests/__init__.py
```diff
@@ -37,9 +38,12 @@ class ElectrumTestCase(SequentialTestCase):
 
     def setUp(self):
         super().setUp()
+        self.asyncio_loop, self._stop_loop, self._loop_thread = util.create_and_start_event_loop()
         self.electrum_path = tempfile.mkdtemp()
 
     def tearDown(self):
```
ecdsa, line 44: why is `super().tearDown()` not called first?

ecdsa:
great work
note: `create_and_start_event_loop` no longer needs to be imported from a few files.

electrum/tests/__init__.py
```diff
@@ -37,9 +38,12 @@ class ElectrumTestCase(SequentialTestCase):
...
+        self.asyncio_loop.call_soon_threadsafe(self._stop_loop.set_result, 1)
```
ecdsa, line 45: why is `super().tearDown()` not called first?

SomberNight: Although in many cases the order does not matter, here it might, depending on the unit test framework running the tests I guess: `SequentialTestCase.setUp()` and `SequentialTestCase.tearDown()` take and release `self.test_lock`, which ensure that tests do not run in parallel. If `super().tearDown()` is called before we stop the asyncio loop, and the test framework supports running tests in parallel, some tests might fail when trying to create an "overlapping" event loop.

The natural order of calling destructors is the reverse of constructors: Create parent first, child next, but destroy child first, parent next (btw this is the order in C++, where the calls happen implicitly). Code in the child might be relying on code in the parent, so the parent shouldn't be destroyed first. I guess we could clean up the rest of the `tearDown` implementations to respect this.

SomberNight:
> note: `create_and_start_event_loop` no longer needs to be imported from a few files.

removed.

PR #7797, opened 2022-05-04
vendor a part of `distutils`, to be removed from stdlib in py3.12

SomberNight:
[`distutils`](https://github.com/python/cpython/tree/9d38120e335357a3b294277fd5eff0a10e46e043/Lib/distutils) (a module in the python standard library that we use) got deprecated in python 3.10,
and is [planned to be removed](https://github.com/python/cpython/blob/9d38120e335357a3b294277fd5eff0a10e46e043/Lib/distutils/__init__.py#L16) from the stdlib in python 3.12.

We only use it in the Qt `update_checker` (for signed version notifications), to compare version numbers.
That is, we only use a very small part of it (only `distutils.version.StrictVersion`).

```
.../electrum/electrum/gui/qt/update_checker.py:7: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils.version import StrictVersion
```

Migration advice in PEP-632 suggests ( https://peps.python.org/pep-0632/#migration-advice ):
> `distutils.version` — use the `packaging` package

Note that [`packaging`](https://github.com/pypa/packaging) is a 3rd party package, i.e. it would be a new dependency.
Also, it does not provide an identical replacement for `distutils.version.StrictVersion`.

Care needs to be taken when changing the semantics of version numbers...
E.g. `packaging.parse` and `packaging.Version` are less strict than what we currently use.
We have to be careful that old code recognises new electrum version numbers as both valid
and numerically greater than their version.

Alternatively, we could reimplement `distutils.version.StrictVersion` ourselves.
Or we could implement something much simpler, but carefully consider the interactions with old electrum versions.

-----

I think the easiest approach is for us to vendor a small part of distutils, as in this PR: only the standalone `version.py` file.
The `LICENSE` file is from cpython.

Re directory structure, this is based on `pip`:
https://github.com/pypa/pip/tree/bab5bfce502be78318ab2a3b364b4923d657c854/src/pip/_vendor
(although the approach here is much more naive ofc)

### DIFF
--- a/electrum/_vendor/distutils/LICENSE
+++ b/electrum/_vendor/distutils/LICENSE
@@ -1,0 +1,279 @@
+A. HISTORY OF THE SOFTWARE
+==========================
+
+Python was created in the early 1990s by Guido van Rossum at Stichting
+Mathematisch Centrum (CWI, see http://www.cwi.nl) in the Netherlands
+as a successor of a language called ABC.  Guido remains Python's
+principal author, although it includes many contributions from others.
+
+In 1995, Guido continued his work on Python at the Corporation for
+National Research Initiatives (CNRI, see http://www.cnri.reston.va.us)
+in Reston, Virginia where he released several versions of the
+software.
+
+In May 2000, Guido and the Python core development team moved to
+BeOpen.com to form the BeOpen PythonLabs team.  In October of the same
+year, the PythonLabs team moved to Digital Creations, which became
+Zope Corporation.  In 2001, the Python Software Foundation (PSF, see
+https://www.python.org/psf/) was formed, a non-profit organization
+created specifically to own Python-related Intellectual Property.
+Zope Corporation was a sponsoring member of the PSF.
+
+All Python releases are Open Source (see http://www.opensource.org for
+the Open Source Definition).  Historically, most, but not all, Python
+releases have also been GPL-compatible; the table below summarizes
+the various releases.
+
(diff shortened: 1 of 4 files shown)


ecdsa:
> Care needs to be taken when changing the semantics of version numbers...
E.g. `packaging.parse` and `packaging.Version` are less strict than what we currently use.
We have to be careful that old code recognises new electrum version numbers as both valid
and numerically greater than their version.

do you have an example of that?

SomberNight:
```
>>> import packaging.version
>>> packaging.version.parse("4.3.0rc1")
<Version('4.3.0rc1')>
>>> from distutils.version import StrictVersion
>>> StrictVersion("4.3.0rc1")
  File "<stdin>", line 1, in <module>
  File "/home/user/.local/lib/python3.8/site-packages/setuptools/_distutils/version.py", line 54, in __init__
    self.parse(vstring)
  File "/home/user/.local/lib/python3.8/site-packages/setuptools/_distutils/version.py", line 157, in parse
    raise ValueError("invalid version number '%s'" % vstring)
ValueError: invalid version number '4.3.0rc1'
```

Let's say, new code started using `packaging.version.parse` and we released a version number with an `rc` string in it. Our new code would not complain that it's not a valid version. Old Electrum however would consider it an invalid version string, log an exception, and not show an update notification.

PR #7809, opened 2022-05-10
Update wording during onboarding wizard

Benjamin-Lee:
I'm a new user and found the wording a little awkward. How's this instead?


ecdsa:
this is broken

PR #7812, opened 2022-05-13
Show message signature as QR code

baam25simo:
Added a simple button to allow the user to get a message signature just by scanning a QR code and not copying it by hand to another device.

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -2895,8 +2895,9 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
         layout.addWidget(QLabel(_('Address')), 2, 0)
         layout.addWidget(address_e, 2, 1)
 
-        signature_e = QTextEdit()
-        signature_e.setAcceptRichText(False)
+        signature_e = ButtonsTextEdit()
+        qr_icon = "qrcode_white.png" if ColorScheme.dark_scheme else "qrcode.png"
+        signature_e.addButton(qr_icon, lambda: self.show_qrcode(signature_e.text()), _("Show QR Code"))
         layout.addWidget(QLabel(_('Signature')), 3, 0)
         layout.addWidget(signature_e, 3, 1)
         layout.setRowStretch(3,1)


SomberNight (changes requested):
Let's not add another top-level button at the bottom for this...
Why not just do `signature_e = ShowQRTextEdit(config=self.config)` on line 2898?

EDIT: note that the text edit needs to be editable for the "verify" use case.
Also, in the "sign" use case, you might want to show a signature-qr-code,
while in the "verify" use case, you might want to scan a signature-qr-code...

SomberNight:
What are you using to scan the signature qr code btw? Is there an app that expects sigs as qr codes, or what specifically is your use case?

baam25simo:
> What are you using to scan the signature qr code btw? Is there an app that expects sigs as qr codes, or what specifically is your use case?

Yes, there is an app which requires to demonstrate that the address is solely controlled by a wallet. To do so, it requires to sign a message containing a code and then requires the signature in order to verify the ownership.

SomberNight:
Right, I did not intend to question the utility of message-signing as a whole. I am questioning the QR code part :)

Are you signing the message on your PC and want to show a QR code there which then you scan with your phone for an app running on the phone? And what about the message - does the phone app give you a message to sign? If so, how do you copy that to the PC?
Or what exactly is the flow?

baam25simo:
> Right, I did not intend to question the utility of message-signing as a whole. I am questioning the QR code part :)
> 
> Are you signing the message on your PC and want to show a QR code there which then you scan with your phone for an app running on the phone? And what about the message - does the phone app give you a message to sign? If so, how do you copy that to the PC? Or what exactly is the flow?

The flow follows:
The app running on the phone requires to demonstrate the ownership of an address. To do so it requires a signature of a custom message which itself provides like "I solely own this address. Secret code: XXXXXXXX" where the secret code can be from 8 to 10 alphanumerical characters, so no effort to type it on pc on contrary to type on the phone the required signature. Fortunately the app provide a QR scanner for the signature. So it could be very helpful and less time consuming allow to display a message signature in QR format.

If you agree I will push your adviced code on line 2898.

SomberNight:
>  The app running on the phone [...] requires a signature of a custom message [...] like "I solely own this address. Secret code: XXXXXXXX" where the secret code can be from 8 to 10 alphanumerical characters, so no effort to type it on pc on contrary to type on the phone the required signature. Fortunately the app provide a QR scanner for the signature. So it could be very helpful and less time consuming allow to display a message signature in QR format.
>
> If you agree I will push your adviced code on line 2898.

All right. You should test it a bit as what I said is insufficient, e.g. also needs `signature_e.setReadOnly(False)` for the "verify" use case.

> The app running on the phone [...] requires a signature of a custom message [...] like "I solely own this address. Secret code: XXXXXXXX" where the secret code can be from 8 to 10 alphanumerical characters, so no effort to type it on pc

Does the app offer showing a QR code for the message?

baam25simo:
> > The app running on the phone [...] requires a signature of a custom message [...] like "I solely own this address. Secret code: XXXXXXXX" where the secret code can be from 8 to 10 alphanumerical characters, so no effort to type it on pc on contrary to type on the phone the required signature. Fortunately the app provide a QR scanner for the signature. So it could be very helpful and less time consuming allow to display a message signature in QR format.
> > If you agree I will push your adviced code on line 2898.
> 
> All right. You should test it a bit as what I said is insufficient, e.g. also needs `signature_e.setReadOnly(False)` for the "verify" use case.
> 
> > The app running on the phone [...] requires a signature of a custom message [...] like "I solely own this address. Secret code: XXXXXXXX" where the secret code can be from 8 to 10 alphanumerical characters, so no effort to type it on pc
> 
> Does the app offer showing a QR code for the message?

No, it does not since the message is easy and short to type on pc.

baam25simo:
Understood.
Could you suggest some test to perform?
I just used electrum so I do not know the code well, as you possibly noted.

Thank you in advance for the support.

electrum/gui/qt/main_window.py
```diff
@@ -2895,7 +2895,8 @@ def sign_verify_message(self, address=''):
         layout.addWidget(QLabel(_('Address')), 2, 0)
         layout.addWidget(address_e, 2, 1)
 
-        signature_e = QTextEdit()
+        signature_e = ShowQRTextEdit(config=self.config)
+        signature_e.setReadOnly(False)
         signature_e.setAcceptRichText(False)
```
SomberNight, line 2900: This line is no longer needed, as `ShowQRTextEdit` inherits `QPlainTextEdit`, so it is incapable of rich text, and this method is undefined.

```
  File "/home/user/wspace/electrum/electrum/gui/qt/address_list.py", line 270, in <lambda>
    menu.addAction(_("Sign/verify message"), lambda: self.parent.sign_verify_message(addr))
  File "/home/user/wspace/electrum/electrum/gui/qt/main_window.py", line 2900, in sign_verify_message
    signature_e.setAcceptRichText(False)
AttributeError: 'ShowQRTextEdit' object has no attribute 'setAcceptRichText'
```

SomberNight: >  Could you suggest some test to perform?

I just meant try signing and verifying a message, show a qr code -- just play around with the functionality that is being changed :)
To catch issues like the one above.

baam25simo: Ok, I will try. Thank you!

electrum/gui/qt/qrcodewidget.py
```diff
@@ -20,7 +20,8 @@ def __init__(self, data = None, fixedSize=False):
         QWidget.__init__(self)
         self.data = None
         self.qr = None
-        self.margin = 0
+        
+        self.margin = 1
```
SomberNight, line 24: > Set `QRCodeWidget` margin not 0 since in dark mode the
> default created QR code is not properly scanned.

Hmm, good observation. Still, it's an independent issue. Could you please split this out into a separate PR?

SomberNight (commented):
Looks good! Just one small thing, see inline.

SomberNight:
(I've force-pushed the branch to squash the commits and to remove a redundant newline.)

Thanks for the PR.

electrum/gui/qt/qrcodewidget.py
SomberNight: (done in https://github.com/spesmilo/electrum/pull/7813)

PR #7813, opened 2022-05-14
Set `QRCodeWidget` margin not null

baam25simo:
This because in dark mode the default created QR code is not properly scanned.


SomberNight:
@ecdsa this was recently touched in https://github.com/spesmilo/electrum/commit/ad70659a666987951061d73aebe4b9410ed96eb0
I guess you changed it with the Receive tab in mind. However note that in many other places, the margin was useful: it made QR codes (when using the dark theme at least) significantly easier to scan.
See e.g.:
txdialog

ecdsa:
indeed, it was for the receive tab.
if it is needed elsewhere, we should pass it as a parameter.

baam25simo:
I agree with you. The best option is pass it as parameter with zero as default value.
If it is ok for you I can do it.

SomberNight:
Right, let's pass it as a parameter then. But the default should be the old value (10) I think.
In the case of the receive tab, we can pass 0. So probably here:
https://github.com/spesmilo/electrum/blob/3cc6c0dd2d5d99506650ce2f990a0d69b6579acb/electrum/gui/qt/main_window.py#L1199-L1201

SomberNight:
>  If it is ok for you I can do it.

Feel free to.

baam25simo:
@SomberNight I tryed to fix the issue adapting the code to your changes. Could you review it and tell me if a further refactoring is required?

ecdsa:
this branch is conflicting.
you need rebase your code against current master, and force push

PR #7817, opened 2022-05-18
Fix RuntimeWarning

LorenzoAncora:
Fix `RuntimeWarning: coroutine 'Interface.run' was never awaited` raised by `tests/test_network.py`.

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -399,7 +399,7 @@ class Interface(Logger):
         self.taskgroup = OldTaskGroup()
 
         async def spawn_task():
-            task = await self.network.taskgroup.spawn(self.run())
+            task = await self.network.taskgroup.spawn(await self.run())
             task.set_name(f"interface::{str(server)}")
         asyncio.run_coroutine_threadsafe(spawn_task(), self.network.asyncio_loop)
 
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -399,7 +399,7 @@ class Interface(Logger):
         self.taskgroup = OldTaskGroup()
 
         async def spawn_task():
-            task = await self.network.taskgroup.spawn(await self.run())
+            task = await self.network.taskgroup.spawn(self.run())
             task.set_name(f"interface::{str(server)}")
         asyncio.run_coroutine_threadsafe(spawn_task(), self.network.asyncio_loop)
 
--- a/electrum/tests/test_network.py
+++ b/electrum/tests/test_network.py
@@ -12,6 +12,11 @@ from electrum import util
 
 from . import ElectrumTestCase
 
+try:
+    from pytest import mark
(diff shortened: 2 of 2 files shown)


electrum/interface.py
```diff
@@ -399,7 +399,7 @@ def __init__(self, *, network: 'Network', server: ServerAddr, proxy: Optional[di
         self.taskgroup = OldTaskGroup()
 
         async def spawn_task():
-            task = await self.network.taskgroup.spawn(self.run())
+            task = await self.network.taskgroup.spawn(await self.run())
```
SomberNight, line 402: `await self.run()` returns None, or at least it would, except it is effectively an infinite-loop which can only exit via exceptions.
`await self.network.taskgroup.spawn(None)` does not make sense and would error, but it never gets to execute.

what you effectively have here between lines 401-404 could then just be replaced with
`asyncio.run_coroutine_threadsafe(self.run(), self.network.asyncio_loop)`

With your change, the interfaces run directly via `run_coroutine_threadsafe` and bypass `network.taskgroup`, which means that they are not closed cleanly if the network is restarted (e.g. due to proxy changes) or during application shutdown.

SomberNight (changes requested):
This does not fix the issue, just hides the warning.
Note that the warning is restricted to the tests as its true cause is specific to the tests.
This change however introduces problems for the main code. See comment inline.

> Fix `RuntimeWarning: coroutine 'Interface.run' was never awaited` raised by `tests/test_network.py`.

The warning is printed because the tests use `MockNetwork`, and `MockNetwork.taskgroup` is indeed not awaited/joined, but when running the main code, `network.taskgroup` is used properly and is unaffected.

LorenzoAncora:
Hi @SomberNight, thank you for sharing your specific knowledge. I thought someone simply forgot the `await` keyword but, if what you wrote is true, this is a bug in the test cases themselves (which should not give phony errors/warnings, obviously). However, we:

1. cannot leave a warning in the test cases;
2. cannot introduce a bug in the main code.

Hence, we need to modify the test case so the warning is not printed. Are you ok with this plan?

SomberNight:
> However, we:
>    cannot leave a warning in the test cases;
>    cannot introduce a bug in the main code.

well I would rather leave the warning in the tests if I had to choose :P

> Hence, we need to modify the test case so the warning is not printed

sure, feel free the have a look at the tests and the mocks

LorenzoAncora:
> Hence, we need to modify the test case so the warning is not printed
>> sure, feel free the have a look at the tests and the mocks

@SomberNight ok, problem solved. I've extended the code to ignore only this specific warning when pytest is used like in the official docs. I've been intentionally a bit more verbose than necessary to ensure the total non-intrusiveness of the new code. Can you merge?

PR #7818, opened 2022-05-18
Filter nodes for receiving:

ecdsa:
- increase MPP_RECEIVE_CUTOFF from 5 to 20 percent
 - filter channels by node_id
 - make num_sats_can_receive consistent with routing hints


electrum/lnworker.py
```diff
@@ -2002,16 +1998,12 @@ def htlc_failed(
...
-        # Even with MPP, below a certain threshold, including these channels probably
-        # hurts more than help, as they lead to many failed attempts for the sender.
-        channels = [chan for chan in channels
-                    if chan.available_to_spend(REMOTE) > (amount_msat or 0) * 0.05]
+        with self.lock:
+            nodes = self.border_nodes_that_can_receive(amount_msat)
+            channels = []
+            for c in self.channels.values():
+                if c.node_id in nodes:
+                    channels.append(c)
```
SomberNight, line 2006: Ah, actually, we might have to distinguish based on whether the sender is paying e2e trampoline or just doing simple legacy payment.

---

Consider scenario: Alice wants to recv 0.5 BTC and has two channels with border-node T1.
- chan1 can recv 1 BTC
- chan2 can recv 0.001 BTC

if the sender is doing a simple legacy payment, including chan2 as an `r` hint in the invoice is harmful.

---

The same affects the other strategy we talked about, where we sort the chans by recv capacity in decreasing order: is it the chans or the "meta-channels" (all chans aggregated for given border node) that should be ordered and considered?

Maybe trampoline_hints should consider meta-channels, and `r` routing_hints should consider channels?

SomberNight: Re scenario1, I think I even confused myself a bit, so let me rephrase:

if the sender is paying e2e trampoline, then as Alice is directly connected to T1, T1 can sort of `sum` Alice's recv capacities - as it is free to resplit the payment. So Alice can receive up to 1.001 BTC.
However, if the sender is paying legacy, it is highly unlikely they manage to send anything over 1 BTC to Alice.

PR #7819, opened 2022-05-18
Fix UI hang when selecting 'Ok' on change password window on a hardware wallet when no hardware is connected

kralverde:
I specifically noticed this on my ledger device; pressing 'ok' when no ledger is connected leads to an infinite hang.

This pr puts the `get_client` method in no interaction mode which raises a failure when no devices are found and then raises a relevant pop up to the user.

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -2553,6 +2553,9 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger):
                 hw_dev_pw = self.wallet.keystore.get_password_for_storage_encryption()
             except UserCancelled:
                 return
+            except ValueError:
+                self.show_error('No hardware device detected')
+                return
             except BaseException as e:
                 self.logger.exception('')
                 self.show_error(repr(e))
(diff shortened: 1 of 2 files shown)


SomberNight:
> Fix UI hang when selecting 'Ok' on change password window on a hardware wallet when no hardware is connected 
> I specifically noticed this on my ledger device; pressing 'ok' when no ledger is connected leads to an infinite hang.

Indeed.
The issue is that `change_password_dialog` runs in the main GUI thread,
which calls `keystore.get_password_for_storage_encryption`, which will schedule a task onto the hardware-device-comms thread and wait for that, blocking the GUI thread. That task, now running in the hwd-thread, if the hwd is disconnected and user-interaction is enabled, will try to show messages in the GUI, asking the user to reconnect the device and will wait for dialog-interaction - these messages are passed via signals to the Qt GUI thread, but the signals are never processed and the messages/dialogs never drawn as the GUI thread is blocking on the task - so it is a deadlock.

Thanks for the report and for looking at the code and implementing a fix; I've opted to fix it differently.

PR #7828, opened 2022-05-24
Fix "== None" linter fails

JeremyRand:
Credit to Zoltan Konder for pointing me to this issue.


JeremyRand:
The Tox fails on Cirrus look unrelated to this PR, I think.

PR #7830, opened 2022-05-24
lnpeer.reestablish_chan: enforce order of replaying commitsig/revack

SomberNight:
When replaying messages during channel-reestablishment,
previously we first resent all update messages, along with potential commitment_signed messages,
and then we potentially resent a single revoke_and_ack.

This can result in incorrect behaviour in case both a commitment_signed and a revoke_and_ack needs to be resent.
When replaying messages, the relative order of commitment_signed and revoke_and_messages needs to be preserved.
(the order of updates (htlc/fee) in relation to the revack messages does not matter)

implements https://github.com/lightning/bolts/pull/810

The logic here is somewhat based on what c-lightning does:
https://github.com/ElementsProject/lightning/blob/01e5f1886e31816e652f417a1ff789a26aaeec3b/channeld/channeld.c#L3059


SomberNight:
note: I also considered including `revoke_and_ack` messages in `unacked_updates` (in lnhtlc's log). However it does not fit well into that model. That data structure is indexed based on remote ctns: for a given remote ctn, it stores a bunch of `update_*` messages followed by at most one `commitment_signed` message. However, there are no restrictions on how many `revoke_and_ack`s we might send "during" any given remote ctn...

PR #7842, opened 2022-06-02
Qt PayToEdit: add a "help" button, next to camera/paste/file

SomberNight:
I think we could add a "help" icon inside the Qt PayToEdit field, explaining some of the supported functionality.
Such an explanation already exist when clicking the "Pay to" label text itself, however I suspect not many users discover that.

Sort of related: https://github.com/spesmilo/electrum/pull/6872

(idea from https://github.com/Electron-Cash/Electron-Cash/commit/3ca8854007556363c6ef201aac5fe275294abf7e )

paytoedit_help_btn

(icon from icons8.com, like others)


ecdsa:
I do not think that adding more buttons to "help users" is the solution. We already added a question mark to the "Expires" label in the Receive tab, on the suspicion that users do not see the clickable links. Was that really useful? Why should that question mark be an icon in the Send tab, and text in the Receive tab? 

Second, I do not think that usability issues arise from the "Pay to" field. What category of users are you  trying to help here? Why would they find it easier to click a question mark icon, than the other icons next to it?

I think most people will understand how to copy and paste in a field, or how to click a camera icon. However, the same people who are able to figure out how to click an icon may have difficulties understanding advanced concepts such as  utxos, rbf, cpfp or lightning channels. I think that is the real challenge. We need to simplify the use of advanced concepts, and to hide their complexity.

PR #7858, opened 2022-06-16
EventListener class to handle callbacks

ecdsa:
This is a refactoring of callbacks in `util.py.` Event listeners must be methods of the EventListener class, and named `'on_event_' + event_name`

It is still a draft, a few things need to be decided:
1. We could add a `@event_listener` decorator to all event listeners, that does nothing but asserts their name starts with `on_event_`, for pedagogic purposes.
2. Alternatively, we could allow arbitrary names, and add a decorator that takes the event's name as parameter.
3. We could try to add a filtering option to the decorator (to avoid starting methods with `if self.wallet != wallet: return`). Not sure if it can be done in a way that removes the current quadratic complexity ("everyone listens to everyone")
4. I am not sure to understand how `QNetworkUpdatedSignalObject` worked before this commit. I might have broken something there.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -37,6 +37,7 @@ from .verifier import SPV
 from .blockchain import hash_header, Blockchain
 from .i18n import _
 from .logging import Logger
+from .util import EventListener, event_listener
 
 if TYPE_CHECKING:
     from .network import Network
@@ -57,7 +58,7 @@ class HistoryItem(NamedTuple):
     balance: int
 
 
-class AddressSynchronizer(Logger):
+class AddressSynchronizer(Logger, EventListener):
     """ address database """
 
     network: Optional['Network']
@@ -181,9 +182,10 @@ class AddressSynchronizer(Logger):
             self.synchronizer = Synchronizer(self)
             self.verifier = SPV(self.network, self)
             self.asyncio_loop = network.asyncio_loop
-            util.register_callback(self.on_blockchain_updated, ['blockchain_updated'])
+            self.register_callbacks()
 
-    def on_blockchain_updated(self, event, *args):
+    @event_listener
+    def on_event_blockchain_updated(self, *args):
         self._get_balance_cache = {}  # invalidate cache
 
     async def stop(self):
@@ -197,7 +199,7 @@ class AddressSynchronizer(Logger):
             finally:  # even if we get cancelled
                 self.synchronizer = None
                 self.verifier = None
-                util.unregister_callback(self.on_blockchain_updated)
+                self.unregister_callbacks()
                 self.db.put('stored_height', self.get_local_height())
 
     def add_address(self, address):
--- a/electrum/gui/qt/channel_details.py
+++ b/electrum/gui/qt/channel_details.py
@@ -5,7 +5,7 @@ import PyQt5.QtWidgets as QtWidgets
 import PyQt5.QtCore as QtCore
 from PyQt5.QtWidgets import QLabel, QLineEdit, QHBoxLayout
 
-from electrum import util
+from electrum.util import EventListener
 from electrum.i18n import _
 from electrum.util import bh2u, format_time
 from electrum.lnutil import format_short_channel_id, LOCAL, REMOTE, UpdateAddHtlc, Direction
@@ -15,6 +15,7 @@ from electrum.bitcoin import COIN
 from electrum.wallet import Abstract_Wallet
 
 from .util import Buttons, CloseButton, ButtonsLineEdit, MessageBoxMixin, WWLabel
+from .util import QtEventListener, qt_event_listener
 
 if TYPE_CHECKING:
     from .main_window import ElectrumWindow
@@ -35,7 +36,8 @@ class LinkedLabel(QtWidgets.QLabel):
         self.linkActivated.connect(on_clicked)
 
 
-class ChannelDetailsDialog(QtWidgets.QDialog, MessageBoxMixin):
+class ChannelDetailsDialog(QtWidgets.QDialog, MessageBoxMixin, QtEventListener):
+
     def make_htlc_item(self, i: UpdateAddHtlc, direction: Direction) -> HTLCItem:
         it = HTLCItem(_('Sent HTLC with ID {}' if Direction.SENT == direction else 'Received HTLC with ID {}').format(i.htlc_id))
         it.appendRow([HTLCItem(_('Amount')),HTLCItem(self.format_msat(i.amount_msat))])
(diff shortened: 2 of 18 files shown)


electrum/gui/qt/util.py
```diff
@@ -1503,6 +1503,31 @@ def resizeEvent(self, e):
...
+        self.qt_callback_signal.disconnect()
+        EventListener.unregister_callbacks(self)
+
+    def on_qt_callback_signal(self, args):
+        func = args[0]
+        return func(self, *args[1:])
+
+# decorator for members of the QtEventListener class
+def run_in_qt_thread(func):
+    def decorator(self, *args):
```
SomberNight, line 1526:
```suggestion
    @wraps(func)
    def decorator(self, *args):
```
(and `from functools import wraps` at the top)

This makes it IDE-friendlier but also, for example, you are logging in `EventListener.register_callbacks`, and this changes that from:
```
I | util | registering callback <bound method run_in_qt_thread.<locals>.decorator of <electrum.gui.qt.main_window.ElectrumWindow object at 0x7f3b9c1a6d30>>
```
to:
```
I | util | registering callback <bound method ElectrumWindow.on_event_verified of <electrum.gui.qt.main_window.ElectrumWindow object at 0x7f27f2512a60>>
```

electrum/address_synchronizer.py
```diff
@@ -181,9 +182,9 @@ def start_network(self, network: Optional['Network']) -> None:
             self.synchronizer = Synchronizer(self)
             self.verifier = SPV(self.network, self)
             self.asyncio_loop = network.asyncio_loop
-            util.register_callback(self.on_blockchain_updated, ['blockchain_updated'])
+            self.register_callbacks()
 
-    def on_blockchain_updated(self, event, *args):
+    def on_event_blockchain_updated(self, event, *args):
```
SomberNight, line 187: This would probably raise as the callback has no arguments no, and this expects at least "event".
```suggestion
    def on_event_blockchain_updated(self, *args):
```

electrum/lnworker.py
```diff
@@ -477,7 +478,7 @@ def choose_preferred_address(addr_list: Sequence[Tuple[str, int, int]]) -> Tuple
         choice = random.choice(addr_list)
         return choice
 
-    def on_proxy_changed(self, event, *args):
+    def on_event_proxy_set(self, event, *args):
```
SomberNight, line 481:
```suggestion
    def on_event_proxy_set(self, *args):
```

SomberNight:
> 1. We could add a `@event_listener` decorator to all event listeners, that does nothing but asserts their name starts with `on_event_`, for pedagogic purposes.
> 2. Alternatively, we could allow arbitrary names, and add a decorator that takes the event's name as parameter.

We could add a `@event_listener` decorator to all event listeners, and the decorator could both assert the func name has given prefix (`on_event_`) and it could mark the method (e.g. set `func._is_event_listener = True`).
Then, in `EventListener._list_callbacks`, we could filter to attributes that both have name prefic `on_event_` AND have been decorated. This way, it would be ok for unrelated fields/methods to have a name that starts with `on_event_` (not saying we want to have such fields but I think it's nice to allow them -- 3rd party libs might have them; e.g. imagine inheriting from a class that is from a 3rd party lib).
```python
for method_name in dir(self):
    if not method_name.startswith('on_event_'):
        continue
    method = getattr(self, method_name)
    if not getattr(method, "_is_event_listener", False):
        continue
```

SomberNight (commented):
stdio and text guis are broken :P

electrum/gui/qt/network_dialog.py
```diff
@@ -53,36 +54,29 @@
...
+    def on_event_network_updated(self):
         self.nlayout.update()
 
     def clean_up(self):
         if self._cleaned_up:
             return
         self._cleaned_up = True
         self.nlayout.clean_up()
-        self.network_updated_signal_obj.network_updated_signal.disconnect()
         self.network_updated_signal_obj = None
```
SomberNight, line 78:
```suggestion
```

electrum/gui/qt/channel_details.py
```diff
@@ -143,17 +138,8 @@ def __init__(self, window: 'ElectrumWindow', chan_id: bytes):
...
-        self.htlc_failed.connect(self.on_htlc_failed)
-        self.state_changed.connect(self.do_state_changed)
-        self.htlc_added.connect(self.on_htlc_added)
-
         # register callbacks for updating
-        util.register_callback(self.htlc_fulfilled.emit, ['htlc_fulfilled'])
-        util.register_callback(self.htlc_failed.emit, ['htlc_failed'])
-        util.register_callback(self.htlc_added.emit, ['htlc_added'])
-        util.register_callback(self.state_changed.emit, ['channel'])
+        self.register_callbacks()
```
SomberNight, line 142: maybe override `def closeEvent(self, event)` and call `unregister_callbacks()` there

electrum/util.py
```diff
@@ -1594,10 +1594,13 @@ def __init__(self):
         self.callbacks = defaultdict(list)      # note: needs self.callback_lock
         self.asyncio_loop = None
 
-    def register_callback(self, callback, events):
+    def register_callback(self, func, events=None):
+        if events is None:
+            assert func.__name__.startswith('on_'), func.__name__
+            events = [func.__name__[3:]]
```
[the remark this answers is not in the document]
ecdsa: no, it is not used anymore, I forgot to remove that.

ecdsa:
Thanks for the feedback. I made the decorator mandatory, I think that is better indeed.
I also use a `dict` to register event listeners, so that we do not call `getattr` on unregistered methods.

accumulator:
>     1. We could add a `@event_listener` decorator to all event listeners, that does nothing but asserts their name starts with `on_event_`, for pedagogic purposes.
> 
>     2. Alternatively, we could allow arbitrary names, and add a decorator that takes the event's name as parameter.

As it is currently implemented in this draft, the event name is implied in the function name (nitpick: this isn't described as such in the above two options). While this is probably almost always fine, there might be a corner-case where this is inconvenient, e.g. when subclassing a class that already registers an event handler, which the subclass wants to override. I'm unsure what the exact behaviour is in this case. Will the superclass method be called, or the overridden method in the subclass, which then maybe doesn't have the guard attribute `_is_event_listener`.

>     3. We could try to add a filtering option to the decorator (to avoid starting methods with `if self.wallet != wallet: return`). Not sure if it can be done in a way that removes the current quadratic complexity ("everyone listens to everyone")

This could work.. But it'll probably look dirty. 
Most `QObjects` in the QML port get the (`QEWallet` wrapped) wallet instance after object instantiation, which is later than the decorator registration. But the decorator could be made 'smart' and probe the object if it can find a `self.wallet` or `self._wallet.wallet` attribute and use that as a filter. Downside is that it isn't very maintainable.

accumulator (approved):

SomberNight:
I've squashed your follow-up commit into the main one, and added a follow-up of mine.

electrum/gui/qt/channel_details.py
SomberNight: done in https://github.com/spesmilo/electrum/commit/be322f7ea7e1789fa044e857202cb2a8f24dadde

accumulator:
> So in such a case, the programmer needs to re-apply the decorator in the child class where the method is overridden. `getattr` works as expected and will find the method implemented in the child.

Ok, this is less of a pitfall than I feared, as long as super() calls are not used and the ordering of event handler registration follows the inheritance.

> Regardless, I say we can postpone (3) entirely.

Agreed.

ecdsa:
Note: I pushed the following commit, that should fix the issue described in the fixme: fe74e4b9c95cc6e83056dd630c6f824830ee57a3

Issue #4823, opened 2018-11-02
replace json on disk, with leveldb/sql/...? wallet file does not scale. partial writes


SomberNight:
json on disk does not scale to large files, as every time we need to flush, the whole file needs to be rewritten.

This is a problem for:
- channel_db (routing table for LN)
- watcher_db (also LN)
- even the wallet file itself...

ecdsa:
this is done for channel_db and watchtower_db
we will not do it for the wallet file (not until we forward payments)

SomberNight:
I looked at this again, specifically for the wallet file, to replace json. Being able to do partial writes will become necessary for large wallets, especially with lightning (and especially when we forward).

The main issue is that we currently encrypt the whole wallet file (by default). This is good for protecting metadata and also privacy reasons. This does not scale however: every time we need to make sure changes are persisted on disk, we serialize the whole wallet file json, compress it with zlib, encrypt it with AES, and write the whole thing to disk. (even with encryption disabled we write the whole json to disk every time)

We could simply replace json with sqlite/leveldb, but we could no longer simply encrypt the whole wallet file.

To be able to do partial writes, and have encryption, we either
1. encrypt individual records (e.g. keys and values if using a kv store), or
2. encrypt transparently between the filesystem layer and the db
The latter would be clearly preferred. (at the very least, there are many footguns and potential mistakes to be made with the first approach)

(note: [ElectrumSV implemented the option1 using sqlite](https://github.com/electrumsv/electrumsv/blob/3ec97a4f12be12c5427b802cc414dd8025af5d6a/electrumsv/wallet_database.py))

To encrypt transparently below the db (option2), we would probably need the db itself to know about the encryption and do it for us. (I guess it's not strictly needed as you could create a virtual filesystem and do the encryption there yourself (see e.g. TrueCrypt/VeraCrypt) but that might be even more complicated)

I have found two viable ways to do option2.

-----

Option2A: SQLCipher
---------------------
https://github.com/sqlcipher/sqlcipher
https://www.zetetic.net/sqlcipher/design/
SQLCipher is an extension to sqlite that allow encrypting the db (using e.g. AES-CBC for each page).
It looks maintained and to have a long track record.
It's a C project, like sqlite, so we would need python bindings.
There seem to be a few competing projects (but none of them seems too reassuring):
https://github.com/leapcode/pysqlcipher
https://github.com/coleifer/pysqlite3
https://github.com/coleifer/sqlcipher3

Option2B: RocksDB
[...]

PR #7867, opened 2022-06-21
storage: only read magic at initialization, not the full wallet contents

accumulator:
This allows fast querying of wallet encryption type regardless of wallet size.

UI could use this to add visual clues about wallet files (encrypted yes/no, HW wallet yes/no etc) without getting hit by long delays loading large wallet files.


ecdsa:
This is redundant with the `jsonpatch` branch. 
see also #4823

accumulator:
> This is redundant with the `jsonpatch` branch. (partial writes, new headers) see also #4823

Ah yes, I see. Much better

PR #7929, opened 2022-08-11
qt/kivy: receive tab: add wallet.ReceiveRequestHelp and refactor

SomberNight:
This PR refactors the Qt and kivy receive request displays in terms of error handling, and moves the common code into `wallet.py`.

"Error-like" help texts are split into two categories:
- errors, such as:
    - `"You must be online to receive Lightning payments."`
    - `"This request cannot be paid on-chain"`
- warnings, such as:
    - `"This address has already been used. For better privacy, do not reuse it for new payments."`

Both the kivy and the Qt GUIs handle these "error-like" scenarios now. If there is an error, the QR code (or corresponding text) is not shown, only the error. If there is a warning, the QR code is still shown, and the warning is also displayed somehow (up to GUI).

Previously the Qt GUI had complex GUI-specific logic for this, and kivy also had some primitive handling, with some duplication between them. Now the duplication is minimised and non-GUI-specific logic is moved to `wallet.py`.

-----

Examples:

<details>
 <summary>kivy warning</summary>

Capture1
</details>

<details>
 <summary>kivy error</summary>

Capture2
</details>

<details>
 <summary>qt warning</summary>

Capture3
</details>

<details>
 <summary>qt error</summary>

Capture4
Capture5
</details>

### DIFF
--- a/electrum/gui/kivy/uix/dialogs/request_dialog.py
+++ b/electrum/gui/kivy/uix/dialogs/request_dialog.py
@@ -6,6 +6,7 @@ from kivy.core.clipboard import Clipboard
 from kivy.app import App
 from kivy.clock import Clock
 from kivy.properties import NumericProperty, StringProperty
+from kivy.uix.tabbedpanel import TabbedPanel
 
 from electrum.gui.kivy.i18n import _
 from electrum.invoices import pr_tooltips, pr_color
@@ -18,6 +19,10 @@ if TYPE_CHECKING:
 Builder.load_string('''
 #:import KIVY_GUI_PATH electrum.gui.kivy.KIVY_GUI_PATH
 
+<TabbedPanelWithHiddenHeader@TabbedPanel>:
+    tab_height: "0dp"
+    tab_width: "1dp"
+
 <RequestDialog@Popup>
     id: popup
     amount_str: ''
@@ -27,6 +32,7 @@ Builder.load_string('''
     key:''
     data:''
     warning: ''
+    error_text: ''
     status_str: ''
     status_color: 1,1,1,1
     shaded: False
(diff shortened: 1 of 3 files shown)


ecdsa:
it looks good.
perhaps in Kivy we could toggle between qr code and text display, like on desktop, when the text is tapped.
(we currently dim the contrast when it is tapped, not sure if that is very useful)

ecdsa:
One minor issue is that the help text may contain "you may have that capacity if you rebalance your channels", even though the kivy interface does not offer that option.

electrum/gui/kivy/uix/dialogs/request_dialog.py
```diff
@@ -159,25 +179,47 @@ def on_mode(self, instance, x):
...
+        elif self.mode == self.MODE_LIGHTNING:
+            if help_texts.ln_is_error:
+                self.error_text = help_texts.ln_help
+            else:
+                self.data = lnaddr
+                self.warning = help_texts.ln_help
+        else:
+            raise Exception(f"unexpected {self.mode=!r}")
+        if self.warning:
+            self.warning = _('Warning') + ': ' + self.warning
```
ecdsa, line 222: Note that self.warning is a kivy property. 
The GUI will be updated everytime it is changed.
This is why the original code was setting a local variable first.

ecdsa: The same applies to `self.error_text`, btw

electrum/gui/kivy/uix/dialogs/request_dialog.py
```diff
@@ -159,25 +179,47 @@ def on_mode(self, instance, x):
             self.ids.qr.opacity = 1
         else:
             self.ids.qr.opacity = 0
```
ecdsa, line 181: changing the qr opacity is not needed anymore, I guess.

ecdsa:
merged manually after rebase

electrum/gui/kivy/uix/dialogs/request_dialog.py
SomberNight: oh right, indeed

electrum/gui/kivy/uix/dialogs/request_dialog.py
SomberNight: right. removed in follow-up https://github.com/spesmilo/electrum/commit/13570a465c0c017803f45cb4ad96f2b349ffc2bb

PR #7932, opened 2022-08-12
No set label

ecdsa:
Note that a sweep_address is still needed for channel backups; a dynamic method is used in that case.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -486,7 +486,6 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         is_mine |= any([self.is_mine(self.adb.get_txin_address(txin)) for txin in tx.inputs()])
         if not is_mine:
             return
-        self._maybe_set_tx_label_based_on_invoices(tx)
         if self.lnworker:
             self.lnworker.maybe_add_backup_from_tx(tx)
         self._update_request_statuses_touched_by_tx(tx_hash)
@@ -993,9 +992,6 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         if not invoice.is_lightning():
             if self.is_onchain_invoice_paid(invoice)[0]:
                 _logger.info("saving invoice... but it is already paid!")
-            with self.transaction_lock:
-                for txout in invoice.get_outputs():
-                    self._invoices_from_scriptpubkey_map[txout.scriptpubkey].add(key)
         self._invoices[key] = invoice
         self.save_db()
 
@@ -1039,18 +1035,8 @@ class Abstract_Wallet(ABC, Logger, EventListener):
     def export_invoices(self, path):
         write_json_file(path, list(self._invoices.values()))
 
-    def _get_relevant_invoice_keys_for_tx(self, tx: Transaction) -> Set[str]:
-        relevant_invoice_keys = set()
-        with self.transaction_lock:
-            for txout in tx.outputs():
-                for invoice_key in self._invoices_from_scriptpubkey_map.get(txout.scriptpubkey, set()):
-                    # note: the invoice might have been deleted since, so check now:
-                    if invoice_key in self._invoices:
-                        relevant_invoice_keys.add(invoice_key)
-        return relevant_invoice_keys
-
-    def get_relevant_invoices_for_tx(self, tx: Transaction) -> Sequence[Invoice]:
-        invoice_keys = self._get_relevant_invoice_keys_for_tx(tx)
+    def get_relevant_invoices_for_tx(self, tx_hash) -> Sequence[Invoice]:
+        invoice_keys = self._invoices_from_txid_map.get(tx_hash, set())
         invoices = [self.get_invoice(key) for key in invoice_keys]
         invoices = [inv for inv in invoices if inv]  # filter out None
         for inv in invoices:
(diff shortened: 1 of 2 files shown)


electrum/lnchannel.py
```diff
@@ -566,10 +551,13 @@ def is_frozen_for_receiving(self) -> bool:
         return False
 
     def is_static_remotekey_enabled(self) -> bool:
-        # Return False so that self.sweep_address will return self._fallback_sweep_address
+        return False
```
SomberNight, line 554: why is this False?
Looks like the old comment might not apply anymore.

ecdsa: indeed, but it does not hurt. 
that way the asserts will fail if executed on a backup.

electrum/wallet.py
```diff
@@ -1107,19 +1099,12 @@ def _is_onchain_invoice_paid(self, invoice: Invoice) -> Tuple[bool, Optional[int
 
     def is_onchain_invoice_paid(self, invoice: Invoice) -> Tuple[bool, Optional[int]]:
         is_paid, conf_needed, relevant_txs = self._is_onchain_invoice_paid(invoice)
+        if is_paid:
+            key = self.get_key_for_outgoing_invoice(invoice)
+            for txid in relevant_txs:
+                self._invoices_from_txid_map[txid].add(key)
```
SomberNight, line 1105: I don't think it's nice to add a side-effect to this method.
I find it difficult to find the code path where this would get called that we actually rely on, for the side-effect. Is it somewhere in the GUI where the invoice status will get checked? We should not rely on that...

electrum/lnchannel.py
SomberNight: I would much prefer explicit `assert not self.is_backup()` lines (either in addition to any existing assert or instead of them, depending on what makes sense) wherever that is relevant :)

> indeed, but it does not hurt.

I think it is confusing. If we keep it as False, at the least there should be a code comment here to explain.

SomberNight: Also, to be clear, we don't want to support backups of legacy channels (that have static_remotekey disabled), right? -- just to clarify whether that's related to what is returned here

electrum/lnchannel.py
```diff
@@ -763,13 +750,22 @@ def is_static_remotekey_enabled(self) -> bool:
         channel_type = ChannelType(self.storage.get('channel_type'))
         return bool(channel_type & ChannelType.OPTION_STATIC_REMOTEKEY)
 
+    @property
+    def sweep_address(self) -> str:
+        # TODO: in case of unilateral close with pending HTLCs, this address will be reused
+        addr = None
+        assert self.is_static_remotekey_enabled()
```
SomberNight, line 757: Re the `"remove support for channels without static remote pubkey"` commit in general, I think we might need to think about this a bit more.

The reason we want to do it, is to simplify the code, and the justification is that most likely hardly anyone has wallets with legacy channels in it (as these were never in a released version). However, what should the behaviour be if someone does have such a channel?
Should we maybe put an assert in `LNWallet.__init__` that all channels are `is_static_remotekey_enabled()`, so that at least there is clear and well-defined behaviour?
(although ofc then the wallet cannot be opened even just to delete those channels)

In particular, I do have a mainnet wallet that has a few legacy channels saved in it -- all of them are already redeemed (deeply-closed). When I open that wallet with this PR, there is no user-visible error in the GUI, but the logs are spammed with AssertionErrors, and in fact I expect lnwatcher to be broken even for the non-legacy `OPEN` channels...

<details>
 <summary>tracebacks from log</summary>

```
E/W | lnwatcher.LNWalletWatcher.[default_wallet-LNW] | Exception in do_breach_remedy: AssertionError()
  File "...\electrum\electrum\util.py", line 1235, in wrapper
  File "...\electrum\electrum\lnwatcher.py", line 443, in do_breach_remedy
    sweep_info_dict = chan.sweep_ctx(closing_tx)
  File "...\electrum\electrum\lnchannel.py", line 264, in sweep_ctx
    our_sweep_info = self.create_sweeptxs_for_our_ctx(ctx)
  File "...\electrum\electrum\lnchannel.py", line 253, in create_sweeptxs_for_our_ctx
    return create_sweeptxs_for_our_ctx(chan=self, ctx=ctx, sweep_address=self.sweep_address)
  File "...\electrum\electrum\lnchannel.py", line 757, in sweep_address
    assert self.is_static_remotekey_enabled()
AssertionError
```
```
E/W | lnwatcher.LNWalletWatcher.[default_wallet-LNW] | Exception in trigger_callbacks: AssertionError()
  File "...\electrum\electrum\util.py", line 1235, in wrapper
  File "...\electrum\electrum\lnwatcher.py", line 208, in trigger_callbacks
    await callback()
  File "...\electrum\electrum\lnwatcher.py", line 225, in check_onchain_situation
    keep_watching = await self.do_breach_remedy(funding_outpoint, closing_tx, spenders)
  File "...\electrum\electrum\util.py", line 1235, in wrapper
  File "...\electrum\electrum\lnwatcher.py", line 443, in do_breach_remedy
    sweep_info_dict = chan.sweep_ctx(closing_tx)
  File "...\electrum\electrum\lnchannel.py", line 264, in sweep_ctx
    our_sweep_info = self.create_sweeptxs_for_our_ctx(ctx)
  File "...\electrum\electrum\lnchannel.py", line 253, in create_sweeptxs_for_our_ctx
    return create_sweeptxs_for_our_ctx(chan=self, ctx=ctx, sweep_address=self.sweep_address)
  File "...\electrum\electrum\lnchannel.py", line 757, in sweep_address
    assert self.is_static_remotekey_enabled()
AssertionError
```
</details>

electrum/wallet.py
ecdsa: Indeed. I guess one could try to detect invoice payment in `on_event_adb_added_tx`, similar to `_update_request_statuses_touched_by_tx`

SomberNight: sounds reasonable

electrum/lnchannel.py
SomberNight: I've added a commit https://github.com/spesmilo/electrum/commit/2175bb6df09938a0d4f62058ad7fe823a1684ffb to the branch that contains a wallet DB upgrade that enforces there are no "legacy" type channels.

electrum/lnchannel.py
ecdsa: since that method is only called for channels, not channel backups, I think it is better to remove it.
ff3c999fcb7befb5c27a0b18d3dfa5fed1b5a18b

ecdsa:
I pushed the commits related to `static_remotepubkey` into master and rebased

electrum/wallet.py
```diff
@@ -1061,16 +1053,25 @@ def _init_requests_rhash_index(self):
...
 
     def _prepare_onchain_invoice_paid_detection(self):
-        # scriptpubkey -> list(invoice_keys)
-        self._invoices_from_scriptpubkey_map = defaultdict(set)  # type: Dict[bytes, Set[str]]
+        self._invoices_from_txid_map = defaultdict(set)  # type: Dict[str, Set[str]]
+        self._invoices_to_txid_map = defaultdict(set)  # type: Dict[str, Tuple[bool, Optional[int], Sequence[str]]]
+        self._detect_paid_invoices()
+
+    def _detect_paid_invoices(self):
         for invoice_key, invoice in self._invoices.items():
```
SomberNight, line 1061: This iterates over all invoices, and is called in `add_transaction`. This is a performance regression that we should avoid IMHO. Instead, I propose: https://github.com/spesmilo/electrum/pull/7942

PR #7938, opened 2022-08-15
Do not index request by receiving address

### LINKED ISSUE #7927: `add_request` command used to reuse addresses of expired payment requests but not anymore
The `add_request` command prior to version 4.3.0 used to reuse addresses that correspond to expired payment requests (but are otherwise unused). On current master however, it now returns False (unless `--force` is set, in which case it goes beyond the gap limit).

This is due to commit https://github.com/spesmilo/electrum/commit/bb2e4109422dc46652bc6579bcfc8149e3c8675

Relevant code for context:
https://github.com/spesmilo/electrum/blob/ea5c49b3ab9cf0737192f1f4e637fd55c75e9b12/electrum/commands.py#L948-L962
https://github.com/spesmilo/electrum/blob/ea5c49b3ab9cf0737192f1f4e637fd55c75e9b12/electrum/wallet.py#L2254-L2260

-----

I guess the change was made with the GUI in mind. For CLI/RPC, I think we should at least add an option to reuse expired addresses.

ecdsa:
Do not index requests by address.
Instead, use the same get_id method as for invoices.
wallet: in `create_request`, reuse use addresses of expired requests (fixes  #7927)
Separate `wallet.get_request` and `wallet.get_request_by_addr,` which takes an address

Note that expired requests in` wallet._requests_addr_to_key` get overwritten when a new one is added

### DIFF
--- a/electrum/invoices.py
+++ b/electrum/invoices.py
@@ -137,6 +137,10 @@ class Invoice(StoredObject):
         # 0 means never
         return self.exp + self.time if self.exp else 0
 
+    def has_expired(self) -> bool:
+        exp = self.get_expiration_date()
+        return bool(exp) and exp < time.time()
+
     def get_amount_msat(self) -> Union[int, str, None]:
         return self.amount_msat
 
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -987,7 +987,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         return invoice
 
     def save_invoice(self, invoice: Invoice, *, write_to_disk: bool = True) -> None:
-        key = self.get_key_for_outgoing_invoice(invoice)
+        key = invoice.get_id()
         if not invoice.is_lightning():
             if self.is_onchain_invoice_paid(invoice)[0]:
                 _logger.info("saving invoice... but it is already paid!")
@@ -1004,7 +1004,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
 
     def clear_requests(self):
         self._receive_requests.clear()
-        self._requests_addr_to_rhash.clear()
+        self._requests_addr_to_key.clear()
         self.save_db()
 
     def get_invoices(self):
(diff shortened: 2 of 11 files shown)


electrum/invoices.py
```diff
@@ -137,6 +137,10 @@ def get_expiration_date(self):
         # 0 means never
         return self.exp + self.time if self.exp else 0
 
+    def has_expired(self):
+        exp = self.get_expiration_date()
+        return exp and exp < time.time()
```
SomberNight, line 142: Let's make sure this method returns a `bool`.
```
>>> 0 and (0 < 5)
0
```

```suggestion
    def has_expired(self) -> bool:
        exp = self.get_expiration_date()
        return bool(exp) and exp < time.time()
```

electrum/wallet.py
```diff
@@ -1055,10 +1055,10 @@ def get_relevant_invoices_for_tx(self, tx: Transaction) -> Sequence[Invoice]:
         return invoices
 
     def _init_requests_rhash_index(self):
-        self._requests_addr_to_rhash = {}
-        for key, req in self._receive_requests.items():
-            if req.is_lightning() and (addr:=req.get_address()):
-                self._requests_addr_to_rhash[addr] = req.rhash
+        self._requests_addr_to_key = {}
+        for req in self._receive_requests.values():
+            if req.is_lightning() and not req.is_expired() and (addr:=req.get_address()):
```
SomberNight, line 1060:
```suggestion
            if req.is_lightning() and not req.has_expired() and (addr:=req.get_address()):
```

electrum/wallet.py
```diff
@@ -2360,12 +2360,12 @@ def get_request_status(self, key):
...
+        key = self._requests_addr_to_key.get(addr)
+        return self._receive_requests.get(key)
+
     def get_request(self, key: str) -> Optional[Invoice]:
-        if req := self._receive_requests.get(key):
-            return req
-        # try 'key' as a fallback address for lightning invoices
-        if (rhash := self._requests_addr_to_rhash.get(key)) and (req := self._receive_requests.get(rhash)):
-            return req
+        return self._receive_requests.get(key)
```
SomberNight, line 2368: note that `wallet.get_request(address)` no longer finds a request now. This is effectively an API change so we need to track down all callers and consider the implications.

For example, consider
- the `delete_request` command
    https://github.com/spesmilo/electrum/blob/21255b33af69a40f42fb9b65870f470af8a213c6/electrum/commands.py#L983-L985
    that currently takes an address (sidenote: we should support deleting by rhash), and would now be broken with this change

electrum/invoices.py
ecdsa: fixed

PR #7942, opened 2022-08-16
wallet: rm `maybe_set_tx_label_based_on_invoices`, just use `get_default_label`

SomberNight:
This is an alternative to https://github.com/spesmilo/electrum/pull/7932
The first three commits are taken from that branch.


ecdsa:
it looks good

Issue #7952, opened 2022-08-24
Exchange rates fetching stops after some time


MrNaif2018:
Opening an issue to document our discussion on IRC

After running in a long-running process for quite some time (usually it happens every day and fixes itself after restart), exchange rates no longer fetch.

Last log lines displayed are fetching exchange rates for USD, and then
`exchange_rate.CoinGecko | failed fx quotes: TimeoutError`

After that, no fetching exchange rates messages ever appear, set_currency doesn't help either, which means that main process got stopped by some exception possibly.

I think it might be because of cf3f92531ec99befd2ab1df628649498931ca78a, 2c57c78ebe77287ba9698225bf8e20880b8fee62 or aiorpcx upgrade

It is something in between a3e1d2e25e78231b30ff1103345095827a909d88 and 73ba00d7dd87af5e63ab2a30617fae1b8d36d0c4 or python version update from 3.7 to 3.9 in our case

MrNaif2018:
I have a feeling that we can patch aiorpcx to wrap the task in it's TimeoutAfter handler in this coroutine:

```
async def _ensure_cancellation(coro):
    watcher = asyncio.Event()

    task = asyncio.ensure_future(coro)
    task.add_done_callback(lambda _: watcher.set())

    try:
        await watcher.wait()
    except asyncio.CancelledError:
        task.cancel()
        await watcher.wait()
        raise

    return task.result()
```

Ref dask/distributed#6235

MrNaif2018:
OldTaskGroup might be impacting too, as it is the one which runs FxThread.run

MrNaif2018:
I have patched `update_safe` this way:

```
if not hasattr(self, "done"):
    with mock.patch.object(self, "get_rates",side_effect=TaskTimeout(5)):
        self.done = True
        self.quotes = await self.get_rates(ccy)
else:
    self.quotes = await self.get_rates(ccy)
```

Is it possible that this exception is raised?
Maybe the issue is exactly because CancelledError bases off BaseException now, so it is not caught anywhere?

MrNaif2018:
Ok, here are the full logs:
`Traceback (most recent call last):
  File "/home/electrum/.local/lib/python3.9/site-packages/el
ectrum/exchange_rate.py", line 98, in update_safe
    self.quotes = await self.get_rates(ccy)
  File "/home/electrum/.local/lib/python3.9/site-packages/el
ectrum/exchange_rate.py", line 337, in get_rates
[...]

PR #7955, opened 2022-08-24
exchange_rate: more robust spot price against temporary network issues

SomberNight:
We poll the fx rate provider every 2.5 minutes (unchanged).
Previously if there was any error during a tick, there would be no fx rate
available in the client until the next tick.
Now, instead, we keep the last rates received with a 10 minute expiry.

One potential drawback is that previously there was instant feedback
to the user when e.g. changing proxy settings, and this is no longer
the case. E.g. consider a provider that bans Tor exit nodes. If a user
enables using a Tor proxy in the network settings, the fxrate used to
disappear immediately - but now the cached rate would still be
available.

### DIFF
--- a/electrum/exchange_rate.py
+++ b/electrum/exchange_rate.py
@@ -48,12 +48,17 @@ def to_decimal(x: Union[str, float, int, Decimal]) -> Decimal:
     return Decimal(str(x))
 
 
+POLL_PERIOD_SPOT_RATE = 150  # approx. every 2.5 minutes, try to refresh spot price
+EXPIRY_SPOT_RATE = 600       # spot price becomes stale after 10 minutes
+
+
 class ExchangeBase(Logger):
 
     def __init__(self, on_quotes, on_history):
         Logger.__init__(self)
         self._history = {}  # type: Dict[str, Dict[str, str]]
-        self.quotes = {}  # type: Dict[str, Optional[Decimal]]
+        self._quotes = {}  # type: Dict[str, Optional[Decimal]]
+        self._quotes_timestamp = 0  # type: Union[int, float]
         self.on_quotes = on_quotes
         self.on_history = on_history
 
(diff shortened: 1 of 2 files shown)


accumulator:
There is one more scenario I ran in to a while back. In certain conditions (I believe a TimeoutError) the FX updates stop completely. It is particularly noticable on Android, as android freezes the app when put in the background, leading to a timeout later.

I have a branch somewhere with applicable logging around the problematic area, but can't seem to find it right now. I'll update this issue when I have some more info..

electrum/exchange_rate.py
```diff
@@ -542,15 +556,15 @@ def ccy_amount_str(self, amount, commas):
...
         while True:
-            # approx. every 2.5 minutes, refresh spot price
+            # every few minutes, refresh spot price
             try:
-                async with timeout_after(150):
+                async def trigger_wait():
                     await self._trigger.wait()
                     self._trigger.clear()
-                # we were manually triggered, so get historical rates
+                await asyncio.wait_for(trigger_wait(), 150)
```
MrNaif2018, line 564:
```suggestion
                await asyncio.wait_for(trigger_wait(), POLL_PERIOD_SPOT_RATE)
```

accumulator:
So, I've reproduced the issue by blocking the exchange service using ip(6)tables.

Flow is as follows:
1. app startup, exchange_rate._trigger is triggered from code, leading to exit from timeout block and request to exchange rate service
2. exchange rate fetch times out after ~45 seconds, 'failed fx quotes: TimeoutError()'
3. loop restarts, entering `async with timeout_after(150)` again, waiting for trigger
4. timeout never happens

It looks like the `aiorpcx.curio.timeout_after` is not reliably timing out (not sure why, might be tied to a request object somewhere?).
The patch I added uses `asyncio.wait_for` instead, and is reliably timing out.

By not using `aiorcx.curio`, are there side effects? Like dangling aoirpcx request objects/state??

MrNaif2018:
Can confirm, the fix does work. We need to ensure nothing else is broken, but it works

Why I wasn't able to reproduce it locally: we need to wait to get asyncio.exceptions.TimeoutError, then we need to wait for timeout_after to timeout and NOT set the trigger by changing currency for example. After that it does get into broken state and even the trigger doesn't help.

From IRC:

`sudo iptables -I OUTPUT -d api.coingecko.com -j DROP`

Change timeout to something like 10 seconds and see it get stuck

The fix works! :rocket:

SomberNight:
I don't like conflating these two issues. The OP is clearly distinct from https://github.com/spesmilo/electrum/issues/7952. Let's treat them independently.

accumulator:
> I don't like conflating these two issues. The OP is clearly distinct from #7952. Let's treat them independently.

agreed. I'll move the patch to a separate PR

PR #7957, opened 2022-08-26
implement timeout using asyncio.wait_for instead of aiorpcx.curio.timeout_after

### LINKED ISSUE #7952: Exchange rates fetching stops after some time

accumulator:
one potential approach to fix #7952

### DIFF
--- a/electrum/exchange_rate.py
+++ b/electrum/exchange_rate.py
@@ -429,7 +429,7 @@ class Biscoint(ExchangeBase):
 class Walltime(ExchangeBase):
 
     async def get_rates(self, ccy):
-        json = await self.get_json('s3.amazonaws.com', 
+        json = await self.get_json('s3.amazonaws.com',
                              '/data-production-walltime-info/production/dynamic/walltime-info.json')
         return {'BRL': to_decimal(json['BRL_XBT']['last_inexact'])}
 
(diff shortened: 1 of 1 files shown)


SomberNight:
I've pushed a fix in https://github.com/spesmilo/electrum/commit/12e628e2e6eb39709361b587a87e35f291200d16 .
The root cause of https://github.com/spesmilo/electrum/issues/7952 is a bug in our existing monkey-patch of `aiorpcx.timeout_after`. That monkey-patch is fixing a bug in that API, so it should stay, however with the newly added changes, the patch is really becoming difficult to understand...
Another approach ofc could be to remove usages of `aiorpcx.{timeout,ignore}_after`, given its apparent unreliableness, removing the need for the monkey-patch, except `aiorpcx` itself uses those APIs internally...

Currently we use `aiorpcx` for two things, broadly speaking:
- jsonrpc session-handling with electrum servers
- its `curio` module implements TaskGroups and timeout_after/ignore_after

As mentioned in https://github.com/spesmilo/electrum/commit/12e628e2e6eb39709361b587a87e35f291200d16, the python stdlib introduces functionality in python 3.11 that I think can deprecate `aiorpcx.curio`. That looks promising to me, and I think at latest by the time we bump the min required python to 3.11, we could drop these patches and stop using `aiorpcx.curio`. Of course that time is still like two years out...

-----

Anyway, my point is mainly that the root cause had to be fixed as the bug might manifest in other places in the code; but regardless if you think it's worthwhile we can also merge your change of moving away from timeout_after here.

accumulator:
> Anyway, my point is mainly that the root cause had to be fixed as the bug might manifest in other places in the code; but regardless if you think it's worthwhile we can also merge your change of moving away from timeout_after here.

Compliments on the thorough root-cause analysis, well-described patch, and future implementation hints!

I think your fix is better even with the monkeypatching, as it's closer tied to the framework used. My patch might fix timeouts now but lead to confusing bugs if future aiorpcx releases have different internal behaviors. I'll close this draft.

PR #7958, opened 2022-08-29
Added new server address

ahsoliman20:
Added new electrum server address to the list of hardcoded servers.


SomberNight:
Thank you for running a server!
If you keep running it, enable peer discovery, and have good uptime, when we next update the default list, the script will likely include it.

PR #7962, opened 2022-08-31
ci: add cirrus qml build on beta/* tags

accumulator:
add a tag to a commit, e.g. `beta-4.3.1`, to trigger additional qml build.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -182,6 +182,19 @@ task:
     path: "dist/*"
 
 task:
+  name: Android build (QML)
+  container:
+    dockerfile: contrib/android/Dockerfile
+    cpu: 8
+    memory: 8G
+  build_script:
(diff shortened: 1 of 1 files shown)


SomberNight:
you could try increasing the buildozer log level to 2, to actually see why the builds failed - unless you managed to figure out the cause, I personally am not sure :P

see https://github.com/spesmilo/electrum/blob/6d77253ac25369815f7b6e1a4019d405e3436a74/contrib/android/make_apk#L48-L51
the log verbosity is low on CI due to historical reasons: travis was failing builds due to logging too much. not sure if cirrus is better, IIRC we never investigated after switching away from travis.

accumulator:
Well, I hit a number of limits (OOM killer, > 1 hour build etc) and it's now failing because of another reason. For now I can fire off some tests within the ci-qml-beta branch, I'll open another PR once it's actually finishing :)

PR #7986, opened 2022-09-22
New approach to implement new wallet wizard

accumulator:
This PR separates the wizard navigation from the logic to validate and create new wallets.

The current implementation is mostly imperative.
`BaseWizard` (in `base_wizard.py`) implements a number of functions/handlers, which decide (depending on earlier choices, or passed parameters) which view to instantiate. The instantiated view gets passed a run_next parameter that determines which handler function follows the to-be shown view. Often some validator function is passed as well, which the view can use to determine if the user is allowed to proceed to the next view.

Specific GUI implementations then subclass `BaseWizard` and implement supporting functions that create views imperatively and sometimes blocking execution until a view is completed.

This imperative approach is quite incompatible with QML, which has a much stronger separation between UI and backend logic. 

This PR adds a new wizard approach, that uses simply view names (strings), linked together in a 'navmap'. The wizard maintains a dict of user entered or wizard added data, which is used at the conclusion of the wizard to create the wallet. The UI specific implementation simply associates the view names with concrete views, and provides a small wrapper to expose the navigation actions.

Also, with this approach the UI is in control, using the wizard as a context and helper, instead of the other way around (wizard (sub)class in control, controlling the UI).

The separation of the navigation also makes it much easier to extend/modify the flow through the views. The old wizard requires subclassing to override behavior (also requiring you to copy potential validation in the subclass), in this approach you simply add views and add/override navigation decisions. For example, the QML implementation for 2fa/trustedcoin defines new views and hooks them into the base NewWalletWizard. (Note: I know we want to get rid of plugins, but for now it'll have to do).

Note 1: validation is currently mostly done in the QML views and in create_storage, this could be moved to the wizard as well.
Note 2: as the wizard state is serializable, this approach should also be able to 'resume' the wizard on a different computer (for example the 2fa offline wallet creation use case). This is not implemented yet.
Note 3: this new wizard can also be used for conversational flows other than new wallet wizards, e.g. collecting the signers in a MuSig scenario


electrum/plugins/trustedcoin/qml.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # extend wizard
+        self.extend_wizard()
+
+    def extend_wizard(self):
+        wizard = self._app.daemon.newWalletWizard
+        self.logger.debug(repr(wizard))
+        views = {
+            'trustedcoin_start': {
+                'gui': '../../../../plugins/trustedcoin/qml/Disclaimer',
```
SomberNight, line 225: what is this relative to -- what is the "current directory" here?

accumulator: Current directory is the directory where the QML file exists that is referencing it. 

SO
```
/path/1.qml: Qt.resolvedUrl('path/2.qml')
/path/path/2.qml: Qt.resolvedUrl('../1.qml')
```

I know, this is ugly as f :( I hope Qt6 improves on that.
I've tried to add more import paths to the QML engine, to have more 'roots', but that doesn't work :(

accumulator: But on the bright side, if we factor out plugins into the main tree, this disappears mostly.

accumulator: Note: I also tried to define a [QQmlAbstractUrlInterceptor](https://doc.qt.io/qt-5/qqmlabstracturlinterceptor.html#details), but that didn't really work in python/PyQt.

accumulator:
If there are no big objections to this PR, I suggest we merge it, as it's getting harder to maintain separate of master.

It's fully working on QML (including complete 2FA wizard implementation and regular 2FA use while still keeping it 'isolated' as a plugin), and merging doesn't impact the existing Qt and Kivy clients, as it is kept separate from the existing (base)wizard.

the `plugins/trustedcoin/qml.py` file is notable as it's a bit different from the other gui-specific trustedcoin code. It avoids methods using the old wizard methods and re-implements some code paths itself. This can be resolved/unified later, once we decide whether we want to migrate the other clients to the new wizard OR dissolve the plugins into the base code, whatever comes first.

electrum/plugins/trustedcoin/qml.py
accumulator: Relative paths can also be resolved by packaging all `qml` files in a resource bundle. 
This however requires one to regenerate the bundle before testing each change.

ecdsa:
@accumulator I will have a look tomorrow

electrum/gui/wizard.py
```diff
@@ -0,0 +1,319 @@
...
+            'create_seed': {
+                'next': 'confirm_seed'
+            },
+            'confirm_seed': {
+                'next': 'wallet_password',
+                'last': self.last_if_single_password
+            },
+            'have_seed': {
+                'next': self.on_have_seed,
+                'last': self.last_if_single_password_and_not_bip39
```
ecdsa, line 156: I guess this item should not have a 'last' method

ecdsa:
Would it be possible to move `wizard.py` to the electrum directory (where `base_wizard.py` resides)? 
The expectation is that any module in `electrum/gui` should be a GUI

accumulator:
> Would it be possible to move `wizard.py` to the electrum directory (where `base_wizard.py` resides)? The expectation is that any module in `electrum/gui` should be a GUI

Yep, no problem.

electrum/gui/wizard.py
accumulator: have_seed is the last page, if using an electrum seed and unified password is enabled. 
If it crashes, it's a bug.

ecdsa:
```
W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/wizard/WCWalletName.qml
W | gui.qml.qeapp | wizard data is now :{}
W | gui.qml.qeapp | wizard data is now :{"wallet_name":"wallet_1"}
W | gui.qml.qeapp | next view: wallet_type
W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/wizard/WCWalletType.qml
W | gui.qml.qeapp | wizard data is now :{"wallet_name":"wallet_1","wallet_type":"standard","seed_type":"segwit"}
W | gui.qml.qeapp | next view: keystore_type
W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/wizard/WCKeystoreType.qml
W | gui.qml.qeapp | wizard data is now :{"seed_type":"segwit","wallet_name":"wallet_1","wallet_type":"standard","keystore_type":"haveseed"}
W | gui.qml.qeapp | next view: have_seed
W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/wizard/WCHaveSeed.qml
E | gui.qml.ElectrumGui | Traceback (most recent call last):
  File "/opt/electrum/electrum/gui/qml/qewizard.py", line 39, in isLast
    return self.is_last_view(self._current.view, wdata)
  File "/opt/electrum/electrum/gui/wizard.py", line 119, in is_last_view
    l = lastnav(view, wizard_data)
  File "/opt/electrum/electrum/gui/wizard.py", line 185, in last_if_single_password_and_not_bip39
    return self.last_if_single_password(view, wizard_data) and not wizard_data['seed_type'] == 'bip39'
KeyError: 'seed_type'
NoneType: None
W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/wizard/WCWalletName.qml:20: TypeError: Cannot call method 'suggestWalletName' of null
W | gui.qml.qeapp | Must construct a QGuiApplication first.


```

ecdsa: note: I already had a wallet open in the above traceback

accumulator: Hmm.. I suspect you're on an older branch. Is your HEAD e32af1230eea9b3b18dc9a28180afc1f30e83375 or 5562cf2683c75ccd30d90c87dff8bf4d043dd210 ?

ecdsa: no, I am not on an older branch.
If I checkout your commit directly, the crash does indeed not occur.
However, if I merge your branch from master, then it occurs.

accumulator:
rebased on master, small conflict resolved

electrum/gui/wizard.py
ecdsa: Have you tested this with `single_password = True`?
You will not get the above traceback if it is not set, because of lazy evaluation.

electrum/wizard.py
```diff
@@ -0,0 +1,319 @@
...
+        }
+        self._daemon = daemon
+
+    def start(self, initial_data = {}):
+        self.reset()
+        self._current = WizardViewState('wallet_name', initial_data, {})
+        return self._current
+
+    def last_if_single_password(self, view, wizard_data):
+        return False # TODO: self._daemon.config.get('single_password')
```
ecdsa, line 182: Please raise a `NotImplementedError` exception here, instead of returning `False`.
it took me a while to understand that this method must be overloaded.

accumulator: The idea was to provide a common implementation that optionally can be overridden, but as we need to combine the password check, config setting and check whether the other wallets use the same password, I guess this will always need a gui specific implementation.. Will modify to `NotimplementedError`.

electrum/gui/wizard.py
accumulator: Even with single_password = True my testcases all seemed to manage to avoid this issue :)
I was able to reproduce it now, and it's fixed in 2adf03477109f774b92e87529b06fd3b71861c8c

ecdsa: hmm, any idea why you did not have the issue? why is this commit fixing it?

accumulator: I had `single_password` config option enabled, but the two test wallets I had in my test-unified folder were unencrypted. This lead to` self._use_single_password = self.daemon.update_password_for_directory(old_password=password, new_password=password)` in qedaemon.py to disable single password behavior

ecdsa: I see. thanks

PR #7999, opened 2022-10-05
build: tighten build system to only use source pkgs in more places

SomberNight:
- the following requirements files are restricted to only include source packages:
    - `requirements.txt`
    - `requirements-build-wine`
    - `requirements-build-mac`
    - `requirements-build-appimage`
    - ~`requirements-build-android`~
- note that the first four (so excluding `-build-android`) have already been used via `pip install --no-binary :all:`
    - see e.g. https://github.com/spesmilo/electrum/blob/488600788e3bf12a2466b586bd20da7d80f76b4f/contrib/build-linux/appimage/make_appimage.sh#L113-L114
    that is, the build scripts were already only using the source dists for these dependencies.
    So there is a larger change for `-build-android` here, where this was not the case before.
        - EDIT: the `-build-android` change was split into a separate PR due to complications...
            - https://github.com/spesmilo/electrum/pull/8000

By only including hashes for the source packages of our dependencies, (and not using prebuilt binaries,) this paves the way towards auditing our dependencies in an easier way
In the future, I would like all deps to be built from source, perhaps with a small handful of whitelisted exceptions (e.g. pyqt5 and cryptography). All the `requirements-*.txt` files should include hashes for only source dists.
Then, as part of running `freeze_packages.sh`, we could audit the diff of the sources of our deps.


SomberNight:
Ok, so the Android build is failing...
I will split that commit out into a separate PR.

SomberNight:
Hmm. The android build is still failing...

SomberNight:
note: planning to have https://github.com/spesmilo/electrum/pull/8002 first

PR #8000, opened 2022-10-05
android build: "req..-build-android.txt" to only use source pkgs

SomberNight:
This was split from https://github.com/spesmilo/electrum/pull/7999 .

Only the last commit is relevant.
TODO: rebase after #7999 is merged


SomberNight:
Ok, so the Android build is failing...

python-for-android needs `pep517<0.7.0`, see
https://github.com/kivy/python-for-android/blob/f4a91b06e24ce9f137b996e5eff5f319b9a6ae80/setup.py#L25
which needs `flit`
https://github.com/pypa/pep517/blob/3d4f9d76408bedf152a66932126f4dcf20fc121d/pyproject.toml#L2

Notice that new `pep517` depends on `flit_core`, which we already depend on,
but old `pep517` depends on `flit`, which is a larger package.
https://github.com/pypa/pep517/commit/8bdb4ef2ee4e9b608486ed1c51062f12462aae4e

We could add `flit` into `base-base.txt`, but ideally p4a would just use newer pep517.
https://github.com/spesmilo/electrum/blob/685f32750ca0d83ae1617198c085285d6a4c398e/contrib/requirements/requirements-build-base.txt#L23

I've opened a PR for p4a:
https://github.com/kivy/python-for-android/pull/2680

PR #8020, opened 2022-10-17
add invoice status to invoice_status callback

accumulator:
This patch adds the invoice status to the `invoice_status` callback;
1) analog to `request_status` callback
2) all consumers of the callback do the `wallet.get_invoice_status(wallet.get_invoice(key))` dance anyway

There's some commented lines in this patch, and some logging, to makes this patch easier to review (will clean up after review).
Also, I've made one assumption here, the `PR_INFLIGHT` status in the `lnworker.pay_to_node()` route loop. As far as I can see, there's no code path that sets a different status while in this loop. If this assumption is incorrrect, we can retrieve the current status from `self.get_invoice_status()` at this point.

### DIFF
--- a/electrum/gui/qml/qewallet.py
+++ b/electrum/gui/qml/qewallet.py
@@ -160,14 +160,7 @@ class QEWallet(AuthMixin, QObject, QtEventListener):
     @event_listener
     def on_event_invoice_status(self, wallet, key, status):
         if wallet == self.wallet:
-            self._logger.debug('invoice status update for key %s' % key)
-            # FIXME event doesn't pass the new status, so we need to retrieve
-            # invoice = self.wallet.get_invoice(key)
-            # if invoice:
-            #     status = self.wallet.get_invoice_status(invoice)
-            #     self.invoiceStatusChanged.emit(key, status)
-            # else:
-            #     self._logger.debug(f'No invoice found for key {key}')
+            self._logger.debug(f'invoice status update for key {key} to {status}')
             self.invoiceStatusChanged.emit(key, status)
 
     @qt_event_listener
--- a/electrum/gui/qml/qewallet.py
+++ b/electrum/gui/qml/qewallet.py
@@ -158,16 +158,17 @@ class QEWallet(AuthMixin, QObject, QtEventListener):
                 self.historyModel.init_model()
 
     @event_listener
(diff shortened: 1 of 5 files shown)


electrum/lnworker.py
```diff
@@ -1268,7 +1268,9 @@ async def pay_to_node(
                         trampoline_onion=trampoline_onion,
                         trampoline_fee_level=self.trampoline_fee_level,
                         trampoline_route=trampoline_route)
-                util.trigger_callback('invoice_status', self.wallet, payment_hash.hex())
+                util.trigger_callback('invoice_status', self.wallet, payment_hash.hex(), PR_INFLIGHT)
+                # s = self.wallet.get_invoice_status(self.wallet.get_invoice(payment_hash.hex()))
+                # self.logger.info(f"invoice status triggered (1) for key {payment_hash.hex()} and status {s}")
```
SomberNight, line 1273: self-note: it's much cleaner to trigger the `'invoice_status'` event in `set_invoice_status` when it actually changes (as-is already done). The reason it is *also* triggered here is to update the attempt-counter and progress for a lightning payment in the GUI.

maybe we should add a comment to that effect here, e.g. "to let GUI update payment progress (e.g. attempt counter)"

electrum/gui/qml/qewallet.py
```diff
@@ -158,16 +158,17 @@ def on_event_request_status(self, wallet, key, status):
                 self.historyModel.init_model()
 
     @event_listener
-    def on_event_invoice_status(self, wallet, key):
+    def on_event_invoice_status(self, wallet, key, status):
```
SomberNight, line 161: `qeinvoicelistmodel.py` also needs updating (?)
https://github.com/spesmilo/electrum/blob/eb00012c951e10855841d9e9daa19125cb3c04a2/electrum/gui/qml/qeinvoicelistmodel.py#L143

SomberNight (commented):
looks good (assuming commented-out old code is removed)

electrum/gui/qml/qewallet.py
accumulator: Yes, the listener in `qeinvoicemodel.py` was added after I made this PR so it needs to be included here

accumulator:
reworked for remarks, rebased on master.

SomberNight:
thanks!

PR #8024, opened 2022-10-19
can_have_lightning should return False for watch-only wallets

accumulator:


accumulator:
Thanks for clarifying.

PR #8034, opened 2022-10-24
make it easier to troubleshoot config parse errors

ln2max:
Change the logging message displayed when the config file can't be
parsed: include the underlying exception text, so that a user who is
attepting to edit the config manually can find and fix any errors.


accumulator:
thanks

PR #8035, opened 2022-10-24
document disabling utxo_freeze_threshold

ln2max:
Add documentation to the docstring of the appropriate function
explaining how to disable the #6960 test for dusting transactions

Why this is a problem: when testing new wallet configurations it is helpful to
make some very low value test transactions to ensure everything works (without
losing too much money if something is not correctly set up and the private key
is not recoverable).

In this case the test transaction may fall below the threshold.

TODO:
1. it may make sense to document passing this option on the command
line as well, since editing the config file by hand is a recipe for
trouble.
Unfortunately this is not as simple as doing:
```
electrum --unconf_utxo_freeze_threshold=0
```
and I can't find how electrum takes config arguments on the command
line. Maybe someone else knows?

2. this is pretty deep in the codebase. Most users who run into this behaviour
and find it to be erroneous may not read this far into the code. An alternate
way to document it would be to include a note about the freezing behaviour in
the error that pops up, if frozen funds are the reason a transaction cannot be
sent.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -1715,6 +1715,13 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         In particular, this test triggers for large "dusting transactions"
         that are used for advertising purposes by some entities.
         see #6960
+        To disable this test, set
+        ```
+            "unconf_utxo_freeze_threshold": 0
+        ```
+        in ~/.electrum/config. Make sure to back up your config file first as
+        JSON syntax is somewhat difficult to get right, and the config file
+        will be overwritten with defaults if you get it wrong.
         """
         # confirmed UTXOs are fine; check this first for performance:
         block_height = utxo.block_height


SomberNight:
> it may make sense to document passing this option on the command line as well, since editing the config file by hand is a recipe for trouble.
> Unfortunately this is not as simple as doing:
> ```
> electrum --unconf_utxo_freeze_threshold=0
> ```
> and I can't find how electrum takes config arguments on the command line. Maybe someone else knows?

There are the `setconfig`/`getconfig` commands.
```
$ ./run_electrum -o getconfig unconf_utxo_freeze_threshold
$ ./run_electrum -o setconfig unconf_utxo_freeze_threshold 1234
true
$ ./run_electrum -o getconfig unconf_utxo_freeze_threshold
1234
$ ./run_electrum -o setconfig unconf_utxo_freeze_threshold null
true
$ ./run_electrum -o getconfig unconf_utxo_freeze_threshold
$
```
Users should not directly edit the config file. Not even power users IMHO.
I get it that a developer might, and that's fine, but this is not behaviour we want to encourage. Like you say, the syntax is non-trivial to get right.

> An alternate way to document it would be to include a note about the freezing behaviour in
the error that pops up, if frozen funds are the reason a transaction cannot be sent.

The Qt GUI already has this.
If you try to send more than available, the error message e.g. says `Not enough funds (0.041 mBTC (0.80 EUR) are frozen)`.
If you click "max" and everything is frozen, you get a similar error.
https://github.com/spesmilo/electrum/blob/fffbccff3b74a937306fdf593dbbbc9b884393f0/electrum/gui/qt/send_tab.py#L278
If you click "max" and some (but not all) coins are frozen, there is a forced tooltip:
https://github.com/spesmilo/electrum/blob/fffbccff3b74a937306fdf593dbbbc9b884393f0/electrum/gui/qt/send_tab.py#L216

> Most users who run into this behaviour and find it to be erroneous may not read this far into the code. 

If using the Qt GUI, as above, there are already hints that the utxo is frozen.
If the user then goes to the Coins tab, they will see that's indeed the case and can manually unfreeze the utxo.

This config variable is not documented and I agree the behaviour might be unintuitive to users, but I think it would be overkill to expose it directly in the settings.

electrum/wallet.py
```diff
@@ -1715,6 +1715,13 @@ def _is_coin_small_and_unconfirmed(self, utxo: PartialTxInput) -> bool:
         In particular, this test triggers for large "dusting transactions"
         that are used for advertising purposes by some entities.
         see #6960
+        To disable this test, set
+        ```
+            "unconf_utxo_freeze_threshold": 0
+        ```
+        in ~/.electrum/config. Make sure to back up your config file first as
+        JSON syntax is somewhat difficult to get right, and the config file
+        will be overwritten with defaults if you get it wrong.
```
SomberNight, line 1724: The advice text here applies to editing the config in general. Should we maybe document `setconfig` and `getconfig` better instead?

ln2max:
On Mon, Oct 24, 2022 at 10:39:56AM -0700, ghost43 wrote:
>There are the `setconfig`/`getconfig` commands.
>```
>$ ./run_electrum -o getconfig unconf_utxo_freeze_threshold
>$ ./run_electrum -o setconfig unconf_utxo_freeze_threshold 1234
>true
>$ ./run_electrum -o getconfig unconf_utxo_freeze_threshold
>1234
>$ ./run_electrum -o setconfig unconf_utxo_freeze_threshold null
>true
>$ ./run_electrum -o getconfig unconf_utxo_freeze_threshold
>$
>```

Oh, that's a much better way to do it!


>> Most users who run into this behaviour and find it to be erroneous may not read this far into the code.
>
>If using the Qt GUI, as above, there are already hints that the utxo is frozen.
>If the user then goes to the Coins tab, they will see that's indeed the case and can manually unfreeze the utxo.

Ah, when I ran into the error (with a new multisig wallet setup I was
creating) the coins were **not** frozen in the Addresses tab (which is
where I went to look to un-freeze them manually).

Since I could not manually un-freeze the UTXO (via the Addresses tab) I
ended up finding the config option (in the source code) and manually
editing the config file.

So I would suggest:
1. having the freeze also show in the Addresses tab, not just in the
Coins tab (which is not shown by default, and I think is not an
intuitive place to look for un-freezing UTXOs).

2. documenting the setconfig/getconfig commands.

I have updated my branch to reflect these changes.

SomberNight:
> So I would suggest:
> 1. having the freeze also show in the Addresses tab, not just in the Coins tab (which is not shown by default, and I think is not an intuitive place to look for un-freezing UTXOs).

In the Coins tab, there is one line per utxo, each with one column for outpoint and one column for address. Users can freeze addresses and coins ~independently. If a coin being frozen is reflected by changing the background colour of the outpoint column, and an address being frozen has different bgcolor in the address column.
In the Addresses tab, it is not clear how to display "some coins are frozen" in the Addresses tab. That functionality is not there atm. I've sketched it out in https://github.com/spesmilo/electrum/pull/8040 but some more interface work would be needed (starting with figuring out how it should look like in the first place).

> 2. documenting the setconfig/getconfig commands.
> I have updated my branch to reflect these changes.

I don't think it's a good approach to teach the reader about getconfig/setconfig in the docstring of every method that might use a config key. There should be a single (more central) place for it, e.g. it could go into `simple_config.py`.

ln2max:
On Thu, Oct 27, 2022 at 12:45:47PM -0700, ghost43 wrote:
>> So I would suggest:
>> 1. having the freeze also show in the Addresses tab, not just in the Coins tab (which is not shown by default, and I think is not an intuitive place to look for un-freezing UTXOs).
>
>In the Coins tab, there is one line per utxo, each with one column for outpoint and one column for address. Users can freeze addresses and coins ~independently. If a coin being frozen is reflected by changing the background colour of the outpoint column, and an address being frozen has different bgcolor in the address column.

I see.

>In the Addresses tab, it is not clear how to display "some coins are frozen" in the Addresses tab. That functionality is not there atm. I've sketched it out in https://github.com/spesmilo/electrum/pull/8040 but some more interface work would be needed (starting with figuring out how it should look like in the first place).

Checkered blue/white? Pale blue?

>> 2. documenting the setconfig/getconfig commands.
>> I have updated my branch to reflect these changes.
>
>I don't think it's a good approach to teach the reader about getconfig/setconfig in the docstring of every method that might use a config key. There should be a single (more central) place for it, e.g. it could go into `simple_config.py`.

Also documented it in `simple_config.py`, see latest commit. Ideally it
would also be in the manpage.

ln2max:
On Thu, Oct 27, 2022 at 12:45:47PM -0700, ghost43 wrote:
>In the Addresses tab, it is not clear how to display "some coins are frozen" in the Addresses tab. That functionality is not there atm. I've sketched it out in https://github.com/spesmilo/electrum/pull/8040 but some more interface work would be needed (starting with figuring out how it should look like in the first place).

Worth adding: it may be worth clarifying that *coins* (in the Coins tab)
can be frozen even when *addresses* are not frozen, in the pop-up
message which warns that there are not sufficient funds, and coins are
frozen. 

By warning the user that they should check the (normally not visible)
Coins tab because coins are frozen, we would eliminate some of the
confusion which drives the need for changing the appearance of addresses
in the Addresses tab.

PR #8040, opened 2022-10-27
Qt AddressList: mark address if some coins are frozen

SomberNight:
If some coins are frozen in the wallet (but not addresses), the Addresses tab of the Qt GUI does not give any clues to the user about this. I think if there is an address which is not frozen itself but some related UTXOs are, the interface should show that somehow.

I don't know *how* to do it nicely though... In this PR, as before, a frozen address has a solid blue background colour; and now if there are some frozen coins, the address has a background in a different style.

pic1

related: https://github.com/spesmilo/electrum/pull/8035#issuecomment-1293535818

-----

I've done some benchmarking on large wallets, and this makes `AddressList.refresh_all()` around 2x slower.
(due to calculating the utxos for an address: `wallet.adb.get_addr_utxo(address)`)
Regardless, I think the slowdown is acceptable.

### DIFF
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -26,7 +26,7 @@
 from enum import IntEnum
 
 from PyQt5.QtCore import Qt, QPersistentModelIndex, QModelIndex
-from PyQt5.QtGui import QStandardItemModel, QStandardItem, QFont
+from PyQt5.QtGui import QStandardItemModel, QStandardItem, QFont, QBrush
 from PyQt5.QtWidgets import QAbstractItemView, QComboBox, QLabel, QMenu
 
 from electrum.i18n import _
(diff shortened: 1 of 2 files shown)


SomberNight:
> In this PR, as before, a frozen address has a solid blue background colour; and now if there are some frozen coins, the address has a background in a different style.

Ahh, this looks really bad when using the dark style. The states are difficult to distinguish.
Actually, in the dark style, I don't particularly like any of the [brush styles](https://doc.qt.io/qt-5/qt.html#BrushStyle-enum) for the background.
I would rather not use a different colour than blue either, as blue is intuitive for frozen (or at least users already learned it).

Maybe we need a different approach; e.g. adding an icon in a new column.

ecdsa:
what is the use case for freezing a coin without freezing the address? 
how about we remove that functionality?

ecdsa:
my point is that this is redundant with the freeze threshold.
the cause of the confusion is that we have both.

SomberNight:
Even if we removed manual utxo freezing from the gui, but kept automatic utxo freezing (for the current use case of unconfirmed dust, and potential future use cases), I think it would be still useful to give a hint to the user in the Addresses tab.

But I would like to keep manual utxo freezing, I think it's a nice power-user feature, which most users likely never interact with but it should not affect them negatively either.

Issue #7897, opened 2022-07-20
Native support for new Ledger app


bigspider:
The upcoming release of the Ledger Bitcoin app will come with miniscript support, but will break compatibility with the legacy protocol (used before app version 2.0.0, and now deprecated). Currently, the new app still basically contains the old app inside, but that's not sustainable in the long term.

**We plan to drop legacy support in the main app by October 2022.**

There will be a way to install the legacy app, so that nobody is locked out − but I would like to encourage the integration of the new API, which brings numerous benefits. It's easier, cleaner and psbt- and descriptor-native; it already supports taproot (and work for miniscript on segwitv0 and on taproot will only be brought to the new API); moreover, it's actively developed.

The good news:
- we have a [python library](https://github.com/LedgerHQ/app-bitcoin-new/tree/experimental/bitcoin_client) that should be quite easy to use and adapt. It was already the base of the [HWI integration](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/ledger.py).
- the library is backwards-compatible with older versions of the app; that is, for simple transactions, it can proxy to the legacy library if the installed app version is legacy.

The (maybe) bad news:
- The new app has a [register-and-use flow](https://blog.ledger.com/bitcoin-2/#:~:text=Using%20multi%2Dsignature%20wallet%20policies) for multisignature wallets; so a _proper_ integrations will require some changes to the client side logic as well. The benefit is a much safer and cleaner UX for multisig wallets (the app can now correctly recognize change addresses and simplify the signing flow accordingly)

Currently, the HWI integration for multisignature is partial (it does not store the required multisig information, forcing the user to re-register it every time); hopefully that can be improved in an upcoming release.

There is also [this PoC for multisig](https://gist.github.com/mflaxman/386990be297e9a595936e8142e815e52) by @mflaxman that was developed independently using the library, so it might be useful.

If you are able to compile and install the app to use with speculos or sideload on a device, you can compile from the current `develop` (default) branch in the [app-bitcoin-new](https://github.com/LedgerHQ/app-bitcoin-new) repository with
```
$ make DEBUG=0 COIN=bitcoin_lite
```
for mainnet, or
```
$ make DEBUG=0 COIN=bitcoin_testnet_lite
```
for testnet.

(The `_lite` targets compile the app without legacy support).

Please feel free to reach out if I can help in any way. As the historical and one of the most commonly used multisig wallets, I would really love to see a good integration within Electrum.

bigspider:
[...]

PR #8041, opened 2022-10-28
Modify Ledger plugin to support the new bitcoin app v2.1.0

bigspider:
The upcoming version 2.1.0 of the Leger bitcoin app removes compatibility with the legacy protocol; see #7897 for context and details.

This PR adds support for the new protocol, while trying to cause minimal disruption for people still running with the legacy protocol.

The plugin now depends on two libraries: `btchip-python` for the legacy protocol, and `ledger_bitcoin` for the new protocol. The old plugin's implementation is still in the `Ledger_Client_Legacy` class, where I tried to keep changes minimal. The new protocol is implemented by `Ledger_Client_New`; the `Ledger_Client` metaclass will choose the appropriate implementation based on the app version.

For the new client, this PR implements:
- support for all common types of transactions; for non-standard paths (which are not allowed by default from the app), the client will register the policy on the device, then is expected to work as usual.
- for multisignature, the new app requires registering the policy on the device. While the information on the policy being signed is available to Electrum, the approach in `Ledger_Client_New` is to deduce the wallet policy from the inputs, based on the approach used in bitcoin-core/HWI.

Some caveats that could be addressed in future PRs:
- A better UX would be obtained by persisting the registered policies within Electrum's storage; currently, registration has to be repeated every time a multisignature wallet is used. For multisignature (and future miniscript-based scripts), it might be wise to allow the user to choose a unique name for the policy, in order to allow to easily distinguish among multiple registered policies.
- The device has to be disconnected and reconnected in order to pair again when a different wallet is used. I couldn't figure out the reason − any suggesstionon how to debug it would be appreciated!
- The library version that is checked is now referring to `ledger_bitcoin`, so this might break for extremely old version of the `btchip-python` library.

I'm not sure I updated correctly the `requirements-hw.txt` file, and I'm not sure how to fix the .

For anyone who wants to test, the new app's release candidate is available in Ledger Live by switching the Manager Provider manager to `4`: (Settings --> Experimental features --> Manager provider, enable and set to `4`). Make sure to switch back to 1 when you're done.

Closes: #7897


electrum/plugins/ledger/ledger.py
```diff
@@ -1,37 +1,42 @@
...
     import hid
-    from btchip.btchipComm import HIDDongleHIDAPI, DongleWait
+    from btchip.btchipComm import HIDDongleHIDAPI
     from btchip.btchip import btchip
-    from btchip.btchipUtils import compress_public_key,format_transaction, get_regular_input_script, get_p2sh_input_script
+    from btchip.btchipUtils import compress_public_key
     from btchip.bitcoinTransaction import bitcoinTransaction
-    from btchip.btchipFirmwareWizard import checkFirmware, updateFirmware
+    from btchip.btchipFirmwareWizard import checkFirmware
     from btchip.btchipException import BTChipException
```
SomberNight, line 40: why can't we just use `ledger_bitcoin.btchip` from the new package? is the old package still needed as a dependency?

contrib/requirements/requirements-hw.txt
```diff
@@ -6,7 +6,8 @@ hidapi<0.11
 trezor[hidapi]>=0.13.0,<0.14
 safet>=0.1.5
 keepkey>=6.3.1
-btchip-python>=0.1.32
+btchip-python>=0.1.32 # used for legacy Ledger Bitcoin app up to versions 2.0.* or below
+ledger-bitcoin=0.1.1 # used for new Ledger Bitcoin app >= v2.1.0 and above
```
SomberNight, line 10: CI is failing because this is not valid pip/setuptools syntax, you need `==`
```suggestion
btchip-python>=0.1.32  # used for legacy Ledger Bitcoin app up to versions 2.0.* or below
ledger-bitcoin==0.1.1  # used for new Ledger Bitcoin app >= v2.1.0 and above
```
but I would like to have a range specified here, e.g. 
`ledger-bitcoin>=0.1.1` or `ledger-bitcoin>=0.1.1,<0.2`

bigspider:
Rebased, minor fixed added to the base commit, and addressed the review comments in two separate commits.

electrum/plugins/ledger/ledger.py
```diff
@@ -1,37 +1,42 @@
...
 
-from electrum import ecc
-from electrum import bip32
-from electrum.crypto import hash_160
-from electrum.bitcoin import int_to_hex, var_int, is_segwit_script_type, is_b58_address
+import ledger_bitcoin
+from ledger_bitcoin import WalletPolicy, MultisigWallet, AddressType, Chain
+from ledger_bitcoin.exception.errors import DenyError, NotSupportedError, SecurityStatusNotSatisfiedError
+from ledger_bitcoin.key import KeyOriginInfo
+from ledgercomm.interfaces.hid_device import HID
```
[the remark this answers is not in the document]
bigspider: Fixed in 4ba45d575fce71fc2f43356b73c676c9500d2196.

electrum/plugins/ledger/ledger.py
```diff
@@ -540,93 +762,557 @@ def sign_transaction(self, tx, password):
...
-        except ImportError:
-            raise
-        except:
-            version = "unknown"
-        if BTCHIP:
-            return version
-        else:
-            raise LibraryFoundButUnusable(library_version=version)
+        # Older versions of the device would rather require the btchip library
+        return ledger_bitcoin.__version__
```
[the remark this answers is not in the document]
bigspider: Fixed in 4ba45d575fce71fc2f43356b73c676c9500d2196.

electrum/plugins/ledger/ledger.py
bigspider: Indeed, that's a good idea. Done in 4ba45d575fce71fc2f43356b73c676c9500d2196.
I removed the call to `checkFirmware` and the corresponding import, as it's not present in `ledger_bitcoin.btchip`.

bigspider:
Sorry for not addressing the failing "Tox Python 3.8" from the CI − I'm not sure how to debug it

electrum/plugins/ledger/ledger.py
```diff
@@ -540,75 +764,548 @@ def sign_transaction(self, tx, password):
...
+        # Non standard policy, so give it a name
+        return WalletPolicy(name, descriptor_template, [key_info])
+
+    def password_dialog(self, msg=None):
+        response = self.handler.get_word(msg)
+        if response is None:
+            return False, None, None
+        return True, response, response
+
+    def _register_policy_if_needed(self, wallet_policy: 'WalletPolicy') -> tuple[bytes, bytes]:
```
SomberNight, line 985: `TypeError: 'type' object is not subscriptable`
[This syntax](https://peps.python.org/pep-0585/) was only added in python 3.9.
(Our minimum atm is 3.8)

bigspider: Indeed, thanks!
Fixed and rebased, crossing fingers for the CI run... :crossed_fingers:

SomberNight (commented):
I am trying to figure out if it's worth the maintenance burden to have both `Ledger_Client_Legacy` and `Ledger_Client_New`. What exactly would it mean if we dropped the old code?
`Ledger_Client_Legacy` supports HW.1, so we would drop support for that. What else?
`nano s`, `nano s plus`, `nano x` would only be supported if the user updated the `bitcoin app` to 2.1, right?

SomberNight:
>  Do you have any advice on how to debug the disconnession issue? Basically I have to unplug and replug the device every time I open a different wallet.

Does this still apply? I could not reproduce.
Note that only one open wallet can have an active connection with the device, if you open another wallet window that tries to connect to the device, try closing the previous window that has the connection.

-----

Slightly off-topic but now that I'm testing with bitcoin app 2.1.0, I am getting warnings re unusual derivation path when I create a new encrypted wallet file, or when decrypting an existing wallet. This was not the case with previous bitcoin app versions.
Could you please whitelist the bip32 path `m/4541509h/1112098098h`?
https://github.com/spesmilo/electrum/blob/master/electrum/storage.py#L41-L44
I believe that path was already whitelisted in prior versions. (see e.g. https://github.com/spesmilo/electrum/pull/6512#issuecomment-678759266)
It's also whitelisted in e.g. the bitbox02 fw btw: https://github.com/spesmilo/electrum/pull/5993#discussion_r385775685

bigspider:
> Does this still apply? I could not reproduce. Note that only one open wallet can have an active connection with the device, if you open another wallet window that tries to connect to the device, try closing the previous window that has the connection.

Ah, I guess it's normal then; I was just launching the 'New wallet' wizard and leaving the old window in background; it works if I close it.
Suggestion for UX improvement: maybe the wizard could detect it and invite the user to close the window?

> Slightly off-topic but now that I'm testing with bitcoin app 2.1.0, I am getting warnings re unusual derivation path when I create a new encrypted wallet file, or when decrypting an existing wallet. This was not the case with previous bitcoin app versions. Could you please whitelist the bip32 path `m/4541509h/1112098098h`?

The new app is a lot more restrictive in the key usage, indeed. Added an exception for electrum [here](https://github.com/LedgerHQ/app-bitcoin-new/commit/313188f732a6f9743fa9de32d488f4b3af33f45f), will be in the final version deployed.

electrum/plugins/ledger/ledger.py
```diff
@@ -540,75 +764,548 @@ def sign_transaction(self, tx, password):
...
-    def show_address(self, sequence, txin_type):
-        client_ledger = self.get_client_dongle_object()
-        address_path = self.get_derivation_prefix()[2:] + "/%d/%d"%sequence
-        self.handler.show_message(_("Showing address ..."))
-        segwit = is_segwit_script_type(txin_type)
-        segwitNative = txin_type == 'p2wpkh'
+    def sign_message(
+            self,
+            address_path: str,
+            sequence: 'AddressIndexGeneric',
```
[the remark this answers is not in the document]
bigspider: I was sure I re-tested signing after the latest changes :-/
Fixed in 80e214d4650dfacdad9043ba1bbe6c5a63844639.

electrum/plugins/ledger/ledger.py
```diff
@@ -540,75 +764,548 @@ def sign_transaction(self, tx, password):
...
     }
+
     SUPPORTED_XTYPES = ('standard', 'p2wpkh-p2sh', 'p2wpkh', 'p2wsh-p2sh', 'p2wsh')
 
     def __init__(self, parent, config, name):
         self.segwit = config.get("segwit")
         HW_PluginBase.__init__(self, parent, config, name)
         self.libraries_available = self.check_libraries_available()
         if not self.libraries_available:
+            print("Library unavailable")
```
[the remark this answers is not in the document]
bigspider: Not sure if there are cases where the `return` is still useful, so I replaced the `print` with the logger in 7c80779903008652103cc90885088ac85963a98a rather than removing it altogether.

bigspider:
Investigating an issue on the new app where signature seems to be requested twice to the device.

bigspider:
> Investigating an issue on the new app where signature seems to be requested twice to the device when sending a transaction.

Fixed in e47270daa50087cbf977a42f4c538996223bb4a0, the signing loop was wrongly indented.

bigspider:
I tried showing an address, signing a transaction and signing a message on both legacy and new app, as of e47270daa50087cbf977a42f4c538996223bb4a0; hopefully all the issues are fixed, now.

electrum/plugins/ledger/ledger.py
```diff
@@ -540,75 +718,547 @@ def sign_transaction(self, tx, password):
...
+
+            self.handler.show_message(_("Confirm Transaction on your Ledger device..."))
+
+            if len(wallets) == 0:
+                # Could not find a WalletPolicy to sign with
+                raise UserFacingException(_('Unable to sign this transaction'))
+
+            # For each wallet, sign
+            for __, (__, wallet, wallet_hmac) in wallets.items():
+                input_sigs = self.client.sign_psbt(psbt, wallet, wallet_hmac)
```
SomberNight, line 1154: This new API is simpler than the old one but has a significant regression: there is no feedback whatsoever to the user regarding progress. see https://github.com/spesmilo/electrum/issues/7516 and in particular https://github.com/spesmilo/electrum/commit/ec98ef5aee72d5b5e5ec119a1c176820c6bb4635
The legacy client gives feedback to the user via callbacks (shown in the UI). The new client is a blackbox.
Note that unlike e.g. trezor, ledger does not show signing progress on the device screen either.
Signing a 70 input tx takes 5+ minutes, during which there is no feedback anywhere with the new client.

---

I've just tested signing a few txs, all p2wpkh inputs and outputs (see commit linked above re phase1/phase2 distinction):
"nano s", bitcoin app 1.6.3
10 txins
- p1: 49 sec
- p2: 17 sec

"nano s", bitcoin app 1.6.3
70 txins
- p1: 328 sec
- p1: 122 sec

---

"nano s plus", bitcoin app 2.1.0rc3
10 txins
- p1: 50 sec
- p2: 57 sec

"nano s plus", bitcoin app 2.1.0rc3
70 txins
- error after 5 mins:
	```
	E | plugins.ledger.ledger | Error while signing
	  File ".../electrum/electrum/plugins/ledger/ledger.py", line 1154, in sign_transaction
		input_sigs = self.client.sign_psbt(psbt, wallet, wallet_hmac)
	  File ".../site-packages/ledger_bitcoin/client.py", line 189, in sign_psbt
		raise DeviceException(error_code=sw, ins=BitcoinInsType.SIGN_PSBT)
	ledger_bitcoin.exception.errors.UnknownDeviceError: ('0x5515', 'Error in <BitcoinInsType.SIGN_PSBT: 4> command', '')
	```
        That is, I could not sign the tx using the new client, only using legacy, as the device errored.

---

This is not a blocker for merging, but it would be good to improve on later. Either the device should show progress on-screen (as trezor), or the API could be extended so that we can pass a callback.

bigspider: > In any case, it's weird for the PIN lock to activate mid-signing :P

Agreed, but this might be tricky because of the constraints from the OS (applications being able to subvert the PIN lock might be considered a breach in the security model).

> (btw timings are: p1: 300 sec, p2: 393 sec)

There is some room for improvement, especially on NanoX/NanoS+; adding [wallet policies](https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/doc/wallet.md) added lots of new possibilities for multisig/miniscript, but focus was on making sure it still fits on Nano S rather than optimizing performance (so there's some redundant work).
Hopefully not too many people frequently send transactions with tens of inputs, yet!

Hopefully I'll find time for making a future PR to further improve the UX for multisignature (persisting the wallet registration on disk would be the main missing feature).

Thanks a lot for your time!

SomberNight:
Please see https://github.com/LedgerHQ/ledgercomm/pull/2 - this too is a regression that works on our current master.

electrum/plugins/ledger/ledger.py
SomberNight: If we have both `Ledger_Client_Legacy` and `Ledger_Client_New`, we might as well keep the HW.1 support for a bit longer. I think the convenient time to remove it is when removing the legacy protocol.

Also, I don't want to remove support for a device in a patch version, and we don't plan to do a large release atm, so keeping support lets us release this in a patch release.

I've pushed a commit to re-add the HW.1 stuff to your branch: https://github.com/bigspider/electrum/commit/9b82eb6d06f8534652849f3370ab8af7d7ca9c7f, along with a new deprecating warning shown in the UI. We can keep depending on `btchip` while the support is kept. I've tried to separate out the hw.1 specific parts more than before. 
I've also tested signing a few txs with a HW.1.

SomberNight (approved):
PR looks good.
I've pushed a clean-up commit to your branch, and also re-added the hw.1 support for the time being (see https://github.com/spesmilo/electrum/pull/8041/files#r1018459663).
If these looks ok to you, then the PR can be merged.
(but please address the [ledgercomm hid bug](https://github.com/spesmilo/electrum/pull/8041#issuecomment-1309418516) and publish a new version with it on PyPI)

SomberNight:
Thank you for the contribution!

IanMichaelHarper:
Does this mean I can now use my ledger in a secure multisig setup using electrum and avoid the pitfalls listed [here](https://benma.github.io/2020/11/05/multisig-xpubs-verification.html)?

bigspider:
> Does this mean I can now use my ledger in a secure multisig setup using electrum and avoid the pitfalls listed [here](https://benma.github.io/2020/11/05/multisig-xpubs-verification.html)?

Yes!
The only caveat right now is that the wallet registration is not persisted, therefore is repeated every time, which is not a great UX.
This PR focused on compatibility, a little more work needed for a better UX.

PR #8050, opened 2022-11-03
CI: try to cache android apk builds

SomberNight:
Trying to get the CI to build the qml apk on every commit...
Let's see the CI.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -174,39 +174,52 @@ task:
     CIRRUS_DOCKER_CONTEXT: contrib/build-wine
 
 task:
-  name: Android build (Kivy arm64)
+  name: Android build (Kivy $APK_ARCH)
   container:
     dockerfile: contrib/android/Dockerfile
     cpu: 2
     memory: 2G
+  env:
+    APK_ARCH: arm64-v8a
+  p4a_cache:
+    folders:
+      - ".buildozer/android/platform/build-$APK_ARCH/packages"
+      - ".buildozer/android/platform/build-$APK_ARCH/build"
+    fingerprint_script:
+      # note: should *at least* depend on Dockerfile and p4a_recipes/, but contrib/android/ is simplest
+      - find contrib/android/ -type f -print0 | sort -z | xargs -0 sha256sum | sha256sum
+      - echo "kivy $APK_ARCH"
+    populate_script: mkdir -p ".buildozer/android/platform/$APK_ARCH"/{packages,build}
   build_script:
-    - ./contrib/android/make_apk.sh kivy arm64-v8a debug
+    - ./contrib/android/make_apk.sh kivy "$APK_ARCH" debug
   binaries_artifacts:
     path: "dist/*"
 
 task:
-  name: Android build (QML arm64)
+  name: Android build (QML $APK_ARCH)
   container:
     dockerfile: contrib/android/Dockerfile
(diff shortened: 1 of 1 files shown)


accumulator:
Very cool! Is there a reason the armeabi-v7a arch is still locked behind the only_if conditional?

SomberNight:
As the task is asking for a beefy container with lots of cpu (and RAM but not sure if that matters) it takes quite some time to get scheduled - regardless of a cache hit. Let's see how well the CI keeps up as-is first.

Do you personally use armeabi-v7a? Is that what runs on your device with alien dalvik? I can't remember clearly.
Most typical phones sold since around 2016 can run both armeabi-v7a and arm64-v8, and starting this year there's a trend that some can only run arm64-v8. If none of us actually uses the old arch, I think there's no point in having the CI build it.

Otherwise if the CI looks to be able to keep up, we can enable it for master. It would be good to not have an extra branch just for this and have it self-contained in master :)

Alternatively, maybe we could build just the old arch and drop the new, if none of us has a device yet that can only run arm64. (I think in practice it's only the pixel7 atm)

SomberNight:
If we used newer p4a, we could build a multi-arch apk and cache it in a single container :P

SomberNight:
> Oh wait, we could already build two apks in the same container...

well, I've tried in https://github.com/spesmilo/electrum/commit/a47c1e791b833a9fd430f51466414f02425a7041
but if there is a cache-miss, the build takes more than 60 minutes, which is the default timeout:
https://cirrus-ci.com/task/6431653531222016
we could increase that (the hard max is 120 minutes) but not sure if it's a good idea :/

So my question is still whether you use the arm7 binary because otherwise it's just not worth the effort.

accumulator:
> Do you personally use armeabi-v7a? Is that what runs on your device with alien dalvik? I can't remember clearly. Most typical phones sold since around 2016 can run both armeabi-v7a and arm64-v8, and starting this year there's a trend that some can only run arm64-v8. If none of us actually uses the old arch, I think there's no point in having the CI build it.

If I remember correctly, my alien dalvik based phone only runs armeabi-v7a. I don't use builds from the CI though, so I can easily live without a 32 bit build :) Even the older android 7 phone I'm occasionally testing on is 64bit, so it should be able to use the arm64-v8 build.

> Otherwise if the CI looks to be able to keep up, we can enable it for master. It would be good to not have an extra branch just for this and have it self-contained in master :)

Yes, I agree. Now that the added cache decimates the build time it should be fine for any branch/PR.

> Alternatively, maybe we could build just the old arch and drop the new, if none of us has a device yet that can only run arm64. (I think in practice it's only the pixel7 atm)

Nah let's stick to arm64. Unless we find a tester that only wants to test on 32bit builds..

> well, I've tried in [a47c1e7](https://github.com/spesmilo/electrum/commit/a47c1e791b833a9fd430f51466414f02425a7041) but if there is a cache-miss, the build takes more than 60 minutes, which is the default timeout: https://cirrus-ci.com/task/6431653531222016 we could increase that (the hard max is 120 minutes) but not sure if it's a good idea :/

For this reason I had the 32bit and 64bit builds in separate tasks. With the matrix build definition, the 1 hour timeout now applies to the entire matrix.

> So my question is still whether you use the arm7 binary because otherwise it's just not worth the effort.

Not from the CI, so let's remove it

PR #8057, opened 2022-11-04
use debian makensis instead of windows makensis executable through wine

### LINKED ISSUE #8051: windows build: instead of running nsis.exe in wine, `apt install nsis` and run it from linux
We use NSIS to build the `electrum-*-setup.exe` distributable for Windows.
Currently we download the official release binary (a win32 `exe`) of nsis from upstream and run it inside `wine`.
Instead, we could `apt install nsis` [on debian](https://packages.debian.org/bullseye/nsis) (running [inside docker](https://github.com/spesmilo/electrum/blob/eb2ff94104baf3b333b82901a208bf93a17a5cd9/contrib/build-wine/Dockerfile#L16)) and run `makensis` from Linux.

debian is packaging `nsis`, reproducibly built. Using that would be better than trusting the binary from upstream.

https://github.com/spesmilo/electrum/blob/eb2ff94104baf3b333b82901a208bf93a17a5cd9/contrib/build-wine/prepare-wine.sh#L5
https://github.com/spesmilo/electrum/blob/eb2ff94104baf3b333b82901a208bf93a17a5cd9/contrib/build-wine/build-electrum-git.sh#L68

accumulator:
closes #8051 

executables are built in CI, but cannot test a.t.m.

### DIFF
--- a/contrib/build-wine/Dockerfile
+++ b/contrib/build-wine/Dockerfile
@@ -30,6 +30,7 @@ RUN dpkg --add-architecture i386 && \
         libtool \
         gettext \
         sudo \
+        nsis \
         && \
     rm -rf /var/lib/apt/lists/* && \
     apt-get autoremove -y && \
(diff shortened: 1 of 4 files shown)


SomberNight (approved):
Looks good! I was expecting more changes to be needed for some reason, but it starts on Win10 and seems to work.
Thank you.

PR #8058, opened 2022-11-07
trezor: optimize signing speed by not serializing transaction in trezor

prusnak:
Since Electrum is not using TxRequestSerializedType.serialized_tx we might ask the device not to serialize transactions by setting SignTx.serialize=False

The trezorlib requirement is bumped to 0.13.4, since that's the version when SignTx.serialize field appeared for the first time.

### DIFF
--- a/electrum/plugins/trezor/trezor.py
+++ b/electrum/plugins/trezor/trezor.py
@@ -363,6 +363,7 @@ class TrezorPlugin(HW_PluginBase):
                                        lock_time=tx.locktime,
                                        version=tx.version,
                                        amount_unit=self.get_trezor_amount_unit(),
+                                       serialize=False,
                                        prev_txes=prev_tx)
         signatures = [(bh2u(x) + '01') for x in signatures]
         tx.update_signatures(signatures)


contrib/requirements/requirements-hw.txt
```diff
@@ -3,7 +3,7 @@
 # Note: newer hidapi should also work.
 hidapi<0.11
 
-trezor[hidapi]>=0.13.0,<0.14
+trezor[hidapi]>=0.13.4,<0.14
```
SomberNight, line 6: note that the version is also checked at runtime here:
https://github.com/spesmilo/electrum/blob/acde8cd0b797dd27492099956b9c7f54b6a4ccb7/electrum/plugins/trezor/trezor.py#L119

prusnak: Thanks!

However, this raises an interesting topic. The library versions 0.13.x (where x < 4) still work, but will not benefit from the speed up.

Should we set the minimum versions in `contrib/requirements/requirements-hw.txt` and `electrum/electrum/plugins/trezor/trezor.py` to 0.13.0 or 0.13.4?

What do you think?

SomberNight: If there is no runtime error (e.g. python exception raised), then either is fine with me.

It might be better to have loose version bounds in this file - that's the spirit of requirements files in general at least.
E.g. to help a distro packager who is figuring out what to package.
OTOH, if someone pip installs with the [`hardware` extra](https://github.com/spesmilo/electrum/blob/acde8cd0b797dd27492099956b9c7f54b6a4ccb7/setup.py#L44), that uses this file, and if they already have e.g. trezor 0.13.3 installed, then that will not be upgraded (unless using `pip install --upgrade`).

If the performance speedup is really significant, that might justify bumping the minimum to provide consistent UX.

There is also the generated file:
https://github.com/spesmilo/electrum/blob/acde8cd0b797dd27492099956b9c7f54b6a4ccb7/contrib/deterministic-build/requirements-hw.txt#L217
which lists what gets bundled into the binaries. I guess that is what affects the majority of users in practice.

So anyway, up to you - but I would prefer if the runtime check and the requirements file are consistent with each other.

prusnak: I agree that since 0.13.4 is already specified in `electrum/contrib/deterministic-build/requirements-hw.txt` we can keep the minimum version to 0.13.0.

Amended the commit in b4e5f35e6624fc5dca5bf6d073880c5b8ce35d71

SomberNight:
Thanks!

PR #8078, opened 2022-11-18
Favor first output address in invoice/request

accumulator:
Favor first output address in invoice/request instead of fallback address in LN invoice.

Without this change, when configuring electrum with `bolt11_fallback` disabled, and calling `invoice.get_address()` on a lightning enabled invoice, it will return `None`, thereby disabling retrieving BIP21 uri or address from the invoice/request.


SomberNight (approved):
Looks good.
This is not only used by the GUI but also in some of the internal wallet logic, and could lead to e.g. address-reuse. Thanks for the fix!

PR #8079, opened 2022-11-18
Update @lukechilds server address

lukechilds:
I migrated from bitcoin.lukechilds.co to bitcoin.lu.ke. I'll keep the old domain up so old versions of electrum still work but it would be great if newer releases use this address.


accumulator:
great domain name :)

lukechilds:
Haha thanks!

PR #8090, opened 2022-12-05
commands: Remove unused requested_amount and fix a typo

zouppen:
Minor fixes. Noticed that requested_amount is not used, it could be cleaned from there.


SomberNight:
thx

PR #8091, opened 2022-12-05
Add command for currency conversions

zouppen:
This adds command convert_currency which allows command line and RPC users to access the currency conversions already existing in Electrum.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1323,14 +1323,22 @@ class Commands:
         }
 
     @command('n')
-    async def convert_currency(self, from_amount, from_ccy = None, to_ccy = 'BTC'):
+    async def convert_currency(self, from_amount=1, from_ccy = None, to_ccy = None):
         """Converts the given amount of currency to another using the
         configured exchange rate source.
         """
         if not self.daemon.fx.is_enabled():
             raise Exception("FX is disabled. To enable, run: 'electrum setconfig use_exchange_rate true'")
+        # Default currencies
+        if from_ccy is None and to_ccy is None:
+            from_ccy = 'BTC'
+            to_ccy = self.daemon.fx.ccy
+        elif from_ccy is None:
+            from_ccy = 'BTC'
+        elif to_ccy is None:
+            to_ccy = 'BTC'
         # Currency codes are uppercase
-        from_ccy = self.daemon.fx.ccy if from_ccy is None else from_ccy.upper()
+        from_ccy = from_ccy.upper()
         to_ccy = to_ccy.upper()
         # Get current rates
         rate_from = self.daemon.fx.exchange.get_cached_spot_quote(from_ccy)
@@ -1342,11 +1350,13 @@ class Commands:
             raise Exception(f'Currency to convert to ({to_ccy}) is unknown or rate is unavailable')
         # Conversion
         try:
-            amount = Decimal(from_amount) / rate_from * rate_to
+            from_amount = Decimal(from_amount)
+            to_amount = from_amount / rate_from * rate_to
         except InvalidOperation:
             raise Exception("from_amount is not a number")
         return {
-            "amount": self.daemon.fx.ccy_amount_str(amount, False, to_ccy),
+            "from_amount": self.daemon.fx.ccy_amount_str(from_amount, False, from_ccy),
+            "to_amount": self.daemon.fx.ccy_amount_str(to_amount, False, to_ccy),
             "from_ccy": from_ccy,
             "to_ccy": to_ccy,
             "source": self.daemon.fx.exchange.name(),
@@ -1380,7 +1390,6 @@ param_descriptions = {
     'redeem_script': 'redeem script (hexadecimal)',
     'lightning_amount': "Amount sent or received in a submarine swap. Set it to 'dryrun' to receive a value",
     'onchain_amount': "Amount sent or received in a submarine swap. Set it to 'dryrun' to receive a value",
-    'from_amount': 'Amount to convert',
 }
 
 command_options = {
@@ -1432,8 +1441,9 @@ command_options = {
     'connection_string':      (None, "Lightning network node ID or network address"),
     'new_fee_rate': (None, "The Updated/Increased Transaction fee rate (in sat/byte)"),
     'strategies': (None, "Select RBF any one or multiple RBF strategies in any order, separated by ','; Options : 'CoinChooser','DecreaseChange','DecreasePayment' "),
+    'from_amount': (None, "Amount to convert (default: 1)"),
     'from_ccy':    (None, "Currency to convert from"),
-    'to_ccy':      (None, "Currency to convert to (default: BTC)"),
+    'to_ccy':      (None, "Currency to convert to"),
 }
 
 
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1323,23 +1323,20 @@ class Commands:
         }
 
     @command('n')
-    async def convert_currency(self, from_amount=1, from_ccy = None, to_ccy = None):
+    async def convert_currency(self, from_amount=1, from_ccy = '', to_ccy = ''):
         """Converts the given amount of currency to another using the
         configured exchange rate source.
         """
         if not self.daemon.fx.is_enabled():
             raise Exception("FX is disabled. To enable, run: 'electrum setconfig use_exchange_rate true'")
-        # Default currencies
-        if from_ccy is None and to_ccy is None:
-            from_ccy = 'BTC'
-            to_ccy = self.daemon.fx.ccy
-        elif from_ccy is None:
-            from_ccy = 'BTC'
-        elif to_ccy is None:
-            to_ccy = 'BTC'
         # Currency codes are uppercase
         from_ccy = from_ccy.upper()
         to_ccy = to_ccy.upper()
+        # Default currencies
+        if from_ccy == '':
+            from_ccy = "BTC" if to_ccy != "BTC" else self.daemon.fx.ccy
+        if to_ccy == '':
+            to_ccy = "BTC" if from_ccy != "BTC" else self.daemon.fx.ccy
         # Get current rates
         rate_from = self.daemon.fx.exchange.get_cached_spot_quote(from_ccy)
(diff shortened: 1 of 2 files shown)


electrum/commands.py
```diff
@@ -1322,6 +1322,34 @@ async def reverse_swap(self, lightning_amount, onchain_amount, wallet: Abstract_
...
+    async def convert_currency(self, from_amount, from_ccy = None, to_ccy = 'BTC'):
+        """Converts the given amount of currency to another using the
+        configured exchange rate source.
+        """
+        # Currency codes are uppercase
+        from_ccy = self.daemon.fx.ccy if from_ccy is None else from_ccy.upper()
+        to_ccy = to_ccy.upper()
+        # Get current rates
+        rate_from = self.daemon.fx.exchange.get_cached_spot_quote(from_ccy)
+        rate_to = self.daemon.fx.exchange.get_cached_spot_quote(to_ccy)
```
SomberNight, line 1335: Note that which ccy rates are available depends on the configured exchange (config key `use_exchange`) and the configured currency (config key `currency`). In particular for quite a few exchanges, only the `fx.ccy`-`BTC` fx rate is available (depends on the `ExchangeBase.get_rates` implementation).
This should be noted in the docstring.

-----

In particular note that the "BTC" to "BTC" exchange rate is not guaranteed to be present.
```suggestion
        rate_from = self.daemon.fx.exchange.get_cached_spot_quote(from_ccy) if from_ccy != "BTC" else Decimal(1)
        rate_to = self.daemon.fx.exchange.get_cached_spot_quote(to_ccy) if to_ccy != "BTC" else Decimal(1)
```
Alternatively, you can modify `exchange.get_cached_spot_quote()` to special-case "BTC" - that might be cleaner actually.

electrum/commands.py
```diff
@@ -1322,6 +1322,34 @@ async def reverse_swap(self, lightning_amount, onchain_amount, wallet: Abstract_
...
+        # Test if currencies exist
+        if rate_from.is_nan():
+              raise Exception('Currency to convert from is unknown')
+        if rate_to.is_nan():
+              raise Exception('Currency to convert to is unknown')
```
SomberNight, line 1340:
```suggestion
        if rate_from.is_nan():
            raise Exception(f'Currency to convert from ({from_ccy}) is unknown or rate is unavailable')
        if rate_to.is_nan():
            raise Exception(f'Currency to convert to ({to_ccy}) is unknown or rate is unavailable')
```

electrum/commands.py
```diff
@@ -1322,6 +1322,34 @@ async def reverse_swap(self, lightning_amount, onchain_amount, wallet: Abstract_
...
+        # Conversion
+        try:
+            amount = Decimal(from_amount) / rate_from * rate_to
+        except InvalidOperation:
+            raise Exception("from_amount is not a number")
+        return {
+            "amount": self.daemon.fx.ccy_amount_str(amount, False, to_ccy),
```
SomberNight, line 1347: I think we could return both `(from_amount, from_ccy)`, `(to_amount, to_ccy)` pairs.
```suggestion
            "from_amount": self.daemon.fx.ccy_amount_str(from_amount, False, from_ccy),
            "to_amount": self.daemon.fx.ccy_amount_str(amount, False, to_ccy),
```
(potentially even a timestamp to show how fresh the quote is, e.g. expose `fx.exchange._quotes_timestamp`, although it's not that important I guess)

electrum/commands.py
```diff
@@ -1322,6 +1322,34 @@ async def reverse_swap(self, lightning_amount, onchain_amount, wallet: Abstract_
             'onchain_amount': format_satoshis(onchain_amount_sat),
         }
 
+    @command('n')
+    async def convert_currency(self, from_amount, from_ccy = None, to_ccy = 'BTC'):
+        """Converts the given amount of currency to another using the
+        configured exchange rate source.
+        """
```
SomberNight, line 1329: I think both `from_ccy` and `to_ccy` should default to `fx.ccy` and "BTC", as appropriate.
```suggestion
    async def convert_currency(self, from_amount, from_ccy=None, to_ccy=None):
        if to_ccy is None:
            to_ccy = "BTC" if from_ccy != "BTC" else self.daemon.fx.ccy
        if from_ccy is None:
            from_ccy = "BTC" if to_ccy != "BTC" else self.daemon.fx.ccy
```
This would make the following queries easy:
```
$ ./run_electrum convert_currency 1000
{
    "from_amount": "1000.00",
    "from_ccy": "EUR",
    "source": "CoinGecko",
    "to_amount": "0.06193907",
    "to_ccy": "BTC"
}
$ ./run_electrum convert_currency 1 --from_ccy BTC
{
    "from_amount": "1.00000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "16144.90",
    "to_ccy": "EUR"
}
```
-----
Alternatively, we could switch around the defaults, so that `convert_currency` without arguments returns the current price:
```suggestion
    async def convert_currency(self, from_amount=1, from_ccy=None, to_ccy=None):
        if from_ccy is None:
            from_ccy = "BTC" if to_ccy != "BTC" else self.daemon.fx.ccy
        if to_ccy is None:
            to_ccy = "BTC" if from_ccy != "BTC" else self.daemon.fx.ccy
```
```
$ ./run_electrum convert_currency
{
    "from_amount": "1.00000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "16144.32",
    "to_ccy": "EUR"
}
$ ./run_electrum convert_currency --from_amount 1000 --from_ccy EUR
{
    "from_amount": "1000.00",
    "from_ccy": "EUR",
    "source": "CoinGecko",
    "to_amount": "0.06194129",
    "to_ccy": "BTC"
}
$ ./run_electrum convert_currency --from_amount 0.25 --from_ccy BTC
{
    "from_amount": "0.25000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "4036.08",
    "to_ccy": "EUR"
}
```
-----
Both choices look reasonable to me.

SomberNight (commented):
Thanks for the PR!
See comments in-line.

electrum/commands.py
zouppen: I decided to modify `exchange.get_cached_spot_quote()` as you suggested, I'll add that commit to this PR.

electrum/commands.py
SomberNight, line 1329: Note that fx is disabled by default, so we should probably give a friendly error message:
```python
if not self.daemon.fx.is_enabled():
    raise Exception("FX is disabled. To enable, run: 'electrum setconfig use_exchange_rate true'")
```

zouppen:
Most of the things were straightforward but the default currency selection was a bit ambiguous, but I hope I got your idea in my recent commits. Thanks for the review!

electrum/commands.py
```diff
@@ -1322,6 +1322,46 @@ async def reverse_swap(self, lightning_amount, onchain_amount, wallet: Abstract_
...
+        if not self.daemon.fx.is_enabled():
+            raise Exception("FX is disabled. To enable, run: 'electrum setconfig use_exchange_rate true'")
+        # Default currencies
+        if from_ccy is None and to_ccy is None:
+            from_ccy = 'BTC'
+            to_ccy = self.daemon.fx.ccy
+        elif from_ccy is None:
+            from_ccy = 'BTC'
+        elif to_ccy is None:
+            to_ccy = 'BTC'
```
SomberNight, line 1339: The defaults should behave in a way such that if one or both of `from_ccy` and `to_ccy` are unset, they should be set to the `(fx.ccy, "BTC")` pair (or its reverse). The current code does not satisfy that:
```
$ ./run_electrum --testnet convert_currency --from_amount 1 --from_ccy BTC
{
    "from_amount": "1.00000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "1.00000000",
    "to_ccy": "BTC"
}
```
-----
Try e.g.:
```suggestion
        if from_ccy is None:
            from_ccy = "BTC" if to_ccy != "BTC" else self.daemon.fx.ccy
        if to_ccy is None:
            to_ccy = "BTC" if from_ccy != "BTC" else self.daemon.fx.ccy
```
With that, I get the following, which looks ok:
```
$ ./run_electrum convert_currency --from_amount 2
{
    "from_amount": "2.00000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "32472.72",
    "to_ccy": "EUR"
}
$ ./run_electrum convert_currency --from_amount 2 --from_ccy BTC
{
    "from_amount": "2.00000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "32472.72",
    "to_ccy": "EUR"
}
$ ./run_electrum convert_currency --from_amount 1000 --from_ccy EUR
{
    "from_amount": "1000.00",
    "from_ccy": "EUR",
    "source": "CoinGecko",
    "to_amount": "0.06159017",
    "to_ccy": "BTC"
}
$ ./run_electrum convert_currency --from_amount 1000 --to_ccy BTC
{
    "from_amount": "1000.00",
    "from_ccy": "EUR",
    "source": "CoinGecko",
    "to_amount": "0.06159017",
    "to_ccy": "BTC"
}
$ ./run_electrum convert_currency --from_amount 2 --to_ccy EUR
{
    "from_amount": "2.00000000",
    "from_ccy": "BTC",
    "source": "CoinGecko",
    "to_amount": "32472.72",
    "to_ccy": "EUR"
}
```
Feel free to write unit tests if you like, e.g. extend `tests/test_commands.py`.

SomberNight: > The defaults should behave in a way such that if one or both of `from_ccy` and `to_ccy` are unset, they should be set to the `(fx.ccy, "BTC")` pair (or its reverse).

Well, it's a bit more complicated, because of the case where the caller specifies exactly one of the ccy args to a third value, e.g. GBP. In that case I think the best behaviour is to default the other ccy to BTC.

zouppen:
Now I'm happy with it, using the logic proposed by @SomberNight. Authoring unit tests is a good idea but I'll put them in a separate PR if I'm doing it. It would require creating a mock exchange to test against.

SomberNight:
Thank you for the contribution.

Issue #8113, opened 2023-01-02
`get_request_by_addr` incorrectly maps addresses to request ids


MrNaif2018:
When using payment requests feature for some time (details below), `get_request_by_addr` returns incorrect id, often expired or already paid and not the one pending, which leads to `request_status` event sending wrong notifications, thus those who listen on that event are unable to correctly process payments.
https://github.com/spesmilo/electrum/blob/30079c06a8b01feeefd713dc18f28219bc4f4842/electrum/wallet.py#L2446-L2448
I think something is wrong here:
https://github.com/spesmilo/electrum/blob/30079c06a8b01feeefd713dc18f28219bc4f4842/electrum/wallet.py#L2358-L2359

How I was able to reproduce it:
Create first invoice, pay it. Create second invoice, pay it. Create third invoice, let it expire. Create fourth one, pay it - `request_status` sends id of the expired invoice instead.

I think I found out why: get_request_by_addr iterates over each payment request ever existing for an address, so while the old id was expired before, when the payment arrived it happens that get_invoice_status returns PR_PAID for an expired invoice (and for non-expired too), expired one is first in the list and gets returned first
Related: #8061, but it is more critical than before because it can be seen easily due to change to use request ids instead of addresses for payment requests

MrNaif2018:
@SomberNight One of the ways to fix this issue is to introduce status to payment requests and store it together in the wallet storage. Then it can just refer to stored status and not compute each time. What do you think?

MrNaif2018:
There's another possible bug I found:
https://github.com/spesmilo/electrum/blob/d52afee6357475fbae194724ca9b3f0b5e4e5ac0/electrum/wallet.py#L2399

If we paid partially, status is still Unpaid, but list of tx hashes isn't updated because of is_paid check here. I think it could be omitted?

SomberNight:
I've pushed a fix/workaround in https://github.com/spesmilo/electrum/commit/9a0fff2571fa4e992b30e2ee5c1aa7bc8ee7b21f. Please see if you can still reproduce.

Indeed the proper fix would be to solve https://github.com/spesmilo/electrum/issues/8061.

> How I was able to reproduce it:
Create first invoice, pay it. Create second invoice, pay it. Create third invoice, let it expire. Create fourth one, pay it - request_status sends id of the expired invoice instead.

This looks needlessly complicated unless I am missing something. 
[...]

PR #8131, opened 2023-01-09
Test payment requests

MrNaif2018:
This PR adds basic regtest tests for payment request functionality. It tests 3 main things: request_status event is fired, getrequest returns correct status and tx_hashes includes the tx hash which paid the request

Note that it still doesn't catch the bug in #8113 because it requires a custom order of conditions to be tested, but at least this provides basic foundation for the existing functionality not to break

cc @SomberNight

### DIFF
--- a/electrum/tests/regtest.py
+++ b/electrum/tests/regtest.py
@@ -1,19 +1,39 @@
-import os
+import asyncio
+import queue
+import subprocess
 import sys
 import unittest
-import subprocess
+from decimal import Decimal
+
+from electrum import util
+from electrum.commands import Commands
+from electrum.constants import set_regtest
+from electrum.daemon import Daemon
+from electrum.simple_config import SimpleConfig
+from electrum.storage import WalletStorage
+from electrum.wallet import Wallet
+from electrum.wallet_db import WalletDB
+
 
 class TestLightning(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


ecdsa:
Thanks for the PR
Is there a reason why you did this using regtest instead of unittest?
Regtest tests are expensive and slow. 
We try to use them only if inter process communication is needed.

MrNaif2018:
Well, how else would I send payments without regtest funds? This test is actually quite fast when run as part of the suite (bootstrap in regtest is slow but it's already done)
The problem is that without such tests in the past there were bugs which weren't caught by unit tests. It tests the whole payment processing flow with electrum.

ecdsa:
@SomberNight recently mentioned that he wants to add tests for invoices, not sure what kind of tests he had in mind

MrNaif2018:
Yes, I've added them by his request. I am really interested in fixing #8113 ASAP, but it requires some core changes or maybe a wallet upgrade
P.S. I am currently online on IRC

SomberNight:
I've added some unit tests in https://github.com/spesmilo/electrum/commit/7dcaa4b2049ee190ab9ff7a6518775955036c048.

ecdsa:
closing in favor of unittests

PR #8132, opened 2023-01-09
Space between pie chart and balance fix

myxmaster:
On Windows the space between pie chart and balance in the status bar was too little, so I added a space char. Looks still ok on Linux, see screenshots.

Before (Windows):
windows_before

Now (Windows):
winfows_fixed

Now (Linux):
linux

### DIFF
--- a/electrum/gui/qt/balance_dialog.py
+++ b/electrum/gui/qt/balance_dialog.py
@@ -107,7 +107,7 @@ class BalanceToolButton(QToolButton, PieChartObject):
 
     def setText(self, text):
         # this is a hack
-        QToolButton.setText(self, '      ' + text)
+        QToolButton.setText(self, '       ' + text)
 
     def paintEvent(self, event):
         QToolButton.paintEvent(self, event)


SomberNight:
It would be better to remove the hack :P  Or at least try to make it less hackish.
E.g. something like this:
```patch
diff --git a/electrum/gui/qt/balance_dialog.py b/electrum/gui/qt/balance_dialog.py
index ab30e3a684..dcb6d02e9d 100644
--- a/electrum/gui/qt/balance_dialog.py
+++ b/electrum/gui/qt/balance_dialog.py
@@ -34,7 +34,7 @@ from PyQt5.QtGui import QPalette, QPen, QPainter, QPixmap
 
 from electrum.i18n import _
 
-from .util import Buttons, CloseButton, WindowModalDialog, ColorScheme, font_height, AmountLabel
+from .util import Buttons, CloseButton, WindowModalDialog, ColorScheme, font_height, AmountLabel, read_QIcon
 
 if TYPE_CHECKING:
@@ -97,19 +97,22 @@ class BalanceToolButton(QToolButton, PieChartObject):
 
     def __init__(self):
         QToolButton.__init__(self)
-        self.size = max(18, font_height())
         self._list = []
-        self.R = QRect(6, 3, self.size, self.size)
+        QToolButton.setIcon(self, read_QIcon('info.png'))  # FIXME use empty transparent icon here
+        QToolButton.setToolButtonStyle(self, Qt.ToolButtonTextBesideIcon)
 
     def update_list(self, l):
         self._list = l
         self.update()
 
-    def setText(self, text):
-        # this is a hack
-        QToolButton.setText(self, '      ' + text)
-
     def paintEvent(self, event):
+        # recalc position
+        full_height = QToolButton.rect(self).height()
+        self.size = icon_height = self.iconSize().height()
+        margin_y = (full_height - icon_height) // 2
+        margin_x = margin_y
+        self.R = QRect(margin_x, margin_y, self.size, self.size)
+        # paint
         QToolButton.paintEvent(self, event)
         PieChartObject.paintEvent(self, event)
```
Only tested on Linux -- would need to test on all three major OSes.

myxmaster:
I totally agree, but didn't have a good idea yet. I can look into that, but for now I think this "update hack" is at least a little better than before, since it looked irritating on windows.

PR #8133, opened 2023-01-09
Added option for disabling watch-only notification

### LINKED ISSUE #2835: Watching-only Notification on Wallet Open
I use a watching-only wallet on my desktop and always get a notification/popup telling me it's a watching-only wallet every time I open Electrum. It's a mild inconvenience, but it would be nice to have an option to disable this.

I understand if you make to make sure the user knows what is going on, but personally it would be nice to have the option to disable these messages, or at least have the option to only display them when the wallet is first imported/opened.

myxmaster:
This fixes #2835.

I added a checkbox in the warning dialog to disable the warning, also added the option in Preferences -> Misc to toggle it.

warning-dialog

preferences

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -541,13 +541,22 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger, QtEventListener):
         self.export_menu.setEnabled(self.wallet.can_export())
 
     def warn_if_watching_only(self):
-        if self.wallet.is_watching_only():
+        if self.wallet.is_watching_only() and not self.config.get('disable_watch-only_warning', False):
             msg = ' '.join([
                 _("This wallet is watching-only."),
                 _("This means you will not be able to spend Bitcoins with it."),
                 _("Make sure you own the seed phrase or the private keys, before you request Bitcoins to be sent to this wallet.")
             ])
-            self.show_warning(msg, title=_('Watch-only wallet'))
+            checkbox = QCheckBox(_("Don't show again"))
+            class DontShowAgainCheckboxResult:
+                isChecked = False
+            result = DontShowAgainCheckboxResult()
+            def checkboxStateChanged(isChecked, result):
+                result.isChecked = isChecked
+            checkbox.stateChanged.connect(partial(checkboxStateChanged, result=result))
+            self.show_warning(msg, title=_('Watch-only wallet'), checkbox=checkbox)
+            if result.isChecked == Qt.Checked:
+                self.config.set_key('disable_watch-only_warning', True, save=True)
 
     def warn_if_testnet(self):
         if not constants.net.TESTNET:
--- a/electrum/gui/qt/settings_dialog.py
+++ b/electrum/gui/qt/settings_dialog.py
@@ -271,7 +271,7 @@ class SettingsDialog(QDialog, QtEventListener):
         show_watchonly_warning_cb.setChecked(not bool(self.config.get('disable_watch-only_warning', False)))
         def on_set_show_watchonly_warning(v):
             self.config.set_key('disable_watch-only_warning', v != Qt.Checked, save=True)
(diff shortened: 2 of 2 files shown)


SomberNight:
I think there is a real risk of a casual user creating a wallet from an address and not realising it's watch-only. They might have previously selected "don't show again" for another wallet, in which case they would not get a warning anymore.

I suppose if you use a watch-only wallet on a regular basis it might become a nuisance to always get the popup... I imagine that's your motivation, right? Maybe we could add the "don't show" option on a per-wallet basis (save into wallet file instead of config?).

Even then, people might forget that a wallet is watching-only. :/

And re exposing the option in the Preferences, that seems overkill. People who would hide the warning in the first place would not likely re-enable it later IMO.

electrum/gui/qt/main_window.py
```diff
@@ -541,13 +541,22 @@ def watching_only_changed(self):
...
+            checkbox = QCheckBox(_("Don't show again"))
+            class DontShowAgainCheckboxResult:
+                isChecked = False
+            result = DontShowAgainCheckboxResult()
+            def checkboxStateChanged(isChecked, result):
+                result.isChecked = isChecked
+            checkbox.stateChanged.connect(partial(checkboxStateChanged, result=result))
+            self.show_warning(msg, title=_('Watch-only wallet'), checkbox=checkbox)
+            if result.isChecked == Qt.Checked:
+                self.config.set_key('disable_watch-only_warning', True, save=True)
```
SomberNight, line 559: Nit but I would prefer not using dashes in config keys. atm IIRC they all use snake-case, and we should keep it consistent. Just convert any `-` to `_`.

myxmaster:
I thought about that and found it is not a serious problem because worst thing that happens is: You cannot send. And then you very likely remember why :)

Also: I had the idea to add "Watch-only Wallet" in the title bar of the main window. That way people who don't remember turning off watch-only-warning can still see this anytime.

electrum/gui/qt/main_window.py
myxmaster: +1

SomberNight:
> I thought about that and found it is not a serious problem because worst thing that happens is: You cannot send. And then you very likely remember why :)

Some people create imported wallets from random addresses they find on forums/etc, to monitor their history. If they mistakenly hand those out to receive on, the money is lost to them (as the address belongs to someone else).

> Also: I had the idea to add "Watch-only Wallet" in the title bar of the main window.

That's already there:
https://github.com/spesmilo/electrum/blob/10e7d8d9264060b3887f9346d4b44eb1d631ebc5/electrum/gui/qt/main_window.py#L535

myxmaster:
> Maybe we could add the "don't show" option on a per-wallet basis (save into wallet file instead of config?).

You are right, that's a little bit more safe.


> Some people create imported wallets from random addresses they find on forums/etc, to monitor their history. If they mistakenly hand those out to receive on, the money is lost to them (as the address belongs to someone else).

Ok, that's true. What about a warning message in case a user tries to create a payment request for a watch-only wallet? Would you then feel comfortable adding this feature? If not, please comment/close issue #2835.


> That's already there:
> 
> https://github.com/spesmilo/electrum/blob/10e7d8d9264060b3887f9346d4b44eb1d631ebc5/electrum/gui/qt/main_window.py#L535

Uhm yeah, didn't even check, sorry :)

ecdsa:
The popup is also a protection against some attacks. There have been reports of users having a watching-only wallet maliciously installed on their computer. It is a very cheap attack, requires no programming skills. Someone with unauthorized access to a computer would certainly disable that popup too.

> Uhm yeah, didn't even check, sorry :)

No, it is not that you did not check, but that you did not see it.
The title bar thing is not visible enough, this is why we have this popup.

ecdsa:
Note: my previous point about attacks is partially addressed by the fact that watching only wallets can be encrypted. In that case, the fact that one has to type a password protects against the wallet being replaced.

PR #8134, opened 2023-01-10
remove trailing decimal point in format_satoshis

myxmaster:
If "Zeros after decimal point" in Preferences is set to 0 and the amount is an integer (e.g. 1), it is currently displayed as `1.` which looks weird. This pull request removes the trailing decimal point in this case.

I noticed the help dialog for "Zeros after decimal point" says explicitly "1." too, so not sure if this was on purpose? I'd strongly suggest to drop the . in this case...

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -753,6 +753,7 @@ def format_satoshis(
         # add leading whitespaces
         target_total_len = target_integer_len + 1 + target_fract_len
         result = " " * (target_total_len - len(result)) + result
+    result = result.removesuffix('.')
     return result
 
 


myxmaster:
Looks like the current formatting is wanted as this change breaks some tests. I think the current formatting is very uncommon and looks like an error. If you agree to this change I will fix the tests as well.

ecdsa:
I think the '.' is desirable in the GUI. 
Even if the base unit is satoshis, LN payments may have millli-satoshi precision. (although this is not rendered now)

SomberNight:
>  Even if the base unit is satoshis, LN payments may have millli-satoshi precision. (although this is not rendered now)

It is rendered if the `amt_precision_post_satoshi` config key is set.

myxmaster:
> I think the '.' is desirable in the GUI.

Even when the decimal point is the last character displayed, are you sure? I never saw that anywhere before and therefore thought its a bug in Electrum.

Regarding the msat precision rendering: Good point, I didn't think about this and just checked how it behaves with my change. So, what happens in this case (if there are msat places) is: the decimal point is displayed again (which is good), but also if its really the last displayed character (which I think is bad) -> see screenshot below. So I would have to check how to also fix it for this case too, if you agree to this change in general now.

trailing_decimal_point

ecdsa:
> Even when the decimal point is the last character displayed, are you sure? I never saw that anywhere before and therefore thought its a bug in Electrum.

I find it better, but I guess that is a question of taste. 
I am sure I have seen that in other applications. In python at least, it tells you that these are objects of the same type. :-)

myxmaster:
Well... For python devs that might look good, but those are not the target audience for Electrum... I hope? :D

Midar:
+1000 to this change. It's super annoying to have the trailing dot when you change the unit to sats, meaning you see it there for everything and you basically never need it.

PR #8137, opened 2023-01-10
allow for translation of age strings

myxmaster:
Strings in the age and time_difference function were not translated, so I fixed that and added "about" where missing.


ecdsa:
this will not work, because you cannot use % in translations.
you need to use format, see other translated strings that take a parameter.

SomberNight:
> this will not work, because you cannot use % in translations.

I've now documented this in https://github.com/spesmilo/electrum/commit/b0e2e893951d98b88dac5c34d39d49e90dac1c88

ecdsa:
thanks. merged manually, after squashing commits

PR #8146, opened 2023-01-14
Fix Weak Cryptography Issues

openrefactory:
In file: digitalbitbox.py, method: stretch_key, the hashing algorithms used a non-random salt. Hash function without an unpredictable salt increases the risk of a Rainbow Table attack (https://en.wikipedia.org/wiki/Rainbow_table). CWE 759 and CWE 760 provide more information about this. I suggested that a secure random number generator should be used to generate a salt. 

Same coding pattern is found in storage.py file line 151.

### DIFF
--- a/electrum/plugins/digitalbitbox/digitalbitbox.py
+++ b/electrum/plugins/digitalbitbox/digitalbitbox.py
@@ -132,6 +132,8 @@ class DigitalBitbox_Client(HardwareClientBase):
 
 
     def stretch_key(self, key: bytes):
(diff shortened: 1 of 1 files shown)


ecdsa:
This PR appears to have been written by a robot, who does  not seem to understand the context.
Should this robot be banned?

openrefactory:
We are triaging results from several SAST tools and reporting the results to maintainers. This is part of an ongoing experiment on how to create a better triaging portal and help maintainers. We would love to understand the context here. This will help us triage better in future. Thanks in advance for your help.

PR #8149, opened 2023-01-14
Add an AppStream metainfo.xml file

sorenstoutner:
AppStream metainfo.xml files are like extended .desktop files used to populate app stores with information about GUI applications.  Debian and Ubunto and other major Linux distributions are starting to push for their inclusions.  This is a fairly minimal example that just covers the required entries.  The text for the description is a slightly modified version of text taken from Electrum's FAQ.

Metainfo files can contain links to screenshots.  I have not included any here, but that might be something someone would like to add in the future.

### DIFF
--- a/org.electrum.electrum.metainfo.xml
+++ b/org.electrum.electrum.metainfo.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2013 Soren Stoutner <soren@stoutner.com> -->
+<component type="desktop-application">
+  <id>org.electrum.electrum</id>
+
+  <name>Electrum</name>
+  <summary>Bitcoin Wallet</summary>
+
+  <metadata_license>MIT</metadata_license>
+  <project_license>MIT</project_license>
+
+  <description>
+    <p>
+      Electrum is a lightweight Bitcoin wallet focused is speed, with low resource usage and simplifying Bitcoin.
+      Startup times are instant because it operates in conjunction with high-performance servers that handle the most complicated parts of the Bitcoin system.
+    </p>
+  </description>
+
+  <launchable type="desktop-id">electrum.desktop</launchable>
+
+  <content_rating type="oars-1.1" />
+</component>


SomberNight:
What is the goal of adding this file to the repo? If you want it also included in the release distributables, that has to be done explicitly. And if so, which distributable do you want it included in? The sdist ones, I guess (?)

sorenstoutner:
The easiest way to think about it is that you want to include the metainfo.xml file next to the .desktop file in Linux packages.

The .desktop file populates the desktop menu on Linux, and the metainfo.xml file populates app stores on Linux.

The AppStream metainfo.xml specification is fairly new.  In the past, app stores would also use the .desktop file to populate their information.  You can read a little more about it at https://appstream.debian.org/sid/main/issues/electrum.html.

Neither the .desktop or the metainfo.xml file would server any purpose being included in builds that are not targeting Linux.  Additionally, the metainfo.xml file is cross-distribution, but it is only of use for packaged builds that are being offered in app stores, so it doesn't serve any purpose for users who are downloading a tarball and installing it directly.

SomberNight:
So to be clear, is it okay for this file *not* to be included in any distributables?

> The easiest way to think about it is that you want to include the metainfo.xml file next to the .desktop file in Linux packages.

Right... I expected you would want the `metainfo.xml` included in the sdist -- we already include the `.desktop` file.

I guess you use this file somehow when creating the debian package... If so, where do you get the file from if not from the sdist? Would your script notice if we modified this `metainfo.xml` in the upstream repository?

On another note, maybe we should move some of these top-level files to a subdirectory.

sorenstoutner:
> So to be clear, is it okay for this file _not_ to be included in any distributables?

Yes.

> Right... I expected you would want the `metainfo.xml` included in the sdist -- we already include the `.desktop` file.

Yes.

> I guess you use this file somehow when creating the debian package... If so, where do you get the file from if not from the sdist? Would your script notice if we modified this `metainfo.xml` in the upstream repository?

I wrote and included the metainfo.xml file in the 4.3.3 Debian package, but it makes more sense for it to be upstream so packagers for other distributions can benefit from it and so that you can modify the contents if desired.  I will remove my local copy and include the upstream version beginning with the next release.  Any changes made upstream would automatically propagate to future Debian releases.

> On another note, maybe we should move some of these top-level files to a subdirectory.

That would be fine.

On a possibly related note, I noticed that beginning with the 4.3.3 release /electrum/plugins/payserver/www includes a .gitignore and a .editorconfig.  I removed these from the Debian package, but if you have an easy way to exclude them from the sourceonly tarball it would make sense to do so.

sorenstoutner:
I see the `metainfo.xml` file in your Git repository, but I don't see it in your release tarballs.  Is that an oversight?

SomberNight:
Maybe I misunderstood your previous comments, e.g.:

>> So to be clear, is it okay for this file not to be included in any distributables?
>
> Yes.

I was specifically [asking](https://github.com/spesmilo/electrum/pull/8149#issuecomment-1401741940) which distributable should include the xml. :P

So I guess now you are saying it should be included in at least the sourceonly tar.gz, but I guess we might as well put it into both tar.gz's.
If so, `org.electrum.electrum.metainfo.xml` has to be listed in `MANIFEST.in`.

sorenstoutner:
> I was specifically [asking](https://github.com/spesmilo/electrum/pull/8149#issuecomment-1401741940) which distributable should include the xml. :P

I misunderstood what you were asking.  It is OK if the file is not included in a distributable in the sense that the program will continue to function without error.  However, for Debian's purposes, it should be included in the sourceonly tarball.  (I do have the option to package Electrum for Debian by pulling directly from Git and create a tarball, but the tools were originally designed to download a tarball and it is easier on my end to continue to do it that way.)

org.electrum.electrum.metainfo.xml
Quoaymag, line 1: /run

PR #8156, opened 2023-01-18
fix adding coins to coincontrol from address list (fixes #8155)

### LINKED ISSUE #8155: crash when trying to select "spend from" option in address list
Trying to select "Spend from" in context menu in address list results in Electrum crashing with:

```
File ".../electrum/gui/qt/address_list.py", line 291, in <lambda>
    menu.addAction(_("Spend from"), lambda: self.parent.utxo_list.set_spend_list(coins))
AttributeError: 'UTXOList' object has no attribute 'set_spend_list'. Did you mean: 'get_spend_list'?
```

Seems that the issue was introduced in commit 975cdca47482c043d846eb2ea1533c65d7d4b3d1, `set_spend_list` was removed.

Rogach:
This RP attempts to fix #8155 by replacing "Spend from" context menu option with "Add to coin control"/"Remove from coin control" options, similar to how it is done in UTXO list.

### DIFF
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -288,7 +288,10 @@ class AddressList(MyTreeView):
 
         coins = self.wallet.get_spendable_coins(addrs)
         if coins:
-            menu.addAction(_("Spend from"), lambda: self.parent.utxo_list.set_spend_list(coins))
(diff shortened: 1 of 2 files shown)


ecdsa:
thanks

PR #8180, opened 2023-02-01
add callback for sync indicator

accumulator:
Add callback 'adb_pending_txs', which indicates when txs are starting to be
requested from the network (passing True to the callback), and finally
again when AddressSynchronizer is_up_to_date (passing False to the callback).

This pending txs callback is useful to indicate when the wallet is synchronizing
(e.g. initial sync)

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -686,11 +686,16 @@ class AddressSynchronizer(Logger, EventListener):
     def up_to_date_changed(self) -> None:
         # fire triggers
         util.trigger_callback('adb_set_up_to_date', self)
+        if self.is_up_to_date():
+            self.pending_txs_changed(False)
 
     def is_up_to_date(self):
         if not self.synchronizer or not self.verifier:
             return False
         return self.synchronizer.is_up_to_date() and self.verifier.is_up_to_date()
+
+    def pending_txs_changed(self, pending) -> None:
+        util.trigger_callback('adb_txs_pending', pending)
 
     def reset_netrequest_counters(self) -> None:
         if self.synchronizer:
--- a/electrum/synchronizer.py
+++ b/electrum/synchronizer.py
@@ -216,9 +216,9 @@ class Synchronizer(SynchronizerBase):
                 continue  # already have complete tx
             transaction_hashes.append(tx_hash)
             self.requested_tx[tx_hash] = tx_height
-            self.adb.pending_txs_changed(True)
 
         if not transaction_hashes: return
+        self.adb.pending_txs_changed(True)
         async with OldTaskGroup() as group:
             for tx_hash in transaction_hashes:
                 await group.spawn(self._get_transaction(tx_hash, allow_server_not_finding_tx=allow_server_not_finding_tx))
--- a/electrum/synchronizer.py
+++ b/electrum/synchronizer.py
@@ -216,6 +216,7 @@ class Synchronizer(SynchronizerBase):
                 continue  # already have complete tx
             transaction_hashes.append(tx_hash)
             self.requested_tx[tx_hash] = tx_height
+            self.adb.pending_txs_changed(True)
 
         if not transaction_hashes: return
         async with OldTaskGroup() as group:


electrum/synchronizer.py
```diff
@@ -216,6 +216,7 @@ async def _request_missing_txs(self, hist, *, allow_server_not_finding_tx=False)
                 continue  # already have complete tx
             transaction_hashes.append(tx_hash)
             self.requested_tx[tx_hash] = tx_height
+            self.adb.pending_txs_changed(True)
 
         if not transaction_hashes: return
```
SomberNight, line 221: I am not quite sure why the function call is inside the loop. Would it not make more sense to call it just once, outside the loop? e.g.:
```suggestion

        if not transaction_hashes: return
        self.adb.pending_txs_changed(True)
```
(although this might depend on what use case you have in mind - see other remark comment)

electrum/address_synchronizer.py
```diff
@@ -686,12 +686,17 @@ def get_tx_height(self, tx_hash: str) -> TxMinedInfo:
...
+        if self.is_up_to_date():
+            self.pending_txs_changed(False)
 
     def is_up_to_date(self):
         if not self.synchronizer or not self.verifier:
             return False
         return self.synchronizer.is_up_to_date() and self.verifier.is_up_to_date()
 
+    def pending_txs_changed(self, pending) -> None:
+        util.trigger_callback('adb_txs_pending', pending)
```
SomberNight, line 698: What is the motivation / why not just use the existing `adb_set_up_to_date` event?

Also, we would probably want to also pass `self`, for multi-wallet scenarios.

SomberNight: note: AFAICS the GUIs are using the `wallet_updated`/`status` events (and not `adb_set_up_to_date`)
https://github.com/spesmilo/electrum/blob/a174e3780b1f15d467fcfc70914d4a9398fced33/electrum/wallet.py#L467-L470
which makes more sense if you want info about the wallet.
However, due line 468, those events are not fired when opening a wallet until it finished synchronising. i.e. there is no initial "up_to_date==false". Is this your motivation?

SomberNight: Also further context, assuming the goal is to update the sync state progress in the qml gui, both the qt(widgets) and the kivy guis are polling `wallet.is_up_to_date()` every 500 ms.

qt gui:
https://github.com/spesmilo/electrum/blob/a174e3780b1f15d467fcfc70914d4a9398fced33/electrum/gui/qt/__init__.py#L134-L136
https://github.com/spesmilo/electrum/blob/a174e3780b1f15d467fcfc70914d4a9398fced33/electrum/gui/qt/main_window.py#L278
https://github.com/spesmilo/electrum/blob/a174e3780b1f15d467fcfc70914d4a9398fced33/electrum/gui/qt/main_window.py#L867-L869

kivy:
https://github.com/spesmilo/electrum/blob/a174e3780b1f15d467fcfc70914d4a9398fced33/electrum/gui/kivy/main_window.py#L444

accumulator: > What is the motivation / why not just use the existing `adb_set_up_to_date` event?

The motivation is that `adb_set_up_to_date`/`is_up_to_date` is flipping between True and False more frequently than I would like to display the Synchronizing in favor of the wallet balance. 

`is_up_to_date` is a composite of 

```
synchronizer.py:
        return (self._init_done
                and not self._adding_addrs
                and not self.requested_addrs
                and not self._handling_addr_statuses
                and not self.requested_histories
                and not self.requested_tx
                and not self._stale_histories
                and self.status_queue.empty())
verifier.py:
        return (not self.requested_merkle
                and not self.wallet.unverified_tx)
```

.. and for a wallet that's basically up to date (i.e. no new transactions when opening a wallet) I think we shouldn't hide the balance in favor of a synchronizing message just because the internal state is not up_to_date.

IMO when we actually need to fetch TXs from the network, this qualifies as synchronizing, as we will need to recalculate the wallet balance then.

> Also, we would probably want to also pass `self`, for multi-wallet scenarios.

Yeah I'm aware, I should've made this PR a draft :). `self` is however not a wallet here (but it should pass one).


> note: AFAICS the GUIs are using the `wallet_updated`/`status` events (and not `adb_set_up_to_date`)
> 
> https://github.com/spesmilo/electrum/blob/a174e3780b1f15d467fcfc70914d4a9398fced33/electrum/wallet.py#L467-L470
> 
> which makes more sense if you want info about the wallet.
> However, due to line 468, those events are not fired when opening a wallet until it finished synchronising. i.e. there is no initial "up_to_date==false". Is this your motivation?

The initial `up_to_date==false` state event could be assumed in the GUI, so it is not essential. 

> Also further context, assuming the goal is to update the sync state progress in the qml gui, both the qt(widgets) and the kivy guis are polling `wallet.is_up_to_date()` every 500 ms.

Polling is a kludge. There's even a higher frequency poll of 100ms in https://github.com/accumulator/electrum/blob/c4fe97238112139691c65d76e8c4084f9b461039/electrum/synchronizer.py#L263 which, if I reduce that loop temporarily from 100ms to e.g. 2 seconds, it reduces the constant 'idle' CPU usage of electrum by 8-10%.

SomberNight: > > What is the motivation / why not just use the existing `adb_set_up_to_date` event?
> 
> The motivation is that `adb_set_up_to_date`/`is_up_to_date` is flipping between True and False more frequently than I would like to display the Synchronizing in favor of the wallet balance.
> 
> `is_up_to_date` is a composite of [...]
> .. and for a wallet that's basically up to date (i.e. no new transactions when opening a wallet)

Note that unlike `adb.is_up_to_date()`, `wallet.is_up_to_date()` should not be flipping between true/false - or if it is, I would like to know and fix it. The wallet has more info about what is going on, so it has a more complete concept of up_to_date. For example, it is supposed to have a single state transition from False to True during an initial sync after restoring the wallet.
Is the `wallet_updated` event not sufficient for your purposes?

> I think we shouldn't hide the balance in favor of a synchronizing message just because the internal state is not up_to_date.

Ok, I don't mind either way. (IIRC the thinking with that was that the balance could arbitrarily change if we are not synced.)

accumulator: > Note that unlike `adb.is_up_to_date()`, `wallet.is_up_to_date()` should not be flipping between true/false - or if it is, I would like to know and fix it. The wallet has more info about what is going on, so it has a more complete concept of up_to_date. For example, it is supposed to have a single state transition from False to True during an initial sync after restoring the wallet. Is the `wallet_updated` event not sufficient for your purposes?

I tested with `wallet_updated` and `status` events. The events that arrive are as follows:

```
 24.29 | E | gui.qml.qewallet | STATUS, sync=False (event wallet_updated)
 24.35 | E | gui.qml.qewallet | STATUS, sync=False (event status)
121.91 | E | gui.qml.qewallet | STATUS, sync=False (event status)
121.94 | E | gui.qml.qewallet | STATUS, sync=False (event status)
122.01 | E | gui.qml.qewallet | STATUS, sync=True (event wallet_updated)
122.02 | E | gui.qml.qewallet | STATUS, sync=True (event status)
122.23 | E | gui.qml.qewallet | STATUS, sync=False (event wallet_updated)
122.25 | E | gui.qml.qewallet | STATUS, sync=False (event status)
```

Note: sync value here is `not wallet.is_up_to_date()`
Note2: `24.29` is after initial sync completes, `121.91` is when the wallet is fully synced for a while, nothing new arrives, but still the state toggles due to `is_up_to_date` flipflop. This is the short period I'd like the GUI not to indicate synchronizing state..


If there's a way to avoid showing the Synchronizing message using existing events I'm fine with that too, but I'm not sure where `wallet_updated` events are used throughout the codebase, and changing the behavior of might lead to side effects elsewhere.. If the main raison d'être of `wallet_updated` is to indicate sync status, I could investigate adapting `wallet_updated` instead?

> > I think we shouldn't hide the balance in favor of a synchronizing message just because the internal state is not up_to_date.
> 
> Ok, I don't mind either way. (IIRC the thinking with that was that the balance could arbitrarily change if we are not synced.)

I agree, the balance can change if we are not synced, I merely separate the address/history querying phase from the fetch and verify txs we don't know about yet phase (`is_up_to_date()` covers both, the new callback covers only the latter phase)

SomberNight:
IIRC we discussed this on IRC and concluded it is not needed

PR #8185, opened 2023-02-03
Support latest Debian libsecp256k1 package

romanz:
https://packages.debian.org/source/testing/libsecp256k1

### DIFF
--- a/electrum/ecc_fast.py
+++ b/electrum/ecc_fast.py
@@ -47,7 +47,7 @@ def load_library():
         library_paths = ('libsecp256k1.so',)
     else:  # desktop Linux and similar
         library_paths = (os.path.join(os.path.dirname(__file__), 'libsecp256k1.so.0'),
-                         'libsecp256k1.so.0')
+                         'libsecp256k1.so.0', 'libsecp256k1.so.1')
 
     exceptions = []
     secp256k1 = None


romanz:
It's available as:
```
# ls -l /usr/lib/x86_64-linux-gnu/libsecp256k1*
-rw-r--r-- 1 root root 1214390 Dec 27 12:11 /usr/lib/x86_64-linux-gnu/libsecp256k1.a
lrwxrwxrwx 1 root root      21 Dec 27 12:11 /usr/lib/x86_64-linux-gnu/libsecp256k1.so -> libsecp256k1.so.1.0.0
lrwxrwxrwx 1 root root      21 Dec 27 12:11 /usr/lib/x86_64-linux-gnu/libsecp256k1.so.1 -> libsecp256k1.so.1.0.0
-rw-r--r-- 1 root root 1202320 Dec 27 12:11 /usr/lib/x86_64-linux-gnu/libsecp256k1.so.1.0.0
```

SomberNight:
Thanks.
I guess this is related to https://github.com/bitcoin-core/secp256k1/pull/1055.
Would the other cases above not have to be updated? Well, ok, not for the debian pkg ofc.

I also don't understand why our own `contrib/make_libsecp256k1.sh` script puts the old version number in the name of the generated file (after bumping the hardcoded git commit there)... Where is this `.1` coming from? I guess autotools or similar is putting it there...

SomberNight:
I've added a follow-up in https://github.com/spesmilo/electrum/commit/dd141c7fa017deaeffabd9b3b6a340b1da1ac487 based on https://github.com/bitcoin-core/secp256k1/pull/1055#issuecomment-1227505189

but I still don't understand why [`contrib/make_libsecp256k1.sh`](https://github.com/spesmilo/electrum/blob/240edb4a77397d568a6f8c37600c0f0fa0b73ab4/contrib/make_libsecp256k1.sh) is creating dists with the old version number...
`libsecp256k1.la`, generated by libtool, still has the old version number for some reason.
https://github.com/spesmilo/electrum/blob/240edb4a77397d568a6f8c37600c0f0fa0b73ab4/contrib/make_libsecp256k1.sh#L65

SomberNight:
ok, figured it out.
fixed in https://github.com/spesmilo/electrum/commit/c66411f47e4cb7344b55d6ee7720bf763472e181

romanz:
Many thanks!

Kixunil:
Oh, this may be a bad idea. `libsecp256k1` is not considered stable and maybe shouldn't have been packaged.

See recent discussion here: https://github.com/rust-bitcoin/rust-secp256k1/pull/566#discussion_r1051424066
That one is Rust-related but unless Electrum has some protection against loaing the wrong library it still applies.

SomberNight:
I don't fully understand what you mean.
- wasn't the point of `libsecp256k1` finally getting releases/tags that it's now ~stable? (at least for a given version number)
    - indeed many old `.0` libs are likely not sufficient for us (recent ones are), but this PR and recent changes did not touch that
- there's been a debian package of libsecp256k1 (and ubuntu, and probably other distros) for many years
- this PR in particular has not changed much; electrum happily loaded any secp256k1 shared lib it found even before this. this PR and follow-ups just extended that to include the (now stable??) `.1` version.
    - but note that a local a .so is always preferred to a system one (so if you run `./contrib/make_libsecp256k1.sh` yourself, or if you just use an appimage, distro installations are ignored)

What part specifically is the bad idea?

Kixunil:
Basically, I'm not really sure it's safe. Pinging @apoelstra if you have time could you give us a hint whether depending on Ubuntu -packaged `libsecp256k1` is OK?

apoelstra:
If it's the recently-released 0.2 then it should be fine.

We deliberately chose 0.2 because as far as we're aware, the "shouldn't have been packaged" packages used 0.1, so it should be easy to tell :).

PR #8190, opened 2023-02-05
signet on bublina

jsarenik:
Adds https://ex.signet.bublina.eu.org/ to the explorers list and electrs at `bublina.eu.org:60003` (TLS) or `bublina.eu.org:50001` (TCP unencrypted) to the 'electrum/servers_signet.json`.

More info:
The server is running natively on IPv4-only, but tunnels public IPv6 using tunnelbroker.net, Yggdrasil Network and CJDNS. Fair packet queuing [done](https://bin.bublina.eu.org/?ae6eaf60a8a39b7d#qqvmBDkVMALdDdMfS4DRoZ7K7FabbZxeHjzD2R9g8ry) with [`sch_cake`](https://www.bufferbloat.net/projects/codel/wiki/Cake/). Server runs in this setup since 2020, raw hardware, not a VPS. [Busybox](https://busybox.net/) on Linux.


SomberNight:
Thanks for the PR.
I assume you are running this service; in which case even more thanks for that.

PR #8204, opened 2023-02-10
Privacy analysis

ecdsa:
I have decided not to display the list of parent labels in the `utxo` tab, because it was getting too crowded. 
Only the direct parent tx label is displayed, as it was the case before (note that I dropped the `get_label_for_address` fallback).
The number of parents is displayed in a new column of the `utxo` tab, and it already a good indicator of privacy.

For now, `get_tx_parents` only does what its name says: it returns parents, and it does not consider address reuse.
The reason is that results need to be cached, and caching becomes complicated if we include address reuse to the analysis.

After a reverse swap, the number of parents is displayed 2, then it goes down to 1 once the swap funding address is dropped by `lnwatcher`. This is weird, and should probably be addressed. I suppose the same issue may exist with channel sweeps.

I do not have a good GUI representation of the parents list, so it is just a list.
I think a graphical representation is desirable, but out of scope for the moment.
A graphical representation should be able to handle address reuse.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -349,7 +349,7 @@ class AddressSynchronizer(Logger, EventListener):
             self.db.add_transaction(tx_hash, tx)
             self.db.add_num_inputs_to_tx(tx_hash, len(tx.inputs()))
             if is_new:
-                util.trigger_callback('adb_added_tx', self, tx_hash)
+                util.trigger_callback('adb_added_tx', self, tx_hash, tx)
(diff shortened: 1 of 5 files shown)


ecdsa:
the results returned by `get_parent_tx` are now filtered, see 4a568e0d6f54fc1eea0165f10501412540d21332

ecdsa:
reverting to draft; that new commit has poor performance.
the cache should probably contain already filtered results.

Issue #5694, opened 2019-10-10
implement output script descriptors


SomberNight:
We should consider implementing support for output script descriptors.

See:
- https://github.com/bitcoin/bitcoin/blob/08ed87e8875d72a1d8b157b67bbd431253d7db24/doc/descriptors.md
- https://github.com/bitcoin/bitcoin/blob/08ed87e8875d72a1d8b157b67bbd431253d7db24/src/script/descriptor.cpp
- https://github.com/bitcoin-core/HWI/blob/95c9387215fd534bb7a7e3e1885d92cc22457847/hwilib/descriptor.py

These would most likely replace the ypubs/zpubs we currently use (deprecate them).
These would allow e.g. importing and spending from p2pk scripts, or non-HD multisig scripts.

Perhaps more importantly however, when handling PSBTs, signers will often want a derivation path prefix for an xpub and the root xpub's fingerprint. This data is not contained in ypubs/zpubs. Hence, e.g. a wallet created from a zpub cannot create a PSBT that a coldcard will sign; and it's even worse in case of multisig wallets (see https://github.com/spesmilo/electrum/issues/5672#issuecomment-540790451).

-----

Unfortunately Ypubs/Zpubs don't map neatly onto output script descriptors. An output script descriptor describes the whole output script, so e.g. often contains multiple xpubs. Hence a Zpub is basically a partial descriptor. 
Further, our wallet code given an xpub always creates an `m/0/i`, `m/1/j` hierarchy (where the latter is for change); but this can only be described via **two output script descriptors, not one**.

The above means we might have to change the wizard (UI) flow, for example.

To illustrate, see example taken from [HWI docs](https://github.com/bitcoin-core/HWI/blob/95c9387215fd534bb7a7e3e1885d92cc22457847/docs/bitcoin-core-usage.md):
```
$ ./hwi.py -f 8038ecd9 getkeypool --wpkh 0 1000
[
  {
    "desc": "wpkh([8038ecd9/84h/0h/0h]xpub6DR4rqx16YnCcfwFqgwvJdKiWrjDRzqxYTY44aoyHwZDSeSB5n2tqt42aYr9qPKhSKUdftPdTjhHrKKD6WGKVbuyhMvGH76VyKKZubg8o4P/0/*)#36sal9a4",
    "internal": false,
    "range": [0, 1000],
    "timestamp": "now",
    "keypool": true,
    "watchonly": true
  },
  {
[...]

PR #8230, opened 2023-02-27
output script descriptors, part 1: change API of transaction.py

SomberNight:
This is the first set of changes towards adding support for output script descriptors (https://github.com/spesmilo/electrum/issues/5694).

- a new module, `descriptor.py`, is added, which is originally from bitcoin-core/HWI but non-trivially changed in parts and extended
- `transaction.py` is adapted to use the new module, changing some APIs there, mostly for `PartialTxInput`
    - the logic in `PartialTxInput` re how to satisfy a locking script is mostly removed; that is now handled by `descriptor.py` instead
    - when creating a txin programmatically, instead of having to set `script_type`, `num_sig`, `pubkeys`, one will just now set a new field, `script_descriptor` instead
    - the `{PartialTxInput,PartialTxOutput}.script_type` field is removed
    - `PartialTxInput.num_sig`, which was a hack for solving multisig scripts, is removed
    - the `.pubkeys` field no longer has to be set explicitly, it is instead a method (property) derived from `.script_descriptor`
- lots of files are touched to adapt to the changed `transaction.py` API
    - in particular, the HWW plugins are not fully adapted - I've added some clutches to help transition them instead, for now
- there are no wallet changes here yet, that will come separately


SomberNight:
Changed to draft as I've found a regression in the trezor plugin.

PR #8273, opened 2023-03-21
qml: pressing "Esc" on desktop to ~simulate "back" button

SomberNight:
Pressing "Esc" on kivy when testing on desktop closes the active popup/dialog/window/application, similar to what pressing the hw/OS "back" button on Android would do.
I want the same behaviour on qml, as it makes testing more convenient.

This change seems to work; and it seems not to break Android either.
The `Popup.CloseOnEscape` modifiers need to be removed as they are catching the signal before it could get to the application-level `Shortcut`, and if I leave them in, one needs to press Esc multiple times to trigger.

### DIFF
--- a/electrum/gui/qml/components/Pin.qml
+++ b/electrum/gui/qml/components/Pin.qml
@@ -23,7 +23,7 @@ ElDialog {
 
     focus: true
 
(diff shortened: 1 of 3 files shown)


accumulator:
LGTM

accumulator:
reverted in 0ce4acd386259e3e337fc33aaa1ccaa2ecd47d20

PR #8275, opened 2023-03-22
use Qt virtual keyboard for Android QML

accumulator:


Bubbap89 (approved):
Appreciate

accumulator:
As the virtual keyboard will probably be a source of contention, as its behaviour and style is a personal preference, we could opt to make its use a configuration setting. Just like the UI language, we would need to show a message that the app needs to be restarted for the setting to take effect.

thoughts? @ecdsa @SomberNight

ecdsa:
no, it should not be optional. we really do not want users to type their seed in a custom keyboard app.
why do you think it will be source of contention?

SomberNight:
This broke running on desktop for me.
```
 ./run_electrum --testnet -g qml
  1.01 | E | gui.qml.qeapp.QEAppController | unable to bind intent: ModuleNotFoundError("No module named 'android'")
  1.10 | W | gui.qml.qeapp | QQmlApplicationEngine failed to load component
  1.10 | W | gui.qml.qeapp | file:///home/user/wspace/electrum/electrum/gui/qml/components/main.qml:273:5: Type InputPanel unavailable
  1.10 | W | gui.qml.qeapp | qrc:/QtQuick/VirtualKeyboard/content/InputPanel.qml:127:5: Type Keyboard unavailable
  1.10 | W | gui.qml.qeapp | qrc:/QtQuick/VirtualKeyboard/content/components/Keyboard.qml:38:1: module "QtQuick.VirtualKeyboard.Plugins" is not installed
```
Do I need to install something? `qml-module-qtquick-virtualkeyboard` from apt (ubuntu) is apparently already installed.

Issue #8288, opened 2023-03-30
Qt on macOS: checkboxes inside QMenu cannot be distinguished from buttons


SomberNight:
Recently we started using checkable items inside QMenus in many places in the Qt GUI, e.g.:
https://github.com/spesmilo/electrum/blob/8a394c3e3f31c8b54e4cd7287fe202328779f425/electrum/gui/qt/my_treeview.py#L78

Unfortunately these look confusing IMO on macOS, when using the native theme: checkable and not-checkable items cannot be distinguished unless they are checked. Empty checkboxes are not drawn. The qdarkstyle dark theme is unaffected as it uses its own custom stylesheet.

We could potentially fix this for the native ("light") theme the same way it works with qdarkstyle, by customising the stylesheet. We already have some such hacks in [stylesheet_patcher.py](https://github.com/spesmilo/electrum/blob/8a394c3e3f31c8b54e4cd7287fe202328779f425/electrum/gui/qt/stylesheet_patcher.py#L29).

Note in pics below how in the QMenu one cannot tell whether items are checkboxes or buttons:
<details>
 <summary>pics1</summary>

![pic1](https://user-images.githubusercontent.com/29142493/228923366-c92597f1-872c-4674-aedb-3efafae76f1f.png)
![pic2](https://user-images.githubusercontent.com/29142493/228923375-7c7b53ea-b1f3-46aa-b825-598643b8cb3e.png)
</details>

Note that some QMenus are out of our control however, and cannot be patched:
<details>
 <summary>pics2</summary>

![pic3](https://user-images.githubusercontent.com/29142493/228924424-6c949525-9060-405e-a69a-a0896869f399.png)
![pic4](https://user-images.githubusercontent.com/29142493/228924588-fa0ede6c-2a98-4021-9f9f-bcd3b8db55eb.png)
</details>

I think this is a very unfortunate choice from the designers of macOS, it is not nice to hide the information whether a button is checkable or not.

SomberNight:
Turns out certain Linux window managers also behave this way.
@ecdsa said he does not have empty checkboxes on Ubuntu MATE either.

(I have tested gnome and kde, and those look good.)

-----

Let's just close this as won't fix for now. It's not worth fully customising QMenu just for this.

ecdsa:
yes, sorry for the extra work; I misunderstood the nature of the issue when we discussed it

PR #8289, opened 2023-03-30
qt stylesheet: patch QMenu checkbox look on macOS

SomberNight:
This adds a custom stylesheet for QMenu on macOS, so that checkable items can be distinguished from non-checkable ones.
Closes https://github.com/spesmilo/electrum/issues/8288

The stylesheet and the icons are from the "light" theme of qdarkstyle (tag v3.1).
In theory, as we already depend on qdarkstyle and it already includes these icons, we might be able to refer to them without including our own copy, but I could not figure out how (if it is possible).

I don't know whether this patch and approach is a good idea... it feels hackish and it's a lot of changes just to fix a bad decision macOS made. I guess it's somewhat self-contained, so we could include it until it causes problems.

pic1


ecdsa:
closing: we will not fix that for now. (I misunderstood the nature of the issue)

Issue #4300, opened 2018-04-23
malformed launcher command


fliteshare:
Linux Mint 18.3 (sylvia)
kernel x86_64 Linux 4.15.0-15-generic
electrum 3.1.3

icon launcher properties
Installed Command:                      [sh -c "PATH=\"\$HOME/.local/bin:\$PATH\" electrum %u"]
throws error: "unknown command %u"
Please change Command to:        [sh -c "PATH=\"\$HOME/.local/bin:\$PATH\" electrum"]
to make it work !

Thank you.

toxeus:
@fliteshare the most important information is missing here: what is the launcher that cannot cope with the `%u`? I have no issues on Gnome3. Also, the launcher file passes validation:
```sh
desktop-file-validate electrum.desktop
electrum.desktop: hint: value item "Finance" in key "Categories" in group "Desktop Entry" can be extended with another category among the following categories: Office
```

Furthermore, launching electrum from the commandline works too.
```sh
gtk-launch electrum.desktop
```

AbdussamadA:
It doesn't work in kde either. Adding a semi-colon after the path variable fixes it for me:

`Exec=sh -c "PATH=\"\\$HOME/.local/bin:\\$PATH\"; electrum %u"`

dmp1ce:
One problem is that for KDE users the launcher fails. To reproduce try:

```
# kioclient5 exec electrum.desktop 
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 7: " "Invalid escape sequence \"\\\"\"."
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 7: " "Invalid escape sequence \"\\\"\"."
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 7: " "Invalid escape sequence \"\\\"\"."
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 7: " "Invalid escape sequence \"\\\"\"."
"KConfigIni: In file /home/david/Projects/electrum/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
[...]

Issue #8004, opened 2022-10-07
Desktop file contains excessive entries in Exec field


sorenstoutner:
The electrum.desktop files contains excessive entries in the Exec field as described at the following URL:

https://appstream.debian.org/sid/main/issues/electrum.html

Specifically, the electrum.desktop file contains the following two Exec entries:

Exec=sh -c "PATH=\"\\$HOME/.local/bin:\\$PATH\"; electrum %u"
Exec=sh -c "PATH=\"\\$HOME/.local/bin:\\$PATH\"; electrum --testnet %u"

My testing indicates that when installed from a package on Debian they function correctly when changed to the following:

Exec=electrum %u
Exec=electrum --testnet %u

Does anyone know if these extended entries provide any benefit on other systems or in environments that do no use an installation package?  If so, I can modify the lines during Debian packaging.

SomberNight:
Looks like setting the PATH there was intentionally added in https://github.com/spesmilo/electrum/pull/4166
I don't know in what specific scenarios it is useful though.

related https://github.com/spesmilo/electrum/issues/4300

sorenstoutner:
It appears the reasoning was that this makes it easier for people who are installing Electrum into `$HOME/.local/bin` but have not included `$HOME/.local/bin` in their path.  As the discussion in https://github.com/spesmilo/electrum/pull/4166 indicates, there are security issues with having an Exec entry in a system-wide .desktop file modify the path.

I would consider the .desktop file to be the wrong place to modify a user's path.  If a user is installing Electrum to `$HOME/.local/bin`, either the installation script should prompt the user to make sure that `$HOME/.local/bin` is in their path or it ought to modify `$HOME/.bashrc` as part of the installation procedure.

SomberNight:
I've just tested pip installing the release tar.gz directly, on ubuntu 22.04. (I don't have a clean/fresh-install test VM atm however)

Looks like it puts the executable to `$HOME/.local/bin/electrum`,
and the desktop file to `$HOME/.local/share/applications/electrum.desktop`.
```
$ which electrum
$
$ wget https://download.electrum.org/4.3.2/Electrum-4.3.2.tar.gz
[... cut for brevity ...]
Saving to: ‘Electrum-4.3.2.tar.gz’

$ python3 -m pip install Electrum-4.3.2.tar.gz
Defaulting to user installation because normal site-packages is not writeable
Processing ./Electrum-4.3.2.tar.gz
[...]

PR #8306, opened 2023-04-17
Fix double quote escaping in desktop entry (#4300)

l-pt:
In the Exec= value, double quotes (") inside a quoted argument must be escaped with two backslashes (just like the dollar sign).
Moreover, this fixes the following errors, also reported in the issue, when launching electrum.desktop with kioclient5 (KDE plasma program runner):
```
$ kioclient exec /usr/share/applications/electrum.desktop 
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 6: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 6: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 6: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 6: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 6: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 6: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
kf.config.core: "KConfigIni: In file /usr/share/applications/electrum.desktop, line 21: " "Invalid escape sequence \"\\\"\"."
```

I have tested this modification with kioclient5 version 5.27.4 and gtk-launch version 3.24.37, with both launchers electrum is starting correctly without errors and with $HOME/.local/bin prepended to $PATH.

Relevant parts of the freedesktop Desktop Entry Specification:

[https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html)
>  Arguments may be quoted in whole. If an argument contains a reserved character the argument must be quoted. The rules for quoting of arguments is also applicable to the executable name or path of the executable program as provided.

> Quoting must be done by enclosing the argument between double quotes and escaping the double quote character, backtick character ("\`"), dollar sign ("$") and backslash character ("\\") by preceding it with an additional backslash character. Implementations must undo quoting before expanding field codes and before passing the argument to the executable program. Reserved characters are space (" "), tab, newline, double quote, single quote ("'"), backslash character ("\\"), greater-than sign (">"), less-than sign ("<"), tilde ("~"), vertical bar ("|"), ampersand ("&"), semicolon (";"), dollar sign ("$"), asterisk ("*"), question mark ("?"), hash mark ("#"), parenthesis ("(") and (")") and backtick character ("\`").

> Note that the general escape rule for values of type string states that the backslash character can be escaped as ("\\\\") as well and that this escape rule is applied before the quoting rule. As such, to unambiguously represent a literal backslash character in a quoted argument in a desktop entry file requires the use of four successive backslash characters ("\\\\\\\\"). Likewise, a literal dollar sign in a quoted argument in a desktop entry file is unambiguously represented with ("\\\\$").

### DIFF
--- a/contrib/build-linux/appimage/make_appimage.sh
+++ b/contrib/build-linux/appimage/make_appimage.sh
@@ -22,6 +22,7 @@ export GCC_STRIP_BINARIES="1"
 PYTHON_VERSION=3.9.15
 PY_VER_MAJOR="3.9"  # as it appears in fs paths
 PKG2APPIMAGE_COMMIT="a9c85b7e61a3a883f4a35c41c5decb5af88b6b5d"
+DESKTOP_FILE_UTILS_VERSION="0.26"
 
 VERSION=$(git describe --tags --dirty --always)
 APPIMAGE="$DISTDIR/electrum-$VERSION-x86_64.AppImage"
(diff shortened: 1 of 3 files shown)


l-pt:
AppImage build failing due to a desktop-file-validate bug that was fixed 2 years ago:
[https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/commit/63840764746e5d7be502c50ee87c5f0b0c42d24b](https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/commit/63840764746e5d7be502c50ee87c5f0b0c42d24b)

SomberNight:
related:
https://github.com/spesmilo/electrum/pull/4166 and https://github.com/spesmilo/electrum/issues/4300
https://github.com/spesmilo/electrum/issues/8004

l-pt:
AppImage build is now passing

It seems the problem was that AppImageTool bundles the desktop-file-validate program, and the bundled version  (0.24) in AppImageTool 13 is buggy (see [here](https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/commit/63840764746e5d7be502c50ee87c5f0b0c42d24b)).

The bundled version [has been updated](https://github.com/AppImage/AppImageKit/pull/1204) but there have not been any more releases of AppImageTool other than the [continuous build](https://github.com/AppImage/AppImageKit/releases/tag/continuous) which we cannot verify the sha256sum of.

Since we already have to extract AppImageKit to patch mksquashfs it seems reasonable to just substitute the bundled version of desktop-file-validate with the newer version 0.26 that does not have known validation bugs. Let me know.

SomberNight:
Thanks for looking into this.
As an alternative, what do you think about reverting to pre-https://github.com/spesmilo/electrum/pull/4166, like https://github.com/spesmilo/electrum/issues/8004 suggests? i.e. removing the PATH-manipulation

contrib/build-linux/appimage/make_appimage.sh
```diff
@@ -40,10 +41,26 @@ verify_hash "$CACHEDIR/functions.sh" "8f67711a28635b07ce539a9b083b8c12d5488c0000
...
+tar xf "$CACHEDIR/desktop-file-utils-$DESKTOP_FILE_UTILS_VERSION.tar.xz" -C "$CACHEDIR"
+(
+    if [ -f "$CACHEDIR/desktop-file-utils-$DESKTOP_FILE_UTILS_VERSION/src/desktop-file-validate" ]; then
+        info "desktop-file-utils already built, skipping"
+        exit 0
+    fi
+    cd "$CACHEDIR/desktop-file-utils-$DESKTOP_FILE_UTILS_VERSION"
+
+    ./autogen.sh --silent
+    make "-j$CPU_CPOUNT" -s
```
SomberNight, line 62: typo `$CPU_CPOUNT`

l-pt:
> Thanks for looking into this. As an alternative, what do you think about reverting to pre-#4166, like #8004 suggests? i.e. removing the PATH-manipulation

Seems right, after a brief Github search I couldn't find any other python projects changing the PATH in the desktop file

SomberNight:
Ok. Good to hear; as I am much more in favour of removing it than keep hacking on top.
Would you like to create another PR that removes the PATH manipulation? Or should I do it?
As the content of this PR is non-trivial, I'd rather keep it just in case and not have it force-pushed.

PR #8308, opened 2023-04-17
Add support for Ledger Stax

vforgeoux-ledger:
LedgerJS source : https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/devices/src/index.ts#L84


SomberNight:
As it looks like this device is using the same APIs as the existing ones, I guess it is sufficient to test things on e.g. a "nano S plus", and then it should work the same on this one, right?

PR #8309, opened 2023-04-18
remove PATH manipulation from .desktop file

l-pt:
Followup from #8306


electrum.desktop
```diff
@@ -3,7 +3,7 @@
```
SomberNight, line 3: maybe add something like
```
# note: this assumes $HOME/.local/bin is in your PATH
```

SomberNight:
Thanks.

PR #8310, opened 2023-04-18
build: rm "contrib" from debian apt sources lists

SomberNight:
unsure if we actually need contrib. let's see how the CI fares.


SomberNight:
Looks like it's indeed not needed, so I will merge this.

PR #8335, opened 2023-04-23
qml: fix channelopener.connectStr qr scan popping under

SomberNight:
With gossip enabled, trying to open a channel in qml. I try to scan a QR code of a connect-string (node id), seemingly the scan dialog pops under the OpenChannelDialog: pointing the device at a QR code results in a successful scan (so the camera is active despite the image not being shown).

This changes the scanning code in OpenChannelDialog to be similar to other dialogs.

@accumulator see if this looks ok. Feel free to close and fix differently if not.

### DIFF
--- a/electrum/gui/qml/components/OpenChannelDialog.qml
+++ b/electrum/gui/qml/components/OpenChannelDialog.qml
@@ -107,13 +107,13 @@ ElDialog {
                         icon.width: constants.iconSizeMedium
                         scale: 1.2
                         onClicked: {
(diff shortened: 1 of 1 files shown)


accumulator:
I'll have a look. Ideally we should use a Dialog to contain the scanning window and z-order it above any dialog that is active, but in the past I ran into a Qt error doing that (something about the VideoOutput not being happy its parent is the Overlay, but I don't remember exactly)

accumulator:
fixed in 6848b8f375f489f47aaf98da6253d2773354c51d

PR #8336, opened 2023-04-23
some qml wizard fixes

SomberNight:
Small fixes for wizard.py. See individual commits.


accumulator:
LGTM.

(defining instance variables on class level was not working as I initially thought due to my inexperience with Python. Earlier, it looked to me like it was giving default values to instance variables, but even though they are referenced using `self`, they are in fact shared among instances. There might be a few other places I used this construct)

PR #8372, opened 2023-05-01
custom keyboard for seed entry

accumulator:
This adds a custom qwerty keyboard to SeedTextArea, so we can use the system keyboard for the rest of the GUI.


ecdsa:
current version looks good to me

prescot-ww:
still not working on version 4.4.1

the issue is very  simple: th**e tor address can't be paste it!**

SomberNight:
This change is only in 4.4.2, try with that.

Issue #8374, opened 2023-05-01
Not asked for password and can't sign


mcnalu:
I'm using Electrum 4.4.1 on a phone with Android 10. I have a standard wallet with a small but worthwhile amount of bitcoin in it which I use infrequently but I now cannot access.

Following some update to Electrum that must have been between 26 Jan 2023 and 30 April 2023 I can no longer send bitcoin. On opening the app it does not ask for my password which it definitely did before. I can view all my past transactions. When I try to send bitcoin it offers me the Sign button but it does nothing when I tap it. If I try to view my seed, it does ask for a password but when I enter it, nothing happens.

I noticed that the App was denied permission to access Storage in the Android settings, so I allowed that but it made no difference. I'm not sure how this became denied nor how it was able to get my public key to show me transactions in the app without it. I have tried clearing the cache but it made no difference.

I created a new standard test wallet and it behaves exactly as I would expect in that it asks for the password when I start using it and shows me the seed etc

If other info would help diagnose what has gone wrong here, let me know what's needed, logs etc

mcnalu:
A few more clues...

I think the previous time I used electrum and it all worked fine was 4.3.3 back in January.

Looking at the logs, I do not see "| W | gui.qml.qeapp | wallet requires password" when loading the default_wallet which was the wallet I was using and which definitely did have a password. I do see this line when loading wallet_test which I created to troubleshoot the issue.

Log lines related to default_wallet:

`05-01 20:49:25.163 13999 14023 I python  :   2.85 | D | gui.qml.qedaemon | load wallet /data/data/org.electrum.electrum/files/data/wallets/default_wallet
05-01 20:49:25.163 13999 14023 I python  :   2.85 | D | gui.qml.qewalletdb | setting path: /data/data/org.electrum.electrum/files/data/wallets/default_wallet
05-01 20:49:25.164 13999 14023 I python  :   2.85 | I | storage.WalletStorage | wallet path /data/data/org.electrum.electrum/files/data/wallets/default_wallet
05-01 20:49:25.189 13999 14023 I python  :   2.88 | D | util.profiler | WalletDB._load_transactions 0.0176 sec
05-01 20:49:25.192 13999 14047 I python  :   2.88 | I | storage.WalletStorage | wallet path /data/data/org.electrum.electrum/files/data/wallets/default_wallet
05-01 20:49:25.193 13999 14023 I python  :   2.88 | D | gui.qml.qeapp.QEAppController | received intent: 'None'
[...]

PR #8382, opened 2023-05-03
qml: rework auth

accumulator:
now handles keystore-only password and defines a new set of auth methods.


electrum/gui/qml/qewallet.py
```diff
@@ -667,12 +671,59 @@ def deleteInvoice(self, key: str):
...
+        if self.wallet.has_keystore_encryption() and not password:
+            self._logger.debug(f'verifyKeystorePassword({password!r}) failed')
+            return False
+
+        try:
+            self.wallet.keystore.check_password(password)
+            self._logger.debug(f'verifyKeystorePassword({password!r}) success')
+            return True
+        except InvalidPassword as e:
+            self._logger.debug(f'verifyKeystorePassword({password!r}) failed')
```
SomberNight, line 700: let's not log the password :/

accumulator: ah sorry, leftovers..

electrum/gui/qml/qeswaphelper.py
```diff
@@ -424,15 +428,19 @@ def executeSwap(self):
...
-            self.do_normal_swap(lightning_amount, onchain_amount)
+            self._do_execute_forward_swap()
+
+    @auth_protect(method='pin', message=_('Confirm Lightning swap?'))
+    def _do_execute_reverse_swap(self):
+        lightning_amount = self._send_amount
+        onchain_amount = self._receive_amount
+        self.do_reverse_swap(lightning_amount, onchain_amount)
+
+    @auth_protect(method='keystore_or_pin', message=_('Confirm Lightning swap?'))
```
SomberNight, line 442: invalid magic string
we need enums or something :P

ecdsa:
This looks complicated, and I don't have time to review it today. 
I think we can consider it for the next major release, but not as a bug fix to be deployed quickly.
To fix #8374, why not simply ask for the password unconditionally on startup?

accumulator:
> This looks complicated, and I don't have time to review it today. I think we can consider it for the next major release, but not as a bug fix to be deployed quickly. To fix #8374, why not simply ask for the password unconditionally on startup?

This is actually less impact than the alternative, which would need a refactor of the wallet loading logic in relation to qewalletdb, expand its scope beyond just wallet db stuff (qewalletdb is used to trigger whether a password is needed before opening the wallet, amongst other things). 
Also it is IMO an improvement because it allows the auth layer to pass along (transient) passwords or other secrets without the decorated objects having to handle/assume passwords themselves. This could also be used for HWW unlock, passphrases etc.

This PR stays much closer to what e.g. the desktop client does, doesn't need refactoring of the qml wallet loading logic, does not need a password at startup for keystore-only passwords (might even be a feature instead of a bug?)

Finally, I don't think I can fully  implement and test the alternative today, so that would mean a delay until after the weekend

electrum/gui/qml/qewallet.py
```diff
@@ -667,12 +671,56 @@ def deleteInvoice(self, key: str):
...
+    @pyqtSlot(str, result=bool)
+    def verifyKeystorePassword(self, password):
+        if not password:
+            password = None
+
+        if self.wallet.has_keystore_encryption() and not password:
+            return False
+
+        try:
+            self.wallet.keystore.check_password(password)
```
SomberNight, line 695: I don't think this always behaves correctly, e.g. for a multisig wallet that has multiple keystores. This only checks the first keystore. I would prefer if we could just use `wallet.check_password()` directly, which has this logic encapsulated. None of the other GUIs call keystore.check_password() AFAICT.

accumulator: agreed

electrum/gui/qml/auth.py
```diff
@@ -1,9 +1,10 @@
 from functools import wraps, partial
 
-from PyQt5.QtCore import pyqtSignal, pyqtSlot, pyqtProperty
```
Capitalxtends, line 3: Capitalfxend.com

Issue #8411, opened 2023-05-08
ValueError('Exceeds the limit (4300 digits


xeverse:
```
E/P | lnpeer.Peer.[LNGossip, 0359ba82d7-f41bd277] | Exception in main_loop: ValueError('Exceeds the limit (4300 digits) for integer string conversion; use sys.set_int_max_str_digits() to increase the limit')
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/electrum/util.py", line 1239, in wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/electrum/lnpeer.py", line 415, in wrapper_func
    return await func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/electrum/lnpeer.py", line 429, in main_loop
    async with self.taskgroup as group:
  File "/usr/lib/python3.11/site-packages/aiorpcx/curio.py", line 304, in __aexit__
    self.secs = secs
  File "/usr/lib/python3.11/site-packages/electrum/util.py", line 1335, in join
    task.result()
  File "/usr/lib/python3.11/site-packages/electrum/lnpeer.py", line 456, in process_gossip
    await self.network.lngossip.process_gossip(chan_anns, node_anns, chan_upds)
  File "/usr/lib/python3.11/site-packages/electrum/lnworker.py", line 603, in process_gossip
    await run_in_thread(process_node_anns)
  File "/usr/lib/python3.11/site-packages/aiorpcx/curio.py", line 57, in run_in_thread
    return await get_event_loop().run_in_executor(None, func, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/electrum/lnworker.py", line 602, in process_node_anns
    self.channel_db.add_node_announcements(node_anns)
  File "/usr/lib/python3.11/site-packages/electrum/channel_db.py", line 611, in add_node_announcements
[...]

PR #8416, opened 2023-05-11
qml: render reserved channel capacity in a darker tone, take frozen into account

accumulator:
Would this be sufficient to have some indication of #8411 ?

### DIFF
--- a/electrum/gui/qml/components/ChannelDetails.qml
+++ b/electrum/gui/qml/components/ChannelDetails.qml
@@ -135,9 +135,10 @@ Pane {
                                 icon.source: '../../icons/share.png'
                                 icon.color: 'transparent'
                                 onClicked: {
-                                    var dialog = app.genericShareDialog.createObject(root,
-                                        { title: qsTr('Channel node ID'), text: channeldetails.pubkey }
-                                    )
+                                    var dialog = app.genericShareDialog.createObject(root, {
+                                        title: qsTr('Channel node ID'),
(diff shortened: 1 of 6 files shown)


SomberNight:
I think to fully address the linked issue(s), we might have to conditionally display some explanatory text somewhere, when can_send is surprisingly small. Example conditions: `(can_send < 0.5 * local_balance) and (local_balance > 0.2 * capacity)`

Regardless, I think this PR goes in the right direction, I like it. Both UI and code look good.

One small refresh bug: if you freeze a channel, and then use the back button to get to the "channels list", the bar is not updated.
If you use `top-left-menu>Channels`, then the bar gets updated.

accumulator:
> I think to fully address the linked issue(s), we might have to conditionally display some explanatory text somewhere, when can_send is surprisingly small. Example conditions: `(can_send < 0.5 * local_balance) and (local_balance > 0.2 * capacity)`

Added a warning above the `ChannelBar` using the condition you proposed.

> One small refresh bug: if you freeze a channel, and then use the back button to get to the "channels list", the bar is not updated. If you use `top-left-menu>Channels`, then the bar gets updated.

This one is a bit weird. I can see the frozen flag propagate correctly to the `ChannelBar` qml component, yet it is not visually updated in the channel list (it is updated in the channel details page). If I trigger a redraw of the layout by attaching the frozen flag to some other element's `visible` property it updates correctly..

PR #8423, opened 2023-05-12
qml: initial support for lightning address

accumulator:
if recipient matches an email-like address, it is converted to a url and submitted to the normal lnurl-pay flow.

TODOs left
- conflicts with openalias (which is currently not supported in QML regardless)
- For the send-screen, we might need to re-add a manual input option.
- more user friendly error reporting? ('This address does not seem to be a Lightning Address') etc

### DIFF
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -17,7 +17,7 @@ from electrum.transaction import PartialTxOutput
 from electrum.util import (parse_URI, InvalidBitcoinURI, InvoiceError,
                            maybe_extract_lightning_payment_identifier, get_asyncio_loop)
 from electrum.lnutil import format_short_channel_id
-from electrum.lnurl import decode_lnurl, request_lnurl, callback_lnurl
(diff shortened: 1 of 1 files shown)


ecdsa:
I think this kind of features should not be developed for a single GUI. 
I have started a new branch a few weeks ago (https://github.com/spesmilo/electrum/commits/payment_identifiers), because I wanted to avoid this kind of code duplication. 

Unfortunately, the QML part of that branch has become difficult to rebase, due to recent changes.
I am afraid that adding new features to QML will only make it harder to to merge.

accumulator:
> I think this kind of features should not be developed for a single GUI. I have started a new branch a few weeks ago (https://github.com/spesmilo/electrum/commits/payment_identifiers), because I wanted to avoid this kind of code duplication.

Agreed, although adding LN address is a very thin addition to LNURLp flow.

The `payment_identifiers` branch is still a work in progress AFAICS, I think it's good to add a draft PR for it, to increase its visibility and have a place to discuss.

PR #8431, opened 2023-05-15
channel_db: raise specific exception when channelDB not loaded, allow…

accumulator:
…ing lnworker to mark payment as failed.

On mobile, it can take a while before channelDB is loaded. If payment is attempted before the DB is fully loaded, this would result in a payment failure, but also leaves the payment attempt in IN_PROGRESS state. This patch adds a more specific ChannelDBNotLoaded exception class, so we can handle this case more gracefully, since we know the payment didn't succeed.


SomberNight (approved):
Looks good.

PR #8437, opened 2023-05-16
Remove Localbitcoins provider

ZenulAbidin:
Localbitcoins went out of business a few months ago, so there is no point in keeping their exchange rate code inside Electrum.

I see that LocalBitcoins is already not shown in the preferences dialog anymore, but this PR will remove some dead backend code.


ZenulAbidin:
Is it normal for Regtest function tests to be failing in the CI/CD?

accumulator:
> Is it normal for Regtest function tests to be failing in the CI/CD?

Yes that happens occasionally.

PR #8454, opened 2023-05-25
config: introduce ConfigVars (take 3)

SomberNight:
A new config API is introduced, and ~all of the codebase is adapted to it. The old API is kept but mainly only for dynamic usage where its extra flexibility is needed.

Using examples, the old config API looked this (API-1):
```py
>>> config.get("request_expiry", 86400)
604800
>>> config.set_key("request_expiry", 86400)
>>>
```

The new config API instead (API-2):
```py
>>> config.WALLET_PAYREQ_EXPIRY_SECONDS
604800
>>> config.WALLET_PAYREQ_EXPIRY_SECONDS = 86400
>>>
```

The old API operated on arbitrary string keys, the new one uses a static ~enum-like list of variables.

With the new API:
- there is a single centralised list of config variables, as opposed to these being scattered all over
- no more duplication of default values (in the getters)
- there is now some (minimal for now) type-validation/conversion for the config values
- better IDE support / static analysis (PyCharm in particular behaves better than with old API)
- later we could also add docstrings for the config vars, using this same framework, if we want

Note: there is yet a third API added here (API-3), for certain niche/abstract use-cases, where we need a reference to the config variable itself. It should only be used when needed:
```py
>>> var = config.cv.WALLET_PAYREQ_EXPIRY_SECONDS
>>> var
<ConfigVarWithConfig key='request_expiry'>
>>> var.get()
604800
>>> var.set(3600)
>>> var.get_default_value()
86400
>>> var.is_set()
True
>>> var.is_modifiable()
True
```

closes https://github.com/spesmilo/electrum/pull/5640
closes https://github.com/spesmilo/electrum/pull/5649

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -959,7 +959,7 @@ class AddressSynchronizer(Logger, EventListener):
         """
         max_conf = -1
         h = self.db.get_addr_history(address)
-        needs_spv_check = not self.config.get("skipmerklecheck", False)
+        needs_spv_check = not self.config.NETWORK_SKIPMERKLECHECK
         for tx_hash, tx_height in h:
             if needs_spv_check:
                 tx_age = self.get_tx_height(tx_hash).conf
--- a/electrum/base_crash_reporter.py
+++ b/electrum/base_crash_reporter.py
@@ -42,7 +42,6 @@ class CrashReportResponse(NamedTuple):
 
 class BaseCrashReporter(Logger):
     report_server = "https://crashhub.electrum.org"
-    config_key = "show_crash_reporter"
     issue_template = """<h2>Traceback</h2>
 <pre>
 {traceback}
--- a/electrum/base_wizard.py
+++ b/electrum/base_wizard.py
@@ -92,7 +92,7 @@ class BaseWizard(Logger):
         self._stack = []  # type: List[WizardStackItem]
         self.plugin = None  # type: Optional[BasePlugin]
         self.keystores = []  # type: List[KeyStore]
-        self.is_kivy = config.get('gui') == 'kivy'
+        self.is_kivy = config.GUI_NAME == 'kivy'
         self.seed_type = None
 
     def set_icon(self, icon):
@@ -697,7 +697,7 @@ class BaseWizard(Logger):
         self.show_xpub_dialog(xpub=xpub, run_next=lambda x: self.run('choose_keystore'))
 
     def choose_seed_type(self):
-        seed_type = 'standard' if self.config.get('nosegwit') else 'segwit'
+        seed_type = 'standard' if self.config.WIZARD_DONT_CREATE_SEGWIT else 'segwit'
         self.create_seed(seed_type)
 
     def create_seed(self, seed_type):
--- a/electrum/coinchooser.py
+++ b/electrum/coinchooser.py
@@ -24,13 +24,16 @@
 # SOFTWARE.
 from collections import defaultdict
 from math import floor, log10
-from typing import NamedTuple, List, Callable, Sequence, Union, Dict, Tuple, Mapping, Type
+from typing import NamedTuple, List, Callable, Sequence, Union, Dict, Tuple, Mapping, Type, TYPE_CHECKING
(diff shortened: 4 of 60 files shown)


SomberNight:
One drawback I see here is that there would now be three(!) concurrent APIs after this change.

The old API (API-1) needs to kept (in some form) because of some dynamic usages we have, e.g. for the `getconfig`/`setconfig` commands. Or when the config keys are constructed via string manipulation, e.g. for the plugins (`f"use_{plugin_name}"`).
And API-3 is used when previously we were just passing a config key around and call config.get()/set_key() with it later (which is not possible via API-2).

An alternative would be to remove API-2 (the proposed new main API of this PR) and convert all its usages to API-3.
This would essentially be the same as https://github.com/spesmilo/electrum/pull/5649 though.

i.e. instead of (API-2):
```py
>>> config.WALLET_PAYREQ_EXPIRY_SECONDS
604800
>>> config.WALLET_PAYREQ_EXPIRY_SECONDS = 86400
```
The main/typical usage would become (API-3):
```py
>>> config.cv.WALLET_PAYREQ_EXPIRY_SECONDS.get()
604800
>>> config.cv.WALLET_PAYREQ_EXPIRY_SECONDS.set(3600)
```
Then we would only have two APIs (1 and 3).

-----

In any case, I think this PR is an improvement compared to what we had before, and I am ~happy with it as-is.
Also, because of this essentially touching the entire codebase, it needs to merged or changed-then-merged soon; if kept lingering it will diverge fast and even changes on master not visible as conflicts can introduce bugs over time.

accumulator:
I think this is a great improvement.

Retention of API-1 might not be necessary. The plugins could define their own `ConfigVar` in their `__init__.py` as part of the plugin descriptor, so there's no dynamic config key generation needed.

w.r.t API-2 and API-3, I would rather describe it as a single API with additional introspection capabilities, so I don't think we have to choose one over the other.

accumulator:
I guess this is a typo:

https://github.com/SomberNight/electrum/blob/24980feab71957df05b1e55f72fbb7dc8cf34036/electrum/simple_config.py#L829

SomberNight:
> I guess `self2` here is a typo:
> 
> https://github.com/SomberNight/electrum/blob/24980feab71957df05b1e55f72fbb7dc8cf34036/electrum/simple_config.py#L829

No, there is another `self` (type: `SimpleConfig`) declared in the method signature that I did not want to re-define. It is in fact the top-level `self` that is used on the following lines.

accumulator:
> No, there is another `self` (type: `SimpleConfig`) declared in the method signature that I did not want to re-define. It is in fact the top-level `self` that is used on the following lines.

Ah, I see. It just looks a bit odd, but it's simply a declaration, not a ref :)

PR looks good to me, a quick scan through the diffs and some casual testing didn't pop up any issues

SomberNight:
>> No, there is another self (type: SimpleConfig) declared in the method signature that I did not want to re-define. It is in fact the top-level self that is used on the following lines.
> 
> Ah, I see. It just looks a bit odd, but it's simply a declaration, not a ref :)

I've pushed a commit to maybe clarify this for future readers: https://github.com/spesmilo/electrum/pull/8454/commits/dfa2b71bc326883da38865aa8609509eaf70eff2,

and another to migrate some more stuff.

> Retention of API-1 might not be necessary. The plugins could define their own `ConfigVar` in their `__init__.py` as part of the plugin descriptor, so there's no dynamic config key generation needed.

You are right that technically we might be able to remove API-1. (or at least make it private)
Some questions do arise though, e.g. what to do with the `getconfig`/`setconfig` commands. In particular, we could dynamically build an inverse map of configvars (string keys -> ConfigVar), and have getconfig/setconfig do lookups in that. And then maybe raise if the user is trying to set/get a config key that does not exist.

There are also some other CLI-related usages of API-1 that I have not migrated here.

Anyway, I think this could be done later, if we feel inclined.
I think we could merge this PR as-is now.

One small point: we can merge this to master, but branch off 4.4.x from an older commit and not-cherry-pick this change -- I think it is better to exclude this from bugfix releases for now, to be safe.

PR #8458, opened 2023-05-30
Payment Identifiers

accumulator:
- this separates GUI from core handling
 - the PaymentIdentifier class handles network requests
 - the GUI is agnostic about the type of PI


ecdsa:
note: payment_identifiers2 is rebased against master, but does not have the QML changes
https://github.com/spesmilo/electrum/tree/payment_identifiers2

accumulator:
closing in favor of #8462

PR #8459, opened 2023-05-30
bitbox02: update to 6.2.0 to fix a bug on macOS 13.3, and ability to display amount in sats on the device

benma:
See the commit messages for details:

```
bitbox02: update to 6.2.0

6.2.0 was released to put a minimum requirement on hidapi 0.14.0,
which includes the fix for this issue:

https://github.com/libusb/hidapi/issues/531

That bug caused hidapi on macOS 13.3 to report 0 as the interface
number for all hid devices, which led to the bitbox02 multi edition being listed
twice instead of once - once for the main HW wallet interface and once erroneously
For the U2F interface (which should not be listed).
```

2nd commit msg:

```
bitbox02: display amounts in sats if Electrum's base unit is sat

The BitBox02 has the ability to display all amounts in sats instead of
BTC. This was introduced in v9.13.0. If Electrum is configured to show
sats, we propagate this config to the BitBox02.

This is backwards compatible: users with older firmware will see the
values in BTC regardless of the config.
```


electrum/plugins/bitbox02/bitbox02.py
```diff
@@ -510,6 +510,12 @@ def sign_transaction(
                 )
 
         keypath_account = full_path[:-2]
+
+        format_unit = bitbox02.btc.BTCSignInitRequest.FormatUnit.DEFAULT
+        # Base unit is configured to be "sat":
+        if self.config.get_decimal_point() == 0:
```
benma, line 516: I avoided `get_base_unit()` here it because I wasn't sure if it could ever change from 'sat' to e.g. 'sats' or 'SAT', breaking this silently.  `get_decimal_point()` looks more robust to me.

SomberNight: Ok, makes sense.

SomberNight:
This would have been problematic without https://github.com/spesmilo/electrum/commit/1b9cb47623fc4b7f7622a7fd02ca7dd35d6c6a87, but it should now be ok.
I guess the CI would not have caught it, as the relevant CI tasks only care about the frozen/pinned deps.

benma:
> This would have been problematic without [1b9cb47](https://github.com/spesmilo/electrum/commit/1b9cb47623fc4b7f7622a7fd02ca7dd35d6c6a87), but it should now be ok. I guess the CI would not have caught it, as the relevant CI tasks only care about the frozen/pinned deps.

Is the explicit hidapi requirement even needed in requirements-hw.txt if the individual hw wallet packages depend on it themselves, like the bitbox02 package does? Or do the ledger packages not depend on it but should?

SomberNight:
DeviceMgr in plugin.py sort of depends on hidapi independent of what the plugins want, so it's added explicitly.
https://github.com/spesmilo/electrum/blob/184281d2fcf6a45a2e375b14a46fed4904b6b878/electrum/plugin.py#L713-L717

And yes, `ledger-bitcoin` specifically pulls in `ledgercomm` for which hidapi is only an extra, but when specifying in `ledger-bitcoin` you can't specify the indirect extra, so it's needed for ledger specifically like that. (and there is also btchip-python, which does pull in hidapi -- but that package is only for old firmware versioned ledger devices, so might get removed at some point)

Anyway, it's all very convoluted and not too important I guess.

SomberNight:
Thanks

PR #8462, opened 2023-05-31
Payment identifiers

accumulator:
- this separates GUI from core handling
 - the PaymentIdentifier class handles network requests
 - the GUI is agnostic about the type of PI

There's a number of open issues still:
- [x] no UI indication of network lookup yet
- [x] Desktop: no UI indication of amount range yet (lnurlp)
- [ ] Desktop: no UI indication if payment identifier resolve step leads to NOT_FOUND state (e.g. openalias not found)
- [x] Desktop: Clear, Save, Pay and Max buttons toggle according to constraints, but probably not consistently in all cases
- [x] Desktop: recipient, description, amount styling not consistent in all cases


electrum/network.py
```diff
@@ -287,7 +287,7 @@ def __init__(self, config: SimpleConfig, *, daemon: 'Daemon' = None):
         self.asyncio_loop = util.get_asyncio_loop()
         assert self.asyncio_loop.is_running(), "event loop not running"
 
-        assert isinstance(config, SimpleConfig), f"config should be a SimpleConfig instead of {type(config)}"
```
accumulator, line 290: @ecdsa I think we can safely remove this assert, daemon isn't checked either and config is dereferenced almost immediately below so that should nicely fail in the unlikely event some other object is passed in.

electrum/gui/qt/send_tab.py
```diff
@@ -632,10 +632,9 @@ def broadcast_thread():
                 return False, repr(e)
             # success
```
ZatmanOptionz24, line 642: OK we all good man

electrum/payment_identifier.py
```diff
@@ -0,0 +1,719 @@
...
+        for k, v in lnaddr.tags:
+            if k == 'd':
+                description = v
+                break
+        else:
+            description = ''
+        amount = lnaddr.get_amount_sat()
+        return pubkey, amount, description
+
+    async def resolve_openalias(self) -> Optional[dict]:
```
SomberNight, line 672: try copy-pasting into the qt send tab payto_edit on *testnet*:
`donate@getmonero.org`
On master, it shows an error popup saying invalid address; on this PR it raises -- in both cases only when you click "pay..." -- maybe it should already complain before, as soon as the DNS host is resolved, not sure.

electrum/gui/qt/send_tab.py
```diff
@@ -297,226 +308,158 @@ def get_frozen_balance_str(self) -> Optional[str]:
...
         self.do_clear()
 
-    def on_pr(self, request: 'paymentrequest.PaymentRequest'):
-        self.payment_request = request
-        if self.payment_request.verify(self.window.contacts):
-            self.payment_request_ok_signal.emit()
+    def set_field_style(self, w, text, validated):
+        from .util import ColorScheme
+        if validated is None:
+            style = ColorScheme.LIGHTBLUE.as_stylesheet(True)
```
SomberNight, line 339: try pasting an lnurl-pay into the qt payto_e, e.g.:
`LNURL1DP68GURN8GHJ7CN5VDCXZ7FWDPEXVTN0WFNJ7TNHV4KXCTTTDEHHWM30D3H82UNVWQHKSUNXLEM35E`
the `amount_e` also becomes light blue. hinting it got frozen -- but it is not frozen. in fact we expect the user to modify the amount there, I guess (?)

electrum/gui/qt/send_tab.py
```diff
@@ -297,226 +308,158 @@ def get_frozen_balance_str(self) -> Optional[str]:
...
-        self.window.show_send_tab()
-        self.payto_e.disable_checks = True
-        for e in [self.payto_e, self.message_e]:
-            e.setFrozen(True)
-        self.lock_amount(True)
+        self.window.show_send_tab() # FIXME why is this here
+        #for e in [self.payto_e, self.message_e]:
+        # self.payto_e.setFrozen(True)
         for btn in [self.save_button, self.send_button, self.clear_button]:
             btn.setEnabled(False)
```
SomberNight, line 329: try pasting an *mainnet* lnurl-pay into the qt payto_e, when running with `--testnet`.
When you click `Pay...`, you get error: `Wrong Lightning invoice HRP bc1u, should be tb`, after which all buttons are disabled (including `Clear`) and the fields are stuck -> forcing the user to restart the program.

electrum/payment_identifier.py
```diff
@@ -0,0 +1,719 @@
...
+                        lndecode(bolt11)
+                        # if we get here, we have a usable bolt11
+                        self.bolt11 = bolt11
+                    except LnInvoiceException as e:
+                        self.logger.debug(_("Error parsing Lightning invoice") + f":\n{e}")
+                    except IncompatibleOrInsaneFeatures as e:
+                        self.logger.debug(_("Invoice requires unknown or incompatible Lightning feature") + f":\n{e!r}")
+                self.set_state(PaymentIdentifierState.AVAILABLE)
+        elif scriptpubkey := self.parse_output(text):
+            self._type = 'spk'
```
SomberNight, line 358: how about making these an enum? ("spk", "bolt11", "bip70", etc)

electrum/gui/qml/qeinvoice.py
```diff
@@ -493,79 +490,54 @@ def validateRecipient(self, recipient):
...
+        if self._pi.type == 'lnurl':
+            self.on_lnurl(self._pi.lnurl_data)
+            return
+
+        if self._pi.type == 'bip70':
+            self._bip70_payment_request_resolved(self._pi.bip70_data)
+            return
+
+        if self._pi.is_available():
+            if self._pi.type == 'spk':
```
SomberNight, line 514: This looks broken atm for non-address spks -- although it's a question whether we even want to support those in qml, or even if so, whether to do it now.
Still, atm it is broken in a weird unintuitive way.
Try pasting `022782ba871eeb444a202f875c639a5e51bbdd38c5a0610111162cec9a2201254b OP_CHECKSIG`.

electrum/gui/qt/send_tab.py
accumulator: Yes, the amount styling is inconsistent still, will add to top comment

electrum/payment_identifier.py
accumulator: Pay.. button is now disabled if the resolved openalias contains an invalid address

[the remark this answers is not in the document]
accumulator: I think trailing whitespace issues are resolved. The domainname style I haven't implemented yet.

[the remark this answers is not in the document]
accumulator: added domainlike payment identifier

electrum/gui/qt/paytoedit.py
```diff
@@ -78,336 +67,169 @@ def __init__(self):
...
 
     def __init__(self, send_tab: 'SendTab'):
+        QObject.__init__(self, parent=send_tab)
         Logger.__init__(self)
         GenericInputHandler.__init__(self)
-        self.line_edit = QLineEdit()
+
         self.text_edit = ResizingTextEdit()
-        self.text_edit.hide()
+        self.text_edit.textChanged.connect(self._handle_text_change)
```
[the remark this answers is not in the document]
accumulator: The split might have been intentional, but I didn't see any functional reason why. Technically it might have had a reason, but after refactoring this approach didn't fit well and it lead to complications keeping these in sync (in combination with textChanged events).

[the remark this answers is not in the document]
accumulator: > Note that there is a toolbar menu in the Send tab (near top right corner) that includes a pay-to-many checkbox (`send_tab.paytomany_menu`). Now with these changes, the checkbox state can easily become out-of-sync.

Not sure what the idea of the menu toggle is. Should this override the size regardless of the contents (i.e. follow content, until user overrides using the toggle?)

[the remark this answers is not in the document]
SomberNight: I think one of the goals in https://github.com/spesmilo/electrum/commit/1f4cedf56a342ed4b1e1e5dea69e037dd1f8083f was that by default the input field is only a single line, pressing enter is suppressed, and it is using the menu toggle (or explicit pasting (ctrl+v or using button)) that you can change to multiline. Though I think this change was subjective.
I think the main reason @ecdsa split the textedit into lineedit+textedit was the "visual appearance" remark in the commit message: on certain linux styles, one-line QTextEdits look significantly different from QLineEdits, and @ecdsa wanted all the input fields on the Send tab to have a homogene look.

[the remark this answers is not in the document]
accumulator: > on certain linux styles, one-line QTextEdits look significantly different from QLineEdits, and @ecdsa wanted all the input fields on the Send tab to have a homogeneous look.

I see.. I've checked most of the styles that come pre-installed with debian kde, and looked at a windows portable build on windows 11. There is indeed a slight difference on linux with some styles, but it's rather minor IMHO. I'd rather not refactor back to the two alternating edit controls, as it's a maintenance hazard..

I can try fiddling with the styling of the QTextEdit, or I can define a custom edit control deriving from QTextEdit and use that also for single line edit controls, so we're sure to have the same styling everywhere. In any case, I don't want to spend too much time on it :)

accumulator:
@ecdsa, @SomberNight If no further objections, I'll merge to master

ecdsa:
> @ecdsa, @SomberNight If no further objections, I'll merge to master

please wait. I will test this week end

ecdsa:
On master, if Electrum is invoked with a bitcoin URI, then the GUI switches to the send tab. This is useful if you click on a bitcoin URI in the web browser. That does not work on this branch: when a payment identifier is passed as parameter to `run_electrum`, the send tab is populated, but the tab is not set.

accumulator:
> On master, if Electrum is invoked with a bitcoin URI, then the GUI switches to the send tab. This is useful if you click on a bitcoin URI in the web browser. That does not work on this branch: when a payment identifier is passed as parameter to `run_electrum`, the send tab is populated, but the tab is not set.

How do you start the Electrum gui and pass an address (or uri)? It thinks the address is a command, and I can't seem to find the right CLI magic..

ecdsa:
see `commands.py` line 1590, where `url` is set, and used in run_electrum

accumulator:
> see `commands.py` line 1590, where `url` is set, and used in run_electrum

Yes I was looking at that already, it assigns 'url' a positional parameter correct? But no matter how I try, I keep getting this:
```
./run_electrum -D /home/sander/projects/electrum/electrum-test --testnet tb1qhrxvt4sdjn76708qmwmj3s68u2z9ytlxfqm3cz
unknown command: tb1qhrxvt4sdjn76708qmwmj3s68u2z9ytlxfqm3cz
```

run_electrum
```diff
@@ -365,12 +365,6 @@ def main():
     if not config_options.get('verbosity'):
         warnings.simplefilter('ignore', DeprecationWarning)
 
-    # check if we received a valid payment identifier
-    uri = config_options.get('url')
-    if uri and not PaymentIdentifier(None, uri).is_valid():
-        print_stderr('unknown command:', uri)
-        sys.exit(1)
-
```
ecdsa, line 372: I see, this explains why it didn't work for testnet.

accumulator: Yes, this confusion caused me to doubt my cmdline fu :)

Might be better to move the passing of payment identifiers via a switched param (e.g. --uri or something similar) in order to disambiguate from a command

electrum/payment_identifier.py
```diff
@@ -0,0 +1,768 @@
...
+            address = self.openalias_data.get('address')
+            name = self.openalias_data.get('name')
+            description = name
+            recipient = key + ' <' + address + '>'
+            validated = self.openalias_data.get('validated')
+            if not validated:
+                self.warning = _('WARNING: the alias "{}" could not be validated via an additional '
+                                 'security check, DNSSEC, and thus may not be correct.').format(key)
+
+        elif self.bolt11 and self.wallet.has_lightning():
```
ecdsa, line 665: I think the output of this function should not depend on `wallet.has_lightning()`. If a wallet does not support LN, it should handle that itself.

I don't like that `PaymentIdentifier` now has a reference to wallet, just checking where it is used and what for.

run_electrum
ecdsa: last time I checked, it was not possible to hook it to firefox with a --uri parameter

ecdsa: <a href="bitcoin:donate.getmonero.org">this is a test</a>

electrum/payment_identifier.py
accumulator: > I think the output of this function should not depend on `wallet.has_lightning()`. If a wallet does not support LN, it should handle that itself.

Agreed. Also, lnurl type payment identifiers  do not check for this either.

> I don't like that `PaymentIdentifier` now has a reference to wallet, just checking where it is used and what for.

Basically only `wallet.config` and `wallet.contacts` are used.

ecdsa: no, there is also get_invoice

run_electrum
ecdsa: maybe we could provide a separate script for firefox, that launches `run_electrum --uri $1`

accumulator: On linux we can use the `.desktop` file spec which allows us to sculpt the cmdline and let firefox delegate to `xdg-open`.

ecdsa: I see. I was referring to firefox's `network.protocol-handlers` in `about:config`, which does not seem to allow that
I'll open a separate issue about it.

electrum/payment_identifier.py
ecdsa: I added a FIXME in the code regarding that.

PR #8467, opened 2023-06-03
submarine swap server plugin

ecdsa:
- uses the same web API as the Boltz backend
 - lnworker and lnpeer are modified to create and handle invoices for which we do not have the preimage (aka 'hold invoices')
 - payment_secret is random for hold invoices, must be persisted
 - for the moment, server fees are hardcoded

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -211,6 +211,7 @@ class LNWorker(Logger, EventListener, NetworkRetryManager[LNPeerAddr]):
         self.lock = threading.RLock()
         self.node_keypair = generate_keypair(BIP32Node.from_xkey(xprv), LnKeyFamily.NODE_KEY)
         self.backup_key = generate_keypair(BIP32Node.from_xkey(xprv), LnKeyFamily.BACKUP_CIPHER).privkey
+        self.payment_secret_key = generate_keypair(BIP32Node.from_xkey(xprv), LnKeyFamily.PAYMENT_SECRET_KEY).privkey
         self._peers = {}  # type: Dict[bytes, Peer]  # pubkey -> Peer  # needs self.lock
         self.taskgroup = OldTaskGroup()
         self.listen_server = None  # type: Optional[asyncio.AbstractServer]
@@ -1817,9 +1818,7 @@ class LNWallet(LNWorker):
         routing_hints, trampoline_hints = self.calc_routing_hints_for_invoice(amount_msat, channels=channels)
         self.logger.info(f"creating bolt11 invoice with routing_hints: {routing_hints}")
         invoice_features = self.features.for_invoice()
-        payment_preimage = self.get_preimage(payment_hash)
-        if payment_preimage is None:  # e.g. when export/importing requests between wallets
-            raise Exception("missing preimage for payment_hash")
+        payment_secret = self.get_payment_secret(payment_hash)
         amount_btc = amount_msat/Decimal(COIN*1000) if amount_msat else None
         if expiry == 0:
             expiry = LN_EXPIRY_NEVER
@@ -1836,11 +1835,14 @@ class LNWallet(LNWorker):
             + routing_hints
             + trampoline_hints,
             date=timestamp,
-            payment_secret=derive_payment_secret_from_payment_preimage(payment_preimage))
+            payment_secret=payment_secret)
         invoice = lnencode(lnaddr, self.node_keypair.privkey)
         pair = lnaddr, invoice
         self._bolt11_cache[payment_hash] = pair
         return pair
+
+    def get_payment_secret(self, payment_hash):
+        return sha256(sha256(self.payment_secret_key) + payment_hash)
 
     def create_payment_info(self, *, amount_msat: Optional[int], write_to_disk=True) -> bytes:
         payment_preimage = os.urandom(32)
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1873,6 +1873,10 @@ class LNWallet(LNWorker):
                 amount_msat, direction, status = self.payment_info[key]
                 return PaymentInfo(payment_hash, amount_msat, direction, status)
 
+    def add_payment_info_for_hold_invoice(self, payment_hash, lightning_amount_sat):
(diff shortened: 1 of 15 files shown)


electrum/lnworker.py
```diff
@@ -1836,12 +1837,24 @@ def get_bolt11_invoice(
...
         invoice = lnencode(lnaddr, self.node_keypair.privkey)
         pair = lnaddr, invoice
         self._bolt11_cache[payment_hash] = pair
         return pair
 
+    def get_payment_secret(self, payment_hash, create=False):
+        preimage = self.get_preimage(payment_hash)
+        derived_secret = derive_payment_secret_from_payment_preimage(preimage) if preimage else None
+        if create and not preimage:
+            self._payment_secrets[payment_hash] = os.urandom(32)
```
SomberNight, line 1850: I think it would be possible to not have to store payment secrets but derive them from other secret data, e.g.
`payment_secret = H(H(H("payment_secret") || H(node_privkey)) || payment_hash)`
Perhaps we could even just switch over completely to the new scheme, even for invoices where we have the preimage.
If we don't want to break the usecase where a user has generated and given out a payment request and then upgraded their client and then is receiving payment, then in lnpeer.maybe_fulfill_htlc we could maybe derive both payment_secrets and accept either.

electrum/submarine_swaps.py
```diff
@@ -194,9 +216,31 @@ async def _claim_swap(self, swap: SwapData) -> None:
             swap.funding_txid = txin.prevout.txid.hex()
             swap._funding_prevout = txin.prevout
             self._add_or_reindex_swap(swap)  # to update _swaps_by_funding_outpoint
+            funding_height = txin.block_height
             spent_height = txin.spent_height
+
+            if swap.is_reverse and swap.preimage is None:
+                if funding_height <= 0:
+                    continue
```
SomberNight, line 224: `txin.block_height` is not SPV-ed.
Tou can do something like `adb.get_tx_height(txin.prevout.txid.hex()).conf >= 1`

electrum/plugins/swapserver/server.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class SwapServer(Logger, EventListener):
+    """
+    public API:
+    - getpairs
+    - createswap
+    """
+
+    WWW_DIR = os.path.join(os.path.dirname(__file__), 'www')
```
SomberNight, line 25: copy-paste left-over?

electrum/plugins/swapserver/server.py
```diff
@@ -0,0 +1,128 @@
...
+            response = {
+                "id": payment_hash.hex(),
+                "acceptZeroConf": False,
+                "expectedAmount": swap.onchain_amount,
+                "timeoutBlockHeight": swap.locktime,
+                "address": swap.lockup_address,
+                "redeemScript": swap.redeem_script.hex()
+            }
+        else:
+            raise Exception('unsupported request type:' + req_type)
```
SomberNight, line 127: what if `req_type` is not even a string? In general `r` is untrusted input.

1. I haven't checked what happens if this method raises -- but obviously that should not kill the swap server.
2. I would be more comfortable if all fields of the request would be validated as early as logically feasible. E.g. what happens if `claimPublicKey` is 1000 bytes long?

electrum/submarine_swaps.py
```diff
@@ -178,6 +176,30 @@ def start_network(self, *, network: 'Network', lnwatcher: 'LNWalletWatcher'):
...
+            for key in list(self.invoices_to_pay):
+                swap = self.swaps.get(key)
+                if not swap:
+                    continue
+                invoice = self.wallet.get_invoice(key)
+                if not invoice:
+                    continue
+                current_height = self.network.get_local_height()
+                delta = current_height - swap.locktime
+                if delta > - 5:
```
SomberNight, line 193: The same security parameter is 60 for the client. Please use the same value and factor them into a global named value.
5 blocks' time is not enough to get a transaction mined.

electrum/submarine_swaps.py
```diff
@@ -178,6 +176,30 @@ def start_network(self, *, network: 'Network', lnwatcher: 'LNWalletWatcher'):
...
+                    # fixme: should consider cltv of ln payment
+                    self.logger.info('locktime too close {key}')
+                    continue
+                success, log = await self.lnworker.pay_invoice(invoice.lightning_invoice, attempts=1)
+                if not success:
+                    self.logger.info('failed to pay invoice {key}')
```
SomberNight, line 199:
```suggestion
                    self.logger.info(f'failed to pay invoice {key}')
```
same, ~4 lines above

electrum/submarine_swaps.py
SomberNight: Arguably, even 60 is low.

electrum/submarine_swaps.py
```diff
@@ -241,6 +289,62 @@ def add_lnwatcher_callback(self, swap: SwapData) -> None:
...
+        from .bitcoin import construct_script
+        from .crypto import ripemd
+        from .lnaddr import lndecode
+        from .invoices import Invoice
+
+        locktime = self.network.get_local_height() + 140
+        privkey = os.urandom(32)
+        our_pubkey = ECPrivkey(privkey).get_public_key_bytes(compressed=True)
+        is_reverse_for_client = (invoice is not None)
+        if is_reverse_for_client:
```
SomberNight, line 302: I am confused by this terminology. Who is the client? Is it the enduser's electrum client (user of the swap service) or is it the electrum process executing this code (server)? By the naming, I first assumed it is the enduser's electrum client, but that does not seem to be the case:

Note that `SwapManager.reverse_swap()`, which is named from the user's perspective, uses `WITNESS_TEMPLATE_REVERSE_SWAP`. But `is_reverse_for_client` here uses `WITNESS_TEMPLATE_SWAP`.

ecdsa: indeed, I misnamed that variable

electrum/submarine_swaps.py
ecdsa: fixed, thanks

Issue #8292, opened 2023-04-02
Consolidate multiple outputs to same address when using RBF


sha-265:
Is it possible to merge multiple outputs to the same address when using RBF?

For example lets say I sent 0.1 BTC to address A, and after sometime while this RBF transaction is not confirmed yet, I want to send additional amount to the same address. The current Electrum's behavior is that the wallet adds a new output to the original transaction with the same address as the first output with the additional amount. 

If it possible the preferable behavior IMO is that the wallet will consolidate all the outputs which send to the same address, to one output with the total amount. It probably will be cheaper (and more private?) to use only one output with the total amount of all the "transactions" to the same address.

Maybe you should ask the users what they want to do in this case, and maybe just add an option to "edit" the current transaction, or increase the sending amount.

Thanks.

sha-265:
https://github.com/spesmilo/electrum/pull/8474

PR #8474, opened 2023-06-06
Merge multiple outputs to same address

sha-265:
Should be address https://github.com/spesmilo/electrum/issues/8292

I'm not a Python expert nor Bitcoin/Electrum expert, but it's working for me. But this changes should be tested more seriously.

I don't know if I should use `copy` or `deepcopy` or neither, but I did use shallow copy for safety. You can change it to whatever you think is proper.

If you want I can add this option to the wallet settings, and it would be opt-in.

Thanks.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -1792,7 +1792,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
                     lower_bound = max(lower_bound_feerate, lower_bound_relayfee)
                     return max(lower_bound, original_fee_estimator(size))
                 txi = base_tx.inputs()
-                txo = list(filter(lambda o: not self.is_change(o.address), base_tx.outputs())) + list(outputs)       
+                txo = list(filter(lambda o: not self.is_change(o.address), base_tx.outputs())) + list(outputs)
                 output_dict = {}
                 for output in txo:
                     if output.address in output_dict:
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -1792,19 +1792,26 @@ class Abstract_Wallet(ABC, Logger, EventListener):
                     lower_bound = max(lower_bound_feerate, lower_bound_relayfee)
                     return max(lower_bound, original_fee_estimator(size))
                 txi = base_tx.inputs()
-                txo = list(filter(lambda o: not self.is_change(o.address), base_tx.outputs()))
+                txo = list(filter(lambda o: not self.is_change(o.address), base_tx.outputs())) + list(outputs)       
+                output_dict = {}
+                for output in txo:
+                    if output.address in output_dict:
+                        output_dict[output.address].value += output.value
+                    else:
+                        output_dict[output.address] = copy.copy(output)
+                txo = list(output_dict.values())
                 old_change_addrs = [o.address for o in base_tx.outputs() if self.is_change(o.address)]
                 rbf_merge_txid = base_tx.txid()
             else:
                 txi = []
-                txo = []
+                txo = list(outputs)
                 old_change_addrs = []
             # change address. if empty, coin_chooser will set it
             change_addrs = self.get_change_addresses_for_new_transaction(change_addr or old_change_addrs)
             tx = coin_chooser.make_tx(
                 coins=coins,
                 inputs=txi,
-                outputs=list(outputs) + txo,
+                outputs=txo,
                 change_addrs=change_addrs,
                 fee_estimator_vb=fee_estimator,
                 dust_threshold=self.dust_threshold())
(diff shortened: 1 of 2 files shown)


z-weed (commented):
Hi @sha-265, these change lgtm. I did try finding tests for this method and there aren't any in `electrum/tests/test_wallet.py`. As a way of increasing reliability of this change and leaving an artifact for future developers, would you be ok adding one? Essentially validating that `outputs`, in context of `coin_chooser.make_tx` sends correct data and a negative case.

electrum/wallet.py
```diff
@@ -1747,19 +1747,26 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
                     lower_bound = lower_bound if not is_local else 0
                     return int(max(lower_bound, original_fee_estimator(size)))
                 txi = base_tx.inputs()
-                txo = list(filter(lambda o: not self.is_change(o.address), base_tx.outputs()))
+                txo = list(filter(lambda o: not self.is_change(o.address), base_tx.outputs())) + list(outputs)
+                output_dict = {}
+                for output in txo:
+                    if output.address in output_dict:
```
SomberNight, line 1753: Should use `output.scriptpubkey` instead, as not all scripts have a corresponding `address` in which case the value would be None.

electrum/wallet.py
```diff
@@ -1747,19 +1747,26 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
...
+                        output_dict[output.address].value += output.value
+                    else:
+                        output_dict[output.address] = copy.copy(output)
+                txo = list(output_dict.values())
```
SomberNight, line 1757: Note: In general re the whole idea of merging same-address outputs in batch-RBF, the same question arises in `PartialTransaction.join_with_other_psbt`.

SomberNight, line 1757: This whole code block could be factored out to a generic function, moved to `transaction.py`: e.g.
```
def merge_tx_outputs(outputs1, outputs2)
```

SomberNight:
> If you want I can add this option to the wallet settings, and it would be opt-in.

I guess in most cases if this PR affects you, you would want the outputs to be merged. There are some niche ~theoretical counter-examples, but atm they are not worth adding an option.

electrum/wallet.py
sha-265: > ```
> def merge_tx_outputs(outputs1, outputs2)
> ```
@SomberNight, why do we need two arguments in the function?

sha-265: https://github.com/spesmilo/electrum/pull/8474/commits/f5a58675f131140de60c38a43194bdd5d6fdde1d

electrum/wallet.py
sha-265: https://github.com/spesmilo/electrum/pull/8474/commits/f5a58675f131140de60c38a43194bdd5d6fdde1d

electrum/wallet.py
sha-265: > This whole code block could be factored out to a generic function, moved to `transaction.py`: e.g.
> 
> ```
> def merge_tx_outputs(outputs1, outputs2)
> ```

@SomberNight, like this https://github.com/spesmilo/electrum/pull/8474/commits/f5a58675f131140de60c38a43194bdd5d6fdde1d?

sha-265: @SomberNight I added this also for PSBTs: https://github.com/spesmilo/electrum/pull/8474/commits/16375d8c2b50dd978fb89e0da0debfa49582e41a. Is that ok?

SomberNight:
I've merged this manually in 66e2c779eaa5358664c9519b0be4fb83f060490d.
Squashed your commits into ae8a546bb957671c8c917d1cd1b47f2a0f2e689e,
and added a follow-up in 227e2574445662abf03bfba7fe9ec0c755f2af4a.

I've changed my mind and instead turned this into an option, off-by-default.
While reviewing your diff, I've realised there can be subtle consequences, potentially even more than I can see now.
E.g. in daemon usage, when running a swap server, I think it's safer not to combine outputs by default.

For a concrete example, besides the batch-rbf and join psbt use cases, there is also the user manually doing pay-to-many (by entering multiple addresses and amounts in the `pay to` field): your diff would override the explicitly user-entered outputs and merge them (if they reuse an address). Another weird scenario is if the user pays their own change address, and then the coinchooser selects the same change address to send the change to -- but I've decided to ignore that now.

So anyway, for now, this is an option, shown near where the batch-rbf option can be enabled.

Thank you for your contribution.

sha-265:
> your diff would override the explicitly user-entered outputs and merge them (if they reuse an address)

@SomberNight amazing coincidence, because today I thought about this exact scenario and tested it, because I thought to add this functionality to `pay to many` if needed. According to my testing, my diff is not apply to pay to many if you not using the `batch unconfirmed transactions` option. I don't know if it better or worst (I don't see why user should be explicitly want not to merge outputs to same address even in `pay to many` situation), but anyway this is the status of this feature.

PR #8478, opened 2023-06-11
:bug: fix #8469 - fiat balance sorting in address list window

### LINKED ISSUE #8469: Adress EUR Balance Sort Bug
In the adress tap you can see all your adresses. When you sort them by balance it works fine. Sort them By EUR Balance doest wort and its mixed up. In Order to the the order By EUR Balance I sort them by Balance which is basically the same. But it annoys me. Sorting by EUR Balance doesnt work.

Small bug
 Electrum 4.4.4 Linux Appimage

thomasleveil:
# Context
As reported in #8469, in v4.4.4, sorting on the FIAT_BALANCE column of the Address list window is done with amounts considered as text.

# Change
This PR  `setData` on the FIAT_BALANCE column with the numerical value so that sorting works as expected

### DIFF
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -262,6 +262,7 @@ class AddressList(MyTreeView):
         address_item[self.Columns.COIN_BALANCE].setText(balance_text)
         address_item[self.Columns.COIN_BALANCE].setData(balance, self.ROLE_SORT_ORDER)
         address_item[self.Columns.FIAT_BALANCE].setText(fiat_balance_str)
+        address_item[self.Columns.FIAT_BALANCE].setData(balance, self.ROLE_SORT_ORDER)
         address_item[self.Columns.NUM_TXS].setText("%d"%num)
         c = ColorScheme.BLUE.as_color(True) if self.wallet.is_frozen_address(address) else self._default_bg_brush
         address_item[self.Columns.ADDRESS].setBackground(c)


electrum/gui/qt/address_list.py
```diff
@@ -255,13 +255,16 @@ def refresh_row(self, key, row):
...
+            fiat_value = fx.fiat_value(balance, rate)
         else:
             fiat_balance_str = ''
+            fiat_value = None
         address_item = [self.std_model.item(row, col) for col in self.Columns]
         address_item[self.Columns.LABEL].setText(label)
         address_item[self.Columns.COIN_BALANCE].setText(balance_text)
         address_item[self.Columns.COIN_BALANCE].setData(balance, self.ROLE_SORT_ORDER)
         address_item[self.Columns.FIAT_BALANCE].setText(fiat_balance_str)
+        address_item[self.Columns.FIAT_BALANCE].setData(fiat_value, self.ROLE_SORT_ORDER)
```
SomberNight, line 267: `fiat_value` can be `Decimal('NaN')`, e.g. if the rate is unavailable because the FX provider is down or it banned your IP address (e.g. CoinGecko via Tor is not reachable sometimes)
In that case, trying to sort this column will blurt out thousands of errors. (`Decimal('NaN') < Decimal('NaN')`)

The FX rate does not matter anyway if you just want to sort, so let's just sort based on the btc values:
`address_item[self.Columns.FIAT_BALANCE].setData(balance, self.ROLE_SORT_ORDER)`

thomasleveil: Thank you for taking the time to review this change. I amended my commit with your suggested changes

SomberNight:
Thanks

PR #8479, opened 2023-06-11
:sparkles: remove thousand separator when copying numbers to clipboard

thomasleveil:
# Context

I've been frustrated when copying btc/fiat values from the different windows into other software. Copied amounts contain _thousand separator spaces_ (if the option is set in config) and most other software do not cope well with this format.

# Change

From contextual menus, when copying data from a numerical column, spaces are removed.

Let me know if this should be handled in a different manner

### DIFF
--- a/electrum/gui/qt/my_treeview.py
+++ b/electrum/gui/qt/my_treeview.py
@@ -24,6 +24,7 @@
 # SOFTWARE.
 
 import asyncio
+import contextlib
 import enum
 import os.path
 import time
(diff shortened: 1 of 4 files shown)


electrum/gui/qt/my_treeview.py
```diff
@@ -451,6 +452,13 @@ def add_copy_menu(self, menu: QMenu, idx) -> QMenu:
         return cc
 
     def place_text_on_clipboard(self, text: str, *, title: str = None) -> None:
+        if title in {
+            "Amount",
+            "Balance",
+        } or title.endswith(" Value") or title.endswith(" Acquisition price") or title.endswith(" Capital Gains"):
```
SomberNight, line 458: I expect this would raise if title was not provided (i.e. is `None`).
Also, these strings are often translated, in which case the logic does not make sense...

ecdsa:
Sorry, it looks like I merged this a bit too fast...

Issue #8324, opened 2023-04-20
Support external pre-signed inputs on Trezor


matejcik:
It is currently not possible to use Trezor to sign a partial transaction where some inputs come pre-signed.

steps to reproduce:
1. set up an Electrum wallet from Trezor, and another one with hot seed
2. in the Trezor-based wallet, generate a transaction, finalize it but do not sign, Export -> For HWW
3. in the hot wallet, generate a transaction, Combine -> Inputs / outputs
4. paste result from step 2
5. sign transaction
6. back in Trezor-based wallet, Load transaction from text, paste result of step 5
7. try to sign the transaction

This is the [error](https://github.com/spesmilo/electrum/blob/1a4e48e2d4799579d1f5fd7ba39e0f27537ed185/electrum/plugins/trezor/trezor.py#L335): `unexpected txin type: unknown`.

coming from `tx_inputs` line 418: https://github.com/spesmilo/electrum/blob/1a4e48e2d4799579d1f5fd7ba39e0f27537ed185/electrum/plugins/trezor/trezor.py#L418

(don't have the actual traceback on me, sorry.)

What I _guess_ is happening is, Electrum loads the previously signed transaction and does not know how to stream the signed inputs into Trezor.

It is necessary to specify script type `EXTERNAL`, and provide the `script_sig`

SomberNight:
I had a look at this. Could you please review https://github.com/spesmilo/electrum/pull/8483 ?
I reproduced your example with a model T, fw 2.6.0, and it seems to work. Before updating the firmware, with fw version 2.1.x, I was getting an error on the device, but I guess that's ~ok.

matejcik:
> with fw version 2.1.x, I was getting an error on the device, but I guess that's ~ok.

Yes, that's very old and the EXTERNAL script type is not supported there. I wouldn't worry about it.

hynek-jina:
@SomberNight Will this be included in 4.4.6?
I don't see it in [release notes](https://github.com/spesmilo/electrum/blob/master/RELEASE-NOTES) 😞

SomberNight:
No, sorry, it was apparently not cherry-picked. 4.4.6 is already tagged and will be released very soon.

hynek-jina:
Yes, but for this feature I need 4.4.7, right? And it isn't there.

SomberNight:
[...]

Issue #3920, opened 2018-02-18
Wrong behaviour of sending back to address when using TREZOR


prusnak:
If one uses TREZOR and unticks "Use change address" in Preferences > Transactions, the transaction being signed by TREZOR does not indicate that "send-back-to-me" output is, in fact, a change address. This can be easily fixed by using `address_n` instead of `address` field in `TxOutputType` structure.

SomberNight:
Can there still be only one change address? It would be easiest to just use `address_n` for all outputs controlled by the wallet and not special case this.

SomberNight:
Well I am not sure what I am doing wrong but I can't get this to work by simply setting `address_n`.

-----

1-in 2-out txn where the "input address" is a receiving address -- the trezor prompts me to confirm both outputs
```
trezor.tx_outputs() [<TxOutputType: {'address': 'tb1qlkl7d8q2t5tzgypa275hfq0l0dm850xg7pztf9', 'amount': 5000, 'script_type': 0}>, <TxOutputType: {'amount': 994859, 'address_n': [2147483732, 2147483649, 2147483648, 0, 5], 'script_type': 4}>]
```

-----

1-in 2-out txn where the "input address" is a change address -- the trezor only prompts me to confirm the non-change output
```
trezor.tx_outputs() [<TxOutputType: {'address': 'tb1qlkl7d8q2t5tzgypa275hfq0l0dm850xg7pztf9', 'amount': 5000, 'script_type': 0}>, <TxOutputType: {'amount': 10059, 'address_n': [2147483732, 2147483649, 2147483648, 1, 6], 'script_type': 4}>]

```

python-trezor 0.9.0
trezor fw 1.6.0

SomberNight:
TBH, I am completely unfamiliar with trezor fw code, but based on these
- https://github.com/trezor/trezor-mcu/blob/e1fa7af1da79e86ccaae5f3cd2a6c4644f546f8a/firmware/signing.c#L551
- https://github.com/trezor/trezor-mcu/blob/v1.6.0/firmware/signing.c#L394
- https://github.com/trezor/trezor-mcu/blob/e1fa7af1da79e86ccaae5f3cd2a6c4644f546f8a/firmware/signing.c#L579
- https://github.com/trezor/trezor-mcu/blob/e1fa7af1da79e86ccaae5f3cd2a6c4644f546f8a/firmware/transaction.c#L252-L254

[...]

PR #8483, opened 2023-06-14
trezor: support external pre-signed inputs, and support multiple change

SomberNight:
- allow multiple change outputs if fw is new enough
    - related https://github.com/spesmilo/electrum/issues/3920
- support external pre-signed inputs
    - closes https://github.com/spesmilo/electrum/issues/8324


matejcik (approved):
looks good from code perspective and I confirm that it fixes #8324 for me.

only note, not sure if it's worth supporting Trezor firmware 1.6.0 from five years ago :woman_shrugging:

PR #8489, opened 2023-06-16
Swapserver plugin

ecdsa:
todo: 
 - fix the hash returned in get_pairs
 - fix GUI grouping for forward swaps with prepay

(continued from https://github.com/spesmilo/electrum/pull/8467)

### DIFF
--- a/electrum/plugins/swapserver/__init__.py
+++ b/electrum/plugins/swapserver/__init__.py
@@ -1,0 +1,6 @@
+from electrum.i18n import _
+
+fullname = _('SwapServer')
+description = ''
+
+available_for = ['qt', 'cmdline']
--- a/electrum/plugins/swapserver/__init__.py
+++ b/electrum/plugins/swapserver/__init__.py
@@ -1,6 +1,15 @@
 from electrum.i18n import _
 
 fullname = _('SwapServer')
-description = ''
+description = """
+Submarine swap server for an Electrum daemon.
+
+Example setup:
+
+  electrum -o setconfig use_swapserver True
+  electrum -o setconfig swapserver_address localhost:5455
+  electrum daemon -v
+
+"""
 
 available_for = ['qt', 'cmdline']
--- a/electrum/plugins/swapserver/server.py
+++ b/electrum/plugins/swapserver/server.py
@@ -103,7 +103,7 @@ class SwapServer(Logger, EventListener):
             their_pubkey=bytes.fromhex(request['claimPublicKey'])
             assert len(payment_hash) == 32
             assert len(their_pubkey) == 33
-            swap, payment_hash, invoice = sm.add_server_swap(
+            swap, payment_hash, invoice, prepay_invoice = sm.add_server_swap(
                 lightning_amount_sat=lightning_amount_sat,
                 payment_hash=payment_hash,
                 their_pubkey=their_pubkey
@@ -111,7 +111,7 @@ class SwapServer(Logger, EventListener):
             response = {
                 'id': payment_hash.hex(),
                 'invoice': invoice,
-                'minerFeeInvoice': None,
+                'minerFeeInvoice': prepay_invoice,
                 'lockupAddress': swap.lockup_address,
                 'redeemScript': swap.redeem_script.hex(),
                 'timeoutBlockHeight': swap.locktime,
@@ -121,7 +121,7 @@ class SwapServer(Logger, EventListener):
             their_invoice=request['invoice']
             their_pubkey=bytes.fromhex(request['refundPublicKey'])
             assert len(their_pubkey) == 33
-            swap, payment_hash, invoice = sm.add_server_swap(
+            swap, payment_hash, invoice, prepay_invoice = sm.add_server_swap(
                 invoice=their_invoice,
                 their_pubkey=their_pubkey
             )
--- a/electrum/plugins/swapserver/server.py
+++ b/electrum/plugins/swapserver/server.py
@@ -1,0 +1,138 @@
+import os
+import asyncio
+import attr
+import random
+from collections import defaultdict
+
+from aiohttp import ClientResponse
+from aiohttp import web, client_exceptions
+from aiorpcx import timeout_after, TaskTimeout, ignore_after
+from aiorpcx import NetAddress
+
+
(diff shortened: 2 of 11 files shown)


electrum/submarine_swaps.py
```diff
@@ -241,6 +296,87 @@ def add_lnwatcher_callback(self, swap: SwapData) -> None:
         callback = lambda: self._claim_swap(swap)
         self.lnwatcher.add_callback(swap.lockup_address, callback)
 
+    def hold_invoice_callback(self, payment_hash):
+        key = payment_hash.hex()
+        if key in self.swaps:
+            swap = self.swaps[key]
+            if swap.funding_txid is None:
+                coro = self.start_normal_swap(swap, None, None)
+                asyncio.run_coroutine_threadsafe(coro, self.network.asyncio_loop)
```
SomberNight, line 305: note: re timing out hold-invoices, we need to consider how some existing timeouts interact with that, e.g.: https://github.com/spesmilo/electrum/blob/079c0ce5b2d7afc86876aa22ca8126b31b0ac6d1/electrum/lnutil.py#L416-L418

electrum/lnpeer.py
```diff
@@ -1790,15 +1790,6 @@ def log_fail_reason(reason: str):
...
             payment_secret_from_onion = None
 
-        if total_msat > amt_to_forward:
-            mpp_status = self.lnworker.check_received_mpp_htlc(payment_secret_from_onion, chan.short_channel_id, htlc, total_msat)
-            if mpp_status is None:
-                return None, None
-            if mpp_status is False:
-                log_fail_reason(f"MPP_TIMEOUT")
-                raise OnionRoutingFailure(code=OnionFailureCode.MPP_TIMEOUT, data=b'')
-            assert mpp_status is True
```
SomberNight, line 1800: As discussed elsewhere, the next block of code is used for trampoline forwarding -- and this block of code ensures that we wait for all htlcs in an MPP, before we forward trampoline, which we should do. So the ordering is critical here.

ecdsa: reverted to previous order

electrum/submarine_swaps.py
ecdsa: Hold invoices are no longer timed out after they have been accepted (see 141cd524bc4881cee21dc5d265c4a80ffe4b033d)

However, there is still a possible interaction between CLTV expiration and swap expiration: 
 - alice initiates reverse swap (offers HTLC)
 - bob funds the swap
 - alice does not publish her claim transaction
 - the CLTV - 72 deadline is reached
 - alice publishes claim transaction (the swap has not expired yet)
 - bob will extract preimage and fulfill the HTLC. at that point, `should_be_closed_due_to_expiring_htlcs` will return `True`.  if bob's channel counterparty does not revoke quickly, I believe the channel might be force closed by bob.

electrum/submarine_swaps.py
```diff
@@ -171,6 +173,30 @@ def start_network(self, *, network: 'Network', lnwatcher: 'LNWalletWatcher'):
...
+                    self.logger.info(f'locktime too close {key}')
+                    continue
+                success, log = await self.lnworker.pay_invoice(invoice.lightning_invoice, attempts=1)
+                if not success:
+                    self.logger.info(f'failed to pay invoice {key}')
+                    continue
+                self.invoices_to_pay.remove(key)
+
     @log_exceptions
     async def _claim_swap(self, swap: SwapData) -> None:
```
accumulator, line 201: This func has many conditional code blocks on `swap.is_reverse`. For readability and maintainability it might be better to split code for forward swaps and reverse swaps.

electrum/plugins/swapserver/swapserver.py
```diff
@@ -0,0 +1,58 @@
...
+class SwapServerPlugin(BasePlugin):
+
+    def __init__(self, parent, config, name):
+        BasePlugin.__init__(self, parent, config, name)
+        self.config = config
+        self.server = None
+
+    @hook
+    def daemon_wallet_loaded(self, daemon, wallet):
+        # we use the first wallet loaded
```
accumulator, line 46: This will cause problems/confusion when using multiple wallets, the last used wallet is opened first, and this might not be the wallet the swap server needs to use.

electrum/plugins/swapserver/__init__.py
```diff
@@ -0,0 +1,6 @@
+from electrum.i18n import _
+
+fullname = _('SwapServer')
+description = ''
```
accumulator, line 4: This needs at least a minimal description

electrum/plugins/swapserver/server.py
```diff
@@ -0,0 +1,138 @@
...
+        self.addr = NetAddress.from_string(self.config.SWAPSERVER_ADDRESS)
+        self.register_callbacks() # eventlistener
+
+        self.pending = defaultdict(asyncio.Event)
+        self.pending_msg = {}
+
+    @ignore_exceptions
+    @log_exceptions
+    async def run(self):
+        self.root = '/root'
```
accumulator, line 41: unused?

electrum/submarine_swaps.py
ecdsa: fair enough. I will restructure that

electrum/tests/regtest/regtest.sh
```diff
@@ -83,10 +96,10 @@ if [[ $1 == "init" ]]; then
     # alice is funded, bob is listening
     if [[ $2 == "bob" ]]; then
         $bob setconfig --offline lightning_listen localhost:9735
-    else
-        echo "funding $2"
-        $bitcoin_cli sendtoaddress $($agent getunusedaddress -o) 1
+        $bob setconfig --offline use_swapserver true
     fi
+    echo "funding $2"
+    $bitcoin_cli sendtoaddress $($agent getunusedaddress -o) 1
```
SomberNight, line 102: Giving bob 1 btc potentially affects all tests that rely on bob's balance.

Fixed in https://github.com/spesmilo/electrum/commit/d9d281338cef699b70e2e816e1724a34586ef778

ecdsa: thanks!

PR #8493, opened 2023-06-19
partial-writes using jsonpatch

ecdsa:
- partial writes are append only.

 - StoredDict objects will append partial writes to the wallet file when items are added, replaced, removed.

 - Lists in the wallet file that have not been registered as StoredObject are converted to StoredList, which overloads append() and remove(). Those methods too will append partial writes to the wallet file.

 - Unlike the old jsonpatch branch, this branch does not support file encryption. Encrypted files always fully rewritten, even if the change before encryption is a partial write.
 

related: https://github.com/spesmilo/electrum/issues/4823

### DIFF
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -26,6 +26,7 @@ import threading
 import copy
 import json
 from typing import TYPE_CHECKING
+import jsonpatch
 
 from . import util
 from .util import WalletFileException, profiler
@@ -80,22 +81,35 @@ def stored_in(name, _type=dict):
     return decorator
 
 
+def key_path(path, key):
+    def to_str(x):
+        if isinstance(x, int):
+            return str(int(x))
+        else:
+            assert isinstance(x, str)
+            return x
+    return '/' + '/'.join([to_str(x) for x in path + [to_str(key)]])
+
 
 class StoredObject:
 
     db = None
+    path = None
 
     def __setattr__(self, key, value):
-        if self.db:
-            self.db.set_modified(True)
+        if self.db and key not in ['path', 'db'] and not key.startswith('_'):
+            if value != getattr(self, key):
+                self.db.add_patch({'op': 'replace', 'path': key_path(self.path, key), 'value': value})
         object.__setattr__(self, key, value)
 
-    def set_db(self, db):
+    def set_db(self, db, path):
         self.db = db
+        self.path = path
 
     def to_json(self):
         d = dict(vars(self))
         d.pop('db', None)
+        d.pop('path', None)
         # don't expose/store private stuff
         d = {k: v for k, v in d.items()
              if not k.startswith('_')}
@@ -112,20 +126,22 @@ class StoredDict(dict):
         self.path = path
         # recursively convert dicts to StoredDict
         for k, v in list(data.items()):
-            self.__setitem__(k, v)
+            self.__setitem__(k, v, patch=False)
 
     @locked
-    def __setitem__(self, key, v):
+    def __setitem__(self, key, v, patch=True):
         is_new = key not in self
         # early return to prevent unnecessary disk writes
-        if not is_new and self[key] == v:
-            return
+        if not is_new and patch:
+            if self.db and json.dumps(v, cls=self.db.encoder) == json.dumps(self[key], cls=self.db.encoder):
+                return
         # recursively set db and path
         if isinstance(v, StoredDict):
+            #assert v.db is None
             v.db = self.db
             v.path = self.path + [key]
             for k, vv in v.items():
-                v[k] = vv
+                v.__setitem__(k, vv, patch=False)
         # recursively convert dict to StoredDict.
         # _convert_dict is called breadth-first
         elif isinstance(v, dict):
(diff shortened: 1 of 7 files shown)


accumulator:
>     * No storage upgrade is needed; it is possible to test this branch with an existing wallet file, and to switch back to master. Before switching back to master, one needs to trigger a consolidated rewrite of the file from the python console: wallet.db._write(wallet.storage) (this must be called twice because the first call will append a change to qt-console-history)

Can the consolidated write be made explicit? The changing behavior between consecutive calls makes this less deterministic. If this assumption is made then this will lead to some unpredictability in a multithreaded context.

Also, this change would benefit from a unit test (if only to catch changes in the new dependency)

electrum/wallet.py
```diff
@@ -3634,6 +3636,9 @@ def __new__(self, db: 'WalletDB', storage: Optional[WalletStorage], *, config: S
         wallet_type = db.get('wallet_type')
         WalletClass = Wallet.wallet_class(wallet_type)
         wallet = WalletClass(db, storage, config=config)
+        # if file exists, consolidate pending changes
+        if storage and storage.file_exists() and db.pending_changes:
+            db._write(storage)
```
SomberNight, line 3641: Shouldn't this be done in Abstract_Wallet.__init__ instead? At least in the unit tests, `Wallet()` is not always used -- e.g. we might construct a `Standard_Wallet` or a `Multisig_Wallet` directly.

electrum/storage.py
```diff
@@ -100,6 +98,18 @@ def write(self, data: str) -> None:
...
+        """ append data to encrypted file"""
+        if self.is_encrypted():
+            self.decrypted += data
+            self.write(self.decrypted)
+            return
+        with open(self.path, "r+") as f:
+            f.seek(0, os.SEEK_END)
+            f.write(data)
+            f.flush()
+            os.fsync(f.fileno())
```
SomberNight, line 111: Can it happen that the OS process crashes in the middle of this, already having some of `data` (but not all) written to disk?
E.g.
the file initially on disk is:
```
{ "foo": [ "bar", "baz" ] }
```
we try to append this diff:
```
[ { "op": "add", "path": "/foo/1", "value": "qux" } ]
```
so we expect to end up with:
```
{ "foo": [ "bar", "baz" ] }
[ { "op": "add", "path": "/foo/1", "value": "qux" } ]
```
but the file will only contain:
```
{ "foo": [ "bar", "baz" ] }
[ { "op": "add", "path": "/fo
```
-----
Basically I have doubts whether we can treat this as an atomic operation.
Consider the effort done atm in `storage.write` to make the writes atomic:
https://github.com/spesmilo/electrum/blob/fbf41b582a151c6bfa74caf0507d3455bddedba9/electrum/storage.py#L82-L101
There (as we rewrite the whole file every time) we first write to a temp file, and the replace the temp and the main files. That approach does not work for appends though.

SomberNight (commented):
just some comments for now

electrum/storage.py
SomberNight: See e.g. https://lwn.net/Articles/789600/ and https://stackoverflow.com/a/61832882
AFAIU writes to disk cannot be assumed to be atomic even with fsync -- metadata changes are atomic on modern filesystems (e.g. ext4) but data changes are not. Though on ZFS, I think, even data changes are atomic. (still, we ofc cannot assume users are using ZFS :P)

ecdsa:
I realize several of the above comments are better addressed if `storage` is an optional field of `db`.
will refactor this PR.

electrum/wallet.py
ecdsa: Indeed this was not the right place.
This code is now in `JsonDB.load_data`, called from `__init__`.

electrum/storage.py
ecdsa: recovery from incomplete write was added, see `JsonDB.maybe_load_incomplete_data`

electrum/wallet.py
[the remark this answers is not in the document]
ecdsa: I added the `@locked` decorator to `_write` and `_append_pending_changes`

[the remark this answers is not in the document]
ecdsa: That's already what `_write` does. It rewrites the file without pending patches.

electrum/storage.py
```diff
@@ -100,6 +98,18 @@ def write(self, data: str) -> None:
         self._file_exists = True
         self.logger.info(f"saved {self.path}")
 
+    def append(self, data: str) -> None:
+        """ append data to encrypted file"""
+        if self.is_encrypted():
+            self.decrypted += data
+            self.write(self.decrypted)
+            return
```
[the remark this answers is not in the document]
ecdsa: Indeed, since we currently do not have encrypted partial writes, we can just call `_write` if the file is encrypted. 
Added this to condition in `JsonDB.write`

electrum/json_db.py
```diff
@@ -268,17 +347,32 @@ def _convert_value(self, path, key, v):
...
+    @locked
+    def write(self):
+        if self.storage.file_exists() and not self.storage.is_encrypted():
+            self._append_pending_changes()
+        else:
+            self._write()
+
+    @locked
+    def _append_pending_changes(self):
+        if threading.currentThread().isDaemon():
```
SomberNight, line 359: these methods are deprecated
```suggestion
        if threading.current_thread().daemon:
```

run_electrum
```diff
@@ -152,17 +152,13 @@ def init_cmdline(config_options, wallet_path, server, *, config: 'SimpleConfig')
         print_stderr("Exposing a single private key can compromise your entire wallet!")
         print_stderr("In particular, DO NOT use 'redeem private key' services proposed by third parties.")
 
-    # will we need a password
-    if not storage.is_encrypted():
-        db = WalletDB(storage.read(), manual_upgrades=False)
```
SomberNight, line 157: This is not the only place where we load the WalletDB. There is similar logic in the qml/kivy GUIs, and notably in `run_offline_command`. Running `run_offline_command` while another daemon has the wallet file open can result in corruption. Some more generic protection against that would be desirable.

ecdsa: The storage object is now passed to the `WalletDB` constructor:
```python 
db = WalletDB(storage.read(), storage=storage, upgrade=upgrade)
```

If we need to load a `WalletDB` for the sole purpose of checking whether it has a password, or if it requires upgrades, I think we should open it in a storage-less mode, which is possible with:
```python 
db = WalletDB(storage.read(), storage=None, upgrade=upgrade)
```

electrum/json_db.py
```diff
@@ -268,17 +347,32 @@ def _convert_value(self, path, key, v):
...
+            self.logger.warning('daemon thread cannot write db')
+            return
+        if not self.pending_changes:
+            return
+        s = ''.join([',\n' + x for x in self.pending_changes])
```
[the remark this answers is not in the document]
ecdsa: Rather than prepending an integer to every line, it would be cleaner to write that integer in the update.
For example:

```diff --git a/electrum/json_db.py b/electrum/json_db.py
index 579a7f7b4..183754e1d 100644
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -241,9 +241,12 @@ class JsonDB(Logger):
         if patches:
             # apply patches
             self.logger.info('found %d patches'%len(patches))
+            for n, patch in enumerate(patches):
+                assert patch['id'] == n
             patch = jsonpatch.JsonPatch(patches)
             data = patch.apply(data)
             self.set_modified(True)
+        self.num_patches = len(patches)
         return data, bool(patches)
 
     def maybe_load_incomplete_data(self, s):
@@ -274,6 +277,8 @@ class JsonDB(Logger):
 
     @locked
     def add_patch(self, patch):
+        patch['id'] = self.num_patches
+        self.num_patches += 1
         self.pending_changes.append(json.dumps(patch, cls=self.encoder))
         self.set_modified(True)
 
@@ -392,3 +397,4 @@ class JsonDB(Logger):
         self.storage.write(json_str)
         self.pending_changes = []
         self.set_modified(False)
+        self.num_patches = 0

```

I am not sure if this is really useful, though. The current code already checks that no other process (or `WalletDB` instance in the same process) appended to the file, by checking the file length every time it appends. It seems to me that this is a stronger protection, because the check is performed earlier (at write time, not at consolidation time).

Anyway, I wrote the above patch because I wanted to compare both approaches.

PR #8536, opened 2023-07-14
fix sweeping chan after local force-close using cb

SomberNight:
scenario:
- user opens a lightning channel and exports an "imported channel backup"
- user closes channel via local-force-close
  - local ctx is published, to_local output has user's funds and they are CSV-locked for days
- user restores wallet file from seed and imports channel backup
- new wallet file should be able to sweep coins from to_local output (after CSV expires)

This was not working previously, as the local_payment_basepoint was not included in the
imported channel backups, and the code was interpreting the lack of this as the channel not
having option_static_remotekey enabled. This resulted in lnutil.extract_ctn_from_tx
using an incorrect funder_payment_basepoint, and lnsweep not recognising the ctx due to
the garbage ctn value.

The imported channel backup serialisation format is slightly changed to include the
previously missing field, and its version number is bumped (0->1). We allow importing
both v0 and v1 backups, however v0 backups cannot handle the above
described scenario (they can only be used to request a remote-force-close).
We only export v1 backups.

Note that we were/are setting the missing local_payment_basepoint to the pubkey of
one of the wallet change addresses, which is bruteforceable if necessary, but I
think it is not worth the complexity to add this bruteforce logic. Also note
that the bruteforcing could only be done after the local-force-close was broadcast.

Ideally people with existing channels and already exported v0 backups should re-export
v1 backups... Not sure how to handle this.

We could maybe warn people about this when importing v0 backups, but again, not sure
how to cleanly do that, and whether it's worth the complexity... :/

closes https://github.com/spesmilo/electrum/issues/8516

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -184,6 +184,7 @@ class AbstractChannel(Logger, ABC):
     funding_outpoint: Outpoint
     node_id: bytes  # note that it might not be the full 33 bytes; for OCB it is only the prefix
     _state: ChannelState
+    sweep_address: str
 
     def set_short_channel_id(self, short_id: ShortChannelID) -> None:
         self.short_channel_id = short_id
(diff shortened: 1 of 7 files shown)


ecdsa:
Thanks, this PR looks good.

I see that `ChannelBackup` was missing `maybe_sweep_revoked_htlc` and `extract_preimage_from_htlc_txin`, but we never received a traceback about that. Here is how an exception can be raised by missing `maybe_sweep_revoked_htlc`

````diff --git a/electrum/tests/regtest/regtest.sh b/electrum/tests/regtest/regtest.sh
index b3d7ebfca..383b708d0 100755
--- a/electrum/tests/regtest/regtest.sh
+++ b/electrum/tests/regtest/regtest.sh
@@ -172,6 +172,7 @@ fi
 
 
 if [[ $1 == "backup_local_forceclose" ]]; then
+    $bob enable_htlc_settle false
     # Alice does a local-force-close, and then restores from seed before sweeping CSV-locked coins
     wait_for_balance alice 1
     echo "alice opens channel"
@@ -181,6 +182,14 @@ if [[ $1 == "backup_local_forceclose" ]]; then
     new_blocks 3
     wait_until_channel_open alice
     backup=$($alice export_channel_backup $channel)
+    echo "alice pays bob"
+    invoice=$($bob add_request 0.04 -m "test" | jq -r ".lightning_invoice")
+    $alice lnpay $invoice --timeout=1 || true
+    unsettled=$($alice list_channels | jq '.[] | .local_unsettled_sent')
+    if [[ "$unsettled" == "0" ]]; then
+        echo "enable_htlc_settle did not work, $unsettled"
+        exit 1
+    fi
     echo "local force close $channel"
     $alice close_channel $channel --force
     sleep 0.5
```

do we want to add that to the test?

PR #8010, opened 2022-10-10
android build: bump targetSdkVersion from 30 to 31

SomberNight:
Re the Android build, this PR bumps the `targetSdkVersion` from 30 to 31, and upgrades the sdk tools used.
(the google play store will require 31, starting 2022-11 (next month))
A few commits are cherry-picked onto our forks of p4a and buildozer from upstream, but the forks are not rebased.

This was split from the more ambitious https://github.com/spesmilo/electrum/pull/8002, which also tried to rebase the forks.

PR #8545, opened 2023-07-19
Qt6

accumulator:
Currently builds a working APK using buildozer HEAD and python-for-android HEAD plus a patchset for qt6 bootstrap and pyqt6.

Builds against API 31, NDK 23c

Additional notes:
- ~build container name temporarily renamed so it's easier to switch between current qt5 and this qt6 build~
- bootstrap now requires a native qt6 build in order to provide tooling and cmake files for android target. This currently cannot be done using distribution provided qt6. This will increase build time.
- the p4a patchset disables NDK version check, as qt6.4.3 officially supports NDK23, while p4a demands NDK25. I've encountered no obvious problems with using NDK23
- when testing on desktop, note that `Pillow` dependency is upped to 8.4.0 for PyQt6 support

# Issues

- [x] Camera is unstable in Qt 6.4 and 6.5. Resolved for now, by using `SimpleScannerActivity` on android platform.
- [x] UI doesn't always resize correctly in case the keyboard is visible.


contrib/android/Dockerfile
```diff
@@ -179,9 +195,10 @@ RUN cd /opt \
     && git remote add sombernight https://github.com/SomberNight/python-for-android \
     && git remote add accumulator https://github.com/accumulator/python-for-android \
     && git fetch --all \
-    # commit: from branch accumulator/electrum_20210421d (note: careful with force-pushing! see #8162)
-    && git checkout "052b9f7945bae557347fa4a4b418040d9da9eaff^{commit}" \
-    && python3 -m pip install --no-build-isolation --no-dependencies --user -e .
+    # commit: from branch accumulator/qt6-wip (note: careful with force-pushing! see #8162) \
+    #
+    && git checkout "46afc71fc0ba501f351c83cfcaac5504c7c935d6^{commit}" \
```
SomberNight, line 200: > APK versionCode is still a TODO. See p4a commits:
>    [f5440f78ce5f1f2492aa87d845ccc163067aeda4](https://github.com/SomberNight/python-for-android/commit/f5440f78ce5f1f2492aa87d845ccc163067aeda4)
>    [72782ed93de7515965388aff66222aa34dbf7ee4](https://github.com/accumulator/python-for-android/commit/72782ed93de7515965388aff66222aa34dbf7ee4)

see also current commit used for already released versions:
https://github.com/accumulator/python-for-android/commit/3a52f805bec49a6818d527286ce5fd09823f738a
so e.g. 4.4.5(.0) for x86_64, I expect has versionCode `4_4_04_05_00 == 44040500`.

1. the versionCode needs to strictly increase between releases
2. there cannot be a collision of versionCodes between different apks
    - as we don't build multi-arch apks, this means versionCodes need to differ between archs 
3. do we want to build apks with different minSDKs? if not, we can leave that out of the versionCode
4. note that the trailing `.0` in the `APK_VERSION` in `electrum/version.py` has been unused for a long time -- it was only used in the pre-reproducible-builds era. so potentially that could be removed
5. as in https://github.com/kivy/python-for-android/issues/1843, the address space is only 32-bit (not sure if signed or unsigned)
    - any "jump" we make, burns address space we never get back

Disregarding (4), I think [f5440f78ce5f1f2492aa87d845ccc163067aeda4](https://github.com/SomberNight/python-for-android/commit/f5440f78ce5f1f2492aa87d845ccc163067aeda4) is the easiest approach.

SomberNight: oh wait, [f5440f78ce5f1f2492aa87d845ccc163067aeda4](https://github.com/SomberNight/python-for-android/commit/f5440f78ce5f1f2492aa87d845ccc163067aeda4) does not deal with (2) either -- unless we will start building multi-arch apks now

Khademr7 (approved):

contrib/android/buildozer_qml.spec
```diff
@@ -138,11 +139,12 @@ android.add_jars = .buildozer/android/platform/*/build/libs_collections/Electrum
...
 # directory containing the files)
-android.add_src = electrum/gui/kivy/data/java-classes/
+#android.add_src = electrum/gui/kivy/data/java-classes/
 
-android.gradle_dependencies = me.dm7.barcodescanner:zxing:1.9.8
+#android.gradle_dependencies = me.dm7.barcodescanner:zxing:1.9.8
+android.gradle_dependencies = com.android.support:support-compat:28.0.0
 
-android.add_activities = org.electrum.qr.SimpleScannerActivity
+#android.add_activities = org.electrum.qr.SimpleScannerActivity
```
SomberNight, line 147: This now commented out zxing stuff was not used prior to this PR either, right?

SomberNight: Ah I see you clarified this in the relevant commit.
I've pushed just this change separately to master: https://github.com/spesmilo/electrum/commit/c3f16615ca167eb6e63415f84a32592285af5da3

accumulator:
I restored the android version_code from 3a52f805bec49a6818d527286ce5fd09823f738a, so it uses the same version_code system of current releases.

contrib/android/buildozer_qml.spec
accumulator: I restored the behavior from https://github.com/accumulator/python-for-android/commit/3a52f805bec49a6818d527286ce5fd09823f738a, so it is as in current releases.

contrib/android/Dockerfile
accumulator: I restored the behavior of https://github.com/accumulator/python-for-android/commit/3a52f805bec49a6818d527286ce5fd09823f738a, so it is as in current releases.

contrib/android/Dockerfile
```diff
@@ -154,23 +165,29 @@ RUN chown --recursive ${USER} ${WORK_DIR} ${ANDROID_SDK_HOME}
...
     && cd buildozer \
     && git remote add sombernight https://github.com/SomberNight/buildozer \
+    && git remote add accumulator https://github.com/accumulator/buildozer \
     && git fetch --all \
-    # commit: from branch sombernight/electrum_20210421 (note: careful with force-pushing! see #8162)
-    && git checkout "6f03256e8312f8d1e5a6da3a2a1bf06e2738325e^{commit}" \
-    && python3 -m pip install --no-build-isolation --no-dependencies --user -e .
+    # commit: from branch master (note: careful with force-pushing! see #8162) \
+    # no, this is master
+    && git checkout "698b48136c55b2495128207c5e98dccc128bb6c2^{commit}" \
```
SomberNight, line 189: looks like you lost commit https://github.com/SomberNight/buildozer/commit/8791a52a02d1c6279504f402098bf6536824ea1c in a rebase/etc, and then reimplemented it yourself in https://github.com/SomberNight/buildozer/commit/698b48136c55b2495128207c5e98dccc128bb6c2 ? >.<

Mine had some references in the commit message, which might be interesting to track.
Also I have an open PR at https://github.com/kivy/buildozer/pull/1511 though I guess there hasn't been much interest in it (maybe we are doing something weird, not sure how this does not affect more people?).

I don't care which commit we use, but I think we should add a reference in the commit msg to https://github.com/kivy/buildozer/pull/1511.

accumulator: Argh, sorry, I must have force-pushed a older branch when I switched to my laptop >.<
I will repair tomorrow

electrum/gui/qml/java_classes/org/electrum/qr/SimpleScannerActivity.java
```diff
@@ -16,18 +25,53 @@
 import com.google.zxing.Result;
 import com.google.zxing.BarcodeFormat;
 
+import org.electrum.electrum.res.R; // package set in build.gradle
```
SomberNight, line 28: Ah, nice solution in https://github.com/accumulator/python-for-android/commit/eb4a3522373e0b4e2749b8a8bc965ff51355ea35 !

Could I ask that you split https://github.com/spesmilo/electrum/pull/8545/commits/3cabc987fece1387fa4b7e58c45ad4fcd1a10a4f into my changes and yours (e.g. by diffing the current commit with https://github.com/SomberNight/electrum/commit/79ff59e88b8989ba901e57e03f049e9d698eb2de)? So it's visible later in the git blame.

accumulator: Done

contrib/android/buildozer_qml.spec
```diff
@@ -84,7 +85,7 @@ android.permissions = INTERNET, CAMERA, WRITE_EXTERNAL_STORAGE
 
 # (int) Android API to use  (compileSdkVersion)
 # note: when changing, Dockerfile also needs to be changed to install corresponding build tools
-android.api = 30
+android.api = 31
 
 # (int) Android targetSdkVersion
 android.target_sdk_version = 31
```
SomberNight, line 91: Note: the google play store requires target api 33 now.
So ideally we should set `android.api = 33` and comment out `android.target_sdk_version` (The latter was added as a crutch last year as we could not bump `android.api` to 31, which was the new minimum at the time. see https://github.com/spesmilo/electrum/commit/0efc8815a4431e000d86779bb49a360f6f486726 and https://github.com/spesmilo/electrum/pull/8010)

However, we might have to bump the qt version to 6.5 for this. :/

I've downloaded [different NDKs](https://github.com/android/ndk/wiki/Unsupported-Downloads), and it looks like
23(x) that this PR sets and that Qt 6.2-6.4 requires only supports compileSdkVersion up to 31.
25x is the first NDK that supports compileSdkVersion 33.
qt 6.5 is the first version that [uses NDK 25](https://doc.qt.io/qt-6.5/android-getting-started.html#android-ndk).

Alternatively, we might be able to get away with the same trick as last year and set targetSdkVersion higher than compileSdkVersion. I think whether that works depends on the exact changes we have to make to accommodate the runtime checks enforced on the app due to the raised targetSdkVersion.

----

Nevertheless, we can do this in a follow-up to the PR. Upgrading to *some version* of Qt6 is already a huge step towards this.

accumulator: I hope we can avoid Qt6.5, as this version has changes in the styling of the controls, which would force us to tidy up a lot of UI (probably not that difficult, just a big diff)

accumulator: As my main testing device was API33 for months and now recently API34, I think we can safely declare compatibility with targetSdkVersion=33

accumulator:
removing Draft status, as there are currently no open issues and this PR can be reviewed as is.

before considering merging, the build container name change should be removed first.

contrib/android/buildozer_qml.spec
SomberNight: > As my main testing device was API33 for months and now recently API34, I think we can safely declare compatibility with targetSdkVersion=33

AFAIU it should not matter what device you are testing on, as long as it has API >= targetSdkVersion.
As in, the app when running on your device behaves as per `min(targetSdkVersion, installed_android_on_device)`.

When bumping the targetSdkVersion, we should read through the documented changes at https://developer.android.com/about/versions/13/behavior-changes-13 and see if we need to adapt anything. I did that just now and it looks to me we should not be affected.

When we [bumped to 31](https://developer.android.com/google/play/requirements/target-sdk#pre13),  I also read through the corresponding changes, and had to backport some changes from upstream p4a. (e.g. https://github.com/sombernight/python-for-android/commit/b475b97b08e8a64b5d978b971d47a5e81d82d66c)  I even missed something for the notifications, as I think they were not really used for kivy (I think there was some code but disabled): https://github.com/kivy/plyer/pull/756

Anyway, I think we might be ok and can just bump it. :)

accumulator: > AFAIU it should not matter what device you are testing on, as long as it has API >= targetSdkVersion. As in, the app when running on your device behaves as per `min(targetSdkVersion, installed_android_on_device)`.

It does matter what API level you are testing against, precisely because of the behavior changes you already mentioned above, and any further restrictions added by newer API levels. `targetSdkVersion` merely suggests that the package is tested against that level.

On the other end of the spectrum, if a newer API level requires a change, it might need additional code to remain 
[backward compatible](https://developer.android.com/develop/ui/views/search/training/backward-compat) or else bumping the `minSdkVersion` in the manifest.

So, ideally, we need to test against ~every~ API level :)

We are fortunately not using bleeding edge API features, or parts of the API that are moving targets, so we can mostly assume things to keep working (taking the published android version changes into account of course)

SomberNight: > targetSdkVersion merely suggests that the package is tested against that level.

Another way of saying what I meant is that for the testing to count in the first place, you have to test while already having set the higher targetSdkVersion.
Unless I completely misunderstood how it works :D

contrib/android/Dockerfile
```diff
@@ -154,23 +165,29 @@ RUN chown --recursive ${USER} ${WORK_DIR} ${ANDROID_SDK_HOME}
...
     && git clone https://github.com/kivy/buildozer \
     && cd buildozer \
     && git remote add sombernight https://github.com/SomberNight/buildozer \
+    && git remote add accumulator https://github.com/accumulator/buildozer \
     && git fetch --all \
-    # commit: from branch sombernight/electrum_20210421 (note: careful with force-pushing! see #8162)
-    && git checkout "6f03256e8312f8d1e5a6da3a2a1bf06e2738325e^{commit}" \
-    && python3 -m pip install --no-build-isolation --no-dependencies --user -e .
+    # commit: from branch master (note: careful with force-pushing! see #8162) \
+    # no, this is master
```
SomberNight, line 188: I guess it is not master afterall! :P

electrum/gui/qml/android_res/layout/scanner_layout.xml
```diff
@@ -0,0 +1,29 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+
+    <TextView
+        android:id="@+id/hint"
+        android:layout_gravity="center|top"
+        android:text="Scan a QR code."
+        android:layout_width="wrap_content"
+        android:textColor="#ffffff"
+        android:layout_height="wrap_content" />
```
SomberNight, line 20: I made this look nicer in https://github.com/sombernight/electrum/commit/638f2d885cd9fb4dae47012d59dc0dbad420f838

Compare:
<details>
 <summary>before</summary>

Screenshot_20231031_190414_Electrum
</details>

<details>
 <summary>after</summary>

Screenshot_20231031_190948_Electrum
</details>

contrib/android/Dockerfile
accumulator: it was for a long time, but good catch :)

electrum/gui/qml/android_res/layout/scanner_layout.xml
accumulator: Looks much nicer indeed. Feel free to just add to this branch directly

contrib/android/buildozer_qml.spec
accumulator: tested with compileSdk 31, targetSdk 33, on SDK 34 device.
We should be good.

accumulator:
rebased and removed the commit changing the qt6 specific build image name, so this PR should be ready to merge.

SomberNight (approved):
I have tried to review and test this now (at commit 047cc6f9579f17647cefe20ec05fdad9dafd0457), and to me it looks good.
Thank you very much for this work.

PR #8546, opened 2023-07-20
Add Vendor/Device IDs for CH340 based DIY Jade devices.

3rdIteration:
Allows use with DIY Jade + Camera based on common, low cost boards like those based on ESP32-Wrover and ESP32-CAM.

### DIFF
--- a/electrum/plugins/jade/jade.py
+++ b/electrum/plugins/jade/jade.py
@@ -350,7 +350,10 @@ class Jade_KeyStore(Hardware_KeyStore):
 class JadePlugin(HW_PluginBase):
     keystore_class = Jade_KeyStore
     minimum_library = (0, 0, 1)
-    DEVICE_IDS = [(0x10c4, 0xea60), (0x1a86, 0x55d4), (0x0403, 0x6001), (0x1a86, 0x7523)]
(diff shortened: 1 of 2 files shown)


electrum/plugins/jade/jade.py
```diff
@@ -350,7 +350,7 @@ def show_address_multi(self, multisig_name, paths):
 class JadePlugin(HW_PluginBase):
     keystore_class = Jade_KeyStore
     minimum_library = (0, 0, 1)
-    DEVICE_IDS = [(0x10c4, 0xea60), (0x1a86, 0x55d4), (0x0403, 0x6001)]
+    DEVICE_IDS = [(0x10c4, 0xea60), (0x1a86, 0x55d4), (0x0403, 0x6001), (0x1a86, 0x7523)]
```
SomberNight, line 353: Could you please add some newlines and comments to help distinguish which code is for what kind of device?
Something like this:
https://github.com/spesmilo/electrum/blob/827792c14cb11cfca7c8c91af2724c8b15fa2df4/electrum/plugins/ledger/ledger.py#L1323-L1335

3rdIteration: You are right, it certainly gets a bit unwieldly without that. Have added them in. :)

PR #8560, opened 2023-08-01
new wizard Qt desktop client

accumulator:
Refactor of wizard to new wizard framework (as in qml).

Notes:
- For multisig Qt client shows master pubkey on cosigner keystore select page, not on separate page.
- inserts extra separate pages for passphrase/ext entry in the base wizard flow, to mimic old wizard. 
- script and derivation path entry shown after HWW select but before device access/unlock

Todo:
- [ ] hw wallets
    - [x] trezor
        - [x] device setup
        - [x] test
    - [x] jade
        - [x] device setup
        - [x] test
    - [x] digital bitbox
        - [x] device setup
        - [x] test
    - [x] bitbox02
        - [x] test
    - [x] coldcard
        - [ ] test
    - [x] ledger
        - [x] test
    - [x] safe_t
        - [x] device setup
        - [x] test
    - [x] keepkey
        - [x] device setup
        - [x] test


electrum/gui/qt/wizard/wizard.py
```diff
@@ -0,0 +1,233 @@
...
+from electrum.i18n import _
+from ..util import Buttons, icon_path
+from electrum.logging import get_logger
+
+if TYPE_CHECKING:
+    from electrum.simple_config import SimpleConfig
+
+
+class QEAbstractWizard(QDialog):
+    _logger = get_logger(__name__)
```
SomberNight, line 18: I think it is nicer to inherit Logger, e.g. `QEAbstractWizard(QDialog, Logger)`, and then
```py
def __init__(self, config: 'SimpleConfig', app: QApplication, daemon):
    Logger.__init__(self)
```

Especially as `QEAbstractWizard` is not a singleton, it could be useful for debugging to have separate instances of it each their own logger. E.g. look at wallet.py/Abstract_Wallet: note it is overriding Logger.diagnostic_name. -- Though if you don't override diagnostic_name, the choice of having a per-class/per-instance logger becomes less important.

I guess this remark also applies to a lot of existing code in the qml GUI.
Anyway, it's more of a nit.

accumulator: Thanks. I guess this stems from my past experience with C++ and Java, where inheritance is more rigid. Pythonic multiple inheritance and mixins still feel a bit 'loose' to me :)

accumulator: BTW, `diagnostic_name` might not be unique unless I use the `id()` of the wizard instance.

SomberNight: > I'm used to loggers being a composition into a class


You are right that inheritance should not be overused.
Hmm, I guess it is mostly more consistent usage of the Logger across the codebase that I wanted, not necessarily inheritance.
We could change the Logger class to be used with composition instead. `diagnostic_name` and `LOGGING_SHORTCUT` could be passed as parameters to `Logging.__init__`,
and then do e.g. `self.logger = Logger(name=self.diagnostic_name(), shortcut=LOGGING_SHORTCUT)` in init.


> `diagnostic_name` might not be unique unless I use the `id()` of the wizard instance

Yes, indeed. See `lnpeer.Peer` for an example where we explicitly use `id()` (through the transport) -- it was added during/after debugging an issue where multiple Peer objects did have the same colliding diagnostic names which was causing confusion.

---

Anyway, it's just a small remark; not too important re this PR.

accumulator:
@ecdsa, @SomberNight I've started adding HWW support for devices I can't test. 
I've added test todos in the top post. Can you please check these devices for 
- [ ] proper handling when HWW uninitialized
- [ ] can use to create standard wallet
- [ ] can use for wallet encrypt by hardware

accumulator:
Removing draft status. Not all HWW that currently support initial device setup are ported, but it's ready to test.

electrum/gui/qt/__init__.py
```diff
@@ -398,6 +403,72 @@ def start_new_window(
...
+            db.put('x3/', wizard.get_wizard_data()['x3/'])
+            db.write()
+
+        wallet = Wallet(db, config=self.config)
+        wallet.start_network(self.daemon.network)
+        self.daemon.add_wallet(wallet)
+        return wallet
+
+
+    def _start_wizard_to_select_or_create_wallet2(self, path) -> Optional[Abstract_Wallet]:
```
ecdsa, line 471: is this unused?

ecdsa:
Can we rename `qt_common` to something else? I am a heavy user of tab-completions, and this new directory is slowing me down significantly; I now need to type "q + t + shift + slash" in order to have a non-ambiguous prefix, where "q + t" used to be sufficient. (before the introduction of the `qml` directory, the sole letter "q" was enough).

accumulator:
> Can we rename `qt_common` to something else? I am a heavy user of tab-completions, and this new directory is slowing me down significantly; I now need to type "q + t + shift + slash" in order to have a non-ambiguous prefix, where "q + t" used to be sufficient. (before the introduction of the `qml` directory, the sole letter "q" was enough).

Done

electrum/gui/qt/__init__.py
accumulator: Yes, I used this to quickly switch between old wizard and new wizard entry. I'll remove it.

electrum/gui/qt/__init__.py
```diff
@@ -398,26 +403,65 @@ def start_new_window(
...
+                return
+            wallet_file = wizard.path
+        else:
+            wallet_file = d['wallet_name']
+
+        storage = WalletStorage(wallet_file)
+        if storage.is_encrypted_with_user_pw() or storage.is_encrypted_with_hw_device():
+            storage.decrypt(d['password'])
+
+        db = WalletDB(storage.read(), storage=storage, manual_upgrades=True)
```
ecdsa, line 435: Actually, `WalletDB` is called in daemon.py with `manual_upgrades=True`
This line is reached only after the user has clicked 'next' in the wizard first screen. 
Thus, `manual_upgrades` was correctly set to `False` in master, and should remain so.

electrum/gui/qt/__init__.py
```diff
@@ -398,26 +403,65 @@ def start_new_window(
...
+        if db.requires_split() or db.requires_upgrade():
+            try:
+                wizard.run_upgrades(db)
+            except UserCancelled:
+                return
+
+        if action := db.get_action():
+            # wallet creation is not complete, 2fa online phase
```
ecdsa, line 443: action handling used to be performed in `run_upgrades`
- why did you move this code here?
- if you move it here, why is the old action handling code not removed from `run_upgrades`?
- is saving and resuming the creation of a multisig wallet (not trustedcoin) supported?

electrum/gui/qt/__init__.py
ecdsa: edit: I realize that run_wizard from installwizard.py is not used anymore, the code has moved

electrum/gui/qt/__init__.py
accumulator: > action handling used to be performed in `run_upgrades`
> 
>     * why did you move this code here?

I need to initialize QENewWalletWizard with the necessary data

>     * if you move it here, why is the old action handling code not removed from `run_upgrades`?

That was leftover old wizard code, which is now removed

>     * is saving and resuming the creation of a multisig wallet (not trustedcoin) supported?

No. Unless that was well hidden I don't think it was implemented for the old wizard :)

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1322 @@
...
+                    if wallet_from_memory:
+                        temp_storage = wallet_from_memory.storage  # type: Optional[WalletStorage]
+                        self.wallet_is_open = True
+                    else:
+                        temp_storage = WalletStorage(path)
+                    self.wallet_exists = temp_storage.file_exists()
+                except (StorageReadWriteError, WalletFileException) as e:
+                    msg = _('Cannot read file') + f'\n{repr(e)}'
+                except Exception as e:
+                    self.logger.exception('')
```
SomberNight, line 297: `self.logger` looks undefined. Probably in some other instances too. In some cases it's a `_logger` vs `logger` issue, but here neither is defined I think.
Maybe we should standardise on one, between `_logger` and `logger`.

electrum/gui/qt/__init__.py
```diff
@@ -398,26 +401,65 @@ def start_new_window(
...
+            wizard.create_storage()
+            if d['wallet_type'] == '2fa' and 'x3/' not in d:
+                return
+            wallet_file = wizard.path
+        else:
+            wallet_file = d['wallet_name']
+
+        storage = WalletStorage(wallet_file)
+        if storage.is_encrypted_with_user_pw() or storage.is_encrypted_with_hw_device():
+            storage.decrypt(d['password'])
```
SomberNight, line 431: `InvalidPassword` should be handled somewhere
```
 28.11 | E | gui.qt.ElectrumGui | 
  File "...\electrum\electrum\gui\qt\__init__.py", line 362, in start_new_window
    wallet = self._start_wizard_to_select_or_create_wallet(path)
  File "...\electrum\electrum\gui\qt\__init__.py", line 431, in _start_wizard_to_select_or_create_wallet
    storage.decrypt(d['password'])
  File "...\electrum\electrum\storage.py", line 172, in decrypt
    s = zlib.decompress(ec_key.decrypt_message(self.raw, enc_magic))
  File "...\electrum\electrum\ecc.py", line 546, in decrypt_message
    raise InvalidPassword()
electrum.util.InvalidPassword: Incorrect password
```

SomberNight: Note: applies to both mistyped user pw, and mismatching hardware device cases

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1322 @@
...
+            self.busy = False
+            self.validate()
+
+        t = threading.Thread(target=get_xpub_task, args=(client, derivation, xtype), daemon=True)
+        t.start()
+
+    def get_xpub_from_client(self, client, derivation, xtype):  # override for HWW specific client if needed
+        _name, _info = self.wizard_data['hardware_device']
+        if xtype not in self.plugin.SUPPORTED_XTYPES:
+            raise ScriptTypeNotSupported(_('This type of script is not supported with {}').format(_info.model_name))
```
SomberNight, line 1290: You use `DeviceInfo.model_name` in multiple user-facing strings. Note that that field is often None.
Look at `HardwareClientBase.device_model_name`: it is only implemented by trezor, jade, and ledger.
Plugin-agnostic code on master IIRC often uses `device_info.model_name or plugin_name`, see e.g. 
https://github.com/spesmilo/electrum/blob/ddad5f1320993a49327267bf790b485182f13ec0/electrum/base_wizard.py#L355

Maybe the `DeviceInfo` class could have a helper method for this:
`device_info.model_name or device_info.plugin_name or "unknown"`, that always returns a string.

electrum/gui/qt/wizard/wallet.py
accumulator: For the `WizardComponent` descendants I used Logger.
For `QENewWalletWizard` that approach still has an issue.

electrum/gui/qt/wizard/wallet.py
accumulator: I'm in favor of mandating plugins to define this field. In absence, we can initialize it to the plugin name. This approach would let us assume it is always present, avoiding a repeat of `or device_info.plugin_name` at every use of the field.

electrum/plugins/trezor/qt.py
```diff
@@ -767,3 +793,109 @@ def slider_released():
...
+
+class WCTrezorInitMethod(WizardComponent, Logger):
+    def __init__(self, parent, wizard):
+        WizardComponent.__init__(self, parent, wizard, title=_('HW Setup'))
+        Logger.__init__(self)
+        self.plugin = None
+
+    def on_ready(self):
+        _name, _info = self.wizard_data['hardware_device']
+        self.plugin = self.plugins.get_plugin(_info.plugin_name)
```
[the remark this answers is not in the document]
accumulator: thx, fixed.

electrum/gui/qt/wizard/wallet.py
accumulator: For now I've done the quick fix in ae8c6dfb6347d98d293b213d71d8014b51028cd4, which adds static device names for all HWW. The method can still return None in case a new HWW doesn't define the method, but we now at least get the hint in the IDE that an abstract method is not overridden.

electrum/gui/qt/__init__.py
accumulator: added in 7940c0725ebfb17b527cc7b95311a76364444740

electrum/gui/qt/wizard/wallet.py
accumulator: added plugin name as default impl for device name

ecdsa:
I made a few tests with trezor. Compared to the old wizard, I think there are unnecessary steps:
 - the "rescan devices" button is enabled by default, instead of "next", even if a Trezor is detected. This forces me to use my mouse, or to press tab.
 - the next screen "unlocking trezor one" concludes by "Hardware successfully unlocked". While this sounds like good news, why on earth do I need to click on "next" in that screen? the previous wizard used to open the wallet directly. This is asked not only on wallet creation, but every time I open the wallet.

electrum/gui/qt/wizard/wallet.py
accumulator: Ah I wasn't aware this would/should raise. I'll remove it then, and just rely on a default impl of the method.

accumulator: changed in 8490a2868889e02323c35b8b65c367ebaf0bbc67
I replaced the raise `NotImplementedError`s with `@abstractmethod`  decorators for the methods that must have a subclass definition.

ecdsa:
@accumulator It looks like you have removed the rescan functionality completely. I suppose I was unclear.

Please look at how it works on master: The dialog displays "to trigger a rescan, press 'Next'".
Initially, this PR added an extra 'Rescan' button, and replaced that message with "To trigger a rescan, press 'Rescan'".

That extra button was useless, and it forced the user to select the right button.
This is why I asked you to remove it, and to revert to using the 'Next' button for triggering a rescan, like on master. 

Now the next button is grayed out, it is no longer possible to trigger a rescan.

ecdsa:
note: I tested with Ledger, it seems to work fine

accumulator:
> @accumulator It looks like you have removed the rescan functionality completely. I suppose I was unclear.
> 
> Please look at how it works on master: The dialog displays "to trigger a rescan, press 'Next'". Initially, this PR added an extra 'Rescan' button, and replaced that message with "To trigger a rescan, press 'Rescan'".
> 
> That extra button was useless, and it forced the user to select the right button. This is why I asked you to remove it, and to revert to using the 'Next' button for triggering a rescan, like on master.
> 
> Now the next button is grayed out, it is no longer possible to trigger a rescan.

Doing it this way is not really compatible with how the new wizard works, as I would have to redefine the Next button behaviour of the wizard from within a wizard component. 

Also I always found that 'press Next to rescan' a UX oddity, not something I'd thought we want to replicate.

Eventually (not in this PR) I was thinking of making the device list more an 'active' component, detecting devices as they are plugged in, not 'passive', as in having to explicitly rescan.

ecdsa:
> 
> Doing it this way is not really compatible with how the new wizard works, as I would have to redefine the Next button behaviour of the wizard from within a wizard component.
> 

So the extra scan button is imposed by the new wizard code?  It would have saved some time if you had explained that earlier, because I had the impression that you were not understanding what I was asking...

I guess we can accept that in that case, because it is not a big issue, and it seems to be difficult to implement. However, if you see a way to adapt the code and remove that button, please do it. In general, I think that code should be adapted to user experience, and not the other way round.

electrum/plugins/keepkey/qt.py
```diff
@@ -310,6 +321,23 @@ def pin_matrix_widget_class(self):
...
+
+    # insert keepkey pages in new wallet wizard
+    def extend_wizard(self, wizard: 'QENewWalletWizard'):
+        super().extend_wizard(wizard)
+        views = {
+            'keepkey_start': {'gui': WCScriptAndDerivation},
+            'keepkey_xpub': {'gui': WCHWXPub},
+            'safet_not_initialized': {'gui': WCKeepkeyInitMethod},
+            'safet_choose_new_recover': {'gui': WCKeepkeyInitParams},
+            'safet_do_init': {'gui': WCKeepkeyInit},
```
SomberNight, line 336: `safet`->`keepkey`? copy-paste

ecdsa:
after initializing keepkey:

```
  File "/opt/electrum/electrum/gui/qt/wizard/wizard.py", line 200, in on_next_button_clicked
    self.load_next_component(next.view, next.wizard_data, next.params)
  File "/opt/electrum/electrum/gui/qt/wizard/wizard.py", line 132, in load_next_component
    comp = self.view_to_component(view)
  File "/opt/electrum/electrum/gui/qt/wizard/wizard.py", line 207, in view_to_component
    return self.navmap[view]['gui']
KeyError: 'gui'

```

ecdsa:
a few crashes with keepkey: 

```
diff --git a/electrum/plugins/keepkey/qt.py b/electrum/plugins/keepkey/qt.py
index acebce3f1..01beff84c 100644
--- a/electrum/plugins/keepkey/qt.py
+++ b/electrum/plugins/keepkey/qt.py
@@ -231,6 +231,7 @@ class KeepkeyInitLayout(QVBoxLayout):
     validChanged = pyqtSignal([bool], arguments=['valid'])
 
     def __init__(self, method, device):
+        QVBoxLayout.__init__(self)
         self.method = method
 
         label = QLabel(_("Enter a label to name your device:"))
@@ -257,8 +258,8 @@ class KeepkeyInitLayout(QVBoxLayout):
                 self.bg.setId(rb, i)
                 hbox1.addWidget(rb)
                 rb.setChecked(True)
-            cb_pin = QCheckBox(_('Enable PIN protection'))
-            cb_pin.setChecked(True)
+            self.cb_pin = QCheckBox(_('Enable PIN protection'))
+            self.cb_pin.setChecked(True)
         else:
             self.text_e = QTextEdit()
             self.text_e.setMaximumHeight(60)
@@ -285,7 +286,7 @@ class KeepkeyInitLayout(QVBoxLayout):
 
         if method in [TIM_NEW, TIM_RECOVER]:
             self.addWidget(WWLabel(RECOMMEND_PIN))
-            self.addWidget(cb_pin)
+            self.addWidget(self.cb_pin)
         else:
             self.addLayout(hbox_pin)
 
@@ -331,9 +332,9 @@ class Plugin(KeepKeyPlugin, QtPlugin):
         views = {
             'keepkey_start': {'gui': WCScriptAndDerivation},
             'keepkey_xpub': {'gui': WCHWXPub},
-            'safet_not_initialized': {'gui': WCKeepkeyInitMethod},
-            'safet_choose_new_recover': {'gui': WCKeepkeyInitParams},
-            'safet_do_init': {'gui': WCKeepkeyInit},
+            'keepkey_not_initialized': {'gui': WCKeepkeyInitMethod},
+            'keepkey_choose_new_recover': {'gui': WCKeepkeyInitParams},
+            'keepkey_do_init': {'gui': WCKeepkeyInit},
             'keepkey_unlock': {'gui': WCHWUnlock}
         }
         wizard.navmap_merge(views)
@@ -640,7 +641,7 @@ class WCKeepkeyInitParams(WizardComponent):
 
     def on_ready(self):
         _name, _info = self.wizard_data['hardware_device']
-        self.settings_layout = KeepkeyInitLayout(self.plugins.device_manager, self.wizard_data['keepkey_init'], _info.device.id_)
+        self.settings_layout = KeepkeyInitLayout(self.wizard_data['keepkey_init'], _info.device.id_)
         self.layout().addLayout(self.settings_layout)
         self.layout().addStretch(1)

```

electrum/plugins/hw_wallet/plugin.py
```diff
@@ -276,10 +257,11 @@ def get_password_for_storage_encryption(self) -> str:
     def device_model_name(self) -> Optional[str]:
         """Return the name of the model of this device, which might be displayed in the UI.
         E.g. for Trezor, "Trezor One" or "Trezor T".
+        If this method is not defined for a plugin, the plugin name is used as default
         """
-        return None
+        return self.plugin.name
 
-    def manipulate_keystore_dict_during_wizard_setup(self, d: dict) -> None:
+    def manipulate_keystore_dict_during_wizard_setup(self, d: dict) -> None:  # noqa: B027
```
SomberNight, line 264: It is suspicious that this method is only called from base_wizard.py, and not from the new wizard.

electrum/plugins/hw_wallet/plugin.py
```diff
@@ -379,3 +361,7 @@ def text_ignore_old_fw_and_continue(self) -> str:
             return str(self) + "\n\n" + suffix
         else:
             return suffix
+
+
+class OperationCancelled(UserFacingException):
```
SomberNight, line 366: How does this differ from `UserCancelled`? If we need both, maybe add a comment here.

electrum/keystore.py
```diff
@@ -58,6 +58,8 @@
 
 class CannotDerivePubkey(Exception): pass
 
+class ScriptTypeNotSupported(Exception): pass
```
SomberNight, line 61: This is already defined in base_wizard.py.
Speaking of; can base_wizard.py be deleted?
How about all the overrides of the old wizard methods in trustedcoin.py?

electrum/gui/qt/installwizard.py
```diff
@@ -1,797 +0,0 @@
...
-            options.append('bip39')
-        if self.opt_slip39:
-            options.append('slip39')
-        title = _('Enter Seed')
-        message = _('Please enter your seed phrase in order to restore your wallet.')
-        return self.seed_input(title, message, test, options)
-
-    @wizard_dialog
-    def confirm_seed_dialog(self, run_next, seed, test):
-        self.app.clipboard().clear()
```
SomberNight, line 505: This behaviour should be replicated.

electrum/plugins/trezor/qt.py
```diff
@@ -376,65 +385,65 @@ def show_expert_settings():
...
+            self.bg_rectype.setId(rb2, RecoveryDeviceType.Matrix)
             hbox_rectype.addWidget(rb2)
 
         # no backup
-        cb_no_backup = None
+        self.cb_no_backup = None
         if method == TIM_NEW:
-            cb_no_backup = QCheckBox(f'''{_('Enable seedless mode')}''')
-            cb_no_backup.setChecked(False)
+            self.cb_no_backup = QCheckBox(f'''{_('Enable seedless mode')}''')
```
SomberNight, line 426: Hmm, it's not related to the PR, but what is going on here... Looks like I wrote this some years ago >_<
Let's just do `QCheckBox(_('Enable seedless mode'))`

electrum/plugins/keepkey/qt.py
accumulator: weird.. I'm pretty sure I fixed this, as well as ecdsa's patch below. Must be git stashed instead of committed.

electrum/plugins/hw_wallet/plugin.py
accumulator: This is only used for coldcard, which isn't fully ported yet.

electrum/keystore.py
accumulator: I avoided referencing `base_wizard.py` as that file is planned to become unused as soon as the kivy client is removed. `trustedcoin.py` still has some leftovers which I will remove.

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1364 @@
...
+                else:
+                    msg = _("This file is already open in memory.") + "\n" \
+                          + _("Press 'Finish' to create/focus window.")
+            if msg is None:
+                msg = _('Cannot read file')
+            msg_label.setText(msg)
+            widget_create_new.setVisible(bool(temp_storage and temp_storage.file_exists()))
+            if user_needs_to_enter_password:
+                pw_label.show()
+                self.pw_e.show()
```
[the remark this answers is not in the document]
accumulator: Ah, there's two scenarios here. 
1) opening electrum when the default wallet is password encrypted or selecting an existing wallet via file picker
2) typing a (new/existing) wallet name

If you have e.g. a wallet named `test` and you're trying to enter `test_2`, your editing is interrupted by a focus change as soon as `test` is entered.

I've added a conditional setFocus for scenario 1.

accumulator:
> A few crashes with keepkey (after reinitializing the device). The following diff fixes them, but is not enough: the next button is grayed out on the last screen.

Should be handled correctly now.

ecdsa:
> > A few crashes with keepkey (after reinitializing the device). The following diff fixes them, but is not enough: the next button is grayed out on the last screen.
> 
> Should be handled correctly now.

thanks. it works for me now.

electrum/keystore.py
accumulator: trustedcoin old wizard leftovers removed.

electrum/gui/qt/__init__.py
accumulator: manual_upgrades=True is just making sure no implicit upgrade is happening in the constructor. As I am here just using WalletDB to test for requires_split() and requires_upgrade() and eventually deferring the actual upgrade to wizard.run_upgrades() where I can show an in progress dialog.

electrum/plugins/hw_wallet/plugin.py
```diff
@@ -205,9 +183,14 @@ def can_recognize_device(self, device: Device) -> bool:
         """
         return device.product_key in self.DEVICE_IDS
 
+    @abstractmethod
```
SomberNight, line 186: decorator is no-op without inheriting from `ABC` class

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1384 @@
...
+        else:
+            self.rescan_button.setFocus()
+
+
+    def failed_getting_device_infos(self, debug_msg, name, e):
+        # nonlocal debug_msg
+        err_str_oneline = ' // '.join(str(e).splitlines())
+        self.logger.warning(f'error getting device infos for {name}: {err_str_oneline}')
+        indented_error_msg = '    '.join([''] + str(e).splitlines(keepends=True))
+        debug_msg += f'  {name}: (error getting device infos)\n{indented_error_msg}\n'
```
SomberNight, line 1142: debug_msg is not getting side-effected correctly. Only the logging is visible to the outside atm.
(Just insert some artificial exceptions somewhere to test. -- I saw some organically)
Maybe something like this could work:
```patch
diff --git a/electrum/gui/qt/wizard/wallet.py b/electrum/gui/qt/wizard/wallet.py
index 6169f9d618..016e4f1184 100644
--- a/electrum/gui/qt/wizard/wallet.py
+++ b/electrum/gui/qt/wizard/wallet.py
@@ -1134,12 +1134,7 @@ class WCChooseHWDevice(WizardComponent, Logger):
             self.rescan_button.setFocus()
 
 
-    def failed_getting_device_infos(self, debug_msg, name, e):
-        # nonlocal debug_msg
-        err_str_oneline = ' // '.join(str(e).splitlines())
-        self.logger.warning(f'error getting device infos for {name}: {err_str_oneline}')
-        indented_error_msg = '    '.join([''] + str(e).splitlines(keepends=True))
-        debug_msg += f'  {name}: (error getting device infos)\n{indented_error_msg}\n'
+
 
     def scan_devices(self):
         self.valid = False
@@ -1153,6 +1148,13 @@ class WCChooseHWDevice(WizardComponent, Logger):
             devmgr = self.plugins.device_manager
             debug_msg = ''
 
+            def failed_getting_device_infos(name, e):
+                nonlocal debug_msg
+                err_str_oneline = ' // '.join(str(e).splitlines())
+                self.logger.warning(f'error getting device infos for {name}: {err_str_oneline}')
+                indented_error_msg = '    '.join([''] + str(e).splitlines(keepends=True))
+                debug_msg += f'  {name}: (error getting device infos)\n{indented_error_msg}\n'
+
             # scan devices
             try:
                 # scanned_devices = self.run_task_without_blocking_gui(task=devmgr.scan_devices,
@@ -1178,15 +1180,15 @@ class WCChooseHWDevice(WizardComponent, Logger):
                         device_infos = devmgr.list_pairable_device_infos(
                             handler=None, plugin=plugin, devices=scanned_devices, include_failing_clients=True)
                     except HardwarePluginLibraryUnavailable as e:
-                        self.failed_getting_device_infos(debug_msg, name, e)
+                        failed_getting_device_infos(name, e)
                         continue
                     except BaseException as e:
                         self.logger.exception('')
-                        self.failed_getting_device_infos(debug_msg, name, e)
+                        failed_getting_device_infos(name, e)
                         continue
                     device_infos_failing = list(filter(lambda di: di.exception is not None, device_infos))
                     for di in device_infos_failing:
-                        self.failed_getting_device_infos(debug_msg, name, di.exception)
+                        failed_getting_device_infos(name, di.exception)
                     device_infos_working = list(filter(lambda di: di.exception is None, device_infos))
                     devices += list(map(lambda x: (name, x), device_infos_working))
             if not debug_msg:
```

accumulator: Thanks, committed.

electrum/plugins/hw_wallet/plugin.py
accumulator: Thx, committed

electrum/wizard.py
```diff
@@ -351,18 +408,102 @@ def keystore_from_data(self, wallet_type, data):
...
+        else:
+            raise Exception(f'unknown seed variant {seed_variant}')
+
+        # check if seed matches wallet type
+        if wallet_type == '2fa' and not is_any_2fa_seed_type(seed_type):
+            seed_valid = False
+        elif wallet_type == 'standard' and seed_type not in ['old', 'standard', 'segwit', 'bip39', 'slip39']:
+            seed_valid = False
+        elif wallet_type == 'multisig' and seed_type not in ['standard', 'segwit', 'bip39']:
+            seed_valid = False
```
SomberNight, line 500: why no `slip39` nesting in multisig? :P
I think the old wizard allows it.

accumulator: I'll have a look

accumulator: probably works already, and the storage creation code is just being conservative.

electrum/plugins/trezor/qt.py
```diff
@@ -767,3 +793,110 @@ def slider_released():
...
+            self.init_done()
+
+        t = threading.Thread(
+            target=initialize_device_task,
+            args=(settings, method, device_id, client.handler),
+            daemon=True)
+        t.start()
+
+    def init_done(self):
+        self.logger.info('Done initialize device')
```
SomberNight, line 898: I am testing with a trezor one, uninitialized.
Regardless of generating a new random seed, or restoring from existing seed, the wizard gets stuck after last word, just showing "done". The next button is disabled.
pic2

note: does not affect keepkey

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1382 @@
...
+                self.xpub = self.get_xpub_from_client(client, derivation, xtype)
+                self.root_fingerprint = client.request_root_fingerprint_from_device()
+                self.label = client.label()
+                self.soft_device_id = client.get_soft_device_id()
+            except UserFacingException as e:
+                self.error = str(e)
+                self.logger.error(repr(e))
+            except Exception as e:
+                self.error = repr(e)  # TODO: handle user interaction exceptions (e.g. invalid pin) more gracefully
+                self.logger.error(repr(e))
```
SomberNight, line 1334: ah was about to comment about this, but I see there is already this TODO.
With trezor1 and keepkey, on invalid PIN, the screen looks like this:
pic3

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1382 @@
...
+                                      allow_multi=False, config=self.wizard.config)
+            self.layout().addLayout(self.slayout)
+
+    def apply(self):
+        text = self.slayout.get_text()
+        cosigner_data = self.wizard.current_cosigner(self.wizard_data)
+        cosigner_data['master_key'] = text
+
+
+class WCMultisig(WizardComponent):
```
SomberNight, line 870: Looks like it is not possible to create a wallet that contains multiple `Hardware_KeyStore`s.
If I try to create a 2of2 wallet from a trezor and a ledger, one of the keystores is going to be a software xpub.
This was not the case in the old wizard.

electrum/plugins/keepkey/qt.py
```diff
@@ -272,30 +285,29 @@ def set_enabled():
...
+
+    def get_settings(self):
+        if self.method in [TIM_NEW, TIM_RECOVER]:
+            item = self.bg.checkedId()
+            pin = self.cb_pin.isChecked()
         else:
-            item = ' '.join(str(clean_text(text)).split())
-            pin = str(pin.text())
+            item = ' '.join(str(clean_text(self.text_e)).split())
+            pin = str(self.pin.text())
```
SomberNight, line 308: Testing with an uninitialized keepkey. When using `TIM_MNEMONIC`  or `TIM_PRIVKEY` to restore, getting:
```
  File "...\electrum\electrum\gui\qt\wizard\wizard.py", line 151, in on_page_updated
	page.apply()
  File "...\electrum\electrum\plugins\keepkey\qt.py", line 656, in apply
	self.wizard_data['keepkey_settings'] = self.settings_layout.get_settings()
  File "...\electrum\electrum\plugins\keepkey\qt.py", line 308, in get_settings
	pin = str(self.pin.text())
AttributeError: 'KeepkeyInitLayout' object has no attribute 'pin'
```

electrum/plugins/keepkey/keepkey.py
```diff
@@ -198,52 +196,8 @@ def get_client(self, keystore, force_pair=True, *,
...
-        except BaseException as e:
-            self.logger.exception('')
-            handler.show_error(repr(e))
-            exit_code = 1
-        finally:
-            wizard.loop.exit(exit_code)
-
     @runs_in_hwd_thread
-    def _initialize_device(self, settings, method, device_id, wizard, handler):
+    def _initialize_device(self, settings, method, device_id, handler):
```
SomberNight, line 200: Restoring with TIM_RECOVER is broken. The value of `word_count` being passed used to be 24, now it is 6 (item is `-1` now). The value does not matter, but it has to be one of {12, 18, 24}.
```patch
diff --git a/electrum/plugins/keepkey/keepkey.py b/electrum/plugins/keepkey/keepkey.py
index ecbb0444e8..5d81b13a37 100644
--- a/electrum/plugins/keepkey/keepkey.py
+++ b/electrum/plugins/keepkey/keepkey.py
@@ -211,7 +211,7 @@ class KeepKeyPlugin(HW_PluginBase):
             client.reset_device(True, strength, passphrase_protection,
                                 pin_protection, label, language)
         elif method == TIM_RECOVER:
-            word_count = 6 * (item + 2)  # 12, 18 or 24
+            word_count = 24  # looks like this value is ignored by the device, but it has to be one of {12,18,24}
             client.step = 0
             client.recovery_device(word_count, passphrase_protection,
                                        pin_protection, label, language)

```

SomberNight (commented):
I've tested with a coldcard, trezor1, trezorT, keepkey now.
coldcard all ok. TrezorT all ok. Trezor1 mostly ok, see comments.
keepkey when initialized already, ok. Otherwise see comments.

electrum/plugins/trezor/qt.py
ecdsa: I hit exactly the same issue with keepkey, see this comment: https://github.com/spesmilo/electrum/pull/8560#issuecomment-1717968858
Apparently it was fixed for keepkey, but not for other devices?

accumulator: Should be fixed in b6fdd128a1e49bff1dfa10223be3f46e6747a354

electrum/gui/qt/wizard/wallet.py
accumulator: These exceptions are coming from the device client libs. It would be unwieldy to catch them all within `WCHWXPub` class as it is a generic component used for all devices, so I think the better approach is that the (device specific) device client class catches these low level exceptions and re-raise them within a `UserFacingException`.

electrum/plugins/keepkey/qt.py
accumulator: fixed in ec4c2487b810cfe8a13b709fc399458f1d29cbce

electrum/plugins/keepkey/keepkey.py
accumulator: This might have already been broken. The changes in this PR don't really change the handling of the radiobuttons

electrum/gui/qt/wizard/wallet.py
accumulator: fixed in 447a684d1876aaa637c67608ee6f2126e0c70110

electrum/gui/qt/wizard/wizard.py
```diff
@@ -0,0 +1,287 @@
...
+
+        self.start_viewstate = start_viewstate
+
+        self.show()
+        self.raise_()
+
+        QTimer.singleShot(40, self.strt)
+
+        # TODO: re-test if needed on macOS
+        # self.refresh_gui()  # Need for QT on MacOSX.  Lame.
```
accumulator, line 109: @SomberNight I suspect this isn't needed anymore, can you check?

SomberNight (commented):
Added some more comments. Also tested on safe-t, looks ok.

electrum/gui/qt/wizard/wizard.py
SomberNight: I guess it is not needed anymore. Tested on macos 10.15 with pyqt 5.15.9.
Does not look needed on master either - it was added in 2016 when still using pyqt4, so a lot has changed since.

electrum/gui/qt/wizard/wallet.py
```diff
@@ -0,0 +1,1386 @@
...
+class WCHWUnlock(WizardComponent, Logger):
+    def __init__(self, parent, wizard):
+        WizardComponent.__init__(self, parent, wizard, title=_('Unlocking hardware'))
+        Logger.__init__(self)
+        self.plugins = wizard.plugins
+        self.plugin = None
+        self._busy = True
+        self.password = None
+
+        self.ok_l = WWLabel(_('Hardware successfully unlocked'))
```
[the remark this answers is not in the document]
accumulator: added decrypt check in WCHWUnlock, so this message doesn't appear late (after Finish)

electrum/wizard.py
```diff
@@ -416,14 +561,29 @@ def create_storage(self, path, data):
...
         else:
             raise Exception('unsupported/unknown keystore_type %s' % data['keystore_type'])
 
         if data['encrypt']:
             if k and k.may_have_password():
                 k.update_password(None, data['password'])
-            storage.set_password(data['password'], enc_version=StorageEncryptionVersion.USER_PASSWORD)
+            enc_version = StorageEncryptionVersion.USER_PASSWORD
+            if 'keystore_type' in data and data['keystore_type'] == 'hardware':
+                enc_version = StorageEncryptionVersion.XPUB_PASSWORD
```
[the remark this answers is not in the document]
accumulator: fixed.

ecdsa:
not directly related, but the `split_account` function does not work on qml. For example, if I try `./run_electrum -w electrum/tests/test_storage_upgrade/client_2_5_4_trezor_multiacc -g qml -v`, a popup is shown, but no split happens.

Since that is not needed on Android, I think this should be removed from qml

electrum/plugins/keepkey/keepkey.py
SomberNight: No, the bug is introduced with this PR.


https://github.com/spesmilo/electrum/blob/44a15951575254572f49d81f189cd4456b488815/electrum/plugins/keepkey/qt.py#L304
`bg.checkedId()`, before the PR, is called in the same method that creates `bg`. After the PR, it is called in a separate method.
Before the PR, it returns `2`. After the PR, it returns `-1`.
I think it is the groupbox `gb`, that gets GC-ed when we go out of scope, because of lack of parenting of objects:
https://github.com/spesmilo/electrum/blob/44a15951575254572f49d81f189cd4456b488815/electrum/plugins/keepkey/qt.py#L246

E.g. with this patch, the issue goes away:
```patch
diff --git a/electrum/plugins/keepkey/qt.py b/electrum/plugins/keepkey/qt.py
index 014cce74c8..efa0dc1255 100644
--- a/electrum/plugins/keepkey/qt.py
+++ b/electrum/plugins/keepkey/qt.py
@@ -241,9 +241,10 @@ class KeepkeyInitLayout(QVBoxLayout):
         hl.addWidget(self.label_e)
         hl.addStretch(1)
         self.addLayout(hl)
+        self._w = QWidget()
 
         if self.method in [TIM_NEW, TIM_RECOVER]:
-            gb = QGroupBox()
+            gb = QGroupBox(self._w)
             hbox1 = QHBoxLayout()
             gb.setLayout(hbox1)
             # KeepKey recovery doesn't need a word count
@@ -260,6 +261,7 @@ class KeepkeyInitLayout(QVBoxLayout):
                 rb.setChecked(True)
             self.cb_pin = QCheckBox(_('Enable PIN protection'))
             self.cb_pin.setChecked(True)
+            print(f"heyheyhey1. {self.bg.checkedId()=!r}")
         else:
             self.text_e = QTextEdit()
             self.text_e.setMaximumHeight(60)
@@ -301,6 +303,7 @@ class KeepkeyInitLayout(QVBoxLayout):
 
     def get_settings(self):
         if self.method in [TIM_NEW, TIM_RECOVER]:
+            print(f"heyheyhey2. {self.bg.checkedId()=!r}")
             item = self.bg.checkedId()
             pin = self.cb_pin.isChecked()
         else:
```

Only TIM_RECOVER is affected. TIM_NEW is not, as `gb` there won't get GC-ed:
https://github.com/spesmilo/electrum/blob/44a15951575254572f49d81f189cd4456b488815/electrum/plugins/keepkey/qt.py#L250-L251

SomberNight: I've pushed the first patch in https://github.com/spesmilo/electrum/commit/0a3dd8e5e53e9a95480cf1700d6ba2a0008f9a37

PR #8584, opened 2023-08-22
invoices: also run amount-validator on setter

SomberNight:
- `@amount_msat.validator` prevents the creation of invoices with e.g. too large amounts
- however the qml gui is mutating invoices by directly setting the `amount_msat` field, and it looks like attrs validators only run during init. We can use `on_setattr` (introduced in attrs==20.1.0).
- a wallet db upgrade is added to rm existing insane invoices

btw the qml gui was already doing its own input validation on the textedit (see qeconfig.btcAmountRegex). However that only limits the input to not have more chars than what is needed to represent 21M BTC (e.g. you can still enter 99M BTC, which the invoice logic does not tolerate later on - but is normally caught).

fixes https://github.com/spesmilo/electrum/issues/8582

### DIFF
--- a/contrib/requirements/requirements.txt
+++ b/contrib/requirements/requirements.txt
@@ -6,7 +6,7 @@ aiohttp>=3.3.0,<4.0.0
 aiohttp_socks>=0.3
 certifi
 bitstring
-attrs>=19.2.0
+attrs>=20.1.0
 
 # Note that we also need the dnspython[DNSSEC] extra which pulls in cryptography,
 # but as that is not pure-python it cannot be listed in this file!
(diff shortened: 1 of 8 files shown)


accumulator:
If we can handle post-construction validation with `on_setattr`, do we need the `set_amount_msat` func?

> the qml gui is mutating invoices

For completeness, the kivy and text UIs do the same

SomberNight:
> Nitpick: if we can handle post-construction validation with `on_setattr`, do we need the refactor to `set_amount_msat` func?

I added `set_amount_msat` so that it also changes `.outputs` for onchain invoices. `.outputs` and `.amount_msat` should be consistent. Try (on master) pasting an onchain address into qml send screen, type some amount X, and then click Save. Then open the same wallet on desktop qt and try paying the saved invoice. Desktop Qt creates a tx with a zero-amt-output because unlike qml it relies on `.outputs`.

Ideally we should limit the public API of this class (e.g. make `.amount_msat` private) but that would probably be easier without attrs.

Dirtyyt66 (approved):

Issue #8586, opened 2023-08-26
Revealer dialog cuts off text with large fonts


sorenstoutner:
I noticed that the Revealer dialog cuts off text with large fonts (especially noticeable on high DPI screens).

![Standard DPI Layout Problems](https://github.com/spesmilo/electrum/assets/26014704/d30c70eb-eb2a-4099-9056-48da0cef18e5)

![High DPI Layout Problems](https://github.com/spesmilo/electrum/assets/26014704/111bddd3-7d1a-478f-94bd-ca550c8ec725)

This is due to some maximum sizing of items being hardcoded into the layout.

If you like, I would be willing to take a crack at removing the layout restrictions so it could auto resize.  My Qt layout experience is mostly with .ui XML files and C++, but I think it would transition to Python.

SomberNight:
Sure, feel free to see if you can fix it. :)

SomberNight:
fixed in https://github.com/spesmilo/electrum/pull/8587

PR #8587, opened 2023-08-27
Fix Revealer Qt layout

sorenstoutner:
This fixes the layout problem discussed at #8586.

### DIFF
--- a/electrum/plugins/revealer/qt.py
+++ b/electrum/plugins/revealer/qt.py
@@ -4,8 +4,9 @@ Revealer
 Do you have something to hide?
 Secret backup plug-in for the electrum wallet.
 
-Tiago Romagnani Silveira, 2017
-
+Copyright:
+    2017 Tiago Romagnani Silveira
+    2023 Soren Stoutner <soren@stoutner.com>
 
 '''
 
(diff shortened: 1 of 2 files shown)


sorenstoutner:
Thanks for explaining what the underscore before a string indicates.  I wondered what that meant but didn't take the time to explore it.

I do understand the desire not to make trivial changes to translated strings.  Although I think the newer version looks better (the main reason I made the change was that the dialog looked weird with a lowercase "e" at the beginning of the second part of the string; I just altered the punctuation to be more expressive since I was in there anyway), I can see this not raising to the level of instantiating a translation change, so I will revert it to the old string.

As a little bit of background, when I started working on these dialogs my intention was only to fix the spacing and clipping of the text.  But once I got into it I realized that the word wrapping text, particularly those that included <br/>, let to problematic layouts on high DPI monitors.  So, I needed to separate those string into two new strings.  After I had done that, I realized that the existing text was so confusing that it took experimentation with the dialog for me to even understand how to use it.  So I rewrote a bunch of the strings to be more expressive for a first-time user.  By that point I had touched almost all the strings, so cleaning up this last one seemed like a natural way to go.

You will notice that, to make the layout adapt correctly, I have removed all the custom WWLabel  (Word Wrap Label) widgets and replaced them with standard QLabels.

electrum/plugins/revealer/qt.py
```diff
@@ -4,8 +4,9 @@
 Do you have something to hide?
 Secret backup plug-in for the electrum wallet.
 
-Tiago Romagnani Silveira, 2017
-
+Copyright:
+    2017 Tiago Romagnani Silveira
+    2023 Soren Stoutner <soren@stoutner.com>
```
SomberNight, line 9: Now that you've also added the word "copyright", I would feel better if we explicitly specified the license.
Before your change, I would say it was implicitly MIT, inherited from the root directory license.

You can just add e.g.
```
Distributed under the MIT software license, see the accompanying
file LICENCE or http://www.opensource.org/licenses/mit-license.php
```
Or I can do it too if you like, just comment that it's ok :)

sorenstoutner: That's a good Idea.  I'm a big fan of all files explicitly stating both their copyright and their license information.

SomberNight:
Thanks

PR #8591, opened 2023-08-30
remove the kivy gui

SomberNight:
We now use the qml gui on Android since version 4.4, and haven't been maintaining the kivy GUI for a while.

As-is, the kivy gui already has small breakages on master,
and I think e.g. https://github.com/spesmilo/electrum/pull/8545 and https://github.com/spesmilo/electrum/pull/8560 will break kivy and the ability the build a kivy apk even further, which we probably should not spend time on fixing.


88700022320126485980782887 (changes requested):
اريد سحب اموالي وارسالها على اسمي ورقمي 
عبدالله عبدالله حسن مقدام
الدولة اليمن 
+967770178557

وبعدذالك الغا الاشتراك

PR #8615, opened 2023-09-15
bip39: add likely script vs derivation path mistakes

accumulator:
bip39: add likely script vs derivation path mistakes for BIP49 and BIP84 paths. See e.g.

https://bitcointalk.org/index.php?topic=5454270.0

### DIFF
--- a/electrum/bip39_wallet_formats.json
+++ b/electrum/bip39_wallet_formats.json
@@ -36,6 +36,30 @@
         "iterate_accounts": true
     },
     {
+        "description": "Non-standard legacy on BIP84 path",
+        "derivation_path": "m/84'/0'/0'",
+        "script_type": "p2pkh",
+        "iterate_accounts": true
+    },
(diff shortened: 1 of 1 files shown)


SomberNight:
I've skimmed the bitcointalkthread.
I wonder, did you create the PR because of that single user creating that thread, or were there other users/reports of UTXOs at mismatched derivation/script addresses?
Do you know if there is a program that creates these addresses, or if it's just user error of manually creating these addresses (afterall, e.g. electrum can be used to create these addresses)?

accumulator:
No other users or reports, but since it's possible using Electrum and a bip39 seed, it's a possbile footgun (and it's trivial to add, I thought why not :)

SomberNight:
Obviously we cannot scan all arbitrary paths for all script types, but these are common paths, just for the incorrect script type. So, sure, we can scan them.

PR #8618, opened 2023-09-18
submarine swaps: register hold invoice early in wait_for_htlcs_and_br…

accumulator:
…oadcast, so a SwapManager.cancel_normal_swap() does not race.

cancel_normal_swap unregisters the hold invoice, which can occur before register_hold_invoice in wait_for_htlcs_and_broadcast if the network call has not returned yet.

~~This can lead to loss of funds, if the UI cancels the swap by calling cancel_normal_swap() but fails to cancel the running future. In this case, the swap 'succeeds', the funding TX is broadcast, but the invoice has not been paid.~~

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -674,6 +674,11 @@ class SwapManager(Logger):
         payment_hash = swap.payment_hash
         refund_pubkey = ECPrivkey(swap.privkey).get_public_key_bytes(compressed=True)
         if self.server_supports_htlc_first:
+            async def callback(payment_hash):
+                await self.broadcast_funding_tx(swap, tx)
+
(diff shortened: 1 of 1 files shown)


ecdsa:
> cancel_normal_swap unregisters the hold invoice, which can occur before register_hold_invoice in wait_for_htlcs_and_broadcast if the network call has not returned yet.

indeed, good point.

> This can lead to loss of funds, if the UI cancels the swap by calling cancel_normal_swap() but fails to cancel the running future. In this case, the swap 'succeeds', the funding TX is broadcast, but the invoice has not been paid.

I do not understand that part; the GUI never calls `cancel_normal_swap` directly- instead, that method is called if the future raises a `Cancelled` exception, so how would the GUI fail to cancel the future?

Issue #8542, opened 2023-07-19
Man Page


sorenstoutner:
I am writing a Man Page for Electrum, which is the primary documentation format used on Debian systems.

https://en.wikipedia.org/wiki/Man_page

I was able to get most of the information I needed from the excellent help commands.  But there are a few things that aren't clear to me.

1.  What are the possible arguments for the `-v VERBOSTIY` global option?  I am used to using log levels between 0-7, but trying any of those numbers produces an error.
2. What are the possible arguments for the `-V VERBOSITY` global option?  It says it should be a shortcut-filter list, but I wasn't able to find any documentation of what that is.
3. Is `electrum gui -g text` expected to work?  On my machine, I receive an error saying that the paperclip module is missing.
4. `electrum gui -L es_ES` does not appear to do anything, as Electrum is still launched in English.  However, the `-L` option does work as expected with the `make_seed` command.
5. `electrum gui -m`, which is supposed to hide the GUI on startup, does not appear to work.  Electrum launches but isn't hidden.  Am I missing something from the syntax?
6. I assume that `clear_ln_blacklist` clears the Lightning blacklist.  Is that correct?
7. `convert_xkey` lists an example of `xpub -> ypub`.  Are there any other types of keys beyond these two?
8. `--seed_type` lists `standard` and `segwit` as possible options.  Are there any other options?  Which is the default if none is specified?
9. Does `createmultisig` create a MultiSig address or a MultiSig wallet?  The documentation in Electrum says MultiSig address, but all the general Bitcoin information I found online only talks about MultiSig wallets.  What are the possible values for the `num` argument?  What is the syntax (delimiter) for using multiple public keys?
10. What does `dumpgraph` do?
11. How is `get_ssl_domain` used?  It returns blank for me, and it is unclear to me what ssl_keyfile and ssl_certificate it is using.
12. What are the possible fee methods that can be passed to `electrum getfeerate --fee_method FEE_METHOD`?
13. What is the syntax for passing a list of wallet addresses to `getprivatekeys`?  All of the variations I tried produced errors.
14. What is the syntax for passing a list of private keys to `signtransaction_with_privkey`?
15. How is `inject_fees` used?
16. Is there a specific syntax for `searchcontacts` or does it just simply match partial strings?
17. Is there a list of configuration variables that can be set with `setconfig`?

[...]

PR #8619, opened 2023-09-18
config: move tooltips from Qt gui into configvars

SomberNight:
related to https://github.com/spesmilo/electrum/issues/8542#issuecomment-1641780165 :
> We recently restructured config variables, and they are now listed in simple_config.py (they used to be scattered around in the code). Maybe we could pass a helpstring to ConfigVar, and make it visible from command line

This PR moves the tooltips into the ConfigVars - but they are not exposed to the CLI yet.

Note that many configvars are missing tooltips - I did not write new ones, only moved existing ones around.
This is because they are either only indirectly exposed to the GUI or not exposed at all.

### DIFF
--- a/electrum/gui/messages.py
+++ b/electrum/gui/messages.py
@@ -5,20 +5,6 @@ def to_rtf(msg):
     return '\n'.join(['<p>' + x + '</p>' for x in msg.split('\n\n')])
 
 
-MSG_RECOVERABLE_CHANNELS = _(
-"""Add extra data to your channel funding transactions, so that a static backup can be recovered from your seed.
-
-Note that static backups only allow you to request a force-close with the remote node. This assumes that the remote node is still online, did not lose its data, and accepts to force close the channel.
-
-If this is enabled, other nodes cannot open a channel to you. Channel recovery data is encrypted, so that only your wallet can decrypt it. However, blockchain analysis will be able to tell that the transaction was probably created by Electrum."""
-)
-
-MSG_CONFIG_INSTANT_SWAPS = _(
-"""If this option is checked, your client will complete reverse swaps before the funding transaction is confirmed.
-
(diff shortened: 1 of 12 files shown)


ecdsa:
Note: it would be nice to avoid lambda expressions everywhere. 

We can define a `_()` function in the `simple_config` module, that would be a no_op. 
That syntax would  trigger translatable strings collection. 
Strings would then be translated in the GUI modules.

SomberNight:
Similar ideas are discussed in the stdlib docs:
https://docs.python.org/3/library/gettext.html#deferred-translations :

> In this case, you are marking translatable strings with the function `N_()`, which won’t conflict with any definition of `_()`. However, you will need to teach your message extraction program to look for translatable strings marked with `N_()`. xgettext, pygettext, pybabel extract, and xpot all support this through the use of the `-k` command-line switch. The choice of `N_()` here is totally arbitrary

---

> We can define a `_()` function in the simple_config module, that would be a no_op.

I find that potentially confusing, and would prefer the `N_()` approach.

Anyway, note that it is more complicated than that, as some strings are constructed from parts, e.g.:
https://github.com/spesmilo/electrum/blob/f8c2a38954e3d4170a2b213024430d3851ad9ee0/electrum/simple_config.py#L1122-L1125
https://github.com/spesmilo/electrum/blob/f8c2a38954e3d4170a2b213024430d3851ad9ee0/electrum/simple_config.py#L1072-L1073

ecdsa:
If strings need to be constructed from parts, it might be better to let the GUI choose how to concatenate them. 

However, I suppose that in most cases, these concatenated strings should really be a single string; the work of translators is not really facilitated when phrases are taken out of their context. We used string concatenation for bad reasons; namely because it looks better indented in our code.

PR #8636, opened 2023-10-03
rbf batching: make sure new feerate > old feerate

ecdsa:
The previous lower bound did not ensure that, sometimes resulting in tx rejection. Note, though, that BIP125 does not explicitly state that the new feerate must be strictly higher than the old feerate.


SomberNight:
I guess this is documented in rule 6 here:
https://github.com/bitcoin/bitcoin/blob/30b3477507c7ac8d022a83d68afd560fe7ffc801/doc/policy/mempool-replacements.md

> 6. The replacement transaction's feerate is greater than the feerates of all directly conflicting transactions.

Looks like it is not in [BIP125](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki) indeed.

PR #8646, opened 2023-10-10
daemon: refactor load_wallet

accumulator:
daemon: refactor load_wallet to not just return None, but raise specific exceptions.

The following exceptions should be expected:
FileNotFoundError: given wallet path does not exist StorageReadWriteError: given file is not readable/writable or containing folder is not writable InvalidPassword: wallet requires a password but no password or an invalid password was given WalletFileException: any internal wallet data issue. specific subclasses can be caught separately:
-  WalletRequiresSplit: wallet needs splitting (split_data passed in Exception)
-  WalletRequiresUpgrade: wallet needs upgrade, and no upgrade=True was passed to load_wallet
-  WalletUnfinished: wallet file contains an action and needs additional information to finalize. (WalletDB passed in exception)

Removed qml/qewalletdb.py

This patch also fixes load_wallet calls in electrum/scripts and adds a qml workaround for dialogs opening and closing so fast that the dialog opened==true property change is missed (which we need to manage the dialog/page stack)


ecdsa:
looks good to me

ecdsa:
This introduced a regression: 
It is no longer possible to start the wizard in wallet creation mode at a specific path:

`./run_electrum -w  <path of wallet to be created>`

This used to start the wizard, now it raises an exception

ecdsa:
actually, this also raises the same exception if I try to create a new electrum directory:

`./run_electrum -D <new directory>`

ecdsa:
the above commit seems to fix it.

Issue #8028, opened 2022-10-20
Base directory for wallet creation changes after creating wallet in subdir


ln2max:
Electrum wallets are by default created in `~/.electrum/wallets/`. However, the user can create subdirectories, e.g `~/.electrum/wallets/experimental_wallets/`.

If such a subdirectory exists, the user may create wallets in that subdirectory from the New/Restore dialog box by typing `experimental_wallets/new_wallet` into the `Wallet:` path box in the first pane of the New/Restore wallet wizard.

If the user does so, creates the wallet, and then goes to create a new wallet in the same directory again by entering  `experimental_wallets/another_new_wallet` in that same box, the user will be presented with an error:
```
Cannot read file
StorageReadWriteError(OSError(FileNotFoundError(2, 'No such file or directory')))
```

and wallet creation will not be possible.

This error arises because Electrum has now changed to the directory `experimental_wallets/` and is using it as the root path for further wallet creation.

The expected behaviour in this situation would be for Electrum to always use `~/.electrum/wallets/` as the root path, so that the behaviour of the New/Restore wallet wizard is always the same.

However, it may be desirable to keep the existing behaviour. Particularly for high-volume advanced users, creating wallets in the same sub-directory could save time.

In this case, one solution would be for the New/Restore wallet wizard to display the current root path above the path entry field, so the user can take any unexpected changes into account.

Issue #8495, opened 2023-06-22
Can't open wallets stored on WinFSP/FUSE mounted drive


tayl:
I'm using Cryptomator to store some things in Google Drive and noticed today that Electrum is unable to read wallet files in a filesystem mounted by Cryptomator using WinFSP (FUSE).

![image](https://github.com/spesmilo/electrum/assets/3814633/ef4ee343-2eb4-4c8b-a171-70f07713bb4b)

Opening the file using the exact same file path in something like Notepad works as expected. Possibly this is related to how Cryptomator is mounting the filesystem, and how Electrum is referencing the path? Unsure..

SomberNight:
Can you get the full traceback from the [debug log](https://electrum.readthedocs.io/en/latest/faq.html#how-to-enable-debug-logging)?

tayl:
The wallet path as logged is slightly different from the path selected in the explorer interface. Accessing this path fails, but accessing the path as provided in the UI works.

`20230630T030814.812366Z |     INFO | i18n | gettext setting initial language to None
20230630T030815.043105Z |     INFO | simple_config.SimpleConfig | electrum directory C:\Users\Taylor\Desktop\Temporary\electrum_data
20230630T030815.082518Z |     INFO | __main__ | get_default_language: detected default as lang='en_UK'
20230630T030815.082518Z |     INFO | i18n | setting language to 'en_UK'
20230630T030815.087501Z |     INFO | logging | Electrum version: 4.4.5 - https://electrum.org - https://github.com/spesmilo/electrum
20230630T030815.151867Z |     INFO | logging | Python version: 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:24:45) [MSC v.1929 32 bit (Intel)]. On platform: Windows-10-10.0.19044-SP0
20230630T030815.151867Z |     INFO | logging | Logging to file: C:\Users\Taylor\Desktop\Temporary\electrum_data\logs\electrum_log_20230630T030815Z_6400.log
20230630T030815.151867Z |     INFO | logging | Log filters: verbosity '', verbosity_shortcuts ''
20230630T030815.155865Z |     INFO | plugin.Plugins | registering hardware bitbox02: ('hardware', 'bitbox02', 'BitBox02')
20230630T030815.160862Z |     INFO | plugin.Plugins | registering hardware coldcard: ('hardware', 'coldcard', 'Coldcard Wallet')
[...]

PR #8651, opened 2023-10-13
config: keep wallets directory stable, regardless of wallet opens in …

accumulator:
config: keep wallets directory stable, regardless of wallet opens in other directories

addresses issue #8028

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -669,7 +669,7 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger, QtEventListener):
         self.recently_visited_menu.setEnabled(bool(len(recent)))
 
     def get_wallet_folder(self):
-        return os.path.dirname(os.path.abspath(self.wallet.storage.path))
+        return os.path.abspath(self.config.get_datadir_wallet_path())
 
     def new_wallet(self):
         try:
--- a/electrum/gui/qt/wizard/wallet.py
+++ b/electrum/gui/qt/wizard/wallet.py
@@ -363,6 +363,8 @@ class WCWalletName(WizardComponent, Logger):
                           + _("Press 'Finish' to create/focus window.")
             if msg is None:
                 msg = _('Cannot read file')
+            if filename and os.path.isabs(relative_path(_path)):
+                msg += '\n\n' + _('Note: this wallet file is outside the default wallets folder.')
             msg_label.setText(msg)
             widget_create_new.setVisible(bool(temp_storage and temp_storage.file_exists()))
             if user_needs_to_enter_password:
--- a/electrum/gui/qt/wizard/wallet.py
+++ b/electrum/gui/qt/wizard/wallet.py
@@ -298,14 +298,13 @@ class WCWalletName(WizardComponent, Logger):
         self.layout().addStretch(1)
 
         temp_storage = None  # type: Optional[WalletStorage]
-        wallet_folder = os.path.dirname(path)
+        datadir_wallet_folder = self.wizard.config.get_datadir_wallet_path()
 
         def relative_path(path):
             new_path = path
             try:
-                datadir_wallet_path = self.wizard.config.get_datadir_wallet_path()
-                commonpath = os.path.commonpath([path, datadir_wallet_path])
-                if commonpath == datadir_wallet_path:
+                commonpath = os.path.commonpath([path, datadir_wallet_folder])
+                if commonpath == datadir_wallet_folder:
                     # below datadir_wallet_path, make relative
                     new_path = os.path.relpath(path, commonpath)
             except ValueError:
@@ -313,7 +312,7 @@ class WCWalletName(WizardComponent, Logger):
             return new_path
 
         def on_choose():
-            _path, __ = QFileDialog.getOpenFileName(self, "Select your wallet file", wallet_folder)
+            _path, __ = QFileDialog.getOpenFileName(self, "Select your wallet file", datadir_wallet_folder)
             if _path:
                 self.name_e.setText(relative_path(_path))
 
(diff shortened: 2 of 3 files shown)


SomberNight:
In the context of subdirectories inside the main `wallets/` folder, I agree this is probably an improvement.

However, also consider saving wallets to some unrelated path. E.g. imagine a user creating `~/dropbox/my_electrum_wallets/` or similar. Say they have all their existing wallet files in that folder. When they click `New/Restore` or `Open`, what is the reasonable thing to do? Is it to use `~/.electrum/wallets/` or `~/dropbox/my_electrum_wallets/`?

-----

An alternative idea would be to customise the default value in `WCWalletName.name_e`.
wizard_wallet_name
If the abs path is a subpath of `~/.electrum/wallets/`, we could show only the relative path,
otherwise we could show the full abs path.

That is,
`~/.electrum/wallets/wallet_47` -> `wallet_47`
`~/.electrum/wallets/lnfun/wallet_47` -> `lnfun/wallet_47`
`~/dropbox/my_electrum_wallets/` -> `~/dropbox/my_electrum_wallets/`  (or resolve the tilde; that's besides the point)

-----

Or as suggested in the linked issue, just do:
> one solution would be for the New/Restore wallet wizard to display the current root path above the path entry field, so the user can take any unexpected changes into account

(and maybe hide the dir path if it is `~/.electrum/wallets/`)

SomberNight:
Side-note, try:
- create a wallet at `/opt/tmp/w1`, set a password and encrypt file
- close program
- start electrum, wizard opens, `WCWalletName.name_e` is prefilled with `w1`
    - if I enter the password and press `Finish`, wallet at `/opt/tmp/w1` opens
    - if I change `WCWalletName.name_e` to `w2` and create a new wallet, it gets created at `~/.electrum/wallets/w2`
    - kind of confusing

This is already the behaviour on existing master; unaffected by this PR.

accumulator:
> In the context of subdirectories inside the main `wallets/` folder, I agree this is probably an improvement.
> 
> However, also consider saving wallets to some unrelated path. E.g. imagine a user creating `~/dropbox/my_electrum_wallets/` or similar. Say they have all their existing wallet files in that folder. When they click `New/Restore` or `Open`, what is the reasonable thing to do? Is it to use `~/.electrum/wallets/` or `~/dropbox/my_electrum_wallets/`? (e.g. see #8495)

In the absence of a user definable wallet folder preference setting (so the wallets dir can be separate from the data dir), we only have the `-D/--dir` cmdline option for that (which also keeps wallets together with config and other runtime files). Creating a separate cmdline option for this pref setting might be useful.

BTW, the cmdline descriptions for` -D, -w` are somewhat confusing, and could be improved (e.g 'electrum path' can be interpreted in a few different ways, 'wallet path' could mean a folder or a file, etc..)

> An alternative idea would be to customise the default value in `WCWalletName.name_e`. wizard_wallet_name If the abs path is a subpath of `~/.electrum/wallets/`, we could show only the relative path, otherwise we could show the full abs path.

Yes this would be logical.

> That is, `~/.electrum/wallets/wallet_47` -> `wallet_47` `~/.electrum/wallets/lnfun/wallet_47` -> `lnfun/wallet_47` `~/dropbox/my_electrum_wallets/` -> `~/dropbox/my_electrum_wallets/` (or resolve the tilde; that's besides the point)
> 
> Or as suggested in the linked issue, just do:
> 
> > one solution would be for the New/Restore wallet wizard to display the current root path above the path entry field, so the user can take any unexpected changes into account
> 
> (and maybe hide the dir path if it is `~/.electrum/wallets/`)

accumulator:
> EDIT: Ah! but it is not the behaviour on 4.4.6, where it creates w2 at `/opt/tmp/w2`. I guess this is a ~regression from the wizard refactor.

It depends from which electrum window the New/Restore is triggered from. I don't think this was changed by the new wizard, as the current window's wallet path was used before as well.

SomberNight:
> > EDIT: Ah! but it is not the behaviour on 4.4.6, where it creates w2 at `/opt/tmp/w2`. I guess this is a ~regression from the wizard refactor.
> 
> It depends from which electrum window the New/Restore is triggered from. I don't think this was changed by the new wizard, as the current window's wallet path was used before as well.

I am talking about directly opening into the wizard, before any main window exists. That is, clean start of electrum, `config.GUI_LAST_WALLET="/opt/tmp/w1"`, which is encrypted, so the wizard opens and prompts for password. The `WCWalletName.name_e` lineedit then shows `w1`, and if I change it to `w2` (instead of entering the pw), then it creates `w2` in a completely different folder than where `w1` is. This is changed between master and 4.4.6.
Anyway, it's not too important; but I would argue it's a bug :)

> In the absence of a user definable wallet folder preference setting (so the wallets dir can be separate from the data dir), we only have the -D/--dir cmdline option for that (which also keeps wallets together with config and other runtime files). Creating a separate cmdline option for this pref setting might be useful.

I don't know... sounds overkill to have another option for this.

accumulator:
> > > EDIT: Ah! but it is not the behaviour on 4.4.6, where it creates w2 at `/opt/tmp/w2`. I guess this is a ~regression from the wizard refactor.
> > 
> > 
> > It depends from which electrum window the New/Restore is triggered from. I don't think this was changed by the new wizard, as the current window's wallet path was used before as well.
> 
> I am talking about directly opening into the wizard, before any main window exists. That is, clean start of electrum, `config.GUI_LAST_WALLET="/opt/tmp/w1"`, which is encrypted, so the wizard opens and prompts for password. The `WCWalletName.name_e` lineedit then shows `w1`, and if I change it to `w2` (instead of entering the pw), then it creates `w2` in a completely different folder than where `w1` is. This is changed between master and 4.4.6. Anyway, it's not too important; but I would argue it's a bug :)

In this branch it shows an absolute path now if the last wallet is outside the datadir wallets subtree


> > In the absence of a user definable wallet folder preference setting (so the wallets dir can be separate from the data dir), we only have the -D/--dir cmdline option for that (which also keeps wallets together with config and other runtime files). Creating a separate cmdline option for this pref setting might be useful.
> 
> I don't know... sounds overkill to have another option for this.

Yeah I'm fine with just having `--dir` too.

accumulator:
any objections left on this PR?

PR #8653, opened 2023-10-17
qml: also show coins in Addresses page, and add a few filter options.…

accumulator:
… Additionally, long press

now activates multi-select mode, and add action to (un)freeze selection.


ecdsa:
the tab look good. can you find a better icon for the freeze button?

accumulator:
> the tab look good. can you find a better icon for the freeze button? (or just remove the seal icon)

done in d6a6f56ddf72396958998634808f921e60e1234f

PR #8659, opened 2023-10-24
request: amount_msat is set to None in Request when creating request …

accumulator:
request: amount_msat is set to None in Request when creating request and amount is 0 or undefined.

`PaymentInfo` declares `amount_msat` as optional, and the only place where `PaymentInfo.amount_msat` is assumed `not None` is through `LNWallet.get_lightning_history()` -> `get_payment_value()`, which I think is probably safe for items in the history.


accumulator:
This fixes `amount_msat` for newly created Requests, but existing no-amount requests still have `amount_msat=0` in the wallet db, so a conversion might be useful to add.

PR #8666, opened 2023-10-30
some qt wizard fixes/cleanup (WCHaveMasterKey, WCHaveSeed)

SomberNight:
- I think this fixes https://github.com/spesmilo/electrum/issues/8665
- adds missing "duplicate master key"/etc validation to WCHaveMasterKey

@accumulator, could you have a look?


accumulator:
LGTM

PR #8668, opened 2023-10-31
Cirrus: Mark E722 linter as mandatory

JeremyRand:
These linter warnings were all fixed in 312f2641e7284dc9ed7bde99a61060b4d9e35dd9, so making CI enforce them will reduce the risk of regressions.


SomberNight:
thx

PR #8669, opened 2023-10-31
Running Regtest functional tests, on AppImage Binaries

robertmin2:
Running Regtest functional tests, on AppImage Binaries

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -80,7 +80,7 @@ task:
   only_if: $CIRRUS_BRANCH == 'master'
 
 task:
-  name: "Regtest functional tests"
+  name: "Regtest functional tests $Binaries"
   compute_engine_instance:
     image_project: cirrus-images
     image: family/docker-builder
@@ -109,6 +109,22 @@ task:
     - electrum/tests/regtest/run_bitcoind.sh
   electrumx_service_background_script:
     - electrum/tests/regtest/run_electrumx.sh
+  matrix:
+    - env:
+        Binaries: 'Tarball'
+      depends_on:
+        - "linter: Flake8 Mandatory"
+    - env:
+        Binaries: 'AppImage'
+      app_image_setup:
+        - apt-get -y install libfuse2
+        - rm run_electrum
+        - VERSION="$(git describe --tags --always | cut -c 1-9)"
+        - curl -o run_electrum https://api.cirrus-ci.com/v1/artifact/build/$CIRRUS_BUILD_ID/build:%20AppImage/binaries/dist/electrum-$VERSION-x86_64.AppImage
+        - chmod u+x run_electrum
+      depends_on:
+        - "build: AppImage"
+        - "linter: Flake8 Mandatory"
   regtest_script:
     - sleep 10s
     - python3 -m unittest electrum/tests/regtest.py
@@ -116,8 +132,6 @@ task:
     ELECTRUM_REQUIREMENTS: contrib/requirements/requirements.txt
     # ElectrumX exits with an error without this:
     ALLOW_ROOT: 1
-  depends_on:
-    - "linter: Flake8 Mandatory"
 
 task:
   container:


.cirrus.yml
```diff
@@ -109,15 +109,29 @@ task:
...
+    - env:
+        Binaries: 'Tarball'
+      depends_on:
+        - "linter: Flake8 Mandatory"
+    - env:
+        Binaries: 'AppImage'
+      app_image_setup:
+        - apt-get -y install libfuse2
+        - rm run_electrum
+        - VERSION="$(git describe --tags --always | cut -c 1-9)"
```
SomberNight, line 122: I am confused about this cut (and btw it looks fragile). Can you give an example of what is being piped into it?

SomberNight: This is how the appimage is getting its name:
https://github.com/spesmilo/electrum/blob/33ac0b45917d1dfcd6d673eab2f7840173c84068/contrib/build-linux/appimage/make_appimage.sh#L26-L27

robertmin2: The initial attempt using this didn't succeed for some unknown reason, but it's functioning properly now.

SomberNight: Okay but why is the cut needed?
Note that the commit hash used by git describe is not guaranteed to have a constant length. As there are more commits in a repository, it tends to get longer and longer.

robertmin2: It was returning a commit hash longer than 9 characters, that's why I used `cut.` I tested  using ` VERSION=$(git describe --tags --dirty --always)` and it works well.

ecdsa:
hi, what is the point of running regtests with the appimage?

robertmin2:
@ecdsa AppImages bundle dependencies to make applications self-contained, but sometimes these bundled dependencies can cause compatibility issues. We can catch such issues early and fast

SomberNight:
Though note that typically the problems with an appimage surface when run on a somewhat different linux config than the build system. E.g. building on debian stable and then running on some flavour of arch. So testing on the build config has limited utility.
Instead of running the regtests, we could maybe just run the `version` command and check it outputs the expected thing.

In practice, I would expect if all source-based tests pass and `$ ./Electrum.AppImage -o version` works, then that implies your newly added tests also pass.

ecdsa:
I think this addition might be a waste of resources. In addition, it increases our maintenance work.
The same applies to #8672. I propose to revert this. 

We do need better tests. But that requires understanding how they work, what they do and what they do not do.
OTOH , this kind of pull request is easy to do, it does not require any understanding of the tests.

accumulator:
I do like the AppImage being tested, however I agree that a full regtest is probably overkill, but just a version test is a bit limited. 

If we can trigger wallet create+open plus a few cryptographic operations that could be sufficient.

JeremyRand:
> I do like the AppImage being tested, however I agree that a full regtest is probably overkill, but just a version test is a bit limited.

@accumulator I'm a little confused how we're defining overkill -- the added Cirrus task completed in under 6 minutes. Is 6 minutes on one Cirrus vCPU core really worth worrying about?

JeremyRand:
> Though note that typically the problems with an appimage surface when run on a somewhat different linux config than the build system. E.g. building on debian stable and then running on some flavour of arch.

@SomberNight This is a good point, I would be supportive of changing the Docker image that runs the AppImage tests. Arch isn't a bad choice, though maybe something even more different could be used... perhaps a distro that uses Musl instead of Glibc? Is the AppImage expected to run there without issues?

accumulator:
> @accumulator I'm a little confused how we're defining overkill -- the added Cirrus task completed in under 6 minutes. Is 6 minutes on one Cirrus vCPU core really worth worrying about?

In september we ran out of our compute allowance on cirrus-ci, so yes, we do need to be a bit careful.

JeremyRand:
> > @accumulator I'm a little confused how we're defining overkill -- the added Cirrus task completed in under 6 minutes. Is 6 minutes on one Cirrus vCPU core really worth worrying about?
> 
> In september we ran out of our compute allowance on cirrus-ci, so yes, we do need to be a bit careful.

@accumulator I see, I wasn't aware how constrained you were on this.

Is there a specific subset of the regtests that you think would sufficiently exercise the critical paths (wallet+crypto?) while minimizing the runtime? I think Robert and I would be happy to narrow it down.

ecdsa:
@JeremyRand what is the specific error that you are trying to catch with these tests?

JeremyRand:
> If you want to test that binaries are working, there should be some way.
> Maybe run the `version_info` command at the end of the build?

@ecdsa If we were to submit a PR for that, would you want `version_info` to be run in a separate container (with different OS) as was suggested above, or should we keep it in the existing AppImage build container?

ecdsa:
I would prefer to have it run in the existing build container, because it looks simpler that way.
However, I have not looked at that issue closely, maybe @SomberNight has?

PR #8672, opened 2023-10-31
Regtest functional tests MacOS

robertmin2:
The functional tests are executed on the tarball build within a macOS environment.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -208,6 +208,59 @@ task:
   depends_on:
     - "unittests: Tox Python 3.8"
 
+task:
+  name: "Regtest functional tests MacOS"
+  macos_instance:
+    image: macos-monterey-xcode
+  pip_cache:
+    folder: ~/.cache/pip
+    fingerprint_script: echo Regtest && echo macOS && cat $ELECTRUM_REQUIREMENTS
+  bitcoind_cache:
+    folder: /tmp/bitcoind
+    populate_script: mkdir -p /tmp/bitcoind
+  install_electrumx_script:
+    - brew install leveldb
+    - export LIBRARY_PATH="$LIBRARY_PATH:$(brew --prefix)/lib"
+    - export CPATH="$CPATH:$(brew --prefix)/include"
+    - CFLAGS="-I/opt/homebrew/include/ -L/opt/homebrew/lib/ -fno-rtti" pip3 install --force-reinstall --global-option="build_ext" plyvel --no-cache-dir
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -208,6 +208,59 @@ task:
...
+  install_electrumx_script:
+    - brew install leveldb
+    - export LIBRARY_PATH="$LIBRARY_PATH:$(brew --prefix)/lib"
+    - export CPATH="$CPATH:$(brew --prefix)/include"
+    - CFLAGS="-I/opt/homebrew/include/ -L/opt/homebrew/lib/ -fno-rtti" pip3 install --force-reinstall --global-option="build_ext" plyvel --no-cache-dir
+    # install e-x some commits after 1.16.0 tag
+    - pip3 install git+https://github.com/spesmilo/electrumx.git@4e66804dc0d668cd6bd4602b547e2f5b2e227e97
+    - PYTHON_VERSION=$(python3 -c "print('.'.join(map(str, __import__('sys').version_info[:2])))")
+    # Workaround for macOS limitation: To prevent ElectrumX from setting the maximum allowed sessions to zero
+    - sed -i '' 's/return value/return 200/g' /Users/$USER/Library/Python/$PYTHON_VERSION/lib/python/site-packages/electrumx/server/env.py
```
SomberNight, line 230: This sounds like something that should be upstream to e-x.

JeremyRand: Upstreaming to ElectrumX would be a bit more complicated since the fix would need to work on all platforms, but no particular objection here to upstreaming first if that's what you prefer?

SomberNight: https://github.com/spesmilo/electrumx/blob/9224d1ea6697790b79ec9e4b9be9e2206ce58bf4/electrumx/server/env.py#L114
I've tested on macOS now and it looks like the issue is that the open file limit by default is very low compared to Linux.
```
>>> resource.getrlimit(resource.RLIMIT_NOFILE)
(256, 9223372036854775807)
```
and indeed `$ ulimit -n` reports 256.
So... not sure how to go about this -- maybe e-x should raise instead of returning 0.
But anyway, instead of hacking with `sed`, a cleaner solution would be to try to raise the limit, by running e.g. `$ ulimit -n 1000`.

ecdsa:
same comment as  #8669

.cirrus.yml
robertmin2: You need to restart the system, if you want to use `ulimit -n`

Issue #8687, opened 2023-11-13
Non-default sighash type


ik105:
I created an unsigned transaction in Sparrow wallet with SIGHASH_NONE, then imported the psbt into Electrum and was surprised that Electrum didn't show any warning about the unusual sighash type. It becomes possible to steal coins with malware even from cold or multisig storage. 

Please add a warning when signing a transaction with a non-default sighash type.

SomberNight:
Good point. Do you know how it works/looks in other programs when trying to sign a tx with a non-sighash_all input?

ik105:
Sparrow wallet doesn't warn, but there is an option to select the type of sighash (I don't think it's enough, you can ignore it). Trezor seems to just not accept such transactions. Ledger shows "Non-default sighash - > Reject if you're not sure -> Continue/Reject". Coldcard shows this:

![photo_2023-11-13_20-26-23](https://github.com/spesmilo/electrum/assets/38519936/ec7de082-6bc7-47ea-87a7-2b19decdcc17)

ecdsa:
I guess there is no use case for `sighash_none` at the moment, so we can forbid it.
There are use cases for `sighash_all|anyonecanpay`

RobinLinus:
There is a use case for `sighash_none`: decaying multi-signatures. E.g., if you have a 2-of-3 multisig, signing a `nLocktime` transaction with one of your keys using `sighash_none` effectively decays your 2-of-3 into a 1-of-3 once the locktime opens.

Is there any way to allow it in Electrum?

stwenhao:
> I guess there is no use case for `sighash_none` at the moment, so we can forbid it.

There are use cases for sighash_none, for example funding a puzzle: https://mempool.space/tx/8349df0753e80cce322322f1b76789e1d0fd6693aed2f4de4e49576423081ae7

[...]

PR #8699, opened 2023-11-20
show warning message in transaction dialog if one of my inputs have non-standard sighash.

accumulator:
solution towards #8687

### DIFF
--- a/electrum/gui/qt/transaction_dialog.py
+++ b/electrum/gui/qt/transaction_dialog.py
@@ -50,8 +50,7 @@ from electrum.bitcoin import base_encode, NLOCKTIME_BLOCKHEIGHT_MAX, DummyAddres
 from electrum.i18n import _
 from electrum.plugin import run_hook
 from electrum import simple_config
-from electrum.transaction import SerializationError, Transaction, PartialTransaction, PartialTxInput, TxOutpoint
-from electrum.transaction import TxinDataFetchProgress
+from electrum.transaction import SerializationError, Transaction, PartialTransaction, TxOutpoint, TxinDataFetchProgress
 from electrum.logging import get_logger
 from electrum.util import ShortID, get_asyncio_loop
 from electrum.network import Network
@@ -77,14 +76,15 @@ _logger = get_logger(__name__)
 dialogs = []  # Otherwise python randomly garbage collects the dialogs...
 
 
-
 class TxSizeLabel(QLabel):
     def setAmount(self, byte_size):
         self.setText(('x   %s bytes   =' % byte_size) if byte_size else '')
 
+
 class TxFiatLabel(QLabel):
     def setAmount(self, fiat_fee):
         self.setText(('≈  %s' % fiat_fee) if fiat_fee else '')
+
 
 class QTextBrowserWithDefaultSize(QTextBrowser):
     def __init__(self, width: int = 0, height: int = 0):
@@ -96,8 +96,8 @@ class QTextBrowserWithDefaultSize(QTextBrowser):
     def sizeHint(self):
         return QSize(self._width, self._height)
 
-class TxInOutWidget(QWidget):
 
+class TxInOutWidget(QWidget):
     def __init__(self, main_window: 'ElectrumWindow', wallet: 'Abstract_Wallet'):
         QWidget.__init__(self)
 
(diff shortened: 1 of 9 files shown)


electrum/gui/qt/transaction_dialog.py
```diff
@@ -245,6 +251,9 @@ def insert_tx_io(
                 tcf_shortid=tcf_shortid,
                 short_id=str(txin.short_id), addr=addr, value=txin_value,
             )
+            if isinstance(txin, PartialTxInput):
+                if txin.sighash != Sighash.ALL and self.wallet.adb.is_mine(addr):
+                    self.sighash_label.setText(_('Warning: Non-standard Sighash!'))
```
SomberNight, line 256: Users won't know what this means. If we don't explain it (e.g. using a mouse-over tooltip), then we should use stronger wording at least. For example, instead of non-standard, we could say "dangerous".

electrum/gui/qt/transaction_dialog.py
```diff
@@ -245,6 +251,9 @@ def insert_tx_io(
                 tcf_shortid=tcf_shortid,
                 short_id=str(txin.short_id), addr=addr, value=txin_value,
             )
+            if isinstance(txin, PartialTxInput):
+                if txin.sighash != Sighash.ALL and self.wallet.adb.is_mine(addr):
```
SomberNight, line 255: txin.sighash is None by default, so as-is, this ~always warns

SomberNight, line 255: I think it would be better to move the checks to the shared lib, have that return a warning, and only add GUI-specific code that can show warnings.

In particular, have you seen `wallet.get_warning_for_risk_of_burning_coins_as_fees(tx)`?
That already exists (not used in qml though), we could maybe just rename and extend that with this new kind of warning.

Another argument for trying to have a generic warning logic instead of different kind of warning tests directly done in the GUI is that `wallet.get_warning_for_risk_of_burning_coins_as_fees` was used in both qt and kivy, but none of us noticed it was not added to qml: fewer things to do in the GUI, the better. :)

SomberNight, line 255: A related but distinct point re adding warnings in the GUIs is what to do about the CLI. E.g. consider calling the `signtransaction` command with a PSBT given to us by someone else (e.g. coinjoin). Perhaps we should just hard fail if the sighashes are not sighash_all, by default.

This can be handled separately and does not have to be solved in this PR though.

electrum/gui/qt/transaction_dialog.py
accumulator: Sure, the text can be changed, but I think it also needs an extra confirmation step, e.g. add a confirmation pop-up before signing. Maybe even one extra safeguard for dangerous sighash types like SIGHASH_NONE, which can only be confirmed if first enabling an option in the preferences for instance.

electrum/gui/qt/transaction_dialog.py
accumulator: ok. handled now.

accumulator: I moved the check to the backend, and made a distinction between confirmation needed and reject. 

Not sure if this should stay within `PartialTransaction` or if `Wallet` is a better place for it..

I don't want to overload `wallet.get_warning_for_risk_of_burning_coins_as_fees` with this, as that func name is rather specific, but we can add a high-level 'do-all-checks' func which calls the specific ones.

accumulator: I'll make a TODO item for `wallet.get_warning_for_risk_of_burning_coins_as_fees` in qml

accumulator: Yeah we should give this some thought, see how we can share as much non-gui specific gui code in shared code (separate layer?)

accumulator:
This PR currently handles SIGHASH_SINGLE as confirm, not as reject, but this might need additional checks for value_in ~== value_out

electrum/gui/qt/transaction_dialog.py
SomberNight: Note that `wallet.get_warning_for_risk_of_burning_coins_as_fees` is closely related in the sense that it too is doing sighash stuff. Specifically, SIGHASH_ALL is not always safe either, depending on legacy/segwit/taproot, the number of inputs, etc, and that is what is being considered there.

accumulator: > Not sure if this should stay within `PartialTransaction` or if `Wallet` is a better place for it..

Moved `check_sighash` to `Abstract_Wallet`, seems more appropriate.

accumulator: > A related but distinct point re adding warnings in the GUIs is what to do about the CLI. E.g. consider calling the `signtransaction` command with a PSBT given to us by someone else (e.g. coinjoin). Perhaps we should just hard fail if the sighashes are not sighash_all, by default.

Added the checks to `wallet.sign_transaction()` with a kwarg to not consider warnings as fatal, which can be set through the `--iknowwhatimdoing` cmdline arg

Issue #8708, opened 2023-11-28
Join PSBTs without duplicate outputs


1440000bytes:
How can we join PSBTs and make sure outputs dont get duplicated? I tried this but it does not work:

```python
    def finalize_coinjoin(self):

        self.config.WALLET_MERGE_DUPLICATE_OUTPUTS = False

        coinjoin_tx = PartialTransaction()

        for psbt_64 in self.round_input_list:
            psbt = PartialTransaction.from_raw_psbt(psbt_64)
            coinjoin_tx.join_with_other_psbt(psbt)

        coinjoin_tx = coinjoin_tx.serialize()

        return coinjoin_tx
```

SomberNight:
`config.WALLET_MERGE_DUPLICATE_OUTPUTS` was only added in a recent commit: https://github.com/spesmilo/electrum/commit/227e2574445662abf03bfba7fe9ec0c755f2af4a
Since then, `tx.join_with_other_psbt()` takes config as an argument.
Also, you probably want to set `config.WALLET_MERGE_DUPLICATE_OUTPUTS=True`, not False.

1440000bytes:
> `config.WALLET_MERGE_DUPLICATE_OUTPUTS` was only added in a recent commit: [227e257](https://github.com/spesmilo/electrum/commit/227e2574445662abf03bfba7fe9ec0c755f2af4a) Since then, `tx.join_with_other_psbt()` takes config as an argument. Also, you probably want to set `config.WALLET_MERGE_DUPLICATE_OUTPUTS=True`, not False.

Will that ensure there are no outputs with same address?

SomberNight:
Here is the description:
https://github.com/spesmilo/electrum/blob/6467db0b7db6a3ed3d8403cf24a10dbfe1c6913c/electrum/simple_config.py#L944-L949

> The amounts should not get merged. So it should be 0.01 and not 0.02 for an address in this example.

Oh so you want something different. What is your use case for this?

1440000bytes:
> Oh so you want something different. What is your use case for this?

Electrum [plugin for joinstr](https://gitlab.com/1440000bytes/joinstr/-/tree/main/plugin?ref_type=heads) follows this process:

1. 5 Peers publish their outputs
2. Same peers create PSBT with 1 input and 5 outputs (SIGHASH ALL | ACP). They publish their signed PSBTs.
3. These are joined and broadcasted as a coinjoin transaction.

SomberNight:
[...]

PR #8709, opened 2023-11-28
add `remove_duplicates` argument in join_with_other_psbt()

1440000bytes:
Fixes https://github.com/spesmilo/electrum/issues/8708

Context: I am using this function in joinstr electrum plugin to join all the PSBTs submitted by different participants. Each PSBT has 1 input and several outputs. The outputs get duplicated (multiple outputs with same address) when joining PSBT which needs to be avoided.


1440000bytes:
Closing as the issue is resolved.

See https://github.com/spesmilo/electrum/issues/8708#issuecomment-1829698029

PR #8710, opened 2023-11-28
Add support for IPv6 for proxy address/hostname

accumulator:
Allows using proxy servers by IPv6 address, and proxy server hostnames that resolve to an IPv6 address.

Splits proxy authentication to --proxyuser and --proxypassword command line options and corresponding config keys, removing them from the single serialized proxy string.

### DIFF
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -40,7 +40,7 @@ import functools
 from enum import IntEnum
 
 import aiorpcx
-from aiorpcx import ignore_after
+from aiorpcx import ignore_after, NetAddress
 from aiohttp import ClientResponse
 
 from . import util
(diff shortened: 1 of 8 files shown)


electrum/network.py
```diff
@@ -168,35 +168,52 @@ class NetworkParameters(NamedTuple):
...
-        n += 1
-    if len(args) > n:
-        proxy["password"] = args[n]
+    if args[0] in proxy_modes:
+        proxy['mode'] = args[0]
+        args = args[1:]
+
+    def is_valid_port(ps: str):
+        try:
+            assert 0 < int(ps) < 65535
```
SomberNight, line 188: I still think it's bad practice to rely on asserts being enabled.
just do:
```py
if not (0 < int(ps) < 65535):
   raise ValueError()
```

electrum/simple_config.py
```diff
@@ -926,6 +926,8 @@ def _default_swapserver_url(self) -> str:
     NETWORK_AUTO_CONNECT = ConfigVar('auto_connect', default=True, type_=bool)
     NETWORK_ONESERVER = ConfigVar('oneserver', default=False, type_=bool)
     NETWORK_PROXY = ConfigVar('proxy', default=None)
+    NETWORK_PROXY_USER = ConfigVar('proxy_user', default=None)
+    NETWORK_PROXY_PASSWORD = ConfigVar('proxy_password', default=None)
```
SomberNight, line 930: can we make these `type_=str`? I don't see why that would cause problems.

SomberNight (approved):
I haven't tested but at high-level it looks good.

accumulator:
merged.

PR #8713, opened 2023-11-30
Allow users to downoad external plugins

ecdsa:
Finally, I have decided to hardcode the hash of external plugins, for security reasons. 
This simplifies things:
 - we do not need 'per wallet activation' anymore
 - thus, we do not need to define an API number
 - since we endorse plugins through their hash. we can propose a download dialog in the GUI, which simplifies UX

Notes:
 - `contrib/make_plugin` creates a zip file and updates `electrum/plugins.json` with the hash.
 - the current zipfile can only contain a single module. we should consider adding python dependencies in the zip file.


electrum/plugin.py
```diff
@@ -162,12 +170,9 @@ def disable(self, name: str) -> None:
...
     def is_plugin_enabler_config_key(cls, key: str) -> bool:
-        if not key.startswith('use_'):
-            return False
-        # note: the 'use_' prefix is not sufficient to check, there are
-        #       non-plugin-related config keys that also have it... hence:
-        name = key[4:]
-        return name in cls.find_all_plugins()
+        # fixme: the 'use_' prefix is not sufficient to check, there are
+        #         non-plugin-related config keys that also have it.
+        return key.startswith('use_')
```
SomberNight, line 175: Maybe we should just do a breaking change and change the config keys.
E.g. reserve the prefix "enable_plugin_" for this, instead of "use_".
As part of the change, we could end up (one-time) disabling all plugins for existing users, but I think that's fine.

ecdsa: indeed, that's probably the best thing to do

ecdsa: Actually, not going to do that.
I think it would be safer to enable plugins not globally, but on a per-wallet basis.
That way, we can write plugin preferences in the encrypted wallet file, and we have some protection against malware.

SomberNight: Do you also want this for the built-in plugins?
Also, for some plugins it might not make sense to be per-wallet, e.g. what about plugins that are used as part of the wizard? E.g. consider the trezor plugin. You want to have that enabled when running the wizard if you want to be able to create a wallet containing a trezor.

ecdsa: I don't like the idea of having several types of plugins. The initial motivation for the plugin system was to allow third parties to develop their stuff without needing our permission. Maybe some of the current plugins, in particular those that are touching the wizard, should be moved to the main codebase?

ecdsa: I am reverting the "plugins per wallet" concept; this makes things too complicated.
thus, I will change the prefix to "enable_plugin_"

ecdsa:
The make_plugin script creates a zip file that can be distributed via http. Here is how to use it:

First, call `./contrib/make_plugin /path/to/yourplugin`

This will create a plugin file named `yourplugin-version.zip`, where `version` is set in `__init__.py` of the plugin.
It will also update the hash and metadata of your plugin in `electrum/plugins.json`.
This hash update is required by Electrum in order to import the plugin.

In order to test changes to your plugin, you need to both move the zip file to your userspace and update the hash.
It is best to perform both operations together:
`./contrib/make_plugin /path/to/yourplugin && mv /path/to/yourplugin-*.zip ~/.electrum/plugins/`

ecdsa:
You can include icons in the plugin zipfile.
example:

```python
from electrum.gui.qt.util import read_QIcon_from_bytes
window.tabs.addTab(tab_widget, read_QIcon_from_bytes(self.read_file("myplugin.png")), _("MyPlugin"))

```

1440000bytes:
> The `contrib/make_plugin` script creates a zip file that can be distributed via http. Here is how to use it:
> 
> First, call `./contrib/make_plugin /path/to/yourplugin`

This is giving me an error for joinstr plugin:

```
creating /home/test/Downloads/electrum-pl/electrum/electrum/plugins/joinstr.zip
added joinstr/./__init__.py
added joinstr/./qt.py
added joinstr/__pycache__/test1.cpython-310.pyc
added joinstr/__pycache__/__init__.cpython-310.pyc
added joinstr/__pycache__/qt.cpython-310.pyc
libsecp256k1 library failed to load. exceptions: [OSError('/home/test/.local/lib/python3.10/site-packages/electrum/libsecp256k1.so.0: cannot open shared object file: No such file or directory'), OSError('libsecp256k1.so.0: cannot open shared object file: No such file or directory')]
Error: Failed to load libsecp256k1.
```

ecdsa:
you should remove the `__pycache__` directory before calling that script

1440000bytes:
This command fixed the error:

```
./contrib/make_libsecp256k1.sh
```

Now I am getting a different error:

```
creating /home/test/Documents/join-pl/joinstr.zip
added joinstr/./__init__.py
added joinstr/./qt.py
  File "/home/test/Downloads/electrum-pl/electrum/contrib/./make_plugin", line 47, in <module>
    'download_url': module.download_url,
AttributeError: module 'joinstr' has no attribute 'download_url'
```

ecdsa:
> This command fixed the error:
> 
> ```
> ./contrib/make_libsecp256k1.sh
> ```
> 

I see. Alternatively, you could install the dependencies with apt:
` sudo apt-get install python3-pyqt5 libsecp256k1-dev python3-cryptography `


> Now I am getting a different error:
> 
> ```
> creating /home/test/Documents/join-pl/joinstr.zip
> added joinstr/./__init__.py
> added joinstr/./qt.py
> Traceback (most recent call last):
>   File "/home/test/Downloads/electrum-pl/electrum/contrib/./make_plugin", line 47, in <module>
>     'download_url': module.download_url,
> AttributeError: module 'joinstr' has no attribute 'download_url'
> ```

That is expected. In the plugin's `__init__.py` you need to define the following new strings:
- author
- licence
- version
- download_url

ecdsa:
Great. Please note that `download_url` is for a URL where the plugin ZIP file is publicly accessible.

electrum/plugin.py
```diff
@@ -117,13 +124,133 @@ def load_plugins(self):
             details = d.get('registers_keystore')
             if details:
                 self.register_keystore(name, gui_good, details)
-            self.descriptions[name] = d
-            if not d.get('requires_wallet_type') and self.config.get('use_' + name):
+            if d.get('requires_wallet_type'):
+                # trustedcoin will not be added to list
+                continue
```
SomberNight, line 129: Why?

This broke the qml gui:
```
  1.35 | E | __main__ | daemon.run_gui errored
  File "/home/user/wspace/electrum/./run_electrum", line 458, in handle_cmd
    d.run_gui()
  File "/home/user/wspace/electrum/electrum/daemon.py", line 623, in run_gui
    self.gui_object = gui.ElectrumGui(config=self.config, daemon=self, plugins=self._plugins)
  File "/home/user/wspace/electrum/electrum/util.py", line 482, in do_profile
    o = func(*args, **kw_args)
  File "/home/user/wspace/electrum/electrum/gui/qml/__init__.py", line 82, in __init__
    self.app = ElectrumQmlApplication(sys.argv, config=config, daemon=daemon, plugins=plugins)
  File "/home/user/wspace/electrum/electrum/gui/qml/qeapp.py", line 428, in __init__
    self.plugins.load_plugin('trustedcoin')
  File "/home/user/wspace/electrum/electrum/plugin.py", line 269, in load_plugin
    raise Exception(f"could not find plugin {name!r}")
Exception: could not find plugin 'trustedcoin'
```

SomberNight: fixed in https://github.com/spesmilo/electrum/commit/96bb1611c5bafdd4a8d82bd8ff0fad4faf925927

PR #8714, opened 2023-11-30
network: async tor probe

accumulator:
Probe for TOR asynchronously if proxy is socks5.

This allows for longer timeouts without blocking the GUI, making detection less probabilistic for slow proxy servers.

Adds a separate TOR icon in the desktop client status bar

### DIFF
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -323,6 +323,8 @@ class Network(Logger, NetworkRetryManager[ServerAddr]):
 
         self._allowed_protocols = {PREFERRED_NETWORK_PROTOCOL}
 
+        self.proxy = None
+        self.tor_proxy = False
         self._init_parameters_from_config()
 
         self.taskgroup = None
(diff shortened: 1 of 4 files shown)


electrum/network.py
```diff
@@ -625,18 +628,29 @@ def _set_default_server(self) -> None:
...
     def _set_proxy(self, proxy: Optional[dict]):
-        self.proxy = proxy
-        dns_hacks.configure_dns_resolver()
-        self.logger.info(f'setting proxy {proxy}')
+        if self.proxy == proxy:
+            return
 
+        self.logger.info(f'setting proxy {proxy}')
+        self.proxy = proxy
         self.tor_proxy = False
```
SomberNight, line 636: Shouldn't this be tri-state instead, and start with None? That way you can potentially distinguish False and dont-know.
(also btw, I see this is a pre-existing name, but I think `is_proxy_tor` or `is_tor_proxy` would be a better name)

accumulator: agreed, done.

electrum/network.py
```diff
@@ -625,18 +628,29 @@ def _set_default_server(self) -> None:
...
 
     def _set_proxy(self, proxy: Optional[dict]):
-        self.proxy = proxy
-        dns_hacks.configure_dns_resolver()
+        if self.proxy == proxy:
+            return
+
         self.logger.info(f'setting proxy {proxy}')
+        self.proxy = proxy
+        self.is_proxy_tor = False
```
SomberNight, line 636: Wouldn't you also want to set it to `None` here?

PR #8719, opened 2023-12-01
wallet db: deduplicate "seed_type" field

SomberNight:
In the db, the 'seed_type' field could be present both at the top-level and inside keystores.

Note:
- both fields had usages
- the top-level field was added in 2.8 re "initial segwit support" (3a64ec0f2ed9c845b5c01daa07fafcbfa26c3cb7)
  - there was no db upgrade for it, so older files are missing it
  - if present, valid values can be electrum types but also other types supported by the wizard, e.g. "bip39"
- the keystore-level field was added in 4.1 (7b7bba22992eca4536a91249758f178b4a7e044a)
  - there was a db upgrade when it was introduced, so old files also have it
  - if present, valid values can only be electrum types
- there is not much value in the top-level one having a non-electrum value, and those values were never used by other parts of the code
  - note that when creating a standard wallet from a bip39 seed, the seed is discarded. Only the derived xprv and the derivation path are kept. If we changed this and also kept the seed, e.g. to display it to the user, then it would make sense to save the seed type (e.g. "bip39"). However storing that seed_type would make more sense at the keystore level (not top-level).

We delete the top-level 'seed_type' field.

```
{
    "keystore": {
        "seed_type": "segwit",
        ...
    },
    "seed_type": "segwit",
    ...
}
```

### DIFF
--- a/electrum/gui/qml/qewallet.py
+++ b/electrum/gui/qml/qewallet.py
@@ -366,7 +366,7 @@ class QEWallet(AuthMixin, QObject, QtEventListener):
 
     @pyqtProperty(str, notify=dataChanged)
     def seedType(self):
-        return self.wallet.db.get('seed_type')
+        return self.wallet.get_seed_type()
 
     @pyqtProperty(bool, notify=dataChanged)
     def isWatchOnly(self):
--- a/electrum/gui/qt/wallet_info_dialog.py
+++ b/electrum/gui/qt/wallet_info_dialog.py
@@ -36,9 +36,7 @@ class WalletInfoDialog(WindowModalDialog):
         seed_available = _('False')
         if wallet.has_seed():
             seed_available = _('True')
-            ks = wallet.keystore
-            assert isinstance(ks, keystore.Deterministic_KeyStore)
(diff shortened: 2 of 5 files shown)


accumulator:
LGTM

>       * note that when creating a standard wallet from a bip39 seed, the seed is discarded. Only the derived xprv and the derivation path are kept. If we changed this and also kept the seed, e.g. to display it to the user, then it would make sense to save the seed type (e.g. "bip39"). However storing that seed_type would make more sense at the keystore level (not top-level).
 
Keeping BIP39 seeds would be a nice addition.

SomberNight:
>> note that when creating a standard wallet from a bip39 seed, the seed is discarded. Only the derived xprv and the derivation path are kept. If we changed this and also kept the seed, e.g. to display it to the user, then it would make sense to save the seed type (e.g. "bip39"). However storing that seed_type would make more sense at the keystore level (not top-level).
>
> Keeping BIP39 seeds would be a nice addition. 

Sure we could do it. However note that every time it is shown to the user, we should also show the type of the seed (and if it is not electrum, be clear about that).

PR #8725, opened 2023-12-04
qt, qml: use lowercase bolt11 when copying or sharing

accumulator:
Be more compatible, uppercase is a QR optimization


SomberNight:
Sure, looks good.

Issue #6440, opened 2020-07-26
gap limit for change addresses is too small and cannot be changed


Fonta1n3:
It seems impossible to increase the range of change addresses (change) that electrum watches for, therefore my balance is completely incorrect. The information is not easy to find in the docs or google, the existing poor answers I have found result in a crash in the electrum console.

SomberNight:
What version of Electrum are you using?
Starting with version 4.0.1, the gap limit for change is 10; was 6 previously.
Changing the gap limit for receiving (external) addresses is possible via CLI/RPC commands (see `changegaplimit`) as there are good usecases for it, but it is not exposed in the GUI as it caused too many misunderstanding for newbies in the past.
Changing the gap limit for change (internal) addresses is only possible via hacking the source code; or using the Qt console (as it is a full featured python interpreter); it is even less exposed as there are even fewer good usecases for modifying it.

Anyway, if you know what you are doing, you can change the gap limit for receiving addresses and the gap limit for change addresses like this, in the Qt console tab:
```
wallet.change_gap_limit(100)
wallet.gap_limit_for_change = 100
wallet.synchronize()
```

Fonta1n3:
> What version of Electrum are you using?
> 
> Starting with version 4.0.1, the gap limit for change is 10; was 6 previously.
> 
> 
> 
> Changing the gap limit for receiving (external) addresses is possible via CLI/RPC commands (see `changegaplimit`) as there are good usecases for it, but it is not exposed in the GUI as it caused too many misunderstanding for newbies in the past.
> 
> Changing the gap limit for change (internal) addresses is only possible via hacking the source code; or using the Qt console (as it is a full featured python interpreter); it is even less exposed as there are even fewer good usecases for modifying it.
> 
> 
> 
> Anyway, if you know what you are doing, you can change the gap limit for receiving addresses and the gap limit for change addresses like this, in the Qt console tab:
> 
> ```
> 
> wallet.change_gap_limit(100)
> 
> wallet.gap_limit_for_change = 100
> 
> wallet.synchronize()
> 
> ```

Even fewer good use cases? Huh... not being able to see my balance or spend my btc is a fairly good use case.

SomberNight:
Sure but this situation only arose because you did something special using another application, so you are now trying to fix a previous mess. Instead, it's better to avoid creating the mess in the first place.

Re not being many good usecases, I meant for changing it in the first place.

May I ask why you have a large gap in your change addresses, and how big it is?

[...]

PR #8726, opened 2023-12-04
wallet: make gap limit for change configurable

Filiprogrammer:
Add the ability to configure the gap limit for change addresses while
retaining that value in the wallet database.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -974,7 +974,7 @@ class Commands:
         return wallet.create_new_address(False)
 
     @command('w')
-    async def changegaplimit(self, new_limit, iknowwhatimdoing=False, wallet: Abstract_Wallet = None):
+    async def changegaplimit(self, new_limit, change=False, iknowwhatimdoing=False, wallet: Abstract_Wallet = None):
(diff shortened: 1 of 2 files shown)


SomberNight:
Hi. What is your use case for this?

note: related issue at https://github.com/spesmilo/electrum/issues/6440

Filiprogrammer:
- Users may need to restore an existing wallet from a seed that includes transactions on change addresses beyond the current hardcoded gap limit. With the current setup, users have to manually change the gap limit for change addresses every time they open the wallet in Electrum. Introducing a configurable option that is saved in the wallet database alleviates this manual task, **providing a more user-friendly experience and reducing the risk of oversight**.
- Another point is **consistency**. Currently, receiving addresses allow for a configurable gap limit, while change addresses are limited by a hardcoded value.
- The ability to change the gap limit for change addresses via the **CLI** `electrum changegaplimit <newlimit> --iknowwhatimdoing --change` offers power users a convenient method for adjusting their wallet settings. And again, since it is already possible to change the gap limit for receiving addresses via the CLI, it makes sense to be consistent and have the same ability for change addresses.

PR #8736, opened 2023-12-11
wallet: since we have jsonpatch, save labels immediately upon setting.

accumulator:


ecdsa:
jsonpatch is only enabled on non-encrypted wallets

accumulator:
> jsonpatch is only enabled on non-encrypted wallets

hmm, yes, that is unfortunate.

ecdsa:
in order to enable this with encrypted wallets, we need to define proper file header.
see the last commits of https://github.com/spesmilo/electrum/compare/master...jsonpatch

PR #8737, opened 2023-12-11
qt,qml: add a welcome page as initial page for server connect wizard

accumulator:
This is much less intimidating than asking if the user wants to use a proxy out of the gate.


ecdsa:
I disagree. it creates an extra screen, and increases the number of user interactions required to setup a wallet.

accumulator:
> I disagree. it creates an extra screen, and increases the number of user interactions required to setup a wallet.

Most of the immediate feedback I got when people used electrum for the first time was "What is a proxy" :smile:

ecdsa:
well, it makes it faster for users who are on default settings, but slower for everyone else.
I think 'default settings' is also kind of opaque, we don't know what we are choosing..

how about the following:
 - keep the logo
 - replace 'Welcome' with 'Network Settings'
 - instead of the 'default settings' checkbox, create two checkboxes: one for 'use a proxy', one for 'select server manually'

That way, the two yes/no screens for proxy and auto-connect can be removed, and we have actually made it faster for everyone.
We would lose the help text that is currently in the auto-connect dialog, but that text it not very good, and part of it can be relocated in the "server" screen.

electrum/gui/qt/wizard/server_connect.py
```diff
@@ -18,18 +22,55 @@ class QEServerConnectWizard(ServerConnectWizard, QEAbstractWizard):
...
-        self.setWindowTitle(_('Network and server configuration'))
+        self.window_title = _('Network and server configuration')
 
         # attach gui classes
         self.navmap_merge({
-            'autoconnect': { 'gui': WCAutoConnect },
-            'proxy_ask': { 'gui': WCProxyAsk },
-            'proxy_config': { 'gui': WCProxyConfig },
-            'server_config': { 'gui': WCServerConfig },
+            'welcome': {'gui': WCWelcome, 'params': {'icon': ''}},
```
SomberNight, line 29: I get the welcome screen on every start, until I manually change the network settings.

1. `$ rm -rf electrum_data/`
2. `$ ./run_electrum -v --testnet -P -g qt`
3. welcome screen opens, click "Finish" (btw, shouldn't it be "Next"?)
4. create a wallet or just quit wizard there
5.  `$ ./run_electrum -v --testnet -P -g qt`
6. welcome screen opens again

See
https://github.com/spesmilo/electrum/blob/b5d8d31723720de1c415a3cce7f3378b7b77d181/electrum/gui/qt/__init__.py#L487-L488
To avoid this, for example, when leaving the welcome screen with "use default settings" checked, we could set NETWORK_AUTO_CONNECT to True or maybe `config.cv.NETWORK_AUTO_CONNECT.get_default_value()`.
Or have another escape clause from the branch, idk.

accumulator:
> well, it makes it faster for users who are on default settings, but slower for everyone else. I think 'default settings' is also kind of opaque, we don't know what we are choosing..
> 
> how about the following:
> 
>     * keep the logo
> 
>     * replace 'Welcome' with 'Network Settings'
> 
>     * instead of the 'default settings' checkbox, create two checkboxes: one for 'use a proxy', one for 'select server manually'
> 
> 
> That way, the two yes/no screens for proxy and auto-connect can be removed, and we have actually made it faster for everyone. We would lose the help text that is currently in the auto-connect dialog, but that text it not very good, and part of it can be relocated in the "server" screen.

@ecdsa What do you think about https://github.com/spesmilo/electrum/pull/8737/commits/65b89d41e531dd78a39e3a13ab0c5b7dead71286 ? (only qml currently)

electrum/wizard.py
```diff
@@ -652,16 +654,13 @@ class ServerConnectWizard(AbstractWizard):
...
                 'accept': self.do_configure_autoconnect,
-                'last': lambda d: d['autoconnect']
-            },
-            'proxy_ask': {
-                'next': lambda d: 'proxy_config' if d['want_proxy'] else 'autoconnect'
+                'last': lambda d: bool(d['autoconnect'] and not d['want_proxy'])
             },
             'proxy_config': {
-                'next': 'autoconnect',
+                'next': 'server_config',
```
SomberNight, line 663: If someone checks `config_proxy` it does not matter whether they check `config_server`, they will see both screens.

SomberNight (commented):
> @ecdsa What do you think about https://github.com/spesmilo/electrum/pull/8737/commits/65b89d41e531dd78a39e3a13ab0c5b7dead71286 ? (only qml currently)

I think it looks good.

accumulator:
@SomberNight thx for reviewing

ecdsa:
it looks good to me. 

Note: testing QML, I noticed that the app hangs if the user cancels the first screen. Sadly, it is also the case on master.

accumulator:
> Note: testing QML, I noticed that the app hangs if the user cancels the first screen. Sadly, it is also the case on master.

fixed

PR #8738, opened 2023-12-12
mac build: migrate to new VM (macOS 10.14 -> 11)

SomberNight:
See individual commits.
In short, Apple removed support for `altool`, which we were using for notarisation, and to be able to use their new tool (`notarytool`), the build machine needs at least macOS 11.


ecdsa:
it works for me: hid.cpython-310-darwin.so is now reproducible between builds.

PR #8741, opened 2023-12-12
add labels_received callback and update labels even when set for item

accumulator:


electrum/plugins/labels/labels.py
```diff
@@ -156,14 +157,14 @@ async def pull_thread(self, wallet: 'Abstract_Wallet', force: bool):
                 result[key] = value
 
         for key, value in result.items():
-            if force or not wallet._get_label(key):
-                wallet._set_label(key, value)
+            wallet._set_label(key, value)
```
ecdsa, line 160: Note that this changes behaviour significantly: it will now overwrite local labels, regardless of the 'force' setting.
Not against this, but:
- has this been tested with two instances of the same wallet?
- is the 'force' checkbox still meaningful?

accumulator: The `force` param is still used by the plugin UI (force download in desktop client), which then retrieves all labels from the labelserver, instead of just the set of labels between `old_nonce ... new_nonce`.

ecdsa: if the plugin is turned off for a while, and then turned on, are the new labels added in that time at risk of being to be overwritten by some older labels from the server?

accumulator: The whole labelsync plugin is pretty basic, so yes, labels will be overwritten.
But the inverse is also true, no existing label is overwritten, leading to differences between wallet instances, which is probably also unexpected behavior for users.

accumulator: We could ask to upload (my labels take priority) or download (server labels take priority) when turning on the labelsync feature, then overwrite labels while the feature is on. This would mean it has to move to a per-wallet option.

Dirtyyt66:
EEKZ573

electrum/plugins/labels/labels.py
ecdsa: > But the inverse is also true, no existing label is overwritten, leading to differences between wallet instances, which is probably also unexpected behavior for users.

good point. I guess we can accept the change of logic.

SomberNight: Note that this introduces more trust in the server: a malicious server can now overwrite/delete labels stored by the client. Also, because the labels are encrypted in a very naive way, a keyed per wallet (not per address), the server can now e.g. mix up and overwrite labels across addresses. but I guess the real solution would be to re-design the labelsync protocol. replace aes-cbc with some AEAD, make the encryption keys different between addresses, and make the nonces per address (instead of per-wallet) - the nonce could simply be a unix timestamp.

PR #8743, opened 2023-12-13
wallet: refactor check_sighash/risk_of_burning_coins_as_fees

SomberNight:
This continues https://github.com/spesmilo/electrum/pull/8699

- risk_of_burning_coins_as_fees is turned into a private (helper) method, only called by check_sighash. UIs should only care about check_sighash.
- check_sighash returns instance of new class "TxSighashDanger" instead of tuple
- made warning levels more fine-grained (FEE_WARNING_SKIPCONFIRM vs FEE_WARNING_NEEDCONFIRM)
- this became more complicated than I had hoped for but I think it is worth it to ~merge check_sighash and risk_of_burning_coins_as_fees into one.

### DIFF
--- a/electrum/gui/qml/components/TxDetails.qml
+++ b/electrum/gui/qml/components/TxDetails.qml
@@ -362,7 +362,7 @@ Pane {
                 onClicked: {
                     if (txdetails.shouldConfirm) {
                         var dialog = app.messageDialog.createObject(app, {
-                            text: qsTr('Confirm signing non-standard transaction?'),
+                            text: qsTr('Confirm signing transaction despite warnings?'),
                             yesno: true
                         })
                         dialog.accepted.connect(function() {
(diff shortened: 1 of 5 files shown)


electrum/gui/qml/components/TxDetails.qml
```diff
@@ -362,7 +362,7 @@ Pane {
                 onClicked: {
                     if (txdetails.shouldConfirm) {
                         var dialog = app.messageDialog.createObject(app, {
-                            text: qsTr('Confirm signing non-standard transaction?'),
+                            text: qsTr('Confirm signing transaction despite warnings?'),
```
SomberNight, line 365: note: I've changed this phrasing as "(non-)standard transaction" has a specific meaning in bitcoind (re relay policy) and hence I found it confusing. Though most users probably don't know that meaning. Still, I think it is better not to overload the term.

PR #8772, opened 2023-12-22
qml: add transaction inputs in TxDetails and …

accumulator:
ConfirmTxDialog, RbfBumpFeeDialog, RbfCancelDialog, CpfpBumpFeeDialog and allow collapse/expand of input/output lists by clicking on label

### DIFF
--- a/electrum/gui/qml/qetxfinalizer.py
+++ b/electrum/gui/qml/qetxfinalizer.py
@@ -6,7 +6,7 @@ from PyQt6.QtCore import pyqtProperty, pyqtSignal, pyqtSlot, QObject
 
 from electrum.logging import get_logger
 from electrum.i18n import _
-from electrum.transaction import PartialTxOutput, PartialTransaction, Transaction
+from electrum.transaction import PartialTxOutput, PartialTransaction, Transaction, TxOutpoint
 from electrum.util import NotEnoughFunds, profiler
 from electrum.wallet import CannotBumpFee, CannotDoubleSpendTx, CannotCPFP, BumpFeeStrategy
 from electrum.plugin import run_hook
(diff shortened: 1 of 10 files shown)


SomberNight:
Looks nice.

Could we apply the same green/yellow colouring to the input addresses that the outputs have?
That is, if the input address is known, and it is a change address, make it yellow, and if it is non-change ismine, make it green.

accumulator:
> Could we apply the same green/yellow colouring to the input addresses that the outputs have? That is, if the input address is known, and it is a change address, make it yellow, and if it is non-change ismine, make it green.

Yeah, I had that initially, but it looked a bit off.

accumulator:
added color to inputs again.
with fontsize the same for inputs and outputs it looks better.

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -210,14 +222,38 @@ def update_from_tx(self, tx):
         self.fee = QEAmount(amount_sat=int(fee))
         self.feeRate = f'{feerate:.1f}'
 
+        self.update_inputs_from_tx(tx)
         self.update_outputs_from_tx(tx)
```
SomberNight, line 226: You need to do something similar in `QETxCpfpFeeBumper.update()`, near (new) line 868.
There only `update_outputs_from_tx` is called atm, so either call either both (`update_inputs_from_tx` and `update_outputs_from_tx`),
or maybe call `update_from_tx` instead, though the other side-effects of that are not relevant.

SomberNight (commented):
Skimmed the code, and then played around it with a bit.

I think it's much better than status quo. Even forgetting about the inputs, I really like seeing the output indices.

Also, and I guess this might have been your motivation, for advanced users it's useful to see the number of inputs when creating/bumping txs.

electrum/gui/qml/qetxfinalizer.py
accumulator: thx, good catch

Jircs1:
https://github.com/spesmilo/electrum/pull/8772#issue-2054098522

Jircs1 (commented):
accumulator:qml_tx_inputs

PR #8786, opened 2024-01-02
Restore ability to pass wallet in-memory

MrNaif2018:
Hello! This pull request restore the ability to pass wallet directly, avoiding loading it in daemon beforehand. This is used with e.g. custom daemons, or in this case specifically when implementing a diskless mode, which does not load the wallet anywhere except memory. As wallet doesn't have storage associated with it, it can't be added to daemon. Before I was able to pass it just fine, but due to added check here: 4e80ef818df0295fa46ee50bbe695b7e6eab1eda I no longer can
When passing wallet object directly as before, I now get that wallet object is not defined. When passing a wallet path it works, but it defeats the purpose of passing wallet object directly and breaks this special in-memory mode

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -140,9 +140,9 @@ def command(s):
                 if cmd.requires_wallet and kwargs.get('wallet') is None:
                     kwargs['wallet'] = daemon.config.get_wallet_path()
                 if 'wallet' in cmd.options:
-                    wallet_path = kwargs.get('wallet', None)
-                    if isinstance(wallet_path, str):
-                        wallet = daemon.get_wallet(wallet_path)
+                    wallet = kwargs.get('wallet', None)
+                    if isinstance(wallet, str):
+                        wallet = daemon.get_wallet(wallet)
                         if wallet is None:
                             raise Exception('wallet not loaded')
                         kwargs['wallet'] = wallet
--- a/electrum/tests/test_daemon.py
+++ b/electrum/tests/test_daemon.py
@@ -1,21 +1,21 @@
 import os
 from typing import Optional, Iterable
 
+from electrum.commands import Commands
 from electrum.daemon import Daemon
 from electrum.simple_config import SimpleConfig
-from electrum.wallet import restore_wallet_from_text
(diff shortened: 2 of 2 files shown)


MrNaif2018:
cc @SomberNight

gruve-p:
utACK https://github.com/spesmilo/electrum/pull/8786/commits/9b5bead6f4c28aeede152ed6c26dda7eb91e205d

ecdsa:
This is ugly, because it creates uncertainty on the type of the `wallet` parameter.

What is the use case? In which context is that diskless mode used?
if you are implementing a custom daemon, I don't think it should call commands from the commands module.
there might be some logic inside the commands that you want to avoid replicating. 
It would be worth looking at the specific commands that you need, and maybe move their logic into wallet.py.

SomberNight:
> This is ugly, because it creates uncertainty on the type of the wallet parameter.

That uncertainty already exists in the `@command` decorator. There is some magic stuff already going on there, irrespective of this PR.

Nevertheless, it is quite difficult to understand what is going on here, and @MrNaif2018 should have provided a small code example and a traceback to illustrate what is going on. More importantly, this is the kind of change that really warrants unit tests. I emphatically ask everyone reading this, please seriously consider writing unit tests for core library code, *especially* when fixing regressions.

I have now written unit tests for this (and will push them), but please, in future, write your own tests.

MrNaif2018:
> if you are implementing a custom daemon, I don't think it should call commands from the commands module.

That's exactly what I have been doing for 4+ years, creating Commands object and calling commands. And it works fine and allows to not write existing code

>  EDIT: or, at least, we ourselves are already abusing the types in unit tests, see e.g.

Yep, that's exactly the way I am calling it from my app, not sure why it works in unit tests though :thinking:  (ah I know, because commands object was created without daemon, hence not entering this branch being fixed here)
Passing wallet object directly is useful to avoid loading it from disk in such scenarios where only a temp wallet is needed without dumping private keys on disk

About unit tests, I fully agree. The thing is, this thing wasn't covered by unit tests before, hence this breaking was indeed not found out (and it's a specific use-case). I will try to write them in the future

Issue #8764, opened 2023-12-21
qml history list shows fiat deltas using today's FX rate (if `historicRates` disabled)


SomberNight:
There is a `historicRates` toggle in the qml preferences, enabled by default.
(1) If enabled, the user can only select FX rate providers that provide historical data, and the history list shows fiat deltas calculated using the historical rates.
(2) If disabled, the user can select any FX rate provider, and the history list shows fiat deltas calculated using the current FX rate.

This is unlike the qt desktop gui, where the history tab does not show any fiat information if historical rates are disabled.

I find (2) very confusing. Some days ago I was not getting FX rates from coingecko (probably because of using Tor), and I disabled historicRates and changed to another provider. Today I was just scrolling through the history, looking for an old tx, and was very confused by the fiat labels -- they were using today's FX rate and I forgot about disabling the toggle. I think it's confusing and kind of pointless to show these fiat labels.

-----

I say we never do fiat conversion for historical stuff using current price. E.g.:

<details>
 <summary>patch1</summary>

```patch
diff --git a/electrum/gui/qml/components/controls/FormattedAmount.qml b/electrum/gui/qml/components/controls/FormattedAmount.qml
index b33880ad3b..27394c5e61 100644
--- a/electrum/gui/qml/components/controls/FormattedAmount.qml
+++ b/electrum/gui/qml/components/controls/FormattedAmount.qml
@@ -43,11 +43,17 @@ GridLayout {
     }
 
     function setFiatValue() {
-        if (showAlt)
-            if (historic && timestamp)
-                fiatLabel.text = '(' + Daemon.fx.fiatValueHistoric(amount, timestamp) + ' ' + Daemon.fx.fiatCurrency + ')'
-            else
+        if (showAlt) {
+            if (timestamp) {
+                if (historic) {
+                    fiatLabel.text = '(' + Daemon.fx.fiatValueHistoric(amount, timestamp) + ' ' + Daemon.fx.fiatCurrency + ')'
+                } else {
+                    fiatLabel.text = ""
+                }
+            } else {
                 fiatLabel.text = '(' + Daemon.fx.fiatValue(amount) + ' ' + Daemon.fx.fiatCurrency + ')'
+            }
+        }
 
     }
 
diff --git a/electrum/gui/qml/components/controls/HistoryItemDelegate.qml b/electrum/gui/qml/components/controls/HistoryItemDelegate.qml
index 79e40b4fd7..84525d4c85 100644
[...]

PR #8788, opened 2024-01-04
qml: don't show fiat amount when timestamp more than a day old and historic rates are disabled

accumulator:
#8764

### DIFF
--- a/electrum/gui/qml/components/controls/HistoryItemDelegate.qml
+++ b/electrum/gui/qml/components/controls/HistoryItemDelegate.qml
@@ -105,7 +105,7 @@ Item {
                     function updateText() {
                         if (!Daemon.fx.enabled) {
                             text = ''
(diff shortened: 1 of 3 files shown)


electrum/gui/qml/components/controls/HistoryItemDelegate.qml
```diff
@@ -108,7 +108,11 @@ Item {
                         } else if (Daemon.fx.historicRates) {
```
SomberNight, line 108:
```suggestion
                        } else if (Daemon.fx.historicRates && model.timestamp) {
```
(pre-existing bug regardless of PR: if historyRates is enabled, for unconfirmed (or local) txs, the fiat prices is not shown)

SomberNight (commented):
Looks good. Just that one comment to address, then merge.

PR #8799, opened 2024-01-08
daemon.load_wallet to raise if given redundant password

SomberNight:
- for a wallet with no password, running commands with a pw given should raise InvalidPassword, instead of ignoring the provided pw
- and for wallet.unlock, it is very confusing to save the given pw in memory, if the wallet does not actually use a pw

(let's see if tests pass on CI)

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -515,6 +515,10 @@ class Daemon(Logger):
         if db.get_action():
             raise WalletUnfinished(db)
         wallet = Wallet(db, config=config)
+        if password:
+            # if a password was given, we check it.
+            # e.g. for unpassworded wallet, if given pw, we should raise.
+            wallet.check_password(password)
         return wallet
 
     @with_wallet_lock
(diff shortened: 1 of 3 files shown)


electrum/daemon.py
```diff
@@ -515,6 +515,10 @@ def _load_wallet(
         if db.get_action():
             raise WalletUnfinished(db)
         wallet = Wallet(db, config=config)
+        if password:
+            # if a password was given, we check it.
+            # e.g. for unpassworded wallet, if given pw, we should raise.
+            wallet.check_password(password)
```
SomberNight, line 521: @ecdsa commented out of band that load_wallet should not care about the keystore, only about the storage (wallet.check_password tests both storage and keystore), and e.g. if a password is given to load_wallet for a wallet that only has keystore encryption, load_wallet should unconditionally raise, even if the provided password matches the keystore

```patch
diff --git a/electrum/daemon.py b/electrum/daemon.py
index 4cfd0012db..3225e590d5 100644
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -510,6 +510,9 @@ class Daemon(Logger):
             if not password:
                 raise InvalidPassword('No password given')
             storage.decrypt(password)
+        else:
+            if password:
+                raise InvalidPassword("password given but wallet storage is not encrypted")
         # read data, pass it to db
         db = WalletDB(storage.read(), storage=storage, upgrade=upgrade)
         if db.get_action():
```

Necmeddinhammami: Yis

PR #8804, opened 2024-01-09
Fix graceful shutdown for lightning swaps

MrNaif2018:
This pull request fixes slow shutdown issues when lightning wallets are loaded. 
Basically when `lnworker` is created, a `SwapManager` object is created. And `lnworker.start_network` calls `swap_manager.start_network`. Which in turn uses it's new taskgroup to spawn an infinite task.
68be7688185affc870fb01226061cbff4936b3f0 made swap manager use it's own taskgroup instead of network one. But network taskgroup is being properly stopped, while SwapManager has no stop method or something, so `self.taskgroup.cancel_remaining()` is not called anywhere
This pull request fixes this

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -968,6 +968,8 @@ class LNWallet(LNWorker):
         if self.lnwatcher:
             await self.lnwatcher.stop()
             self.lnwatcher = None
+        if self.swap_manager:  # may not be present in tests
+            await self.swap_manager.stop()
 
     async def wait_for_received_pending_htlcs_to_get_removed(self):
         assert self.stopping_soon is True
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -210,6 +210,9 @@ class SwapManager(Logger):
         coro = self.pay_pending_invoices()
         asyncio.run_coroutine_threadsafe(self.taskgroup.spawn(coro), self.network.asyncio_loop)
 
+    async def stop(self):
+        await self.taskgroup.cancel_remaining()
+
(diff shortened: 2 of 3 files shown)


ecdsa:
thanks; looks good to me

SomberNight:
The change looks ok.
Though I am a bit confused by the pre-existing way of how taskgroups are used in the SwapManager. For example, the taskgroup is never actually joined...

Relatedly, you say

> This pull request fixes slow shutdown issues when lightning wallets are loaded.

I don't really see slow shutdown issues even without the diff. How can I reproduce? Are you using the swapserver plugin?

MrNaif2018:
I tried to reproduce it with electrum darmon indeed, but didn't find a way yet. Was able to with my custom daemon. Compared between 2 electrum versions I had pinned, one had no slowdown, new one had. Then step by step I figured out that it was only reproduced when wallet is restored from seed. From seed means lightning is activated. Disabling lnworker helped. Checked the diff of changes and found out submarine swaps one to be a possible candidate, and indeed if I didn't call swap_manager.start_network, it stopped fine. I am not sure how cancellation properly works in asyncio, but I guess those OldTaskgroup require to manually call this method to be properly stopped.
Maybe if to add logs in swap server taskgroup like inside pay_pending_invoices (ads any kind of print) see if it still prints longer than it should. Technically it should be reproducible when calling stop_wallet, where everything is logged stopped but this thing still prints, which means the task is still rubbish

ecdsa:
merged manually in 7369145a493b055b559bcecdcab6c0ad7fb55f5f

PR #8818, opened 2024-01-16
ci: only run build tasks once per day (or on manual trigger)

SomberNight:
The android+windows+appimage build tasks make up around 2/3 of our CI time usage currently.
(in particular the android build is using a lot of credits: the build caching for it seems to have broken around the qt6 changes. I guess the ~6 GBs we are trying to store in the cache is just too much...)

I think we could de-prioritise these binary builds to only be done once per day, instead of on every commit.

---

(untested)
related https://github.com/cirruslabs/cirrus-ci-docs/discussions/949

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -149,8 +149,16 @@ task:
         ELECTRUM_LINTERS_IGNORE: ""
       allow_failures: true
 
+
+# Cron jobs configured in https://cirrus-ci.com/settings/...
+# - job "nightly" on branch "master" at "0 30 2 * * ?"  (every day at 02:30Z)
 task:
   name: "build: Windows"
+  matrix:
+    - trigger_type: manual
+      only_if: $CIRRUS_CRON == ""
+    - trigger_type: automatic
+      only_if: $CIRRUS_CRON == "nightly"
   container:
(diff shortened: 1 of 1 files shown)


accumulator:
I already removed the build tree for `hostqt6` in https://github.com/accumulator/python-for-android/commit/d976b539dc7c808d103a7edfe6dc1fd300582a89, which reduces the data to be cached from 10GB to 6GB. We could do the same for target qt6 and see if that takes us under the threshold.

SomberNight:
> I already removed the build tree for hostqt6 in https://github.com/accumulator/python-for-android/commit/d976b539dc7c808d103a7edfe6dc1fd300582a89, which reduces the data to be cached from 10GB to 6GB. We could do the same for target qt6 and see if that takes us under the threshold.

Right, I saw the commit; and I think you are right, it just might be enough.
For reference, right before the qt6 merge, with qt5, the p4a cache was around 4200 MB. (e.g. this build 
https://cirrus-ci.com/task/5174511004286976 still has logs)

Still, I think it's kind of overkill to build on every commit in any case. We should manually trigger builds for PRs/commits that we think relevant, and otherwise just do daily builds.

Issue #8088, opened 2022-12-04
Remove RBF from user visible settings


llamasoft:
Bitcoin 24.0 adds a new [`mempoolfullrbf` option](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-24.0.md#notice-of-new-option-for-transaction-replacement-policies) that allows for replace-by-fee even if the original transaction didn't set the RBF flag.

Currently, the [`get_tx_info`](https://github.com/spesmilo/electrum/blob/e8c5ff5d2d4b3ecf821baf0fb1880ece0e276d02/electrum/wallet.py#L762) and [`dscancel`](https://github.com/spesmilo/electrum/blob/e8c5ff5d2d4b3ecf821baf0fb1880ece0e276d02/electrum/wallet.py#L2066) functions only consider a double-spend-cancel viable if the transaction has the RBF flag set (i.e. `if not tx.is_final()`).  
As the linked release notes mention, full RBF will now be a mainstream feature even though many alternate implementations have supported it for a while.  Either `is_final()` should always return False or it should depend on if the transaction has been confirmed or not.

junderw:
From a usability perspective, what you really want to avoid is:

1. Electrum says we can cancel.
2. Electrum fails to cancel at a high rate due to lack of adoption of the mempoolfullrbf flag (currently defaults to off)

Maybe introduce the logic change as a simple flag that advanced users can toggle, and the default can be switched to "always return False" after the success rate is high enough (enough of the network adopts the mempoolfullrbf flag)

ecdsa:
I think there is no need to update the current logic for the moment. 
As @junderw pointed out, lack of adoption would create usability issues.
Electrum has RBF by default, that should be enough for most use cases.

SomberNight:
> Maybe introduce the logic change as a simple flag that advanced users can toggle

We could add a config key that allows RBF-ing non-signalling txs.

petertodd:
Counterpoint re: usability issues: the times when users are most likely to need this, mempool congestion, are times when due to txs being dropped full-rbf doesn't need any miner support to be fairly often succesful. In fact, you don't even need to be connected to full-rbf peers as a dropped tx isn't being double-spent from their perspective.

ecdsa:
> We could add a config key that allows RBF-ing non-signalling txs.

I do not think it is the right approach to add new stuff every time we want to fix a rare issue. 
Lets stop trying to please everyone all the time.

[...]

Issue #7072, opened 2021-02-26
Consider setting RBF flag for Lightning funding transactions


SomberNight:
Currently we intentionally don't set the RBF flag for LN funding transactions (channel opening txs).
https://github.com/spesmilo/electrum/blob/84326cf1f78f60201e5bac44ff306c240147adde/electrum/lnpeer.py#L686

This means that if the user lowballs the fee, the funding tx can get stuck in the mempool in limbo for a very long time, and the user has essentially no recourse: all they can do is wait.

To be clear, with current the Lightning/BOLT v1 spec, it is not possible to non-destructively bump the fee of a funding transaction. That is, if we bump the fee and keep the funding output, the txid changes which invalidates the channel id, and the remote peer will not find the funding outpoint -- the funds then can only be recovered manually through with the help of the channel counterparty.
Bumping the fee should be possible in the future with the channel v2 spec (https://github.com/lightningnetwork/lightning-rfc/pull/524), but naturally only in a collaborative way (the peers need to exchange messages).

Code in `wallet.py` only allows bumping the fee of a tx that both (1) opted in to RBF, and (2) is not a LN funding tx for one of our channels:
https://github.com/spesmilo/electrum/blob/84326cf1f78f60201e5bac44ff306c240147adde/electrum/wallet.py#L618
https://github.com/spesmilo/electrum/blob/84326cf1f78f60201e5bac44ff306c240147adde/electrum/wallet.py#L647-L648

As said at the beginning, we don't set RBF for our funding txs, so check (2) is actually not relied on right now.

The issue is that not opting in to RBF not only disables `bump_fee` but also e.g. double-spend cancelling the tx:
https://github.com/spesmilo/electrum/blob/84326cf1f78f60201e5bac44ff306c240147adde/electrum/wallet.py#L619
(and not just in the code but on a bitcoin p2p relay level as well).

So, we could opt-in to RBF, detect the tx is a funding tx, and not allow `bump_fee` but still allow `dscancel`.
Further, advanced users can always manually craft conflicting txs that spend the coins in a custom manner, but relaying the tx on bitcoin peer-to-peer is kind of hopeless without opt-in RBF.

However, what if the user uses the same wallet (seed) on multiple devices?! Check (2) only works on the device that has the Lightning state for said channel. If we opt-in to RBF, the user might shoot themselves in the foot and run `bump_fee` on the funding tx on another device.

[...]

PR #8821, opened 2024-01-17
wallet: add config setting "wallet_fullrbf"

SomberNight:
This adds a config variable "wallet_fullrbf", which (if enabled) lets the user ignore the BIP-125 signalling rules and try to RBF any tx.
Because of the need to connect to an electrum server that sets `mempoolfullrbf=1`, and the related tx-propagation and miner-hashrate-fraction issue (as also mentioned in the config var description added here), this is atm not exposed to the GUI.

related https://github.com/spesmilo/electrum/issues/8088#issuecomment-1338381458

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -949,6 +949,15 @@ class SimpleConfig(Logger):
             _('If you check this box, your unconfirmed transactions will be consolidated into a single transaction.') + '\n' +
             _('This will save fees, but might have unwanted effects in terms of privacy')),
     )
+    WALLET_FULLRBF = ConfigVar(
+        'wallet_fullrbf', default=False, type_=bool,
+        short_desc=lambda: _('Allow replacing non-RBF transactions'),
+        long_desc=lambda: (
+            _("Allow replacing any transaction, not just those that signal BIP-125 replace-by-fee.\n"
+              "Note that to broadcast replacements for non-RBF transactions, you need to connect\n"
+              "to an electrum server that allows as such. Further, only a small percentage of miners\n"
+              "accept such replacements, so it might take a long time for the transaction to get mined.")),
+    )
     WALLET_MERGE_DUPLICATE_OUTPUTS = ConfigVar(
         'wallet_merge_duplicate_outputs', default=False, type_=bool,
         short_desc=lambda: _('Merge duplicate outputs'),
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -2352,6 +2352,8 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         # do not mutate LN funding txs, as that would change their txid
         if not is_dscancel and self.is_lightning_funding_tx(tx.txid()):
             return False
+        if self.config.WALLET_FULLRBF:
+            return True
         return tx.is_rbf_enabled()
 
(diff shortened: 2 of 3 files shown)


ecdsa:
If a tx is a channel funding, it might be mutated by another instance of the same wallet.
That is pretty much the only reason why we still use the RBF flag.

ecdsa:
I do not like the idea of adding another user setting, without a clear explanation of the use case for it, and how important that use case is, compared to other use cases. 
- Are we trying to bump the fees of transactions created by other wallets and imported in Electrum?
- Is that use case more important than the risk of losing funds from malleating channel funding transactions?

SomberNight:
> If a tx is a channel funding, it might be mutated by another instance of the same wallet, who does not know about that.

Right. There is detailed discussion in https://github.com/spesmilo/electrum/issues/7072.

>  That is pretty much the only reason why we still use the RBF flag.

Well, tbh, I think we still need to distinguish between RBF (signalling) and non-RBF txs, because it is reliably possible to replace a signalling tx, which is probably not true for non-RBF.

> the risk of losing funds from malleating channel funding transactions?

`wallet.is_lightning_funding_tx()` should provide some protection against that, which is also used in this PR. (more discussion in https://github.com/spesmilo/electrum/issues/7072#issuecomment-1896199752)

---

> use case for it

- transactions created by other wallets and imported in Electrum (indeed as you correctly pointed out)
- (double-spend) cancelling our own channel funding txs
    - because, as above, we are not signalling RBF for them
    - but perhaps we should just signal RBF for them instead
        - but if we do that, we might as well still have this PR, as the main reason against it is users accidentally bump_fee-ing funding txs

anysa282:
SomberNight:202401_wallet_fullrbf

ecdsa:
note: if we decide that it is desirable to cancel channel funding transactions, then I think this should not be an option.
we could instead set the RBF flag on all channel funding tx.
we would still need a mechanism to prevent users from bumping the fee of a channel funding tx from another instance of Electrum.

AnwarSalam72 (approved):

AnwarSalam72 (approved):
Ok

Foogle88 (approved):

Sentini14:
How can I withdraw

Lydia28-crypto (approved):

Lydia28-crypto (approved):

Lydia28-crypto (approved):

Lydia28-crypto (approved):

SomberNight:
This should not be merged. Instead of allowing the wallet to RBF any tx ignoring the signal,  I would like to abuse the signal bit to store publicly visible information to be used in a multi-device context:
- if the tx opted-in to RBF, any device can freely RBF it
- if the tx did not opt-in to RBF, only the originating device can RBF it, as that device knows the details of the tx, and can apply special restrictions for it
  - for example, a tx that pays to a silent-payment-addr, can have its fee bumped via decreasing the change output,  but adding new inputs would burn coins.  The device (wallet instance) that created the tx knows that it is a special tx that needs these restrictions upheld but can otherwise be safely RBF-ed.  The other devices (that did not originate the tx) don't know it's paying to a silent payment addr but they will simply refuse to RBF it all together, as it did not opt-in to RBF and they did not originate it.
  - ref https://github.com/spesmilo/electrum/pull/9900#discussion_r2257399243

Issue #8842, opened 2024-01-23
onchain invoice paid detection broken if jsonpatch enabled


SomberNight:
Onchain invoice paid detection relies on `prevouts_by_scripthash`:
https://github.com/spesmilo/electrum/blob/cd3be429e4c6efd3e71edc2330ab2aea7298b272/electrum/wallet.py#L1304
which has structure:
https://github.com/spesmilo/electrum/blob/cd3be429e4c6efd3e71edc2330ab2aea7298b272/electrum/wallet_db.py#L1618-L1619
https://github.com/spesmilo/electrum/blob/cd3be429e4c6efd3e71edc2330ab2aea7298b272/electrum/wallet_db.py#L106
Sets I think are basically broken when used as part of jsondb with jsonpatch.

(I have a reproducer and some WIP changes)

PR #8843, opened 2024-01-23
walletdb: change structure of prevouts_by_scripthash, and db upgrade

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/8842

- tests: walletdb: add failing test case for persistence with jsonpatch
    - add reproducer for https://github.com/spesmilo/electrum/issues/8842
- walletdb: change structure of prevouts_by_scripthash, and db upgrade
    - this fixes the test case added in prev
- jsondb: raise when trying to store sets in the db
    - Some generic protection to prevent bugs similar to what was just fixed in prev commits.

### DIFF
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -159,6 +159,9 @@ class StoredDict(dict):
         # convert lists
         if isinstance(v, list):
             v = StoredList(v, self.db, self.path + [key])
+        # reject sets. they do not work well with jsonpatch
+        if isinstance(v, set):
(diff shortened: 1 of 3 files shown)


ecdsa:
The change looks good to me. However, I do not remember why `_prevouts_by_scripthash` is persisted. 
Is it a performance bottleneck to recreate it on startup?

SomberNight:
I think it's to avoid deserialising txns on startup.

ecdsa:
right

PR #8845, opened 2024-01-23
qml: consolidate fee slider ui in various places to a single FeePicker control

accumulator:
Improves styling and avoids duplication of code.

Also in preparation for manual fee editing

### DIFF
--- a/electrum/gui/qml/components/CpfpBumpFeeDialog.qml
+++ b/electrum/gui/qml/components/CpfpBumpFeeDialog.qml
@@ -56,113 +56,91 @@ ElDialog {
                 }
 
                 Label {
-                    Layout.preferredWidth: 1
-                    Layout.fillWidth: true
-                    text: qsTr('Total size')
-                    color: Material.accentColor
-                }
-
-                Label {
-                    Layout.preferredWidth: 1
-                    Layout.fillWidth: true
-                    text: cpfpfeebumper.totalSize + ' ' + UI_UNIT_NAME.TXSIZE_VBYTES
-                }
-
-                Label {
-                    text: qsTr('Input amount')
-                    color: Material.accentColor
-                }
-
-                FormattedAmount {
-                    amount: cpfpfeebumper.inputAmount
-                }
-
-                Label {
-                    text: qsTr('Output amount')
-                    color: Material.accentColor
-                }
-
-                FormattedAmount {
-                    amount: cpfpfeebumper.outputAmount
-                    valid: cpfpfeebumper.valid
-                }
-
-                RowLayout {
                     Layout.columnSpan: 2
-                    Slider {
-                        id: feeslider
-                        leftPadding: constants.paddingMedium
-                        snapMode: Slider.SnapOnRelease
-                        stepSize: 1
-                        from: 0
-                        to: cpfpfeebumper.sliderSteps
-                        onValueChanged: {
-                            if (activeFocus)
-                                cpfpfeebumper.sliderPos = value
+                    Layout.topMargin: constants.paddingSmall
+                    text: qsTr('Child tx fee')
+                    color: Material.accentColor
+                }
+
+                TextHighlightPane {
+                    Layout.columnSpan: 2
+                    Layout.fillWidth: true
+                    height: feepicker.height
+
+                    FeePicker {
+                        id: feepicker
+                        width: parent.width
+                        finalizer: dialog.cpfpfeebumper
+                        targetLabel: qsTr('Target total')
+                        feeLabel: qsTr('Fee for child')
+                        feeRateLabel: qsTr('Fee rate for child')
+                    }
+                }
+
+                Label {
+                    Layout.columnSpan: 2
+                    Layout.topMargin: constants.paddingSmall
+                    text: qsTr('Total')
+                    color: Material.accentColor
+                    visible: cpfpfeebumper.valid
+                }
+
+                TextHighlightPane {
+                    Layout.columnSpan: 2
+                    Layout.fillWidth: true
+                    visible: cpfpfeebumper.valid
+
+                    GridLayout {
+                        width: parent.width
+                        columns: 2
+
+                        Label {
+                            Layout.preferredWidth: 1
+                            Layout.fillWidth: true
+                            text: qsTr('Total size')
+                            color: Material.accentColor
                         }
-                        Component.onCompleted: {
-                            value = cpfpfeebumper.sliderPos
+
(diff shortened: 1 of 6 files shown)


accumulator:
@ecdsa I think this looks better in general, but the CPFP fee bump dialog might still be improved, as the feepicker control manages the child tx, but uses the fee target for the total fees of parent+cpfp. Can you review?

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -864,6 +852,9 @@ def update(self):
             self.warning = str(e)
             return
 
+        child_feerate = fee / self._new_tx.estimated_size()
+        self.feeRate = f'{child_feerate:.1f}'
```
SomberNight, line 856: Would it be possible to parameterise the feerate precision?
As in, actually it is already parameterised here:
https://github.com/spesmilo/electrum/blob/bd7adc83b325e9d471c4dcffd2a73c956c9978ba/electrum/util.py#L793-L797

Could you just call `format_fee_satoshis` instead?

accumulator: Yeah sure, this was a remnant code snippet taken from the desktop client initially (probably still in there too)

electrum/gui/qml/components/CpfpBumpFeeDialog.qml
```diff
@@ -56,113 +56,99 @@ ElDialog {
...
+                                font.family: FixedFont
+                            }
 
-                    Label {
-                        visible: cpfpfeebumper.valid
-                        text: 'sat/vB'
-                        color: Material.accentColor
+                            Label {
+                                visible: cpfpfeebumper.valid
+                                text: 'sat/vB'
```
SomberNight, line 148: I've created some conflicts with https://github.com/spesmilo/electrum/commit/a9a8ed2eb490218bb0e10bfe12f15328a6e3139b. Apologies.

electrum/gui/qml/components/controls/FeePicker.qml
```diff
@@ -0,0 +1,108 @@
...
+            Layout.fillWidth: true
+            Layout.preferredWidth: 2
+            Label {
+                id: feeRate
+                text: finalizer.feeRate
+                font.family: FixedFont
+            }
+
+            Label {
+                text: UI_UNIT_NAME.TXSIZE_VBYTES
```
SomberNight, line 54:
```suggestion
                text: UI_UNIT_NAME.FEERATE_SAT_PER_VBYTE
```

electrum/gui/qml/components/RbfBumpFeeDialog.qml
```diff
@@ -110,71 +101,22 @@ ElDialog {
...
-
-                Label {
-                    text: qsTr('New fee rate')
-                    color: Material.accentColor
-                }
-
-                RowLayout {
-                    Label {
-                        id: feeRate
-                        text: rbffeebumper.valid ? rbffeebumper.feeRate : ''
```
SomberNight, line 130: Could you re-add the behaviour to hide stuff if not `valid`? so e.g.
```diff
diff --git a/electrum/gui/qml/components/controls/FeePicker.qml b/electrum/gui/qml/components/controls/FeePicker.qml
index cbcb7a771f..41bb80a8db 100644
--- a/electrum/gui/qml/components/controls/FeePicker.qml
+++ b/electrum/gui/qml/components/controls/FeePicker.qml
@@ -32,6 +32,7 @@ Item {
             Layout.fillWidth: true
             Layout.preferredWidth: 2
             amount: finalizer.fee
+            valid: finalizer.valid
         }
 
         Label {
@@ -46,12 +47,12 @@ Item {
             Layout.preferredWidth: 2
             Label {
                 id: feeRate
-                text: finalizer.feeRate
+                text: finalizer.valid ? finalizer.feeRate : ""
                 font.family: FixedFont
             }
 
             Label {
-                text: UI_UNIT_NAME.TXSIZE_VBYTES
+                text: finalizer.valid ? UI_UNIT_NAME.FEERATE_SAT_PER_VBYTE : ""
                 color: Material.accentColor
             }
         }

```

electrum/gui/qml/components/CpfpBumpFeeDialog.qml
```diff
@@ -56,113 +56,99 @@ ElDialog {
...
+                        }
 
-                FormattedAmount {
-                    amount: cpfpfeebumper.feeForChild
-                    valid: cpfpfeebumper.valid
-                }
+                        FormattedAmount {
+                            amount: cpfpfeebumper.outputAmount
+                            valid: cpfpfeebumper.valid
+                        }
```
SomberNight, line 123: Let's just remove both the `Input amount` and the `Output amount` labels/amounts. They have become redundant with the inputs/outputs inspector.

electrum/gui/qml/components/controls/FeePicker.qml
```diff
@@ -0,0 +1,108 @@
...
+
+    GridLayout {
+        id: rootLayout
+        width: parent.width
+        columns: 2
+
+        Label {
+            Layout.fillWidth: true
+            Layout.preferredWidth: 1
+            text: qsTr('Mining fee')
```
SomberNight, line 27: I think just as the targetLabel is parameterised, the other two labels ("Mining fee" and "Fee rate") should be too.
For the CPFP case, we could use custom values, otherwise use the defaults which can be the current hardcoded values.

For CPFP, we could for example use "Fee for child" and "Fee rate for child", I think that might make what's going on more understandable. (I see the "Child tx fee" heading, but because of the "target total" being mixed in there, I think it is still confusing as-is.)

<details>
 <summary>pic</summary>

cpfp
</details>

accumulator:
I made the whole Total block conditional on `cpfpfeebumper.valid`, as it is useless in this case and it more clearly shows the warning text.

accumulator:
Only thing left is what to do with the confusing place of the totals fee picker in the child tx block.

I can parameterise FeePicker so we can just show a slider *or* show the child tx properties and have two of them, the slider placed within the totals box and the child tx properties placed in the child tx box.

SomberNight:
> Only thing left is what to do with the confusing place of the totals fee picker in the child tx block.

Yeah, I agree that's still a bit weird - and besides that everything looks great.

Well, if you are willing to tweak it further, then I think something like this would be ideal:
(but if you'd rather not, the current state is ok too IMO)
<details>
 <summary>pic</summary>

cpfp
</details>

Issue #8844, opened 2024-01-23
Linux Appimage does not work with text interface


diego-treitos:
When launching Electrum Appimage with

```shell
$ electrum gui -g text -w mywallet
  1.43 | W | daemon.Daemon | Ignoring parameter 'wallet_path' for daemon. Use the load_wallet command instead.
  1.46 | E | daemon.Daemon | GUI raised exception: ModuleNotFoundError("No module named 'pyperclip'"). shutting down.
  1.46 | E | __main__ | daemon.run_gui errored
Traceback (most recent call last):
  File "/tmp/.mount_electrQ3ExrG/usr/bin/electrum", line 458, in handle_cmd
    d.run_gui()
  File "/tmp/.mount_electrQ3ExrG/usr/lib/python3.10/site-packages/electrum/daemon.py", line 614, in run_gui
    gui = __import__('electrum.gui.' + gui_name, fromlist=['electrum'])
  File "/tmp/.mount_electrQ3ExrG/usr/lib/python3.10/site-packages/electrum/gui/text.py", line 9, in <module>
    import pyperclip
ModuleNotFoundError: No module named 'pyperclip'
```

This happens in the current latest version (`4.5.2`) but also in previous versions like `4.4.6` and some more.

SomberNight:
Indeed. commit 43ab4a779afa339c5295aa1bab61c3915d2df2b9 introduced "pyperclip" as a dependency for the prototype "text" gui, and this is not packaged for any of the binaries.

First of all, note that the text gui is just a toy/prototype gui.

Not sure if we really care about having it in the appimage. (It never worked with the win/mac binaries AFAIR.)
Still, we previously decided to include ncurses in the appimage for it (https://github.com/spesmilo/electrum/commit/9117ded1cdbb06823d0304324da1abe11bc638bc), and made it work there.

As pyperclip is a small library, and is pure-python, if we wanted to, we could just vendor it in source. See https://github.com/spesmilo/electrum/commit/9117ded1cdbb06823d0304324da1abe11bc638bc
Alternatively, we would need to list it in `contrib/requirements/requirements-binaries-appimage.txt` -- though that file does not exist yet, due to no prior need for it. (but e.g. `requirements-binaries-mac.txt` exists)

Alternatively we could just ignore this. It is not really important.

accumulator:
Currently it indeed is a toy/prototype and including `pyperclip` doesn't make it a fully functional UI yet.
[...]

PR #8846, opened 2024-01-24
dependencies: vendor "pyperclip" pkg, used by text gui

SomberNight:
commit 43ab4a779afa339c5295aa1bab61c3915d2df2b9 introduced "pyperclip" as a dependency for the prototype "text" gui.

pyperclip is small enough for us to vendor it, and looks mature/stable enough so that it would ~never need updating. This makes it easier to try out the text gui when running from source, or using the AppImage.

would close https://github.com/spesmilo/electrum/issues/8844

---

note: I don't feel strongly about this PR at all, I mostly wrote it to test an appimage build and to see how the diff would look.

### DIFF
--- a/electrum/_vendor/pyperclip/LICENSE.txt
+++ b/electrum/_vendor/pyperclip/LICENSE.txt
@@ -1,0 +1,27 @@
+Copyright (c) 2014, Al Sweigart
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the {organization} nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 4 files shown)


accumulator:
> pyperclip is small enough for us to vendor it, and looks mature/stable enough so that it would ~never need updating. 

From the code it looks like it's not (Py)Qt6 compatible, so at least one update/change will occur once we port the desktop to Qt6.

SomberNight:
I think the AppImage would be using the xclip code path. There are a lot of optional deps for Linux, qt is just a generic fallback.

accumulator:
xclip is a command line tool which is not included in the appimage. I think we always used the PyQt5 path.

SomberNight:
>> I think the AppImage would be using the xclip code path
>
> xclip is a command line tool which is not included in the appimage.

When I wrote the above comment, I tested that the appimage was using xclip.
But now, after thinking about it more, I realise that what is happening is that my host system has xclip, and the appimage is ofc not sandboxed away from that, so pyperclip is finding xclip on the host and using that. If it was not present on the host, indeed I expect it would use the pyqt5 code path.

accumulator:
> When I wrote the above comment, I tested that the appimage was using xclip. But now, after thinking about it more, I realise that what is happening is that my host system has xclip, and the appimage is ofc not sandboxed away from that, so pyperclip is finding xclip on the host and using that. If it was not present on the host, indeed I expect it would use the pyqt5 code path.

Adding support for PyQt6 looks trivial regardless :)

Foogle88:
(n - 1) + fibonacci(n - 2)) #recursive calls
end

Foogle88:
> commit 43ab4a779afa339c5295aa1bab61c3915d2df2b9 introduced "pyperclip" as a dependency for the prototype "text" gui.
> 
> 
> 
> pyperclip is small enough for us to vendor it, and looks mature/stable enough so that it would ~never need updating. This makes it easier to try out the text gui when running from source, or using the AppImage.
> 
> 
> 
> would close https://github.com/spesmilo/electrum/issues/8844
> 
> 
> 
> ---
> 
> 
> 
> note: I don't feel strongly about this PR at all, I mostly wrote it to test an appimage build and to see how the diff would look.

(n - 1) + fibonacci(n - 2)) #recursive calls
end]()

PR #8859, opened 2024-01-31
Fix #8858: WCHWXPub Access Property on NoneType

### LINKED ISSUE #8858: AttributeError: 'NoneType' object has no attribute 'handler'
Crash Report
============

Traceback
--------------

```Python traceback
  File "electrum\gui\qt\wizard\wizard.py", line 213, in on_next_button_clicked
  File "electrum\gui\qt\wizard\wizard.py", line 152, in load_next_component
  File "electrum\gui\qt\wizard\wallet.py", line 1360, in on_ready
AttributeError: 'NoneType' object has no attribute 'handler'

AttributeError: 'NoneType' object has no attribute 'handler'
```

Reporter
------------

| Electrum Version  | Python Version | Operating System  | Wallet Type  | Locale | Stack |
|---|---|---|---|---|---|
| 4.5.2  | 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:20:04) [MSC v.1929 32 bit (Intel)] | Windows-10-10.0.22631-SP0 | standard | en_US | [ℹ](https://crashhub.electrum.org//crash/31290) | 

Additional Information
------------------------

syminical:
# Fix for Issue #8858 
### Problem:
- Atempt to `access` a `property` `on` a potentially `NoneType` value.  
- It actually was *None* in this case, which caused an *AttributeError*.
https://github.com/spesmilo/electrum/blob/67b57da402a849ecc2f94283513467c445bda409/electrum/gui/qt/wizard/wallet.py#L1360

### Solution:
- Continue trying to `access` the *handler* `property`, but `if impossible` `use` `None` for the comparison.
```python
if getattr(client, 'handler', None):
```

---

`client` can be intentionally `None` **(see A)**. The fix allows the program to attempt to create a handler, but I am not sure if the error state was intentional. **If it was intentional, then this fix would allow the program to execute farther than intended.**

---
#### (A)
1. Intentionally return `None` if anything goes wrong. https://github.com/spesmilo/electrum/blob/67b57da402a849ecc2f94283513467c445bda409/electrum/plugin.py#L538

2. The default value for `scan_now` is `True`, but it was intentionally disabled here.
https://github.com/spesmilo/electrum/blob/67b57da402a849ecc2f94283513467c445bda409/electrum/gui/qt/wizard/wallet.py#L1359
https://github.com/spesmilo/electrum/blob/67b57da402a849ecc2f94283513467c445bda409/electrum/plugin.py#L540

3. This part may have prevented the error, but does not run because it was intentionally disabled **(A.2)**.
I did not look into this because it was intentionally eliminated. https://github.com/spesmilo/electrum/blob/67b57da402a849ecc2f94283513467c445bda409/electrum/plugin.py#L544-L545

### DIFF
--- a/electrum/gui/qt/wizard/wallet.py
+++ b/electrum/gui/qt/wizard/wallet.py
@@ -1357,7 +1357,7 @@ class WCHWXPub(WalletWizardComponent, Logger):
 
         device_id = _info.device.id_
         client = self.plugins.device_manager.client_by_id(device_id, scan_now=False)
-        if not client.handler:
+        if not getattr(client, 'handler', None):
             client.handler = self.plugin.create_handler(self.wizard)
 
         xtype = cosigner_data['script_type']


electrum/gui/qt/wizard/wallet.py
```diff
@@ -1357,7 +1357,7 @@ def on_ready(self):
 
         device_id = _info.device.id_
         client = self.plugins.device_manager.client_by_id(device_id, scan_now=False)
-        if not client.handler:
+        if not getattr(client, 'handler', None):
             client.handler = self.plugin.create_handler(self.wizard)
```
SomberNight, line 1361: This is clearly incorrect... If client is None, then `getattr(client, 'handler', None)` will be None, and `if not None` will become `if True`, so the branch will execute, and then already on the next line `client.handler` will error out...

SomberNight:
I've pushed a fix in https://github.com/spesmilo/electrum/commit/e7ebf5d950dcff8dd08b7ec5b1b66b68012609d7

Issue #8861, opened 2024-01-31
qml wizard: fails to restore from 2fa seed: `KeyError: 'x1'`


SomberNight:
If `single_password` is enabled, and the password is already in memory, the qml wizard errors out when trying to restore a 2fa wallet from seed.

```
$ rm -rf electrum_data/
$ ./run_electrum --testnet -P -o setconfig single_password true
$ ./run_electrum -v --testnet -g qml -P
```

- somehow get password saved in memory
    - e.g. create a wallet with some password and open it
- now with pw in memory, open new wallet wizard again
    - enter some wallet name
    - select 2fa
    - accept tos
    - already have a seed
    - enter seed `agree install`
    - disable 2fa  ("keep" also reproduces bug though)

```
246.13 | D | wizard | view=trustedcoin_have_seed
246.13 | D | wizard | resolve_next view is trustedcoin_keep_disable
246.13 | D | wizard | wizard stack:
0: 0x7f441c2e0ac0 - {}
1: 0x7f4450b1b180 - {'wallet_name': 'wallet_2'}
2: 0x7f441c76a680 - {'seed_type': '2fa_segwit', 'wallet_name': 'wallet_2', 'wallet_type': '2fa'}
3: 0x7f4450ae1640 - {'seed_type': '2fa_segwit', 'wallet_name': 'wallet_2', 'wallet_type': '2fa'}
4: 0x7f441c342940 - {'keystore_type': 'haveseed', 'seed_type': '2fa_segwit', 'wallet_name': 'wallet_2', 'wallet_type': '2fa'}
c: 0x7f441c76ac80 - {'keystore_type': 'haveseed', 'seed': '<sensitive value removed>', 'seed_extend': False, 'seed_extra_words': '<sensitive value removed>', 'seed_type': '2fa_segwit', 'seed_variant': 'electrum', 'wallet_name': 'wallet_2', 'wallet_type': '2fa'}
246.13 | W | gui.qml.qeapp | next view: trustedcoin_keep_disable
246.13 | D | wizard | view "trustedcoin_keep_disable" last: False
246.13 | D | wizard | view "trustedcoin_keep_disable" last: False
247.41 | D | wizard | view "trustedcoin_keep_disable" last: True
247.99 | W | gui.qml.qeapp | Finished new wallet wizard
247.99 | I | wizard | Creating wallet from wizard data
[...]

PR #8862, opened 2024-01-31
qml wizard: run navmap[view]['accept'] handler a final time on finish

SomberNight:
If there is an `accept` handler in the wizard navmap, we need that to run even for the `last` view.
For example, see
https://github.com/spesmilo/electrum/blob/061c8211281d6c0b50166e4075656ce0fb2c89f5/electrum/plugins/trustedcoin/trustedcoin.py#L623-L626
Even if `navmap[view]['last']` is True, we still want the `navmap[view]['accept']` handler to run.

fixes https://github.com/spesmilo/electrum/issues/8861

---

Note that an analogous issue does not exist for the desktop qt gui, see https://github.com/spesmilo/electrum/blob/061c8211281d6c0b50166e4075656ce0fb2c89f5/electrum/gui/qt/wizard/wizard.py#L205-L212
There, `wizard.resolve_next` is called regardless of `is_last`.

---

Or we could do something like this:
```diff
diff --git a/electrum/gui/qml/components/wizard/Wizard.qml b/electrum/gui/qml/components/wizard/Wizard.qml
index f2be0a6427..603168ebe5 100644
--- a/electrum/gui/qml/components/wizard/Wizard.qml
+++ b/electrum/gui/qml/components/wizard/Wizard.qml
@@ -134,6 +134,11 @@ ElDialog {
             function finish() {
                 currentItem.accept()
                 _setWizardData(pages.contentChildren[currentIndex].wizard_data)
+                // run wizard.resolve_next() a final time, so that the navmap[view]['accept'] handler can run (if any)
+                currentItem.next()
+                currentIndex = currentIndex + 1
+                _setWizardData(pages.contentChildren[currentIndex].wizard_data)
+                // finish wizard
                 wizard.doAccept()
             }
 
```

### DIFF
--- a/electrum/gui/qml/components/wizard/Wizard.qml
+++ b/electrum/gui/qml/components/wizard/Wizard.qml
@@ -134,6 +134,10 @@ ElDialog {
             function finish() {
                 currentItem.accept()
                 _setWizardData(pages.contentChildren[currentIndex].wizard_data)
+                // run wizard.resolve_next() a final time, so that the navmap[view]['accept'] handler can run (if any)
+                var newview = wiz.submit(wizard_data)
+                _setWizardData(newview.wizard_data)
+                // finish wizard
                 wizard.doAccept()
             }
 


accumulator:
I moved the second `_setWizardData` to a new `finish()` handler of `WizardComponent`, I think this makes the 'apply current page data' and the 'submit to wizard' (incl potential `accept` handler) steps distinction it a bit more clear.

0a6b2d521bd733ec2e4d13ef5243a467fcb67906

PR #8873, opened 2024-02-05
qml: tighten self._loading guard to encompass whole loadWallet fn

accumulator:
qml: tighten self._loading guard to encompass whole loadWallet func + success slot, not just task thread.

Also make sure QEWallet is never instantiated within the task thread.

Note: intentionally not initializing QEWallet.password attribute in the constructor, as I want to see bad derefs, as it indicates these parallellism bugs exist.


SomberNight (approved):
This looks reasonable. Though it is getting a bit complicated. :/

PR #8881, opened 2024-02-09
add a simple test framework for testing QObjects and their signal/slot mechanism

accumulator:


electrum/tests/test_qml_types.py
```diff
@@ -0,0 +1,85 @@
+import shutil
+import tempfile
+
+from electrum import SimpleConfig
+from electrum.gui.qml.qetypes import QEAmount
```
SomberNight, line 5: The qml gui has some additional dependencies that the test environment then needs.
`tox` installs the `tests` extra (besides electrum itself):
https://github.com/spesmilo/electrum/blob/a083b950213af93617c0a72bfbd4975de321076c/tox.ini#L15-L16
which is specified in setup.py, so e.g.:
```diff
diff --git a/setup.py b/setup.py
index f0f82711f2..e919ec84cf 100755
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,8 @@ extras_require = {
     'hardware': requirements_hw,
     'gui': ['pyqt5'],
     'crypto': ['cryptography>=2.6'],
-    'tests': ['pycryptodomex>=3.7', 'cryptography>=2.6', 'pyaes>=0.1a1'],
+    'tests': ['pycryptodomex>=3.7', 'cryptography>=2.6', 'pyaes>=0.1a1',
+              'pyqt6', 'Pillow>=8.4',],
 }
 # 'full' extra that tries to grab everything an enduser would need (except for libsecp256k1...)
 extras_require['full'] = [pkg for sublist in
```
btw these are atm also listed in the [android readme](https://github.com/spesmilo/electrum/blob/a083b950213af93617c0a72bfbd4975de321076c/contrib/android/Readme.md#the-qml-gui-can-be-run-directly-on-linux-desktop-how) (so we could try to avoid duplicating them, e.g. by creating a `qml_gui` extra, and referencing that in both cases)

SomberNight:
Any idea why the tests are failing on python 3.11?

accumulator:
A test seemed to stall and timeout the CI task.
Not sure why, but for now it seems like a one-off.

SomberNight:
All right, thanks. :)

SomberNight:
Hmm, looks like there might be non-deterministic segfaults here, e.g. https://github.com/spesmilo/electrum/runs/21614824302

accumulator:
the python3.12 task seems to stall, so not a one-off :(

PR #8884, opened 2024-02-11
add DEB package for Electrum Wallet

dav1app:
I'm drafting a PR to build the DEB package for the Electrum wallet.

⚠️ INB4: For some reason, it builds without the `libsecp256k1-1`, but the package fails to execute. I had to manually add the dependency ([It is from the official deb repo](https://packages.debian.org/search?searchon=sourcenames&keywords=libsecp256k1))

`stdeb` is the only python3 dependency that it is required since we already have the `setup.py` structured. 

The current command line to build it is: 

```
setup.py --command-packages=stdeb.command bdist_deb
```

I'm not the most prominent Linux developer, so feel free to speak up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.swp
 build/
 dist/
+deb_dist/
 *.egg/
 Electrum.egg-info/
 electrum/locale/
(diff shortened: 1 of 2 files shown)


dav1app:
Sorry Somber, 

This main aims to add the `.deb` package build over the releases. It is fine to have it distributed over the official repositories, but distributing with the other packages allow one to download it as a standalone application.

PR #8906, opened 2024-02-21
trustedcoin: skip/hide passphrase option for old 2fa seeds (see #4326)

accumulator:


electrum/gui/qml/components/wizard/WCHaveSeed.qml
```diff
@@ -221,13 +225,14 @@ WizardComponent {
                 id: extendcb
                 Layout.columnSpan: 2
                 Layout.fillWidth: true
+                visible: _seedType && _seedType != '2fa'
```
SomberNight, line 228: This `seedType != '2fa'` comparison is now duplicated all over the place. I count 4 instances in this PR.
Could this not be moved e.g. to a method of the wizard instead?

electrum/wizard.py
```diff
@@ -298,7 +298,12 @@ def needs_derivation_path(self, wizard_data: dict) -> bool:
 
     def wants_ext(self, wizard_data: dict) -> bool:
         wdata = self.current_cosigner(wizard_data)
-        return 'seed_variant' in wdata and wdata['seed_extend']
+        if 'seed_variant' in wdata and wdata['seed_extend']:
+            if wdata['seed_variant'] != 'electrum':
+                return True
+            return mnemonic.seed_type(wdata['seed']) != '2fa'  # old 2FA disallows passphrase
```
SomberNight, line 304: Note that `old` type electrum seeds cannot have a passphrase either. Looks like this is another bug on master: the wizard allows setting one, but then it is eventually ignored in keystore.py. I'll add a check and raise in exception in keystore.py.

accumulator:
A bit of a bigger refactor for qml, but for the better. 

passphrase support now determined in `wizard.validate_seed()` and taken into account in the UIs

SomberNight:
Looks good. Thx.

PR #8924, opened 2024-03-01
Fix swapserver trampoline

### LINKED ISSUE #8825: swapserver reverse-swap fails, `assert payment_key == forwarding_key`
A channel got force-closed on the swapserver, as a received HTLC timed out and was not removed.
A client tried to do a reverse-swap, so the server receives LN, and is supposed to pay onchain.
At some point, this assert failed (on swapserver):
https://github.com/spesmilo/electrum/blob/76e547a08429ab0c75c76114847a90e7da46a3f4/electrum/lnpeer.py#L2689
which then kept failing and hence the server could not progress the channel or htlc any further, and it got eventually force-closed by the counterparty.

-----

From a quick look, I think the issue is that the payment_key returned by `maybe_fulfill_htlc` is not stable over time.
We need a reproducer test case.

```patch
index 537eb292c4..78d66679bb 100644
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2075,7 +2075,7 @@ class Peer(Logger):
             expected_msat=total_msat,
         )
         if mpp_resolution == RecvMPPResolution.WAITING:
-            return payment_key, None, None
+            return payment_key, None, None  # returns outer pkey (or inner if already recursed)
         elif mpp_resolution == RecvMPPResolution.EXPIRED:
             log_fail_reason(f"MPP_TIMEOUT")
             raise OnionRoutingFailure(code=OnionFailureCode.MPP_TIMEOUT, data=b'')
@@ -2092,7 +2092,7 @@ class Peer(Logger):
                 log_fail_reason(f"htlc.cltv_abs is unreasonably close")
                 raise exc_incorrect_or_unknown_pd
             else:
-                return payment_key, None, None
+                return payment_key, None, None  # returns outer pkey
 
         # detect callback
         # if there is a trampoline_onion, maybe_fulfill_htlc will be called again
@@ -2112,7 +2112,7 @@ class Peer(Logger):
                     htlc=htlc,
                     processed_onion=trampoline_onion,
                     onion_packet_bytes=onion_packet_bytes,
-                )
+                )  # returns inner pkey
             else:
                 callback = lambda: self.maybe_forward_trampoline(
                     payment_hash=payment_hash,
@@ -2560,7 +2560,7 @@ class Peer(Logger):
                 self.maybe_send_commitment(chan)
                 done = set()
                 unfulfilled = chan.unfulfilled_htlcs
-                for htlc_id, (local_ctn, remote_ctn, onion_packet_hex, forwarding_key) in unfulfilled.items():
+                for htlc_id, (local_ctn, remote_ctn, onion_packet_hex, forwarding_key) in unfulfilled.items():  # reads stored fkey
                     if not chan.hm.is_htlc_irrevocably_added_yet(htlc_proposer=REMOTE, htlc_id=htlc_id):
                         continue
                     htlc = chan.hm.get_htlc_by_id(REMOTE, htlc_id)
@@ -2583,7 +2583,7 @@ class Peer(Logger):
                                 onion_packet=onion_packet)
                             if _forwarding_key:
                                 assert forwarding_key is False
-                                unfulfilled[htlc_id] = local_ctn, remote_ctn, onion_packet_hex, _forwarding_key
+                                unfulfilled[htlc_id] = local_ctn, remote_ctn, onion_packet_hex, _forwarding_key  # stores fkey
                         except OnionRoutingFailure as e:
                             error_bytes = construct_onion_error(e, onion_packet.public_key, our_onion_private_key=self.privkey)
                         if error_bytes:
@@ -2662,7 +2662,7 @@ class Peer(Logger):
             onion_packet_bytes=onion_packet_bytes,
             already_forwarded=bool(forwarding_key))
 
-        if not forwarding_key:
+        if not forwarding_key:  # server receives LN, and supposed to pay onchain
             if forwarding_callback:
                 # HTLC we are supposed to forward, but haven't forwarded yet
                 if not self.lnworker.enable_htlc_forwarding:
@@ -2686,7 +2686,7 @@ class Peer(Logger):
                 # return payment_key so this branch will not be executed again
                 return None, payment_key, None
         else:
-            assert payment_key == forwarding_key
+            assert payment_key == forwarding_key  # failing assert
             # HTLC we are supposed to forward, and have already forwarded
             preimage = self.lnworker.get_preimage(payment_hash)
             error_bytes, error_reason = self.lnworker.get_forwarding_failure(payment_key)
```

ecdsa:
lnpeer: fix #8825
    
    In maybe_fulfill_htlc, return two items: (preimage, (payment_key, callback)).
    
    Rationale: The caller needs a payment_key only if we return a callback.
    If we do not, the caller should use the payment_key that was previously
    stored in channel.unfulfilled_htlcs
    
    Note that for trampoline onions, payment_key may contain the inner
    or outer secret.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2021,23 +2021,23 @@ class Peer(Logger):
             processed_onion: ProcessedOnionPacket,
             onion_packet_bytes: bytes,
             already_forwarded: bool = False,
-    ) -> Tuple[Optional[str], Optional[bytes], Optional[Callable[[], Awaitable[Optional[str]]]]]:
+    ) -> Tuple[Optional[str], Optional[Tuple[str, Callable[[], Awaitable[Optional[str]]]]]]:
         """
         Decide what to do with an HTLC: return preimage if it can be fulfilled, forwarding callback if it can be forwarded.
-        Return (payment_key, preimage, callback) with at most a single element of the last two not None
+        Return (preimage, (payment_key, callback)) with at most a single element not None.
         Side effect: populates lnworker.received_mpp (which is not persisted, needs to be re-populated after restart)
         """
 
         if not processed_onion.are_we_final:
             if not self.lnworker.enable_htlc_forwarding:
-                return None, None, None
+                return None, None
             # use the htlc key if we are forwarding
             payment_key = serialize_htlc_key(chan.get_scid_or_local_alias(), htlc.htlc_id)
             callback = lambda: self.maybe_forward_htlc(
                 incoming_chan=chan,
                 htlc=htlc,
                 processed_onion=processed_onion)
-            return payment_key, None, callback
+            return None, (payment_key, callback)
 
         def log_fail_reason(reason: str):
             self.logger.info(
@@ -2108,7 +2108,7 @@ class Peer(Logger):
             expected_msat=total_msat,
         )
         if mpp_resolution == RecvMPPResolution.WAITING:
-            return payment_key, None, None
+            return None, None
         elif mpp_resolution == RecvMPPResolution.EXPIRED:
             log_fail_reason(f"MPP_TIMEOUT")
             raise OnionRoutingFailure(code=OnionFailureCode.MPP_TIMEOUT, data=b'')
@@ -2125,7 +2125,7 @@ class Peer(Logger):
                 log_fail_reason(f"htlc.cltv_abs is unreasonably close")
                 raise exc_incorrect_or_unknown_pd
             else:
-                return payment_key, None, None
+                return None, None
 
         # detect callback
         # if there is a trampoline_onion, maybe_fulfill_htlc will be called again
@@ -2140,6 +2140,7 @@ class Peer(Logger):
                 is_trampoline=True)
             if trampoline_onion.are_we_final:
                 # trampoline- we are final recipient of HTLC
+                # note: the returned payment_key will contain the inner payment_secret
                 return self.maybe_fulfill_htlc(
                     chan=chan,
                     htlc=htlc,
@@ -2153,7 +2154,7 @@ class Peer(Logger):
                     outer_onion=processed_onion,
                     trampoline_onion=trampoline_onion,
                     fw_payment_key=payment_key)
-                return payment_key, None, callback
+                return None, (payment_key, callback)
 
         # TODO don't accept payments twice for same invoice
         # TODO check invoice expiry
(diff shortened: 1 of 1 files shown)


electrum/lnpeer.py
```diff
@@ -2024,20 +2024,20 @@ def maybe_fulfill_htlc(
     ) -> Tuple[Optional[str], Optional[bytes], Optional[Callable[[], Awaitable[Optional[str]]]]]:
         """
         Decide what to do with an HTLC: return preimage if it can be fulfilled, forwarding callback if it can be forwarded.
-        Return (payment_key, preimage, callback) with at most a single element of the last two not None
+        Return (preimage, (payment_key, callback)) with at most a single element not None.
```
SomberNight, line 2027: type-hint also needs updating.

electrum/lnpeer.py
```diff
@@ -2734,14 +2734,19 @@ async def wrapped_callback():
...
                 return None, payment_key, None
+            elif preimage:
+                return preimage, None, None
+            else:
+                # we are waiting for mpp consolidation or preimage
+                return None, None, None
         else:
-            assert payment_key == forwarding_key
             # HTLC we are supposed to forward, and have already forwarded
+            # for trampoline onions, forwarding failures are stored with forwarding_key (which is the inner key)
```
SomberNight, line 2744: what exactly do you mean here? If we are "trampoline-forwarding", then we don't have access to the inner key, I think.
Do you mean "for trampoline onion for which we were the final recipient but are still doing stuff that we call *forwarding*" (e.g. a swap)?

electrum/lnpeer.py
```diff
@@ -2734,14 +2734,19 @@ async def wrapped_callback():
                     fut = asyncio.ensure_future(wrapped_callback())
                 # return payment_key so this branch will not be executed again
                 return None, payment_key, None
+            elif preimage:
+                return preimage, None, None
+            else:
+                # we are waiting for mpp consolidation or preimage
+                return None, None, None
         else:
```
SomberNight, line 2742: I see that now if the first part of the `if` is entered, we are guaranteed to return.
Would it be clearer to remove the `else:` and indent everything below one to the left?
Or keep the `else:` and indent everything to the right, to be inside?

SomberNight (commented):
some comments for now but will think a bit more about this

SomberNight:
> some comments for now but will think a bit more about this

Ok, I think the change is logically correct. Though this part of the code (process_unfulfilled_htlc and friends) is a bit hard to review...

electrum/lnpeer.py
ecdsa: I think it is clearer to keep the outer `else:`. Note that the inner else branch that executes `return None, None, None` was added for clarity. simply writing `return preimage, None, None` would have been sufficient.

ecdsa: ah, yes, I see what you mean, some code can be right-indented.

electrum/lnpeer.py
ecdsa: yes, I meant the latter

electrum/lnpeer.py

electrum/lnpeer.py
ecdsa: done

PR #8934, opened 2024-03-08
add wallet rename functionality, and add to qml client

accumulator:


ErnestoLara (approved):

electrum/daemon.py
```diff
@@ -538,6 +538,21 @@ def get_wallet(self, path: str) -> Optional[Abstract_Wallet]:
...
+
+        path = wallet.db.storage.path
+        wallet_key = self._wallet_key_from_path(path)
+        wallet = self._wallets.get(wallet_key)
+        assert wallet
+
+        new_path = wallet.rename_wallet(new_basename)
+        new_wallet_key = self._wallet_key_from_path(new_path)
+        self._wallets.pop(wallet_key)
+        self._wallets[new_wallet_key] = wallet
```
SomberNight, line 554: As a completely different approach, have you considered unloading the wallet, changing the name of file on disk, and then loading the wallet again? so e.g. `daemon._stop_wallet()`, `os.rename()`, `daemon.load_wallet()`

Disregarding GUI concerns, from a trying-to-avoid-subtle-bugs and keeping things simple point of view, that seems to me like a potentially more interesting approach :P  but maybe that has its own complications (?)

accumulator: I think this has complications. There are many references held to `Wallet` instances, but very few to `WalletDB` instances, making the above approach a bit better manageable. 

Desktop should probably destroy the toplevel window associated with the wallet, which might be problematic when it's the only/last window (app closing?), and visually disturbing. 

On Android/QML it would also be quite tricky to get right. 

Unloading+loading is quite a complex code path..

accumulator:
Closing this, as the assumption that `wallet.storage` references are not kept outside `wallet` is violated since `TxBatcher`

PR #8946, opened 2024-03-13
build: split make_osx.sh script into two: "build" and "sign" parts

SomberNight:
(untested)


ecdsa:
can we add `security -v unlock-keychain login.keychain` at the beginning to the signing script?
that command is needed for signing.

ecdsa:
merged manually

PR #9000, opened 2024-04-11
ecc: add bindings for schnorr sign/verify, and refactor

SomberNight:
see messages for individual commits


electrum/ecc_fast.py
```diff
@@ -127,6 +132,31 @@ def load_library():
...
+        # --enable-module-schnorrsig
+        try:
+            secp256k1.secp256k1_schnorrsig_sign32.argtypes = [c_void_p, c_char_p, c_char_p, c_char_p, c_char_p]
+            secp256k1.secp256k1_schnorrsig_sign32.restype = c_int
+
+            secp256k1.secp256k1_schnorrsig_verify.argtypes = [c_void_p, c_char_p, c_char_p, c_size_t, c_char_p]
+            secp256k1.secp256k1_schnorrsig_verify.restype = c_int
+        except (OSError, AttributeError):
+            raise LibModuleMissing('libsecp256k1 library found but it was built '
+                                   'without required module (--enable-module-schnorrsig)')
```
SomberNight, line 144: note: the CI "Regtest functional tests" job is failing with this atm

SomberNight: Looks like ubuntu 22.04 LTS does not ship new enough libsecp for this module. This might be a problem for users as well...
debian 12 (current stable) and ubuntu 24.04 LTS do ship new enough versions at least.
Note that the optional modules we need are enabled by default starting with libsecp256k1 version 0.2.

SomberNight: (fixed CI issue by building our own secp library)

SomberNight: @ecdsa said that for now we should not hard fail at import time if just the schnorr module is missing, and instead we should fail at runtime when a related method is called.
I made that change in https://github.com/spesmilo/electrum/pull/9000/commits/fae672c60cbe30448da743abbcfb12731adbceb7.

ecdsa:
I manually started the android build

PR #9003, opened 2024-04-11
AppArmor Profiles for Electrum Tarball and AppImage

robertmin2:


SomberNight:
Could you please provide some context what this is, and how it is supposed to be used?

robertmin1:
AppArmor is a Mandatory Access Control (MAC) system which confines programs to a limited set of resources. AppArmor confinement is provided via profiles loaded into the kernel. 
To load/add a profile you need to copy the profile to `/etc/apparmor.d` e.g `sudo cp -R -L profiles/* /etc/apparmor.d`
I can add a `Readme` but I wanted to get intial feedback first
More info : https://ubuntu.com/tutorials/beginning-apparmor-profile-development#1-overview

SomberNight:
I see. Thanks.

Is this something that an "upstream" project should provide (e.g.  are these files Linux distribution agnostic)? Do you perhaps have examples of other projects that do so?
In general I have no problem with this including something like this. It should probably go into a more descriptively named folder, e.g. `contrib/apparmor/`.

contrib/profiles/abstractions/electrum
```diff
@@ -0,0 +1,39 @@
...
+include <abstractions/user-tmp>
+include <abstractions/X>
+include <abstractions/wayland>
+include <abstractions/mesa>
+include <abstractions/dri-enumerate>
+include <abstractions/nameservice>
+include <abstractions/openssl>
+include <abstractions/vulkan>
+include <abstractions/python>
+include <abstractions/evince>
```
SomberNight, line 12: How is evince related? Is it because of the menu option to open bitcoin.pdf?

robertmin1: Yes! using `evince`, we can view the bitcoin.pdf. 
Additionally, `xdg-open` enables us to open the `Report bug` link. 
The inclusion of the browser imports `ubuntu-browsers` and `snap_browsers` facilitates the opening of browser links like `Documentation and Official website`. However, for opening the `Report bug` link, `xdg-open` is still required.

robertmin1:
Debian, Ubuntu & OpenSUSE derivatives, come preconfigured with AppArmor as the default, so it should be ready to use for these
It appears that Monero was in the process of adding one, but didn't [complete it](https://github.com/monero-project/monero/pull/6758)
Also, mysql does provide [a profile](https://github.com/mysql/mysql-server/tree/824e2b4064053f7daf17d7f3f84b7a3ed92e5fb4/packaging/deb-in/extra)

contrib/apparmor/abstractions/electrum
```diff
@@ -0,0 +1,43 @@
...
+include <abstractions/mesa>
+include <abstractions/dri-enumerate>
+include <abstractions/nameservice>
+include <abstractions/openssl>
+include <abstractions/vulkan>
+include <abstractions/python>
+include <abstractions/evince>
+include <abstractions/xdg-open> 
+include <abstractions/ubuntu-browsers>
+include <abstractions/snap_browsers>
```
SomberNight, line 15: What about when not using Ubuntu, but instead e.g. Debian?
Are these defined on Debian? Or how do these includes work -- is not finding one of these a hard error or just a warning/silent ignore?

robertmin1: Yes, they are configured on Debian (I tested on Debian 12 and also Kali). All the profiles are predefined, and if one is missing, AppArmor produces a hard error. 
We can also set them as `include if exists`  (last four) , as they are only needed by the help tab and don't cause any hard errors if blocked.
The strange thing I noticed is that `Kali` didn't come with the `evince profile`, but this was because `evince` wasn't installed, after installing the profile was installed.

SomberNight: > We can also set them as `include if exists` (last four) , as they are only needed by the help tab and don't cause any hard errors if blocked.

I think we should that do then, yes.

> The strange thing I noticed is that Kali didn't come with the evince profile, but this was because evince wasn't installed, after installing the profile was installed.

It also does not make sense to tie us to a specific PDF reader program.
Specifically about evince, is it even installed if you choose a different DE than GNOME? E.g. if you install debian with KDE? Anyway, best not require it.

SomberNight:
>  To load/add a profile you need to copy the profile to /etc/apparmor.d e.g for our case sudo cp -R -L profiles/* /etc/apparmor.d
I can add a Readme, but I wanted to get intial feedback first

Could you add a short readme file (in same folder) saying what this is and how to use/test it?

robertmin1:
Alright! Added the Readme

SomberNight:
Looks good. Thanks.

robertmin1:
Thank you for the quick feedback. I'll continue testing the profiles on various OSs in the coming weeks

PR #9013, opened 2024-04-17
build: update bundled python version

SomberNight:


SomberNight:
Hmm, pretty sure I've tested not only building on macos, but reproducibility as well. But I will try to re-test then...

SomberNight:
Tested again. Builds for me.

Perhaps you have something old cached. E.g. try again after `$ git clean -ffxd`

PR #9031, opened 2024-04-29
bitcoin.py/transaction.py: API changes: rm most hex usage

SomberNight:
Clean up APIs in `bitcoin.py` and `transaction.py`:
- instead of some functions operating with hex strings, and others using bytes, this consolidates most things to use bytes
  - scripts, pubkeys, and signatures, should be bytes in almost all places now
- `bitcoin.rev_hex` and `bitcoin.int_to_hex` are removed

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1614,7 +1614,7 @@ class Channel(AbstractChannel):
     def force_close_tx(self) -> PartialTransaction:
         tx = self.get_latest_commitment(LOCAL)
         assert self.signature_fits(tx)
-        tx.sign({self.config[LOCAL].multisig_key.pubkey.hex(): (self.config[LOCAL].multisig_key.privkey, True)})
+        tx.sign({self.config[LOCAL].multisig_key.pubkey: self.config[LOCAL].multisig_key.privkey})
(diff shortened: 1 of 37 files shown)


electrum/lnchannel.py
```diff
@@ -1167,16 +1167,16 @@ def _verify_htlc_sig(self, *, htlc: UpdateAddHtlc, htlc_sig: bytes, htlc_directi
                                                           commit=ctx,
                                                           ctx_output_idx=ctx_output_idx,
                                                           htlc=htlc)
-        preimage_hex = htlc_tx.serialize_preimage(0)
-        pre_hash = sha256d(bfh(preimage_hex))
+        pre_hash = htlc_tx.serialize_preimage(0)
```
ecdsa, line 1170: that renaming is a bit confusing, `pre_hash` is not a hash anymore

SomberNight: Oooooh! So that stood for "hash of preimage"?
I thought it was pre-hash, as in before-hashing.
That's why I renamed it, because with that reading it was a misnomer :D

ecdsa: got it :-)

CarolinaMomma843 (approved):

PR #9034, opened 2024-04-30
cmdline: add listconfig/helpconfig

ecdsa:
This adds commands to view the list of config variables

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -359,6 +359,17 @@ class Commands:
             cv.set(value)
 
     @command('')
+    async def listconfig(self):
+        """Returns the list of all configuration variables. """
+        return self.config.list_config_vars()
+
+    @command('')
+    async def helpconfig(self, key, more=False):
+        """Returns help about a configuration variable. """
+        cv = self.config.cv.from_key(key)
+        return cv.get_long_desc() if more else cv.get_short_desc()
+
+    @command('')
     async def make_seed(self, nbits=None, language=None, seed_type=None):
         """Create a seed"""
         from .mnemonic import Mnemonic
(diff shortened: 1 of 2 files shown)


accumulator:
These commands should not require a running daemon, but currently require either a running daemon or an explicit `--offline` (but I guess this applies to more commands, as e.g. `setconfig`/`getconfig` also require `--offline` currently)

accumulator:
minor nitpick: `helpconfig use_gossip` returns `Use trampoline routing`

ecdsa:
The `--offline` syntax requirement is off-topic. This syntax is explicitly needed for all commands, regardless of whether they use the network, because it clarifies whether the command uses the running daemon or not.

electrum/commands.py
```diff
@@ -352,6 +352,17 @@ async def setconfig(self, key, value):
...
+    @command('')
+    async def listconfig(self):
+        """Returns the list of all configuration variables. """
+        return self.config.list_config_vars()
+
+    @command('')
+    async def helpconfig(self, key, more=False):
+        """Returns help about a configuration variable. """
+        cv = self.config.cv.from_key(key)
+        return cv.get_long_desc() if more else cv.get_short_desc()
```
SomberNight, line 364: I am not sure we need the `--more` option. How about combining the two descriptions?
e.g.:
```suggestion
    async def helpconfig(self, key):
        """Returns help about a configuration variable. """
        cv = self.config.cv.from_key(key)
        short = cv.get_short_desc()
        long = cv.get_long_desc()
        if short and long:
            return short + "\n---\n\n" + long
        else:
            return short or long
```

ecdsa: indeed, that's better

PR #9039, opened 2024-05-06
onion messages

accumulator:
This PR is ready for review.

- added CLI commands
    - `get_blinded_path_via` to construct a blinded path 
        - only with direct peers as introduction point currently
            - longer routes should be trivial to add
        - optionally add dummy hops to obfuscate true blinded path length
    - `send_onion_message` to send a text `onionmsg_tlv.message` payload to a node_id or blinded path
        - receiving electrum node will log these on `info` level
- route finding over onion message capable nodes using channels as edges
    - tested against c-lightning
    - not tested against electrum peers, as electrum does not publish to the channel graph
    - known issue: valid paths with offline hops are a problem
- ~direct connect to destination peer/introduction point if no path can be found over channel graph~
- `OnionMessageManager` class implementing queues for rate limiting
    - request-reply queue for matching replies to requests, 
    - retries in case no reply within x seconds etc. (not needed for text `onionmsg_tlv.message` payloads, which are fire-and-forget, but eventually bolt12 will need this for e.g. 'invoice_request')
        - known issue: not trying alternate routes yet
    - forwarding queue, for limiting amount and pacing of onion message forwards
- serialization and deserialization of `subtype` declarations in wire definitions

### Testing

- Testing done with sending onion messages across electrum instances `Alice <-> Bob <-> Carol <-> Dave`  using node_ids and blinded paths as destination.
    - no route finding, see below
    - typical test:
        1. `alice get_blinded_path_via $(bob nodeid)`
        2. `carol send_onion_message <blinded path> hello`
- Testing done with sending onion messages and invoice_requests to c-lightning instances `Alice <-> cln01 <-> cln02 <-> cln03 <-> Bob`
    - with routing finding (so offer with introduction point not a direct peer)
    - reply matching with request (using `bolt12` branch to issue `invoice_request` from `offer`)

#### Testing setup with c-lightning

Why?
- test interop with different implementation
- provide gossip to electrum clients, direct electrum <-> electrum peers don't exchange gossip

Note: this assumes a few things
- pre-existing named docker containers `cln01`, `cln02`, `cln03` running regtest c-lightning
- a `bitcoind` that not just listens on localhost, but is reachable from the docker network bridge (listen 0.0.0.0)
- regtest bitcoind and electrumx running
- regtest.sh with additional command `new_blocks <n>`
```
                                                                  
 ┌───────────┐     ┌───────────┐     ┌───────────┐    ┌──────────┐
 │           │     │           │     │           │    │          │
 │   cln01   ◄─────┼   cln02   ◄─────┼   cln03   ◄────┼   dave   │
 │           │     │           │     │           │    │          │
 └─────▲─────┘     └─────▲─────┘     └─────▲─────┘    └──────────┘
       │                 │                 │                      
       │                 │                 │                      
       │                 │                 │                      
 ┌─────┴─────┐     ┌─────┼─────┐     ┌─────┼─────┐                
 │           │     │           │     │           │                
 │   alice   │     │    bob    │     │   carol   │                
 │           │     │           │     │           │                
 └───────────┘     └───────────┘     └───────────┘                

```

```bash
alias alice='./run_electrum --regtest -D /tmp/alice'
alias bob='./run_electrum --regtest -D /tmp/bob'
alias carol='./run_electrum --regtest -D /tmp/carol'
alias dave='./run_electrum --regtest -D /tmp/dave'
alias cln01='docker exec cln01 lightning-cli'
alias cln02='docker exec cln02 lightning-cli'
alias cln03='docker exec cln03 lightning-cli'

tests/regtest/regtest.sh init alice
tests/regtest/regtest.sh init bob
tests/regtest/regtest.sh init carol
tests/regtest/regtest.sh init dave
bitcoin-cli sendtoaddress $(cln01 newaddr|jq -r .bech32) 1
bitcoin-cli sendtoaddress $(cln02 newaddr|jq -r .bech32) 1
bitcoin-cli sendtoaddress $(cln03 newaddr|jq -r .bech32) 1
tests/regtest/regtest.sh new_blocks 101
sleep 15
cln02 connect $(cln01 getinfo|jq -r .id) localhost $(cln01 getinfo|jq -r .binding[0].port)
cln02 fundchannel $(cln01 getinfo|jq -r .id) 15000000
cln03 connect $(cln02 getinfo|jq -r .id) localhost $(cln02 getinfo|jq -r .binding[0].port)
cln03 fundchannel $(cln02 getinfo|jq -r .id) 15000000
tests/regtest/regtest.sh start bob
tests/regtest/regtest.sh start alice
tests/regtest/regtest.sh start carol
tests/regtest/regtest.sh start dave
tests/regtest/regtest.sh new_block
alice open_channel $(cln01 getinfo|jq -r .id)@localhost:$(cln01 getinfo|jq -r .binding[0].port) 0.015
bob open_channel $(cln02 getinfo|jq -r .id)@localhost:$(cln02 getinfo|jq -r .binding[0].port) 0.015
carol open_channel $(cln03 getinfo|jq -r .id)@localhost:$(cln03 getinfo|jq -r .binding[0].port) 0.015
dave open_channel $(cln03 getinfo|jq -r .id)@localhost:$(cln03 getinfo|jq -r .binding[0].port) 0.015
tests/regtest/regtest.sh new_blocks 3
```

### known issues

- route finding over channel graph is unreliable, as nodes can be offline and no alternative routes are tried and no direct peer connection is attempted once a valid but unusable path is found.
    - it should be possible to determine from gossip if an edge is usable
- direct peer connect to dest/introduction point is more reliable, but zero privacy if not using TOR.
- retries in case of request-reply is not using alternative routes yet
- no support for trampoline


accumulator:
this branch is now `onion_messages` only, all bolt12 stuff has been moved to the `bolt12` branch, which builds on top of this branch/PR

accumulator:
@ecdsa, @SomberNight, this PR is getting ready to be reviewed.

- I have successfully sent onion messages across electrum instances `Alice <-> Bob <-> Carol <-> Dave`  using node_ids and blinded paths as destination.
- added CLI commands
    - `get_blinded_path_via` to construct a blinded path (only via a direct peer currently, but longer routes should be trivial)
    - `send_onion_message` to send a text `onionmsg_tlv.message` payload to a node_id or blinded path
- route finding over onion message capable nodes using channels as edges (however not yet tested as regtest with only electrum nodes lacks a channel graph)
- initial rough `OnionMessageManager` class implementing a send queue which should match replies to requests, rate-limiting, alternate route retries in case no reply within x seconds etc. (not needed for text `onionmsg_tlv.message` payloads, but eventually bolt12 will need this)
- serialization and deserialization of `subtype` declarations in wire definitions
- TODOs added where applicable
- Of note, there's extra variants of `new_onion_packet` and `get_shared_secrets_along_route` that need to be unified..

electrum/lnmsg.py
```diff
@@ -348,9 +370,116 @@ def __init__(self, *, for_onion_wire: bool = False):
...
+        else:
+            raise Exception(f"unexpected field count: {count!r}")
+        if count == 0:
+            return b""
+
+        parsedlist = []
+
+        # since we don't know the length of the field in advance, and io.BytesIO doesn't provide a look-ahead,
+        # we use this crappy way to detect EOF; read all, and if len(left) > 0, re-wrap in BytesIO stream
+        while left := fd.read(-1):
```
SomberNight, line 457: Can't we use the existing `_num_remaining_bytes_to_read` function?
Or maybe I don't understand the issue.

electrum/lnpeer.py
```diff
@@ -2795,3 +2795,7 @@ def process_onion_packet(
         if self.network.config.TEST_FAIL_HTLCS_WITH_TEMP_NODE_FAILURE:
             raise OnionRoutingFailure(code=OnionFailureCode.TEMPORARY_NODE_FAILURE, data=b'')
         return processed_onion
+
+    def on_onion_message(self, payload):
+        if getattr(self.lnworker, 'onion_message_manager'):  # only on LNWallet
```
SomberNight, line 2800: I don't think this works. Maybe with `hasattr`, otherwise this probably raises AttributeError.

Looks like in some other places we are using `if self.lnworker == self.lnworker.network.lngossip` (negation would be needed here), or we could use `isinstance(self.lnworker, LNWallet)` (though LNWallet cannot be a top-level import in this file). All these options are somewhat ugly, but maybe they should be consistent at least -- we could factor it out into a method of lnpeer.

accumulator: > I don't think this works. Maybe with `hasattr`, otherwise this probably raises AttributeError.

Yep indeed. For now replaced with `hasattr`, as the other approaches are at least equally unelegant :)

electrum/lnmsg.py
accumulator: Not sure why I thought seeking was not possible on `io.BytesIO`..

Onnielassy28 (approved):

accumulator:
I have a few commits queued that need some cleaning, also needs rebasing. marking draft for now

tests/test_onion_message.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def test_blinding_privkey(self):
+        a = blinding_privkey(bfh('4141414141414141414141414141414141414141414141414141414141414141'),
+                             bfh('031b84c5567b126440995d3ed5aaba0565d71e1834604819ff9c17f5e9d5dd078f'))
+        self.assertEqual(a, bfh('7e959bf6bdd3a98caf26cbbee7b69678381d5fa2882c6c12eb2042c2367264b0'))
+
+    def test_create_blinded_path(self):
+        pubkey = bfh('02eec7245d6b7d2ccb30380bfbe2a3648cd7a942653f5aa340edcea1f283686619')  # alice
+        session_key = bfh('3030303030303030303030303030303030303030303030303030303030303030')
```
ecdsa, line 164: any reason why session_key is '303030..' and not '030303...' here?  is this a typo?

tests/test_onion_message.py
```diff
@@ -0,0 +1,255 @@
...
+            OnionHopsDataSingle(tlv_stream_name='onionmsg_tlv', payload={'message': {'text': b'hello'}},
+                                blind_fields={'padding': {'padding': b''},
+                                              'path_id': {'data': bfh('deadbeefbadc0ffeedeadbeefbadc0ffeedeadbeefbadc0ffeedeadbeefbadc0')},
+                                              'unknown_tag_65535': {'data': bfh('06c1')}
+                                              }
+                                )
+        ]
+
+        hop_shared_secrets, blinded_node_ids = get_shared_secrets_along_route(payment_path_pubkeys, session_key,
+                                                                              with_blinded_node_ids=True)
```
ecdsa, line 72: why is this called twice? this looks unnecessary

electrum/lnonion.py
```diff
@@ -211,6 +259,29 @@ def new_onion_packet(
         hmac=next_hmac)
 
 
+def encrypt_encrypted_data_tlv(*, shared_secret, **kwargs):
```
ecdsa, line 262: This name is unfortunate. It suggests that we are encrypting a second time something that was already encrypted. It is enough to call this method `encrypt_data_tlv` (and to call the next one `decrypt_data_tlv`). And, yes, I am aware that the name `encrypted_data_tlv` is part of the spec. That's not a reason.

electrum/lnpeer.py
```diff
@@ -1462,7 +1461,7 @@ def send_node_announcement(self, alias:str):
             addrlen=len(addresses),
             addresses=addresses)
         h = sha256d(raw_msg[64+2:])
-        signature = ecc.ECPrivkey(self.privkey).ecdsa_sign(h, sigencode=ecdsa_sig64_from_r_and_s)
+        signature = ECPrivkey(self.privkey).ecdsa_sign(h, sigencode=ecdsa_sig64_from_r_and_s)
```
ecdsa, line 1464: This kind of cosmetic change can be committed to master independently of this pull request. It is not useful to include it in the pull request.

electrum/lnpeer.py
```diff
@@ -2807,8 +2806,8 @@ def process_onion_packet(
         try:
             processed_onion = process_onion_packet(
                 onion_packet,
-                associated_data=payment_hash,
                 our_onion_private_key=self.privkey,
+                associated_data=payment_hash,
```
ecdsa, line 2810: same here, no reason to overload a PR with parameter reordering.

electrum/onion_message.py
```diff
@@ -0,0 +1,742 @@
...
+            node_filter=lambda x, y: True if x == lnwallet.node_keypair.pubkey else is_onion_message_node(x, y),
+            my_sending_channels=my_sending_channels
+        ): return path
+
+    # strat2: dest node has host:port in channel_db? then open direct peer connection
+    if lnwallet.channel_db:
+        if peer_addr := lnwallet.channel_db.get_last_good_address(node_id):
+            peer = await lnwallet.add_peer(str(peer_addr))
+            await peer.initialized
+            return [PathEdge(short_channel_id=None, start_node=None, end_node=node_id)]
```
ecdsa, line 169: Is this really needed?  Bolt-04 suggests that onion messages should use already existing connections, not create new ones:

> Onion messages allow peers to use existing connections 

I would rather not have methods that create extra peer connections as side-effect, unless there is a clear use case.

ecdsa:
Another thing that needs to be improved is the unit tests provided in this PR. We should refrain from copy-pasting long hexadecimal strings in our code. Since there is a json file provided by the RFC, why not use that file directly? Here is a branch where I started to do that:  
https://github.com/spesmilo/electrum/tree/onion_messages_unittest

ecdsa:
Also note that some of the provided unit tests are not very useful (`test_decode_onion_message_packet` seems useless).
OTOH, some important functions are not tested. It would be nice to have a unit tests that covers `send_onion_message_to`

ecdsa:
Renaming `reqrpy` to `requestreply` is an improvement. We could even use `reply`, as there is no reason to use two words here.

ecdsa:
A general note on formatting: if we use multiple lines for the parameeters of a function, we should avoid indenting them at the level of the opening parenthesis, but use an extra line instead. Example:

```diff
-                blinded_path = OnionWireSerializer._read_complex_field(fd=blinded_path_fd,
-                                                                       field_type='blinded_path',
-                                                                       count=1)
+                blinded_path = OnionWireSerializer._read_complex_field(
+                    fd=blinded_path_fd,
+                    field_type='blinded_path',
+                    count=1)
```

Why? because changing the method name will trigger reindentation of all the lines in the first case, not in the second case. This results in larger diffs. See for example the commit where `submit_reqrpy` is renamed to `submit_requestreply`

tests/test_onion_message.py
accumulator: indeed, not necessary

electrum/lnonion.py
accumulator: Agreed, this looks confusing. I would suggest `encrypt_onionmsg_data_tlv` then, to indicate it is an onion message record.

accumulator:
> The current version of `get_shared_secrets_along_route` seems unnecessarily complicated. We should be very careful about introducing complexity, as it may become difficult to get rid of it down the road.

Yes I don't like it either. 

> I suggest to rewrite `get_shared_secrets_along_route` as follows (this is a diff against master, not against this PR):
> 
> ```diff
>  def get_shared_secrets_along_route(payment_path_pubkeys: Sequence[bytes],
> -                                   session_key: bytes) -> Sequence[bytes]:
> +                                   session_key: bytes) -> Tuple[Sequence[bytes], Sequence[bytes]]:
>      num_hops = len(payment_path_pubkeys)
>      hop_shared_secrets = num_hops * [b'']
> +    hop_blinded_node_ids = num_hops * [b'']
>      ephemeral_key = session_key
>      # compute shared key for each hop
>      for i in range(0, num_hops):
>          hop_shared_secrets[i] = get_ecdh(ephemeral_key, payment_path_pubkeys[i])
> +        hop_blinded_node_ids[i] = get_blinded_node_id(payment_path_pubkeys[i], hop_shared_secrets[i])
>          ephemeral_pubkey = ecc.ECPrivkey(ephemeral_key).get_public_key_bytes()
>          blinding_factor = sha256(ephemeral_pubkey + hop_shared_secrets[i])
>          blinding_factor_int = int.from_bytes(blinding_factor, byteorder="big")
>          ephemeral_key_int = int.from_bytes(ephemeral_key, byteorder="big")
>          ephemeral_key_int = ephemeral_key_int * blinding_factor_int % ecc.CURVE_ORDER
>          ephemeral_key = ephemeral_key_int.to_bytes(32, byteorder="big")
> -    return hop_shared_secrets
> +    return hop_shared_secrets, hop_blinded_node_ids
> ```
> 
> This is a lot simpler than the current PR, and I believe this is sufficient. If we need to create a blinded path with an override (for the moment this is only needed in the unit tests AFAICT), then it is easy to do that by calling the above method twice, and concatenate the results:
> 
> ```python
>         hop_shared_secrets1, blinded_node_ids1 = get_shared_secrets_along_route([ALICE_PUBKEY], BLINDING_SECRET)                                                                              
>         hop_shared_secrets2, blinded_node_ids2 = get_shared_secrets_along_route([BOB_PUBKEY, CAROL_PUBKEY, DAVE_PUBKEY], BLINDING_OVERRIDE_SECRET)                                            
>         hop_shared_secrets = hop_shared_secrets1 + hop_shared_secrets2                                                                                                                        
>         blinded_node_ids = blinded_node_ids1 + blinded_node_ids2
> ```

Yes this is much better

electrum/lnpeer.py
accumulator: This was done to be consistent with `ECPubkey`, which was already imported and used without `ecc.`

accumulator:
> Another thing that needs to be improved is the unit tests provided in this PR. We should refrain from copy-pasting long hexadecimal strings in our code. Since there is a json file provided by the RFC, why not use that file directly? Here is a branch where I started to do that: https://github.com/spesmilo/electrum/tree/onion_messages_unittest

merged.

ecdsa:
Hi, I know that I have already asked this, but could you rebase this branch on current master?

It is currently based on 8f7c11f295887675116714a24c36485556cb416d, which does not pass regtest.
In fact, on that commit, regtest just times out after 1 hour.
As a result, everytime you push in this branch, regtests are failing and unnecessarily consuming cirrus compute credits.

ecdsa:
`TestOnionMessageManager` is unacceptably slow (98 seconds here)
When we test timeouts, we do not use the same time constants as on mainnet.
Please have a look at `MPP_EXPIRY` in `test_lnpeer.py`, line 149

electrum/lnmsg.py
```diff
@@ -413,13 +552,21 @@ def read_tlv_stream(self, *, fd: io.BytesIO, tlv_stream_name: str) -> Dict[str,
...
+                        if field_type in self.subtypes:
+                            parsed[tlv_record_name][field_name] = self._read_complex_field(
+                                fd=tlv_record_fd,
+                                field_type=field_type,
+                                count=field_count)
+                        else:
+                            parsed[tlv_record_name][field_name] = _read_field(
+                                fd=tlv_record_fd,
+                                field_type=field_type,
+                                count=field_count)
```
SomberNight, line 569: This looks a bit redundant.
Can't we do this instead (?):
```diff
diff --git a/electrum/lnmsg.py b/electrum/lnmsg.py
index da849f22fd..52a3555f72 100644
--- a/electrum/lnmsg.py
+++ b/electrum/lnmsg.py
@@ -388,7 +388,14 @@ class LNSerializer:
     def _write_complex_field(self, *, fd: io.BytesIO, field_type: str, count: Union[int, str],
                              value: Union[List[Dict[str, Any]], Dict[str, Any]]) -> None:
         assert fd
-        assert field_type in self.subtypes, f"unknown subtype {field_type}"
+        if field_type not in self.subtypes:
+            _write_field(
+                fd=fd,
+                field_type=field_type,
+                count=count,
+                value=value,
+            )
+            return
 
         if isinstance(count, int):
             assert count >= 0, f"{count!r} must be non-neg int"
@@ -428,22 +435,21 @@ class LNSerializer:
                 if subtype_field_name not in record:
                     raise Exception(f'complex field type {field_type} missing element {subtype_field_name}')
 
-                if subtype_field_type in self.subtypes:
-                    self._write_complex_field(
-                        fd=fd,
-                        field_type=subtype_field_type,
-                        count=subtype_field_count,
-                        value=record[subtype_field_name])
-                else:
-                    _write_field(
-                        fd=fd,
-                        field_type=subtype_field_type,
-                        count=subtype_field_count,
-                        value=record[subtype_field_name])
+                self._write_complex_field(
+                    fd=fd,
+                    field_type=subtype_field_type,
+                    count=subtype_field_count,
+                    value=record[subtype_field_name])
 
     def _read_complex_field(self, *, fd: io.BytesIO, field_type: str, count: Union[int, str])\
             -> Union[bytes, List[Dict[str, Any]], Dict[str, Any]]:
         assert fd
+        if field_type not in self.subtypes:
+            return _read_field(
+                fd=fd,
+                field_type=field_type,
+                count=count,
+            )
         if isinstance(count, int):
             assert count >= 0, f"{count!r} must be non-neg int"
         elif count == "...":
@@ -468,16 +474,10 @@ class LNSerializer:
                     vars_dict=parsed,
                     allow_any=True)
 
-                if subtype_field_type in self.subtypes:
-                    parsed[subtype_field_name] = self._read_complex_field(
-                        fd=fd,
-                        field_type=subtype_field_type,
-                        count=subtype_field_count)
-                else:
-                    parsed[subtype_field_name] = _read_field(
-                        fd=fd,
-                        field_type=subtype_field_type,
-                        count=subtype_field_count)
+                parsed[subtype_field_name] = self._read_complex_field(
+                    fd=fd,
+                    field_type=subtype_field_type,
+                    count=subtype_field_count)
             parsedlist.append(parsed)
 
         return parsedlist if count == '...' or count > 1 else parsedlist[0]
@@ -503,18 +503,11 @@ class LNSerializer:
                                                            vars_dict=kwargs[tlv_record_name],
                                                            allow_any=True)
                         field_value = kwargs[tlv_record_name][field_name]
-                        if field_type in self.subtypes:
-                            self._write_complex_field(
-                                fd=tlv_record_fd,
-                                field_type=field_type,
-                                count=field_count,
-                                value=field_value)
-                        else:
-                            _write_field(
-                                fd=tlv_record_fd,
-                                field_type=field_type,
-                                count=field_count,
-                                value=field_value)
+                        self._write_complex_field(
+                            fd=tlv_record_fd,
+                            field_type=field_type,
+                            count=field_count,
+                            value=field_value)
                     else:
                         raise Exception(f"unexpected row in scheme: {row!r}")
                 _write_tlv_record(fd=fd, tlv_type=tlv_record_type, tlv_val=tlv_record_fd.getvalue())
@@ -557,16 +550,10 @@ class LNSerializer:
                             vars_dict=parsed[tlv_record_name],
                             allow_any=True)
                         #print(f">> count={field_count}. parsed={parsed}")
-                        if field_type in self.subtypes:
-                            parsed[tlv_record_name][field_name] = self._read_complex_field(
-                                fd=tlv_record_fd,
-                                field_type=field_type,
-                                count=field_count)
-                        else:
-                            parsed[tlv_record_name][field_name] = _read_field(
-                                fd=tlv_record_fd,
-                                field_type=field_type,
-                                count=field_count)
+                        parsed[tlv_record_name][field_name] = self._read_complex_field(
+                            fd=tlv_record_fd,
+                            field_type=field_type,
+                            count=field_count)
                     else:
                         raise Exception(f"unexpected row in scheme: {row!r}")
                 if _num_remaining_bytes_to_read(tlv_record_fd) > 0:

```

accumulator: So, basically allow `_{read,write}_complex_field` to handle primitive types as well?

That will likely work, but then the func name might be slightly confusing. Maybe then rename this func to `_{read,write}_field` and rename the `_{read,write}_field` outside the `LNSerializer` class to `_{read,write}_primitive_field` ?

ecdsa: sure, it could be renamed like that

accumulator: Done in 269600f

electrum/onion_message.py
```diff
@@ -0,0 +1,759 @@
...
+
+    return result
+
+
+class Timeout(Exception): pass
+
+
+class OnionMessageRequest(NamedTuple):
+    future: asyncio.Future
+    payload: bytes
```
SomberNight, line 416: maybe?
```suggestion
    payload: dict
```

electrum/onion_message.py
```diff
@@ -0,0 +1,759 @@
...
+from electrum.lnonion import (get_bolt04_onion_key, OnionPacket, process_onion_packet,
+                              OnionHopsDataSingle, decrypt_onionmsg_data_tlv, encrypt_onionmsg_data_tlv,
+                              get_shared_secrets_along_route, new_onion_packet)
+from electrum.lnutil import LnFeatures
+from electrum.util import OldTaskGroup, log_exceptions
+
+
+# do not use util.now, because it rounds to integers
+def now():
+    return time.time()
```
SomberNight, line 48: Ok, but then `util.now` is not used anywhere? :)
Different use cases have different needs for "now", so maybe it does not make sense to have a `util.now`.

Some use cases might want higher than second precision, others might want `time.monotonic()`, etc.

(In fact you might want to use `time.monotonic()` in this file because looks like only time delays are being measured, absolute time is not used. But anyway. We can always improve this later.)

electrum/onion_message.py
```diff
@@ -0,0 +1,759 @@
...
+            next_public_key_int = ecc.ECPubkey(blinding) * blinding_factor_int
+            next_blinding = next_public_key_int.get_public_key_bytes()
+
+        if is_dummy_hop:
+            self.process_onion_message_packet(next_blinding, onion_packet)
+            return
+
+        self.submit_forward(onion_packet=onion_packet, blinding=next_blinding, node_id=next_node_id)
+
+    def on_onion_message(self, payload: dict) -> None:
```
SomberNight, line 719: Would it be possible to write unit test(s) that exercise this method, and the forwarding logic?

Atm these methods are not covered:
```
process_forward_queue
submit_forward
on_onion_message_received
on_onion_message_received_unsolicited
on_onion_message_forward
on_onion_message
process_onion_message_packet
```

electrum/onion_message.py
```diff
@@ -0,0 +1,759 @@
...
+            shared_secret=shared_secret,
+            encrypted_recipient_data=payload['encrypted_recipient_data']['encrypted_recipient_data']
+        )
+
+        self.logger.debug(f'parsed recipient_data: {recipient_data!r}')
+
+        if processed_onion_packet.are_we_final:
+            self.on_onion_message_received(recipient_data, payload)
+        else:
+            self.on_onion_message_forward(recipient_data, processed_onion_packet.next_packet, blinding, shared_secret)
```
SomberNight, line 759: When we are just a light client, e.g. running on a phone, do we want to forward messages?
Payment forwarding is gated behind a config variable. Should this also be?

electrum/onion_message.py
accumulator: certainly.

electrum/onion_message.py
accumulator: `util.now` was introduced in this PR, but left in after changing to `onion_message.now`. I'll remove it from the PR.

electrum/onion_message.py
accumulator: Although it would be unlikely that a mobile electrum will be included in a blinded path, we should probably do it for consistency :)

electrum/onion_message.py
accumulator: done in 14fd28e4eb4cd948a1630cd9690fc038dbb8d429

electrum/onion_message.py
accumulator: done in 14fd28e4eb4cd948a1630cd9690fc038dbb8d429

electrum/onion_message.py
SomberNight: I was thinking about attack surface reduction.
Anyway, if you agree there is no point for a mobile node to forward, I think we could potentially reuse `config.EXPERIMENTAL_LN_FORWARD_PAYMENTS` for this (instead of adding a new config var).

electrum/onion_message.py
accumulator: Note that the logic for determining whether a message is unsolicited has changed slightly. Before, the absence of `path_id` was the determinant. Now, looking at the message to Dave in the test vectors, it is clear that the constraint had to be interpreted slightly different, and also unknown `path_id`s in the message must not lead to dropping the message, but be regarded as unsolicited.

```
  - if `path_id` is set and corresponds to a path the reader has previously published in a `reply_path`:
    - if the onion message is not a reply to that previous onion:
      - MUST ignore the onion message
  - otherwise (unknown or unset `path_id`):
    - if the onion message is a reply to an onion message which contained a `path_id`:
      - MUST respond (or not respond) exactly as if it did not send the initial onion message.
```

electrum/simple_config.py
```diff
@@ -1047,6 +1047,7 @@ def __setattr__(self, name, value):
     )
 
     LIGHTNING_NODE_ALIAS = ConfigVar('lightning_node_alias', default='', type_=str)
+    LIGHTNING_FORWARD_ONION_MESSAGES = ConfigVar('lightning_forward_onion_messages', default=False, type_=bool)
```
SomberNight, line 1050: Not sure if we've ever linked config vars together, but perhaps it is time :D
```suggestion
    LIGHTNING_FORWARD_ONION_MESSAGES = ConfigVar('lightning_forward_onion_messages', default=lambda config: config.EXPERIMENTAL_LN_FORWARD_PAYMENTS, type_=bool)
```
As in, if someone has enabled payment forwarding but have not touched this config var, we could default enable this too.

accumulator: For a moment there I thought I committed a copypaste error :D

accumulator: > As in, if someone has enabled payment forwarding but have not touched this config var, we could default enable this too.

This will probably lead to confusion somewhere.. We could use a single config option for forwarding in general, payments and messages?

SomberNight: > confusion

maybe..

Well in that case I would prefer to just keep the current "EXPERIMENTAL_LN_FORWARD_PAYMENTS" config var as-is, and reuse it for onion msg forwarding. No need to rename it I mean, I think.

tests/test_onion_message.py
```diff
@@ -263,6 +266,7 @@ class MockNetwork:
     def __init__(self):
         self.asyncio_loop = get_asyncio_loop()
         self.taskgroup = OldTaskGroup()
+        self.config = SimpleConfig({'lightning_forward_onion_messages': True})
```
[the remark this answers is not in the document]
accumulator: done

electrum/simple_config.py
accumulator: renaming it is a PITA, as we then need to port older config keys.. :(

ecdsa: Please do not create unneeded config variables. I do not think there is any scenario where we would want to forward payments and not onion messages. And if we need to do that one day, then it will be time to define a new config variable. But for now, please use the current config var.

electrum/onion_message.py
```diff
@@ -0,0 +1,764 @@
...
+
+        # TODO: we should try alternate paths when retrying, this is currently not done.
+        # (send_onion_message_to decides path, without knowledge of prev attempts)
+        send_onion_message_to(self.lnwallet, node_id_or_blinded_path, final_payload)
+
+    def _path_id_from_payload_and_key(self, payload: dict, key: bytes) -> bytes:
+        # TODO: use payload to determine prefix?
+        return b'electrum' + key
+
+    def _get_request_for_path_id(self, recipient_data: dict) -> Request:
```
SomberNight, line 610:
```suggestion
    def _get_request_for_path_id(self, recipient_data: dict) -> Optional[Request]:
```

ecdsa:
LGTM, but it would be nice not to delete my comment about the now function. Any reason why you deleted it?
This kind of knowledge is easily lost if we do not comment about it, and it took me a while to figure out what was going on.

accumulator:
> LGTM, but it would be nice not to delete my comment about the now function. Any reason why you deleted it? This kind of knowledge is easily lost if we do not comment about it, and it took me a while to figure out what was going on.

Did I? I removed the commit from the PR introducing `now()` in `util.py` (you know, clean history and all), maybe the associated comment disappeared?

tests/test_onion_message.py
```diff
@@ -0,0 +1,474 @@
...
+
+        onionmsg = bfh(test_vectors['onionmessage']['onion_message_packet'])
+        try:
+            t.on_onion_message({
+                'blinding': bfh(test_vectors['route']['blinding']),
+                'len': len(onionmsg),
+                'onion_message_packet': onionmsg
+            })
+        finally:
+            await asyncio.sleep(TIME_STEP)
```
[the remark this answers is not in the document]
accumulator: without the sleep I got crashes w.r.t the event loop, it seemed adding an await avoided that problem. However I can now remove the sleep without problems.

What is the error you're getting?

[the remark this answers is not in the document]
ecdsa: the issue is that the sleep is too short, not that it should be removed. 
Replacing the value by 2*TIME_STEP as suggested looks ok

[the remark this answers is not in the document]
accumulator: no, you're missing the point. The issue is that it's racy, the fix is probabilistic.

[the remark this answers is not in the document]
ecdsa: ah, I see. Well, if you have a way to remove the race, please go for it.
note: on my machine the test passes if that line is removed.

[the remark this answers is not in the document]
SomberNight: > What is the error you're getting?

The last assert fails.
```py
  File "...\Python310\lib\asyncio\base_events.py", line 649, in run_until_complete
    return future.result()
  File ...\electrum\tests\test_onion_message.py", line 448, in test_forward
    self.assertTrue(self.was_sent)
AssertionError: False is not true
```

> The issue is that it's racy, the fix is probabilistic.

You're right.
Without a sleep, it passes.
With `sleep(1*TIME_STEP)`, it fails.
And with `sleep(2*TIME_STEP)`, it passes again.

Issue #9033, opened 2024-04-30
Acinq trampoline fees have been raised recently


ecdsa:
According to @t-bast, the trampoline routing fees of the Acinq node were recently raised to 0.4%
The highest fee level tried by Electrum is 0.3% plus a base fee of 100 sats:
```
    {
        'fee_base_msat': 100000,
        'fee_proportional_millionths': 3000,
	'cltv_expiry_delta': 576,
    },

```

t-bast:
That is correct, we indeed raised our trampoline fees to 0.4% + 4 sats. We do reject payments that use a lower fee. This can be an issue in the multi-trampoline setting, but as was discussed with the introduction of blinded paths, there won't be a need for multi-trampoline payments in the future: a single trampoline node paying to a blinded path guarantees privacy while being more efficient than using multiple trampoline nodes.

PR #9041, opened 2024-05-06
trampoline: use exponential search for fees, capped by configurable budget

SomberNight:
See individual commit messages.

closes https://github.com/spesmilo/electrum/issues/9033

### DIFF
--- a/electrum/trampoline.py
+++ b/electrum/trampoline.py
@@ -13,45 +13,6 @@ from .logging import get_logger
 
 _logger = get_logger(__name__)
 
-# trampoline nodes are supposed to advertise their fee and cltv in node_update message
-TRAMPOLINE_FEES = [
-    {
-        'fee_base_msat': 0,
-        'fee_proportional_millionths': 0,
-        'cltv_expiry_delta': 576,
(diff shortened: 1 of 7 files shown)


electrum/trampoline.py
```diff
@@ -185,17 +131,47 @@ def _extend_trampoline_route(
...
+) -> None:
+    # calculate budget_to_use, based on given max available "budget"
+    if trampoline_fee_level == 0:
+        budget_to_use = 0
+    else:
+        assert trampoline_fee_level > 0
+        MAX_LEVEL = 6
+        if trampoline_fee_level > MAX_LEVEL:
+            raise NoPathFound()
+        budget_to_use = budget.fee_msat // (2 ** (MAX_LEVEL - trampoline_fee_level))
```
SomberNight, line 161: Note: this logic means that the lowest total fee we will try is the user configured max divided by 32. E.g. by default the max is 1% of payment amount, i.e. `fee=1/3200*payment_amount`.

If the user configures a higher max budget, that raises the corresponding the lowest possible fee we will try.

PR #9059, opened 2024-05-20
slip39: Implement extendable backups

onvej-sl:
This pull request implements a [recent revision](https://github.com/satoshilabs/slips/commit/8d060706b549af6443e04f55605b71f65c981663) of the [SLIP-39 specification](https://github.com/satoshilabs/slips/blob/master/slip-0039.md). The highest bit of the iteration exponent has been repurposed for a new "extendable backup flag". Setting the flag indicates that the random identifier is not used as salt in the encryption of the master secret. This makes it possible to create multiple sets of shares, such that each set of shares uses a different identifier and each set of shares leads to the same master secret for every passphrase. This is a desirable property, which unfortunately was not considered in the initial specification. It allows users to start working with their wallet by creating a single-share (1-of-1) scheme and later upgrade to a multi-share scheme while maintaining the same encrypted master secret and passphrases. Existing SLIP-39 shares remain valid. New shares created using the revised specification use a different checksum customization string and are not compatible with older implementations.

### DIFF
--- a/electrum/slip39.py
+++ b/electrum/slip39.py
@@ -63,11 +63,19 @@ def _xor(a: bytes, b: bytes) -> bytes:
 _ID_LENGTH_BITS = 15
 """The length of the random identifier in bits."""
 
-_ITERATION_EXP_LENGTH_BITS = 5
+_ITERATION_EXP_LENGTH_BITS = 4
 """The length of the iteration exponent in bits."""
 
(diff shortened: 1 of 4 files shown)


SomberNight:
related: https://github.com/trezor/python-shamir-mnemonic/issues/48

SomberNight:
Looks good. Thx.

PR #9073, opened 2024-05-30
qml: don't unbind/unregister the ActivityResultListener from within the ActivityResultListener handler func

accumulator:
don't unbind/unregister the ActivityResultListener from within the ActivityResultListener handler func. instead, schedule a queued finished signal to unregister the listener after the handler has finished. See PythonActivity.java in P4A for why this probably causes the most often occurring crash we see on the Play Store:

```
Exception java.lang.RuntimeException:
  at android.app.ActivityThread.deliverResults (ActivityThread.java:5164)
  at android.app.ActivityThread.handleSendResult (ActivityThread.java:5205)
  at android.app.servertransaction.ActivityResultItem.execute (ActivityResultItem.java:51)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2136)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:236)
  at android.app.ActivityThread.main (ActivityThread.java:8061)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:656)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:967)
Caused by java.util.ConcurrentModificationException:
  at java.util.ArrayList$Itr.next (ArrayList.java:860)
  at org.kivy.android.PythonActivity.onActivityResult (PythonActivity.java:218)
  at android.app.Activity.dispatchActivityResult (Activity.java:8501)
  at android.app.ActivityThread.deliverResults (ActivityThread.java:5157)
```

### DIFF
--- a/electrum/gui/qml/qeqrscanner.py
+++ b/electrum/gui/qml/qeqrscanner.py
@@ -1,6 +1,6 @@
 import os
 
-from PyQt6.QtCore import pyqtProperty, pyqtSignal, pyqtSlot, QObject
+from PyQt6.QtCore import pyqtProperty, pyqtSignal, pyqtSlot, QObject, Qt
 from PyQt6.QtGui import QGuiApplication
 
 from electrum.util import send_exception_to_crash_reporter, UserFacingException
@@ -10,7 +10,7 @@ from electrum.i18n import _
 
 
 if 'ANDROID_DATA' in os.environ:
-    from jnius import autoclass, cast
+    from jnius import autoclass
     from android import activity
 
     jpythonActivity = autoclass('org.kivy.android.PythonActivity').mActivity
@@ -23,10 +23,18 @@ class QEQRScanner(QObject):
 
     found = pyqtSignal()
 
+    finished = pyqtSignal()
+
     def __init__(self, parent=None):
         super().__init__(parent)
         self._hint = _("Scan a QR code.")
         self._scan_data = ""  # decoded qr code result
+        self.finished.connect(self._unbind, Qt.ConnectionType.QueuedConnection)
+
+        self.destroyed.connect(lambda: self.on_destroy())
+
+    def on_destroy(self):
+        self._unbind()
 
     @pyqtProperty(str)
     def hint(self):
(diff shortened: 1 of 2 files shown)


electrum/gui/qml/qeqrscanner.py
```diff
@@ -49,34 +57,34 @@ def open(self):
...
+            send_exception_to_crash_reporter(e)
+        finally:
+            self.finished.emit()
+
     @pyqtSlot()
-    def close(self):
-        pass
+    def _unbind(self):
+        if 'ANDROID_DATA' in os.environ:
+            activity.unbind(on_activity_result=self.on_qr_activity_result)
```
SomberNight, line 81: Is it ok for this to run twice?
`finished` and `destroyed` might both trigger it.

SomberNight: https://github.com/kivy/python-for-android/blob/5dc17e16217620daf76d58ac6d20e4f9a9d4ff8b/pythonforandroid/recipes/android/src/android/activity.py#L57-L59
Looks like it is not thread-safe but otherwise fine to call it multiple times.

accumulator: yes, that's ok, see https://github.com/kivy/python-for-android/blob/5dc17e16217620daf76d58ac6d20e4f9a9d4ff8b/pythonforandroid/recipes/android/src/android/activity.py#L52-L63

accumulator: > Looks like it is not thread-safe but otherwise fine to call it multiple times.

The java counterpart is 'thread-safe' by using a `synchronized` block, although judging from the exception it isn't really blocking on the lock (maybe due to python-java boundary crossing?)

SomberNight: We could add our own locking via threading.Lock() around `activity.unbind`. Though maybe the pyqt signalling already makes it indirectly invoked from the safe thread, I'm not sure. Anyway, up to you, if you think it's worth it. Otherwise the diff looks good.

accumulator: the Qt signal handlers are called in-sequence, from the UI thread. I think we're good here. 
If we do want to add a lock, I think the proper place is in P4A

PR #9076, opened 2024-05-30
qml: "wallet_else_pin" auth should only use the wallet pw if is unified

SomberNight:
The config PIN is a global, while the wallet passwords are per wallet (unless the passwords have been unified).
Knowing the per-wallet password should not be sufficient to change the global config PIN.

related https://github.com/spesmilo/electrum/pull/9074


accumulator:
LGTM

accumulator (approved):

PR #9083, opened 2024-06-07
Persist MPP resolution status in wallet file.

ecdsa:
If we accept a MPP and we forward the payment (trampoline or swap), we need to persist the payment accepted status, or we might wrongly release htlcs on the next restart.

lnworker.received_mpp_htlcs used to be cleaned up in maybe_cleanup_forwarding, which only applies to forwarded payments. However, since we now persist this dict, we need to clean it up also in the case of payments received by us. This part of maybe_cleanup_forwarding has been migrated to lnworker.maybe_cleanup_mpp

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2750,6 +2750,7 @@ class Peer(Logger):
                 # return payment_key so this branch will not be executed again
                 return None, payment_key, None
             elif preimage:
+                self.lnworker.maybe_cleanup_mpp(chan.get_scid_or_local_alias(), htlc)
                 return preimage, None, None
             else:
                 # we are waiting for mpp consolidation or preimage
@@ -2761,7 +2762,10 @@ class Peer(Logger):
             preimage = self.lnworker.get_preimage(payment_hash)
             error_bytes, error_reason = self.lnworker.get_forwarding_failure(payment_key)
             if error_bytes or error_reason or preimage:
-                self.lnworker.maybe_cleanup_forwarding(payment_key, chan.get_scid_or_local_alias(), htlc)
+                cleanup_keys = self.lnworker.maybe_cleanup_mpp(chan.get_scid_or_local_alias(), htlc)
+                is_htlc_key = ':' in payment_key
+                if is_htlc_key or payment_key in cleanup_keys:
+                    self.lnworker.maybe_cleanup_forwarding(payment_key)
             if error_bytes:
                 return None, None, error_bytes
             if error_reason:
--- a/electrum/lnsweep.py
+++ b/electrum/lnsweep.py
@@ -297,10 +297,12 @@ def create_sweeptxs_for_our_ctx(
         ctn=ctn)
     for (direction, htlc), (ctx_output_idx, htlc_relative_idx) in htlc_to_ctx_output_idx_map.items():
         if direction == RECEIVED:
-            if chan.lnworker.get_payment_status(htlc.payment_hash) == PR_PAID:
-                preimage = chan.lnworker.get_preimage(htlc.payment_hash)
-            else:
+            if chan.lnworker.is_incomplete_mpp(htlc.payment_hash):
                 # do not redeem this, it might publish the preimage of an incomplete MPP
+                continue
+            preimage = chan.lnworker.get_preimage(htlc.payment_hash)
+            if not preimage:
+                # we might not have the preimage if this is a hold invoice
                 continue
         else:
             preimage = None
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -414,6 +414,16 @@ class LNWalletWatcher(LNWatcher):
         self.lnworker = lnworker
         LNWatcher.__init__(self, lnworker.wallet.adb, network)
 
+    @event_listener
+    async def on_event_blockchain_updated(self, *args):
+        # overload parent method with cache invalidation
+        # we invalidate the cache on each new block because
+        # some processes affect the list of sweep transactions
+        # (hold invoice preimage revealed, MPP completed, etc)
+        for chan in self.lnworker.channels.values():
+            chan._sweep_info.clear()
+        await self.trigger_callbacks()
+
     def diagnostic_name(self):
         return f"{self.lnworker.wallet.diagnostic_name()}-LNW"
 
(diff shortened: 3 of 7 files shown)


electrum/lnworker.py
```diff
@@ -169,18 +170,25 @@ class PaymentInfo(NamedTuple):
     status: int
 
 
-class RecvMPPResolution(Enum):
-    WAITING = enum.auto()
-    EXPIRED = enum.auto()
-    ACCEPTED = enum.auto()
-    FAILED = enum.auto()
```
SomberNight, line 176: We should add a comment here about these values being persisted in the wallet file, and hence being sensitive to changes. As in:
https://github.com/spesmilo/electrum/blob/02a9ab80be819468352f773bd7120421868d38f0/electrum/lnchannel.py#L77-L79
Btw, like with the channel states, we could potentially persist them by name instead of value. It only matters if we end up changing this enum, and how often. The ChannelState enum has been changed several times and hence in that case it was quite helpful to use the names.

electrum/lnworker.py
```diff
@@ -2351,47 +2359,46 @@ def update_mpp_with_received_htlc(
...
         if key not in mpp_status.htlc_set:
             mpp_status.htlc_set.add(key)  # side-effecting htlc_set
-        self.received_mpp_htlcs[payment_key] = mpp_status
+        self.received_mpp_htlcs[payment_key.hex()] = mpp_status
 
     def set_mpp_resolution(self, *, payment_key: bytes, resolution: RecvMPPResolution):
-        mpp_status = self.received_mpp_htlcs[payment_key]
-        self.received_mpp_htlcs[payment_key] = mpp_status._replace(resolution=resolution)
+        mpp_status = self.received_mpp_htlcs[payment_key.hex()]
+        self.logger.info(f'set_mpp_resolution {resolution.name} {len(mpp_status.htlc_set)}')
```
SomberNight, line 2366: Log line not very useful in non-trivial scenarios. It should also contain the payment_key or the payment_hash, or sth ~id-like.
Same goes for other log lines touched.

electrum/lnworker.py
```diff
@@ -2351,47 +2359,46 @@ def update_mpp_with_received_htlc(
...
 
     def get_first_timestamp_of_mpp(self, payment_key: bytes) -> int:
-        mpp_status = self.received_mpp_htlcs.get(payment_key)
+        mpp_status = self.received_mpp_htlcs.get(payment_key.hex())
         if not mpp_status:
             return int(time.time())
         return min([_htlc.timestamp for scid, _htlc in mpp_status.htlc_set])
 
-    def maybe_cleanup_forwarding(
+    def maybe_cleanup_mpp(
```
SomberNight, line 2382: What about channels that get closed, do we also want to clean-up the `received_mpp_htlcs` dict for those?

ecdsa: We still need the information for sweep.
I think it is safer not to clean this up, at least until the channel is REDEEMED

electrum/lnworker.py
ecdsa: done

electrum/lnsweep.py
```diff
@@ -297,11 +297,13 @@ def create_txns_for_htlc(
         ctn=ctn)
     for (direction, htlc), (ctx_output_idx, htlc_relative_idx) in htlc_to_ctx_output_idx_map.items():
         if direction == RECEIVED:
-            if chan.lnworker.get_payment_status(htlc.payment_hash) == PR_PAID:
-                preimage = chan.lnworker.get_preimage(htlc.payment_hash)
-            else:
+            if chan.lnworker.is_incomplete_mpp(htlc.payment_hash):
```
SomberNight, line 300: If the resolution is `EXPIRED` or `FAILED`, and we have the preimage, this will release it.
I am not sure if that is safe to do.
Is that an intended behaviour change on this line?

I mean, just consider lnworker.stopping_soon. That can transition the mpp to EXPIRED, and then someone force-closes, and we have the preimage (e.g. because this is a simple invoice we created), then we release it onchain even for a partial payment?

electrum/lnwatcher.py
```diff
@@ -414,6 +414,16 @@ def __init__(self, lnworker: 'LNWallet', network: 'Network'):
...
 
+    @event_listener
+    async def on_event_blockchain_updated(self, *args):
+        # overload parent method with cache invalidation
+        # we invalidate the cache on each new block because
+        # some processes affect the list of sweep transactions
+        # (hold invoice preimage revealed, MPP completed, etc)
+        for chan in self.lnworker.channels.values():
+            chan._sweep_info.clear()
+        await self.trigger_callbacks()
```
SomberNight, line 425: maybe do `super().on_event_blockchain_updated(*args)` instead (of last line)?

electrum/lnsweep.py
ecdsa: oh indeed. that was not intended.

SomberNight, line 300: another note is that this line duplicated 130 lines below, because of the symmetry, and both instances need to be changed, not just this one

electrum/lnpeer.py
```diff
@@ -2750,6 +2750,7 @@ async def wrapped_callback():
                 # return payment_key so this branch will not be executed again
                 return None, payment_key, None
             elif preimage:
+                self.lnworker.maybe_cleanup_mpp(chan.get_scid_or_local_alias(), htlc)
```
SomberNight, line 2753: It is too early to clean-up here. In general we should wait until there has been a revack in both directions.
Specifically, while there is any non-revoked commitment tx (of either party) that contains the htlc, we should not clean-up. This is because if such a commitment tx is broadcast, and the htlc mpp set had resolution ACCEPTED and we have the preimage, then we should use the preimage to claim the htlc onchain.

SomberNight: note: as this is about HTLCs we received, it is us who will fulfill or fail them, so maybe the clean-up (which needs to check `chan.hm.is_htlc_irrevocably_removed_yet(ctx_owner=None, ...)` can be done after *we* send revack, as the expected flow is:

```
LOCAL         REMOTE
   ---fulfill -->
   ---  sig  --->
   <---revack ---
   <---  sig  ---
   --- revack--->
*and now we can clean-up the htlc
```

PR #9086, opened 2024-06-09
chore: fix some comments

longxiangqiao:


electrum/gui/qt/rate_limiter.py
```diff
@@ -98,7 +98,7 @@ def _invoke(self, args, kwargs):
                 self.timer.start(int(diff*1e3))
                 #self.logger.debug("deferring")
         else:
-            # We had a timer active, which means as future call will occur. So return early and let that call happenin the future.
+            # We had a timer active, which means as future call will occur. So return early and let that call happening the future.
```
SomberNight, line 101: This sentence does not make sense.
Instead it was a space that was missing.

longxiangqiao: > This sentence does not make sense. Instead it was a space that was missing.

Thanks. Modified

PR #9091, opened 2024-06-10
qml: fix handling of channel backup import on lightning-disabled wallets

accumulator:
importing a channel backup to a second bip39/xprv wallet with lightning disabled now shows proper message instead of exception text.

We could also provide the option to the user here to enable lightning?

### DIFF
--- a/electrum/gui/qml/components/WalletMainView.qml
+++ b/electrum/gui/qml/components/WalletMainView.qml
@@ -43,7 +43,9 @@ Item {
 
         // Android based send dialog if on android
         var scanner = app.scanDialog.createObject(mainView, {
-            hint: qsTr('Scan an Invoice, an Address, an LNURL-pay, a PSBT or a Channel backup'),
+            hint: Daemon.currentWallet.isLightning
+                ? qsTr('Scan an Invoice, an Address, an LNURL-pay, a PSBT or a Channel Backup')
+                : qsTr('Scan an Invoice, an Address, an LNURL-pay or a PSBT')
         })
         scanner.onFound.connect(function() {
             var data = scanner.scanData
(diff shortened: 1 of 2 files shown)


electrum/gui/qml/components/WalletMainView.qml
```diff
@@ -52,7 +54,7 @@ Item {
                 app.stack.push(Qt.resolvedUrl('TxDetails.qml'), { rawtx: data })
             } else if (Daemon.currentWallet.isValidChannelBackup(data)) {
                 var dialog = app.messageDialog.createObject(app, {
-                    title: qsTr('Import Channel backup?'),
+                    title: qsTr('Import Channel Backup?'),
```
SomberNight, line 57: is this spelling change worth it to nuke the translations? :P

SomberNight (approved):
Looks good. Feel free to merge.

SomberNight:
> We could also provide the option to the user here to enable lightning?

To be clear, there is already an option in the wallet details to enable lightning, and you just mean we could also offer to enable it here, right?
I have no strong opinion on that, but I think it's fine to keep it as-is.

electrum/gui/qml/components/WalletMainView.qml
accumulator: I was doubting whether to do it, but decided to go ahead for consistency (although there's the plural form left still, e.g. 'Channel backups'), and include it in this PR to be reviewed :)

accumulator:
> > We could also provide the option to the user here to enable lightning?
> 
> To be clear, there is already an option in the wallet details to enable lightning, and you just mean we could also offer to enable it here, right? 

Correct

>I have no strong opinion on that, but I think it's fine to keep it as-is.

No strong opinion here either, just wanted to mention it.

PR #9100, opened 2024-06-20
create submodules for ecc and aionostr

ecdsa:
The `ecc` submodule is there twice: 
- electrum/ecc
- electrum/aionostr/ecc

The install script copies `libsecp256k1.so.2` in only one of them.
That seems to be enough, because of the way `loadLibrary` works.
However, we might forget about that. Should we copy it twice instead?`


ecdsa:
`aionostr` uses the `websockets` package. 
apparently, that package does not work with a proxy. 
https://stackoverflow.com/questions/47225488/client-websocket-connect-through-proxy
maybe we could use `websocket-client` instead.

accumulator:
> The `ecc` submodule is there twice:
> 
>     * electrum/ecc
> 
>     * electrum/aionostr/ecc
> 

The submodules could be added to the python interpreter search path, similar to what is done in `run_electrum`:

```
if is_local or is_android:
    sys.path.insert(0, os.path.join(script_dir, 'packages'))
```

ecdsa:
Closing this. We should distribute proper python package.
New project in https://github.com/spesmilo/electrum-ecc

PR #9118, opened 2024-07-03
trezor: support trezorlib==0.13.9, add compat for <=0.13.8. fixes #9116

### LINKED ISSUE #9116: trezor v0.13.9 breaks the plugin
Just a heads up that trezor v0.13.9 breaks the plugin for Electrum ABC, and  I would guess Electrum is affected too.

Here is where I expect it to fail  (import `RecoveryDeviceType` )
https://github.com/spesmilo/electrum/blob/master/electrum/plugins/trezor/trezor.py#L32

```
  File "/work/electrum/electrumabc_plugins/trezor/trezor.py", line 28, in <module>
[Step 1/1]     from trezorlib.messages import (
[Step 1/1] ImportError: cannot import name 'RecoveryDeviceType' from 'trezorlib.messages' (/usr/local/lib/python3.9/dist-packages/trezorlib/messages.py)
[Step 1/1] .....................................................................................................................s...................
[Step 1/1] ...s........................[trezor] Library version for 'trezor' is incompatible.
[08:14:38]
[Step 1/1] Installed: 0.13.9, Needed: 0.13.8 <= x < 0.14
[08:14:38]
[Step 1/1] ...................................................................................................................s.........sss.............................s.s...[dummy name] Library version for 'dummy name' is incompatible.
[Step 1/1] Installed: 0.13.9, Needed: 0.13.8 <= x < 0.14
[Step 1/1] ======================================================================
[Step 1/1] ERROR: test_electrum_tx_to_txtype (electrumabc_plugins.trezor.tests.test_trezor.TestTrezorPlugin)
[Step 1/1] ----------------------------------------------------------------------
[Step 1/1] Traceback (most recent call last):
[Step 1/1]   File "/work/electrum/electrumabc_plugins/trezor/tests/test_trezor.py", line 53, in test_electrum_tx_to_txtype
[Step 1/1]     self.assertEqual(plugin.electrum_tx_to_txtype(tx, xpub_path), expected_txtype)
[Step 1/1]   File "/work/electrum/electrumabc_plugins/trezor/trezor.py", line 597, in electrum_tx_to_txtype
[Step 1/1]     t = TransactionType()
[Step 1/1] NameError: name 'TransactionType' is not defined
[Step 1/1] ----------------------------------------------------------------------
[Step 1/1] Ran 330 tests in 110.698s
[Step 1/1] FAILED (errors=1, skipped=8)
```

accumulator:


accumulator:
I prefer dropping support for the old version, however I haven't had time to test against 0.13.9 so this PR is just meant as a convenience to test against 0.13.9

Sentini14:
I prefer dropping support for the old version, however I haven't had time
to test against 0.13.9 so this PR is just meant as a convenience to test
against 0.13.9

—


***@***.***

PiRK:
On my end the reasoning was that 0.13.8 is well tested and I would rather use it for another release or two while people running from source can start testing 0.13.9 in real life scenarios and confirm that there is no other hidden broken compatibility issue.

dlitz:
I just noticed this after submitting my PR, but if you want a simpler change that only supports 0.13.8 and 0.13.9, I made a patch that does that in #9141.

SomberNight:
I have merged the other PR (https://github.com/spesmilo/electrum/pull/9141).
That still has the "compat" part but in a cleaner way IMO.
Thanks for the comments and code itself in both PRs.

> Do we we really want to support incompatible versions of trezorlib?
> I would rather drop support for the old version, unless we have a good reason to keep supporting it.

In general we don't, I agree, but in this case it can be done relatively cleanly.

> Do we we really want to support incompatible versions of trezorlib?
> Has this been thought through?

One reason to support somewhat older python-trezor, would be that debian is backporting new versions of electrum much more frequently than new versions of python-trezor. Atm debian stable (backports) [has electrum 4.5.5](https://packages.debian.org/bookworm-backports/electrum) but [only python-trezor 0.12.4](https://packages.debian.org/bookworm/python3-trezor). i.e. they are incompatible ([debian ticket](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1062442)). Anyway, clearly the gap is too large and this specific change does not help.

Still, in general I don't think it is worth a lot of code complexity or effort.

bosomt:
Any plans to release version that will include trezorlib v0.13.9 ?
Current Electrum still does not support new model TS5 :(

matejcik:
of course we can also add the compat name into trezorlib 0.13.10, so that this is not breaking :woman_shrugging:

accumulator:
> of course we can also add the compat name into trezorlib 0.13.10, so that this is not breaking 🤷‍♀️

That would make sense, yes. Then perhaps remove once 0.14.x is released, to follow semver-like versioning.

SomberNight:
> of course we can also add the compat name into trezorlib 0.13.10, so that this is not breaking 🤷‍♀️

That would only help users running from source.
In any case we need to build new binaries that bundle newer trezorlib than 0.13.8.

@accumulator I guess we could try to do a release soon. From the 4.5.x branch, just bugfixes, might be easier (though it would be nice to eventually release from master...)

accumulator:
> @accumulator I guess we could try to do a release soon. From the 4.5.x branch, just bugfixes, might be easier (though it would be nice to eventually release from master...)

I backported my bugfixes to the 4.5.x branch.

PR #9122, opened 2024-07-04
README: update install instructions for python virtualenv

accumulator:
Current instructions are outdated, python no longer allows global or --user install by default on current distributions

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,20 +18,61 @@ _(If you've come here looking to simply run Electrum,
 [you may download it here](https://electrum.org/#download).)_
 
 Electrum itself is pure Python, and so are most of the required dependencies,
-but not everything. The following sections describe how to run from source, but here
-is a TL;DR:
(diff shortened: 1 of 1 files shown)


accumulator:
still need to check if install instructions for windows and mac are still valid

mohsenrajei:
bc1qxvnqcahh3tm48dcry068svvrszakvn7hfd8h65

mohsenrajei:
bc1qxvnqcahh3tm48dcry068svvrszakvn7hfd8h65

gdx370:
These commands from the README don't work anymore without a force flag (`--break-system-packages`) due to _externally-managed-environment_ error:
```shell
$ ELECTRUM_ECC_DONT_COMPILE=1 python3 -m pip install --user ".[gui,crypto]"
$ python3 -m pip install --user -e .
```
I think it should be updated to use _venv_

PR #9141, opened 2024-07-19
trezor: Fix for trezor library version 0.13.9

dlitz:
This enables support for the Trezor Safe 5.


SomberNight (approved):
Thanks. Looks good. I also tested it a bit.

Issue #9151, opened 2024-07-24
Jade blockstream stopped working


evgeniy-emelyanchyk:
I've tried to setup Jade+Electrum multisig 2/2 for BTC Testnet  and it was setup without any problem, but when I tried to sign transaction I've got the error: 

```
145.65 | E | gui.qt.main_window.[payment_megatest2] | on_error
Traceback (most recent call last):
  File "electrum/gui/qt/util.py", line 913, in run
  File "electrum/wallet.py", line 2399, in sign_transaction
  File "electrum/plugin.py", line 362, in wrapper
  File "electrum/plugin.py", line 355, in run_in_hwd_thread
  File "concurrent/futures/_base.py", line 446, in result
  File "concurrent/futures/_base.py", line 391, in __get_result
  File "concurrent/futures/thread.py", line 58, in run
  File "electrum/plugins/jade/jade.py", line 287, in sign_transaction
    multisig_name = _register_multisig_wallet(wallet, self, txout.address)
  File "electrum/plugins/jade/jade.py", line 64, in _register_multisig_wallet
    signers.append({'fingerprint': bytes.fromhex(fingerprint),
TypeError: fromhex() argument must be str, not None
347.01 | E | gui.qt.main_window.[payment_megatest2] | on_error
Traceback (most recent call last):
  File "electrum/gui/qt/util.py", line 913, in run
  File "electrum/wallet.py", line 2399, in sign_transaction
  File "electrum/plugin.py", line 362, in wrapper
  File "electrum/plugin.py", line 355, in run_in_hwd_thread
  File "concurrent/futures/_base.py", line 446, in result
  File "concurrent/futures/_base.py", line 391, in __get_result
  File "concurrent/futures/thread.py", line 58, in run
  File "electrum/plugins/jade/jade.py", line 287, in sign_transaction
    multisig_name = _register_multisig_wallet(wallet, self, txout.address)
  File "electrum/plugins/jade/jade.py", line 64, in _register_multisig_wallet
    signers.append({'fingerprint': bytes.fromhex(fingerprint),
TypeError: fromhex() argument must be str, not None
```

Then I try to setup standard wallet electrum+Jade and got another error when I try to sign trx:
```
[...]

PR #9183, opened 2024-09-04
Jade native psbt

JamieDriver:
1.  latest jade api to recognise more recently added hw
2.  remove code which walked psbt/tx data and mapped it into Jade's legacy 'sign_tx' json format - instead pass PSBT directly to Jade for signing, since this has been supported in fw now for well over a year and seems to be working well.

### DIFF
--- a/electrum/plugins/jade/jade.py
+++ b/electrum/plugins/jade/jade.py
@@ -7,7 +7,7 @@ from electrum import bip32, constants
 from electrum.crypto import sha256
 from electrum.i18n import _
 from electrum.keystore import Hardware_KeyStore
-from electrum.transaction import Transaction
+from electrum.transaction import PartialTransaction, Transaction
 from electrum.wallet import Multisig_Wallet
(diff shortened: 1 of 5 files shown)


JamieDriver:
btw I had no probs using my Jade before or after this change ... I assume from your comment above that you also tested with a physical Jade, and it worked ok ?   

I ask because of: #9151 (mirrored in jade repo as https://github.com/Blockstream/Jade/issues/147)

Are you aware of any issues with Jade ?  Thanks.

SomberNight:
> I assume from your comment above that you also tested with a physical Jade, and it worked ok ?

Yes, I tested a bit. Now I've tested a bit more. Created singlesig and multisig wallets and signed txs, on linux/windows/macos, using master and latest Electrum release (though not full combinatorial matrix), with a physical device.
No issues here.

JamieDriver:
Yep, I too could not repro the issues.  Thanks for double-checking, much appreciated.

Issue #8007, opened 2022-10-08
Qt desktop GUI: upgrade to Qt6 (`gui/qt/`, QtWidgets)


SomberNight:
Opening this issue to discuss if it is time to upgrade to Qt6, for the desktop Qt GUI.

Qt5 is not getting any more updates.
The latest Qt version atm is 6.4. Qt6.2 was an LTS version, so I expect Qt6 is mature now.

Just like `pyqt5`, `pyqt6` is also available on PyPI, from the same maintainer.

There is the question of compatibility with OSes:
- Windows
    - Qt6 dropped support of win7 and 8.1 (see https://doc.qt.io/qt-6.2/supported-platforms.html#windows)
        - we already don't support win7 (see https://github.com/spesmilo/electrum/issues/7728)
        - would have to drop support of win8.1
    - Qt6 dropped support of win32 (x86, 32-bit), it requires amd64
        - see https://github.com/spesmilo/electrum/issues/6598
- macOS
    - we currently build on macOS 10.14, and support >=10.13 as targets
    - this would bump the min supported version to 10.14
        - see https://github.com/spesmilo/electrum/issues/3685#issuecomment-508556343
    - unlike pyqt5, pyqt6 has native binaries for (apple-flavoured-)arm64
- Linux
    - Debian
        - debian 11 does not have `python3-pyqt6` 
        - debian 12 (stable, released 2023-06) has `python3-pyqt6` 
            - https://packages.debian.org/bookworm/python3-pyqt6
    - Ubuntu
        - ubuntu 22.04 LTS does not have `python3-pyqt6` 
        - ubuntu 22.10 (unreleased) has `python3-pyqt6`
            - https://packages.ubuntu.com/kinetic/python3-pyqt6
    - Arch Linux
        - has `python-pyqt6`
            - https://archlinux.org/packages/extra/x86_64/python-pyqt6/  
    - Manjaro
        - has `python-pyqt6`
            - https://software.manjaro.org/package/python-pyqt6

SomberNight:
As long as debian stable, and the latest ubuntu LTS do not have packages, I think it is too early to upgrade to Qt6.

accumulator:
As Electrum-QML is only targeting android, not having PyQt6 available in the distro is only an issue for developers. The android build container builds PyQt itself.

SomberNight:
Right, but I think we can upgrade the desktop Qt GUI and the android qml GUI independently.
[...]

Issue #6598, opened 2020-09-16
consider building 64-bit Windows binaries (x64, x86_64, amd64, 64 bit, architecture)


SomberNight:
Currently we distribute three different binaries for Windows (portable, standalone, setup).
All of these target 32-bit Windows (but they can run on 64-bit arch too).

We build 32-bit binaries for compatibility reasons mainly, but this means the process cannot use >4 GB memory.
There are many `MemoryError` bug reports collected by Travis, where I believe users are not running OOM completely, they are simply reaching 4 GB.

see collection of issues at https://github.com/spesmilo/electrum/issues/5627
but also e.g.
https://github.com/spesmilo/electrum/issues/6212
https://github.com/spesmilo/electrum/issues/6272
https://github.com/spesmilo/electrum/issues/6303
https://github.com/spesmilo/electrum/issues/6378
https://github.com/spesmilo/electrum/issues/4995

We should consider either building 64-bit binaries exclusively instead, or also building those in addition to current.
(but we already build 3 binaries, so building two versions of each seems overkill)

It would be good to know how many users there are using 32 bit Windows.

See:
https://stackoverflow.com/a/41497867
https://store.steampowered.com/hwsurvey/

Among the participants of the Steam HW survey (2020-08), AFAICT 32 bit Windows among Windows users is less than 0.51%:
```
Windows 10 64 bit:   92.22%
Windows 7 64 bit:     5.44%
Windows 8.1 64 bit:   1.70%
Windows 7:            0.29%
Windows 10:           0.14%
Windows 8 64 bit:     0.13%
Other:                0.08%
```

SomberNight:
See https://www.bleepingcomputer.com/news/microsoft/rip-microsoft-to-drop-support-for-windows-10-on-32-bit-systems/

> "Beginning with Windows 10, version 2004, all new Windows 10 systems will be required to use 64-bit builds and Microsoft will no longer release 32-bit builds for OEM distribution. This does not impact 32-bit customer systems that are manufactured with earlier versions of Windows 10; Microsoft remains committed to providing feature and security updates on these devices, including continued 32-bit media availability in non-OEM channels to support various upgrade installation scenarios."

ysangkok:
I think that, depending on how the native code has been compiled, the limit may be as low as 2 GiB. See [stackoverflow comments](https://stackoverflow.com/questions/41497783/how-common-are-32-bit-windows-installations-as-of-now/41497867?noredirect=1#comment113036057_41497867)

[...]

Issue #3685, opened 2018-01-09
Electrum 3.0.5 not opening on Mac OS 10.9


lorepieri8:
As in the title, Electrum is not opening in Mac OS 10.9.

I had a similar issue with Electrum LTC, that was solved with an additional build (3.5.0.1-1, not available from the main page):

[https://github.com/pooler/electrum-ltc/issues/125](url)

Hopefully the solution for BTC should be similar.

lorepieri8:
It may be related to #3673

lzsaver:
So does it related to https://github.com/pooler/electrum-ltc/issues/125?

lorepieri8:
It seems so. In the console I get the same error codes:

16/01/18 19:32:57,176 gkbisd[144]: Unable to collect cdhash for /Applications/Electrum.app (error code 100024)

16/01/18 19:32:57,315 com.apple.launchd.peruser.501[157]: (Electrum.86976[62489]) Exited with code: 255

tessus:
I've just tried Electrum 3.1.1 on MacOSX 10.9.5, but I only get the following message (nothing from gkbisd):

```
Exited with code:255
```

How can I debug this? Or does anybody know what the problem might be?

tessus:
Is there any update on this? Maybe @ecdsa can create a build that works on 10.9.5.

I tried to create one on my 10.9.5 machine, but it didn't work. After a few hours I gave up.

bauerj:
Yeah, we should add that to the download page but currently on El Capitan (10.11.6) or newer is supported. To build a version that is compatible to older versions, we would need an installation of that version, which we do not have.

May I ask why you need to use Mavericks?

tessus:
Long story short, I still use a MacBook (Late 2008) and I'm afraid to upgrade that from 10.9.5.

(I also have another MacBook that runs 10.13.4)

SomberNight:
Note: need to keep an eye on this when increasing the PyQt version
See https://github.com/spyder-ide/qtpy/issues/155

SomberNight:
For future reference:

Qt 5.8 needs [minimum MacOS 10.9](https://doc.qt.io/archives/qt-5.8/supported-platforms.html)
Qt 5.9 needs [minimum MacOS 10.10](https://doc.qt.io/archives/qt-5.9/supported-platforms.html)
Qt 5.10 needs [minimum MacOS 10.11](https://doc.qt.io/archives/qt-5.10/supported-platforms.html)
[...]

PR #9189, opened 2024-09-10
Qt desktop GUI: upgrade to Qt6

SomberNight:
This PR migrates the Qt Widgets GUI used on desktop from Qt5 to Qt6.

Notes:
- the latest debian stable and ubuntu lts now both package python3-pyqt6, meaning most distros should now have it
- raises the minimum Windows version needed (for users) to win10 x64 (from 8.1 32 bit?)
  - see https://github.com/spesmilo/electrum/issues/6598
  - I think this is OK.
- raises the min macOS version needed to 11 (from 10.13)
  - related https://github.com/spesmilo/electrum/issues/3685
  - 10.13 was released in 2017-09, and 11 released in 2020-11. That's a significant bump, and the new minimum is only ~4 years old. Is this OK?
    - I think we bumped to 10.13 (from 10.11) around https://github.com/spesmilo/electrum/commit/095464b620e3e1bb699d174d969526f78ad356b9, which was released in 4.0.1 in [2020-07](https://github.com/spesmilo/electrum/blob/0f6898ed90be7777254b6868e1e1b0b5af79372f/RELEASE-NOTES#L1), at which point 10.13 was ~3 years old. Hence, the current bump is less strict and I think is OK.
- the win/mac/appimage binaries grew by ~20 MB :/
  - qt6 takes up more space. I tried to strip out some unneeded new modules (and adjusted our existing stripping-out to layout changes) - without that the size diff would be even greater.
- when running the AppImage on a system that has both wayland and xwayland, AFAICT previously QT_QPA_PLATFORM=xcb was the default, but now it is QT_QPA_PLATFORM=wayland
  - compare running `$ QT_QPA_PLATFORM=xcb ./run_electrum` vs `$ QT_QPA_PLATFORM=wayland ./run_electrum`
  - note that they look a bit different
	- in particular the titlebar for e.g. QMessageBox is missing on wayland
	  - e.g. try in console tab: `>>> window.question(msg="msg")`

closes https://github.com/spesmilo/electrum/issues/8007


electrum/gui/qt/qrreader/__init__.py
```diff
@@ -11,7 +11,7 @@
 #   to access the camera
 # - zbar fails to access the camera on macOS
 # - qtmultimedia seems to support more cameras on Windows than zbar
-# - qtmultimedia is often not packaged with PyQt5
+# - qtmultimedia is often not packaged with PyQt
 #   in particular, on debian, you need both "python3-pyqt5" and "python3-pyqt5.qtmultimedia"
```
accumulator, line 15:
```suggestion
#   in particular, on debian, you need both "python3-pyqt6" and "python3-pyqt6.qtmultimedia"
```

accumulator:
Even though it seems to work, `QDarkStyle` uses `load_stylesheet_pyqt5()`.

https://github.com/SomberNight/electrum/blob/c3a869d5e1dd0d0884e333a44694eb14b16c7a4f/electrum/gui/qt/__init__.py#L179

There's a `load_stylesheet_pyqt6()` func since `qdarkstyle>=3.2`

https://github.com/SomberNight/electrum/blob/c3a869d5e1dd0d0884e333a44694eb14b16c7a4f/contrib/requirements/requirements.txt#L3

accumulator (changes requested):

electrum/gui/common_qt/__init__.py
```diff
@@ -10,7 +10,7 @@ def get_qt_major_version() -> int:
     _GUI_QT_VERSION = getattr(sys, '_GUI_QT_VERSION', None)
     if _GUI_QT_VERSION is None:
         # used by pyinstaller when building (analysis phase)
-        _GUI_QT_VERSION = 5
+        _GUI_QT_VERSION = 6
```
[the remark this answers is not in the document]
SomberNight: No reason to keep it. Removed now.

SomberNight:
> There's a load_stylesheet_pyqt6() func since qdarkstyle>=3.2

Thanks, done.

----

At a higher level, what do you think about doing this migration now? :)  AFAICS there are no real blockers.

accumulator:
> > There's a load_stylesheet_pyqt6() func since qdarkstyle>=3.2
> 
> Thanks, done.
> 
> At a higher level, what do you think about doing this migration now? :) AFAICS there are no real blockers.

It looks fully functional. The only thing I ran into testing the branch was an error when grabbing QR from screen. That was on both wayland and xcb backends.

contrib/build-linux/appimage/make_appimage.sh
SomberNight, line 1: > when running the AppImage on a system that has both wayland and xwayland, AFAICT previously QT_QPA_PLATFORM=xcb was the default, but now it is QT_QPA_PLATFORM=wayland 

> I didn't test the AppImage yet. There might be dependency issues between pyqt6 and debian buster.

I've now tried running the appimage on debian 10 (oldoldstable), and am getting an error with wayland.

```
  4.16 | D | util.profiler | Plugins.__init__ 0.0422 sec
  4.16 | I | daemon.Daemon | launching GUI: qt
  4.76 | I | gui.qt.ElectrumGui | Qt GUI starting up... Qt=6.7.1, PyQt=6.7.1
/tmp/.mount_electrFlGFOt/usr/bin/python3: symbol lookup error: /tmp/.mount_electrFlGFOt/usr/lib/python3.11/site-packages/PyQt6/Qt6/plugins/platforms/../../lib/libQt6WaylandClient.so.6: undefined symbol: wl_proxy_marshal_flags
```

If I explicitly specify `QT_QPA_PLATFORM=xcb`, it starts and works as expected. But it picks wayland by default.
I found https://bugreports.qt.io/browse/QTBUG-114635 and it looks like even debian 11 might be affected.

One commenter recommends setting the env var to prefer x11 at runtime:
```
import os
if sys.platform == "linux":
    os.environ.setdefault("QT_QPA_PLATFORM", "xcb")
```
What do you think of that?

Or should we perhaps try to package a newer libwayland into the appimage? I am not sure if that is realistic.

accumulator: Maybe just document this, and let users of oldoldstable era distro's set the `xcb` override. This way we don't have to nag users on more recent distro's who probably expect it to properly detect the window system.

SomberNight: Well based on that thread, what is needed is `libwayland-client0>=1.19`, which is too old even for debian 11 (oldstable), or for ubuntu 20.04 lts. Meaning users would need current debian stable or ubuntu 22+.

What if we scoped it to `if sys.platform == "linux" and os.environ.get("APPIMAGE")`?

Ideally we could tell Qt to try the xcb plugin as a fallback...

accumulator: > What if we scoped it to `if sys.platform == "linux" and os.environ.get("APPIMAGE")`?

Yeah, better to only do this for AppImage builds

SomberNight: done in https://github.com/spesmilo/electrum/pull/9189/commits/0fb20cc816f83037664e08d701e260a13fb50785

SomberNight:
> The only thing I ran into testing the branch was an error when grabbing QR from screen. That was on both wayland and xcb backends.

I think that functionality is already broken on Linux if wayland is being used as window manager (regardless of the qpa plugin used). It might still work if wayland is not used at all.
On Windows and macOS, it works though.

Issue #9186, opened 2024-09-09
Consider replacing Source Sans Pro with Source Sans 3 in Revealer


sorenstoutner:
Revealer currently embeds the Source Sans Pro font.  However, this has been superseded upstream by Source Sans 3.

https://github.com/adobe-fonts/source-sans/issues/192

If you are interested, I would be willing to create a pull request that makes this change (which would slightly simplify my Debian packaging).

SomberNight:
Feel free to PR the change if it helps you.

sorenstoutner:
I have created a PR at:

https://github.com/spesmilo/electrum/pull/9190

PR #9190, opened 2024-09-11
Update Source Sans Pro font to the current Source Sans 3.

sorenstoutner:
This switches the deprecated Source Sans Pro font for the current Source Sans 3 as discussed at:

https://github.com/spesmilo/electrum/issues/9186


electrum/plugins/revealer/qt.py
```diff
@@ -416,7 +416,7 @@ def seed_img(self, is_seed = True):
             max_lines = 9
             max_words = int(max_letters/4)
 
-        font = QFont('Source Sans Pro', fontsize, QFont.Bold)
+        font = QFont('Source Sans 3', fontsize, QFont.Bold)
```
SomberNight, line 419: What about lines 551-557? I guess those too should be changed.

electrum/plugins/revealer/SourceSans3-Bold.otf
SomberNight, line 1: note: file matches https://github.com/adobe-fonts/source-sans/blob/82818d477056d06ae48f9c68970f9bffd56ce139/OTF/SourceSans3-Bold.otf

electrum/plugins/revealer/qt.py
SomberNight: side-note: I wonder if line 557 even works correctly, given it's trying to use `QFont.Normal` instead of `QFont.Bold`, and we are only including the bold font

sorenstoutner: That is a good point.  I didn't notice that this was also used during the calibration.  It should be easy to edit the two bold instances.  But you are right that we would need to include the normal font for the third entry.  Would we also need to call `QFontDatabase.addApplicationFont` inside of `def calibration_pdf`, or does it inherit it from `def seed_img`?

Note that in my testing, I am using a system that has the full Source Sans font installed as a Debian package.  The fonts have been removed from the Electrum package, so `QFontDatabase.addApplicationFont` doesn't do anything.  But I am currently patching the font name at line 419.

SomberNight: > Would we also need to call `QFontDatabase.addApplicationFont` inside of `def calibration_pdf`, or does it inherit it from `def seed_img`?

`QFontDatabase.addApplicationFont` side-effects the global state of the process. You are right that the `QFont()` calls in `calibration_pdf` might happen before `seed_img` runs (if it runs at all), so without `addApplicationFont` ever being called, in which case Qt probably falls back to an available generic font.
So actually it would probably make more sense to move all `QFontDatabase.addApplicationFont` calls (there are two in this file) to `Plugin.__init__`.

> But you are right that we would need to include the normal font for the third entry. 

I'd rather not include even more fonts. And in the case of the calibration, the fonts being used there do not matter at all. So instead of including the regular source sans font, we can just change that line to e.g. `font = QFont("", 7, QFont.Normal)`, and let Qt decide (which I think is exactly what that line has been doing if the requested font is not found).

sorenstoutner: OK.  I can make these changes, test them, and update the PR next week.

sorenstoutner: I have updated the PR.  Would you check over my syntax in __init__.py to make sure I did it correctly?

SomberNight: No, your change won't work, that `__init__` method will never run. I meant in `qt.py`, so this current file, `Plugin.__init__` (around line 50). However I have now tried doing what I meant, and it segfaults at runtime -- it is too early for loading the fonts at that point.

So anyway, I tried another way and manually merged your changes in
https://github.com/spesmilo/electrum/commit/47370d773c7f8eaf8a5fced048a6852925b735bc
https://github.com/spesmilo/electrum/commit/815e18302d21341726bc999c8826b9e81c559777

Thanks.

Issue #9193, opened 2024-09-13
Feature Request: Testnet 4 support


jlopp:
Testnet4 has been live for several months and will be available in Bitcoin Core with the next release in a month or so. Would be great for Electrum client and servers to support it.

https://testnet4.com/

SomberNight:
done in https://github.com/spesmilo/electrum/pull/9197

accumulator:
Is there a testnet4 faucet somewhere?

wakiyamap:
The first link, testnet4.com, has a list of faucets.

PR #9197, opened 2024-09-13
Add suport testnet4

wakiyamap:
ref. https://github.com/spesmilo/electrum/issues/9193

I have confirmed that there are no problems with sending and receiving.
I will consider Lightning after eclair or lnd supports it.


wakiyamap:
I'm going to fix tox errors. It seems that lndecode needs to be fixed.

SomberNight:
> I'm going to fix tox errors. It seems that lndecode needs to be fixed.

I think that part of lndecode could be simplified a bit, and `BOLT11_HRP_INV_DICT` removed as a whole. That should fix your problem (which stems from such an inverse dictionary not being possible anymore due to the collision between the two testnets). See if this looks ok:
```patch
diff --git a/electrum/lnaddr.py b/electrum/lnaddr.py
index e3f94f1c28..5c9a07cde8 100644
--- a/electrum/lnaddr.py
+++ b/electrum/lnaddr.py
@@ -108,9 +108,6 @@ def parse_fallback_addr(data5: Sequence[int], net: Type[AbstractNet]) -> Optiona
     return addr
 
 
-BOLT11_HRP_INV_DICT = {net.BOLT11_HRP: net for net in constants.NETS_LIST}
-
-
 def tagged5(char: str, data5: Sequence[int]) -> Sequence[int]:
     assert len(data5) < (1 << 10)
     return [CHARSET_INVERSE[char], len(data5) >> 5, len(data5) & 31] + data5
@@ -420,18 +417,16 @@ def lndecode(invoice: str, *, verbose=False, net=None) -> LnAddr:
 
     addr = LnAddr()
     addr.pubkey = None
+    addr.net = net
 
-    m = re.search("[^\\d]+", hrp[2:])
-    if m:
-        addr.net = BOLT11_HRP_INV_DICT[m.group(0)]
-        amountstr = hrp[2+m.end():]
-        # BOLT #11:
-        #
-        # A reader SHOULD indicate if amount is unspecified, otherwise it MUST
-        # multiply `amount` by the `multiplier` value (if any) to derive the
-        # amount required for payment.
-        if amountstr != '':
-            addr.amount = unshorten_amount(amountstr)
+    amountstr = hrp[2+len(net.BOLT11_HRP):]
+    # BOLT #11:
+    #
+    # A reader SHOULD indicate if amount is unspecified, otherwise it MUST
+    # multiply `amount` by the `multiplier` value (if any) to derive the
+    # amount required for payment.
+    if amountstr != '':
+        addr.amount = unshorten_amount(amountstr)
 
     addr.date = int_from_data5(data5_remaining[:7])
     data5_remaining = data5_remaining[7:]
```

wakiyamap:
Thank you. There seems to be no problem.
I will check that lightning actually works on testnet3 and then remove the draft.

SomberNight (approved):
Looks good. Thanks.

PR #9201, opened 2024-09-17
qt: fix scanning multi (privkeys, addresses) from QR.

accumulator:


SomberNight:
Note: this is a regression from https://github.com/spesmilo/electrum/commit/bde066f9ce50b305f86671a777f8206808be5f83, which added one too many indent level to the setText call. I am wondering if we could somehow programmatically catch these kind of bugs. PyCharm actually correctly flags the variable as being unused in the first branch. So clearly this kind of static analysation is possible. Maybe we could introduce something in the CI, if they have a not too high false-positive rate.

PR #9203, opened 2024-09-17
qml: Sweep from privkeys

accumulator:
Existing feature on desktop ported to QML


electrum/gui/qml/qetxfinalizer.py
```diff
@@ -868,3 +871,101 @@ def update(self):
...
+
+    @privateKeys.setter
+    def privateKeys(self, private_keys):
+        if self._private_keys != private_keys:
+            self._private_keys = private_keys
+            self.update_privkeys()
+            self.privateKeysChanged.emit()
+
+    def make_sweep_tx(self):
+        address = self._wallet.wallet.get_unused_address()  # TODO: dont fail
```
SomberNight, line 903: so let's just use `get_receiving_address` instead, to resolve that todo? :)

SomberNight (commented):
Looks great!
Just that one remark re get_unused_address, but otherwise can merge.

electrum/gui/qml/qetxfinalizer.py
accumulator: > so let's just use `get_receiving_address` instead, to resolve that todo? :)

for desktop, it is implemented with number of fallbacks:

https://github.com/spesmilo/electrum/blob/8c9263fb200fc45f0b74def98dd6c1df9f2af9bd/electrum/gui/qt/main_window.py#L2388-L2394

SomberNight: All that complexity is only there because of that "Get wallet address" button, which gives you a "new" address every time you click it. (not sure why we have that button tbh)
If you only want a single address, it is much easier, you can just call `wallet.get_receiving_address()`
pic1

accumulator: ack

accumulator:
What do you think about disabling the sweep option for `imported` wallets? It doesn't really make sense, unless the privkey is potentially compromised and no HD wallet is available, or it is a watch-only `imported` wallet with owned addresses..

SomberNight:
> What do you think about disabling the sweep option for imported wallets? It doesn't really make sense, unless the privkey is potentially compromised and no HD wallet is available, or it is a watch-only imported wallet with owned addresses..

Or the user might be sweeping lots of privkeys/utxos, and just wants to consolidate them.
Well, I don't know, but I think it's fine not to restrict this.

I saw that you disabled it already for watchonly wallets. Even that could be argued for or against. :)

Just imagine finding a random printed WIF key, e.g. as part of a scavenger hunt, you want to be able to scan and send the coins to whatever wallet you have :D
Though this is not the usecase to optimise for :D

accumulator:
> I saw that you disabled it already for watchonly wallets. Even that could be argued for or against. :)
> 
> Just imagine finding a random printed WIF key, e.g. as part of a scavenger hunt, you want to be able to scan and send the coins to whatever wallet you have :D Though this is not the usecase to optimise for :D

finding a random printed WIF key hmmmmm.. I'm sold! :-]
94c54f7057716ab490404954aca090705ef0ae86

PR #9208, opened 2024-09-19
qml: do not display success if swap failed..

ecdsa:
I am not able to test this commit because the qml gui is broken for me on master:
```
  1.82 | W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/WalletMainView.qml:332:9: Unable to assign QEWallet to QEWallet
  1.82 | W | gui.qml.qeapp | file:///opt/electrum/electrum/gui/qml/components/WalletMainView.qml:327:9: Unable to assign QEWallet to QEWallet

```
not sure what causes that issue.


ecdsa:
thanks. indeed it works for me with 6.5

Issue #9210, opened 2024-09-20
artifact url improvment for f-droid binaries


thecockatiel:
Hi
I'm trying to do the necessary work for releasing this app on f-droid, regarding issue #1700 
it would be nice if there could be some adjustments made to the website to help ease stuff a bit for reproducible builds
currently the url is like this: `https://download.electrum.org/4.5.5/Electrum-4.5.5.0-arm64-v8a-release.apk`

it would be nice if you can make `https://download.electrum.org/4.5.5.0/Electrum-4.5.5.0-arm64-v8a-release.apk` also point to the same file ( `4.5.5.0` -> `4.5.5` redirection)

I'll also open a PR for adding the fastlane metadata files for f-droid, which can help you control the metadata without opening MRs on f-droid
if you don't like the idea of metadata living inside the repo, I can still add it directly to f-droid repo, It's your decision

thecockatiel:
I have made the required metadata to build the app, but I'm having a hard time reproducing the exact apk
any tips ?
I have used the same env, sdk, ndk ,ant and steps of the dockerfile inside the gitlab runner, but it's producing different apk by a few kilobytes
most notably the pyc files inside the apk have a slightly different size than the one produced from docker image
the python version seems to be the same, so I am having a hard time tracking down the issue

thecockatiel:
can you list possible things that can cause the build to vary ?

thecockatiel:
I even tried replacing the base image with `registry.gitlab.com/fdroid/fdroidserver:buildserver-bookworm` and deleting the existing user with id 1000 and tried building the image and it still fails

thecockatiel:
the `private.tar` that is included in the apk is one part of the issue. the path at which the project is built affects the reproduction. for example inside the `main.pyc` inside the `private.tar` that is included in the assets of the apk we see this path:
![image](https://github.com/user-attachments/assets/68bc5be2-14d3-4d59-bcc3-06904cc2a57f)

which should not be there I think, but is.

SomberNight:
Thank you for looking into this!

>  it would be nice if there could be some adjustments made to the website to help ease stuff a bit for reproducible builds
currently the url is like this: https://download.electrum.org/4.5.5/Electrum-4.5.5.0-arm64-v8a-release.apk
>
[...]

PR #9211, opened 2024-09-20
chore: add fastlane metadata for f-droid

thecockatiel:
related to #9210


SomberNight:
Could you give some context: why is this called "fastlane"? Did you generate these files (the file hierarchy at least) using the [similarly named tool](https://github.com/fastlane/fastlane)?

Have you downloaded these from the google play store?
(maybe we should update that icon there, it's one of the worse looking ones :P)

Maybe this should go into `contrib/android/`.

thecockatiel:
Hi
Sure,
this is the structure f-droid expects the repo to have: https://f-droid.org/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-applications-source-repository

we can put it also in `metadata/` instead of `fastlane/...`

thecockatiel:
and no i did not use a tool to generate these, I made them by hand
and yes i did download the images from google play store

sure thing, you can adjust the metadata as you see fit, its just a starting point for f-droid to pickup the metadata

fastlane/metadata/android/en-US/full_description.txt
SomberNight, line 1: I guess we also need a `short_description.txt` (?)
[Your link](https://f-droid.org/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-applications-source-repository) says it is mandatory.

SomberNight:
> we can put it also in metadata/ instead of fastlane/..., but we cant put it in contrib/android/

OK. Then let's keep it as-is.

fastlane/metadata/android/en-US/full_description.txt
thecockatiel: oops
I did not see a short desc on google play thats why I omitted it
I'll add one

thecockatiel: done

SomberNight: Same link says it needs to be below 80 chars, so yours is too long.

> `│   ├── short_description.txt   (short description, max 80 chars, mandatory)`

AFAICT on the play store we have:
> `Fast and self-custodial wallet for Bitcoin and the Lightning Network`

It is visible when you search, before opening the page for any specific app:
https://web.archive.org/web/20240926162441/https://play.google.com/store/search?q=org.electrum.electrum&c=apps

thecockatiel: sorry for the mistakes and thanks for the help
I've updated the short description with the given text

SomberNight: No problem. Thanks for working on this!

thecockatiel:
when will this PR get merged ? just asking because this has to be inside the repo before the MR on f-droid

thecockatiel:
thanks

PR #9216, opened 2024-09-28
Unix: Allow external plugins listed in /etc/electrum/trusted_plugins

ecdsa:


electrum/plugin.py
```diff
@@ -72,22 +76,45 @@ def __init__(self, config: SimpleConfig, gui_name):
...
+        d.update(self.internal_plugin_metadata)
+        d.update(self.external_plugin_metadata)
+        d.update(self.installed_plugin_metadata)
+        return d
+
+    def is_installed(self, name):
+        return name in self.installed_plugin_metadata
+
+    def can_be_upgraded(self, name):
+        from packaging.version import Version
```
SomberNight, line 107: `packaging` is a 3rd-party library, not part of the stdlib.
Looks like we indirectly depend on it currently via qdarkstyle, but if possible I think it would still be preferable to avoid making it a direct dependency also. Is it needed here?
```
  - qdarkstyle [required: >=3.2, installed: 3.2.3]
    - qtpy [required: >=2, installed: 2.1.0]
      - packaging [required: Any, installed: 23.1]
```

SomberNight: For trivial int-only version numbers (are those not sufficient here?) we could use `electrum.util.versiontuple`.
For more advanced strings there is `electrum._vendor.distutils.version.StrictVersion` (though I don't know how exactly it compares to `packaging.version.Version`).

ecdsa: fixed

9764319764319764: 1000

9764319764319764: hamedmensvri@gmail.com

gburkhow:
userspace_plugins

gburkhow:
userspace_plugins

PR #9221, opened 2024-09-30
android: rm APK_VERSION, and change versionCode calculation

SomberNight:
This drops the trailing .0 from the version number of android apks. It is a historical thing that we don't need/utilise anymore. (Some years ago, prior to reproducible builds at least, sometimes we had .1 and .2 releases just for android-specific bugs.)
Related https://github.com/spesmilo/electrum/issues/9210#issuecomment-2376572832

Also, the android `versionCode` calculation is changed to accommodate the changed version string, but also to have the architecture part in the least significant digits, as requested in https://github.com/spesmilo/electrum/issues/9210#issuecomment-2380559324

@thecockatiel Could you check if this looks good to you? (re version string and int version code construction)

### DIFF
--- a/contrib/android/get_apk_versioncode.py
+++ b/contrib/android/get_apk_versioncode.py
@@ -1,0 +1,61 @@
+#!/usr/bin/python3
+
+import importlib.util
+import os
+import sys
+
+ARCH_DICT = {
+    "x86_64": "4",
+    "arm64-v8a": "3",
+    "armeabi-v7a": "2",
+    "x86": "1",
+}
+
+
+def get_electrum_version() -> str:
+    project_root = os.path.abspath(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
+    version_file_path = os.path.join(project_root, "electrum", "version.py")
+    # load version.py; needlessly complicated alternative to "imp.load_source":
+    version_spec = importlib.util.spec_from_file_location('version', version_file_path)
+    version_module = version = importlib.util.module_from_spec(version_spec)
+    version_spec.loader.exec_module(version_module)
+    return version.ELECTRUM_VERSION
+
+
+def get_android_versioncode(*, arch_name: str) -> int:
+    version_code = 0
+    # add ELECTRUM_VERSION
+    app_version = get_electrum_version()
+    app_version_components = app_version.split('.')
+    assert len(app_version_components) == 3, f"version str expected to have 3 components, but got {app_version!r}"
+    for i in app_version_components:
+        version_code *= 100
+        version_code += int(i)
+    # add arch
+    arch_code = ARCH_DICT[arch_name]
+    assert len(arch_code) == 1
+    version_code *= 10
+    version_code += int(arch_code)
+    # compensate for legacy scheme
+    # note: up until version 4.5.5, we used a different scheme for version_code.
+    #       4_______________4_05_05_00
+    #       ^ android arch, ^ app_version (4.5.5.0)
+    # This offset ensures that all new-scheme version codes are larger than the old-scheme version codes.
+    offset_due_to_legacy_scheme = 45_000_000
+    version_code += offset_due_to_legacy_scheme
+    return version_code
+
+
+if __name__ == '__main__':
+    try:
+        android_arch = sys.argv[1]
+    except Exception:
+        print(f"usage: {os.path.basename(__file__)} <android_arch>", file=sys.stderr)
+        sys.exit(1)
+    if android_arch not in ARCH_DICT:
+        print(f"usage: {os.path.basename(__file__)} <android_arch>", file=sys.stderr)
+        print(f"error: unknown {android_arch=}", file=sys.stderr)
+        print(f"       should be one of: {list(ARCH_DICT.keys())}", file=sys.stderr)
+        sys.exit(1)
+    version_code = get_android_versioncode(arch_name=android_arch)
+    print(version_code, file=sys.stdout)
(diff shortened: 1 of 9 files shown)


thecockatiel:
I'm not sure if offset is needed, we simply can make a major release (I know there's no breaking change, but it would help android versioning to be much simpler to reason about). then first digit will be `5`, which is higher than previously `x86_64`'s `4`:
```python
def get_android_versioncode(*, arch_name: str) -> int:
    version_code = 0
    # add ELECTRUM_VERSION
    app_version = "5.0.0" # <--------------------------- added for testing
    app_version_components = app_version.split('.')
    assert len(app_version_components) == 3, f"version str expected to have 3 components, but got {app_version!r}"
    for i in app_version_components:
        version_code *= 100
        version_code += int(i)
    # add arch
    version_code *= 100
    arch_code = ARCH_DICT[arch_name]
    assert len(arch_code) == 1
    version_code *= 10
    version_code += int(arch_code)
    # note: up until version 4.5.5, we used a different scheme for version_code.
    #       4_______________4_05_05_00
    #       ^ android arch, ^ app_version (4.5.5.0)
    # legacy scheme is now compensated by a major release:
    # 44050500 ---> 50000004
    return version_code
```

contrib/android/get_apk_versioncode.py
```diff
@@ -0,0 +1,61 @@
...
+    assert len(arch_code) == 1
+    version_code *= 10
+    version_code += int(arch_code)
+    # compensate for legacy scheme
+    # note: up until version 4.5.5, we used a different scheme for version_code.
+    #       4_______________4_05_05_00
+    #       ^ android arch, ^ app_version (4.5.5.0)
+    # This offset ensures that all new-scheme version codes are larger than the old-scheme version codes.
+    offset_due_to_legacy_scheme = 45_000_000
+    version_code += offset_due_to_legacy_scheme
```
SomberNight, line 45: > I'm not sure if offset is needed, we simply can make a major release (I know there's no breaking change, but it would help android versioning to be much simpler to reason about, and easier to predict). then first digit will be `5`, which is higher than previously `x86_64`'s `4`

As long as the major version number is a single digit (which will be true for the foreseeable future), this 45 million offset is equivalent to a simple "45" string prefix. I don't think that's too complicated.

I don't think we want to do a major version number bump just for this.

thecockatiel: hmm, are you sure you don't want to put room for patch? if patch ever goes to two digits (ex, `4..5.10`), it will fail.
currently for `4.5.5` it yields `45405054`

thecockatiel: it's actually fine, I ran some more examples and all looked fine to me, even though I thought it would not be fine

thecockatiel:
Also it would be nice if we could include the base version code in the git tag name, so it can be picked up by f-droid for `VercodeOperation`

SomberNight:
> include the base version code in the git tag name, so it can be picked up by f-droid for VercodeOperation

Do you mean having git tag names such as `4.5.5-45405050`? I have not seen other projects doing that.
Why is that need specifically?

I searched for VercodeOperation and found https://f-droid.org/en/docs/Build_Metadata_Reference/#UpdateCheckMode but it's still not clear why we should put the full versionCode into the tag. Setting the tag to the version string should be just as good I think. (We have been naming git tags as basically `ELECTRUM_VERSION` (`x.y.z`) for some time. And now with this PR removing the APK_VERSION vs ELECTRUM_VERSION distinction, tags would directly correspond to apk version strings.)

Btw re release workflow, we typically push a git tag, and then start building binaries after. The build takes a long time and we also cross-compare it across multiple people doing builds (to check reproducibility) and only then upload them to the webserver. So the http links for a new version only become live ~1-2 days after a git tag is pushed.

thecockatiel:
hmmm, the delay could cause some issues, not necessarily as f-droid build cycles are like 7 days apart. the worst is it gets disabled until manually enabled through a MR again

about the version code. we need some way to either: 1. take the version code directly from a file 2. calculate it based on some **number**
the issue is this specifically: https://gitlab.com/fdroid/fdroidserver/-/issues/1228
until they give us a way to turn the dot separated numbers into a single number, it's not possible

thecockatiel:
checking the other metadata files, it seems they also support loading the version from a remote url:
https://gitlab.com/fdroid/fdroiddata/-/blame/master/metadata/at.tomtasche.reader.yml#L297
https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/apps.amine.bou.readerforselfoss.yml#L520

SomberNight:
Well there is https://electrum.org/version. It is a json file.
It is used by the desktop gui to check for updates, so this file is guaranteed to only be updated once binaries are available on download.electrum.org.

SomberNight:
> Well there is https://electrum.org/version. It is a json file.

It is also available in the website git repo btw, along with the rest of the website:
https://github.com/spesmilo/electrum-web/blob/master/version

It is generated as part of the release process, in [`contrib/release_www.sh`](https://github.com/spesmilo/electrum/blob/b1db99fde85811c7cb76fdf9ff674fe81b0a0044/contrib/release_www.sh#L46). There we could call `get_apk_versioncode.py` for e.g. the "null" arch, and put that base versionCode somewhere into the `version` json file.

SomberNight:
Re this PR in particular, if there are no objections (?), I will merge it.

thecockatiel:
> Re this PR in particular, if there are no objections (?), I will merge it.

LGTM

> https://github.com/spesmilo/electrum-web/blob/master/version

the problem is the time lag makes it difficult, as I don't know when f-droid will try to access the url and the binaries. if f-droid picks up the tag before the release & version update, the version code extracted from the file will be out dated and invalid. maybe f-droid retries ? we have to ask to know if we can use it. but sounds good overall.

Issue #9209, opened 2024-09-19
Small translation issues on Transaction dialog


lionkmp:
Hi,

On the transaction details dialog there is a popup menu on the addresses. That looks like this in Hungarian:

![screenshot-20240919_213135](https://github.com/user-attachments/assets/1829825f-727e-4112-9a92-f1d22695bc05)

Issues:

1. "Copy Outpoint" menu is built up from two language strings `_("Copy") + " " + _("Outpoint")`, so it cannot have the right order of words and letter casing (and suffices) in Hungarian (like the two menus below, where "copy" (másolás) is the last word. The menu appears in two code lines:

https://github.com/spesmilo/electrum/blob/master/electrum/gui/qt/transaction_dialog.py#L342

https://github.com/spesmilo/electrum/blob/master/electrum/gui/qt/transaction_dialog.py#L386

2. There are "Copy", "Copy Link Location" and "Select all" menus that are English and I cannot see in the source code. 
3. Tx Input #0, and Tx Output #0 are also English.
4. On same dialog, "Locktime" "(height)" are also English

(And CrowdIn is 100% translated, except "Get wallet address", where I added a question asking what that is and where it is in the program. ;-) )

Thank you in advance.

accumulator:
Note: Electrum uses `gettext` for its translations, and `createStandardContextMenu()` is using the default `QTranslator` set for the application, so any translations originating in the standard Qt library will translated through the Qt translation system, and not through `gettext`.

The simplest fix is probably to just simply use `gettext` for everything and add the limited number of strings that are added by `createStandardContextMenu()` to the translatable strings. 

This is not ideal, but the alternative, routing some translations through Qt and others through `gettext` is also unnecessarily complex with two parallel translation systems and no clear routing heuristics for when to use which.

PR #9232, opened 2024-10-07
qt/transaction_dialog: fix translation issues

accumulator:
related: #9209


SomberNight:
Looks good.
I noticed that if fx.is_enabled() but historical rates are not, the historical values are still shown in the tx dialog, potentially with "No Data". But I see it is pre-existing, so nvm.

PR #9233, opened 2024-10-07
contrib/release_www.sh: put android versionCode in "version" file

SomberNight:
This updates the `version` file hosted at `https://electrum.org/version` to also include the android `versionCode` (or rather, a closely related value, as that would depend on the target architecture).
This can be used by the F-Droid buildserver to detect new versions and release updates automatically.

ref:
https://github.com/spesmilo/electrum/pull/9221#issuecomment-2386549086
https://f-droid.org/en/docs/Build_Metadata_Reference/#UpdateCheckMode

@thecockatiel does this look good? AFAICT UpdateCheckMode/HTTP cannot parse json files, but I think a regex should be sufficient to extract the version number from this structure.

Example file:
```
{
    "version": "4.5.5",
    "signatures": {"13xjmVAB1EATPP8RshTE8S8sNwwSUM9p1P": "H29AgDpgA0jFGPq6AiL8DAZsyq0TSyzmbHGxT7dwimRTeF28qc93agiG+oR6K+81gNVqNxvDRO2rOC1BmsW3yhE="},
    "extradata": {
        "android_versioncode_nullarch": 45405050
    }
}
```

### DIFF
--- a/contrib/android/get_apk_versioncode.py
+++ b/contrib/android/get_apk_versioncode.py
@@ -9,6 +9,7 @@ ARCH_DICT = {
     "arm64-v8a": "3",
     "armeabi-v7a": "2",
     "x86": "1",
+    "null": "0",
 }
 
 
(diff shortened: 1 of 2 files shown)


thecockatiel:
LGTM

Yes it seems to only support regex and I also think it should be sufficient in our case

PR #9234, opened 2024-10-07
use `electrum_ecc` package

SomberNight:
Move `ecc.py` and `ecc_fast.py` to separate package named [`electrum_ecc`](https://github.com/spesmilo/electrum-ecc), distributed [on PyPI](https://pypi.org/project/electrum-ecc/).

The motivation is to let other packages also use our ecc bindings, e.g. our fork of aionostr, or electrumx.


SomberNight:
note: I have tested the binary builds and also that there are no issues at runtime, however I have not tested reproducibility. In theory it should be unaffected.

PR #9235, opened 2024-10-07
Unix: Import external plugins from /opt/electrum_plugins

ecdsa:
This replaces #9216.
Plugin zip files must be placed in `/opt/electrum_plugins` and have root permissions.


SomberNight:
btw, now that we decided not to do integrity verification/etc of these plugins and just rely on root permissions, do we still want to use zip files for them? If we asked users to unpack folders, and have one folder per plugin, we could also allow non-pure python stuff.

accumulator:
Why not import gpg signing keys into the config and validate plugins against that (e.g. `.zip` + `.asc` files)?

Needing `root` feels not quite like the right fit for this..

SomberNight:
> Why not import gpg signing keys into the config and validate plugins against that (e.g. .zip + .asc files)?

This PR is like the 5th iteration of ideas over the last year :D  and at least the 3rd Thomas actually implemented.
Unfortunately most of the discussion is not documented I think.

The fundamental question is whether it would be okay to allow a process that has write-access to the datadir to add/rm/modify plugins (i.e. code). Note that such a process can modify the contents of the config (including state of a plugin being enabled if stored there).

The naive approach of just nominating a folder inside the datadir to be for external plugins and ask users to copy-paste plugins into that has this issue.

> Why not import gpg signing keys into the config and validate plugins against that (e.g. .zip + .asc files)?

As above, if you trust the config, you might as well just trust the whole datadir.

If we want integrity/authenticity checks for plugins,
- we could hardcode into the source code
  - hashes of plugins, or
  - pubkeys of plugin authors
- save plugin hashes or pubkeys into each wallet file (which can be password-protected)

Note that we already trust the installed code of Electrum. How is that different than trusting e.g. the datadir? Well, I guess in part because of access control. (e.g. root permissions if installed via `apt`, root perms if using setup.exe on Windows (inside Program Files) -- though different mechanisms for other binaries). That's the inspiration/justification for the idea in this current PR.

Issue #9236, opened 2024-10-07
Default settings allow using `electrum lnpay $invoice` without any confirmation or password


Midar:
The default settings allow using `electrum lnpay $invoice` without any confirmation or password. That seems a bit risky, as anyone who can get access to the socket gets to spend all available LN balance.

SomberNight:
> anyone who can get access to the socket gets to spend all available LN balance

Note: on Linux/macOS, the default is to use unix sockets for the RPC, on Windows we listen on a TCP port on localhost.
In either case, the RPC is protected with a random password, which is stored in the config file. So in practice to be able to run commands such as `lnpay`, the conditions are that (1) the wallet has to be open, (2) read access for the config file is needed.
(read access to the config file implies access to the unix socket or localhost tcp also)

Midar:
Hm, isn't the wallet always encrypted when using a HW wallet? In any case I'm seeing the lock symbol in Electrum and I can just use `lnpay` without any confirmation whatsoever.

SomberNight:
> Hm, isn't the wallet always encrypted when using a HW wallet?

No, you can click the lock and disable file encryption.

> I can just use lnpay without any confirmation whatsoever.

On master? Can you give repro steps?

Midar:
No, this was on 4.5.5, waiting for the next signed release before I let this touch my wallet ;). What I meant to say is that before this change at least, it already had access, even with a HW wallet.

SomberNight:
> Before this change, all commands worked without password-authentication (compare e.g. signmessage (which started a CLI hw handler), and lnpay (which just executed), both of which worked). Quirk: the "requires_password" commands required the user to explicitly pass a dummy password.

Yes that's already what I meant there ^

SomberNight:
The mentioned changes were included in the recent releases.
I am not fully satisfied with how it works now, but not sure how we can improve on this easily. We can revisit if there is a better idea.

Midar:
I checked now and it told me to unlock my wallet. It was unlocked, but a HW wallet, so I would guess this means it's WAI now, except for maybe a slightly confusing error? (I guess "Command `lnpay` cannot be used with a hardware wallet" would be more accurate).

PR #9238, opened 2024-10-08
commands: require wallet password for lnpay command

accumulator:
ref #9236

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1227,15 +1227,15 @@ class Commands:
             self.network.path_finder.liquidity_hints.reset_liquidity_hints()
             self.network.path_finder.clear_blacklist()
 
-    @command('wnl')
-    async def close_channel(self, channel_point, force=False, wallet: Abstract_Wallet = None):
+    @command('wnpl')
+    async def close_channel(self, channel_point, force=False, password=None, wallet: Abstract_Wallet = None):
         txid, index = channel_point.split(':')
         chan_id, _ = channel_id_from_funding_tx(txid, int(index))
         coro = wallet.lnworker.force_close_channel(chan_id) if force else wallet.lnworker.close_channel(chan_id)
         return await coro
 
-    @command('wnl')
-    async def request_force_close(self, channel_point, connection_string=None, wallet: Abstract_Wallet = None):
+    @command('wnpl')
+    async def request_force_close(self, channel_point, connection_string=None, password=None, wallet: Abstract_Wallet = None):
         """
         Requests the remote to force close a channel.
         If a connection string is passed, can be used without having state or any backup for the channel.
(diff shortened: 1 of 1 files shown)


accumulator:
might be useful to also require password for e.g. channel closes and reverse swaps

electrum/commands.py
```diff
@@ -1159,8 +1159,10 @@ async def decode_invoice(self, invoice: str):
         invoice = Invoice.from_bech32(invoice)
         return invoice.to_debug_json()
 
-    @command('wnl')
-    async def lnpay(self, invoice, timeout=120, wallet: Abstract_Wallet = None):
+    @command('wnpl')
+    async def lnpay(self, invoice, timeout=120, password=None, wallet: Abstract_Wallet = None):
+        if wallet.has_password():
+            wallet.check_password(password)
```
SomberNight, line 1165: Conceptually this should not be done here I think.
I moved this check to the `@command` decorator in https://github.com/spesmilo/electrum/commit/37d090c621006cc4caa7e0fb9d080b3e6a9da05c.

SomberNight:
> might be useful to also require password for e.g. channel closes and reverse swaps

Well, yes, maybe. But also something to consider is what exactly is the threat-model, because at the end of the day we might as well password-protect every command :P


```
>>> [cn for (cn, c) in electrum.commands.known_commands.items() if c.requires_wallet and not c.requires_password]
```
`['get', 'listunspent', 'freeze', 'unfreeze', 'freeze_utxo', 'unfreeze_utxo', 'ismine', 'getpubkeys', 'getbalance', 'getmpk', 'onchain_history', 'lightning_history', 'setlabel', 'listcontacts', 'getalias', 'searchcontacts', 'listaddresses', 'get_request', 'get_invoice', 'list_requests', 'list_invoices', 'createnewaddress', 'changegaplimit', 'getminacceptablegap', 'getunusedaddress', 'add_request', 'addtransaction', 'delete_request', 'delete_invoice', 'clear_requests', 'clear_invoices', 'is_synchronized', 'removelocaltx', 'get_tx_status', 'add_peer', 'list_peers', 'lnpay', 'nodeid', 'list_channels', 'enable_htlc_settle', 'close_channel', 'request_force_close', 'export_channel_backup', 'import_channel_backup', 'get_channel_ctx', 'get_watchtower_ctn', 'rebalance_channels', 'reverse_swap']`

1. Dangerous commands:
`lnpay, get_channel_ctx`

2. Not too dangerous but still kind of:
`close_channel, request_force_close, export_channel_backup, reverse_swap, rebalance_channels`

3. And then there is also stuff like `getmpk` (privacy) or `changegaplimit` (DOS-potential).

Anyway, I would say let's password-protect the "dangerous" list (1).

Do you think list (2) should be password-protected as well? I am not sure.

electrum/commands.py
accumulator: agreed

accumulator:
> Do you think list (2) should be password-protected as well? I am not sure.

These are currently inconsistent across GUIs (PIN on android, if enabled. no auth on desktop). 
I'm inclined to add password requirement for (2)

SomberNight:
>  I'm inclined to add password requirement for (2), though only for CLI

Okay, let's do that then.

SomberNight (approved):
Thanks!

PR #9243, opened 2024-10-10
qt,qml: show option for single server in ServerConfig

accumulator:


electrum/wizard.py
```diff
@@ -704,14 +704,17 @@ def do_configure_server(self, wizard_data: dict):
...
         if not wizard_data['autoconnect']:
             try:
                 server = ServerAddr.from_str_with_inference(wizard_data['server'])
                 if not server:
                     raise Exception('failed to parse server %s' % wizard_data['server'])
             except Exception:
                 return
-        net_params = net_params._replace(server=server, auto_connect=wizard_data['autoconnect'])
+        else:
+            oneserver = False
```
SomberNight, line 716: Can we avoid duplicating the default value here?
Would this work?
```suggestion
            oneserver = net_params.oneserver 
```

electrum/wizard.py
```diff
@@ -704,14 +704,17 @@ def do_configure_server(self, wizard_data: dict):
         self._logger.debug(f'configuring server: {wizard_data!r}')
         net_params = self._daemon.network.get_parameters()
         server = ''
+        oneserver = wizard_data['one_server']
```
SomberNight, line 707: you also need to set this parameter in the desktop qt gui if you refer to it here.
```
  6.15 | D | wizard | view "server_config" last: True
  6.15 | D | wizard | view=server_config
  6.15 | D | wizard | configuring server: {'use_defaults': False, 'want_proxy': False, 'autoconnect': True, 'server': ''}
  6.15 | E | gui.qt.exception_window.Exception_Hook | exception caught by crash reporter
  File "/home/user/wspace/electrum/electrum/gui/qt/wizard/wizard.py", line 214, in on_next_button_clicked
    self.submit(wd)
  File "/home/user/wspace/electrum/electrum/gui/qt/wizard/wizard.py", line 232, in submit
    view = self.resolve_next(self._current.view, wdata)
  File "/home/user/wspace/electrum/electrum/wizard.py", line 74, in resolve_next
    view_accept(wizard_data)
  File "/home/user/wspace/electrum/electrum/wizard.py", line 707, in do_configure_server
    oneserver = wizard_data['one_server']
KeyError: 'one_server'
```

electrum/wizard.py
accumulator: That was indeed the intention :)

I added a patch.
Note that it's still possible to `setconfig` both options to `true`, but it will honor `oneserver` over `auto_connect` at startup and at changing network parameters.

electrum/gui/qt/wizard/server_connect.py
```diff
@@ -110,3 +110,4 @@ def __init__(self, parent, wizard):
     def apply(self):
         self.wizard_data['autoconnect'] = self.sw.server_e.text().strip() == ''
         self.wizard_data['server'] = self.sw.server_e.text()
+        self.wizard_data['one_server'] = False
```
SomberNight, line 113: We can avoid duplicating the default value of the config setting, by delegating it to the config:
```suggestion
        self.wizard_data['one_server'] = self.wizard.config.NETWORK_ONESERVER
```

electrum/network.py
```diff
@@ -684,7 +695,14 @@ async def set_parameters(self, net_params: NetworkParameters):
                 int(proxy['port'])
         except Exception:
             return
-        self.config.NETWORK_AUTO_CONNECT = net_params.auto_connect
+        auto_connect = net_params.auto_connect
+        if auto_connect and net_params.oneserver:
+            # enabling both oneserver and auto_connect doesn't really make sense
+            # assume oneserver is enabled for privacy reasons, disable auto_connect
+            self.logger.warning(f'both "oneserver" and "auto_connect" options enabled, disabling "auto_connect".')
+            auto_connect = False
```
SomberNight, line 703: I don't know... I think respecting the auto_connect setting and disabling oneserver might be better.
If they did not disable auto_connect, what's the guarantee that the current server is selected correctly? If it is effectively a "random" server, now disabling auto_connect makes sure we won't switch away from it.

If `auto_connect and oneserver` we should either transition to `auto_connect and not oneserver`
or just kill the network by transitioning to `server==localhost:1:s and not auto_connect and not oneserver`.

SomberNight: >  or just kill the network by transitioning to server==localhost:1:s and not auto_connect and not oneserver.

well actually maybe to `server==localhost:1:s and not auto_connect and oneserver`

accumulator:
changed to draft, this needs work on desktop ui.

ecdsa:
This should come with a proper security warning.

accumulator:
> This should come with a proper security warning.

You mean showing a confirmation popup when enabling, like disabling trampoline on android?
Note that there's a help button showing the config option's long_desc

ecdsa:
> Note that there's a help button showing the config option's long_desc

The problem is not only about lagging and forks. 

The `--oneserver` option makes SPV useless. If you get your info from a single server, it can mine blocks that include fake transactions, and your client will not detect it. For example, an attacker could make you believe you received a large payment, in exchange for goods/fiat/other cryptos. They only need to spend as much PoW as in our last checkpoint, and drive the difficulty down from there.

The dangers of `--oneserver` are not well understood

accumulator:
Yes the option is only intended for connecting to your own node, the text should indicate that.

electrum/simple_config.py
```diff
@@ -614,7 +614,12 @@ def __setattr__(self, name, value):
 
     # config variables ----->
     NETWORK_AUTO_CONNECT = ConfigVar('auto_connect', default=True, type_=bool)
-    NETWORK_ONESERVER = ConfigVar('oneserver', default=False, type_=bool)
+    NETWORK_ONESERVER = ConfigVar(
+        'oneserver', default=False, type_=bool,
+        short_desc=lambda: _('Connect only to a single Electrum Server'),
+        long_desc=lambda: _('Connect only to a single Electrum Server. This is only intended for connecting to your own node.') + ' ' +
+                          _('Using this option on a public server is a security risk and is discouraged.')
```
ecdsa, line 621: When splitting messages for localization, please have a think about the work of translators.
 - "Connect only to a single Electrum Server" will be translated twice.
 - "This is only intended for connecting to your own node" and "Using this option on a public server is a security risk and is discouraged" are provided as two separate messages, meaning translators will not have the context of the first message when translating the second.

PR #9245, opened 2024-10-11
qt: refactor SeedLayout/KeysLayout to SeedWidget/KeysWidget, remove t…

accumulator:
…he hacks left over from old to new wizard and

update validation in all cases (e.g. validate electrum seed when switching from bip39 to electrum in options dialog)

### DIFF
--- a/electrum/gui/qt/seed_dialog.py
+++ b/electrum/gui/qt/seed_dialog.py
@@ -69,66 +69,10 @@ def seed_warning_msg(seed):
     ]).format(len(seed.split()))
 
 
-class SeedLayout(QVBoxLayout):
+class SeedWidget(QWidget):
 
     updated = pyqtSignal()
-
-    def seed_options(self):
-        dialog = QDialog()
-        dialog.setWindowTitle(_("Seed Options"))
-        vbox = QVBoxLayout(dialog)
-
-        seed_types = [
-            (value, title) for value, title in (
-                ('electrum', _('Electrum')),
-                ('bip39', _('BIP39 seed')),
-                ('slip39', _('SLIP39 seed')),
-            )
-            if value in self.options or value == 'electrum'
-        ]
-
-        if 'ext' in self.options:
-            cb_ext = QCheckBox(_('Extend this seed with custom words'))
-            cb_ext.setChecked(self.is_ext)
-            vbox.addWidget(cb_ext)
-
-        if len(seed_types) >= 2:
-            def on_selected(idx):
-                self.seed_type = seed_type_choice.selected_key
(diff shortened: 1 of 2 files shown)


electrum/gui/qt/seed_dialog.py
```diff
@@ -287,15 +307,13 @@ def on_edit(self):
...
+            valid = self.slip39_seed is not None
             self.update_share_buttons()
         else:
+            valid = self.is_seed(s)
             t = calc_seed_type(s)
             label = _('Seed Type') + ': ' + t if t else ''
-            if t and not b:  # electrum seed, but does not conform to dialog rules
-                # FIXME we should just accept any electrum seed and "redirect" the wizard automatically.
-                #       i.e. if user selected wallet_type=="standard" but entered a 2fa seed, accept and redirect
-                #            if user selected wallet_type=="2fa" but entered a std electrum seed, accept and redirect
```
SomberNight, line 298: Looks like this FIXME is still applicable though?
I thought you fixed it, but no, just removed the comment :D

accumulator: Not sure I agree it is a FIXME, even though this is how the old wizard behaved. 
To be fair, I thought this FIXME comment was older than it actually is (77c55d78b79ad83b825fc98a891ffabf0e8c4066), but if you really want to replicate this behavior, I think it should have its own github issue, otherwise it can easily stay off the radar for a long time.

SomberNight: Well okay, maybe it is a TODO and not a FIXME. That line is blurry and often subjective.

It is a quality of life feature - users sometimes do not know what kind of seeds they have. In the past it happened quite often that users tried restoring a 2fa seed just clicking "next"-"next"-"next" (so as a "standard wallet") and then they went to the forums/irc complaining that the "next" button is disabled.

The warning shown (next few lines of code) is a clutch to sort of handle this, at least it removes the support burden on us. So it is good enough. (Though note that the error is even more obscure and hard to understand when using the qml gui)
But in ideal world, the wizard should probably just automagically accept "any" seed.

> I think it should have its own github issue, otherwise it can easily stay off the radar for a long time.

It is not urgent at all, not even that important. It is a nice to have. I think it is useful to put TODOs/FIXMEs in the code if there is an obvious place to put them (and maybe also create an issue for them sure, and if so, probably refer to that issue from the TODO).
Just imagine a potential contributor running into this flow while using the GUI and getting annoyed by it or seeing an opportunity to improve things: if I were them, I would grep the codebase for the text shown by the GUI (e.g. `"this dialog does not support such seeds"`) and immediately find this comment. I think that is desirable.

So anyway, I am a fan of well-written TODOs/FIXMEs :P

PR #9248, opened 2024-10-12
coldcard: rebase #7682 and fix CC bugs

accumulator:


electrum/gui/qt/main_window.py
```diff
@@ -258,7 +258,7 @@ def add_optional_tab(tabs, tab, icon, description):
         QShortcut(QKeySequence("Ctrl+PgDown"), self, lambda: wrtabs.setCurrentIndex((wrtabs.currentIndex() + 1)%wrtabs.count()))
 
         for i in range(wrtabs.count()):
-            QShortcut(QKeySequence("Alt+" + str(i + 1)), self, lambda i=i: wrtabs.setCurrentIndex(i))
+            QShortcut(QKeySequence("Alt+" + str(i + 1)), self, lambda i: wrtabs.setCurrentIndex(i))
```
SomberNight, line 261: Looks unrelated to coldcard stuff?

More importantly it is incorrect. The existing code here works and this change breaks it.
(Interestingly PyCharm marks the old code as an error, but its static analyser is just not good enough I guess.)

electrum/plugins/coldcard/qt.py
```diff
@@ -48,23 +52,59 @@ def wallet_info_buttons(self, main_window: 'ElectrumWindow', dialog):
...
+        return buttons
 
-        return btn
+    def import_multisig_wallet_to_cc(self, main_window, coldcard_keystores):
+        from io import StringIO
+        from ckcc.protocol import CCProtocolPacker
 
-    def export_multisig_setup(self, main_window, wallet):
+        index = main_window.query_choice(
+            _("Please select which {} device to use:".format(self.device)),
```
SomberNight, line 84:
```suggestion
            _("Please select which {} device to use:").format(self.device),
```
and same issue everywhere

accumulator: Ah yes. Interestingly this still seems to work (?)

electrum/gui/qt/main_window.py
accumulator: Hm, I thought is was caused by me hitting a key with the cursor in the wrong spot.

electrum/plugins/coldcard/qt.py
SomberNight: What do you mean by "work"? My claim is that `_("Please select which {} device to use:".format(self.device))` cannot be translated. I don't expect other runtime problems besides that.

I think these two are equivalent:
```
(1)  f"My name: {name}"
(2)  "My name: {name}".format(name=name)
```
Meaning these two are equivalent:
```
(3)  _(f"My name: {name}")
(4)  _("My name: {name}".format(name=name))
```

And regarding (3), we clearly have a comment saying that does not work, so (4) should not work either:
https://github.com/spesmilo/electrum/blob/ffdd45ee6f7d49501582e4c935c17acd1a5a75ac/electrum/i18n.py#L46-L48


Or do you mean you tested translations and even that works like that? If so, I don't understand how.

accumulator: Heh, no need for deep semantic discussions, 'seems to work' was merely meant as not crashing

PR #9250, opened 2024-10-14
network: use TOR stream isolation

accumulator:
also refactor, for proxy instantiation, use Network instance, not a proxy dict.

### DIFF
--- a/electrum/lntransport.py
+++ b/electrum/lntransport.py
@@ -9,7 +9,7 @@ import re
 import hashlib
 import asyncio
 from asyncio import StreamReader, StreamWriter
-from typing import Optional
+from typing import Optional, TYPE_CHECKING
 from functools import cached_property
 from typing import NamedTuple, List, Tuple, Mapping, Optional, TYPE_CHECKING, Union, Dict, Set, Sequence
 
@@ -17,11 +17,17 @@ from aiorpcx import NetAddress
 import electrum_ecc as ecc
 
 from .crypto import sha256, hmac_oneshot, chacha20_poly1305_encrypt, chacha20_poly1305_decrypt, get_ecdh, privkey_to_pubkey
-from .util import MySocksProxy
+from .util import ESocksProxy
+
 
 class LightningPeerConnectionClosed(Exception): pass
 class HandshakeFailed(Exception): pass
 class ConnStringFormatError(Exception): pass
+
+
+if TYPE_CHECKING:
+    from electrum.network import Network
+
 
 class HandshakeState(object):
     prologue = b"lightning"
@@ -342,18 +348,18 @@ class LNTransport(LNTransportBase):
     """Transport initiated by local party."""
 
     def __init__(self, privkey: bytes, peer_addr: LNPeerAddr, *,
-                 proxy: Optional[dict]):
+                 e_proxy: Optional['ESocksProxy']):
         LNTransportBase.__init__(self)
         assert type(privkey) is bytes and len(privkey) == 32
         self.privkey = privkey
         self.peer_addr = peer_addr
-        self.proxy = MySocksProxy.from_proxy_dict(proxy)
+        self.e_proxy = e_proxy
 
     async def handshake(self):
-        if not self.proxy:
+        if not self.e_proxy:
             self.reader, self.writer = await asyncio.open_connection(self.peer_addr.host, self.peer_addr.port)
         else:
-            self.reader, self.writer = await self.proxy.open_connection(self.peer_addr.host, self.peer_addr.port)
+            self.reader, self.writer = await self.e_proxy.open_connection(self.peer_addr.host, self.peer_addr.port)
         hs = HandshakeState(self.peer_addr.pubkey)
         # Get a new ephemeral key
         epriv, epub = create_ephemeral_key()
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1938,7 +1938,7 @@ class NetworkRetryManager(Generic[_NetAddrType]):
         self._last_tried_addr.clear()
 
 
-class MySocksProxy(aiorpcx.SOCKSProxy):
+class ESocksProxy(aiorpcx.SOCKSProxy):
     # note: proxy will not leak DNS as create_connection()
     # sets (local DNS) resolve=False by default
 
@@ -1952,12 +1952,17 @@ class MySocksProxy(aiorpcx.SOCKSProxy):
         return reader, writer
 
     @classmethod
-    def from_proxy_dict(cls, proxy: dict = None) -> Optional['MySocksProxy']:
-        if not proxy:
+    def from_network_settings(cls, network: Optional['Network']) -> Optional['ESocksProxy']:
+        if not network or not network.proxy:
             return None
+        proxy = network.proxy
         username, pw = proxy.get('user'), proxy.get('password')
         if not username or not pw:
-            auth = None
+            # is_proxy_tor is tri-state; None indicates it is still probing the proxy to test for TOR
+            if network.is_proxy_tor:
+                auth = aiorpcx.socks.SOCKSRandomAuth()
+            else:
+                auth = None
         else:
             auth = aiorpcx.socks.SOCKSUserAuth(username, pw)
         addr = aiorpcx.NetAddress(proxy['host'], proxy['port'])
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -44,7 +44,7 @@ from aiorpcx.jsonrpc import JSONRPC, CodeMessageError
 from aiorpcx.rawsocket import RSClient
 import certifi
 
-from .util import (ignore_exceptions, log_exceptions, bfh, MySocksProxy,
+from .util import (ignore_exceptions, log_exceptions, bfh, ESocksProxy,
                    is_integer, is_non_negative_integer, is_hash256_str, is_hex_str,
                    is_int_or_float, is_non_negative_int_or_float, OldTaskGroup)
 from . import util
@@ -375,7 +375,7 @@ class Interface(Logger):
 
     LOGGING_SHORTCUT = 'i'
 
-    def __init__(self, *, network: 'Network', server: ServerAddr, proxy: Optional[dict]):
+    def __init__(self, *, network: 'Network', server: ServerAddr):
         self.ready = network.asyncio_loop.create_future()
         self.got_disconnected = asyncio.Event()
         self.server = server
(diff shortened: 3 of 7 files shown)


accumulator:
we should perhaps optionally shield this feature by a config setting.

electrum/util.py
```diff
@@ -1952,12 +1952,16 @@ async def open_connection(self, host=None, port=None, **kwargs):
...
+        proxy = network.proxy
         username, pw = proxy.get('user'), proxy.get('password')
         if not username or not pw:
-            auth = None
+            if network.is_proxy_tor:
+                auth = aiorpcx.socks.SOCKSRandomAuth
+            else:
+                auth = None
         else:
             auth = aiorpcx.socks.SOCKSUserAuth(username, pw)
```
SomberNight, line 1966: Notice how auth have different types here. This should raise red flags in the reader's mind.

Indeed the random auth is not working. aiorpcx has a bad implementation here, it should really really raise an exception instead of silently ignoring the random auth.

https://github.com/kyuupichan/aiorpcX/blob/05ab957077f961db7ab4acd3f06c690d63eda6ab/aiorpcx/socks.py#L184
https://github.com/kyuupichan/aiorpcX/blob/05ab957077f961db7ab4acd3f06c690d63eda6ab/aiorpcx/socks.py#L201

You need `auth = aiorpcx.socks.SOCKSRandomAuth()`
not `auth = aiorpcx.socks.SOCKSRandomAuth`

The second one is a class, the first one is an instance of the class.

You can test with
```patch
diff --git a/aiorpcx/socks.py b/aiorpcx/socks.py
index 4a6cefa..2dc3da7 100755
--- a/aiorpcx/socks.py
+++ b/aiorpcx/socks.py
@@ -182,6 +182,7 @@ class SOCKS5(SOCKSBase):
         super().__init__()
         self._dst_bytes = SOCKS5._destination_bytes(remote_address.host, remote_address.port)
         self._auth_bytes, self._auth_methods = SOCKS5._authentication(auth)
+        print(f"{self._auth_bytes=}, {self._auth_methods=}")
 
     @staticmethod
     def _destination_bytes(host, port):
```

electrum/lntransport.py
```diff
@@ -342,12 +348,12 @@ class LNTransport(LNTransportBase):
     """Transport initiated by local party."""
 
     def __init__(self, privkey: bytes, peer_addr: LNPeerAddr, *,
-                 proxy: Optional[dict]):
+                 network: Optional['Network']):
```
SomberNight, line 351: I don't think we should pass the network to LNTransport, they should really not care about each other.
How about passing a MySocksProxy instance. Which by the way we should probably rename :D maybe `ESocksProxy` and the field could be named `e_proxy`, to distinguish it from proxy dicts

So e.g.
```patch
diff --git a/electrum/interface.py b/electrum/interface.py
index eeafd67efa..86aeee6ba7 100644
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -44,7 +44,7 @@ from aiorpcx.jsonrpc import JSONRPC, CodeMessageError
 from aiorpcx.rawsocket import RSClient
 import certifi
 
-from .util import (ignore_exceptions, log_exceptions, bfh, MySocksProxy,
+from .util import (ignore_exceptions, log_exceptions, bfh, ESocksProxy,
                    is_integer, is_non_negative_integer, is_hash256_str, is_hex_str,
                    is_int_or_float, is_non_negative_int_or_float, OldTaskGroup)
 from . import util
@@ -396,7 +396,7 @@ class Interface(Logger):
             self.logger.info(f"looks like localhost: not using proxy for this server")
             self.proxy = None
         else:
-            self.proxy = MySocksProxy.from_network_settings(network)
+            self.proxy = ESocksProxy.from_network_settings(network)
 
         # Latest block header and corresponding height, as claimed by the server.
         # Note that these values are updated before they are verified.
diff --git a/electrum/lntransport.py b/electrum/lntransport.py
index f79b5d0847..0f40daa42f 100644
--- a/electrum/lntransport.py
+++ b/electrum/lntransport.py
@@ -17,7 +17,7 @@ from aiorpcx import NetAddress
 import electrum_ecc as ecc
 
 from .crypto import sha256, hmac_oneshot, chacha20_poly1305_encrypt, chacha20_poly1305_decrypt, get_ecdh, privkey_to_pubkey
-from .util import MySocksProxy
+from .util import ESocksProxy
 
 
 class LightningPeerConnectionClosed(Exception): pass
@@ -348,18 +348,18 @@ class LNTransport(LNTransportBase):
     """Transport initiated by local party."""
 
     def __init__(self, privkey: bytes, peer_addr: LNPeerAddr, *,
-                 network: Optional['Network']):
+                 e_proxy: Optional['ESocksProxy']):
         LNTransportBase.__init__(self)
         assert type(privkey) is bytes and len(privkey) == 32
         self.privkey = privkey
         self.peer_addr = peer_addr
-        self.proxy = MySocksProxy.from_network_settings(network)
+        self.e_proxy = e_proxy
 
     async def handshake(self):
-        if not self.proxy:
+        if not self.e_proxy:
             self.reader, self.writer = await asyncio.open_connection(self.peer_addr.host, self.peer_addr.port)
         else:
-            self.reader, self.writer = await self.proxy.open_connection(self.peer_addr.host, self.peer_addr.port)
+            self.reader, self.writer = await self.e_proxy.open_connection(self.peer_addr.host, self.peer_addr.port)
         hs = HandshakeState(self.peer_addr.pubkey)
         # Get a new ephemeral key
         epriv, epub = create_ephemeral_key()
index 9a5253905a..67464b6ea9 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -31,7 +31,7 @@ from electrum_ecc import ecdsa_der_sig_from_ecdsa_sig64
 
 from . import constants, util
 from . import keystore
-from .util import profiler, chunks, OldTaskGroup
+from .util import profiler, chunks, OldTaskGroup, ESocksProxy
 from .invoices import Invoice, PR_UNPAID, PR_EXPIRED, PR_PAID, PR_INFLIGHT, PR_FAILED, PR_ROUTING, LN_EXPIRY_NEVER
 from .invoices import BaseInvoice
 from .util import NetworkRetryManager, JsonRPCClient, NotEnoughFunds
@@ -346,7 +346,7 @@ class LNWorker(Logger, EventListener, NetworkRetryManager[LNPeerAddr]):
         if node_id == self.node_keypair.pubkey:
             raise ErrorAddingPeer("cannot connect to self")
         transport = LNTransport(self.node_keypair.privkey, peer_addr,
-                                network=self.network)
+                                e_proxy=ESocksProxy.from_network_settings(self.network))
         peer = await self._add_peer_from_transport(node_id=node_id, transport=transport)
         assert peer
         return peer
@@ -3040,7 +3040,7 @@ class LNWallet(LNWorker):
         async def _request_fclose(addresses):
             for host, port, timestamp in addresses:
                 peer_addr = LNPeerAddr(host, port, node_id)
-                transport = LNTransport(privkey, peer_addr, network=self.network)
+                transport = LNTransport(privkey, peer_addr, e_proxy=ESocksProxy.from_network_settings(self.network))
                 peer = Peer(self, node_id, transport, is_channel_backup=True)
                 try:
                     async with OldTaskGroup(wait=any) as group:
diff --git a/electrum/network.py b/electrum/network.py
index d9ba81217d..e0a0ff24a5 100644
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -323,7 +323,7 @@ class Network(Logger, NetworkRetryManager[ServerAddr]):
 
         self._allowed_protocols = {PREFERRED_NETWORK_PROTOCOL}
 
-        self.proxy = None
+        self.proxy = None  # type: Optional[dict]
         self.is_proxy_tor = None
         self._init_parameters_from_config()
 
diff --git a/electrum/util.py b/electrum/util.py
index 1828c44908..3ff7887675 100644
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1938,7 +1938,7 @@ class NetworkRetryManager(Generic[_NetAddrType]):
         self._last_tried_addr.clear()
 
 
-class MySocksProxy(aiorpcx.SOCKSProxy):
+class ESocksProxy(aiorpcx.SOCKSProxy):
     # note: proxy will not leak DNS as create_connection()
     # sets (local DNS) resolve=False by default
 
@@ -1952,7 +1952,7 @@ class MySocksProxy(aiorpcx.SOCKSProxy):
         return reader, writer
 
     @classmethod
-    def from_network_settings(cls, network: Optional['Network']) -> Optional['MySocksProxy']:
+    def from_network_settings(cls, network: Optional['Network']) -> Optional['ESocksProxy']:
         if not network or not network.proxy:
             return None
         proxy = network.proxy
diff --git a/tests/test_lntransport.py b/tests/test_lntransport.py
index b522c83c15..ac057fbd26 100644
--- a/tests/test_lntransport.py
+++ b/tests/test_lntransport.py
@@ -78,7 +78,7 @@ class TestLNTransport(ElectrumTestCase):
             responder_shaked.set()
         async def connect(port: int):
             peer_addr = LNPeerAddr('127.0.0.1', port, responder_key.get_public_key_bytes())
-            t = LNTransport(initiator_key.get_secret_bytes(), peer_addr, network=None)
+            t = LNTransport(initiator_key.get_secret_bytes(), peer_addr, e_proxy=None)
             await t.handshake()
             async with OldTaskGroup() as group:
                 await group.spawn(read_messages(t, messages_sent_by_server))

```

SomberNight (changes requested):

electrum/util.py
accumulator: Yes I know the difference. Nice that it is silently ignored :)

electrum/lntransport.py
accumulator: Passing an `ESocksProxy` instance is actually even better than a `Network`..

As long as the horrible `dict` is gone

SomberNight (approved):
looks good :)

electrum/util.py
```diff
@@ -1952,12 +1952,16 @@ async def open_connection(self, host=None, port=None, **kwargs):
...
-    def from_proxy_dict(cls, proxy: dict = None) -> Optional['MySocksProxy']:
-        if not proxy:
+    def from_network_settings(cls, network: Optional['Network']) -> Optional['ESocksProxy']:
+        if not network or not network.proxy:
             return None
+        proxy = network.proxy
         username, pw = proxy.get('user'), proxy.get('password')
         if not username or not pw:
-            auth = None
+            if network.is_proxy_tor:
```
SomberNight, line 1961: could you add a comment/fixme here saying that `is_proxy_tor` is actually tri-state, and if it's value is None, then we don't yet know if it is Tor
This actually makes all this prone to timing (`is_proxy_tor` gets set to None on network settings changes, briefly).

accumulator: Done. It is briefly None only on proxy changes, not on other network changes.

not a FIXME, a tri-state here is actually useful, and one can hook the 'tor_probed' callback to defer actions until the tor probe is done.

PR #9254, opened 2024-10-16
tests: initial wizard test cases

accumulator:


tests/test_wizard.py
```diff
@@ -0,0 +1,249 @@
...
+        v = w._current
+        d = v.wizard_data
+        self.assertEqual('keystore_type', v.view)
+
+        d.update({'keystore_type': 'createseed'})
+        v = w.resolve_next(v.view, d)
+        self.assertEqual('create_seed', v.view)
+
+        d.update({'seed': '9dk', 'seed_type': 'segwit', 'seed_extend': False, 'seed_variant': 'electrum',
+                  'seed_extra_words': False})
```
SomberNight, line 147: Just to clarify, these fields are populated by the GUIs and the base lib, that's why you need to set them all manually here, right?
E.g. at first glance I would have expected not to have to specify the "seed" in the "createseed" flow, but I guess that must be the reason.
So these tests check whether the fields are properly processed and test if "resolve_next" works, but they cannot test if population of the fields is working?

SomberNight (commented):
Thanks a lot for writing these!!

tests/test_wizard.py
accumulator: > Just to clarify, these fields are populated by the GUIs and _not_ the base lib, that's why you need to set them all manually here, right? E.g. at first glance I would have expected not to have to specify the "seed" in the "createseed" flow, but I guess that must be the reason. So these tests check whether the fields are properly processed and test if "resolve_next" works, but they cannot test if population of the fields is working?

Correct. To test populating the wizard_data, we should create testcases that use the UI classes.

The testcases in this commit are simply testing the wizard flow.

PR #9256, opened 2024-10-18
qt: replace BlockingWaitingDialog with RunCoroutineDialog

ecdsa:
RunCoroutineDialog has a run() method that blocks the thread without blocking the GUI (using exec), and a Cancel button that cancels the coroutine.

main_window.run_coroutine_dialog() is a wrapper that returns the coroutine result and may raise exceptions.

BlockingWaitingDialog was removed is transaction_dialog, where it was not particularly useful.

### DIFF
--- a/electrum/gui/qt/confirm_tx_dialog.py
+++ b/electrum/gui/qt/confirm_tx_dialog.py
@@ -532,7 +532,7 @@ class TxEditor(WindowModalDialog):
             if self.not_enough_funds:
                 self.io_widget.update(None)
             self.set_feerounding_visibility(False)
-            self.messages = []
+            self.messages = [_('Preparing transaction...')]
         else:
             self.messages = self.get_messages()
             self.update_fee_fields()
@@ -619,7 +619,7 @@ class ConfirmTxDialog(TxEditor):
             title=_("New Transaction"), # todo: adapt title for channel funding tx, swaps
             allow_preview=allow_preview)
 
-        self.update()
+        self.trigger_update()
 
     def _update_amount_label(self):
         tx = self.tx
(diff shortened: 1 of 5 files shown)


electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -619,7 +619,7 @@ def __init__(self, *, window: 'ElectrumWindow', make_tx, output_value: Union[int
             title=_("New Transaction"), # todo: adapt title for channel funding tx, swaps
             allow_preview=allow_preview)
 
-        BlockingWaitingDialog(window, _("Preparing transaction..."), self.update)
```
SomberNight, line 622: > BlockingWaitingDialog was removed is transaction_dialog, where it was not particularly useful.

I have a testnet wallet where `wallet.make_unsigned_transaction()` takes 17 seconds (which is called from self.update).
That was actually the motivation I added this wrapper in https://github.com/spesmilo/electrum/pull/5833 IIRC

ecdsa: why is that useful? if it takes 17 seconds to sign the transaction, then BlockingWaitingDialog is going to block the GUI as well...

SomberNight: The point of BlockingWaitingDialog was to give feedback to the user about what is going on. If you remove it without a replacement, the GUI still freezes, but now without even telling the user why.

ecdsa: fixed in 6635e40a2c0d89ef68c7964d205f94287448f8aa

PR #9260, opened 2024-10-19
Swaps over nostr

ecdsa:
Swaps over Nostr:
- Separation between SwapManager and its transport: Legacy transpport uses http, Nostr uses websockets
- The transport uses a context to open/close connections. This context is not async, because it needs to be called from the GUI
- Initialize swapserver fees values to None instead of 0 (so that they cannot be used before swap manager is initialized)
- Remove swapserver fees disk caching (swap_pairs file)
- Regtests set config 'nostr_relays' to an empty string, so that the swap manager falls back to using the http transport

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -8,7 +8,15 @@ import time
 
 import attr
 import aiohttp
+
+import electrum_ecc as ecc
 from electrum_ecc import ECPrivkey
+
+import electrum_aionostr as aionostr
+from electrum_aionostr.util import to_nip19
+
+from collections import defaultdict
+
 
 from . import lnutil
 from .crypto import sha256, hash_160
@@ -16,7 +24,7 @@ from .bitcoin import (script_to_p2wsh, opcodes,
                       construct_witness)
 from .transaction import PartialTxInput, PartialTxOutput, PartialTransaction, Transaction, TxInput, TxOutpoint
 from .transaction import script_GetOp, match_script_against_template, OPPushDataGeneric, OPPushDataPubkey
-from .util import log_exceptions, BelowDustLimit, OldTaskGroup
+from .util import log_exceptions, BelowDustLimit, OldTaskGroup, age
 from .lnutil import REDEEM_AFTER_DOUBLE_SPENT_DELAY
 from .bitcoin import dust_threshold, DummyAddress
 from .logging import Logger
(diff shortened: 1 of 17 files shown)


ecdsa:
Notes
- both this pull request and #9261 are based on [828c71f](https://github.com/spesmilo/electrum/pull/9260/commits/828c71f14d3afcd1bad29033439769335bf65f12)
- the current `electrum_aionostr` package uses `websockets`, which does not support a proxy

electrum/submarine_swaps.py
```diff
@@ -193,38 +211,62 @@ def __init__(self, *, wallet: 'Abstract_Wallet', lnworker: 'LNWallet'):
...
-        # api url
-        self.api_url = wallet.config.SWAPSERVER_URL
-        # init default min & max
-        self.init_min_max_values()
+        self.is_server = self.config.get('enable_plugin_swapserver', False)
+        self.is_initialized = asyncio.Event()
+
+    @property
+    def use_nostr(self):
+        return bool(self.config.NOSTR_RELAYS)
```
accumulator, line 219: I don't like this. `use_nostr` is now defined implicitly as an empty set of nostr relays.

in `create_sm_transport` above the user is asked whether to enable nostr, but this is conditional on `use_nostr` and the existence of a selected npub. if no npub is found or selected, the user is presented with this question again later. 

Better to make this a separate bool config option, and expose in the preferences. We then only have to ask this once.

electrum/submarine_swaps.py
```diff
@@ -193,38 +211,62 @@ def __init__(self, *, wallet: 'Abstract_Wallet', lnworker: 'LNWallet'):
...
-        try:
-            async with self.taskgroup as group:
-                await group.spawn(self.pay_pending_invoices())
-        except Exception as e:
-            self.logger.exception("taskgroup died.")
-        finally:
-            self.logger.info("taskgroup stopped.")
+        tasks = [self.pay_pending_invoices()]
+        if self.is_server:
+            if self.config.SWAPSERVER_PORT:
```
accumulator, line 250: same as above, http choice implicit on value of SWAPSERVER_PORT

electrum/submarine_swaps.py
ecdsa: The goal is not to let the user choose between `http` and `nostr`. `http` is kept only for testing. 
So there is no need to expose another setting.

electrum/submarine_swaps.py
ecdsa: Contrary to the client, the server, does not choose between http and nostr. it may offer both.

electrum/submarine_swaps.py
ecdsa: ok, I changed this slightly, now the client-side choice between http and nostr depends on `SWAPSERVER_URL` being set. That way, we don't have to nuke the relay set in order to enable http.

PR #9261, opened 2024-10-19
Psbt nostr

ecdsa:
Replacement for cosignerpool plugin.
Encryption uses xpubs.


electrum/plugins/psbt_nostr/qt.py
```diff
@@ -0,0 +1,269 @@
...
+        self.logger.info(f'relays {self.relays}')
+        self.obj = QReceiveSignalObject()
+        self.obj.cosigner_receive_signal.connect(self.on_receive)
+
+        self.cosigner_list = []  # type: List[Tuple[str, bytes, str]]
+
+        for key, keystore in wallet.keystores.items():
+            xpub = keystore.get_master_public_key()  # type: str
+            privkey = sha256('nostr_psbt:' + xpub)
+            pubkey = ecc.ECPrivkey(privkey).get_public_key_bytes()[1:]
```
SomberNight, line 138: Not sure if we want to use ypub/zpubs here for a new protocol.
but I'll think about this a bit more.

ecdsa: yes, good point.

PR #9264, opened 2024-10-20
Anchor commitments 2024

ecdsa:
this was rebased in july 2024 from #7509
I guess it needs to be rebased again :-) 

Here are my personal notes regarding this rebase:
- <s> some tests are broken (`REDEEMED` state reached too early in test_breach)</s>
- <s> we should use password in memory (daemon `unlock` command) </s>
- if no password in memory (non-android GUI): maybe warn user?
- stop LN support for watching-only and hardware wallets

### DIFF
--- a/electrum/lnsweep.py
+++ b/electrum/lnsweep.py
@@ -56,16 +56,18 @@ def txs_their_ctx_watchtower(chan: 'Channel', ctx: Transaction, per_commitment_s
     # prep
     ctn = extract_ctn_from_tx_and_chan(ctx, chan)
     pcp = ecc.ECPrivkey(per_commitment_secret).get_public_key_bytes(compressed=True)
-    this_conf, other_conf = get_ordered_channel_configs(chan=chan, for_us=False)
-    other_revocation_privkey = derive_blinded_privkey(other_conf.revocation_basepoint.privkey,
-                                                      per_commitment_secret)
-    to_self_delay = other_conf.to_self_delay
-    this_delayed_pubkey = derive_pubkey(this_conf.delayed_basepoint.pubkey, pcp)
+    breacher_conf, watcher_conf = get_ordered_channel_configs(chan=chan, for_us=False)
+    watcher_revocation_privkey = derive_blinded_privkey(
+        watcher_conf.revocation_basepoint.privkey,
+        per_commitment_secret
+    )
+    to_self_delay = watcher_conf.to_self_delay
+    breacher_delayed_pubkey = derive_pubkey(breacher_conf.delayed_basepoint.pubkey, pcp)
     txs = []
-    # to_local
-    revocation_pubkey = ecc.ECPrivkey(other_revocation_privkey).get_public_key_bytes(compressed=True)
+    # create justice tx for breacher's to_local output
+    revocation_pubkey = ecc.ECPrivkey(watcher_revocation_privkey).get_public_key_bytes(compressed=True)
     witness_script = make_commitment_output_to_local_witness_script(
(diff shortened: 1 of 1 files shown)


electrum/lnsweep.py
```diff
@@ -139,56 +201,79 @@ def create_sweeptx_for_their_revoked_ctx(
...
+    htlc_outputs_idxs = [idx for idx, output in enumerate(htlc_tx.outputs()) if output.address == htlc_address]
+    if not htlc_outputs_idxs:
+        return {}
+
+    index_to_sweepinfo = {}
+    for output_idx in htlc_outputs_idxs:
+        # generate justice transactions
+        gen_tx = lambda: tx_sweep_htlctx_output(
+            sweep_address=sweep_address,
+            output_idx=output_idx,
```
SomberNight, line 244: Oh wow, amazing that the linter caught this!
```
./electrum/lnsweep.py:244:24: B023 Function definition does not bind loop variable 'output_idx'.
            output_idx=output_idx,
                       ^
1     B023 Function definition does not bind loop variable 'output_idx'.
```

electrum/lnchannel.py
```diff
@@ -851,15 +868,20 @@ def sweep_address(self) -> str:
...
         assert self.is_static_remotekey_enabled()
         our_payment_pubkey = self.config[LOCAL].payment_basepoint.pubkey
-        addr = make_commitment_output_to_remote_address(our_payment_pubkey)
-        if self.lnworker:
-            assert self.lnworker.wallet.is_mine(addr)
+        addr = make_commitment_output_to_remote_address(our_payment_pubkey, has_anchors=self.has_anchors())
+        # this assert fails with anchor output channels
+        #if self.lnworker:
+        #    assert self.lnworker.wallet.is_mine(addr)
         return addr
```
SomberNight, line 875: Something weird happen in a rebase here.
The original branch looks like this: https://github.com/bitromortac/electrum/blob/5cb97c077c799aa6ebd06492645b9c4f738f85f8/electrum/lnchannel.py#L325
which looks okay.

However this new implementation is not good. There is a reason that assert is there, and it failing signals a problem.
The `sweep_address()` method is supposed to return a wallet address that we can send coins to as part of closing this channel. It is called from a lot of places, e.g. code that is trying to sweep HTLCs will send coins here, or even the mutual close will send our coins here. Clearly we don't want the mutual close to send coins to a CSV-locked non-wallet address :)

I've tried to clarify this method in https://github.com/spesmilo/electrum/commit/dd140df17dbba95108f1383c2e5d4c7c8d3e3b9e

ecdsa: fixed in ccd4e2bcf08901a95a0a46d4367b8a10e8976d56

electrum/lnsweep.py
ecdsa: replaced by `partial`

ecdsa:
Notes about this merge: 
 - commits were merged individually (there is no merge commit).
 - anchor outputs can be activated by config variable `enable_anchor_channels`
 - unit tests are performed both with and without anchors. However, regtest are run only with anchors. (unlike in the original PR)
 - the commit that reorders functions in `lnsweep` was left out due to difficulty rebasing.
 - password-storing code was left out. Instead, htlc sweeping assumes that the wallet is unlocked, and calls `wallet.get_unlocked_password()`
 - a confusion between sats and msats was fixed (see `FIXED_ANCHOR_SAT`)
 - the `for` loop calling `extract_preimage_from_htlc` in f4ebe6e75f07049934438c537816d5fd94a6a941 was left out. I believe it is not needed.

PR #9265, opened 2024-10-20
lnworker: reserve wallet addresses also for chan backups

SomberNight:
We were already reserving wallet addresses for full channels. Now we also do the same for imported channel backups. (but not for onchain, as we don't have enough info for that)

Without this, if the same seed is used on multiple devices (with each device having its own set of LN channels), the wallet instances will reuse keys (specifically the payment_basepoint, which for static_remotekey chans is used as the to_remote output so it is trivially visible onchain). Now with this change, at least if the wallet instances have imported channel backups of each other, this reuse is avoided.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -93,6 +93,7 @@ if TYPE_CHECKING:
     from .network import Network
     from .exchange_rate import FxThread
     from .submarine_swaps import SwapData
+    from .lnchannel import AbstractChannel
 
 
 _logger = get_logger(__name__)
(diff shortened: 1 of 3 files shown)


electrum/wallet.py
```diff
@@ -2033,13 +2034,20 @@ def is_address_reserved(self, addr: str) -> bool:
...
             return
         with self.lock:
+            has_changed = (addr in self._reserved_addresses) != reserved
             if reserved:
                 self._reserved_addresses.add(addr)
             else:
                 self._reserved_addresses.discard(addr)
-            self.db.put('reserved_addresses', list(self._reserved_addresses))
+            if has_changed:
+                self.db.put('reserved_addresses', list(self._reserved_addresses))
```
SomberNight, line 2046: We could remove this whole data structure from the wallet DB. With this change I am effectively generating it on startup anyway, so we don't need the persistence.

ecdsa: I' always for removing things :-)

PR #9269, opened 2024-10-21
Make it possible to turn off automatic txout ordering

### LINKED ISSUE #9252: Allow disabling of BIP69 on transaction send window
Unfortunately the world created a protocol, which requires the end-user to send transaction in a very specific format (the outputs ordering and the exact spot of the memo (OP_RETURN script) matters.

The protocol is called Thorchain and you can see the docs here: https://dev.thorchain.org/concepts/sending-transactions.html#utxo-chains

As you can see, they say very explicitly in which order the transaction has to order the outputs, and they even have a danger field saying that otherwise funds will be lost. It's kind of unfortunate they call this "randomised" order, because the whole point of BIP69 is to be consistent, not random, but well, thor people are runed :)

Would it be possible to add a feature to electrum similar to "Enable output value rounding" called "Do not reorder outputs (disable BIP69)"? Would a PR to that effect be appreciated?

errge:
Closes #9252

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -991,6 +991,11 @@ class SimpleConfig(Logger):
         short_desc=lambda: _('Send change to Lightning'),
         long_desc=lambda: _('If possible, send the change of this transaction to your channels, with a submarine swap'),
     )
+    WALLET_SORT_TX_OUTPUTS = ConfigVar(
+        'sort_tx_outputs', default=True, type_=bool,
+        short_desc=lambda: _('Sort transaction outputs'),
+        long_desc=lambda: _('Provides protection against client fingerprinting, but smart-contract like protocols can depend on UTXO ordering'),
+    )
 
     FX_USE_EXCHANGE_RATE = ConfigVar('use_exchange_rate', default=False, type_=bool)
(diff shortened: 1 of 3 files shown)


electrum/simple_config.py
```diff
@@ -991,6 +991,11 @@ def _default_swapserver_url(self) -> str:
         short_desc=lambda: _('Send change to Lightning'),
         long_desc=lambda: _('If possible, send the change of this transaction to your channels, with a submarine swap'),
     )
+    WALLET_SORT_TX_OUTPUTS = ConfigVar(
+        'sort_tx_outputs', default=True, type_=bool,
+        short_desc=lambda: _('Sort transaction outputs'),
+        long_desc=lambda: _('Provides protection against client fingerprinting, but smart-contract like protocols can depend on UTXO ordering'),
```
ecdsa, line 997: This description is cryptic. If this is going to be exposed in the GUI, it should be improved.

errge: Done, thanks for the comment, the old text was really cryptic, I agree.

Could you please check the new one and give me ideas if you know how to make it better?

ecdsa:
I do not think this is the right approach regarding thorchain, see my comment in #9252

PR #9270, opened 2024-10-21
Make lntransport not require lnutil.

ecdsa:
This will be useful if we decide to ship lntransport as a separate package. It is also a conceptual cleanup.

Notes:
 - lntransport still requires crypto.py
 - parsing node id from a bolt11 invoice is not supported.

### DIFF
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -13,7 +13,6 @@ import sys
 import electrum_ecc as ecc
 from electrum_ecc import CURVE_ORDER, ecdsa_sig64_from_der_sig, ECPubkey, string_to_number
 import attr
-from aiorpcx import NetAddress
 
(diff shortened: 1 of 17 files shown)


electrum/lnutil.py
```diff
@@ -414,10 +413,7 @@ def formatted_tuple(self):
 
 
 class LightningError(Exception): pass
-class LightningPeerConnectionClosed(LightningError): pass
 class UnableToDeriveSecret(LightningError): pass
-class HandshakeFailed(LightningError): pass
-class ConnStringFormatError(LightningError): pass
 class RemoteMisbehaving(LightningError): pass
```
SomberNight, line 417: note: Hmm looks like we never rely on these being subclasses of `LightningError`

SomberNight (commented):
Looks reasonable.

PR #9275, opened 2024-10-22
qml: qetxfinalizer: use wallet.get_tx_fee_warning in all subclasses

SomberNight:
Show high fee warnings also for RBF/dscancel in qml gui.


accumulator:
LGTM

Issue #9089, opened 2024-06-09
Seed extension prompt does not support empty extension


AbdussamadA:
The window prompting for the seed extension says to leave the field blank if you don't know what it's for but if you do that you can't proceed to the next step because the next button is greyed out. It should not advise people to enter an empty extension if it can't accept them.

PR #9279, opened 2024-10-24
qt: wizard: implement user feedback TODOs, consistently use wizard.check_multisig_constraints

accumulator:


accumulator:
This also allows empty seed extension even if seed extension switch is checked (#9089)

PR #9285, opened 2024-10-28
qt: additional validation for master keys in WCHaveMasterKey in wallet wizard

accumulator:


ecdsa:
If I understand correctly, the current wizard allows the user to enter the wrong key type. For example:
 - it is currently possible to choose 'multisig', and to enter zpubs instead of Zpubs
 - it is possible to choose 'standard' and to enter a Zpub instead of zpub

In both cases, the exception raised by `create_storage` is what prevents the creation of the wallet. 
Thus, I think that the sanity check in `create_storage` should not be removed, if that is what you mean by "deduplicate".

accumulator:
> In both cases, the exception raised by `create_storage` is what prevents the creation of the wallet. Thus, I think that the sanity check in `create_storage` should not be removed, if that is what you mean by "deduplicate".

No the check should not be removed from `create_storage`, I only mean the code could be deduplicated as there's overlapping checks (`xpub_type` in the context of `wallet_type` etc). There also is an existing related TODO w.r.t refactor to `keystore_from_data`, which is another deduplication candidate.

PR #9298, opened 2024-11-12
wallet: RBF batch payments manager

ecdsa:
note: this is  not "reorg safe", meaning that there is no guarantee that all payments will end up in in the blockchain.
However, I believe this is "double send safe", meaning that we will never send a payment twice.


electrum/txengine.py
```diff
@@ -0,0 +1,321 @@
...
+        self.logger.info(f'locktime: {locktime}')
+        outputs += to_pay
+        inputs += self.get_change_inputs(self.batch_parent_tx) if self.batch_parent_tx else []
+        tx = self.wallet.create_transaction(
+            base_tx=base_tx,
+            inputs=inputs,
+            outputs=outputs,
+            password=password,
+            locktime=locktime,
+            BIP69_sort=False,
```
SomberNight, line 223: why disable bip69 sort?

tests/test_sswaps.py
```diff
@@ -1,77 +0,0 @@
...
-class TestSwapTxs(ElectrumTestCase):
-    TESTNET = True
-
-    def setUp(self):
-        super().setUp()
-        self.config = SimpleConfig({'electrum_path': self.electrum_path})
-        self.config.FEE_EST_DYNAMIC = False
-        self.config.FEE_EST_STATIC_FEERATE = 1000
-
-    def test_claim_tx_for_successful_reverse_swap(self):
```
SomberNight, line 18: Why delete these tests?
Just because `SwapManager._create_and_sign_claim_tx` is being removed?

The tests could be adapted to test `create_claim_txin` instead. Maybe feed the txin to `wallet.create_transaction` along with some hardcoded values.
The point is to test that the txin is constructed correctly, but the easiest way of doing that is comparing a full tx.

SomberNight (commented):
some preliminary comments

electrum/submarine_swaps.py
```diff
@@ -1077,46 +1037,32 @@ def add_txin_info(self, txin: PartialTxInput) -> None:
...
-            txin=txin,
-            witness_script=swap.redeem_script,
-            address=swap.receive_address,
-            amount_sat=amount_sat,
-            locktime=locktime,
-        )
-        cls.sign_tx(tx, swap)
-        return tx
+        self.add_txin_info(swap, txin)
+        txin.privkey = swap.privkey
```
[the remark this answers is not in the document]
ecdsa: see `sign_transaction` in `wallet.py`

electrum/txengine.py
ecdsa: when sweeping htlc outputs with sighash single-anyonecanpay, the order must be preserved.

tests/test_sswaps.py
ecdsa: > The point is to test that the txin is constructed correctly, but the easiest way of doing that is comparing a full tx.

Creating a full tx in order to check the construction of a tx input is indeed the easiest way, but it is completely overkill. 
I think we should try to put a little bit more efforts in writing tests.

electrum/txengine.py
```diff
@@ -0,0 +1,321 @@
...
+# When a batch is forgotten but not mined (because the server returned an error), we no longer bump its fee.
+# However, the current code does not theat the next batch as a CPFP when computing the fee.
+#
+# 2. Reorgs:
+# This code does not guarantee that a payment or a sweep will happen.
+# This is fine for sweeps; it is the responsibility of the caller (lnwatcher) to add them again.
+# To make payments reorg-safe, we would need to persist more data and redo failed payments.
+
+
+class TxEngine(Logger):
```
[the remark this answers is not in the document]
ecdsa: ok let's call it TxBatcher

tests/test_sswaps.py
ALANthemoor: RELEASE-NOTES

electrum/txengine.py
ecdsa: note that regtests would be failing if this was set to True

ALANthemoor: What happened 😔

tests/test_sswaps.py
ecdsa: sweeping from swaps is now tested in test_txbatcher.py

f321x:
I get this exception and electrum crashes when batching is enabled and trying to fully spend an unconfirmed utxo from the "Coins" tab with right click -> fully spend -> send to address in clipboard.
```
  File "/home/user/code/electrum-fork/electrum/gui/qt/confirm_tx_dialog.py", line 116, in timer_actions
    self.update()
    ~~~~~~~~~~~^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/confirm_tx_dialog.py", line 120, in update
    self.update_tx()
    ~~~~~~~~~~~~~~^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/confirm_tx_dialog.py", line 641, in update_tx
    self.tx = self.make_tx(fee_estimator, confirmed_only=confirmed_only)
              ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/send_tab.py", line 320, in <lambda>
    make_tx = lambda fee_est, *, confirmed_only=False: self.wallet.make_unsigned_transaction(
                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        coins=get_coins(nonlocal_only=nonlocal_only, confirmed_only=confirmed_only),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        outputs=outputs,
        ^^^^^^^^^^^^^^^^
        fee=fee_est,
        ^^^^^^^^^^^^
        is_sweep=is_sweep)
        ^^^^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/util.py", line 492, in do_profile
    o = func(*args, **kw_args)
  File "/home/user/code/electrum-fork/electrum/wallet.py", line 1879, in make_unsigned_transaction
    base_tx = self.get_unconfirmed_base_tx_for_batching(outputs, coins)
  File "/home/user/code/electrum-fork/electrum/wallet.py", line 1774, in get_unconfirmed_base_tx_for_batching
    output_amount = sum(o.value for o in outputs)
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

ecdsa:
>   File "/home/user/code/electrum-fork/electrum/wallet.py", line 1879, in make_unsigned_transaction
>     base_tx = self.get_unconfirmed_base_tx_for_batching(outputs, coins)

This does not correspond to the code in the `batch_payment_manager` branch.
`get_unconfirmed_base_tx_for_batching` is not used anymore (and should probably be removed)

electrum/txbatcher.py
```diff
@@ -0,0 +1,324 @@
...
+                self._base_tx = tx
+                self.logger.info(f'found base_tx {last_txid}')
+
+        self._parent_tx = None
+        self.batch_processing = set()  # list of inputs we are sweeping (until spending tx is confirmed)
+
+
+    def add_batch_payment(self, output: 'PartialTxOutput'):
+        # todo: maybe we should raise NotEnoughFunds here
+        self.batch_payments.append(output)
```
SomberNight, line 98: Is this class supposed to be thread-safe?
Calling this from GUI and asyncio thread at the "same time" might result in some hard-to-debug issues atm.

ecdsa: thanks, I added a lock

ecdsa:
Note: This PR does not work well with fee settings.

When we change the fee settings in `ConfirmTxDialog`, this side effects the fee settings in `config`, which is in turn used by `TxBatcher`. I think we should use a `FeeEstimator` object that holds fee settings and provides fee estimates, without committing the settings to config.

electrum/txbatcher.py
```diff
@@ -0,0 +1,354 @@
...
+        # this ensure that we can accept an input again
+        # if the spending tx has been removed from the blockchain
+        # fixme: what if there are several batches?
+        for txin in tx.inputs():
+            if txin.prevout in self._unconfirmed_sweeps:
+                self._unconfirmed_sweeps.remove(txin.prevout)
+
+    def _start_new_batch(self, tx):
+        use_change = tx and tx.has_change() and any([txout in self.batch_payments for txout in tx.outputs()])
+        self.batch_payments = self._to_pay_after(tx)
```
SomberNight, line 296: does this not need the lock?
```suggestion
        with self.lock:
            self.batch_payments = self._to_pay_after(tx)
```

electrum/txbatcher.py
```diff
@@ -0,0 +1,354 @@
...
+        self.batch_inputs = self._to_sweep_after(tx)
```
SomberNight, line 297: should we not take the lock everywhere `self.batch_inputs` is iterated on, mutated, or reassigned?
Well, maybe that's too aggressive but note that just because it is a dict, that does not suddenly solve all thread-safety.

Anyway, as-is, I think if another thread calls `add_sweep_info` while `_to_sweep_after` is being executed, we might lose the side-effect of `add_sweep_info`.

electrum/txbatcher.py
```diff
@@ -0,0 +1,354 @@
...
+        self.wallet = wallet
+        self.config = wallet.config
+        self.db = wallet.db
+        self.lock = threading.Lock()
+        self.batch_payments = []       # list of payments we need to make
+        self.batch_inputs = {}         # list of inputs we need to sweep
+
+        # list of tx that were broadcast. Each tx is a RBF replacement of the previous one. Ony one can get mined.
+        self.batch_txids = self.db.get_stored_item("batch_txids", [])
+        self._base_tx = None           # current batch tx. last element of batch_txids
```
SomberNight, line 90: TxBatcher needs to be taught the 100 KB max tx size standardness rule. It must not try creating txs that are too large.

electrum/lnwatcher.py
```diff
@@ -256,119 +266,24 @@ async def sweep_commitment_transaction(self, funding_outpoint, closing_tx) -> bo
...
-                tx_was_added = False
-            # set future tx regardless of tx_was_added, because it is not persisted
-            # (and wanted_height can change if input of CSV was not mined before)
-            self.adb.set_future_tx(new_tx.txid(), wanted_height=wanted_height)
-        if tx_was_added:
-            self.lnworker.wallet.set_label(new_tx.txid(), name)
-            if old_tx and old_tx.txid() != new_tx.txid():
-                self.lnworker.wallet.set_label(old_tx.txid(), None)
-            util.trigger_callback('wallet_updated', self.lnworker.wallet)
+        self.lnworker.wallet.txbatcher.add_sweep_info(sweep_info)
```
SomberNight, line 289: In general it is not safe to batch all lightning sweeps together, due to (1) pinning attacks and (2) split views of mempools due to distributed nature of network.

---

Perhaps the scariest problem is to do with revocations:
- counterparty force-closes using old state
- ctx has some htlcs
- counterparty creates a 2nd stage htlc tx ("tx1"), that spends one or more of the ctx's htlc outputs, and they add a lot of inputs/outputs to this tx (which they can do when using anchor channels due to sighash_single). They ensure tx1 is large (in vbytes), pays a low feerate so it does not confirm, but it pays a high absolute fee (as it is large).
- the local client then tries to punish the counterparty: we want to sweep `tx1:0` (the first output) but we also want to sweep `to_local`
- if we batch the sweeping of those two utxos, to create "tx2", as "tx1" might ~never confirm, "tx2" won't get mined either.
- and in the meantime, the relative timelock (to_self_delay_ on `to_local` is ticking. If that expires, the remote can just sweep it for themselves.

(see https://github.com/lightning/bolts/pull/803)

The easiest workaround I believe would be to avoid batching all revocations. That is, any sweep where we use the revocation path, should be a completely separate 1-in-1-out tx.

---

However, it's not just about revocations. In general, I believe it is not safe to batch the sweeping of txos where some txos are spendable by both us and the remote, and some only by us. Perhaps we should have a separate "safe" batch for sweeping txos that are only spendable by us, and another for the risky txos.

One complication is that some utxos transition from "only spendable by us" to "spendable by either party" over time (~already when a timelock gets close to expiry, but especially after it expires).

I am not yet sure exactly how to handle this.

---

Also, btw, to simplify reasoning, I think we should consider strictly separating the batching of user-initiated txs (e.g. simple sends), and lnwatcher/swap_manager stuff.

SomberNight (changes requested):

electrum/txbatcher.py
ecdsa: done

electrum/txbatcher.py
ecdsa: indeed. I made the locking more aggressive.

electrum/lnwatcher.py
SomberNight: > (2) split views of mempools due to distributed nature of network

Example:
- at block 0, remote counterparty force-closes small-value chan1, the to_self_delay is 1000
- we are offline
- at block 1000, remote force-closes high-value chan2, the to_self_delay is 1000
- at block 1100, we come online - and the remote somehow learns of this instantly
  - they try to partition the mempool
  - "at the same time":
    - remote broadcasts tx1, that sweeps the to_local output of their chan1
      - they try to pin this tx to the mempool: it is large in size, low feerate, but highish absolute fee
    - we broadcast tx2, that sweeps both to_locals (from chan1 and chan2) in a single batched tx
- if the remote is successful, the mempool is now partitioned so that most nodes and miners have tx1, but our electrum server has tx2
  - the attacker's goal here is that neither tx1 nor tx2 gets mined
- now the attacker waits.
- at block 2000, the CSV for "high-value" chan2 has expired, and the remote broadcasts a 1-in-1-out tx that spends just that to_local

ecdsa: ok, I believe this is addressed:
 - `TxBatcher` now uses separated TXBatche instances for swaps and lnwatcher.
 - `SweepInfo` has a `can_be_batched` boolean field, which is set only for inputs that only us can spend. If it is not set, a TxBatch instance is created for sweeping only that input.
 - Notably, second-stage htlc transactions are not batched. (we might want to batch them in some cases, but this is left for future work)

electrum/wallet.py
```diff
@@ -3327,6 +3337,32 @@ def get_frozen_balance_str(self) -> Optional[str]:
...
+            # (and wanted_height can change if input of CSV was not mined before)
+            self.adb.set_future_tx(txid, wanted_height=wanted_height)
+            return
+        name = sweep_info.name
+        tx = self.create_transaction(
+            inputs=[txin],
+            outputs=[],
+            password=None,
+            fee_policy=FixedFeePolicy(0),
+        )
```
SomberNight, line 3356: no outputs and zero fee? what is this supposed to do?
and in any case, maybe we should make make_unsigned_transaction enforce there is at least one output (that's a consensus rule btw)

ecdsa: Given no outputs, `make_unsigned_transaction` will send all the coins to a change address.
Zero fee: this is a future transaction, its purpose is only to provide feedback to the user. 
A fee will be added when the actual tx is broadcast.

SomberNight: > Given no outputs, make_unsigned_transaction will send all the coins to a change address.

Ah right... ok, makes sense. Could you add a comment "this will send all the coins to a change address"?

Note that `create_transaction` tries to sign this by default. Is that intentional in this case? (with zero fee?)
(indeed we should just kill `wallet.create_transaction`)

ecdsa: indeed, it is better not to sign. I passed `sign=False`
(the future tx used to have a fee, and to be signed so that the user can broadcast early)

ALANthemoor: Screenshot_20250206-123829

PR #9304, opened 2024-11-15
submarine_swaps: implement swaps-over-nostr for qml

accumulator:


FELLIX000 (approved):

accumulator:
depends on [newer electrum-aionostr](https://github.com/spesmilo/aionostr/pull/3)

SomberNight:
> depends on [newer electrum-aionostr](https://github.com/spesmilo/aionostr/pull/3)

done in https://github.com/spesmilo/electrum/commit/40842fad399e694539041de3e1fb78896e07878e

ecdsa:
thanks for rebasing. 
Note: there is an issue with the nostr event definition of master.
@f321x is going to change the event number and subtype, so that events can properly expire.

ecdsa:
note: there seems to be a bug in the GUI: 
if I disable nostr and set swapserver_url, the slider is disabled for normal swaps, only reverse swaps are allowed.

accumulator:
> note: there seems to be a bug in the GUI: if I disable nostr and set swapserver_url, the slider is disabled for normal swaps, only reverse swaps are allowed.

Can't immediately reproduce.. Do you have enough non-frozen receiving capacity?

FELLIX000 (approved):

ecdsa:
fixed and merged manually

ecdsa:
> Can't immediately reproduce.. Do you have enough non-frozen receiving capacity?

I would not post that comment if there was no difference between Qt and QML.

The problem was caused by an incorrect computation of the max amount in QML, and was not related to this PR.
Fixed in 5600e79eddcad1c646993ec460776f8a274440a2

PR #9321, opened 2024-11-24
qml: add config setting for max brightness on qr display

jgeden:
In #9170 a feature was added that enables max screen brightness when displaying QR codes. I don't usually use the QR code and prefer to just copy the address from this screen. This puts the max brightness feature behind a config setting so I can stop accidentally blinding myself every time I open the request screen lol.

### DIFF
--- a/electrum/gui/qml/qeapp.py
+++ b/electrum/gui/qml/qeapp.py
@@ -211,6 +211,10 @@ class QEAppController(BaseCrashReporter, QObject):
         jpythonActivity.startActivity(it)
 
     @pyqtSlot()
+    def isMaxBrightnessOnQrDisplayEnabled(self):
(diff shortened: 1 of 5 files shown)


accumulator:
Moved setting to 'User Interface' section, merged in 185f1976f3451b4d67ab82c75b23bf390eec0143

Thanks!

electrum/gui/qml/qeapp.py
```diff
@@ -210,6 +210,10 @@ def doShare(self, data, title):
         it = jIntent.createChooser(sendIntent, cast('java.lang.CharSequence', jString(title)))
         jpythonActivity.startActivity(it)
 
+    @pyqtSlot()
```
SomberNight, line 213: missing result type for the decorator, resulting in setMaxScreenBrightness() never actually getting called.

fixed in https://github.com/spesmilo/electrum/commit/7fa204f16b08a3c58ec403d9e07661531b91b773

PR #9358, opened 2024-12-12
Change thousands separator to comma

cowsaymoe:


electrum/util.py
```diff
@@ -767,7 +767,7 @@ def format_satoshis_plain(
 
 # DECIMAL_POINT = locale.localeconv()['decimal_point']  # type: str
 DECIMAL_POINT = "."
-THOUSANDS_SEP = " "
+THOUSANDS_SEP = ","
```
keharv, line 770: Thank you for making this change. This has been bothering me for awhile.
Can this get merged ASAP?

keharv (approved):

SomberNight:
related https://github.com/spesmilo/electrum/pull/8139

It was a deliberate choice not to use "." and "," for dp and ts, as some countries use "." and "," and others do the exact opposite with "," and ".".
I think it is a bad idea to hardcode any one of those.
At most, we could do what the system locale suggests. Or keep the status quo.

Btw, you should provide some justification for this kind of change and take the time to find, read, and link back to prior discussion.

PR #9369, opened 2024-12-17
adding timestamp on coins tab for easier utxo time sorting

wiederdev:
Choose utxo according to its age is now easier with this patch that adds ability to sort also with the utxo creation time.

### DIFF
--- a/electrum/gui/qt/utxo_list.py
+++ b/electrum/gui/qt/utxo_list.py
@@ -35,7 +35,7 @@ from electrum.i18n import _
 from electrum.bitcoin import is_address
 from electrum.transaction import PartialTxInput, PartialTxOutput
 from electrum.lnutil import MIN_FUNDING_SAT
-from electrum.util import profiler
+from electrum.util import profiler, format_time
 
 from .util import ColorScheme, MONOSPACE_FONT, EnterButton
 from .my_treeview import MyTreeView, MySortModel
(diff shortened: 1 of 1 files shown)


wiederdev:
Does anyone understand what the regtest failure is all about? The patch is pretty simple, and I don't see how it is related to the failure.

ecdsa:
Thanks for the PR. the regtest failure does not come from your commit, it also exists on master. I am investigating it.

ecdsa:
on my system, the window seems to need a manual resize in order to display the new column properly. not sure why..

SomberNight:
> adds ability to sort also with the utxo creation time

you can already achieve the same sort order using the outpoint column though?

I don't really see the value in adding a date column.

If we made the columns configurable -- so that right clicking in the header would open a context menu allowing hiding/showing columns, then it would be easier to make the case to add more utility columns such as this. Without that, each column takes up space for everyone.

wiederdev:
> I don't really see the value in adding a date column.

I live in a county which taxes the capital gains differently according to the date of purchase, and this way I could easily see the date of the utxo and know if it is relevant for selling.
 
> If we made the columns configurable -- so that right clicking in the header would open a context menu allowing hiding/showing columns, then it would be easier to make the case to add more utility columns such as this. Without that, each column takes up space for everyone.

This would be perfect.
I am not sure how to do that, but on the history tab, there is a similar solution using a box on the top right where one can choose show/hide for some columns.

PR #9397, opened 2025-01-02
Lossless image asset compression

Awqre:
Optimization of PNG, JPEG, ICO and GIF files

Combination of different optimizers (ECT, PNGOut, DeflOpt...)


ecdsa:
please provide the exact recipes to do that, and we will try to reproduce them.

not merging this because of trust issues.

PR #9412, opened 2025-01-09
Add filehash of external plugins to PluginDialog 

f321x:
This PR adds the sha256 hash of the zipfile of external plugins to the PluginDialog so users can see it when enabling a plugin. 

Screenshot From 2025-01-09 18-07-20

The hash is calculated when searching for plugins and split with a whitespace every 16 chars to allow for line wrapping and easier readability.


SomberNight (approved):
Looks good. Thanks!

PR #9417, opened 2025-01-10
fix: add extra check to script_GetOp

thecockatiel:
Hi
This check is done by bitcoinj when parsing chunks and there was a test case for it in that lib, so you can consider this a thought

test cases from bitcoinj for context:
```
  ["0x4c01","0x01 NOP", "P2SH,STRICTENC","BAD_OPCODE", "PUSHDATA1 with not enough bytes"],
  ["0x4d0200ff","0x01 NOP", "P2SH,STRICTENC","BAD_OPCODE", "PUSHDATA2 with not enough bytes"],
  ["0x4e03000000ffff","0x01 NOP", "P2SH,STRICTENC","BAD_OPCODE", "PUSHDATA4 with not enough bytes"],
```

feel free to close this if you think this is unnecessary, I would certainly be happy to know your opinion on this

### DIFF
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -620,7 +620,7 @@ def script_GetOp(_bytes : bytes):
                 i += 4
                             
             if nSize > _bytes_len - i:
-                raise MalformedBitcoinScript(f"Push of data element that is larger than remaining data: {nSize} vs {_bytes_len - i}")    
+                raise MalformedBitcoinScript(f"Push of data element that is larger than remaining data: {nSize} vs {_bytes_len - i}")
             
             vch = _bytes[i:i + nSize]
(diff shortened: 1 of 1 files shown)


SomberNight:
Indeed. This seems to be a missing check. Thank you!

I've [squashed this](https://github.com/spesmilo/electrum/commit/ddb67d9bca8289cbfa1c903d7b38a6115c984ae3) and added a [follow-up commit](https://github.com/spesmilo/electrum/commit/043be2439ef84fd7fb52b300e5ba95981f326228) and merged this manually in https://github.com/spesmilo/electrum/commit/ce843fd4546fb66ff353e9a64059cca495dc23af

PR #9427, opened 2025-01-14
Limit payment attempts if client uses trampoline

f321x:
This PR sets the payment retry limit to 20 if the client is using trampoline to prevent excessive retries until the timeout.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1487,6 +1487,9 @@ class LNWallet(LNWorker):
         self.wallet.set_label(key, lnaddr.get_description())
         self.set_invoice_status(key, PR_INFLIGHT)
         budget = PaymentFeeBudget.default(invoice_amount_msat=amount_to_pay, config=self.config)
+        if attempts is None and self.uses_trampoline():
+            # we don't expect lots of failed htlcs with trampoline, so we can fail sooner
+            attempts = 30
         success = False
         try:
             await self.pay_to_node(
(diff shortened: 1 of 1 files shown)


electrum/lnworker.py
```diff
@@ -1497,7 +1497,7 @@ async def pay_invoice(
                 min_final_cltv_delta=min_final_cltv_delta,
                 r_tags=r_tags,
                 invoice_features=invoice_features,
-                attempts=attempts,
+                attempts=20 if attempts is None and self.uses_trampoline() else attempts,
```
SomberNight, line 1500: Note: "attempts" is counting the number of failed htlcs, so this number should not be set to too low.
However in the case of trampoline, the client does not split the payment too agressively. (It splits it more when using gossip. In the case of trampoline it is assumed the trampoline forwarder node will split it for us in a smarter way.)

Still, note that it is common (almost expected) for ~3-4 htlcs to fail, due to trampoline_fee_insufficient. Just to be safe I would suggest using a bit higher value, let's say 30.

Also, (nit) I think this would be more readable not inlined, and then we can also add a comment, e.g.:
```diff
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index ecff33e3f3..8d37ba8376 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1493,6 +1493,9 @@ class LNWallet(LNWorker):
         self.wallet.set_label(key, lnaddr.get_description())
         self.set_invoice_status(key, PR_INFLIGHT)
         budget = PaymentFeeBudget.default(invoice_amount_msat=amount_to_pay, config=self.config)
+        if attempts is None and self.uses_trampoline():
+            # we don't expect lots of failed htlcs with trampoline, so we can fail sooner
+            attempts = 30
         success = False
         try:
             await self.pay_to_node(

```

f321x: thx for the review, changed it according to your comment.

PR #9430, opened 2025-01-14
lightning: change derivation of funding_pubkey

SomberNight:
Ideally, given an on-chain backup, after the remote force-closes, we should be able to spend our anchor output, to CPFP the remote commitment tx (assuming the channel used OPTION_ANCHORS). To spend the anchor output, we need to be able to sign with the local funding_privkey.

Previously we derived the funding_key from the channel_seed (which comes from os.urandom). Prior to anchors, there was no use case for signing with the funding_key given a channel backup. Now with anchors, we should make its derivation deterministic somehow, in a way so that it can be derived given just an on-chain backup.
- one way would be to put some more data into the existing OP_RETURN
  - uses block space
  - the OP_RETURNs can be disabled via "use_recoverable_channels"
  - only the initiator can use OP_RETURNs (so what if channel is in incoming dir?)
- instead, new scheme for our funding_key:
  - we derive the funding_privkey from the lnworker root secret (derived from our bip32 seed)
  - for outgoing channels:
    - lnworker_root_secret + remote_node_id + funding_tx_nlocktime
  - for incoming channels:
    - lnworker_root_secret + remote_node_id + remote_funding_pubkey
  - a check is added to avoid reusing the same key between channels: not letting to user open more than one channel with the same peer in a single block
  - only the first 16 bytes of the remote_node_id are used, as the onchain backup OP_RETURNs only contain that
  - note: we could even derive channel_seed deterministically this way
	- that would make avoiding reuse really security-critical
		(e.g. rev_acks across different chans reusing ~same revocation secrets)
	- for the funding_key in isolation, a potential key reuse is only a privacy risk.
- as the funding_privkey cannot be derived from the channel_seed anymore, it is included in the imported channel backups, which in turn need a new version defined
  - a **wallet db upgrade** is used to update already stored imported cbs
  - alternatively we could keep the imported cbs as-is, so no new version, no new funding_privkey field, as it is clearly somewhat redundant given on-chain backups can reconstruct it
    - however adding the field seems easier
      - otherwise the existing code would try to derive the funding_privkey from the channel_seed
      - also note: atm there is no field in the imported backups to distinguish anchor channels vs static-remotekey channels

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -415,6 +415,10 @@ class AbstractChannel(Logger, ABC):
     def get_funding_address(self) -> str:
         pass
 
+    def get_funding_tx(self) -> Optional[Transaction]:
+        funding_txid = self.funding_outpoint.txid
+        return self.lnworker.lnwatcher.adb.get_transaction(funding_txid)
+
     @abstractmethod
     def get_sweep_address(self) -> str:
         """Returns a wallet address we can use to sweep coins to.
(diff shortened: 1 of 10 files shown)


SomberNight:
(can be merged after review)

ecdsa:
There is a bit of duplication between `derive_multisig_funding_key_if_we_opened` and `derive_multisig_funding_key_if_they_opened`. 
Would it make sense to use a single function `derive_multisig_funding_key`, that accepts both parameters `locktime` and `remote_funding_pubkey`, and asserts that exactly one of them is not None?

PR #9432, opened 2025-01-15
Increase attempts when forwarding trampoline payment

f321x:
Currently forwarding trampoline payments is only done in one attempt. The attempts have to be increased to make forwarding more reliable.


ecdsa:
Note: it might make sense to adapt the max number of htlc attempts to the amount we need to pay, but for now I think we can live with a constant.

PR #9433, opened 2025-01-15
Change OnionFailureCode returned when failing trampoline forwarding

f321x:
This changes the OnionFailureCode returned from ```maybe_forward_trampoline()``` on ```PaymentFailure``` to ```TRAMPOLINE_FEE_INSUFFICIENT``` so the client will retry the payment with higher trampoline fee instead of giving up.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2016,7 +2016,7 @@ class Peer(Logger, EventListener):
             self.logger.debug(
                 f"maybe_forward_trampoline. PaymentFailure for {payment_hash.hex()=}, {payment_secret.hex()=}: {e!r}")
             # FIXME: adapt the error code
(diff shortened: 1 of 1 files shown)


ecdsa:
Note: This reverts eda9097e894b4e94b17160701cf242ad8da76635. That commit is related to a discussion in https://github.com/lightning/bolts/pull/829

The FIXME indicates that there are some cases where it would make sense to return unknown_next_peer.
Specifically:
 - we should return TRAMPOLINE_FEE_INSUFFICIENT if we could not find a route because of too low fees
 - we should return UNKNOWN_NEXT_PEER if we cannot find the next trampoline in the graph (per https://github.com/lightning/bolts/pull/829#issuecomment-756868196 )

PR #9435, opened 2025-01-15
Add fallback list of signet nodes and use it for ln peer discovery

f321x:
This PR adds a fallback list of signet lightning nodes and uses it in ```lnworker._get_next_peers_to_try()``` in case there are no previous peers in the ChannelDB. 
As peers only get added to the ChannelDB on connection if trampoline is already disabled this will help to directly find peers when disabling trampoline payments (enabling gossip) as the ChannelDB doesn't contain any previous peers yet (they would be added on restart when reconnecting with loaded ChannelDB).

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -152,6 +152,20 @@ FALLBACK_NODE_LIST_MAINNET = [
     LNPeerAddr(host='2001:41d0:1:b40d::1', port=9735, pubkey=bfh('026726a4b043d413b45b334876d17b8a98848129604429ec65532ba286a42efeac')),
 ]
 
+FALLBACK_NODE_LIST_SIGNET = (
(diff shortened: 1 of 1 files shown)


ecdsa:
note: it would be nice to move these lists out of `lnworker.py` into a json file

SomberNight:
> note: it would be nice to move these lists out of lnworker.py into a json file

Or alternatively refactor (and rename?) constants.py and put this there.
The list of blockchain-specific block explorers could also go there (e.g. the `util.mainnet_block_explorers` stuff)

Look at e.g. https://github.com/spesmilo/electrum/pull/9197 for reference.
The LN fallback peers and the block explorers could be made fields of AbstractNet (or similar).

Or we could do both: move into json file, and reference json file from AbstractNet.

PR #9437, opened 2025-01-15
Reload existing peers into ChannelDB when enabling gossip

f321x:
Currently when enabling gossip (disabling trampoline) in the qt settings dialog the ChannelDB will be created and doesn't contain peers. Peers get added to the ChannelDB by lnworker on connection, but not when enabling gossip. So the newly created gossip has to access fallback nodes because the ChannelDB is empty even though we already know about our channel peers. 
This PR will trigger a reload of the existing peers into the ChannelDB when disabling trampoline so we don't solely rely on the fallback nodes.

### DIFF
--- a/electrum/gui/qt/settings_dialog.py
+++ b/electrum/gui/qt/settings_dialog.py
@@ -127,6 +127,9 @@ class SettingsDialog(QDialog, QtEventListener):
             self.config.LIGHTNING_USE_GOSSIP = not use_trampoline
             if not use_trampoline:
                 self.network.start_gossip()
+                if self.wallet.lnworker is not None:
+                    # add existing peers to ChannelDB
+                    self.wallet.lnworker.reload_peers_into_channeldb()
             else:
                 self.network.run_from_another_thread(
                     self.network.stop_gossip())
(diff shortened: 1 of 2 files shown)


electrum/gui/qt/settings_dialog.py
```diff
@@ -127,6 +127,9 @@ def on_trampoline_checked(_x):
             self.config.LIGHTNING_USE_GOSSIP = not use_trampoline
             if not use_trampoline:
                 self.network.start_gossip()
+                if self.wallet.lnworker is not None:
+                    # add existing peers to ChannelDB
+                    self.wallet.lnworker.reload_peers_into_channeldb()
```
SomberNight, line 132: What about e.g. the qml GUI? Maybe this should be done inside the `network.start_gossip` function instead.

Also, there might be multiple wallets open in the GUI.
`self.wallet` is used for wallet-specific settings in this dialog, but note that some settings are global and apply to all wallets (such as this one, toggling the gossip).

In principle you could do something like this:
```py
        for wallet in network.daemon.get_wallets().values():
            if wallet.lnworker is not None:
                wallet.lnworker.reload_peers_into_channeldb()
```

f321x: I see it makes sense to remove it from the GUI, i think it also would also be useful to first check if there are already peers available (otherwise the reload would be unnecessary) as start_gossip will also get called from other points than the GUI.
In the following commit added a function to the gossip taskgroup which then also can first await readiness of the ChannelDB too.

PR #9458, opened 2025-01-20
Bump allowed aiorpcx version to 0.24

f321x:
Increase the allowed version of [aiorpcx](https://github.com/kyuupichan/aiorpcX) to the new 0.24 version, which uses the new websockets version, so it's possible to get rid of the websockets deprecation warning when starting electrum.
Changelog of aiorpcX 0.24: [Changelog](https://github.com/kyuupichan/aiorpcX/blob/master/docs/changelog.rst#version-024-16-jan-2024)


SomberNight:
Thanks :)

PR #9465, opened 2025-01-21
Use getlocale() instead of getdefaultlocale() in crash reporter

f321x:
Changes usage of the depreciated [```locale.getdefaultlocale()```](https://docs.python.org/3/library/locale.html#locale.getdefaultlocale) to the new [```locale.getlocale()```](https://docs.python.org/3/library/locale.html#locale.getlocale) to fix the following depreciation warning:
```
../electrum/base_crash_reporter.py:127: DeprecationWarning: 'locale.getdefaultlocale' is deprecated and slated for removal in Python 3.15. Use setlocale(), getencoding() and getlocale() instead.
  "locale": locale.getdefaultlocale()[0] or "?",
```

### DIFF
--- a/electrum/base_crash_reporter.py
+++ b/electrum/base_crash_reporter.py
@@ -124,7 +124,7 @@ class BaseCrashReporter(Logger):
             "python_version": sys.version,
             "os": describe_os_version(),
             "wallet_type": "unknown",
-            "locale": locale.getdefaultlocale()[0] or "?",
+            "locale": locale.getlocale()[0] or "?",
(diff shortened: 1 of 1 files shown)


SomberNight:
See e.g. https://github.com/python/cpython/issues/82986 and https://github.com/pytest-dev/pytest-nunit/issues/67

SomberNight:
Well, okay, this locale API looks like a mess... but I guess we don't need formatting and can just do a breaking change here, as these values are only there to be human-readable in the report.

PR #9466, opened 2025-01-21
Separate between fee related NoPathFound and other causes when doing lightning payments/forwards

f321x:
Related to https://github.com/spesmilo/electrum/pull/9433#issuecomment-2592754899. Adds a bool flag to ```NoPathFound``` exception that can be used to determine if the exception could be caused by a too low fee or if it has another cause when catching it. This information is then be used to send either ```OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT``` or ```OnionFailureCode.UNKNOWN_NEXT_PEER``` when failing to forward a trampoline payment, so we don't always return ```TRAMPOLINE_FEE_INSUFFICIENT``` even if maybe there is no channel available at all (so raising fees would be useless). 
The flag could also be used client side, for example to show the user that a payment maybe failed because of a too low fee instead of just showing that it failed without reason as we use the ```NoPathFound``` exception too.
However, maybe there is a more elegant version than a bool flag, like an additional exception (```NoPathTooLowFee```)?


ecdsa:
I think it would be more elegant to raise another exception, rather than adding this flag

PR #9470, opened 2025-01-22
Remove query_short_channel_ids zlib compression

f321x:
We send the query_short_channel_ids gossip message with zlib compression encoding enabled by default. However this was removed/forbidden in the bolt spec in this PR https://github.com/lightning/bolts/pull/981 and must not be done anymore, see https://github.com/lightning/bolts/blob/master/07-routing-gossip.md#query-messages. I removed the compression part completely as i see no case where it could be used anymore.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -605,16 +605,17 @@ class Peer(Logger, EventListener):
         await self.querying.wait()
         self.querying.clear()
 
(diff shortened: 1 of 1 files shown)


accumulator:
Yep, I ran into the same issue when testing onion messages against c-lightning and had a similar patch parked in a local branch for testing. Good to see this is indeed part of the spec.

PR #9522, opened 2025-02-04
Update mainnet ln fallback nodes

f321x:
The mainnet ln fallback nodes were so outdated that it was not possible to bootstrap from them anymore. These are all large nodes from the mempool.space liquidity ranking.


SomberNight:
thx

PR #9523, opened 2025-02-04
qt: show warning on expired invoice on send tab.

accumulator:
Also disables Save button when invoice has expired.

### DIFF
--- a/electrum/gui/qt/send_tab.py
+++ b/electrum/gui/qt/send_tab.py
@@ -25,7 +25,7 @@ from electrum.submarine_swaps import SwapServerError
 from .amountedit import AmountEdit, BTCAmountEdit, SizedFreezableLineEdit
 from .paytoedit import InvalidPaymentIdentifier
 from .util import (WaitingDialog, HelpLabel, MessageBoxMixin, EnterButton, char_width_in_lineedit,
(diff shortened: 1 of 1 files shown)


electrum/gui/qt/send_tab.py
```diff
@@ -464,9 +470,14 @@ def update_fields(self):
...
         self.send_button.setEnabled(not pi_unusable and amount_valid and not pi.has_expired())
-        self.save_button.setEnabled(not pi_unusable and not is_spk_script and \
+        self.save_button.setEnabled(not pi_unusable and not is_spk_script and not pi.has_expired() and \
                                     pi.type not in [PaymentIdentifierType.LNURLP, PaymentIdentifierType.LNADDR])
 
+        if pi.has_expired():
+            self.invoice_error.setText(_('Expired'))
+        else:
+            self.invoice_error.setText('')
+
```
ecdsa, line 480: please avoid code duplication
```python
msg = _('Expired') if pi.has_expired() else ''
self.invoice_error.setText(msg)
```

SomberNight: this is short enough to be highly subjective and down to personal style IMO :)

PR #9534, opened 2025-02-06
fix Coldcard msg signing for segwit v0 address types

scgbckbone:
* this is bug where electrum showed correct address (i.e `p2wpkh` or `p2sh-p2wpkh`) but Coldcard was always showing `p2pkh` address. Msg verifies OK (sig correct) as all the address types for specific path have same public key, but address verification on device is not possible


accumulator:
thanks!

PR #9551, opened 2025-02-11
Change nostr swap event type, add PoW, rebroadcast events

f321x:
* Changes the event type for nostr swap announcements to 30315 (NIP38 user status), we now also filter directly for the chain hash and add an expiry tag.
Example event:
```json
{
  "id": "49ca58b57fda5dc384a6b2f3d1dcf83ec33160670e439ccd95701135d06f171e",
  "pubkey": "c70d7bc9de7e98280c039e6b741c075b66f3f56e4798e3d9c3b4c93dc0511f27",
  "created_at": 1739351024,
  "kind": 30315,
  "tags": [
    [
      "d",
      "electrum-swapserver-2"
    ],
    [
      "r",
      "net:signet"
    ],
    [
      "expiration",
      "1739351634"
    ]
  ],
  "content": "{\"percentage_fee\": 0.5, \"normal_mining_fee\": 952, \"reverse_mining_fee\": 1071, \"claim_mining_fee\": 952, \"min_amount\": 20000, \"max_amount\": 10000000, \"relays\": \"wss://nos.lol,wss://relay.damus.io,wss://brb.io,wss://nostr.mom,wss://relay.primal.net,wss://ftp.halifax.rwth-aachen.de/nostr,wss://eu.purplerelay.com,wss://nostr.einundzwanzig.space\", \"pow_nonce\": \"0x6b5ad6b5ad6b5ad6b5ad6b5ad6b5ad6b5ad6b5ad6b5ad6b5ad6b5ad6ba0c3f10\"}",
  "sig": "6a1a58d48be40aec49026b925e0daf431eeb51eba496013b0993bc85bc8c3a9c973167a29d41a4350d2107c03e5533b6d3b078a8f799c9f018818f5b355125dc"
}
```

* Rebroadcast events to relays that we use but are not announced in the event of the swapserver.

* Add more default relays.

* Add Proof of Work to swapserver events, and filter for the PoW on client side.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import os
+import ssl
 from typing import TYPE_CHECKING, Optional, Dict, Union, Sequence, Tuple, Iterable
 from decimal import Decimal
 import math
@@ -13,6 +14,7 @@ import electrum_ecc as ecc
 from electrum_ecc import ECPrivkey
 
 import electrum_aionostr as aionostr
+from electrum_aionostr.event import Event
 from electrum_aionostr.util import to_nip19
 
 from collections import defaultdict
@@ -24,7 +26,8 @@ from .bitcoin import (script_to_p2wsh, opcodes,
                       construct_witness)
 from .transaction import PartialTxInput, PartialTxOutput, PartialTransaction, Transaction, TxInput, TxOutpoint
 from .transaction import script_GetOp, match_script_against_template, OPPushDataGeneric, OPPushDataPubkey
-from .util import log_exceptions, ignore_exceptions, BelowDustLimit, OldTaskGroup, age
+from .util import (log_exceptions, ignore_exceptions, BelowDustLimit, OldTaskGroup, age, ca_path,
+                   gen_nostr_ann_pow, get_nostr_ann_pow_amount)
 from .lnutil import REDEEM_AFTER_DOUBLE_SPENT_DELAY
 from .bitcoin import dust_threshold, DummyAddress
 from .logging import Logger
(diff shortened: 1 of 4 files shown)


electrum/util.py
```diff
@@ -2116,3 +2118,69 @@ def truncate_text(text: str, *, max_len: Optional[int]) -> str:
...
+
+def get_nostr_ann_pow_amount(nostr_pubk: bytes, nonce: Optional[int]) -> int:
+    """Return the amount of leading zero bits for a nostr announcement PoW."""
+    if not nonce:
+        return 0
+    hash_function = hashlib.sha256
+    hash_len_bits = 256
+    hash_preimage = b'electrum-' + nostr_pubk
+
+    digest = hash_function(hash_preimage + nonce.to_bytes(8, 'big')).digest()
```
SomberNight, line 2184: If we use sha256 for the pow function, a 64-bit nonce is not enough :D

Mainnet already burns through around 80 bits, every 10 minutes.
https://blockstream.info/block/000000000000000000003ac0a58830f247c3773a2958bdb3d924ba7ab94b99ec

We don't expect that kind of hashpower being used for this, but still, when designing something new, let's at least try to learn from past mistakes: the nonce field in the Bitcoin block header is only 4 bytes, and because of that, miners have to do all kinds of ugly hacks to gain extra nonce space.

Let's just allocate 32 bytes for the nonce here: AFAICT it does not cost anything for us to do that.

electrum/submarine_swaps.py
```diff
@@ -1466,6 +1518,21 @@ async def get_pairs(self):
             self.sm.update_pairs(pairs)
             self.server_relays = content['relays'].split(',')
 
+    async def rebroadcast_event(self, event: Event, server_relays: Sequence[str]):
+        """If the relays of the origin server are different from our relays we rebroadcast the
+        event to our relays so it gets spread more widely."""
```
ecdsa, line 1523: note: nothing prevent us from rebroadcasting an event to a relay that already sent us that event.

PR #9557, opened 2025-02-13
Verify nostr events in submarine_swaps

f321x:
Currently we don't call the verify method on nostr events in the submarine swap client, so we would accept forged events that are not actually created by the pubkey they reference. 
This PR adds .verify() calls on the event. 
**Note**, this will only work when using aionostr with this fix https://github.com/spesmilo/electrum-aionostr/pull/4 as the verify method in the current version (0.0.7) of aionostr is broken.


f321x:
All events will be verified directly in aionostr

PR #9560, opened 2025-02-13
Pass proxy settings to aionostr manager for submarine swaps

f321x:
This PR relies on https://github.com/spesmilo/electrum-aionostr/pull/5
Passes the electrum proxy configuration to the aionostr library so the websocket connections go through the proxy.
Can be tested by configuring tor as proxy and using these onion nostr relays in the config:
`ws://oxtrdevav64z64yb7x6rjg4ntzqjhedm5b5zjqulugknhzr46ny2qbad.onion,ws://bitcoinr6de5lkvx4tpwdmzrdfdpla5sya2afwpcabjup2xpi5dulbad.onion,ws://nostrland2gdw7g3y77ctftovvil76vquipymo7tsctlxpiwknevzfid.onion`

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -27,7 +27,8 @@ from .bitcoin import (script_to_p2wsh, opcodes,
 from .transaction import PartialTxInput, PartialTxOutput, PartialTransaction, Transaction, TxInput, TxOutpoint
 from .transaction import script_GetOp, match_script_against_template, OPPushDataGeneric, OPPushDataPubkey
 from .util import (log_exceptions, ignore_exceptions, BelowDustLimit, OldTaskGroup, age, ca_path,
-                   gen_nostr_ann_pow, get_nostr_ann_pow_amount)
+                   gen_nostr_ann_pow, get_nostr_ann_pow_amount, make_aiohttp_proxy_connector, get_running_loop,
+                   get_asyncio_loop)
 from .lnutil import REDEEM_AFTER_DOUBLE_SPENT_DELAY
 from .bitcoin import dust_threshold, DummyAddress
 from .logging import Logger
@@ -53,6 +54,7 @@ if TYPE_CHECKING:
     from .lnworker import LNWallet
     from .lnchannel import Channel
     from .simple_config import SimpleConfig
+    from aiohttp_socks import ProxyConnector
 
 
 
@@ -1322,8 +1324,8 @@ class NostrTransport(SwapServerTransport):
         self.nostr_private_key = to_nip19('nsec', keypair.privkey.hex())
         self.nostr_pubkey = keypair.pubkey.hex()[2:]
         self.dm_replies = defaultdict(asyncio.Future)  # type: Dict[bytes, asyncio.Future]
-        ssl_context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=ca_path)
-        self.relay_manager = aionostr.Manager(self.relays, private_key=self.nostr_private_key, log=self.logger, ssl_context=ssl_context)
+        self.ssl_context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=ca_path)
+        self.relay_manager = None
         self.taskgroup = OldTaskGroup()
         self.server_relays = None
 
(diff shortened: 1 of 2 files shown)


electrum/network.py
```diff
@@ -1498,3 +1499,31 @@ async def prune_offline_servers(self, hostmap):
...
+        password = self.proxy.get('password', "")
+
+        return ProxyConnector(
+            proxy_type=mode,
+            host=host,
+            port=port,
+            username=user,
+            password=password,
+            rdns=True
+        )
```
SomberNight, line 1529: Most of this code already exists in `util.make_aiohttp_session`.
It would be good to refactor to avoid duplication.

electrum/network.py
```diff
@@ -1498,3 +1499,31 @@ async def prune_offline_servers(self, hostmap):
...
+        user = self.proxy.get('user', "")
+        password = self.proxy.get('password', "")
+
+        return ProxyConnector(
+            proxy_type=mode,
+            host=host,
+            port=port,
+            username=user,
+            password=password,
+            rdns=True
```
SomberNight, line 1528: note: we should also set `ssl=ssl_context` (as done in `util.make_aiohttp_session`)

electrum/network.py
```diff
@@ -1498,3 +1499,31 @@ async def prune_offline_servers(self, hostmap):
         servers_dict = {k: v for k, v in hostmap.items()
                         if k in servers_replied}
         return servers_dict
+
+    def get_aiohttp_proxy_connector(self) -> Optional[ProxyConnector]:
```
SomberNight, line 1503: note the comment on line 1459:
```
# methods used in scripts
```

although it is not exactly "visible" :)

The other methods below that line are only used from `electrum/scripts/`.

electrum/submarine_swaps.py
```diff
@@ -1335,7 +1336,13 @@ def __init__(self, config, sm, keypair):
         self.nostr_pubkey = keypair.pubkey.hex()[2:]
         self.dm_replies = defaultdict(asyncio.Future)  # type: Dict[bytes, asyncio.Future]
         ssl_context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=ca_path)
-        self.relay_manager = aionostr.Manager(self.relays, private_key=self.nostr_private_key, log=self.logger, ssl_context=ssl_context)
+        self.proxy: Optional[ProxyConnector] = sm.network.get_aiohttp_proxy_connector()
```
SomberNight, line 1339: This only works by accident. On line 55, you only import `ProxyConnector` inside `if TYPE_CHECKING`.
At runtime, those branches are never entered. As in, `TYPE_CHECKING` is always False.

You should either move the import outside, to the top level; or use quotes:
```
self.proxy: Optional['ProxyConnector'] = sm.network.get_aiohttp_proxy_connector()
```

I think the only reason this does not crash at runtime currently, is because the module gets already imported by other modules by the time code execution reaches this line.

Try this in a fresh local interpreter:
```py
>>> from typing import TYPE_CHECKING
>>> if TYPE_CHECKING:
...   from aiohttp_socks import ProxyConnector
...
>>> t: ProxyConnector = 2
  File "<stdin>", line 1, in <module>
NameError: name 'ProxyConnector' is not defined
```

electrum/network.py
f321x: Ok, removed `network.get_aiohttp_proxy_connector` and made use of the code in `util.make_aiohttp_session` in https://github.com/spesmilo/electrum/pull/9560/commits/3144ebdc537fe6e0cd4ee26dbfaba88fe5130c35

electrum/submarine_swaps.py
f321x: oh ok, so i put it in quotes now. in https://github.com/spesmilo/electrum/pull/9560/commits/05da97caa4abd73e7f9eb6087f91a7bf8f7e22bc

electrum/network.py
f321x: Also moved the instantiation of the relay_manager from the NostrTransport `__init__` into the async `main_loop` as the ProxyConnector also seems to require being instantiated inside an asyncio loop in the newer aiohttp version, like the `ClientSession`.

SomberNight (approved):
Looks good. Thanks.
This is now blocked on a new release of aionostr, and bumping the corresponding min version.

electrum/util.py
```diff
@@ -1313,6 +1313,18 @@ def format_short_id(short_channel_id: Optional[bytes]):
         + 'x' + str(int.from_bytes(short_channel_id[6:], 'big'))
 
 
+def make_aiohttp_proxy_connector(proxy: 'ProxySettings', ssl_context: Optional[ssl.SSLContext] = None) -> ProxyConnector:
```
ecdsa, line 1316: also used for non-http stuff

PR #9564, opened 2025-02-15
Fix: Handle UTF-8 decode error and overflow error in wallet storage

krishnakant2607:
I've fixed two issues in the wallet storage handling:

- A UnicodeDecodeError that occurs when reading wallet files with invalid UTF-8 encoding
 - An OverflowError that happens when trying to display large error messages

The changes include:

- Better exception handling for UTF-8 decode errors in storage.py
 - Added size limit checks to prevent overflow when displaying error messages
 - More user-friendly error messages
- Graceful fallback for oversized error messages

Changed files:

1. electrum/storage.py
2. electrum/gui/qt/wizard/wallet.py


SomberNight:
is this LLM-generated?

PR #9566, opened 2025-02-17
lnworker: imports, whitespace

accumulator:
Submitted as PR, as there's a small import conflict with #9542

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -7,16 +7,14 @@ import os
 from decimal import Decimal
 import random
 import time
-import operator
-import enum
-from enum import IntEnum, Enum
-from typing import (Optional, Sequence, Tuple, List, Set, Dict, TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


SomberNight:
Thanks.  This kind of thing of going toward a low-entropy state is appreciated :)  however I think ultimately we need to add automation to avoid regressing. Without automation, the clutter will come back naturally. Maybe pycodestyle or similar.
Or perhaps the extreme solution of using [`black`](https://github.com/psf/black) or similar.

accumulator:
> Thanks. This kind of thing of going toward a low-entropy state is appreciated :) however I think ultimately we need to add automation to avoid regressing. Without automation, the clutter will come back naturally. Maybe pycodestyle or similar. Or perhaps the extreme solution of using [`black`](https://github.com/psf/black) or similar.

"The Uncompromising Code Formatter".. heh, nice :)

PR #9568, opened 2025-02-18
Ignore odd bits if even bit is set too in list_enabled_bits

f321x:
We use `list_enabled_bits()` to get the features we signal in ln/gossip messages. Bolt 9 specifies: 

> MUST NOT set both the optional and mandatory bits.

We however set both, optional and required features in messages because `list_enabled_bits()` returns them both if they are set in a LnFeatures instance. This caused the invalid announcement signature warning by eclair because eclair correctly reserialized our message which was different from what we signed, so the signature didn't match anymore.

Example for LnFeatures(175922003768259):
This is what we sent before:
```
['OPTION_DATA_LOSS_PROTECT_REQ', 'OPTION_DATA_LOSS_PROTECT_OPT', 'GOSSIP_QUERIES_REQ', 'GOSSIP_QUERIES_OPT', 'VAR_ONION_REQ', 'VAR_ONION_OPT', 'OPTION_STATIC_REMOTEKEY_REQ', 'OPTION_STATIC_REMOTEKEY_OPT', 'PAYMENT_SECRET_REQ', 'PAYMENT_SECRET_OPT', 'BASIC_MPP_OPT', 'OPTION_SUPPORT_LARGE_CHANNEL_OPT', 'OPTION_ANCHORS_ZERO_FEE_HTLC_OPT', 'OPTION_SHUTDOWN_ANYSEGWIT_OPT', 'OPTION_CHANNEL_TYPE_OPT', 'OPTION_SCID_ALIAS_OPT']
```

This is what we send after this fix:
```
['OPTION_DATA_LOSS_PROTECT_REQ', 'GOSSIP_QUERIES_REQ', 'VAR_ONION_REQ', 'OPTION_STATIC_REMOTEKEY_REQ', 'PAYMENT_SECRET_REQ', 'BASIC_MPP_OPT', 'OPTION_SUPPORT_LARGE_CHANNEL_OPT', 'OPTION_ANCHORS_ZERO_FEE_HTLC_OPT', 'OPTION_SHUTDOWN_ANYSEGWIT_OPT', 'OPTION_CHANNEL_TYPE_OPT', 'OPTION_SCID_ALIAS_OPT']
```

### DIFF
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -1468,28 +1468,28 @@ class LnFeatures(IntFlag):
 
     def for_init_message(self) -> 'LnFeatures':
         features = LnFeatures(0)
-        for flag in list_enabled_bits(self):
+        for flag in list_enabled_ln_feature_bits(self):
             if LnFeatureContexts.INIT & _ln_feature_contexts[1 << flag]:
                 features |= (1 << flag)
         return features
 
     def for_node_announcement(self) -> 'LnFeatures':
         features = LnFeatures(0)
-        for flag in list_enabled_bits(self):
+        for flag in list_enabled_ln_feature_bits(self):
             if LnFeatureContexts.NODE_ANN & _ln_feature_contexts[1 << flag]:
                 features |= (1 << flag)
         return features
 
     def for_invoice(self) -> 'LnFeatures':
         features = LnFeatures(0)
-        for flag in list_enabled_bits(self):
+        for flag in list_enabled_ln_feature_bits(self):
             if LnFeatureContexts.INVOICE & _ln_feature_contexts[1 << flag]:
                 features |= (1 << flag)
         return features
 
     def for_channel_announcement(self) -> 'LnFeatures':
         features = LnFeatures(0)
-        for flag in list_enabled_bits(self):
+        for flag in list_enabled_ln_feature_bits(self):
             ctxs = _ln_feature_contexts[1 << flag]
             if LnFeatureContexts.CHAN_ANN_AS_IS & ctxs:
                 features |= (1 << flag)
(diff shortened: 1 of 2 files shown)


electrum/util.py
```diff
@@ -1760,11 +1760,17 @@ def is_private_netaddress(host: str) -> bool:
...
+    """e.g. 77 (0b1001101) --> (0, 2, 6), ignores odd bit if even bit is
+    set if not `include_redundant_opt`."""
     binary = bin(x)[2:]
     rev_bin = reversed(binary)
-    return tuple(i for i, b in enumerate(rev_bin) if b == '1')
+    enabled_bits = tuple()
+    for index, bit in enumerate(rev_bin):
+        if (bit == '1' # bit is set, and we either want redundant options or it's not redundant
+                and (include_redundant_opt or not (index % 2 != 0 and index - 1 in enabled_bits))):
+            enabled_bits += (index,)
```
SomberNight, line 1772: I think this might make it quadratic time complexity instead of linear (in size of rev_bin). You are not supposed to "append" to a tuple. Should use a list instead, and convert it to a tuple at the end of the loop.

electrum/util.py
```diff
@@ -1760,11 +1760,17 @@ def is_private_netaddress(host: str) -> bool:
     return False
 
 
-def list_enabled_bits(x: int) -> Sequence[int]:
-    """e.g. 77 (0b1001101) --> (0, 2, 3, 6)"""
+def list_enabled_bits(x: int, include_redundant_opt: bool = False) -> Sequence[int]:
+    """e.g. 77 (0b1001101) --> (0, 2, 6), ignores odd bit if even bit is
+    set if not `include_redundant_opt`."""
```
SomberNight, line 1765: This function used to be a pure mathematical "operation", and now you mix BOLT-9 into it :P
It is in util.py, so not even near LN-related code :)

Though true, atm `list_enabled_bits` is only used for BOLT-9.

Maybe this is subjective, but I would prefer if we left `list_enable_bits` as-is, and maybe create a wrapper function around it `list_enabled_ln_features` or sth, which calls `list_enabled_bits` and filters the redundant bits out.

SomberNight (commented):
Nice find! Thanks for tracking down this bug.

electrum/util.py
f321x: Ok i added `list_enabled_ln_feature_bits` to lnutil in https://github.com/spesmilo/electrum/pull/9568/commits/9394b18b4e87efb4a0de004d2ac8461a3d8ee416 and restored the `list_enabled_bits` method.

electrum/util.py
f321x: Thanks for the hint. I now use a set for the inclusion check and return a tuple in https://github.com/spesmilo/electrum/pull/9568/commits/9394b18b4e87efb4a0de004d2ac8461a3d8ee416

SomberNight:
Thanks.

PR #9573, opened 2025-02-20
Simplify submarine swap onchain fee model to single base fee 'mining_fee'

f321x:
This PR simplifies the fees used in submarine swaps to use a single base fee `mining_fee`. This way the user sees the same base fee independent of reverse swap or forward swap.


ecdsa:
looks good. 
we should also display the flat fee in the qml  GUI

electrum/gui/qml/qeswaphelper.py
```diff
@@ -502,7 +498,7 @@ def swap_slider_moved(self):
         self.toreceive = QEAmount(amount_sat=self._receive_amount)
         # fee breakdown
         self.serverfeeperc = f'{swap_manager.percentage:0.1f}%'
-        server_miningfee = swap_manager.lockup_fee if self.isReverse else swap_manager.normal_fee
+        server_miningfee = swap_manager.mining_fee
```
f321x, line 501: > we should also display the flat fee in the qml GUI - @ecdsa

Here I changed the qml gui to just show the `mining_fee` base fee instead of different fees depending on the slider direction. Is this what you mean, or do you want to combine it with the client mining fee? 
This is how it looks right now, 150 sat on both sides: 
Screenshot From 2025-02-21 08-38-48
Screenshot From 2025-02-21 08-38-57

ecdsa: oh no, I mean in the dialog where the user chooses a swap provider

f321x: Ah i see 👍

PR #9580, opened 2025-02-24
reintroduce separate request types for lightning and onchain

ecdsa:
This PR removes on-chain fallback addresses in bolt11 invoices, and removes gap limit restrictions on the number of lightning invoices that can be created. In the GUI, it reintroduces the previous design with two buttons in the receive tab, which is simpler.

The QML component `ReceiveDialog.qml` needs further simplification.
The payserver plugin needs to be adapted too.

### DIFF
--- a/electrum/gui/qml/components/ReceiveDetailsDialog.qml
+++ b/electrum/gui/qml/components/ReceiveDetailsDialog.qml
@@ -106,6 +106,7 @@ ElDialog {
             }
             FlatButton {
                 Layout.fillWidth: true
+                enabled: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > amountBtc.textAsSats.satsInt
                 text: qsTr('Lightning')
                 icon.source: '../../icons/lightning.png'
                 onClicked: { dialog.isLightning = true; doAccept() }
--- a/electrum/gui/qml/components/ReceiveDetailsDialog.qml
+++ b/electrum/gui/qml/components/ReceiveDetailsDialog.qml
@@ -11,12 +11,13 @@ import "controls"
 ElDialog {
     id: dialog
 
-    title: qsTr('Receive payment')
+    title: qsTr('Create Invoice')
     iconSource: Qt.resolvedUrl('../../icons/tab_receive.png')
 
     property alias amount: amountBtc.text
     property alias description: message.text
     property alias expiry: expires.currentValue
+    property bool isLightning: false
 
     padding: 0
 
@@ -93,11 +94,22 @@ ElDialog {
             }
         }
 
-        FlatButton {
-            Layout.fillWidth: true
-            text: qsTr('Create request')
-            icon.source: '../../icons/confirmed.png'
-            onClicked: doAccept()
+        GridLayout {
+            width: parent.width
+	    columns: 2
+
+            FlatButton {
+                Layout.fillWidth: true
+                text: qsTr('Onchain')
+                icon.source: '../../icons/bitcoin.png'
+                onClicked: { dialog.isLightning = false; doAccept() }
+            }
+            FlatButton {
+                Layout.fillWidth: true
+                text: qsTr('Lightning')
+                icon.source: '../../icons/lightning.png'
+                onClicked: { dialog.isLightning = true; doAccept() }
+            }
         }
     }
 
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -2833,6 +2833,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         return invoice
 
     def create_request(self, amount_sat: Optional[int], message: Optional[str], exp_delay: Optional[int], address: Optional[str]):
+        """ will create a lightning request if address is None """
         # for receiving
         amount_sat = amount_sat or 0
         assert isinstance(amount_sat, int), f"{amount_sat!r}"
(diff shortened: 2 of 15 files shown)


electrum/wallet.py
```diff
@@ -2841,9 +2842,11 @@ def create_request(self, amount_sat: Optional[int], message: Optional[str], exp_
         address = address or None  # converts "" to None
         exp_delay = exp_delay or 0
         timestamp = int(Request._get_cur_time())
-        payment_hash = None  # type: Optional[bytes]
-        if self.has_lightning():
+        if address is None:
+            assert self.has_lightning()
             payment_hash = self.lnworker.create_payment_info(amount_msat=amount_msat, write_to_disk=False)
+        else:
+            payment_hash = None
```
SomberNight, line 2849: This makes the `WALLET_BOLT11_FALLBACK` config setting useless. That setting is exposed to the GUIs.

From the context, and also seeing as you removed the unit test that was checking it such addresses are reused as expected, I assume you just forgot to remove this config key?

electrum/gui/qml/components/ReceiveDetailsDialog.qml
```diff
@@ -106,6 +106,7 @@ ElDialog {
             }
             FlatButton {
                 Layout.fillWidth: true
+                enabled: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > amountBtc.textAsSats.satsInt
```
SomberNight, line 109: This seems to have a subtle bug.
If `Daemon.currentWallet.isLightning` is true, but e.g. the wallet does not have any channels, before `amountBtc` is edited for the first time, the button is enabled.
If I type text into `amountBtc` and then clear it, then the button is correctly disabled.

The root cause seems to be that `amountBtc.textAsSats` is null, and when `null.satsInt` is evaluated, instead of some kind of error, `enabled` becomes true. Or rather, I guess `enabled` is true by default, and the error is completely suppressed somewhere. This is very confusing.

Either one of patch1 or patch2 "fixes" the issue, however I am not sure I fully understand what is going on. Could you please take a look, @accumulator?

Patch1:
```patch
diff --git a/electrum/gui/qml/components/controls/BtcField.qml b/electrum/gui/qml/components/controls/BtcField.qml
index 53b0e61c50..8a9ec4ccf0 100644
--- a/electrum/gui/qml/components/controls/BtcField.qml
+++ b/electrum/gui/qml/components/controls/BtcField.qml
@@ -31,4 +31,8 @@ TextField {
                 : ''
         }
     }
+
+    Component.onCompleted: {
+        textAsSats = Config.unitsToSats("")
+    }
 }

```

Patch2:
```patch
diff --git a/electrum/gui/qml/components/ReceiveDetailsDialog.qml b/electrum/gui/qml/components/ReceiveDetailsDialog.qml
index 23c840e9e3..b5cf90b66b 100644
--- a/electrum/gui/qml/components/ReceiveDetailsDialog.qml
+++ b/electrum/gui/qml/components/ReceiveDetailsDialog.qml
@@ -106,7 +106,7 @@ ElDialog {
             }
             FlatButton {
                 Layout.fillWidth: true
-                enabled: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > amountBtc.textAsSats.satsInt
+                enabled: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > (amountBtc.textAsSats ? amountBtc.textAsSats.satsInt : 0)
                 text: qsTr('Lightning')
                 icon.source: '../../icons/lightning.png'
                 onClicked: { dialog.isLightning = true; doAccept() }

```

accumulator: Good catch. It took me a while to find a wallet with an open channel with remote balance, and after additional changes forgot to test the no remote balance case. (Is `htlc.me` broken? It used to be handy for pushing out some sats to the remote side on testnet)

Both patches could do, but I rather have the `textAsSats` property always initialized, so we don't have to add the conditional wherever we want to dereference the amount. So patch1 plz

accumulator: > The root cause seems to be that `amountBtc.textAsSats` is null

It is actually `undefined`, which is != `null`, which is also a possible value in QML javascript (sigh).

> when `null.satsInt` is evaluated, instead of some kind of error, `enabled` becomes true. Or rather, I guess `enabled` is true by default, and the error is completely suppressed somewhere. This is very confusing.

QML is weird. I think what happens is that `undefined`.`property` evaluates to `undefined` again (not an error *I think*, but a warning is logged), which in the `>` comparator then evaluates to `true`

accumulator: One more note about patch1. *replacing* an instance, like we do for `textAsSats` in `BtcField`, can lead to problems if another component creates a *binding* to that property or its subproperties. It depends a bit how the binding is created, but in some cases the binding is not updated to the new instance, especially if the binding is created dynamically, or the reference to the property is passed around, or a property of the instance (e.g.`satsInt`) is bound.

That's why from python scope, I usually try to keep the `QEAmount` instance, but write its property contents (like `satsInt`) instead.

electrum/wallet.py
ecdsa: right, we should remove that config key

Issue #7922, opened 2022-08-08
protobuf breaking changes between 3.20.x and 4.21 (4.x)


SomberNight:
Google made breaking changes in protobuf (see e.g. [here](https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates)), in the generation and parsing of `_pb2.py` generated files.

Re the python [`protobuf` package](https://pypi.org/project/protobuf/#history), the 3.20.1 release is followed by the 4.21.1 release.

AFAICT protoc >=3.19 generates `_pb2.py` files in the new format, while older protoc generates the old format.
AFAICT
- protobuf 3.20.x is able to parse both the old and the new format,
- `protobuf<3.20` can only parse the old format,
- `protobuf>=3.21` (i.e. `protobuf>=4`) can only parse the new format

We rely on `protobuf` to parse `_pb2.py` files at runtime:
- directly for [`paymentrequest_pb2.py`](https://github.com/spesmilo/electrum/blob/73ba00d7dd87af5e63ab2a30617fae1b8d36d0c4/electrum/paymentrequest_pb2.py)
- indirectly via `bitbox02` (hww plugin)
- indirectly via `keepkey` (hww plugin)

-----

- Currently, our `_pb2.py` file(s) is in the old format, and we require `protobuf>=3.12,<4`.
    - see https://github.com/spesmilo/electrum/issues/7833
    - and https://github.com/spesmilo/electrum/commit/52b73880f95be5cf51742fec10cde0a2e4b23de4
- bitbox02 has `_pb2.py` file(s)
    - for bitbox02<6.1.0, in the old format
    - for bitbox02==6.1.0, in the new format
        - and it requires "protobuf>=3.21" (i.e. "protobuf>=4")
			- related [PR](https://github.com/digitalbitbox/bitbox02-firmware/pull/951)
			- ~~would be nice if they relaxed this to accept 3.20.x~~ (done in [PR](https://github.com/digitalbitbox/bitbox02-firmware/pull/964), thanks!)
	- for bitbox02>=6.1.1, in the new format
		- and it requires "protobuf>=3.20"
- keepkey has `_pb2.py` file(s) in the old format 
	- I've asked them to regenerate in new format and publish a new release
		- see [issue](https://github.com/keepkey/python-keepkey/issues/146)

rt121212121:
[...]

PR #9581, opened 2025-02-24
plugins: keepkey: vendor keepkeylib

SomberNight:
This adds keepkeylib into our git tree. We will vendor it that way from now, instead of declaring it as an external dependency.

Upstream keepkeylib is effectively unmaintained -- but the only problem with it is that it requires an old version of `protobuf`. (see https://github.com/keepkey/python-keepkey/issues/146)
By vendoring it, we can fix that ourselves.

keepkeylib is [taken from PyPI](https://files.pythonhosted.org/packages/30/38/558d9a2dd1fd74f50ff4587b4054496ffb69e21ab1138eb448f3e8e2f4a7/keepkey-6.3.1.tar.gz), at keepkey==6.3.1, which is the version we have [sha256 pinned](https://github.com/spesmilo/electrum/blob/30901212d0e4cb503bf0f76f6a5a3c1b0473de2c/contrib/deterministic-build/requirements-hw.txt#L284) for 5 years.

This also vendors the corresponding source protobuf files, which were not part of the keepkeylib sdist. Those are taken from the git repo, at the [corresponding commit](https://github.com/keepkey/python-keepkey/commit/2170fc4479bf5a4b63d2067801ecf041430db769) (for v6.3.1).

Subsequent commits
- strip out some of the shitcoin crap from keepkeylib, though only the parts that was easy to remove
- regenerate the pb2 files using `protoc-21.12` (which is the version currently in debian stable, already much newer than what was previously used)

closes https://github.com/spesmilo/electrum/issues/7922


SomberNight:
note: we could alternatively make this a git submodule,

or we could also just rm the keepkey support.

SomberNight:
> Upstream keepkeylib is effectively unmaintained 

Hmm, looks like there is actually some recent work in branches:
https://github.com/keepkey/python-keepkey/tree/btc-only
https://github.com/keepkey/python-keepkey/tree/feature-p2tr

PR #9584, opened 2025-02-27
wip: Improve JIT channel opening

f321x:
* make it possible to have multiple JIT channels with single LSP by using the deterministic remote scid alias created on channel open instead of the static node id hash as scid alias of the jit channel on LSP side
* fix race between htlc_switch / process_htlc that removed htlc key from active_forwardings while it was still needed causing Exception with slow forwarding callbacks (like jit channel open), and also introduce persisted `dont_settle_htlc_keys` dict on lnworker to block htlc_keys from getting settled back even if we already got the preimage. This way we can manually decide if we want to settle the htlc incoming to the LSP or fail it (e.g. if the channel broadcasting fails) instead of it happening concurrently.
* Only add zeroconf route hint to invoice if we really need a channel, not always, if `ZEROCONF_TRUSTED_NODE` is set.
* disable mpp feature flags for JIT requesting invoices to prevent multiple channels from getting opened for multiple htlcs
* add error handling to correctly wind down and remove channels and local txs if the jit opening failed (e.g. if lsp didn't get preimage from client)
* disable JIT/zeroconf on LSP side if exceptions occur to prevent further issues
* check channels with TX_HEIGHT_LOCAL funding on zeroconf related issues to notice if LSP didn't broadcast funding tx or requested close on channel after failure (see `update_unfunded_state()`), remove trusted LSP from config if there was a serious issue like local balance on an unfunded channel
* freezing inputs of jit funding tx to prevent conflicting channel open txs/double spends, unfreezing on issues unrelated to created tx (`wallet.set_frozen_state_of_tx_inputs()`)
* more extensive regtest tests, with multiple channels, client failing htlc
* accept non recoverable zeroconf channels on client even if recoverable channels are enabled as zeroconf peers are supposed to be trusted and hardcoded, so clients can use JIT without completely disabling recoverable channels. Alternatively we could make the LSP open OP_RETURN channels?

#### Todos:
* improve gui(s)
* `lnworker:1205 -> # fixme: some checks are missing`: find additional relevant checks
* more testing, (w/wo) anchors, (w/wo) trampoline, signet/regtest


f321x:
Note: we should only signal `OPTION_ZEROCONF_OPT` in the init messages if the peer is also our configured trusted zeroconf node.

f321x:
will replace this, it's pretty outdated.

PR #9586, opened 2025-02-28
Add mechanism to block htlcs from settling back

f321x:
Adds a persisted dictionary of htlc keys (= result of `serialize_htlc_key()`) to lnworker to allow adding htlcs which should not be settled back even if we got the required preimage to do so. 
Currently if we open a JIT channel and get the preimage from the client we concurrently settle the incoming htlc to us, however we may still want to fail it in the process of opening the JIT channel which is not finished yet, for example if we were unable to broadcast the funding tx. With this change we can block the settling until the forwarding callback has returned.
<s>Also fixes a key error exception that happened if the incoming htlc got settled before the callback returned, `wrapped_callback` then tried to access `active_forwardings[payment_key]` which has already been removed after settling the incoming htlc.</s>


ecdsa:
note: the key error exception handling was removed from this PR, as it needs to be addressed separately

PR #9588, opened 2025-02-28
Refector on_channel_open to accept incoming channels if the source is a trusted zeroconf node

f321x:
Allows the client to accept an incoming channel even if recoverable channels are enabled if the originator is a trusted zeroconf node. This way the client doesn't have to completely disable recoverable channels to get incoming JIT channels and as the trusted nodes are supposed to be hardcoded the client is able to recover them. 
If we want to move away from hardcoding the providers we could look into opening recoverable jit channels.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -1042,7 +1042,7 @@ class Peer(Logger, EventListener):
                 raise Exception("sender has sent a channel type we don't support")
 
         if self.is_channel_type():
-            is_zeroconf = channel_type & channel_type.OPTION_ZEROCONF
+            is_zeroconf = bool(channel_type & ChannelType.OPTION_ZEROCONF)
             if is_zeroconf and not self.network.config.ZEROCONF_TRUSTED_NODE.startswith(self.pubkey.hex()):
                 raise Exception(f"not accepting zeroconf from node {self.pubkey}")
         else:
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -1041,9 +1041,12 @@ class Peer(Logger, EventListener):
             if not channel_type.complies_with_features(self.features):
                 raise Exception("sender has sent a channel type we don't support")
 
-        is_zeroconf = channel_type & channel_type.OPTION_ZEROCONF
-        if is_zeroconf and not self.network.config.ZEROCONF_TRUSTED_NODE.startswith(self.pubkey.hex()):
-            raise Exception(f"not accepting zeroconf from node {self.pubkey}")
+        if self.is_channel_type():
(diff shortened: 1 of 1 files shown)


electrum/lnpeer.py
```diff
@@ -1055,6 +1041,26 @@ async def on_open_channel(self, payload):
             if not channel_type.complies_with_features(self.features):
                 raise Exception("sender has sent a channel type we don't support")
 
+        is_zeroconf = channel_type & channel_type.OPTION_ZEROCONF
```
SomberNight, line 1044: hmm looks like this is a pre-existing bug also on master, but if what if `channel_type` is None?

SomberNight:
> as the trusted nodes are supposed to be hardcoded the client is able to recover them

What do you mean by hardcoded? Atm it is the `config.ZEROCONF_TRUSTED_NODE`, isn't it? In theory the user could change it arbitrarily.

electrum/lnpeer.py
f321x: Addressed in https://github.com/spesmilo/electrum/pull/9588/commits/bc90f517d110efd8babf38f7ef72ce6c9beda112. I check if `self.is_channel_type()` because it could happen that we have `channel_type` but don't deserialize it if not `self.is_channel_type()`.

f321x:
> > as the trusted nodes are supposed to be hardcoded the client is able to recover them
> 
> What do you mean by hardcoded? Atm it is the `config.ZEROCONF_TRUSTED_NODE`, isn't it? In theory the user could change it arbitrarily.

After discussing about this with @ecdsa we came to the conclusion that it should be ok to accept non recoverable zeroconf as long as it is not exposed to the GUI. When we ship it in gui we probably need to hardcode them / don't allow for change in the gui due to the trust relation in zeroconf channels.

electrum/lnpeer.py
```diff
@@ -1055,6 +1041,29 @@ async def on_open_channel(self, payload):
             if not channel_type.complies_with_features(self.features):
                 raise Exception("sender has sent a channel type we don't support")
 
+        if self.is_channel_type():
+            is_zeroconf = channel_type & channel_type.OPTION_ZEROCONF
```
SomberNight, line 1045: two suggestions:
- use `ChannelType.OPTION_ZEROCONF` instead of `channel_type.OPTION_ZEROCONF`
- cast to bool for sanity - without that it is an int
```suggestion
            is_zeroconf = bool(channel_type & ChannelType.OPTION_ZEROCONF)
```

f321x: Thanks, yeah this looks better. Fixed in cb56f0873dc65dccfd3111d1ca675bc8991a31bb

PR #9590, opened 2025-02-28
Handle unfunded zeroconf channels in update_unfunded_state

f321x:
This adds handling of zeroconf channels with funding tx height `TX_HEIGHT_LOCAL`. This way the client removes channels if the LSP fails to broadcast the funding transaction instead of trying to continue using it. It also allows the LSP to remove the channel on client side by sending shutdown instead of broadcasting the funding tx, the client will then remove the channel as it is in `CLOSING` state, this is useful if the LSP wants to abort the opening procedure, e.g. if the client fails to provide the preimage.

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -55,7 +55,7 @@ from .lnutil import (Outpoint, LocalConfig, RemoteConfig, Keypair, OnlyPubkeyKey
                      ShortChannelID, map_htlcs_to_ctx_output_idxs,
                      fee_for_htlc_output, offered_htlc_trim_threshold_sat,
                      received_htlc_trim_threshold_sat, make_commitment_output_to_remote_address, FIXED_ANCHOR_SAT,
-                     ChannelType, LNProtocolWarning)
+                     ChannelType, LNProtocolWarning, ZEROCONF_TIMEOUT)
 from .lnsweep import sweep_our_ctx, sweep_their_ctx
 from .lnsweep import sweep_their_htlctx_justice, sweep_our_htlctx, SweepInfo
 from .lnsweep import sweep_their_ctx_to_remote_backup
@@ -345,7 +345,11 @@ class AbstractChannel(Logger, ABC):
     def update_unfunded_state(self) -> None:
         self.delete_funding_height()
         self.delete_closing_height()
-        if self.get_state() in [ChannelState.PREOPENING, ChannelState.OPENING, ChannelState.FORCE_CLOSING] and self.lnworker:
+        if not self.lnworker:
+            return
+        chan_age = now() - self.storage.get('init_timestamp', 0)
+        state = self.get_state()
+        if state in [ChannelState.PREOPENING, ChannelState.OPENING, ChannelState.FORCE_CLOSING]:
             if self.is_initiator():
                 # set channel state to REDEEMED so that it can be removed manually
                 # to protect ourselves against a server lying by omission,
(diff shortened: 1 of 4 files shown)


electrum/lnchannel.py
```diff
@@ -365,15 +369,35 @@ def update_unfunded_state(self) -> None:
                             self.set_state(ChannelState.REDEEMED)
                             break
             else:
-                if self.lnworker and (now() - self.storage.get('init_timestamp', 0) > CHANNEL_OPENING_TIMEOUT):
+                if chan_age > CHANNEL_OPENING_TIMEOUT:
                     self.lnworker.remove_channel(self.channel_id)
+        elif self.is_zeroconf() and state in [ChannelState.OPEN, ChannelState.CLOSING, ChannelState.FORCE_CLOSING]:
+            # handling zeroconf channels with no funding tx, can happen if broadcasting fails on LSP side
+            # or if the LSP did double spent the funding tx/never published it intentionally
+            if state != ChannelState.OPEN or chan_age > 60 * 10 and 'init_timestamp' in self.storage:
```
ecdsa, line 377: `init_timestamp` is always in storage for recently created channels.
we should just have an assert for that one line above.

ecdsa: maybe define `ZEROCONF_TIMEOUT = 60 * 10`

electrum/lnchannel.py
```diff
@@ -365,15 +369,35 @@ def update_unfunded_state(self) -> None:
...
+                # or if the channel is not seeing any funding tx after 10 minutes to prevent further usage (limit damage)
+                self.set_state(ChannelState.REDEEMED, force=True)
+                local_balance_sat = int(self.balance(LOCAL) // 1000)
+                if local_balance_sat > 0:
+                    self.logger.warning(
+                        f"we may have been scammed out of {local_balance_sat} sat by our "
+                        f"JIT provider: {self.lnworker.config.ZEROCONF_TRUSTED_NODE} or he didn't use our preimage")
+                    self.lnworker.config.ZEROCONF_TRUSTED_NODE = ''
+                self.lnworker.lnwatcher.unwatch_channel(self.get_funding_address(), self.funding_outpoint.to_str())
+                self.lnworker.lnwatcher.adb.remove_transaction(self.funding_outpoint.txid)
```
ecdsa, line 388: why `remove_transaction` ?

electrum/lnchannel.py
ecdsa: what if the client is offline for more than 10 minutes?

f321x: In https://github.com/spesmilo/electrum/pull/9590/commits/5995299caf334ff7dec01dd88606f060ff9b84d3:
Changed it so `init_timestamp` is accessed directly instead through get, so KeyError would be raised if it doesn't exist. 
Added `ZEROCONF_TIMEOUT` constant in lnutil.
Added a check if `network.is_connected()` when we check `channel_age` so it only allows to remove the channel if it is timed out and we are online, so this should prevent the client from accidentally removing an `OPEN` channel if it went offline right after the opening procedure and not seeing the funding tx in the mempool.

electrum/lnchannel.py
f321x: `remove_transaction` is called to get rid of all potential leftover local transactions of the failed channel opening, mostly to remove the invalid closing transaction the client would try to broadcast otherwise if it has been sent shutdown by the LSP.

PR #9591, opened 2025-02-28
Make question dialog of trampoline checkbox child of SettingsDialog

f321x:
Passes the SettingsDialog as parent to the question dialog shown when disabling the trampoline checkbox. This makes the question dialog show in front of the SettingsDialog and the SettingsDialog stay in focus. Before this the SettingsDialog would move behind the main window after answering the question dialog which is confusing.


SomberNight:
thx

PR #9597, opened 2025-03-03
separate fee policy from config

ecdsa:
This will allow tx batching to use different fee policies for different batches.
It is also a conceptual cleanup.

 - Wallet.make_unsigned_transaction takes a FeePolicy parameter
 - fee sliders act on a FeePolicy instead of config
 - different fee policies may be used for different purposes
 - do not detect dust outputs in lnsweep, delegate that to lnwatcher


electrum/commands.py
```diff
@@ -761,12 +761,25 @@ async def verifymessage(self, address, signature, message):
         message = util.to_bytes(message)
         return bitcoin.verify_usermessage_with_address(address, sig, message)
 
+    def _get_fee_policy(self, fee, feerate):
+        if fee and feerate:
```
SomberNight, line 765:
```suggestion
        if fee is not None and feerate is not None:
```
setting `fee=0` is a perfectly valid setting that is not None

electrum/fee_policy.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def closest_index(value, array) -> int:
+    dist = list(map(lambda x: abs(x - value), array))
+    return min(range(len(dist)), key=dist.__getitem__)
+
+
+class FeeMethod(IntEnum):
+    FIXED = 0
+    STATIC = 1
```
SomberNight, line 38: please at least add some comments, the names are not clear:
```suggestion
    FIXED = 0  # fixed absolute fee
    STATIC = 1  # static fee rate
```
perhaps we should even rename them.
e.g. `STATIC` -> `STATICRATE`
or `FIXED` -> `ABS`

electrum/fee_policy.py
```diff
@@ -0,0 +1,369 @@
...
+# warn user if fee/amount for on-chain tx is higher than this
+FEE_RATIO_HIGH_WARNING = 0.05
+
+
+def closest_index(value, array) -> int:
+    dist = list(map(lambda x: abs(x - value), array))
+    return min(range(len(dist)), key=dist.__getitem__)
+
+
+class FeeMethod(IntEnum):
```
SomberNight, line 36:
```suggestion
# note: careful changing the enum names! they appear in the config files.
class FeeMethod(IntEnum):
```

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -67,21 +68,19 @@ def sliderPos(self, sliderPos):
     methodChanged = pyqtSignal()
     @pyqtProperty(int, notify=methodChanged)
     def method(self):
-        return self._method
+        # -1 because the first element is FeeMethod.FIXED
+        return int(self._fee_policy.method) - 1
```
SomberNight, line 72: magic mapping between unrelated enums, huh?

Please do the sensible thing and have an explicit conversion:
```diff
diff --git a/electrum/gui/qml/components/controls/FeeMethodComboBox.qml b/electrum/gui/qml/components/controls/FeeMethodComboBox.qml
index 56cbb829f3..8c4ef250f9 100644
--- a/electrum/gui/qml/components/controls/FeeMethodComboBox.qml
+++ b/electrum/gui/qml/components/controls/FeeMethodComboBox.qml
@@ -12,9 +12,9 @@ ElComboBox {
     valueRole: 'value'
 
     model: [
-        { text: qsTr('ETA'), value: 1 },
-        { text: qsTr('Mempool'), value: 2 },
-        { text: qsTr('Static'), value: 0 }
+        { text: qsTr('ETA'), value: FeeSlider.FSMethod.ETA },
+        { text: qsTr('Mempool'), value: FeeSlider.FSMethod.MEMPOOL },
+        { text: qsTr('Static'), value: FeeSlider.FSMethod.STATIC }
     ]
     onCurrentValueChanged: {
         if (activeFocus)
diff --git a/electrum/gui/qml/qeapp.py b/electrum/gui/qml/qeapp.py
index 6a8334082c..375252f7ae 100644
--- a/electrum/gui/qml/qeapp.py
+++ b/electrum/gui/qml/qeapp.py
@@ -29,7 +29,7 @@ from .qeqr import QEQRParser, QEQRImageProvider, QEQRImageProviderHelper
 from .qeqrscanner import QEQRScanner
 from .qebitcoin import QEBitcoin
 from .qefx import QEFX
-from .qetxfinalizer import QETxFinalizer, QETxRbfFeeBumper, QETxCpfpFeeBumper, QETxCanceller, QETxSweepFinalizer
+from .qetxfinalizer import QETxFinalizer, QETxRbfFeeBumper, QETxCpfpFeeBumper, QETxCanceller, QETxSweepFinalizer, FeeSlider
 from .qeinvoice import QEInvoice, QEInvoiceParser
 from .qerequestdetails import QERequestDetails
 from .qetypes import QEAmount
@@ -407,6 +407,7 @@ class ElectrumQmlApplication(QGuiApplication):
         qmlRegisterType(QETxCanceller, 'org.electrum', 1, 0, 'TxCanceller')
         qmlRegisterType(QETxSweepFinalizer, 'org.electrum', 1, 0, 'SweepFinalizer')
         qmlRegisterType(QEBip39RecoveryListModel, 'org.electrum', 1, 0, 'Bip39RecoveryListModel')
+        qmlRegisterType(FeeSlider, 'org.electrum', 1, 0, 'FeeSlider')
 
         # TODO QT6: these were declared as uncreatable, but that doesn't seem to work for pyqt6
         # qmlRegisterUncreatableType(QEAmount, 'org.electrum', 1, 0, 'Amount', 'Amount can only be used as property')
diff --git a/electrum/gui/qml/qetxfinalizer.py b/electrum/gui/qml/qetxfinalizer.py
index 26eed4398f..ea7ae4cc70 100644
--- a/electrum/gui/qml/qetxfinalizer.py
+++ b/electrum/gui/qml/qetxfinalizer.py
@@ -1,10 +1,11 @@
 import copy
+from enum import IntEnum
 import threading
 from decimal import Decimal
 from typing import Optional, TYPE_CHECKING
 from functools import partial
 
-from PyQt6.QtCore import pyqtProperty, pyqtSignal, pyqtSlot, QObject
+from PyQt6.QtCore import pyqtProperty, pyqtSignal, pyqtSlot, QObject, pyqtEnum
 
 from electrum.logging import get_logger
 from electrum.i18n import _
@@ -25,13 +26,35 @@ if TYPE_CHECKING:
 
 
 class FeeSlider(QObject):
+
+    @pyqtEnum
+    class FSMethod(IntEnum):
+        STATIC = 0
+        ETA = 1
+        MEMPOOL = 2
+
+        def to_fee_method(self) -> 'FeeMethod':
+            return {
+                self.STATIC: FeeMethod.STATIC,
+                self.ETA: FeeMethod.ETA,
+                self.MEMPOOL: FeeMethod.MEMPOOL,
+            }[self]
+
+        @classmethod
+        def from_fee_method(cls, fm: FeeMethod) -> 'FeeSlider.FSMethod':
+            return {
+                FeeMethod.STATIC: cls.STATIC,
+                FeeMethod.ETA: cls.ETA,
+                FeeMethod.MEMPOOL: cls.MEMPOOL,
+            }[fm]
+
     def __init__(self, parent=None):
         super().__init__(parent)
 
         self._wallet = None  # type: Optional[QEWallet]
         self._sliderSteps = 0
         self._sliderPos = 0
-        self._fee_policy = None
+        self._fee_policy = None  # type: Optional[FeePolicy]
         self._target = ''
         self._config = None  # type: Optional[SimpleConfig]
 
@@ -68,13 +91,13 @@ class FeeSlider(QObject):
     methodChanged = pyqtSignal()
     @pyqtProperty(int, notify=methodChanged)
     def method(self):
-        # -1 because the first element is FeeMethod.FIXED
-        return int(self._fee_policy.method) - 1
+        fsmethod = self.FSMethod.from_fee_method(self._fee_policy.method)
+        return int(fsmethod)
 
     @method.setter
-    def method(self, method):
-        # +1 because the first element is FeeMethod.FIXED
-        method = FeeMethod(method + 1)
+    def method(self, method: int):
+        fsmethod = self.FSMethod(method)
+        method = fsmethod.to_fee_method()
         if self._fee_policy.method != method:
             self._fee_policy.set_method(method)
             self.update_slider()

```

electrum/fee_policy.py
```diff
@@ -0,0 +1,369 @@
...
+    def __repr__(self):
+        return self.get_descriptor()
+
+    def get_descriptor(self) -> str:
+        return self.method.name.lower() + ':' + str(self.value)
+
+    def set_method(self, method: FeeMethod):
+        assert isinstance(method, FeeMethod)
+        self.method = method
+        self.value = 0
```
SomberNight, line 65: why 0?
A value of 0 means that, in the GUI, when selecting a different FeeMethod, the slider jump to one of the extremes:
- for mempool or ETA, it jumps to a very high feerate (0 block ETA, or 0 MB from tip)
- from static rate, it jumps to a very low feerate (0 s/b)

Maybe we should have sensible defaults for each `FeeMethod` instead, somewhere in the middle of the slider.

electrum/fee_policy.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+class FeePolicy:
+    # object associated to a fee slider
+
+    def __init__(self, descriptor: str):
+        try:
+            name, value = descriptor.split(':')
+            self.method = FeeMethod[name.upper()]
+            self.value = int(value)
```
SomberNight, line 50:
```suggestion
            self.value = int(value)  # target (e.g. num blocks, nbytes from mempool tip, sat/kbyte)
```

electrum/gui/qt/fee_slider.py
```diff
@@ -5,16 +5,16 @@
...
 
 class FeeComboBox(QComboBox):
 
     def __init__(self, fee_slider):
         QComboBox.__init__(self)
-        self.config = fee_slider.config
         self.fee_slider = fee_slider
         self.addItems([_('Static'), _('ETA'), _('Mempool')])
-        self.setCurrentIndex((2 if self.config.use_mempool_fees() else 1) if self.config.is_dynfee() else 0)
+        self.setCurrentIndex(int(self.fee_slider.fee_policy.method) - 1)
```
SomberNight, line 17: ah we have the same implicit magic mapping here too :/

electrum/fee_policy.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def fee_per_byte(self, network: 'Network') -> Optional[int]:
+        """Returns sat/vB fee to pay for a txn.
+        Note: might return None.
+        """
+        fee_per_kb = self.fee_per_kb(network)
+        return fee_per_kb / 1000 if fee_per_kb is not None else None
+
+    def estimate_fee(self, network: 'Network', size: Union[int, float, Decimal], *,
+                     allow_fallback_to_static_rates: bool = False) -> int:
```
SomberNight, line 202: Let's make network kwarg-only.
Also, looks like it should be `Optional['Network']`.
```suggestion
    def estimate_fee(
        self,
        size: Union[int, float, Decimal],
        *,
        network: Optional['Network'],
        allow_fallback_to_static_rates: bool = False,
    ) -> int:
```

electrum/wallet.py
```diff
@@ -1876,23 +1871,12 @@ def make_unsigned_transaction(
                 i_max_sum += weight
                 i_max.append((weight, i))
 
-        if fee is None and self.config.fee_per_kb() is None:
-            raise NoDynamicFeeEstimates()
```
SomberNight, line 1880: Something similar to this check is needed.
```
  File "...\electrum\electrum\gui\qt\confirm_tx_dialog.py", line 119, in timer_actions
    self.update()
  File "...\electrum\electrum\gui\qt\confirm_tx_dialog.py", line 123, in update
    self.update_tx()
  File "...\electrum\electrum\gui\qt\confirm_tx_dialog.py", line 635, in update_tx
    self.tx = self.make_tx(fee_policy, confirmed_only=confirmed_only)
  File "...\electrum\electrum\gui\qt\send_tab.py", line 328, in <lambda>
    make_tx = lambda fee_policy, *, confirmed_only=False: self.wallet.make_unsigned_transaction(
  File "...\electrum\electrum\util.py", line 492, in do_profile
    o = func(*args, **kw_args)
  File "...\electrum\electrum\wallet.py", line 1921, in make_unsigned_transaction
    tx = coin_chooser.make_tx(
  File "...\electrum\electrum\coinchooser.py", line 343, in make_tx
    all_buckets = self.bucketize_coins(coins, fee_estimator_vb=fee_estimator_vb)
  File "...\electrum\electrum\coinchooser.py", line 123, in bucketize_coins
    constant_fee = fee_estimator_vb(2000) == fee_estimator_vb(200)
  File "...\electrum\electrum\fee_policy.py", line 213, in estimate_fee
    return self.estimate_fee_for_feerate(fee_per_kb, size)
  File "...\electrum\electrum\fee_policy.py", line 220, in estimate_fee_for_feerate
    fee_per_kb = Decimal(fee_per_kb)
TypeError: conversion from NoneType to Decimal is not supported
```

electrum/lnworker.py
```diff
@@ -1345,16 +1348,17 @@ def suggest_funding_amount(self, amount_to_pay, coins):
         min_funding_sat = max(min_funding_sat, 100_000) # at least 1mBTC
         if min_funding_sat > self.config.LIGHTNING_MAX_FUNDING_SAT:
             return
-        fee_est = partial(self.config.estimate_fee, allow_fallback_to_static_rates=True)  # to avoid NoDynamicFeeEstimates
+        #fee_est = partial(self.config.estimate_fee, allow_fallback_to_static_rates=True)  # to avoid NoDynamicFeeEstimates
+        fee_policy = FeePolicy(self.config.FEE_POLICY)
```
SomberNight, line 1352: soooo, what about NoDynamicFeeEstimates?
We could make `allow_fallback_to_static_rates` a field in `FeePolicy`. Then we could pass that to the constructor here.

electrum/wallet.py
```diff
@@ -1840,7 +1835,7 @@ def make_unsigned_transaction(
             coins: Sequence[PartialTxInput],
```
SomberNight, line 1835: changes needed, ~13 lines above. `config.estimate_fee` does not exist anymore:
```py
fee = partial(self.config.estimate_fee, allow_fallback_to_static_rates=True) # to avoid NoDynamicFeeEstimates
```
(I cannot comment directly on that line as it is too far away from any diff)

electrum/simple_config.py
```diff
@@ -1048,11 +673,7 @@ def __setattr__(self, name, value):
     TEST_SHUTDOWN_FEE_RANGE = ConfigVar('test_shutdown_fee_range', default=None)
     TEST_SHUTDOWN_LEGACY = ConfigVar('test_shutdown_legacy', default=False, type_=bool)
 
-    FEE_EST_DYNAMIC = ConfigVar('dynamic_fees', default=True, type_=bool)
```
SomberNight, line 1051: This is still used in `test_lnpeer.TestPeerDirect.test_close_upfront_shutdown_script`.

```
        w1.network.config.FEE_EST_DYNAMIC = False
        w2.network.config.FEE_EST_DYNAMIC = False
        w1.network.config.FEE_EST_STATIC_FEERATE = 5000
        w2.network.config.FEE_EST_STATIC_FEERATE = 1000
```
Unfortunately the ConfigVar logic is not smart enough to error on this :/  I wonder if I could fix that.

SomberNight: ref https://github.com/spesmilo/electrum/pull/9600

electrum/gui/qt/fee_slider.py
SomberNight: you've changed this in https://github.com/spesmilo/electrum/pull/9597/commits/b4e95f7d7a4c2c7b43f67b8ee53c617cc4094c7b, but it is still not fixed. The whole thing still implicitly depends on the order of items in the `addItems` call:
```
self.addItems([_('Static'), _('ETA'), _('Mempool')])
```
See e.g.:
https://github.com/spesmilo/electrum/commit/ccc1897f36b3f8017b669dd15e38c96534a6d524

electrum/gui/qml/qetxfinalizer.py
SomberNight: you've changed this in https://github.com/spesmilo/electrum/pull/9597/commits/b4e95f7d7a4c2c7b43f67b8ee53c617cc4094c7b, but it is still not fixed.

See this:
https://github.com/spesmilo/electrum/blob/2cae6bb0ab940e69097ee93b072a3ec2662ae93f/electrum/gui/qml/components/controls/FeeMethodComboBox.qml#L15-L17
If someone changes the 0-1-2 magic numbers there, e.g. just reorders them, the whole thing breaks. That is, you have an implicit invisible mapping. This is what I want to avoid, and this is what my example patch above would fix.

electrum/fee_policy.py
ecdsa: added more sensible values

electrum/gui/qt/fee_slider.py
ecdsa: now uses `self.addItems([x.name_for_GUI() for x in FeeMethod.slider_values()])`

electrum/wallet.py
ecdsa:
```python
        if fee_policy.method is not FeeMethod.FIXED and fee_policy.fee_per_kb(self.network) is None:
            raise NoDynamicFeeEstimates()


```

electrum/lnworker.py
ecdsa: not done; this would not work everywhere. 
I think it is better to pass `allow_fallback_to_static_rates` as a parameter of `estimate_fee`
Note that a fallback to static feerates is still hard-coded for regtest, see FeePolicy.fee_per_kb

ecdsa: note: before this commit, `allow_fallback_to_static_rates` was not passed to `mktx_for_open_channel` when called from `open_channel_with_peer`, because it used `fee_est=None` .

PR #9598, opened 2025-03-03
Handle zeroconf lightning requests in QT gui

f321x:
Adds handling of zeroconf payment requests to qt gui by checking if user can receive through zeroconf channel. If possible it shows a warning that has to be accepted first before the invoice is shown. Prefers swaps over zeroconf channels as they require less trust and are more capital efficient.

https://github.com/user-attachments/assets/0a37b82c-8e42-4ce7-85d9-51ed1e8687f0

image

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -53,6 +53,7 @@ from .i18n import _
 from .bip32 import BIP32Node, convert_bip32_intpath_to_strpath, convert_bip32_strpath_to_intpath
 from .crypto import sha256
 from . import util
+from .lntransport import extract_nodeid
 from .util import (NotEnoughFunds, UserCancelled, profiler, OldTaskGroup, ignore_exceptions,
                    format_satoshis, format_fee_satoshis, NoDynamicFeeEstimates,
                    WalletFileException, BitcoinException,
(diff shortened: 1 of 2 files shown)


electrum/wallet.py
```diff
@@ -3309,21 +3322,33 @@ def get_help_texts_for_receive_request(self, req: Request) -> ReceiveRequestHelp
...
-                        ln_help += '\n\n' + _('You may have that capacity if you rebalance your channels.')
-                    elif bool(ln_swap_suggestion):
-                        ln_help += '\n\n' + _('You may have that capacity if you swap some of your funds.')
+                    # prefer to use swaps over JIT channels if possible
+                    if can_get_zeroconf_channel and not bool(ln_rebalance_suggestion) and not bool(ln_swap_suggestion):
+                        if amount_sat < MIN_FUNDING_SAT:
+                            ln_is_error = True
+                            ln_help = (_('Cannot receive this payment. Request at least {} sat '
+                                        'to purchase a Lightning channel from your service provider.')
+                                       .format(MIN_FUNDING_SAT))
```
ecdsa, line 3340: could we use the base unit set by the user here?

f321x: Yes, fixed in https://github.com/spesmilo/electrum/pull/9598/commits/5134d07b147e30f46949c6bb9e32beeb2eb6c269

f321x: Also changed it in the qml gui https://github.com/spesmilo/electrum/pull/9606

PR #9599, opened 2025-03-03
Cache font-ids that were already loaded

oren-z0:
`get_font_id` also saves the trouble of specifying the entire path of the fonts.

### DIFF
--- a/electrum/gui/common_qt/util.py
+++ b/electrum/gui/common_qt/util.py
@@ -1,12 +1,26 @@
 from typing import Optional
+import os.path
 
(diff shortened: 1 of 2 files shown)


ecdsa:
I think this should be moved to `common_qt`, because it uses Qt

oren-z0:
@ecdsa I've moved the function but now the fonts fail to load and I don't understand why.

oren-z0:
Moved the function to `common_qt`

ecdsa:
thanks

PR #9606, opened 2025-03-05
Handle just in time channels in qml request creation

f321x:
Allows to create invoices in QML gui if client has configured trusted zeroconf channels. Shows according warning when a channel is going to get opened or when the amount is too low to receive the payment.
Screenshot From 2025-03-05 11-03-09
Screenshot From 2025-03-05 11-03-18

### DIFF
--- a/electrum/gui/qml/components/ReceiveDetailsDialog.qml
+++ b/electrum/gui/qml/components/ReceiveDetailsDialog.qml
@@ -106,10 +106,42 @@ ElDialog {
             }
             FlatButton {
                 Layout.fillWidth: true
-                enabled: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > amountBtc.textAsSats.satsInt
+                enabled: Daemon.currentWallet.isLightning && (Daemon.currentWallet.lightningCanReceive.satsInt
+                            > amountBtc.textAsSats.satsInt || Daemon.currentWallet.canGetZeroconfChannel)
                 text: qsTr('Lightning')
                 icon.source: '../../icons/lightning.png'
-                onClicked: { dialog.isLightning = true; doAccept() }
+                onClicked: {
+                    if (Daemon.currentWallet.lightningCanReceive.satsInt > amountBtc.textAsSats.satsInt) {
+                        // can receive on existing channel
+                        dialog.isLightning = true
+                        doAccept()
+                    } else if (Daemon.currentWallet.canGetZeroconfChannel && amountBtc.textAsSats.satsInt
+                                >= Daemon.currentWallet.minChannelFunding.satsInt) {
+                        // ask for confirmation of zeroconf channel to prevent fee surprise
+                        var confirmdialog = app.messageDialog.createObject(dialog, {
+                            title: qsTr('Confirm just-in-time channel'),
+                            text: [qsTr('Receiving this payment will purchase a Lightning channel from your service provider.'),
+                                   qsTr('Fees will be deducted from the payment.'),
+                                   qsTr('Do you want to continue?')].join(' '),
+                            yesno: true
+                        })
+                        confirmdialog.accepted.connect(function () {
+                            dialog.isLightning = true
+                            doAccept()
+                        })
+                        confirmdialog.open()
+                    } else {
+                        // show error that amnt > 200k is neccessary to get zeroconf channel
+                        var confirmdialog = app.messageDialog.createObject(dialog, {
+                            title: qsTr("Amount too low"),
+                            text: [qsTr("You don't have channels with enough inbound liquidity to receive this payment."),
+                                   qsTr("Request at least %1 to open a channel just-in-time.").arg(
+                                       Config.formatSats(Daemon.currentWallet.minChannelFunding.satsInt, true))].join(' ')
+                        })
+                        confirmdialog.open()
+                    }
+                    // can't get zeroconf channel and doesn't have enough inbound liquidity
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


electrum/gui/qml/qewallet.py
```diff
@@ -458,6 +459,11 @@ def canSignWithoutCosigner(self):
     def canSignMessage(self):
         return not isinstance(self.wallet, Multisig_Wallet) and not self.wallet.is_watching_only()
 
+    zeroconfAvailabilityChanged = pyqtSignal()
+    @pyqtProperty(bool, notify=zeroconfAvailabilityChanged)
+    def canGetZeroconfChannel(self) -> bool:
```
accumulator, line 464: The convention is to name the signal the same as the property if it's used exclusively for that property, e.g. `canGetZeroconfChannel` property -> `canGetZeroconfChannelChanged` signal.

Also note that this property is evaluated the moment `ReceiveDetailsDialog` is constructed, but not updated later if the zeroconf peer connects/disconnects. The signal should ideally be emitted if the zeroconf peer connected state changes, to trigger re-evaluation of the binding in qml. However, I don't think this is cricital for this PR, as `ReceiveDetailsDialog` is _usually_ short-lived, but if the user keeps it open for longer, the likelyhood of the connected state getting outdated increases.

electrum/gui/qml/qerequestdetails.py
```diff
@@ -118,9 +118,13 @@ def expiration(self):
...
     def bolt11(self):
-        can_receive = self._wallet.wallet.lnworker.num_sats_can_receive() if  self._wallet.wallet.lnworker else 0
-        if self._req and can_receive > 0 and (self._req.get_amount_sat() or 0) <= can_receive:
-            bolt11 = self._wallet.wallet.get_bolt11_invoice(self._req)
+        wallet = self._wallet.wallet
+        amount_sat = self._req.get_amount_sat() or 0 if self._req else 0
+        can_receive = wallet.lnworker.num_sats_can_receive() if wallet.lnworker else 0
+        will_req_zeroconf = wallet.lnworker.receive_requires_jit_channel(amount_msat=amount_sat*1000)
+        if self._req and ((can_receive > 0 and amount_sat <= can_receive)
+                            or (will_req_zeroconf and amount_sat >= 200_000)):
```
accumulator, line 126: hardcoded 200_000 limit should probably be defined in a symbol.

electrum/lnworker.py
```diff
@@ -2734,17 +2734,27 @@ def receive_requires_jit_channel(self, amount_msat: Optional[int]) -> bool:
...
             return True
         return False
 
+    def can_get_zeroconf_channel(self) -> bool:
+        if not self.config.ACCEPT_ZEROCONF_CHANNELS and self.config.ZEROCONF_TRUSTED_NODE:
+            # check if zeroconf is accepted and client has trusted zeroconf node configured
+            return False
+        try:
+            node_id = extract_nodeid(self.wallet.config.ZEROCONF_TRUSTED_NODE)[0]
+        except Exception:
```
accumulator, line 2752: You can catch `ConnStringFormatError` instead here.

accumulator (changes requested):

electrum/lnworker.py
f321x: Changed in https://github.com/spesmilo/electrum/pull/9606/commits/7a073e65b08814910a4c9157dc39c6217372f3ff

electrum/gui/qml/qerequestdetails.py
f321x: Changed in 7a073e65b08814910a4c9157dc39c6217372f3ff, using the MIN_FUNDING_SAT from lnutil now as it serves the same purpose

electrum/gui/qml/qewallet.py
f321x: Thanks for the hint, changed the naming in 7a073e65b08814910a4c9157dc39c6217372f3ff.

electrum/gui/qml/components/ReceiveDetailsDialog.qml
```diff
@@ -106,10 +106,41 @@ ElDialog {
...
+                            dialog.isLightning = true
+                            doAccept()
+                        })
+                        confirmdialog.open()
+                    } else {
+                        // show error that amnt > 200k is neccessary to get zeroconf channel
+                        var confirmdialog = app.messageDialog.createObject(dialog, {
+                            title: qsTr('Amount too low'),
+                            text: qsTr("You don't have channels with enough inbound liquidity to receive ")
+                                + qsTr("this payment. Request at least 200 000 sat to open a channel just-in-time."),
```
accumulator, line 138: The text here should be parameterized as well, e.g. `qsTr('%1 sat').arg(Config.formatSats(Daemon.currentWallet.minChannelFundingSat))`

accumulator (changes requested):

electrum/gui/qml/qewallet.py
```diff
@@ -505,6 +512,10 @@ def lightningCanReceive(self):
             self._lightningcanreceive.satsInt = int(self.wallet.lnworker.num_sats_can_receive())
         return self._lightningcanreceive
 
+    @pyqtProperty(int, notify=dataChanged)
```
accumulator, line 515: Not a problem here, as `MIN_FUNDING_SAT` will probably always fit 32 bit unsigned integer, but FYI:
See 7d8a5cc259724449b572ea7b537a6152978ed0c8

accumulator (changes requested):

electrum/gui/qml/components/ReceiveDetailsDialog.qml
f321x: thx, missed that, fixed in https://github.com/spesmilo/electrum/pull/9606/commits/257b63c996fcb88a3e5fb24db90e70bf802c5e66

electrum/gui/qml/qewallet.py
f321x: Oh good to know, thx. I changed it to `QEAmount` in https://github.com/spesmilo/electrum/pull/9606/commits/257b63c996fcb88a3e5fb24db90e70bf802c5e66 to have it consistent with the other methods, won't hurt at least.

electrum/gui/qml/components/ReceiveDetailsDialog.qml
accumulator: sorry, my short hint was a bit confusing, I added a commit. This makes the strings better translatable (i.e. no space suffix, and don't split paragraphs)

PR #9608, opened 2025-03-05
asyncio: hold our own strong refs for tasks and futures

SomberNight:
see https://docs.python.org/3.13/library/asyncio-task.html#asyncio.create_task :

> Important
> Save a reference to the result of this function, to avoid a task
> disappearing mid-execution. The event loop only keeps weak references
> to tasks. A task that isn’t referenced elsewhere may get garbage
> collected at any time, even before it’s done. For reliable
> “fire-and-forget” background tasks, gather them in a collection

ref https://github.com/python/cpython/issues/91887

### DIFF
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -755,7 +755,6 @@ class Network(Logger, NetworkRetryManager[ServerAddr]):
 
         proxy = self.proxy
         if proxy and proxy.enabled and proxy.mode == 'socks5':
-            # FIXME GC issues? do we need to store the Future?
             asyncio.run_coroutine_threadsafe(tor_probe_task(proxy), self.asyncio_loop)
 
     @log_exceptions
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1652,6 +1652,7 @@ def create_and_start_event_loop() -> Tuple[asyncio.AbstractEventLoop,
             _asyncio_event_loop = None
 
     loop.set_exception_handler(on_exception)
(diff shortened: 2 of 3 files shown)


SomberNight:
for example, this kind of usage is not safe (without this PR):
https://github.com/spesmilo/electrum/blob/b88d9f9d066c2ae245cdc8735814ee916f9bd539/electrum/lnpeer.py#L3079

accumulator:
Interesting read...

Does this make this existing specific strong reference storing code redundant?

https://github.com/spesmilo/electrum/blob/0b3a28358681a0fe101bd6fc3feb5ff981fb68ce/electrum/util.py#L1883-L1893

SomberNight:
> Does this make this existing specific strong reference storing code redundant?
> 
> https://github.com/spesmilo/electrum/blob/0b3a28358681a0fe101bd6fc3feb5ff981fb68ce/electrum/util.py#L1883-L1893

Well, I don't know.
I had an earlier version of this commit where I undid that, yes.
But ultimately I don't know if `run_coroutine_threadsafe` is affected by this GC issue at all. 
I tried to ask in https://github.com/python/cpython/issues/88831#issuecomment-1426863907, but never got a reply.

If it is affected, this current PR might not help it. When writing the unit test, I also tried to check if `run_coroutine_threadsafe` is going through the task factory and it seems it is not...

SomberNight:
Anyway, I'll merge this. It cannot hurt.

PR #9611, opened 2025-03-07
Additional regtest tests for just in time channels

f321x:
Depends on https://github.com/spesmilo/electrum/pull/9584, to be evaluated after 9584 is merged.

Adds 2 additional regtest test for just in time opening, one for unwinding a failed opening procedure, the other one for a payment with exiting channels to test if the client correctly determines if it should request a jit channel or not. The existing test now does two payments which should open two channels to see if multiple jit openings to the same client work.

### DIFF
--- a/tests/regtest.py
+++ b/tests/regtest.py
@@ -135,8 +135,17 @@ class TestLightningJIT(TestLightning):
         }
     }
 
-    def test_just_in_time(self):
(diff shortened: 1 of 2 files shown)


f321x:
https://github.com/spesmilo/electrum/pull/10463 extends one test

PR #9613, opened 2025-03-07
regtests: in py ctx, refer to configvars directly, not as str literals

SomberNight:
though in the bash script, we cannot do this

### DIFF
--- a/tests/regtest.py
+++ b/tests/regtest.py
@@ -2,8 +2,13 @@ import os
 import sys
 import unittest
 import subprocess
+from typing import Mapping, Any
+
+from electrum.simple_config import SimpleConfig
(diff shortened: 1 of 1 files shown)


ecdsa:
While I understand your desire to have less free variables, I do not approve this particular change. 
I think this lacks practical thinking, and it is becoming counter-productive. 
 - the code is less readable.
 - it does work for all settings, so things become heterogeneous.
 - these are parameters to `setconfig`. (this if why we use `'false'` and not the boolean `False` as the value, for example.)  When invoking `setconfig`, we are in a bash context. I think it is a lot more consistent to write what is passed to `setconfig` here.

PR #9623, opened 2025-03-10
lnworker: pass PaySession to route building methods.

accumulator:
Simplifies method signatures and groups quite a large set of parameters that are mostly passed down along the call stack but functionally belong together. Also removes bolt11 specific parameters in method signatures, e.g. r_tags, which are not relevant in e.g. a bolt12 context.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -3,6 +3,7 @@
 # file LICENCE or http://www.opensource.org/licenses/mit-license.php
 
 import asyncio
+import copy
 import os
 from decimal import Decimal
 import random
@@ -680,6 +681,8 @@ class LNGossip(LNWorker):
 
 
 class PaySession(Logger):
+    __immutable = False
+
     def __init__(
             self,
             *,
@@ -719,6 +722,17 @@ class PaySession(Logger):
         self._amount_inflight = 0  # what we sent in htlcs (that receiver gets, without fees)
         self._nhtlcs_inflight = 0
         self.is_active = True  # is still trying to send new htlcs?
+
+    def immutable(self):
+        dc = copy.copy(self)
+        dc.sent_htlcs_q = None
+        dc.__immutable = True
+        return dc
+
+    def __setattr__(self, name, value):
+        if self.__immutable:
+            raise AttributeError(f"{name} is immutable")
+        return super().__setattr__(name, value)
 
     def diagnostic_name(self):
         pkey = sha256(self.payment_key)
@@ -1618,7 +1632,7 @@ class LNWallet(LNWorker):
                     # graph updates might occur during the computation
                     remaining_fee_budget_msat = (budget.fee_msat * amount_to_send) // amount_to_pay
                     routes = self.create_routes_for_payment(
-                        paysession=paysession,
+                        paysession=paysession.immutable(),
                         amount_msat=amount_to_send,
                         full_path=full_path,
                         fwd_trampoline_onion=fwd_trampoline_onion,
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1897,10 +1897,8 @@ class LNWallet(LNWorker):
         self,
         *,
         amount_msat: int,
-        final_total_msat: int,
         my_active_channels: Sequence[Channel],
-        invoice_features: LnFeatures,
-        r_tags,
+        paysession: PaySession,
     ) -> List['SplitConfigRating']:
         channels_with_funds = {
             (chan.channel_id, chan.node_id): int(chan.available_to_spend(HTLCOwner.LOCAL))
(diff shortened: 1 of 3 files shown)


electrum/trampoline.py
```diff
@@ -197,19 +197,13 @@ def _choose_second_trampoline(
 def create_trampoline_route(
         *,
         amount_msat: int,
-        min_final_cltv_delta: int,
-        invoice_pubkey: bytes,
-        invoice_features: int,
+        paysession: PaySession,
```
SomberNight, line 200: You don't need to move the whole PaySession class just to be able to type-hint it here.

You can just use the `if TYPE_CHECKING:` construct, import PaySession there, and then type-hint the function param as a string. E.g.:

```diff
diff --git a/electrum/trampoline.py b/electrum/trampoline.py
index 04777406d5..182b184d8f 100644
--- a/electrum/trampoline.py
+++ b/electrum/trampoline.py
@@ -1,7 +1,7 @@
 import io
 import os
 import random
-from typing import Mapping, DefaultDict, Tuple, Optional, Dict, List, Iterable, Sequence, Set, Any
+from typing import Mapping, DefaultDict, Tuple, Optional, Dict, List, Iterable, Sequence, Set, Any, TYPE_CHECKING
 
 from .lnutil import LnFeatures, PaymentFeeBudget, FeeBudgetExceeded
 from .lnonion import calc_hops_data_for_payment, new_onion_packet, OnionPacket
@@ -11,6 +11,9 @@ from .lntransport import LNPeerAddr
 from . import constants
 from .logging import get_logger
 
+if TYPE_CHECKING:
+    from .lnworker import PaySession
+
 
 _logger = get_logger(__name__)
 
@@ -197,6 +200,7 @@ def _choose_second_trampoline(
 def create_trampoline_route(
         *,
         amount_msat: int,
+        paysession: 'PaySession',
         min_final_cltv_delta: int,
         invoice_pubkey: bytes,
         invoice_features: int,

```

electrum/lnworker.py
```diff
@@ -2062,10 +2049,7 @@ async def create_routes_for_payment(
                                 partial(
                                     self.create_route_for_single_htlc,
                                     amount_msat=part_amount_msat,
-                                    invoice_pubkey=paysession.invoice_pubkey,
-                                    min_final_cltv_delta=paysession.min_final_cltv_delta,
-                                    r_tags=paysession.r_tags,
-                                    invoice_features=paysession.invoice_features,
+                                    paysession=paysession,
```
ecdsa, line 2052: Note that the fields of `PaySession` are mutable.  We are passing them to a function that is run asynchronously (`await run_in_thread`). This happens to be fine in that case, because the fields we are passing (`invoice_pubkey, invoice, r_tags, min_final_cltv_delta`) happen to not be mutated during the lifetime of the session. However, this seems to be lucky. What would happen if we had been passing `paysession._amount_inflight,` or some other variable field? Then the value at the time it is evaluated in the coroutine might differ from the value passed as parameter.

accumulator: We could use [dataclasses](https://docs.python.org/3/library/dataclasses.html) for `PaySession` and force immutability using the `frozen` flag

accumulator: `NamedTuple` seems to be too limited in this case AFAICS

accumulator: We could store the relevant (immutable) fields in a `NamedTuple` within `PaySession` and pass that instead. This way there is no access to the mutable fields of `PaySession` in the routing methods

SomberNight: > Note that the fields of PaySession are mutable

Good point.

> store the relevant (immutable) fields in a NamedTuple within PaySession

(btw besides NamedTuple, there is also `@dataclass(frozen=True)` in the [stdlib](https://docs.python.org/3/library/dataclasses.html). It is similar to attrs, a tiny bit less flexible, but attrs I would eventually like to remove from the dependencies)

accumulator: > (btw besides NamedTuple, there is also `@dataclass(frozen=True)` in the [stdlib](https://docs.python.org/3/library/dataclasses.html). It is similar to attrs, a tiny bit less flexible, but attrs I would eventually like to remove from the dependencies)

Yes I have looked at [dataclasses](https://docs.python.org/3/library/dataclasses.html), but this is too inflexible, as there are mutable fields in `PaySession` that should remain mutable, and I haven't seen an option to selectively un-immute (heh) a subset of fields. Subclassing also doesn't allow mutability as the restriction propagates down the hierarchy.

accumulator: I added a method to return an immutable copy and pass that instead. See https://github.com/spesmilo/electrum/pull/9623/commits/dc1e5a024432325e583077b81553e964d1940435

There's probably nicer ways to do this, e.g. externalize the `__setattr__` override

accumulator: Hmm.. I'm not really happy with the last commit. If we copy the PaySession it does not really need to be immutable, and the added code is.. not very nice.

Maybe it is nicer to add a `NamedTuple` within `PaySession`, and just pass that immutable tuple down to the routing functions.

ecdsa: indeed I would prefer not to pass `paysession` to that method, but a namedtuple that can be a field of `paysession`.

The other call (`create_trampoline_route_and_onion`) needs to pass a mutable field (`failed_trampoline_routes`) in addition to immutable fields. That mutable field should be passed explicitly, in addition to the namedtuple.

accumulator: as I'm reworking the patch to pass a `NamedTuple`, it looks like there are too many exceptions to really benefit from this encapsulation :(

passing the `NamedTuple` and extra fields of `PaySession` along is not really an improvement of the old situation IMO.

PR #9628, opened 2025-03-12
Fail incoming htlcs if we forwarded and chan was force-closed (fix #8547)

### LINKED ISSUE #8547: LN forwarding: if forwarded htlc expires on outgoing channel, after timing out htlc onchain, we should send update_fail_htlc on incoming channel. to avoid cascading channel closes
Consider an A->B->C forwarding, from the PoV of B.

The CLTV at the B->C link is e.g. at height 100, and the CLTV at the A->B link is at height 200. At block 101 we have B force-close the B->C channel if C has not timed out the htlc yet (see logic in [`lnchannel.should_be_closed_due_to_expiring_htlcs`](https://github.com/spesmilo/electrum/blob/0e5cb194085297421cfc3bc70dcb2b70bf8999e8/electrum/lnchannel.py#L1649-L1650)).

The risk of B not doing anything is that the height of 200 is approaching, at which point B will not be able to pull funds from A (as either B is supposed to remove the htlc on A->B, or A will force-close). If B cannot pull the funds from A, and the B->C htlc is still lingering, and then C fulfills that, then B sent out coins, got the preimage, but the preimage is not useful anymore.

---

The problem is that B force-closes the B->C channel at height=101, but then B will not send update_fail_htlc to the A->B channel.
Specifically, if the htlc on the B->C channel expires, B should force-close that channel, and reclaim the funds in the htlc by timing it out on-chain. After the htlc timeout is mined, B should send update_fail_htlc to Alice for the A->B channel.

Instead, I think with the current code, B will not send update_fail_htlc to the A->B channel, and then at block 201, A will force-close the A->B channel. That is, if there is a forwarding with k nodes: N_1 -> N_2 -> ... -> N_k, and N_k becomes unresponsive, all the k-1 channels might eventually get force-closed (assuming they all run the current code), instead of only the N_(k-1) -> N_k channel getting closed and the update_fail_htlc message propagating back on the rest.

---

Note that after B force-closes the B->C channel after the htlc is expired, I said B would time out the htlc on-chain -- however C might also come back online at this point and use the preimage to fulfill the htlc on-chain. In which case B can use the preimage and fulfill back to Alice. This scenario is already implemented in `lnchannel.extract_preimage_from_htlc_txin`.

ecdsa:
call lnworker.htlc_failed once the htlc_tx is deeply mined.

In the case of a forwarding, this will fail incoming htlcs. (fixes #8547)

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -321,7 +321,7 @@ class AbstractChannel(Logger, ABC):
     def maybe_sweep_htlcs(self, ctx: Transaction, htlc_tx: Transaction) -> Dict[str, SweepInfo]:
         return {}
 
-    def extract_preimage_from_htlc_txin(self, txin: TxInput) -> None:
+    def extract_preimage_from_htlc_txin(self, txin: TxInput, *, is_deeply_mined: bool) -> None:
         return
 
     def update_onchain_state(self, *, funding_txid: str, funding_height: TxMinedInfo,
@@ -658,7 +658,7 @@ class ChannelBackup(AbstractChannel):
     def maybe_sweep_htlcs(self, ctx: Transaction, htlc_tx: Transaction) -> Dict[str, SweepInfo]:
         return {}
 
-    def extract_preimage_from_htlc_txin(self, txin: TxInput) -> None:
+    def extract_preimage_from_htlc_txin(self, txin: TxInput, *, is_deeply_mined: bool) -> None:
         return None
 
     def get_funding_address(self):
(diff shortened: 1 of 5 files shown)


electrum/lnchannel.py
```diff
@@ -1350,42 +1350,75 @@ def receive_revocation(self, revocation: RevokeAndAck):
                     error_bytes, failure_message = None, None
                 self.lnworker.htlc_failed(self, htlc.payment_hash, htlc.htlc_id, error_bytes, failure_message)
 
-    def extract_preimage_from_htlc_txin(self, txin: TxInput) -> None:
+
+    def extract_preimage_from_htlc_txin(self, txin: TxInput, is_confirmed: bool) -> None:
```
SomberNight, line 1354: can we have a more descriptive name than `is_confirmed`? e.g. `is_deeply_mined` or `is_deep`?
also, make it kwarg-only
```suggestion
    def extract_preimage_from_htlc_txin(self, txin: TxInput, *, is_deeply_mined: bool) -> None:
```

electrum/lnwatcher.py
```diff
@@ -374,3 +369,10 @@ async def maybe_redeem(self, prevout, sweep_info: 'SweepInfo', name: str) -> Non
             if old_tx and old_tx.txid() != new_tx.txid():
                 self.lnworker.wallet.set_label(old_tx.txid(), None)
             util.trigger_callback('wallet_updated', self.lnworker.wallet)
+
+    def maybe_extract_preimage(self, chan, spender_tx, prevout):
```
SomberNight, line 373: pretty please add type hints for function args and return values for new code
```patch
diff --git a/electrum/lnwatcher.py b/electrum/lnwatcher.py
index c8e5f8e2d6..5d72063805 100644
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -22,6 +22,7 @@ if TYPE_CHECKING:
     from .network import Network
     from .lnsweep import SweepInfo
     from .lnworker import LNWallet
+    from .lnchannel import AbstractChannel
 
 
 class TxMinedDepth(IntEnum):
@@ -370,7 +371,7 @@ class LNWalletWatcher(LNWatcher):
                 self.lnworker.wallet.set_label(old_tx.txid(), None)
             util.trigger_callback('wallet_updated', self.lnworker.wallet)
 
-    def maybe_extract_preimage(self, chan, spender_tx, prevout):
+    def maybe_extract_preimage(self, chan: 'AbstractChannel', spender_tx: Transaction, prevout: str):
         txin_idx = spender_tx.get_input_idx_that_spent_prevout(TxOutpoint.from_str(prevout))
         assert txin_idx is not None
         spender_txin = spender_tx.inputs()[txin_idx]

```

electrum/lnutil.py
```diff
@@ -858,67 +858,88 @@ def make_received_htlc(
...
+        opcodes.OP_1,
+        opcodes.OP_CHECKSEQUENCEVERIFY,
+        opcodes.OP_DROP
+    ] if anchors else [
+    ]) + [
+        opcodes.OP_ENDIF
+    ]
+
+WITNESS_TEMPLATE_RECEIVED_HTLC = witness_template_received_htlc(anchors=False)
+WITNESS_TEMPLATE_RECEIVED_HTLC_ANCHORS = witness_template_received_htlc(anchors=True)
```
SomberNight, line 942: I have extracted this into a separate commit and already pushed it to master, in https://github.com/spesmilo/electrum/commit/77c180f793831b3f056bbd30e64c6d969ef1bd5a

electrum/lnchannel.py
```diff
@@ -1350,42 +1350,75 @@ def receive_revocation(self, revocation: RevokeAndAck):
...
+        # is_confirmed is True if tx was deeply mined (100 blocks)
+        # Note: this is probably excessive, given that our default
+        # CLTV delta is 144, and the remote might not always be online
```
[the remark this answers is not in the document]
ecdsa: decreased to 20, as discussed

electrum/lnchannel.py
ecdsa: done

PR #9629, opened 2025-03-13
Allow plugins to register CLI commands 

### LINKED ISSUE #1814: Plugins that add command-line commands
I'd like to make a plugin that defines a new command-line command. At this point, that doesn't seem possible, and existing plugins are only adding support for hardware wallets so they can interact with standard command-line commands.

Challenges include:
- plugins are not initialized before `lib.commands.get_parser` is called, so they are not listed in `known_commands` when the [command-line parser is constructed](https://github.com/spesmilo/electrum/blob/master/lib/commands.py#L780).
- not clear if we are allowed to use `lib.commands.command` decorator outside of that file.
- there is an `unknown command:` error message hacked into URI handling for the GUI: [electrum:310](https://github.com/spesmilo/electrum/blob/master/electrum#L310)
- when the command arguments are parsed, we do not yet know if starting the GUI or running in command-line mode
- the configuration is not fully-defined while parsing arguments, so not clear if a plugin is enabled.

My proposed changes (which I would make and submit as a PR for further review):

step 1:
- define a custom `argparse.Action()` handler for `url` argument so it's only set if starts with `bitcoin:`
- remove/replace the "unknown command" error message associated with URI
- do not start the GUI if unparsed command line remains; print error instead (ie. unknown command)
- AFAIK this will preserve the existing behaviour of entering the GUI on:  `electrum bitcoin:1foo` 

step 2A:
- add a new "type" of plugin that is used only to define command line stuff: perhaps `argparser`
- like `qt`, `cmdline` are now, plugins may define `plugins/FOO/argparser.py` and it gets hooked into the argument processing stuff, based on the `available_for` value defined in `plugins/FOO/__init__.py`
- plugins of the new type would be loaded very early before the command line is parsed
- a new base class would be used to expose an interface (wouldn't use BasePlugin)
- they can define new commands and existing optparser code would detect them in `knowncommands`

**OR** a different approach: step 2B:
- new plugin type, perhaps called `early`
- this type of plugin gets loaded very early, before the command-line parser is setup
- add some new hooks for updating the arg parser (maybe)
- new plugin type could do almost anything, but I would use it just to register new commands into `known_commands`

From my point of view, approach 2A is better, but 2B is more general purpose and flexible. The only real difference is that 2A is "more objected oriented".

What do you think of these changes? Other ideas?

f321x:
This PR adds functionality which allows plugins to register cli commands. 
CLI commands can be registered from within a `commands.py` file at the plugins root by decorating them with the `@plugin_command(requirements, plugin_name)` decorator.

Example:
```python
file: plugins/demo_plugin/commands.py

from electrum.commands import plugin_command

plugin_name = "demo_plugin"

@plugin_command("", plugin_name=plugin_name)
async def plugin_demo_command(self: 'Commands'):
    plugin: 'DemoPlugin' = self.daemon._plugins.get_plugin(plugin_name)
    plugin.do_something()
    return "plugin did something (this is the cli output)"
```

This works for internal and external plugins. Plugin commands have to begin with `plugin_` and are not allowed to overwrite other commands. 

Here is a external demo plugin which registers 3 commands for testing: 
[nwc-0.0.1.zip](https://github.com/user-attachments/files/19228224/nwc-0.0.1.zip)
to be enabled with `setconfig enable_plugin_nwc true`. See the commands with `help | grep nwc`.

closes #1814

### DIFF
--- a/electrum/plugin.py
+++ b/electrum/plugin.py
@@ -65,19 +65,20 @@ class Plugins(DaemonThread):
     pkgpath = os.path.dirname(plugins.__file__)
 
     @profiler
-    def __init__(self, config: SimpleConfig = None, gui_name = None, cmd_only: bool = False):
+    def __init__(self, config: SimpleConfig, gui_name = None, cmd_only: bool = False):
+        self.config = config
         self.cmd_only = cmd_only  # type: bool
         self.internal_plugin_metadata = {}
         self.external_plugin_metadata = {}
         self.loaded_command_modules = set()  # type: set[str]
         if cmd_only:
             # only import the command modules of plugins
+            Logger.__init__(self)
             self.find_plugins()
             return
         DaemonThread.__init__(self)
         self.device_manager = DeviceMgr(config)
         self.name = 'Plugins'  # set name of thread
-        self.config = config
         self.hw_wallets = {}
         self.plugins = {}  # type: Dict[str, BasePlugin]
         self.gui_name = gui_name
@@ -85,15 +86,6 @@ class Plugins(DaemonThread):
         self.load_plugins()
         self.add_jobs(self.device_manager.thread_jobs())
         self.start()
-
-    def __getattr__(self, item):
-        # to prevent accessing of a cmd_only instance of this class
-        if self.cmd_only:
-            if item == 'logger':
-                # if something tries to access logger and it is not initialized it gets initialized here
-                Logger.__init__(self)
-                return self.logger
-            raise Exception(f"This instance of Plugins is only for command importing, cannot access {item}")
 
     @property
     def descriptions(self):
@@ -107,6 +99,8 @@ class Plugins(DaemonThread):
             #       once as data and once as code. To honor the "no duplicates" rule below,
             #       we exclude the ones packaged as *code*, here:
             if loader.__class__.__qualname__ == "PyiFrozenImporter":
+                continue
+            if self.cmd_only and self.config.get('enable_plugin_' + name) is not True:
                 continue
             full_name = f'electrum.plugins.{name}' + ('.commands' if self.cmd_only else '')
             spec = importlib.util.find_spec(full_name)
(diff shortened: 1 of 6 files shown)


electrum/plugin.py
```diff
@@ -65,41 +65,60 @@ class Plugins(DaemonThread):
...
             #       we exclude the ones packaged as *code*, here:
             if loader.__class__.__qualname__ == "PyiFrozenImporter":
                 continue
-            full_name = f'electrum.plugins.{name}'
+            full_name = f'electrum.plugins.{name}' + ('.commands' if self.cmd_only else '')
             spec = importlib.util.find_spec(full_name)
-            if spec is None:  # pkgutil found it but importlib can't ?!
+            if spec is None:
+                if self.cmd_only:
+                    continue # no commands module in this plugin
```
SomberNight, line 115: Could you add a command to one of the existing plugins to showcase this functionality?
For example, for the `labels` plugin a force-pull command, similar to the GUI buttons.

See
https://github.com/spesmilo/electrum/blob/e62a2c43c5e1e9769e40ea0fcc2591f5c48166a4/electrum/plugins/labels/qt.py#L51
https://github.com/spesmilo/electrum/blob/e62a2c43c5e1e9769e40ea0fcc2591f5c48166a4/electrum/plugins/labels/labels.py#L180

ecdsa:
I added two commits to your branch:
 - the first commit adds a method that parses the command line.
 - the second commit parses the command line twice: first in order to figure out which plugins are enabled, then once plugin commands have been added to the list of known commands.
That way, we do not need to import command files of plugins that are not enabled, and to remove them later.

For clarity I think we should enforce that all plugin commands start with the plugin name followed by an underscore:
```
./run_electrum nwc_list_connections
./run_electrum nwc_remove_connection
./run_electrum nwc_get_connection_string
```
I think `argparse` even allows subcommands, so we could use a space:
```
./run_electrum nwc list_connections
./run_electrum nwc remove_connection
./run_electrum nwc get_connection_string
```

electrum/plugin.py
ecdsa: done

PR #9634, opened 2025-03-13
wallet: consider "future" coins as frozen by default

SomberNight:
alternative to https://github.com/spesmilo/electrum/pull/9630

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -1966,6 +1966,9 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         if frozen is not None:  # user has explicitly set the state
             return bool(frozen)
         # State not set. We implicitly mark certain coins as frozen:
+        tx_mined_status = self.adb.get_tx_height(utxo.prevout.txid.hex())
+        if tx_mined_status.height == TX_HEIGHT_FUTURE:
+            return True
         if self._is_coin_small_and_unconfirmed(utxo):
             return True
         return False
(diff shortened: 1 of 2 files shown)


electrum/wallet.py
```diff
@@ -1964,8 +1964,14 @@ def is_frozen_coin(self, utxo: PartialTxInput) -> bool:
         #       True/False if the user explicitly set it,
         #       None otherwise
         if frozen is None:
```
SomberNight, line 1966: note: if the user explicitly sets the frozen state via the GUI, whatever they set is respected. Also, the GUI does not let the user transition back to the `frozen is None` state (the GUI only allows setting `frozen=True` or `frozen=False`).

I've pushed a separate commit to help testing: https://github.com/spesmilo/electrum/commit/8ccd31fe4952a77f63e506b740c1b32e56d1db60 (on master).
So now using the Console, one can manually transition back to `frozen=None`, e.g.:
```
wallet.set_frozen_state_of_coins(["f4f1bfb718c8f3d94b0af6a6cc8d03cb2c6e7738bc3c8aba70498b9d055128e0:0"], None)
```

ecdsa:
I think tri-state variables are bad UX. the 3 states are not shown to the user.

SomberNight:
@f321x noted that one remaining issue is that as the future tx status is not persisted, if you e.g. start electrum while offline, and try to create a tx, the default frozen state for the future coins will be false (as we don't know they are future txs, instead we treat them as "local"). If you then come online, they will correctly get marked as future, and frozen by default. The solution to this would be to persist the future vs local tx distinction. If we want that, it can be done separately later.

> I think tri-state variables are bad UX. the 3 states are not shown to the user.

I think exposing the tri-state in the GUI is somewhat overkill, yes. However internally I argue we want it to be tri-state: forced-off, forced-on, and "application can decide". I see your point that a user manually freezing and then unfreezing would expect to have undone their action and to have returned to the original state - which is not the case now, however I find both alternatives unsatisfactory (making it a two-state everywhere, or fully exposing the tri-state in the GUI).
I think the current status quo is fine.

Issue #7497, opened 2021-09-13
add option to avoid forced address reuse (e.g. dust attack)


SomberNight:
We could add an option to auto-freeze UTXOs that belong to addresses from which we have already spent.

Bitcoin Core wallet and JoinMarket already has such an option, see references below.

-----

- https://github.com/bitcoin/bitcoin/pull/13756
    > A new wallet flag `avoid_reuse` has been added (default off). When enabled, a wallet will distinguish between used and unused addresses, and default to not use the former in coin selection. [...] Together with "avoid partial spends" (added in Bitcoin Core v0.17.0), this can eliminate a serious privacy issue where a malicious user can track spends by sending small payments to a previously-paid address that would then be included with unrelated inputs in future payments. (#13756)

- https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/release-notes/release-notes-0.6.2.md#auto-freeze-of-deposits-to-reused-addresses
    > Auto-freeze of deposits to reused addresses.
    >
    > New deposits should always be to unused addresses, so to prevent [forced address reuse attacks](https://en.bitcoin.it/wiki/Privacy#Forced_address_reuse), (sometimes called dust attacks, but that's a poor name), we freeze any deposit to an already-used address by default.
    > 
    > The user can easily simple unfreeze these utxos both on CLI and in the Qt GUI (see the "Coins" tab; right click a utxo to unfreeze it). Also, the user can set a finite threshold above which such deposits will *not* be frozen by setting the value of `max_sats_freeze_reuse` in the `[POLICY]` section to a positive integer (`-1` means no limit).

ecdsa:
I agree, we should have such an option.
I think that "forced address reuse" is a poor name too.

PR #9636, opened 2025-03-14
wallet: add new config option "FREEZE_REUSED_ADDRESS_UTXOS"

SomberNight:
Adds a new config option: `WALLET_FREEZE_REUSED_ADDRESS_UTXOS`. This is based on [Bitcoin Core's "avoid_reuse"](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.19.0.1.md#wallet) wallet flag.

This opt-in feature, if enabled:
> Automatically freeze coins received to already used addresses.
> This can eliminate a serious privacy issue where a malicious user can track your spends by sending small payments
> to a previously-paid address of yours that would then be included with unrelated inputs in your future payments.

Note that currently we only have a single coinchooser policy, `CoinChooserPrivacy`, which interacts well with this option, as it spends all coins from any selected address. However, if we later add a different coinchooser policy, which allowed "partial spends", care should be taken re e.g. disallowing using that when this option is set.

Also note that this PR adds this as a config option, but arguably it could be wallet-specific instead, such as `use_change`.

closes https://github.com/spesmilo/electrum/issues/7497

### DIFF
--- a/electrum/gui/qt/confirm_tx_dialog.py
+++ b/electrum/gui/qt/confirm_tx_dialog.py
@@ -426,6 +426,7 @@ class TxEditor(WindowModalDialog):
         add_cv_action(self.config.cv.WALLET_MERGE_DUPLICATE_OUTPUTS, self.toggle_merge_duplicate_outputs)
         add_cv_action(self.config.cv.WALLET_SPEND_CONFIRMED_ONLY, self.toggle_confirmed_only)
         add_cv_action(self.config.cv.WALLET_COIN_CHOOSER_OUTPUT_ROUNDING, self.toggle_output_rounding)
+        add_cv_action(self.config.cv.WALLET_FREEZE_REUSED_ADDRESS_UTXOS, self.toggle_freeze_reused_address_utxos)
         self.pref_button = QToolButton()
         self.pref_button.setIcon(read_QIcon("preferences.png"))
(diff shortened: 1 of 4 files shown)


SomberNight:
btw I can write a unit test for this, just did not want to spend more time on it in case there was opposition to the feature itself :D

electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -426,6 +426,7 @@ def add_cv_action(configvar: 'ConfigVarWithConfig', action: Callable[[], None]):
         add_cv_action(self.config.cv.WALLET_MERGE_DUPLICATE_OUTPUTS, self.toggle_merge_duplicate_outputs)
         add_cv_action(self.config.cv.WALLET_SPEND_CONFIRMED_ONLY, self.toggle_confirmed_only)
         add_cv_action(self.config.cv.WALLET_COIN_CHOOSER_OUTPUT_ROUNDING, self.toggle_output_rounding)
+        add_cv_action(self.config.cv.WALLET_FREEZE_REUSED_ADDRESS_UTXOS, self.toggle_freeze_reused_address_utxos)
```
SomberNight, line 429: Unclear if the TxEditor is the best place to put this.
We could put it into the utxo_list toolbar menu instead.
(or both)

PR #9640, opened 2025-03-14
swaps: Update submarine swap nostr relays dynamically and remove redundant query

f321x:
Dynamically update the nostr relays used for submarine swaps so the client connects to relays provided by the swap provider. This allows the client to use a swap provider even if they have non overlapping relays and the client just got the announcement of the provider because it spreaded to the clients relay.
Also removes redundant query (`get_pairs()` and renames previous `get_offers()` to `get_pairs()`) when initializing `NostrTransport`, there is no need to query all announcements and additionally query the announcement from a specific npub, we can just use the announcement returned by the broad query we run anyways.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -202,6 +202,7 @@ class SwapManager(Logger):
                 self.prepayments[swap.prepay_hash] = bytes.fromhex(k)
         self.is_server = False # overriden by swapserver plugin if enabled
         self.is_initialized = asyncio.Event()
+        self.pairs_updated = asyncio.Event()
 
     def start_network(self, network: 'Network'):
         assert network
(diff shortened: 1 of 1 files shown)


hany246:
0

electrum/submarine_swaps.py
```diff
@@ -1543,5 +1548,30 @@ async def handle_request(self, request):
         else:
             raise Exception(method)
         r['reply_to'] = event_id
-        self.logger.info(f'sending response id={event_id}')
-        await self.send_direct_message(event_pubkey, relays, json.dumps(r))
+        self.logger.debug(f'sending response id={event_id}')
+        await self.send_direct_message(event_pubkey, json.dumps(r))
+
+    def _store_last_swapserver_relays(self, relays: Sequence[str]):
```
ecdsa, line 1554: this could use util.read_json_file, write_json_file

ecdsa: oops, sorry, I forgot that these utility functions call MyEncoder. Please revert. my bad

f321x: Why is it an issue that write_json_file calls the MyEncoder, doesn't it just default back to the default JsonEncoder if the passed content is no special type? It seems to work for the relay strs.

ecdsa: Not really an issue, but this is adding unneeded complexity.  The point of this util function is to use `MyEncoder`. 
I think it makes the code more obscure if we use it where it is not needed.

f321x: Ok, reverted the functions back to not using the util functions, 21e3fd91ddf2866b100b7bd0fbea01b3d287b0d7

braicool (approved):

braicool (approved):

PR #9647, opened 2025-03-17
ln_utxo_reserve

ecdsa:
When we send max, decrease sent amount in order to keep some reserve utxo, in order to be able to sweep lightning channels.


electrum/wallet.py
```diff
@@ -1845,6 +1845,17 @@ def get_new_sweep_address_for_channel(self) -> str:
         assert is_address(selected_addr), f"not valid bitcoin address: {selected_addr}"
         return selected_addr
 
+    def should_keep_reserve_utxo(self, to_be_sent: int, is_channel_opening: bool) -> bool:
```
SomberNight, line 1848: we should try to suffix amounts in function params with the unit:
```suggestion
    def should_keep_reserve_utxo(self, to_be_sent_sat: int, is_channel_opening: bool) -> bool:
```

electrum/wallet.py
```diff
@@ -1845,6 +1845,17 @@ def get_new_sweep_address_for_channel(self) -> str:
...
+        if self.lnworker and self.config.ENABLE_ANCHOR_CHANNELS:
+            available = sum(map(lambda c: c.value_sats(), self.get_spendable_coins()))
+            assert to_be_sent <= available
+            if to_be_sent < available:
+                # coin control was used
+                return False
```
SomberNight, line 1854: what if the total value of the unselected coins is below the reserve?
e.g. only a 3000 sat utxo remains

electrum/wallet.py
```diff
@@ -1961,6 +1973,17 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
             # note: Actually, it might be the case that not all UTXOs from the wallet are
             #       being spent if the user manually selected UTXOs.
             sendable = sum(map(lambda c: c.value_sats(), coins))
+            if self.should_keep_reserve_utxo(sendable, is_channel_opening):
```
SomberNight, line 1976: what about non-"spend max" branch?
the user might try to send an amount very close to the available balance, where the only remaining utxo after the tx is the change output, which is below the reserve

electrum/wallet.py
```diff
@@ -1845,6 +1845,17 @@ def get_new_sweep_address_for_channel(self) -> str:
         assert is_address(selected_addr), f"not valid bitcoin address: {selected_addr}"
         return selected_addr
 
+    def should_keep_reserve_utxo(self, to_be_sent: int, is_channel_opening: bool) -> bool:
+        if self.lnworker and self.config.ENABLE_ANCHOR_CHANNELS:
```
SomberNight, line 1849: note that the config var config.ENABLE_ANCHOR_CHANNELS might get enabled then disabled, etc, by the user

electrum/wallet.py
```diff
@@ -1845,6 +1845,17 @@ def get_new_sweep_address_for_channel(self) -> str:
...
+            available = sum(map(lambda c: c.value_sats(), self.get_spendable_coins()))
+            assert to_be_sent <= available
+            if to_be_sent < available:
+                # coin control was used
+                return False
+            if is_channel_opening:
+                return True
```
SomberNight, line 1856: just because we have `config.ENABLE_ANCHOR_CHANNELS` enabled, does not mean the channel to be opened is going to use anchors.
the remote might not support anchors, in which case we open a SRK channel with them

electrum/wallet.py
```diff
@@ -1845,6 +1845,17 @@ def get_new_sweep_address_for_channel(self) -> str:
...
+            return any(not chan.is_redeemed() for chan in self.lnworker.channels.values())
```
SomberNight, line 1857: not sure if we want to filter to anchor channels

f321x:
What if the user tries to spend max to himself (manual coin consolidation), is it useful to create an reserve change output in this case?

ecdsa:
> What if the user tries to spend max to himself (manual utxo consolidation), is it useful to still create an reserve change output in this case?

Good point. We should not add change in the case, no

electrum/wallet.py
ecdsa: indeed. should_keep_reserve_utxo is now called in both branches of the if

electrum/wallet.py
```diff
@@ -1845,6 +1845,25 @@ def get_new_sweep_address_for_channel(self) -> str:
...
 
+    def should_keep_reserve_utxo(self, to_be_spent_sat: int, is_anchor_channel_opening: bool) -> bool:
+        if not self.lnworker:
+            return False
+        is_reserve_needed = any(chan.has_anchors() and not chan.is_redeemed() for chan in self.lnworker.channels.values())
+        is_reserve_needed |= is_anchor_channel_opening
+        if not is_reserve_needed:
+            return False
+        available = sum(map(lambda c: c.value_sats(), self.get_spendable_coins()))
+        assert available - to_be_spent_sat >= 0
```
SomberNight, line 1856: is this assert sound?
e.g. consider this: wallet has 1 confirmed utxo (1 btc) and 1 unconfirmed utxo (1 btc). config has WALLET_SPEND_CONFIRMED_ONLY set to true. User uses manual coin selection to select both utxos, and clicks spend max when trying to open the channel. Then `to_be_spent_sat` I guess is 2 btc, and available is 1 btc.

so maybe call `wallet.get_spendable_coins()` with explicit `nonlocal_only=False`, `confirmed_only=False`

electrum/wallet.py
```diff
@@ -1960,22 +1988,39 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
...
+            distribute_amount(to_distribute)
+
+            amount_retained = sum(o.value for o in outputs if self.is_mine(o.address))
+            to_be_spent = input_amount - amount_retained # this includes the fee
+
+            if self.should_keep_reserve_utxo(to_be_spent, is_anchor_channel_opening):
+                self.logger.info(f'Adding change output to meet utxo reserve requirements')
+                change_addr = self.get_change_addresses_for_new_transaction(change_addr)[0]
+                change = PartialTxOutput.from_address_and_value(change_addr, self.config.LN_UTXO_RESERVE)
+                change.is_utxo_reserve = True # for GUI, using hasattr()
```
SomberNight, line 2019: why using hasattr? would it not be cleaner to just declare it in transaction.py?

```diff
diff --git a/electrum/transaction.py b/electrum/transaction.py
index 3abc4879db..78ab6c35ab 100644
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -1875,6 +1875,7 @@ class PartialTxOutput(TxOutput, PSBTSection):
         self._script_descriptor = None  # type: Optional[Descriptor]
         self.is_mine = False  # type: bool  # whether the wallet considers the output to be ismine
         self.is_change = False  # type: bool  # whether the wallet considers the output to be change
+        self.is_utxo_reserve = False  # type: bool  # whether the wallet added this as anchor reserve
 
     @property
     def pubkeys(self) -> Set[bytes]:

```

electrum/wallet.py
```diff
@@ -1960,22 +1988,39 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
...
+            input_amount = sum(c.value_sats() for c in coins)
+            allocated_amount = sum(o.value for o in outputs if not parse_max_spend(o.value))
+            to_distribute = input_amount - allocated_amount - fee
+            distribute_amount(to_distribute)
+
+            amount_retained = sum(o.value for o in outputs if self.is_mine(o.address))
+            to_be_spent = input_amount - amount_retained # this includes the fee
+
+            if self.should_keep_reserve_utxo(to_be_spent, is_anchor_channel_opening):
+                self.logger.info(f'Adding change output to meet utxo reserve requirements')
```
SomberNight, line 2016: could you add a unit test that enters this branch?
just lie about `is_anchor_channel_opening=True` and mimic one of the existing tests in test_wallet_vertical

electrum/wallet.py
```diff
@@ -1952,6 +1972,14 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
                     amount = change[0].value
                     if amount <= self.lnworker.num_sats_can_receive():
                         tx.replace_output_address(change[0].address, DummyAddress.SWAP)
+
+            self.logger.debug(f'coinchooser returned tx with {len(tx.inputs())} inputs and {len(tx.outputs())} outputs')
+            to_be_consumed = sum(c.value_sats() for c in tx.inputs() if c in coins)
+            amount_retained = sum(o.value for o in tx.outputs() if self.is_mine(o.address))
+            to_be_spent = to_be_consumed - amount_retained
+            if to_be_spent > 0 and self.should_keep_reserve_utxo(to_be_spent, is_anchor_channel_opening):
+                raise NotEnoughFunds()
```
SomberNight, line 1981: could you add a unit test that enters this branch?

electrum/wallet.py
```diff
@@ -1952,6 +1972,14 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
                     amount = change[0].value
                     if amount <= self.lnworker.num_sats_can_receive():
                         tx.replace_output_address(change[0].address, DummyAddress.SWAP)
+
+            self.logger.debug(f'coinchooser returned tx with {len(tx.inputs())} inputs and {len(tx.outputs())} outputs')
+            to_be_consumed = sum(c.value_sats() for c in tx.inputs() if c in coins)
```
SomberNight, line 1977: why is this only looking at `coins`? what about `inputs`?

electrum/wallet.py
```diff
@@ -1960,22 +1988,39 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
...
+                for (weight, i) in i_max:
+                    # fixme: this does not check that value >= dust_threshold
+                    val = int((amount/i_max_sum) * weight)
+                    outputs[i].value = val
+                    distr_amount += val
+                (x,i) = i_max[-1]
+                outputs[i].value += (amount - distr_amount)
+
+            distribute_amount(0)
             tx = PartialTransaction.from_io(list(coins), list(outputs))
```
SomberNight, line 2004: pre-existing bug on master from https://github.com/spesmilo/electrum/commit/8bec974a39a7692475db85b94c227b21a57bab31:
what about `inputs`? should those not be included in the spend max branch? unit test needed :)

ecdsa: Good point. There is currently no exec path where inputs are passed to `make_unsigned_transaction` and we spend max. Maybe we can just add an `assert inputs == []` in the spend max branch?

electrum/wallet.py
ecdsa: the parameter `inputs` is used for sweeping coins that are not in the wallet. These inputs should not be added to `to_be_consumed`, because this would overestimate how much the transaction is spending. 

We could add an assert that inputs are not wallet coins.

electrum/wallet.py
SomberNight: That works as a stopgap, yes. Though it is logically valid to try to spend max with some fixed inputs, you could just raise an exception saying it is not yet implemented.

electrum/wallet.py
ecdsa: right. I refactored  `should_keep_reserve_utxo` so that it takes `tx_inputs` and `tx_outputs`, and the available amount considers all coins in the wallet.

electrum/wallet.py
ecdsa: done

electrum/wallet.py
ecdsa: I added new tests.

electrum/wallet.py
ecdsa: test added

electrum/wallet.py
ecdsa: ok, now using `inputs + coins`. the refactoring of `should_keep_reserve_utxo` makes it possible

electrum/gui/messages.py
```diff
@@ -74,3 +74,5 @@ def to_rtf(msg):
     _("If you go offline before broadcasting the claiming transaction and let the swap time out, "
       "you will not get back the already pre-paid mining fees.")
 )
+
+MSG_LN_UTXO_RESERVE = _("You do not have enough on-chain funds to protect your Ligtning channels. You should have at least {} on-chain in order to be able to sweep channel outputs.")
```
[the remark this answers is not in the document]
ecdsa: fixed

electrum/wallet.py
```diff
@@ -1845,6 +1845,25 @@ def get_new_sweep_address_for_channel(self) -> str:
...
+        is_reserve_needed = any(chan.has_anchors() and not chan.is_redeemed() for chan in self.lnworker.channels.values())
+        is_reserve_needed |= is_anchor_channel_opening
+        if not is_reserve_needed:
+            return False
+        available = sum(map(lambda c: c.value_sats(), self.get_spendable_coins()))
+        assert available - to_be_spent_sat >= 0
+        if available - to_be_spent_sat >= self.config.LN_UTXO_RESERVE:
+            # there will be enough remaining after we send
+            return False
+        # we will need to substract the reserve
```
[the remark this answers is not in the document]
ecdsa: fixed

electrum/wallet.py
```diff
@@ -1952,23 +1989,38 @@ def fee_estimator(size: Union[int, float, Decimal]) -> int:
...
+                    val = int((amount/i_max_sum) * weight)
+                    outputs[i].value = val
+                    distr_amount += val
+                (x,i) = i_max[-1]
+                outputs[i].value += (amount - distr_amount)
+
+            tx_inputs = inputs + coins # these do not overlap, see above
+            distribute_amount(0)
+            tx = PartialTransaction.from_io(list(tx_inputs), list(outputs))
             fee = fee_estimator(tx.estimated_size())
```
f321x, line 2007: Here the fee is calculated before potentially adding the reserve output, if the fee is already at the lower limit here (1sat/vb) and a reserve output is added below the resulting tx fee will not be updated and is be below the relay limit.

ecdsa: fixed

electrum/wallet.py
```diff
@@ -1837,6 +1837,37 @@ def get_new_sweep_address_for_channel(self) -> str:
         assert is_address(selected_addr), f"not valid bitcoin address: {selected_addr}"
         return selected_addr
 
+    def should_keep_reserve_utxo(
+            self,
+            tx_inputs: List[PartialTxInput],
+            tx_outputs: List[PartialTxOutput],
+            is_anchor_channel_opening: bool,
+    ) -> bool:
```
SomberNight, line 1845: I would like to suggest making most parameters in most functions kw-only in general, except maybe the first one or two parameters that are trivially deducible just from the function name.
```suggestion
    def should_keep_reserve_utxo(
            self,
            *,
            tx_inputs: List[PartialTxInput],
            tx_outputs: List[PartialTxOutput],
            is_anchor_channel_opening: bool,
    ) -> bool:
```

electrum/wallet.py
```diff
@@ -1837,6 +1837,37 @@ def get_new_sweep_address_for_channel(self) -> str:
...
+
+        coins_in_wallet = self.get_spendable_coins(nonlocal_only=False, confirmed_only=False)
+        prevout_coins_in_wallet = set(c.prevout for c in coins_in_wallet)
+        amount_in_wallet = sum(c.value_sats() for c in coins_in_wallet)
+
+        amount_consumed = sum(c.value_sats() for c in tx_inputs if c.prevout in prevout_coins_in_wallet)
+        amount_retained = sum(o.value for o in tx_outputs if self.is_mine(o.address))
+        to_be_spent_sat = amount_consumed - amount_retained
+
+        assert amount_in_wallet - to_be_spent_sat >= 0
```
SomberNight, line 1860: This assert can fail e.g. due to multi-threading races. Consider that
- thread1: make_unsigned_transaction gets called,
- thread2: spends some coins or we learn of a new outgoing tx from the network,
- thread1: make_unsigned_transaction reaches this assert but `coins_in_wallet` is now of much lesser value than at step1

Not sure how to practically fix. Just wanted to point out the conceptual issue.
If you added a code comment about this here with a fixme, I would say that's enough.

SomberNight (approved):
I've rechecked the base lib part and it looks ok.

electrum/wallet.py
ecdsa: sorry, didn't see this comment, you added it while I was merging. 
I guess the proper way to fix it is to add a proper locking mechanism on everything that affects available coins.
The scope of such a mechanism is potentially large.

PR #9648, opened 2025-03-17
config vars for plugins

ecdsa:
- probably needs a better way to pass default value and type
 - needs to set `SimpleConfig._init_done`:  we are currently passing the config object to the constructor of Plugins(). That might be easier if we passed Plugins to the constructor of SimpleConfig

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -814,12 +814,8 @@ Warning: setting this to too low will result in lots of payment failures."""),
 
     # connect to remote submarine swap server
     SWAPSERVER_URL = ConfigVar('swapserver_url', default='', type_=str)
-    # run submarine swap server locally
-    SWAPSERVER_PORT = ConfigVar('swapserver_port', default=None, type_=int)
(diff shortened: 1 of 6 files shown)


electrum/simple_config.py
```diff
@@ -814,12 +814,8 @@ def __setattr__(self, name, value):
 
     # connect to remote submarine swap server
     SWAPSERVER_URL = ConfigVar('swapserver_url', default='', type_=str)
-    # run submarine swap server locally
-    SWAPSERVER_PORT = ConfigVar('swapserver_port', default=None, type_=int)
-    SWAPSERVER_FEE_MILLIONTHS = ConfigVar('swapserver_fee_millionths', default=5000, type_=int)
     TEST_SWAPSERVER_REFUND = ConfigVar('test_swapserver_refund', default=False, type_=bool)
     SWAPSERVER_NPUB = ConfigVar('swapserver_npub', default=None, type_=str)
-    SWAPSERVER_ANN_POW_NONCE = ConfigVar('swapserver_ann_pow_nonce', default=0, type_=int)
```
SomberNight, line 822: This is used in `submarine_swaps.py`

PR #9651, opened 2025-03-18
Use manifest.json instead of loading init file for plugin registration

f321x:
This PR changes the way how plugins are discovered and loaded. Plugins now have to define a `manifest.json` which contains the same variables as previously the `__init__.py`. This way its possible find plugins and access their metadata without loading the init file upfront.


electrum/plugin.py
```diff
@@ -314,7 +290,18 @@ def load_plugin_by_name(self, name) -> 'BasePlugin':
                 zipfile = zipimport.zipimporter(metadata['path'])
                 init_spec = zipfile.find_spec(name)
             self.exec_module_from_spec(init_spec, base_name)
+            if metadata.get('requires_wallet_type'):
+                # removes trustedcoin after loading to not show it in the list of plugins
```
ecdsa, line 294: lets just use the plugin name for that

Issue #9370, opened 2024-12-17
"ModuleNotFoundError: No module named 'btchip'", however, it also won't let me install 'btchip'


gitaccount798:
Environment: Ubuntu 24.04, Electrum 4.5.8

Full error message:
```

File "/home/computer/.local/lib/python3.12/site-packages/electrum/plugins/ledger/ledger.py", line 42, in <module>
    from btchip.btchipComm import HIDDongleHIDAPI
ModuleNotFoundError: No module named 'btchip'
  9.41 | W | plugins.ledger.qt.Plugin | Library version for 'ledger' is incompatible.
Installed: 0.3.0, Needed: 0.2.0 <= x < 0.4.0
```

It won't also let me install btchip:
```

python3 -m pip install btchip-python
Defaulting to user installation because normal site-packages is not writeable
Collecting btchip-python
  Using cached btchip-python-0.1.32.tar.gz (20 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      error in btchip-python setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```
I have installed the following (as per https://github.com/spesmilo/electrum-docs/blob/master/hardware-linux.rst):

```
sudo apt-get install libusb-1.0-0-dev libudev-dev
python3 -m pip install hidapi ecdsa ledger-bitcoin  #(but it would not accept btchip-python)

```

apt show libudev-dev
Package: libudev-dev
Version: 255.4-1ubuntu8.4
Priority: optional
Section: admin
Source: systemd
Origin: Ubuntu
Maintainer: Ubuntu Developers [ubuntu-devel-discuss@lists.ubuntu.com](mailto:ubuntu-devel-discuss@lists.ubuntu.com)
Original-Maintainer: Debian systemd Maintainers [pkg-systemd-maintainers@lists.alioth.debian.org](mailto:pkg-systemd-maintainers@lists.alioth.debian.org)
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 197 kB
[...]

PR #9652, opened 2025-03-18
plugins: ledger: rm support for hw.1

SomberNight:
This removes support for Ledger HW.1 and "Nano" (non-S) devices. These were manufactured/sold around 2015-2016, and are long unsupported by the upstream vendor.

We previously added a [deprecation warning](https://github.com/spesmilo/electrum/commit/9b82eb6d06f8534652849f3370ab8af7d7ca9c7f) to the GUI, released in 4.3.3 (2023-01-02), to warn owners of these devices. This PR now fully removes support.

As a consequence, the unmaintained btchip-python dependency can now be removed, which solves https://github.com/spesmilo/electrum/issues/9370.
(Instead of btchip-python, we now only use ledger_bitcoin, which includes a stripped down version of btchip-python. However one of the things that was ["stripped out" is support for hw.1](https://github.com/LedgerHQ/app-bitcoin-new/blob/97f596e17b64619ed0f545c0482fb1cf94c6d282/bitcoin_client/ledger_bitcoin/btchip/README.md#L9))


gitaccount798:
Many thanks for fixing this!

PR #9654, opened 2025-03-19
Improve cmdline help

ecdsa:


electrum/commands.py
```diff
@@ -1598,108 +1943,26 @@ def eval_bool(x: str) -> bool:
...
+
 arg_types = {
-    'num': int,
-    'nbits': int,
-    'imax': int,
-    'year': int,
-    'from_height': int,
-    'to_height': int,
+    'int': int,
+    'bool': eval_bool,
```
f321x, line 1958: Could it be dangerous that `eval_bool()` always returns True for inputs not correctly representing a bool?
e.g.:
```
eval_bool("Falsee") -> True
eval_bool("Some other misplaced argument") -> True
```

ecdsa: indeed; let it raise an exception

PR #9656, opened 2025-03-19
plugin ConfigVars: define vars less dynamically

SomberNight:
- define vars less dynamically
- enforce "key" starts with name of plugin
    - this depends on https://github.com/spesmilo/electrum/pull/9655 to work properly

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -45,6 +45,7 @@ class ConfigVar(property):
         convert_getter: Callable[[Any], Any] = None,
         short_desc: Callable[[], str] = None,
         long_desc: Callable[[], str] = None,
+        plugin: Optional[str] = None,
     ):
         self._key = key
         self._default = default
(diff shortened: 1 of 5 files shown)


electrum/simple_config.py
```diff
@@ -56,6 +57,13 @@ def __init__(
         assert long_desc is None or callable(long_desc)
         self._short_desc = short_desc
         self._long_desc = long_desc
+        if plugin:  # enforce "key" starts with name of plugin
+            pkg_prefix = "electrum.plugins."  # for internal plugins
+            if plugin.startswith(pkg_prefix):
+                plugin = plugin[len(pkg_prefix):]
+            assert "." not in plugin, plugin
+            key_prefix = plugin + "_"
+            assert key.startswith(key_prefix), f"ConfigVar {key=} must be prefixed with the plugin name ({key_prefix})"
```
ecdsa, line 66: if a user-installed plugin triggers this crash, a non-programmer needs to see a nicer message.
For example "remove plugin {full path}".
Alternatively, we could disable the plugin instead of crashing Electrum.

SomberNight: This does not crash electrum.
Assuming we merge https://github.com/spesmilo/electrum/pull/9655, it only results in the plugin not getting loaded.
Without that PR, we *even* load the plugin! (in a half-broken state)

PR #9660, opened 2025-03-21
swaps: Announce available liquidity instead of static maximum value

f321x:
Makes swap provider announce the actual available liquidity for forward and reverse swaps instead of a static hardcoded max value. This should increase reliability of swaps as clients will know upfront if the provider even has enough liquidity to fulfill their request.

Depends on https://github.com/spesmilo/electrum/pull/9647

### DIFF
--- a/electrum/plugins/swapserver/server.py
+++ b/electrum/plugins/swapserver/server.py
@@ -66,7 +66,9 @@ class HttpSwapServer(Logger, EventListener):
                 "BTC/BTC": {
                     "rate": 1,
                     "limits": {
-                        "maximal": sm._max_amount,
+                        "maximal": min(sm._max_forward, sm._max_reverse),  # legacy
+                        "max_forward_amount": sm._max_forward,  # new version, uses 2 separate limits
+                        "max_reverse_amount": sm._max_reverse,
                         "minimal": sm._min_amount,
                     },
                     "fees": {
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -26,8 +26,8 @@ from .bitcoin import (script_to_p2wsh, opcodes,
 from .transaction import PartialTxInput, PartialTxOutput, PartialTransaction, Transaction, TxInput, TxOutpoint
 from .transaction import script_GetOp, match_script_against_template, OPPushDataGeneric, OPPushDataPubkey
 from .util import (log_exceptions, ignore_exceptions, BelowDustLimit, OldTaskGroup, age, ca_path,
-                   gen_nostr_ann_pow, get_nostr_ann_pow_amount, make_aiohttp_proxy_connector, get_running_loop,
-                   get_asyncio_loop)
+                   gen_nostr_ann_pow, get_nostr_ann_pow_amount, make_aiohttp_proxy_connector,
+                   get_running_loop, get_asyncio_loop, wait_for2)
 from .lnutil import REDEEM_AFTER_DOUBLE_SPENT_DELAY
 from .bitcoin import dust_threshold, DummyAddress
 from .logging import Logger
@@ -134,7 +134,8 @@ class SwapFees:
     percentage = attr.ib(type=int)
     mining_fee = attr.ib(type=int)
     min_amount = attr.ib(type=int)
-    max_amount = attr.ib(type=int)
+    max_forward = attr.ib(type=int)
+    max_reverse = attr.ib(type=int)
 
 @stored_in('submarine_swaps')
 @attr.s
@@ -175,7 +176,8 @@ class SwapManager(Logger):
         self.mining_fee = None
         self.percentage = None
         self._min_amount = None
-        self._max_amount = None
+        self._max_forward = None
+        self._max_reverse = None
 
         self.wallet = wallet
         self.config = wallet.config
@@ -202,6 +204,7 @@ class SwapManager(Logger):
         self.is_server = False # overriden by swapserver plugin if enabled
         self.is_initialized = asyncio.Event()
         self.pairs_updated = asyncio.Event()
+        self._liquidity_changed = asyncio.Event()
 
     def start_network(self, network: 'Network'):
         assert network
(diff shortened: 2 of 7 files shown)


f321x:
In `server_update_pairs()` i round down to the next 1000 sat:
```python
        oc_balance = sum([coin.value_sats() for coin in self.wallet.get_spendable_coins()])
        max_forward = min(int(self.lnworker.num_sats_can_receive()), oc_balance, 10000000)
        max_reverse = min(int(self.lnworker.num_sats_can_send()), 10000000)
        self._max_forward = (max_forward // 1000) * 1000
        self._max_reverse = (max_reverse // 1000) * 1000
```
e.g. changing 984323 to 984000. We could also randomly round down in a specified range... Let me know if you have any more ideas how we could improve privacy for providers without reducing functionality.

accumulator:
Maybe zero out after 1 or 2 significant numbers, e.g. 984323 -> 900000 or 980000

Also, maybe there's also a case to be made to offer at most 50% of available balance, to be able to service 2 clients if balance is below 10Msat cutoff point. 

Maybe the anchors UTXO reserve also should be taken into account? (haven't checked)

f321x:
I changed it to reduce the precision of the announced amounts to 2 leading digits, and deduct the anchor reserve, if required, when calculating the available onchain amount.

> Also, maybe there's also a case to be made to offer at most 50% of available balance, to be able to service 2 clients if balance is below 10Msat cutoff point.

I'm not sure how this improves the usability, the provider announces the new liquidity as soon as it changed, so the other clients see what is still available. I imagine it could prevent one rejected swap if two clients request a swap > 50% of the available liquidity at the same time (which should happen rarely), but at the same time it would also prevent clients from requesting swaps > 50% even if the provider could satisfy the swap easily (which probably happens more often in practice).

electrum/simple_config.py
```diff
@@ -846,6 +846,13 @@ def __setattr__(self, name, value):
     ACCEPT_ZEROCONF_CHANNELS = ConfigVar('accept_zeroconf_channels', default=False, type_=bool)
     ZEROCONF_TRUSTED_NODE = ConfigVar('zeroconf_trusted_node', default='', type_=str)
     ZEROCONF_MIN_OPENING_FEE = ConfigVar('zeroconf_min_opening_fee', default=5000, type_=int)
+    LN_UTXO_RESERVE = ConfigVar(
+        'ln_utxo_reserve',
+        default=10000,
+        type_=int,
+        short_desc=lambda: _("Amount that must be kept on-chain in order to sweep anchor output channels"),
+        long_desc=lambda: _("Do not set this below dust limit"),
+    )
```
ecdsa, line 856: this is from another branch...

ecdsa:
note: this PR seems to require `LN_UTXO_RESERVE`

electrum/submarine_swaps.py
```diff
@@ -907,17 +924,35 @@ def _add_or_reindex_swap(self, swap: SwapData) -> None:
...
         """ for server """
-        self.percentage = float(self.config.SWAPSERVER_FEE_MILLIONTHS) / 10000
+        self.percentage = float(self.config.SWAPSERVER_FEE_MILLIONTHS) / 10000  # type: ignore
         self._min_amount = 20000
-        self._max_amount = 10000000
+        anchor_reserve = self.config.LN_UTXO_RESERVE \
+                                if (any(chan.has_anchors() and not chan.is_redeemed()
+                                for chan in self.lnworker.channels.values())) else 0
+        oc_balance = max(sum([coin.value_sats() for coin in self.wallet.get_spendable_coins()])
+                                - anchor_reserve, 0)
```
ecdsa, line 933: I think computing this `oc_balance` should really be a wallet method.

electrum/plugins/swapserver/server.py
```diff
@@ -66,7 +66,9 @@ async def get_pairs(self, r):
                 "BTC/BTC": {
                     "rate": 1,
                     "limits": {
-                        "maximal": sm._max_amount,
+                        "maximal": min(sm._max_forward, sm._max_reverse),  # legacy
```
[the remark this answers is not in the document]
f321x: yes i think so, it allows old clients (without nostr) to use a new swapserver, also the old client crashes completely if the 'maximal' key is not served when requesting a swap from the configured http server.

electrum/submarine_swaps.py
f321x: I moved the calculation to wallet.py in https://github.com/spesmilo/electrum/pull/9660/commits/9e04d9d5281249de556d48a152739a999b4c5d5f
Is this good or should the anchor reserve be a completely separated concern?

electrum/plugins/swapserver/server.py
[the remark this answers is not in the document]
ecdsa: oh right

electrum/submarine_swaps.py
ecdsa: I wanted to avoid code duplication.
for example `has_anchor_channels` could be called from `should_keep_reserve_utxo`

electrum/wallet.py
```diff
@@ -1069,6 +1069,27 @@ def get_balance(self, **kwargs):
...
+    def anchor_reserve(self) -> int:
+        if self.lnworker is None or not isinstance(self.lnworker, LNWallet):
+            return 0
+        if not self.lnworker.has_anchor_channels():
+            return 0
+        if self.config.LN_UTXO_RESERVE is not None:
+            return self.config.LN_UTXO_RESERVE
+        else:
+            # if we have anchor channels LN_UTXO_RESERVE should be set
+            raise Exception("LN_UTXO_RESERVE not set in config")
```
ecdsa, line 1081: not sure if that branch is useful. 
if the user 'unsets' a config var, it returns its default value, not None:
```
voegtlin@z:~/electrum$ $bob -o getconfig ln_utxo_reserve
10000
voegtlin@z:~/electrum$ $bob -o setconfig ln_utxo_reserve 20000
voegtlin@z:~/electrum$ $bob -o getconfig ln_utxo_reserve
20000
voegtlin@z:~/electrum$ $bob -o setconfig ln_utxo_reserve None
voegtlin@z:~/electrum$ $bob -o getconfig ln_utxo_reserve
10000

```

f321x: Yeah i see it shouldn't happen that this is None, removed the branch in https://github.com/spesmilo/electrum/pull/9660/commits/6e8bdb346a9009caf1a5e22732b7df65c4ae8d81

PR #9681, opened 2025-03-28
wallet: make txbatcher wait for network connection

f321x:
Currently txbatcher deletes a batch when there is no network connection as a `NoDynamicFeeEstimates()` exception is raised. We can make the txbatcher wait for network connection like this PR does to prevent this, or is there a point in txbatcher iterating even if no network connection is available? Conceptually i think broadcasting transactions requires network anyways so i don't see a point to keep it iterating without network.

Sample exception:
```
 30.89 | I | txbatcher.TxBatch | to_sweep: [<TxOutpoint("70f84911582b4053a67009ae164238dd44e5055c8c37aba0e29fc7f7b5be82b2:2")>, <TxOutpoint("8d613ced3b30b9dd8733c0053ab801b6fb9090dc6b1dc895c5b59f5d33682788:1")>]
 30.89 | I | txbatcher.TxBatch | to_pay: []
 30.89 | I | txbatcher.TxBatch | locktime: None
 30.91 | E | txbatcher.TxBatcher | TxBatch error: NoDynamicFeeEstimates()
  File "/home/user/code/electrum-fork/electrum/txbatcher.py", line 146, in run
    await txbatch.run_iteration(password)
  File "/home/user/code/electrum-fork/electrum/txbatcher.py", line 310, in run_iteration
    tx = self.create_next_transaction(base_tx, password)
  File "/home/user/code/electrum-fork/electrum/txbatcher.py", line 355, in create_next_transaction
    tx = self._create_batch_tx(base_tx, to_sweep_now, to_pay, password)
  File "/home/user/code/electrum-fork/electrum/txbatcher.py", line 394, in _create_batch_tx
    tx = self.wallet.make_unsigned_transaction(
        fee_policy=self.fee_policy,
    ...<5 lines>...
        merge_duplicate_outputs=False,
    )
  File "/home/user/code/electrum-fork/electrum/util.py", line 492, in do_profile
    o = func(*args, **kw_args)
  File "/home/user/code/electrum-fork/electrum/wallet.py", line 1941, in make_unsigned_transaction
    tx = coin_chooser.make_tx(
        coins=coins,
    ...<4 lines>...
        dust_threshold=self.dust_threshold(),
        BIP69_sort=BIP69_sort)
  File "/home/user/code/electrum-fork/electrum/coinchooser.py", line 349, in make_tx
    all_buckets = self.bucketize_coins(coins, fee_estimator_vb=fee_estimator_vb)
  File "/home/user/code/electrum-fork/electrum/coinchooser.py", line 123, in bucketize_coins
    constant_fee = fee_estimator_vb(2000) == fee_estimator_vb(200)
                   ~~~~~~~~~~~~~~~~^^^^^^
  File "/home/user/code/electrum-fork/electrum/fee_policy.py", line 250, in estimate_fee
    raise NoDynamicFeeEstimates()
electrum.util.NoDynamicFeeEstimates: Dynamic fee estimates not available
 30.92 | I | txbatcher.TxBatcher | deleting TxBatch lnwatcher

```


electrum/txbatcher.py
```diff
@@ -139,7 +142,8 @@ async def run(self):
         while True:
             await asyncio.sleep(self.SLEEP_INTERVAL)
             password = self.wallet.get_unlocked_password()
-            if self.wallet.has_keystore_encryption() and not password:
+            if (self.wallet.has_keystore_encryption() and not password
+                    or not self.wallet.network.is_connected()):
```
ecdsa, line 146: Priority precedence of boolean operators makes parenthesis unnecessary in this expression, yet you added them, so I guess you added them for clarity? In that case, the parenthesis should be closed after `password`.

f321x: They are added by my IDE (PyCharm) when breaking lines (instead of `\` ). Switched to `\` to make it clearer, and added check for existence of network.

SomberNight: I personally like this kind of style most:
```
if self.wallet.has_keystore_encryption() and not password:
	continue
if not (self.wallet.network and self.wallet.network.is_connected()):
	continue
```
As in, start a new "if" for each logically bundled list of checks, and just duplicate the early exit.
I think it is the most readable, and least error-prone. It categorically removes operator precedence issues for example.

electrum/txbatcher.py
```diff
@@ -139,7 +142,8 @@ async def run(self):
         while True:
             await asyncio.sleep(self.SLEEP_INTERVAL)
             password = self.wallet.get_unlocked_password()
-            if self.wallet.has_keystore_encryption() and not password:
+            if self.wallet.has_keystore_encryption() and not password \
+                    or not self.wallet.network or not self.wallet.network.is_connected():
                 continue
             for key, txbatch in list(self.tx_batches.items()):
                 try:
```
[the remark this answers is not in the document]
f321x: I now catch `NoDynamicFeeEstimates` explicitly in `run_iteration()` to prevent it from happening if we have network but there are no fees available, which seems quite likely right after we get network:
```python
        try:
            tx = self.create_next_transaction(base_tx, password)
        except NoDynamicFeeEstimates:
            self.logger.debug('no dynamic fee estimates available')
            return
```


I think it still makes sense to wait for network connection in the loop to also prevent starting a new batch here if broadcasting the tx fails (which it will if there is no network):
```python
run_iteration():
...
else:
            # most likely reason is that base_tx is not replaceable
            # this may be the case if it has children (because we don't pay enough fees to replace them)
            # or if we are trying to sweep unconfirmed inputs (replacement-adds-unconfirmed error)
            self.logger.info(f'cannot broadcast tx {tx}')
            if base_tx:
                self.logger.info(f'starting new batch because could not broadcast')
                self._start_new_batch(base_tx)
```

electrum/txbatcher.py
```diff
@@ -141,6 +144,8 @@ async def run(self):
             password = self.wallet.get_unlocked_password()
             if self.wallet.has_keystore_encryption() and not password:
                 continue
+            if not (self.wallet.network or not self.wallet.network.is_connected()):
+                continue
```
SomberNight, line 148: The boolean logic is incorrect. The value of `is_connected` would never get checked.
You need either:
```py
if not (self.wallet.network and self.wallet.network.is_connected()):
    continue
```
or
```py
if not self.wallet.network or not self.wallet.network.is_connected():
    continue
```

f321x: Oh yeah, thanks for catching this, fixed in https://github.com/spesmilo/electrum/pull/9681/commits/b6137539fea0608ba1c5d66772b9e4e858c04d89

PR #9682, opened 2025-03-29
Android >=13 permissions support

accumulator:


electrum/gui/qml/qeapp.py
```diff
@@ -173,6 +183,42 @@ def bindIntent(self):
         except Exception as e:
             self.logger.error(f'unable to bind intent: {repr(e)}')
 
+    @pyqtSlot(str, result=bool)
+    def hasPermission(self, permissionFqcn: str) -> bool:
```
SomberNight, line 187: what is "fqcn" short for? "fully qualified .. name"? is this a term from the android dev docs, or just something you came up with in the spirit of FQDN?

SomberNight (approved):
I haven't tested this -- but it looks good!

electrum/gui/qml/qeapp.py
accumulator: it's a bit of a java-ism. fully qualified constant name. no objection to change.

PR #9684, opened 2025-03-31
qt: use system theme as default theme and add 'System' theme option

f321x:
Adds a theme setting to qt gui to use the system theme color instead of a fixed theme, also uses it as default config.

Edit: see https://github.com/spesmilo/electrum/issues/9690, will wait for the qt function to handle this properly

### DIFF
--- a/electrum/gui/qt/settings_dialog.py
+++ b/electrum/gui/qt/settings_dialog.py
@@ -36,7 +36,7 @@ from electrum.util import base_units_list, event_listener
 
 from electrum.gui import messages
 
-from .util import ColorScheme, HelpLabel, Buttons, CloseButton, QtEventListener
+from .util import ColorScheme, HelpLabel, Buttons, CloseButton, QtEventListener, is_system_dark_mode
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


electrum/gui/qt/settings_dialog.py
```diff
@@ -252,15 +252,26 @@ def on_video_device(x):
         qr_combo.currentIndexChanged.connect(on_video_device)
 
         colortheme_combo = QComboBox()
-        colortheme_combo.addItem(_('Light'), 'default')
+        colortheme_combo.addItem(_('Light'), 'light')
         colortheme_combo.addItem(_('Dark'), 'dark')
+        colortheme_combo.addItem(_('System'), 'default')
```
SomberNight, line 257: When running on a system that has a dark theme, what exactly happens when the user sets GUI_QT_COLOR_THEME to "light"? Presumably the user would then expect Electrum to run in a light theme - but that's not what happens, is it? I suppose it is only a few colours here-and-there that gets changed.

What I am getting at is that the cornerstone of the "dark" theme is the qdarkstyle package, and the "system" ("default") theme is just whatever qt does by default. -- and on top of all that we have some hardcoded colors that depend on `GUI_QT_COLOR_THEME` (see `util.ColorScheme`)

electrum/gui/qt/util.py
```diff
@@ -1328,6 +1328,17 @@ def font_height(widget: QWidget = None) -> int:
     return QFontMetrics(widget.font()).height()
 
 
+def is_system_dark_mode() -> bool:
+    """Returns True if the system theme is in dark mode, False otherwise."""
```
SomberNight, line 1332: What is the motivation for this PR? Is it that when running on a dark system, you want an option to run electrum with the system qt theming but with `util.ColorScheme` selecting the dark colors? If so, there is already some related logic here that seems to duplicate your `is_system_dark_mode()` function (perhaps that logic is flawed and yours is better -- but maybe then it should be fixed instead :) ):
https://github.com/spesmilo/electrum/blob/1559129016a21dfa74a34287c35a9e6cecf270ff/electrum/gui/qt/__init__.py#L195-L198
https://github.com/spesmilo/electrum/blob/1559129016a21dfa74a34287c35a9e6cecf270ff/electrum/gui/qt/util.py#L1144-L1150

electrum/gui/qt/settings_dialog.py
f321x: When the user sets `GUI_QT_COLOR_THEME` to light and has its system theme in dark mode electrum will start in the regular light theme respecting the configuration (nothing changes). 

The only place i see that is dependent on the value of `GUI_QT_COLOR_THEME` is `reload_app_stylesheet()`:
```python
        use_dark_theme = self.config.GUI_QT_COLOR_THEME == 'dark' \
                            or self.config.GUI_QT_COLOR_THEME == 'default' and is_system_dark_mode()
        if use_dark_theme:
            try:
                import qdarkstyle
                self.app.setStyleSheet(qdarkstyle.load_stylesheet_pyqt6())
            except BaseException as e:
                use_dark_theme = False
                self.logger.warning(f'Error setting dark theme: {repr(e)}')
        else:
            self.app.setStyleSheet(self._default_qtstylesheet)
```

The only thing that changes in this pr is that `use_dark_theme` gets an additional condition, if `GUI_QT_COLOR_THEME` is set to default ('System' setting in GUI) and if the system theme is dark (determined by `is_system_dark_mode()`) then also use the dark mode, else just go the same path as if the configuration is `Light`.

electrum/gui/qt/util.py
f321x: The motivation for this PR is that its annoying having to set the theme manually in every application instead of just switching the system theme and have all applications follow it, especially when switching often due to lighting conditions. So this seemed like a small useful change to allow electrum automatically switch between the `Light` and `Dark` setting depending on the current system theme it is starting in. Am i missing some point here?

Oh yeah i didn't see `has_dark_background()`, seems like it does the same thing. Will try to unify them.

f321x:
Just noticed this doesn't work on gnome (always detects system theme as light), will change the approach of detection...

SomberNight:
(somewhat related: https://github.com/spesmilo/electrum/issues/9690)

PR #9685, opened 2025-03-31
cli: add hold invoice cli functionality

f321x:
Adds four new commands to the CLI which allow handling hold invoices

* add_hold_invoice
* cancel_hold_invoice
* check_hold_invoice
* settle_hold_invoice

Users have to bring their own preimage.

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -25,6 +25,7 @@
 import io
 import sys
 import datetime
+import time
 import argparse
 import json
 import ast
@@ -60,10 +61,8 @@ from .synchronizer import Notifier
 from .wallet import Abstract_Wallet, create_new_wallet, restore_wallet_from_text, Deterministic_Wallet, BumpFeeStrategy, Imported_Wallet
 from .address_synchronizer import TX_HEIGHT_LOCAL
 from .mnemonic import Mnemonic
-from .lnutil import SENT, RECEIVED
-from .lnutil import LnFeatures
 from .lntransport import extract_nodeid
-from .lnutil import channel_id_from_funding_tx
+from .lnutil import channel_id_from_funding_tx, LnFeatures, SENT, RECEIVED, MIN_FINAL_CLTV_DELTA_FOR_INVOICE
 from .plugin import run_hook, DeviceMgr, Plugins
 from .version import ELECTRUM_VERSION
 from .simple_config import SimpleConfig
@@ -77,6 +76,7 @@ from . import descriptor
 if TYPE_CHECKING:
     from .network import Network
     from .daemon import Daemon
+    from electrum.lnworker import PaymentInfo
 
 
 known_commands = {}  # type: Dict[str, Command]
@@ -995,7 +995,6 @@ class Commands(Logger):
     def get_year_timestamps(self, year:int):
         kwargs = {}
         if year:
-            import time
             start_date = datetime.datetime(year, 1, 1)
             end_date = datetime.datetime(year+1, 1, 1)
             kwargs['from_timestamp'] = time.mktime(start_date.timetuple())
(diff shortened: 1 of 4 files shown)


electrum/commands.py
```diff
@@ -1346,6 +1344,127 @@ async def add_request(self, amount, memo='', expiry=3600, lightning=False, force
...
 
+    @command('wnl')
+    async def add_hold_invoice(
+            self,
+            payment_hash,
+            amount = None,
+            memo = "",
+            expiry = 3600,
+            min_final_cltv_expiry_delta = MIN_FINAL_CLTV_DELTA_FOR_INVOICE * 2,
+            callback_url = None,
```
ecdsa, line 1355: Do we really need to support `callback_url`?
I thought the idea was to let the user poll using check_hold_invoice.
if we drop this option, I believe we do not need to add a new persisted dict for `cli_hold_invoice_callbacks`

f321x: I removed the callback URL handling, now there is no need for an additional persisted dict anymore. 

Some form of event notification system would make sense for use cases where lots of hold invoices have to get checked at the same time (e.g. robosats). However i think it would make sense to implement something more generic which could also be used for other notifications, so this should be in a separate PR.

ecdsa: indeed, a generic callback handling would make sense; we could use it for other things, such as receiving onion messages.

electrum/commands.py
```diff
@@ -1349,6 +1348,134 @@ async def add_request(self, amount, memo='', expiry=3600, lightning=False, force
...
+    @command('wnl')
+    async def add_hold_invoice(
+            self,
+            preimage: str,
+            amount: Optional[Decimal] = None,
+            memo: str = "",
+            expiry: int = 3600,
+            min_final_cltv_expiry_delta: int = MIN_FINAL_CLTV_DELTA_FOR_INVOICE * 2,
+            wallet: Abstract_Wallet = None
+        ) -> dict:
```
SomberNight, line 1360: I don't understand this API.
How is one supposed to use it? What is an example intended use case?

If we know the preimage, why would we hold the htlcs? We should just fulfill them and claim the money.

Shouldn't `add_hold_invoice` take a `payment_hash` instead of a `preimage`,
and `settle_hold_invoice` take both a `payment_hash` and a `preimage`?
Isn't the whole point of a "hold invoice" that the htlc-destination party does not initially have the preimage but is willing to hold the htlcs, and then later it somehow (perhaps externally) discovers the preimage and settles the htlcs?

I wanted to write a regtest utilising this. With a pending "hold-invoice" and pending htlc, I closed a channel, and wanted the sender to time out the htlcs, but of course the recipient at that point "breached the contract" and just settled the hold-invoice onchain. Which makes perfect sense as it had the preimage, because of how the API is written. but if it's ok to claim onchain, we could also just claim offchain, and at that point it's no longer a hold-invoice.

f321x: The use case i had in mind was giving out a hold invoice and then being able to settle or fail it later depending on some other condition. This is what e.g. sms4sats or robosats do, essentially being able to refund the sender without additional interaction (providing new invoice...) or do some short term escrow. For these things the receiver usually already has the preimage from the beginning.

But yeah i see that it prevents usage in scenarios like you describe which seem useful as well, i think we can also just use a hash to create the hold invoice so it can be used to do this too. 
Then we also have to move this block a bit higher so `LNPeer` doesn't fail the htlcs if no preimage is available:
https://github.com/spesmilo/electrum/blob/50d48aee2d42ec6064d94c6cf1301ff0e29175c6/electrum/lnpeer.py#L2556

>  and settle_hold_invoice take both a payment_hash and a preimage?

for `settle_hold_invoice` the preimage as argument should be enough as it can internally calculate the hash, or do you see an issue with doing that?

ecdsa: indeed, add_hold_invoice should take a payment hash.

SomberNight: >>  and settle_hold_invoice take both a payment_hash and a preimage?
>
> for `settle_hold_invoice` the preimage as argument should be enough as it can internally calculate the hash, or do you see an issue with doing that?

Sure, that's fine too.

SomberNight: > This is what e.g. sms4sats or robosats do, essentially being able to refund the sender without additional interaction (providing new invoice...) or do some short term escrow. For these things the receiver usually already has the preimage from the beginning.

Ok, I see. Indeed for a temporary trusted bond kind of thing, it makes sense that we hold an htlc that we can either settle or fail, and decide which to do later.
However in that case, is it ok that we will settle it if the channel gets closed?  (maybe it is, not sure.)

f321x: So i had a look at other projects, LND essentially does it as you expected it to work (supply payment hash first and preimage later), while [this CLN plugin](https://github.com/daywalker90/holdinvoice) does it as we do right now.
Providing the hash first and not settling automatically if the channel gets closed seems more reasonable i think, so I opened a PR https://github.com/spesmilo/electrum/pull/9998 that changes the cli.

PR #9689, opened 2025-04-01
lightning: don't use fallback feerates for lightning fee estimation

f321x:
Stops using a fallback feerate for lightning fee estimation function (`current_target_feerate_per_kw()`) and prevents trying channel reestablishment before we have fee estimates available and are fully synced to prevent errors like this and resulting causing force closes:
```
INFO | lnpeer.Peer.[LNWallet, ...] | remote peer sent error [DO NOT TRUST THIS MESSAGE]: 'local/remote feerates are too different: remoteFeeratePerKw=37500 localFeeratePerKw=788'. chan_id=.... is_known_chan_id=True
```

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -23,8 +23,8 @@ from .crypto import sha256, sha256d, privkey_to_pubkey
 from . import bitcoin, util
 from . import constants
 from .util import (bfh, log_exceptions, ignore_exceptions, chunks, OldTaskGroup,
-                   UnrelatedTransactionException, error_text_bytes_to_safe_str, AsyncHangDetector)
-from .util import event_listener, EventListener
+                   UnrelatedTransactionException, error_text_bytes_to_safe_str, AsyncHangDetector,
+                   NoDynamicFeeEstimates, event_listener, EventListener)
 from . import transaction
 from .bitcoin import make_op_return, DummyAddress
 from .transaction import PartialTxOutput, match_script_against_template, Sighash
@@ -993,7 +993,9 @@ class Peer(Logger, EventListener):
             raise Exception('Not a trampoline node: ' + str(self.their_features))
 
         channel_flags = CF_ANNOUNCE_CHANNEL if public else 0
-        feerate = self.lnworker.current_target_feerate_per_kw()
+        feerate: Optional[int] = self.lnworker.current_target_feerate_per_kw()
+        if feerate is None:
+            raise NoDynamicFeeEstimates()
         # we set a channel type for internal bookkeeping
         open_channel_tlvs = {}
         assert self.their_features.supports(LnFeatures.OPTION_STATIC_REMOTEKEY_OPT)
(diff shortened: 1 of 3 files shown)


electrum/lnpeer.py
```diff
@@ -993,7 +993,9 @@ async def channel_establishment_flow(
             raise Exception('Not a trampoline node: ' + str(self.their_features))
 
         channel_flags = CF_ANNOUNCE_CHANNEL if public else 0
-        feerate = self.lnworker.current_target_feerate_per_kw()
+        feerate: Optional[int] = self.lnworker.current_target_feerate_per_kw()
+        if not feerate:
```
SomberNight, line 997:
```suggestion
        if feerate is None:
```

electrum/lnpeer.py
```diff
@@ -2590,12 +2597,16 @@ def maybe_update_fee(self, chan: Channel):
             return
         if chan.get_state() != ChannelState.OPEN:
             return
-        feerate_per_kw = self.lnworker.current_target_feerate_per_kw()
-        def does_chan_fee_need_update(chan_feerate: Union[float, int]) -> bool:
+        feerate_per_kw: Optional[int] = self.lnworker.current_target_feerate_per_kw()
+        if not feerate_per_kw:
```
SomberNight, line 2601:
```suggestion
        if feerate_per_kw is None:
```

electrum/lnworker.py
```diff
@@ -3000,21 +3000,28 @@ async def reestablish_peers_and_channels(self):
                 else:
                     await self.taskgroup.spawn(self.reestablish_peer_for_given_channel(chan))
 
-    def current_target_feerate_per_kw(self) -> int:
+    def current_target_feerate_per_kw(self, allow_fallback_to_static_rates: bool = False) -> Optional[int]:
```
SomberNight, line 3003: This `allow_fallback_to_static_rates` parameter was my suggestion, but maybe it is unnecessary: looks like it is not set anywhere.

-----

On a generic note: this a good example for a function parameter that IMO should be keyword-only.
(`def current_target_feerate_per_kw(self, *, allow_fallback_to_static_rates: bool = False)`)

I think in many cases it is preferable to force params to be kw-only.
If the function name makes it obvious what the parameters are, then ofc they do not need to be kw-only, e.g.
`def sum(iterable)` or `def sum(a, b)`

but for `def current_target_feerate_per_kw(self, allow_fallback_to_static_rates: bool = False)`,
the reader has no idea what `current_target_feerate_per_kw(True)` does (without looking up the function signature).

electrum/lnpeer.py
f321x: Fixed in https://github.com/spesmilo/electrum/pull/9689/commits/eb37821df35b572a40193a6d683efb5f558521e9

electrum/lnpeer.py
f321x: Fixed in https://github.com/spesmilo/electrum/pull/9689/commits/eb37821df35b572a40193a6d683efb5f558521e9

electrum/lnworker.py
f321x: Removed the parameter in https://github.com/spesmilo/electrum/pull/9689/commits/eb37821df35b572a40193a6d683efb5f558521e9
Thanks for the hint, i see parameters like this are less confusing if kw only.

tests/test_lnpeer.py
```diff
@@ -95,6 +95,10 @@ async def broadcast_transaction(self, tx):
     async def try_broadcasting(self, tx, name):
         await self.broadcast_transaction(tx)
 
+    def populate_fee_estimates(self):
+        for target in FEE_ETA_TARGETS[1:]:
```
SomberNight, line 99: why the `[1:]`? why is the first element skipped?

f321x: Because `FeeTimeEstimates.has_data()` checks for `len(FEE_ETA_TARGETS) - 1` items as we don't request the next block fee (see comment):
```python
    def has_data(self):
        # we do not request estimate for next block fee
        return len(self.data) == len(FEE_ETA_TARGETS) - 1
```
However i guess its better to do [:-1] instead of [1:] to really skip the next block target instead of the 25 block target.

PR #9691, opened 2025-04-02
swaps: improve send change to lightning and swapserver error handling

f321x:
If the user tries to send change to lightning in an onchain payment and the swap amount is below the supported amount by the provider it will send `invoiceAmount` None to the server, this causes the swapserver taskgroup to stop due to an exception and the client will be stuck in the 'Requesting swap invoice...' dialog waiting for a response. 
This pr adds a error response to the swapserver so the client is able to handle errors of the server and checks if `invoiceAmount` is not none before even trying to request a swap.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -123,7 +123,13 @@ def check_reverse_redeem_script(
 
 
 class SwapServerError(Exception):
+    def __init__(self, message=None):
+        self.message = message
+        super().__init__(message)
+
     def __str__(self):
+        if self.message:
+            return self.message
         return _("The swap server errored or is unreachable.")
 
 def now():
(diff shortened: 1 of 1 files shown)


Smilley27 (approved):

electrum/submarine_swaps.py
```diff
@@ -1515,7 +1528,15 @@ async def check_direct_messages(self):
...
-                await self.handle_request(content)
+                try:
+                    await self.handle_request(content)
+                except Exception as e:
+                    self.logger.exception(f"failed to handle request: {content}")
+                    error_response = json.dumps({
+                        "error": str(e)[:100],
+                        "reply_to": event.id,
+                    })
+                    await self.send_direct_message(event.pubkey,[], error_response)
```
ecdsa, line 1539: what is this for? it two servers are talking to each other, could it trigger an infinite loop?

PR #9694, opened 2025-04-03
psbt_nostr: split generic and UI parts, implement for qml

accumulator:
also some boilerplate for plugins extending QML GUI

### DIFF
--- a/electrum/plugins/psbt_nostr/qml.py
+++ b/electrum/plugins/psbt_nostr/qml.py
@@ -36,11 +36,13 @@ from electrum.util import EventListener, event_listener
 
 from electrum.gui.qml.qewallet import QEWallet
 
-from .psbt_nostr import PsbtNostrPlugin, CosignerWallet
+from .psbt_nostr import PsbtNostrPlugin, CosignerWallet, now
 
 if TYPE_CHECKING:
     from electrum.wallet import Abstract_Wallet
     from electrum.gui.qml import ElectrumQmlApplication
+
+USER_PROMPT_COOLDOWN = 10
 
 
 class QReceiveSignalObject(QObject):
@@ -69,6 +71,13 @@ class QReceiveSignalObject(QObject):
         if not cosigner_wallet:
             return
         cosigner_wallet.accept_psbt(event_id)
+
+    @pyqtSlot(QEWallet, str)
+    def rejectPsbt(self, wallet: 'QEWallet', event_id: str):
+        cosigner_wallet = self._plugin.cosigner_wallets[wallet.wallet]
+        if not cosigner_wallet:
+            return
+        cosigner_wallet.reject_psbt(event_id)
 
 
 class Plugin(PsbtNostrPlugin):
@@ -103,13 +112,18 @@ class QmlCosignerWallet(EventListener, CosignerWallet):
         self.plugin = plugin
         self.register_callbacks()
 
-        self.pending = None
+        self.tx = None
+        self.user_prompt_cooldown = None
 
     @event_listener
-    def on_event_psbt_nostr_received(self, wallet, pubkey, event, tx: 'PartialTransaction'):
+    def on_event_psbt_nostr_received(self, wallet, pubkey, event_id, tx: 'PartialTransaction'):
         if self.wallet == wallet:
-            self.plugin.so.cosignerReceivedPsbt.emit(pubkey, event, tx.serialize())
-            self.on_receive(pubkey, event, tx)
+            self.tx = tx
+            if not (self.user_prompt_cooldown and self.user_prompt_cooldown > now()):
+                self.plugin.so.cosignerReceivedPsbt.emit(pubkey, event_id, tx.serialize())
+            else:
+                self.mark_pending_event_rcvd(event_id)
+                self.add_transaction_to_wallet(self.tx, on_failure=self.on_add_fail)
 
     def close(self):
         super().close()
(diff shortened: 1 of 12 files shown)


electrum/plugins/psbt_nostr/qml.py
```diff
@@ -34,40 +43,101 @@
...
+        except concurrent.futures.CancelledError:
+            pass
+        except Exception as e:
+            self.plugin.so.sendPsbtFailed.emit(str(e))
+
+    def on_receive(self, pubkey, event_id, tx):
+        self.pending = (pubkey, event_id, tx)
+
+    def accept_psbt(self, my_event_id):
+        pubkey, event_id, tx = self.pending
```
f321x, line 140: Got the following exception when accepting 2 incoming cosigner requests at the same time (by broadcasting from 2 signers in a 3-of-3, then opening the third signer which will show 2 overlapping popups, clicking yes on both produces this exception):
```
  File "/home/user/code/electrum-fork/electrum/plugins/psbt_nostr/qml.py", line 71, in acceptPsbt
    cosigner_wallet.accept_psbt(event_id)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/plugins/psbt_nostr/qml.py", line 140, in accept_psbt
    pubkey, event_id, tx = self.pending
    ^^^^^^^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable NoneType object
```

f321x:
Tested QT and QML.
IIUC the plugin is disabled by default and there is no way to enable it from the QML gui (on Android), should there be a config variable set for Android in `run_electrum` or some switch in the settings?

I think it would also be useful to allow the user to configure relays (like in the QT gui), as they maybe want to use their own relay for privacy or the default relays could all be unavailable, rendering this function unusable until the next release.

electrum/plugins/psbt_nostr/qml.py
accumulator: Yes, good catch, testing this was indeed left on my TODO list :)

I wasn't sure if the desktop client handled this case, but I was thinking it might be better to put the responsibility on the plugin to emit only one PSBT, until the PSBT/event is accepted. This way we don't have to implement this in every GUI.

accumulator:
> Tested QT and QML. IIUC the plugin is disabled by default and there is no way to enable it from the QML gui (on Android), should there be a config variable set for Android in `run_electrum` or some switch in the settings?

Correct, I haven't implemented this yet. We can hard-enable this, or create a Preferences option for it.

> I think it would also be useful to allow the user to configure relays (like in the QT gui), as they maybe want to use their own relay for privacy or the default relays could all be unavailable, rendering this function unusable until the next release.

Agreed

electrum/plugins/psbt_nostr/qml.py
accumulator: correction: it seems it is not blocking the entire UI thread, but the signal handler itself is blocking on the `window.question()`
https://github.com/spesmilo/electrum/blob/78c6f7c77f28f28de084298c3499277d8faa3fba/electrum/plugins/psbt_nostr/qt.py#L138-L145

regardless, this is the result of a little spamming on desktop 

image

What to do if the user answers no? Present the next psbt event popup (potentially many)? Or stop asking altogether? (maybe skip all received events?)

accumulator: It looks more like an inbox pattern than an event pattern.. Maybe receive all to wallet and add another tab 'Sign Requests'? 😄

accumulator: This would also solve supercedes, e.g. remove older psbts that spend the same utxos, due to re-send of RbF-ed or same tx.

ecdsa: Maybe if the user says 'no', the event should be added to `self.known_events`. That would reduce the spam.

Also, if a tx has more than 2 cosigners, we should try to show only the most recent version of the tx, or the one that has the most signatures.

accumulator:
I've added an `Event.wait()` in `check_direct_messages` to pace the incoming PSBTs. 

If the user responds No to 'open psbt now', it will add the transaction to the wallet and set a short cooldown timeout to process existing PSBTs in the nostr-queue. 

Some extra PSBT handling is still TODO, e.g. check to only add 'newer' txs with more signatures, or auto-combine signatures if PSBTs are coming from multiple cosigners. Also GUI-wise there's some fighting between transaction dialog open and the popup for another PSBT opening at the same time.

ecdsa:
> If the user responds No to 'open psbt now', it will add the transaction to the wallet and set a short cooldown timeout to process existing PSBTs in the nostr-queue without asking the user.

On qt, if I click 'No', the history tab is not refreshed after the tx is added to the wallet. Also, the plugin is not informing me that it added the tx to my history. Maybe, instead of a yes/no dialog, we could have 3 buttons: 'Discard', 'Save to wallet' and 'Open'.

accumulator:
> Maybe, instead of a yes/no dialog, we could have 3 buttons: 'Discard', 'Save to wallet' and 'Open'.

Ok, that sounds better than an additional popup mentioning it saved to wallet, and also removes the need for a cooldown. It will instead show this popup for each PSBT in the queue in sequence.

accumulator:
Also, this plugin might be usable for watch-only wallets to request signing by the seeded wallet.

ecdsa:
> Also, this plugin might also be usable for watch-only wallets to request signing by the seeded wallet.

Yeah. I renamed it 'Nostr Multisig' because PSBT is cryptic and too vague. but we could call it 'Nostr Cosigner' ?

PR #9703, opened 2025-04-07
qml: call QEChannelOpener.validate() on updateMaxAmount

f321x:
Calls `QEChannelOpener.validate()` after setting the max amount slider in the channel open dialog so that the `Open Channel...` button gets activated when max amount was set. 
Right now if max amount gets set and no other value gets changed the `Open Channel...` button stays deactivated.

image


accumulator:
thanks

PR #9704, opened 2025-04-07
tx / gui: Disable output value rounding for 0 fee tx and don't show relay fee warning in gui

f321x:
Disable output value rounding for txs intended to have 0 fees (future txs) in `_change_amounts()` so it doesn't add a small fee due to rounding. 
Also stops showing the below relay fee warning in the gui if the tx is a future tx.

Before:
Screenshot_20250407_114617
After:
Screenshot_20250407_114724

### DIFF
--- a/electrum/coinchooser.py
+++ b/electrum/coinchooser.py
@@ -201,7 +201,10 @@ class CoinChooserBase(Logger):
         # Last change output.  Round down to maximum precision but lose
         # no more than 10**max_dp_to_round_for_privacy
         # e.g. a max of 2 decimal places means losing 100 satoshis to fees
-        max_dp_to_round_for_privacy = 2 if self.enable_output_value_rounding else 0
+        # don't round if the fee estimator is set to 0 fixed fee, so a 0 fee tx remains a 0 fee tx
+        is_zero_fee_tx = True if fee_estimator_numchange(1) == 0 else False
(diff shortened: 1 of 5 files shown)


electrum/coinchooser.py
```diff
@@ -201,7 +201,10 @@ def trailing_zeroes(val):
         # Last change output.  Round down to maximum precision but lose
         # no more than 10**max_dp_to_round_for_privacy
         # e.g. a max of 2 decimal places means losing 100 satoshis to fees
-        max_dp_to_round_for_privacy = 2 if self.enable_output_value_rounding else 0
+        # don't round if the fee estimator is set to 0 fixed fee, so a 0 fee tx remains a 0 fee tx
+        is_zero_fee_tx = True if fee_estimator_numchange(1) == 0 else False
+        output_value_rounding = self.enable_output_value_rounding and not is_zero_fee_tx
+        max_dp_to_round_for_privacy = 2 if output_value_rounding else 0
```
SomberNight, line 207: Could you write a unit test for this into `test_coinchooser.py`? :)

f321x: Added test `test_doesnt_round_output_value_with_zerofee_estimator` in https://github.com/spesmilo/electrum/pull/9704/commits/42bd334f5b381ca4ebbebcf6e81bad7c74958434

PR #9705, opened 2025-04-07
qml: add config option WALLET_FREEZE_REUSED_ADDRESS_UTXOS

accumulator:


electrum/gui/qml/qeconfig.py
```diff
@@ -120,6 +120,16 @@ def spendUnconfirmed(self, checked):
         self.config.WALLET_SPEND_CONFIRMED_ONLY = not checked
         self.spendUnconfirmedChanged.emit()
 
+    freezeReusedAddressUtxosChanged = pyqtSignal()
+    @pyqtProperty(bool, notify=freezeReusedAddressUtxosChanged)
+    def freezeReusedAddressUtxos(self):
+        return not self.config.WALLET_FREEZE_REUSED_ADDRESS_UTXOS
```
SomberNight, line 126: I am confused. Why is this negated?

accumulator: indeed!

electrum/gui/qml/qeconfig.py
```diff
@@ -120,6 +120,16 @@ def spendUnconfirmed(self, checked):
...
         self.spendUnconfirmedChanged.emit()
 
+    freezeReusedAddressUtxosChanged = pyqtSignal()
+    @pyqtProperty(bool, notify=freezeReusedAddressUtxosChanged)
+    def freezeReusedAddressUtxos(self):
+        return self.config.WALLET_FREEZE_REUSED_ADDRESS_UTXOS
+
+    @freezeReusedAddressUtxos.setter
+    def freezeReusedAddressUtxos(self, checked):
+        self.config.WALLET_FREEZE_REUSED_ADDRESS_UTXOS = not checked
```
SomberNight, line 130: Pretty sure this negation is not needed either.

Also, could we call `setDirty` on the address list model, or something to that effect, to get it to refresh?

accumulator: > Pretty sure this negation is not needed either.

Oh my.. :)

> Also, could we call `setDirty` on the address list model, or something to that effect, to get it to refresh?

Yes, however that's a bit less trivial (QE)Config is globally available in QML context, but not in python context..

SomberNight:
Great! Thx.

PR #9710, opened 2025-04-07
config: Raise instead of overwriting the config file on syntax error

f321x:
When there is a syntax error in the config file, even a single trailing `,`, the config is being overwritten by an empty one. Instead i think it is a better idea to raise an exception, considering that the config can contain lots of variables that have to be set up again.

With this pr an exception like this will get raised and the user can just fix the config file:
```
user@fedora:~/code/electrum-fork$ $signet4
  File "/home/user/code/electrum-fork/./run_electrum", line 580, in <module>
    main()
    ~~~~^^
  File "/home/user/code/electrum-fork/./run_electrum", line 375, in main
    tmp_config = SimpleConfig(config_options)
  File "/home/user/code/electrum-fork/electrum/simple_config.py", line 203, in __init__
    self.user_config = read_user_config_function(self.path)
                       ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/simple_config.py", line 873, in read_user_config
    result = json.loads(data)
  File "/usr/lib64/python3.13/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/usr/lib64/python3.13/json/decoder.py", line 345, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/json/decoder.py", line 361, in raw_decode
    obj, end = self.scan_once(s, idx)
               ~~~~~~~~~~~~~~^^^^^^^^
json.decoder.JSONDecodeError: Illegal trailing comma before end of object: line 62 column 38 (char 2223)
  0.27 | W | simple_config | Cannot read config file at /home/user/wallets/signet4/signet/config: Illegal trailing comma before end of object: line 62 column 38 (char 2223)

```


accumulator:
:+1: I've been bitten by this on multiple occasions..

electrum/simple_config.py
```diff
@@ -873,7 +873,8 @@ def read_user_config(path: Optional[str]) -> Dict[str, Any]:
         result = json.loads(data)
     except Exception as exc:
         _logger.warning(f"Cannot read config file at {config_path}: {exc}")
-        return {}
+        raise
```
SomberNight, line 876: If a user is running a binary, e.g. exe on Windows or dmg on macOS, they will not see stderr. For them, Electrum would just silently crash.

Just brainstorming here, but an alternative would be to move the current config file to e.g. `f"config.{unix_time}.bak"`, and then start a new file.

Then again, if this code block raises, they likely manually modified the file or changed file permissions/etc. So maybe it's ~ok to silently crash.

f321x: In https://github.com/spesmilo/electrum/pull/9710/commits/afdbcb29babdb45cb1e1670182d98a4e0e152d73 I  have added a simple gui dialog that shows up if the exception is raised and electrum is started in GUI mode on desktop, on android it will raise as before (which should never happen), this should prevent that issue.

run_electrum
```diff
@@ -372,7 +402,13 @@ def main():
...
         # parse first without plugins
         config_options = parse_command_line(simple_parser=True)
-        tmp_config = SimpleConfig(config_options)
+        try:
+            tmp_config = SimpleConfig(config_options)
+        except Exception as e:
+            if parse_command_line().get("cmd") == 'gui':
+                # show an error dialog on GUI
+                show_simple_qt_error_dialog(e)
+            raise
```
ecdsa, line 411: this feels like a hack. `run_electrum` should be gui agnostic

f321x: Ok i have removed the gui functionality from run_electrum into the gui utils in https://github.com/spesmilo/electrum/pull/9710/commits/2577ae39af749f3d0389309150d20aa110473620.

ecdsa: thanks. the `invalid_config_file` field feels a bit hackish.
why not keep raising an exception, like in the initial version of this PR, and catch it in run_electrum:

define a read_config function and call it from main_loop:
```
    try:
        config_options, config = read_config()
    except InvalidConfigError:
        ... show exception dialog in gui mode, raise in cmdline mode

```

f321x: Ok i now call a read_config method from main, which returns a config or, if the config is invalid, tries to show a gui dialog and then raises an exception.

PR #9713, opened 2025-04-08
wallet,gui: improve not_enough_funds_mentioning_frozen

accumulator:


electrum/wallet.py
```diff
@@ -3417,11 +3417,28 @@ def unlock(self, password):
...
+        if for_amount is not None:
+            if frozen_bal := sum(self.get_frozen_balance()):
+                frozen_str = None
+                if isinstance(for_amount, int):
+                    if frozen_bal + self.get_spendable_balance_sat() > for_amount:
+                        frozen_str = self.config.format_amount_and_units(frozen_bal)
+                elif for_amount == '!':
+                    frozen_str = self.config.format_amount_and_units(frozen_bal)
+                if frozen_str:
+                    text = _('Not enough funds ({} are frozen)').format(frozen_str)
```
SomberNight, line 3439: maybe build this string from the existing parts?
```suggestion
                    text = _('Not enough funds') + " " + _('({} are frozen)').format(frozen_str)
```

electrum/wallet.py
```diff
@@ -3417,11 +3417,28 @@ def unlock(self, password):
...
 
-    def get_text_not_enough_funds_mentioning_frozen(self) -> str:
+    def get_text_not_enough_funds_mentioning_frozen(
+            self,
+            *,
+            for_amount: Optional[Union[int, str]] = None,
+            hint: Optional[str] = None
+    ) -> str:
+        """Generate 'Not enough funds' text.
+        Include mention of frozen coins (and append optional hint), iff unfreezing would satifsy for_amount
```
SomberNight, line 3427:
```suggestion
        Include mention of frozen coins (and append optional hint), iff unfreezing would satisfy for_amount
```

SomberNight (commented):
Looks ok, just trivial comments.

electrum/wallet.py
accumulator: The idea was that building text from fragments is less well translatable. But in this case it's probably ok to concat.

PR #9715, opened 2025-04-08
gui: Add option to show warnings on wallet close, add warning for swaps

f321x:
Allows to register a potential warning shown when the user tries to close the wallet GUI. 
A `callback: Callable[[], Optional[str]]` can be registered using `Abstract_Wallet.register_closing_warning_callback(callback)`.

On QT wallet close the callbacks will get called, if the callback returns a `str` it is presented to the user like this (str is the bottom message):
Screenshot_20250408_162916
 
TODO: implement for QML if this gets merged


accumulator:
On Android this is not really a good option. There's no guarantee the app goes through a confirm-close sequence, as the app can be destroyed at any time by the OS. The current confirm-close is there, and it will save the wallet file, but one can easily close the app through other means.

Ideally the app should run an android background service to monitor chain and lightning events, so the UI can pop in and out of existence, but that would require a big overhaul..

electrum/gui/qt/main_window.py
```diff
@@ -31,6 +31,7 @@
 import csv
 from decimal import Decimal
 import base64
+from email.policy import default
```
ecdsa, line 34: what is this used for?

electrum/gui/qt/main_window.py
```diff
@@ -2619,8 +2620,35 @@ def settings_dialog(self):
...
+                    buttons=buttons,
+                    defaultButton=QMessageBox.StandardButton.Cancel,
+                )
+                if result == QMessageBox.StandardButton.Cancel:
+                    break
+        else:
+            # user chose to cancel all warnings or there were no warnings
+            return True
+        return False
+
```
ecdsa, line 2646: I would prefer to show a single dialog outside the for loop, showing the list of warnings, rather than several dialogs.

electrum/gui/qt/main_window.py
```diff
@@ -2809,7 +2837,16 @@ def on_swap_result(self, txid: Optional[str], *, is_reverse: bool):
...
-                msg += messages.MSG_FORWARD_SWAP_FUNDING_MEMPOOL
+                msg += messages.MSG_FORWARD_SWAP_FUNDING_MEMPOOL + messages.MSG_FORWARD_SWAP_HTLC_EXPIRY
+
+            def wallet_closing_warning_callback() -> Optional[str]:
+                # gets called when the wallet GUI is closed
+                warning = messages.MSG_REVERSE_SWAP_FUNDING_MEMPOOL if is_reverse \
+                    else messages.MSG_FORWARD_SWAP_FUNDING_MEMPOOL
+                if self.wallet.adb.get_tx_height(txid).height < 1:
+                    return _("Ongoing submarine swap") + ": " + warning
+            self.wallet.register_closing_warning_callback(wallet_closing_warning_callback)
```
ecdsa, line 2848: Currently,  `register_closing_warning_callback` is a wallet method, but it is called from the GUI.
If we call it from here, I think it should be a main_window method, and not a wallet method.

OTOH, if we want to keep it as a wallet method (for example, because we want other GUIs to use it) , then I think we should call register callback from `submarine_swap.py`, not from here.

electrum/gui/qt/main_window.py
f321x: Nothing, this got smuggled in by my IDE. removed.

electrum/gui/qt/main_window.py
f321x: As it is used only by QT right now and not really feasible for QML due to the nature of closing apps on mobile (see the comment by @accumulator above) i moved it to main_window. If it turns out we want to use it for something else we can trivially move it.

ecdsa:
it would be nice if we could avoid adding a new persisted dict.
it should be possible to reconstruct what we need from the already existing swap manager data.

electrum/submarine_swaps.py
```diff
@@ -1279,6 +1279,15 @@ def get_group_id_for_payment_hash(self, payment_hash):
         if swap:
             return swap.spending_txid if swap.is_reverse else swap.funding_txid
 
+    def has_pending_swaps(self) -> bool:
```
ecdsa, line 1282: it would be better to return a list of pending swaps, then the GUI can adapt its message depending on the direction of the swap(s)

f321x: Ok, i changed it to return a list and made the warning message adapt to the amount and type of pending swaps.

PR #9717, opened 2025-04-08
swaps: fix race condition causing initialize_swap_manager assert to fail 

f321x:
Fixes race condition that caused the `assert sm.is_initialized.is_set()` in `initialize_swap_manager()` to fail because `is_initialized.set()` is called on the asyncio event loop and the assert was reached before the event was set.  
Note: https://github.com/spesmilo/electrum-aionostr/pull/9 is also required for the timeout value in `initialize_swap_manager()` to be higher than the timeout value awaited when connecting, so this PR alone doesn't completely solve the issue as the current aionostr adds +2s to the timeout val

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Optional, Dict, Union, Sequence, Tuple, Iterab
 from decimal import Decimal
 import math
 import time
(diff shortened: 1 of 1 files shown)


SomberNight:
see https://github.com/spesmilo/electrum/commit/4b9d874d131418c5bc77154b805c354db33cbde4

Issue #9701, opened 2025-04-06
Userspace plugins (version 2)


ecdsa:
### Description

This is a new proposal for userspace plugins.
Previous (reverted) attempt: #8713 . 
Related: #9673 

Reminder: We want to let user install 3rd party plugins, but we want to prevent plugin installation by malware. For this reason, on current `master`, plugins installed by the user must reside in a directory that has root permissions.

This has several disadvantages:
 - plugins reside in a system directory `/opt/electrum_plugins` that is not cleaned up if Electrum in uninstalled.
 - the user must get admin rights every time a plugin in installed/updated/uninstalled. Thus, these operations cannot be performed in Electrum.
 - this creates heterogeneity between platforms (works only on linux/bsd systems for now)

I think we could replace this with a "plugin password", that would offer a similar protection, while allowing the user to install/remove 3rd party plugins from the application:

- User plugins reside in the user's data directory, e.g. `~/.electrum/plugins` on Linux.
- The user chooses a 'plugin password'.
-  An EC keypair is derived from the password: `plugins_privkey/plugins_pubkey`. The derivation should use a salt saved in config.
- `plugins_pubkey` is saved in the system, with root privileges. The exact way it is saved may be platform dependent.
- In order to enable a plugin installed by the user, the user enters his plugin password. The plugin hash (hash of the zip file or recursive hash of the plugin directory) is signed using `plugins_privkey`. This signature is written in the config file, or in a dedicated clear text file.
- Every time a plugin is loaded, we verify the signature against the plugin hash. This disables the plugin if the code has been modified, or if `plugins_pubkey` has changed.

This makes life considerably easier for the user: 
 - they need admin permissions only once, instead of on every plugin install/uninstall. 
 - plugins can be downloaded/removed from the GUI. This reduces the potential for errors.

Having a 'plugins password' does not introduce much complexity, because that password can be lost without adverse consequences; it can be reset with the system's admin password.

PR #9718, opened 2025-04-09
Userspace plugins:

ecdsa:
- Allow plugins saved as zipfiles in user data dir
 - plugins are authorized with a user chosen password
 - pubkey derived from password is saved with admin permissions

implements #9701

### DIFF
--- a/electrum/gui/qt/plugins_dialog.py
+++ b/electrum/gui/qt/plugins_dialog.py
@@ -1,7 +1,7 @@
 from typing import TYPE_CHECKING, Optional
 from functools import partial
 
-from PyQt6.QtWidgets import QLabel, QVBoxLayout, QGridLayout, QPushButton, QWidget, QScrollArea, QCheckBox, QFormLayout
+from PyQt6.QtWidgets import QLabel, QVBoxLayout, QGridLayout, QPushButton, QWidget, QScrollArea, QFormLayout
 
 from electrum.i18n import _
 from electrum.plugin import run_hook
@@ -11,12 +11,13 @@ from .util import WindowModalDialog, Buttons, CloseButton, WWLabel, insert_space
 
 if TYPE_CHECKING:
     from .main_window import ElectrumWindow
+    from electrum_cc import ECPrivkey
 
 
 class PluginDialog(WindowModalDialog):
 
-    def __init__(self, name, metadata, cb: 'QCheckBox', window: 'ElectrumWindow', index:int):
-        display_name = metadata.get('display_name', '')
+    def __init__(self, name, metadata, status_button: Optional['PluginStatusButton'], window: 'ElectrumWindow'):
+        display_name = metadata.get('fullname', '')
         author = metadata.get('author', '')
         description = metadata.get('description', '')
         requires = metadata.get('requires')
@@ -24,14 +25,13 @@ class PluginDialog(WindowModalDialog):
         zip_hash = metadata.get('zip_hash_sha256', None)
 
         WindowModalDialog.__init__(self, window, 'Plugin')
-        self.setMinimumSize(400,250)
-        self.index = index
+        self.setMinimumSize(400, 250)
         self.window = window
         self.metadata = metadata
         self.plugins = self.window.plugins
         self.name = name
-        self.cb = cb
-        p = self.plugins.get(name) # is installed
+        self.status_button = status_button
+        p = self.plugins.get(name)  # is enabled
         vbox = QVBoxLayout(self)
         form = QFormLayout(None)
         form.addRow(QLabel(_('Name') + ':'), QLabel(display_name))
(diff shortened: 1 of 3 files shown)


electrum/gui/qt/plugins_dialog.py
```diff
@@ -11,11 +11,12 @@
...
 if TYPE_CHECKING:
     from .main_window import ElectrumWindow
+    from electrum_cc import ECPrivkey
 
 
 class PluginDialog(WindowModalDialog):
 
-    def __init__(self, name, metadata, cb: 'QCheckBox', window: 'ElectrumWindow', index:int):
+    def __init__(self, name, metadata, status_button: Optional['PluginStatusButton'], window: 'ElectrumWindow'):
         display_name = metadata.get('display_name', '')
```
f321x, line 20: Here you get 'display_name' but in `PluginsDialog.show_list()` you get `fullname`: ```display_name = metadata.get('fullname')```.
So if a plugin (like the virtualkeyboard) only defines fullname its name is not shown in the `PluginDialog` and if it defines `display_name` its not shown in the List?

electrum/gui/qt/plugins_dialog.py
```diff
@@ -70,63 +121,145 @@ def __init__(self, window: 'ElectrumWindow'):
...
+        d = NewPasswordDialog(self, msg=msg)
+        pw = d.run()
+        if not pw:
+            return
+        privkey = self.plugins.derive_privkey(pw)
+        pubkey = privkey.get_public_key_bytes()
+        keyfile_path, keyfile_help = self.plugins.get_keyfile_path()
+        msg = ''.join([
+            _('Your plugins key is:'), '\n\n', pubkey.hex(), '\n\n',
+            _('Please save this key in'), '\n\n' + keyfile_path, '\n\n', keyfile_help
```
f321x, line 180: Screenshot_20250409_123136

I think it would be helpful if this dialog either contained some platform specific hints on how to do this (e.g. some cli commands) or a hyperlink to some documentation explaining it.

It may also be useful to differentiate between the file and the directory in the sentence as the user might confuse `plugins_key` with a directory as there is no visible file extension.
Something like this:
*Please save this key in a file called plugins_key located inside the directory /etc/electrum. Create this file as root user or set the owner to root after creating it using command ...*

ecdsa: The message is actually platform specific. On windows it says to use RegEdit

ecdsa:
Note: the config file structure could be improved w.r.t. plugin data.

We now store:
 - `plugins_salt`: hexadecimal salt
 - `plugin_enabled_<plugin_name>`: boolean
 - `plugin_authorized_<plugin_name>`: hexadecimal signature
 - `<plugin_name>_<configvarname>:` config vars set by plugin

we could instead store a dict named `plugins`, keyed by plugin name.

electrum/gui/qt/plugins_dialog.py
f321x: RE docs, we can link to https://electrum.readthedocs.io/en/latest/plugins.html in this dialog, https://github.com/spesmilo/electrum-docs/pull/154 updates them with guides on how to set up the plugins_key

ecdsa:
> We could also add the addition of a publisher pubkey to the authorization mechanism in the PR, in the same fashion as the plugin.

Yes, we could make the addition a publisher key conditional to the password. but again, what is the point of doing that? like I said, we are not going to grant permissions to that publisher key, so the only thing this could be used for is to show info to the user. Not sure if it is worth the extra complexity.

I am not against doing that, but this is out of scope for this PR. If you want to discuss the plugins architecture, please use #9701

electrum/plugin.py
```diff
@@ -333,6 +371,44 @@ def load_plugin_by_name(self, name: str) -> 'BasePlugin':
     def close_plugin(self, plugin):
         self.remove_jobs(plugin.thread_jobs())
 
+    def derive_privkey(self, pw: str) -> ECPrivkey:
+        from hashlib import pbkdf2_hmac
+        salt = self.config.get('plugins_salt')
+        if salt is None:
+            salt = os.urandom(32).hex()
+            self.config.set_key('plugins_salt', salt)
```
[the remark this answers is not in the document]
ecdsa: done in [edba58b](https://github.com/spesmilo/electrum/pull/9718/commits/edba58ba45c3392d77f01fb243c66df91ed0f678)

electrum/gui/qt/plugins_dialog.py
```diff
@@ -70,63 +121,161 @@ def __init__(self, window: 'ElectrumWindow'):
...
+            _('Its purpose is to prevent unauthorized users (or malware) from installing plugins.'),
+        ])
+        d = NewPasswordDialog(self, msg=msg)
+        pw = d.run()
+        if not pw:
+            return
+        salt = os.urandom(32)
+        privkey = self.plugins.derive_privkey(pw, salt)
+        pubkey = privkey.get_public_key_bytes()
+        key_hex = salt.hex() + pubkey.hex()
```
SomberNight, line 187: I think we should version this. That would for example allow changing the KDF in the future.

Just prefix it with e.g. `b"epp1".hex()`
(short for electrum plugin password)

ecdsa: done, see `PLUGIN_PASSWORD_VERSION`

electrum/plugin.py
```diff
@@ -435,10 +514,11 @@ def run(self):
         self.on_stop()
 
 
-def get_file_hash256(path: str) -> str:
+def get_file_hash256(path: str) -> bytes:
     '''Get the sha256 hash of a file in hex, similar to `sha256sum`.'''
```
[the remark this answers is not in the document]
ecdsa: done

electrum/plugin.py
```diff
@@ -183,35 +184,83 @@ def load_plugins(self):
...
+                    return None, None
+        elif sys.platform in ['linux', 'darwin'] or sys.platform.startswith('freebsd'):
+            if not os.path.exists(self.keyfile_linux):
+                return None, None
+            if not self._has_root_permissions(self.keyfile_linux):
+                return
+            with open(self.keyfile_linux) as f:
+                key_hex = f.read()
+        else:
+            return None, None
```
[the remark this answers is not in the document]
ecdsa: done: detecting windows and android, treating all other platforms as linux-like  (also in `get_keyfile_path`)

PR #9720, opened 2025-04-09
qml: simplify QEConfig and QEDaemon use.

accumulator:
force QEDaemon singleton, and refer to QEDaemon.instance where possible In cases where we would run into circular dependencies, pass the instance

also refer to singleton QEConfig instead of passing instance in qeapp.py

### DIFF
--- a/electrum/gui/qml/qedaemon.py
+++ b/electrum/gui/qml/qedaemon.py
@@ -121,6 +121,8 @@ class QEWalletListModel(QAbstractListModel):
 
 
 class QEDaemon(AuthMixin, QObject):
+    instance = None  # type: Optional[QEDaemon]
+
     _logger = get_logger(__name__)
 
     _available_wallets = None
@@ -149,6 +151,9 @@ class QEDaemon(AuthMixin, QObject):
 
     def __init__(self, daemon: 'Daemon', plugins: 'Plugins', parent=None):
         super().__init__(parent)
+        if QEDaemon.instance:
+            raise RuntimeError('There should only be one QEDaemon instance')
+        QEDaemon.instance = self
         self.daemon = daemon
         self.plugins = plugins
         self.qefx = QEFX(daemon.fx, daemon.config)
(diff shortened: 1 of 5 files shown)


electrum/gui/qml/qedaemon.py
```diff
@@ -149,6 +151,9 @@ class QEDaemon(AuthMixin, QObject):
 
     def __init__(self, daemon: 'Daemon', plugins: 'Plugins', parent=None):
         super().__init__(parent)
+        if QEDaemon.instance:
+            raise RuntimeError('There should only be one QEDaemon instance')
+        QEDaemon.instance = self
```
SomberNight, line 156: In general I think we should only introduce singletons sparingly. E.g. they can make testing significantly more complicated.
However if I understood you correctly, in [another thread](https://github.com/spesmilo/electrum/pull/9705#discussion_r2031486360), you said these are already effectively singletons in the QML context? In which case I guess we might as well make them singletons in python too.

accumulator: The `QE*` variants are singletons, the underlying `SimpleConfig` and `Daemon` not necessarily, but the QML UI uses a single `QEDaemon` and a single `QEConfig`. The backend wallet could well remain non-singleton and allow flexibility, although I think the backend also mostly assumes a single daemon and passes around the same SimpleConfig instance everywhere.

SomberNight: note: in the non-GUI code, atm I think only the Network object is a singleton (and even that I don't really like! :P).
For a while the SimpleConfig used to be a singleton as well, but we managed to untangle that, and the unit tests in particular are now creating configs left-and-right :P

Oh, the asyncio_loop is also a singleton. That is probably also ok, as having multiple asyncio event loops is a bug magnet AFAIU.

But again, if the claim is that QEDaemon and QEConfig were already effectively singletons in QML just not in python, maybe it's fine to make it homogeneous.

PR #9722, opened 2025-04-10
lightning: enable anchor channels by default

f321x:
Now that anchor channels are ready to use and sufficiently tested we could make them default.

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -842,7 +842,7 @@ Warning: setting this to too low will result in lots of payment failures."""),
     )
 
     # anchor outputs channels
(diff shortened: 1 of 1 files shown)


ecdsa:
I am not satisfied with the current 'password in memory' situation for the Qt GUI. 
I think the wallet should not be unlocked by default. Unlocking should be possible, but optional.
If the wallet is locked and a transaction needs to be signed, the status bar should prompt the user for taking action.

The threat model here is a lightning user going away from their computer while the wallet is unlocked. (malware is an issue too, but malware is more likely to steal seed or password using a keylogger, because that is easier for their perspective.)

I think `txbatcher` can be adapted to interact with the GUI. Draft PR: #9751

PR #9723, opened 2025-04-10
lightning: prevent usage of unavailable channels for payment

f321x:
If `get_shortest_path_hops()` finds a channel that is not in `my_sending_channels` but still uses our node id (`nodeA`) as starting node it will include it in the route and cause the payment to fail. Using this route will cause `pay_to_route()` to fail as we try to use a channel from our node that we have no access to (its not in our channel list).

I noticed this in form of an assertion error when trying to pay an invoice from a seed i have opened channels with in different wallet instances (same seed, different wallet).
```
 26.68 | E | lnworker.LNWallet.[default_wallet] | Exception in pay_invoice: AssertionError(<ShortID: 242311x96x2>)
  File "/home/user/code/electrum-fork/electrum/util.py", line 1199, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/lnworker.py", line 1551, in pay_invoice
    await self.pay_to_node(
    ...<11 lines>...
    )
  File "/home/user/code/electrum-fork/electrum/lnworker.py", line 1651, in pay_to_node
    await self.pay_to_route(
    ...<5 lines>...
    )
  File "/home/user/code/electrum-fork/electrum/lnworker.py", line 1724, in pay_to_route
    assert chan, ShortChannelID(short_channel_id)
           ^^^^
AssertionError: 242311x96x2
```

### DIFF
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -526,7 +526,7 @@ class LNPathFinder(Logger):
     def get_shortest_path_hops(
             self,
             *,
-            nodeA: bytes,
+            nodeA: bytes, # nodeA is expected to be our node id if channels are passed in my_sending_channels
             nodeB: bytes,
             invoice_amount_msat: int,
             my_sending_channels: Dict[ShortChannelID, 'Channel'] = None,
@@ -589,10 +589,11 @@ class LNPathFinder(Logger):
                     if not node_filter(edge_startnode, node_info):
                         continue
                 is_mine = edge_channel_id in my_sending_channels
-                if is_mine:
-                    if edge_startnode == nodeA:  # payment outgoing, on our channel
-                        if not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
-                            continue
+                if edge_startnode == nodeA and my_sending_channels:  # payment outgoing, on our channel
+                    if edge_channel_id not in my_sending_channels:
+                        continue
+                    if not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
+                        continue
                 edge_cost, fee_for_edge_msat = self._edge_cost(
                     short_channel_id=edge_channel_id,
                     start_node=edge_startnode,


electrum/lnrouter.py
```diff
@@ -531,7 +531,8 @@ def get_shortest_path_hops(
             invoice_amount_msat: int,
             my_sending_channels: Dict[ShortChannelID, 'Channel'] = None,
             private_route_edges: Dict[ShortChannelID, RouteEdge] = None,
-            node_filter: Optional[Callable[[bytes, NodeInfo], bool]] = None
+            node_filter: Optional[Callable[[bytes, NodeInfo], bool]] = None,
+            our_node_id: Optional[bytes] = None,
```
ecdsa, line 535: do we really need to pass `our_node_id`? it should be the same as `nodeA`

electrum/lnrouter.py
```diff
@@ -593,6 +594,11 @@ def get_shortest_path_hops(
                     if edge_startnode == nodeA:  # payment outgoing, on our channel
                         if not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
                             continue
+                else:
+                    if our_node_id and our_node_id == edge_startnode:
+                        # channel is not in my_sending_channels but startnode is our node,
+                        # this can happen if there is a channel in another wallet with the same node id (same seed)
+                        continue
```
ecdsa, line 601:
```diff
                     if not node_filter(edge_startnode, node_info):
                         continue
                 is_mine = edge_channel_id in my_sending_channels
-                if is_mine:
-                    if edge_startnode == nodeA:  # payment outgoing, on our channel
-                        if not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
-                            continue
+                if edge_startnode == nodeA:  # payment outgoing, on our channel
+                    if edge_channel_id not in my_sending_channels:
+                        continue
+                    if not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
+                        continue
                 edge_cost, fee_for_edge_msat = self._edge_cost(
                     short_channel_id=edge_channel_id,
                     start_node=edge_startnode,

```

f321x: This also fixes the issue. Changed the PR to this patch.

Issue #9531, opened 2025-02-05
Crowdin moved(?) download link for translations, again.


mercurytoxic:
This problem https://github.com/spesmilo/electrum/issues/4559 seems to be happening again. The URL https://crowdin.com/backend/download/project/electrum.zip goes to a login page a the moment.

There is a thread in crowdin that refers to this problem with seemingly no solution: https://community.crowdin.com/t/api-get-project-id-from-slug-download-public-files-without-token/9471

SomberNight:
Indeed, thanks for pointing this out.
Looks like the endpoint still works if I set cookies for my account to authenticate, but it no longer works without auth.

I tried to comment in that thread but the auto-moderator put my account on hold, lol.
Here is the comment I wrote, so that I can maybe try to copy-paste it back later, if they lift the ban:

```
I would like to second OP's point. The `hxxps://crowdin.com/backend/download/project/<PROJECT_NAME>.zip` HTTP endpoint used to not require any authentication and provide a way to download a zip of all translations. This was the case for at least 6 years (since 2018).

Our project has been downloading translations as part of our [build scripts](hxxps://github.com/spesmilo/electrum-locale/blob/39132e77229bf9bc7b1a617b58deab855060d732/update.py#L21) from `hxxps://crowdin.com/backend/download/project/electrum.zip` for many years.
[Before that](hxxps://github.com/spesmilo/electrum/issues/4559), the `hxxps://crowdin.com/download/project/electrum.zip` URL served the same purpose (we used that endpoint [starting 2015](hxxps://github.com/spesmilo/electrum/commit/557286ff792aa687e0dd08f337a9f31a0ab4f0c5), but it was killed in 2018).

To be clear, I just tested, and the `hxxps://crowdin.com/backend/download/project/<PROJECT_NAME>.zip` endpoint works as before *if I provide cookies/etc to authenticate*. However, previously it worked *without* auth.
Could you confirm this is expected?

If so, is there an alternative new endpoint where the translations can be downloaded without auth?
Such functionality used to exist for more than 9 years.

-----

(note: I don't have permissions to post links, hence hxxps)
```

mercurytoxic:
Could you check if the ban has been lifted?

SomberNight:
> Could you check if the ban has been lifted?

I am still banned and no one ever got back to me. I got a message from the "system", "Account temporarily on hold" and it really looks like there is nothing obvious I can do about it. My replies to that message might have ended up in the void.

[...]

PR #9726, opened 2025-04-10
mv git submodule electrum-locale from contrib to electrum/locale

SomberNight:
- this merges `contrib/deterministic-build/locale` and `electrum/locale`
- it is now once again possible have translations when running from a local git clone
    - which was already possible in the past before crowdin removed their unauthenticated APIs
        - see https://github.com/spesmilo/electrum/issues/9531
    - however, the translations available are the often-old frozen strings from electrum-locale
        - while previously one could just download the latest strings from crowdin


contrib/locale/build_cleanlocale.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
+
+CONTRIB_LOCALE="$(dirname "$(readlink -e "$0")")"
```
gruve-p, line 5: The version of readlink installed on macOS does not support the -e flag, so there's a lot of noise when running argbash commands:

```
💬 INFO:  preparing electrum-locale.
readlink: illegal option -- e
usage: readlink [-fn] [file ...]
```

On a mac you can install a version of readlink that argbash expects: Using homebrew, `brew install coreutils` will install `greadlink` which supports the `-e` flag.

gruve-p: @SomberNight ^

SomberNight: Thanks.
See https://github.com/spesmilo/electrum/commit/3a5815d8548b7fe0318de4f7c489a896bebe2f61

PR #9731, opened 2025-04-14
lightning: remove LIGHTNING_LEGACY_ADD_TRAMPOLINE config option

f321x:
The config option (`LIGHTNING_LEGACY_ADD_TRAMPOLINE`) to add a 2nd trampoline hop to legacy trampoline payments requires too much space in the trampoline onion. Because the trampoline onion has only limited space of 400b and the payload with 2nd t hop and no r_tags already requires 392b it is practically unusable for payments that require routing hints to be included in the onion (e.g. because the receiver has no direct channel to the 2nd trampoline node) as there is no space left to include even a single routing hint (1r_tag == ~52b).

As discussed, this pr removes the config option from SimpleConfig and the GUI, however it keeps the logic of use_second_trampoline so we could in theory enable it again if the conditions (available space?) change. I also added a TODO to completely remove the code in 2027 if it is not useful for anything else again until then.

### DIFF
--- a/electrum/gui/qt/settings_dialog.py
+++ b/electrum/gui/qt/settings_dialog.py
@@ -130,19 +130,10 @@ class SettingsDialog(QDialog, QtEventListener):
             else:
                 self.network.run_from_another_thread(
                     self.network.stop_gossip())
-            legacy_add_trampoline_cb.setEnabled(use_trampoline)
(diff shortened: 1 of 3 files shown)


SomberNight (approved):
Looks good!

PR #9733, opened 2025-04-14
lightning: stop including all invoice r_tags in legacy trampoline onion

f321x:
This change makes `create_trampoline_onion()` only include as many of the available r_tags as there is space left in the trampoline onion.

Previously it tried to include all passed invoice r_tags into the trampoline onion payload which caused an user facing exception and payment failure as the onion can only store a max of 400 bytes.
A single, single hop r_tag is ~52 bytes (encoded) and the payload without r_tags is already at ~280 bytes. So usually there is enough space for 2 single hop r_tags.
The implementation shuffles the r_tags on each call so the payment will try different route hints on the attempts (fee level increase or user retry).

It also adds a test which would fail with the previous implementation so this issue (including too many r_tags) would be be noticed.

I have logged the following byte sizes of the trampoline onion (`hop_data.to_bytes() + 32 byte HMAC`) with a 2 trampoline onion hop and changing amounts of r_tags:

3 rtags:
payload size [0]: 113 (hop size: 81)
payload size [1]: 440 (hop size: 295) ( 52 bytes/rtag ) 
payload size [2]: 550 (hop size: 78)

2 rtags:
payload size [0]: 113 (hop size: 81)
payload size [1]: 386 (hop size: 241) ( 52 bytes/rtag ) 
payload size [2]: 496 (hop size: 78)

1 rtag:
payload size [0]: 113 (hop size: 81)
payload size [1]: 334 (hop size: 189) ( 52 bytes/rtag ) 
payload size [2]: 444 (hop size: 78)

0 rtags:
payload size [0]: 113 (hop size: 81)
payload size [1]: 282 (hop size: 137)
payload size [2]: 392 (hop size: 78)

As can be seen in the data, using 2 trampoline hops there is not enough space for even a single r_tag which is why this option is being removed too in https://github.com/spesmilo/electrum/pull/9731.


electrum/trampoline.py
```diff
@@ -305,11 +310,32 @@ def create_trampoline_onion(
...
             payload["payment_data"] = {
                 "payment_secret": payment_secret,
                 "total_msat": total_msat
             }
+
+    if (index := routing_info_payload_index) is not None:
+        # fill the remaining payload space with available routing hints (r_tags)
+        payload: dict = hops_data[index].payload
+        invoice_routing_info: MutableSequence[bytes] = route[index].invoice_routing_info
+        random.shuffle(invoice_routing_info)  # try different r_tag order on each attempt
```
SomberNight, line 323: I don't like side-effecting the `route` argument. As-is, it feels we are doing it for little benefit.
- if you intentionally want to side-effect `route`, because e.g. you want it visible in the `TrampolineEdge` object what routing info was ended up being chosen here, then the loop below should also operate directly on `route[index].invoice_routing_info`. As in, the r_tags that we skip should also be removed from that list.
    - also, in general, unless it's obvious from the function name / list of args / return type, I think it's worth it to signal somehow that an argument is being mutated by a given function. It can e.g. be a note either in the docstring of the function, or a comment near the argument.
    e.g.:
	```py
	def create_trampoline_onion(
		*,
		route: LNPaymentTRoute,  # note: we mutate invoice_routing_info
	```
- alternatively, don't side-effect the `TrampolineEdge`, and e.g. use `util.random_shuffled_copy` to take a copy of the r_tags.

```patch
diff --git a/electrum/lnrouter.py b/electrum/lnrouter.py
index 4e52d48113..73d69711f6 100644
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -25,7 +25,7 @@
 
 import queue
-from typing import Sequence, Tuple, Optional, Dict, TYPE_CHECKING, Set, Callable, MutableSequence
+from typing import Sequence, Tuple, Optional, Dict, TYPE_CHECKING, Set, Callable
 import time
 import threading
 from threading import RLock
@@ -115,7 +115,7 @@ class RouteEdge(PathEdge):
 
 @attr.s
 class TrampolineEdge(RouteEdge):
-    invoice_routing_info = attr.ib(type=MutableSequence[bytes], default=None)
+    invoice_routing_info = attr.ib(type=Sequence[bytes], default=None)
     invoice_features = attr.ib(type=int, default=None)
     # this is re-defined from parent just to specify a default value:
     short_channel_id = attr.ib(default=ShortChannelID(8), repr=lambda val: str(val))
index 953e716334..404d32446e 100644
--- a/electrum/trampoline.py
+++ b/electrum/trampoline.py
@@ -12,6 +12,7 @@ from .lnutil import NoPathFound
 from .lntransport import LNPeerAddr
 from . import constants
+from .util import random_shuffled_copy
 
 
@@ -319,8 +320,8 @@ def create_trampoline_onion(
     if (index := routing_info_payload_index) is not None:
         # fill the remaining payload space with available routing hints (r_tags)
         payload: dict = hops_data[index].payload
-        invoice_routing_info: MutableSequence[bytes] = route[index].invoice_routing_info
-        random.shuffle(invoice_routing_info)  # try different r_tag order on each attempt
+        # try different r_tag order on each attempt
+        invoice_routing_info = random_shuffled_copy(route[index].invoice_routing_info)
         remaining_payload_space = TRAMPOLINE_HOPS_DATA_SIZE \
                                   - sum(len(hop.to_bytes()) + PER_HOP_HMAC_SIZE for hop in hops_data)
         routing_info_to_use = []
```

SomberNight (commented):
Looks good. Just one small comment.

electrum/trampoline.py
f321x: Yes using a copy seems more sensible, changed in https://github.com/spesmilo/electrum/pull/9733/commits/8d79c58c5e54b72be3b2aa545bbaae054cbbcdc1

PR #9740, opened 2025-04-16
swaps: Improve accuracy of swapserver liquidity announcement.

f321x:
Changes the attempt to liquidity updates in the swapserver to a simple loop which regularly (30 sec) checks if the liquidity has changed, and broadcasts a new swapserver announcement if so.
This is simpler and more reliable than trying to trigger an update on each change of liquidity.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -210,7 +210,6 @@ class SwapManager(Logger):
         self.is_server = False # overriden by swapserver plugin if enabled
         self.is_initialized = asyncio.Event()
         self.pairs_updated = asyncio.Event()
-        self._liquidity_changed = asyncio.Event()
 
     def start_network(self, network: 'Network'):
         assert network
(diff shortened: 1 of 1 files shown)


electrum/submarine_swaps.py
```diff
@@ -479,6 +484,7 @@ async def hold_invoice_callback(self, payment_hash: bytes) -> None:
                 output = self.create_funding_output(swap)
                 self.wallet.txbatcher.add_payment_output('swaps', output, self.config.FEE_POLICY_SWAPS)
                 swap._payment_pending = True
+                self.server_maybe_trigger_liquidity_update()
```
ecdsa, line 487: shouldn't this be triggered by htlc added/settled in lnwatcher?

f321x: this would make sense yes, however i intended to keep the liquidity update logic contained in submarine_swaps.py. I now found a more accurate solution by using the event callbacks `on_event_channel` and `on_event_new_transaction`. 
`on_event_channels` triggers for essentially all interesting channel changes that could change the available liquidity.

electrum/submarine_swaps.py
```diff
@@ -974,17 +980,39 @@ def server_maybe_trigger_liquidity_update(self) -> None:
...
-                self.logger.debug(f"liquidity changed, updating announcement")
-                self._liquidity_changed.set()
-                self._liquidity_changed.clear()
+        # set bool to prevent triggering multiple redundant updates at the same time if this gets called in
+        # quick succession (e.g. lnwatcher triggers _claim_swap on new block for multiple swaps)
+        self._ongoing_liquidity_update = True
+        async def trigger_update():
+            # some sleep makes this more reliable as the trigger may is called close to the
+            # action that is supposed to change the liquidity, but the action can take some time to
+            # complete and change the available liquidity data (e.g. settling htlcs after preimage got added)
```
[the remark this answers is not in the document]
f321x: Yes, see fix in other comment. I would still keep the sleep though as the liquidity update is not very time critical and it allows to call `server_maybe_trigger_liquidity_update()` without risking to get rate limited by relays or wasting resources on recalculating the amounts very often.

electrum/submarine_swaps.py
```diff
@@ -240,7 +239,9 @@ async def run_nostr_server(self):
                 await transport.publish_offer(self)
                 try:
                     await wait_for2(
-                        self._liquidity_changed.wait(),
+                        self.server_monitor_liquidity(
```
ecdsa, line 242: a while loop in a while loop? there must be a simpler solution

f321x: moved the liquidity logic into the outer loop and added an update for changed mining fees.

Issue #496, opened 2013-12-11
Option to make each request to server in separate TCP connection


ectomens:
Currently, information about all addresses is requested in one TCP connection, so bad Electrum server can group addresses belonging to one person. This is unwanted in some cases.

I think, we need an option for making each request in separate TCP connection. Then servers located in .onion could not group addresses neither by same TCP connection nor by IP.

ecdsa:
@kyuupichan has implemented interface with select.select instead of threads.
once this is merged, it might be possible to do that

JeremyRand:
Since this issue is fairly old and I don't see a definitive answer, I'd like to ask before we expend resources on implementation: would a PR be accepted that creates an interface for each address subscription (with a randomly chosen server for each of those interfaces)?  The behavior would be optional (maybe the user wants to use a trusted server); whether it's enabled by default is open to negotiation.  It seems like a major privacy benefit, but I'm not really qualified to evaluate whether it would put too much load on the network.  One of the Namecoin developers is interested in implementing this, if the concept is considered mergeable.

Also please note that it's feasible to create connections via multiple Tor exit relays; that's what https://github.com/spesmilo/electrum/pull/5470 does.  So, the suggested attack where a server correlates multiple TCP connections by looking for a common exit IP is not an attack that we need to worry about.

ecdsa:
@JeremyRand it is difficult to assess the privacy benefit of this idea. 
1. with TOR, a single connection per address is an improvement. However, performance might be severely degraded.
2. without TOR, a server can still correlate your IP to a subset of your addresses. So, instead of leaking all your addresses to a single server, you are going to leak part of them it to multiple entities. 
 2.1: if auto-connect is used (i.e. servers are selected randomly), you want to add a strategy that ensures that you disclose the same subset of your addresses to a given server. if you don't do that, you will eventually leak more information to more servers than if you use a single server.
 2.2: if auto-connect is not used, the user needs to be asked which servers will see which part of their addresses, and asked again if a server is unavailable. That seems really bad in terms of usability, unless the "subset of their addresses" if an entire wallet.

SomberNight:
> That seems really bad in terms of usability, unless the "subset of their addresses" if an entire wallet.

Note that currently if a user has multiple wallets open, all addresses in all wallets will be subscribed to using the same interface. So one idea is to at least separate the wallets; use a separate interface for each wallet.

JeremyRand:
[...]

PR #9746, opened 2025-04-17
feat: add manual history fetching mode

thecockatiel:
This small PR aims to add a simple preference for privacy, in which the app does not retrieve any information about any address unless manually requested by user.

Naming stuff is not my strong suit, so feel free to change it as you see fit

I have added a button to the wizard as well to make this setting accessible before the wallet is loaded, so user can opt-in and opt-out at any time, as it is a new feature and users must not need to open wallet in offline mode or use cli to just change this setting

I created `_FETCH` to make this change smaller and easier to implement and review, feel free to change if you have any better idea

~~we can maybe move the `Manual Query Mode` to bottom of the window as an icon, since the balance is there as well~~ applied

related:
https://github.com/spesmilo/electrum/issues/496#issuecomment-44718842


SomberNight:
Could you list some example use cases for this? How would it be used and by what kind of users?
I guess it must be for a narrow niche of users with high technical understanding but who don't run their own server...

Also note: if you have LN channels, it part of the security model that you learn about e.g. your channel-counterparty force-closing the channel within some number of blocks. So we would need to treat lnwatcher differently. Or maybe disallow enabling this feature in a wallet that has lightning channels. (LNWatcher and Abstract_Wallet share the same `adb`, hence the same Synchronizer/Verifier)

thecockatiel:
For those who don't want to let a single Electrum server know all of their other addresses
when you send a group of addresses together to a single server, the server can send this information for analysis and now all of your addresses are tied together to a single identity

It's most likely only technical users at the moment who know about the details, but what's the harm in adding such feature, it may cause other non-technical users to become curious and go and see what it is in details and end up using it. It's not that complicated either, my first paragraph is simple enough for most users to understand it I believe.

about LN channels, I don't know how LN stuff work, so I don't have any idea about what to do about that. how critical is it and how fast does things happen in LN ? is it possible to let user refresh the related addresses with an extra button or some other way? or limit actions when this is enabled and require manual refresh in certain circumstances that need it? after-all the whole point of this PR is to make the control of retrieving information about an address manual so user can decide when to do that

thecockatiel:
@SomberNight @accumulator @ecdsa 
any comments ? can we have this feature? is there anything I can do to get this merged?

thecockatiel:
this is less than a hundred line of change, and if it's for lightning, we can disable it for these users. this is not complex
I don't see why an option adds complexity to entirety of the user base

why should I be forced to run a node, when such solution can exist to enhance privacy without too much effort? not everyone can run and use a private node correctly, but everyone can understand "not sending all of your addresses at once helps your privacy" and just use the option

I'm not picking a fight here, I just really want to contribute features to electrum that will boost it's privacy reputation, even if it's small, because it's a really good software and I love it :)

accumulator:
> this is less than a hundred line of change, and if it's for lightning, we can disable it for these users. this is not complex I don't see why an option adds complexity to entirety of the user base

What you're introducing here is a potential incomplete view of the wallet state w.r.t the on-chain reality. Huge can of worms.

> why should I be forced to run a node, when such solution can exist to enhance privacy without too much effort? not everyone can run and use a private node correctly, but everyone can understand "not sending all of your addresses at once helps your privacy" and just use the option

Servers can over time correlate all these partial address set requests anyway, so it is at best a false sense of security to have this as an option.

> I'm not picking a fight here, I just really want to contribute features to electrum that will boost it's privacy reputation, even if it's small, because it's a really good software and I love it :)

We appreciate the effort, but this is simply not a good solution

thecockatiel:
> Servers can over time correlate all these partial address set requests anyway, so it is at best a false sense of security to have this as an option.

this can be improved further, by only asking servers the same address and avoid sending other addresses to the same server.

but I understand that this can, and will probably cause a lot of confusion for users if not implemented with care

so how about introducing a different kind of syncing as an option, where we download blocks from public nodes and take related information out of it, just like other wallets do for privacy? it will increase traffic, but it will increase privacy a lot, and it wont have the same issue of an incomplete view of the wallet state
would you be open to such contribution ? (as an extra feature, living as an option alongside the current behavior)

thecockatiel:
since this PR would be used by advanced users mostly, can we at least have this behind a flag, like by setting a special env variable, instead of settings ? that way normal users won't get confused either

thecockatiel:
how does one write middlewares for Electrum ? I don't remember seeing something like that in the code base
any pointers is appreciated

ecdsa:
by middleware I mean something that simulates an electrum server, like EPS.

PR #9749, opened 2025-04-18
add harden_memory_linux: harder for other processes to read our memory

SomberNight:
This module tries to restrict the ability of other processes to access the memory of our process. This is Linux-specific.

The immediate motivation is that, when having LN channels with anchor outputs, we need to be able to spend onchain utxos to bring fees to e.g. second-stage htlc transactions or to spend anchors, and for this the current idea is to keep the wallet password in memory.

-----

Traditionally, on Linux, one process can access the memory of another arbitrary process if both are running as the same user (uid). (Root can ofc access the memory of ~any process)
Programs can opt-out from this by setting `prctl(PR_SET_DUMPABLE, 0);`

- see longer comment at the top of the new module
- see https://man.archlinux.org/man/PR_SET_DUMPABLE.2const.en

-----

Also, from https://unix.stackexchange.com/a/518452 :

In a terminal window:
```
% echo $$  # show our pid
6744
% read -sp 'secret password: '; echo
secret password:
%
```
Then in another terminal window:
```
% grep heap /proc/6744/maps
01bb7000-01c3e000 rw-p 00000000 00:00 0                                  [heap]
% dd if=/proc/6744/mem bs=1 skip=$((0x01bb7000)) count=$((0x01c3e000-0x01bb7000)) status=none |
    strings | less
...
% dd if=/proc/6744/mem bs=1 skip=$((0x01bb7000)) count=$((0x01c3e000-0x01bb7000)) status=none |
    strings | grep obiwan
obiwan_kenobi  # "secret password"
```

### DIFF
--- a/electrum/harden_memory_linux.py
+++ b/electrum/harden_memory_linux.py
@@ -1,0 +1,94 @@
+# Copyright (C) 2020 cptpcrd
+# Copyright (C) 2025 The Electrum developers
+# Distributed under the MIT software license, see the accompanying
+# file LICENCE or http://www.opensource.org/licenses/mit-license.php
+#
+# based on https://github.com/cptpcrd/pyprctl/blob/578ed3e81066a8a61dede912454d5eeaef37eeea/pyprctl/ffi.py#L28
+#
+# This module tries to restrict the ability of other processes to access the memory of our process.
+# Traditionally, on Linux, one process can access the memory of another arbitrary process
+# if both are running as the same user (uid). (Root can ofc access the memory of ~any process)
+# Programs can opt-out from this by setting prctl(PR_SET_DUMPABLE, 0);
+#
+# Besides PR_SET_DUMPABLE, there are ways to globally restrict this for all processes:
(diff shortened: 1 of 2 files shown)


electrum/harden_memory_linux.py
```diff
@@ -0,0 +1,94 @@
...
+        _logger.error(f"libc.prctl(PR_SET_DUMPABLE, {flag}) errored: {e}")
+
+
+def get_dumpable() -> bool:
+    _load_libc()
+    res = _libc.prctl(PR_GET_DUMPABLE, 0, 0, 0, 0)
+    if res < 0:
+        eno = ctypes.get_errno()
+        raise OSError(eno, os.strerror(eno), None, None, None)
+    return res != 0
```
accumulator, line 94: This func is never used? Maybe call this right after `set_dumpable()` above to verify it was actually set?

accumulator:
Tested AppImage and on debian w. 6.14, no issues.

PR #9755, opened 2025-04-22
plugin: psbt_nostr: use proxy in psbt cosigning plugin

f321x:
Use the network proxy in the psbt cosigning plugin for nostr relay connections.

### DIFF
--- a/electrum/plugins/psbt_nostr/psbt_nostr.py
+++ b/electrum/plugins/psbt_nostr/psbt_nostr.py
@@ -38,11 +38,13 @@ from electrum.i18n import _
 from electrum.logging import Logger
 from electrum.plugin import BasePlugin
 from electrum.transaction import PartialTransaction, tx_from_any
-from electrum.util import log_exceptions, OldTaskGroup, ca_path, trigger_callback, event_listener
+from electrum.util import (log_exceptions, OldTaskGroup, ca_path, trigger_callback, event_listener,
+                           make_aiohttp_proxy_connector)
 from electrum.wallet import Multisig_Wallet
 
 if TYPE_CHECKING:
     from electrum.wallet import Abstract_Wallet
+    from aiohttp_socks import ProxyConnector
 
 # event kind used for nostr messages (with expiration tag)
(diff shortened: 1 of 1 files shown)


electrum/plugins/psbt_nostr/psbt_nostr.py
```diff
@@ -123,6 +125,15 @@ def on_event_wallet_updated(self, wallet):
             self.logger.debug('starting handling of PSBTs')
             self.wallet_uptodate.set()
 
+    @event_listener
+    async def on_event_proxy_set(self, *args):
+        if not (self.network and self.nostr_pubkey):
+            return
+        await self.stop()
+        await asyncio.sleep(2)
```
accumulator, line 133: is this sleep necessary? According to `self.stop() -> taskgroup.cancel_remaining()` the function only returns once all tasks have ended.

accumulator: Other than this minor remark, PR LGTM

f321x: I mostly put the sleep there to leave the relays some time to cleanup the subscription and leave other, more important tasks (like ln peers) some compute after switching proxy. I'll remove it to simplify this.

accumulator: Ah I see. Well if this leads to races in other parts of the code I'd rather know about it ASAP :) 
Issues with slow relays might still happen (if I understood you correctly), but that should I think be preferably handled by the aionostr manager code

accumulator: > I don't know of any specific races/issues, one thing i noticed that there is a lots of stuff going on when toggling the proxy, causing noticeable lag in the gui (the checkbox takes some time to check/uncheck), and maybe letting some less important things sleep for a some time would distribute this load a bit. 

It's probably either ending up from `trigger_callback('proxy_set')` in long-running blocking calls on the UI thread, or it triggers `main_window.update_tabs()`, which also takes some time while blocking the UI.

PR #9757, opened 2025-04-23
qml: add nostr relay config in NetworkOverview and add relay edit dialog

accumulator:


electrum/gui/qml/components/NostrConfigDialog.qml
```diff
@@ -0,0 +1,118 @@
...
+
+    property bool valid: true
+
+    function clean_array(text) {
+        var relays = []
+        const fragments = text.split("\n")
+        fragments.forEach(function(fragment) {
+            fragment = fragment.trim()
+            if (fragment != "")
+                relays.push(fragment)
```
f321x, line 29: Maybe clean_array should do some deduplication of the relay urls? Its possible to add the same relay multiple times.

f321x:
Works well on my Pixel 9.
Is there a reason we still set the fallback URL for http swaps on android (in run_electrum) but not for qt, or could this be unified?

electrum/gui/qml/components/NostrConfigDialog.qml
accumulator: Since dedup is also implemented in the desktop GUI, maybe we should consolidate this instead to the `ConfigVar` setter (and then read back the config value for the GUI)?

accumulator:
> Works well on my Pixel 9. Is there a reason we still set the fallback URL for http swaps on android (in run_electrum) but not for qt, or could this be unified?

No, I think this is a leftover. It effectively blocks using Nostr based swaps in `swap_manager.create_transport()`. However, I could still see the use of a HTTP fallback in case no Nostr swap providers are present, which is currently not possible as the values of these config variables are implicit toggles :(

PR #9765, opened 2025-04-28
plugins: add functionality to allow setting plugin pubkey from gui

f321x:
Adds functionality that allows the user to store the plugin authorization pubkey without having to edit files/registry manually. On Linux systems it spawns the commands in a subprocess with polkits pkexec which will trigger an OS prompt to execute the commands as root. The user sees the executed commands and can either authorize with the root password or decline.
On windows it uses the windows `ShellExecuteExW` api to execute reg.exe with elevated permissions to edit the registry, this also triggers an OS dialog the user can accept or decline (UAC dialog).
There is also functionality to reset the key again, which works in the same way.
I tested this from source on both plattforms and as AppImage and Windows installer.
On macos it wraps the shell command in applescript and runs the applescript as subprocess which will also start a OS prompt the user has to accept.

The windows OS prompt:

Screenshot_20250428_145953

The Linux (KDE) prompt:

Screenshot_20250428_160719

### DIFF
--- a/electrum/gui/qt/plugins_dialog.py
+++ b/electrum/gui/qt/plugins_dialog.py
@@ -3,7 +3,8 @@ from functools import partial
 import shutil
 import os
 
-from PyQt6.QtWidgets import QLabel, QVBoxLayout, QGridLayout, QPushButton, QWidget, QScrollArea, QFormLayout, QFileDialog, QMenu, QApplication
+from PyQt6.QtWidgets import QLabel, QVBoxLayout, QGridLayout, QPushButton, QWidget, QScrollArea, \
+    QFormLayout, QFileDialog, QMenu, QApplication, QMessageBox
 from PyQt6.QtCore import QTimer
 
 from electrum.i18n import _
@@ -15,7 +16,7 @@ from .util import (WindowModalDialog, Buttons, CloseButton, WWLabel, insert_spac
 
 if TYPE_CHECKING:
     from . import ElectrumGui
-    from electrum_cc import ECPrivkey
+    from electrum_ecc import ECPrivkey
     from electrum.simple_config import SimpleConfig
     from electrum.plugin import Plugins
 
@@ -174,6 +175,7 @@ class PluginsDialog(WindowModalDialog, MessageBoxMixin):
         scroll_w.setLayout(self.grid)
         vbox.addWidget(scroll)
         add_button = QPushButton(_('Add'))
+        add_button.setMinimumWidth(40)  # looks better on windows, no difference on linux
         menu = QMenu()
         for name, item in self.plugins.internal_plugin_metadata.items():
             fullname = item['fullname']
@@ -194,7 +196,7 @@ class PluginsDialog(WindowModalDialog, MessageBoxMixin):
         pubkey, salt = self.plugins.get_pubkey_bytes()
         if not pubkey:
             self.init_plugins_password()
-            return
+            return None
         # ask for url and password, same window
         pw = self.password_dialog(
             msg=' '.join([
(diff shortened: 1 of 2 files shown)


electrum/plugin.py
```diff
@@ -199,6 +199,178 @@ def get_keyfile_path(self) -> Tuple[str, str]:
...
+            assert f.read() == key_hex, f'file content mismatch: {f.read()} != {key_hex}'
+        self.logger.debug(f'file saved successfully to {self.keyfile_linux}')
+
+    def _write_key_to_regedit_windows(self, key_hex: str) -> None:
+        """
+        Writes the key to the Windows registry with windows UAC prompt.
+        """
+        from winreg import ConnectRegistry, OpenKey, QueryValue, HKEY_LOCAL_MACHINE
+
+        key_path = r'HKEY_LOCAL_MACHINE\SOFTWARE\Electrum\PluginsKey'
```
SomberNight, line 255: can this be deduped somehow with `self.keyfile_windows`?

electrum/gui/qt/plugins_dialog.py
```diff
@@ -243,7 +260,22 @@ def init_plugins_password(self):
         ])
         clipboard = QApplication.clipboard()
         clipboard.setText(key_hex)
-        self.show_message(msg)
+
+        auto_setup_btn = QPushButton(_('Try Auto-Setup'))
+        def on_auto_setup_clicked():
+            if not self.plugins.try_auto_key_setup(key_hex):
+                self.show_error(_("Auto-Setup not possible. Try the manual setup."))
```
SomberNight, line 267: ideally we should still show the manual setup instructions if this happens, but the popup gets closed

f321x: Fixed this, it now also shows a shell command in the prompt that can be used to setup the key.

electrum/gui/qt/plugins_dialog.py
```diff
@@ -243,10 +266,30 @@ def init_plugins_password(self):
...
-        self.show_message(msg)
+
+        while True:
+            exit_dialog = True
+            # the button has to be recreated inside the loop, as qt destroys it when the dialog is closed
+            auto_setup_btn = QPushButton(_('Try Auto-Setup'))
+            def on_auto_setup_clicked():
+                nonlocal exit_dialog
+                if not self.plugins.try_auto_key_setup(key_hex):
+                    self.show_error(_("Auto-Setup not possible. Try the manual setup."))
```
ecdsa, line 277: This popup is not needed after clicking cancel

electrum/gui/qt/plugins_dialog.py
```diff
@@ -208,18 +210,39 @@ def get_plugins_privkey(self) -> Optional['ECPrivkey']:
...
-            return
+            keyfile_path, _keyfile_help = self.plugins.get_keyfile_path(None)
+
+            while True:
+                exit_dialog = True
+                auto_reset_btn = QPushButton(_('Try Auto-Reset'))
+                def on_try_auto_reset_clicked():
+                    nonlocal exit_dialog
+                    if not self.plugins.try_auto_key_reset():
+                        self.show_error(_("Auto-Reset not possible. Delete the file manually."))
```
ecdsa, line 224: same thing here

electrum/gui/qt/plugins_dialog.py
f321x: The issue is that the (OS) dialog is not in our control, so we only see the status code of the spawned subprocess as response, which is the same (127 on linux) for entering a wrong password or clicking cancel (and maybe other things that potentially can go wrong depending on the environment), so from the applications perspective we cannot differentiate what went wrong exactly.

Alternatively we could just not show it in either case (wrong password or clicking cancel), but that would be confusing as the user then may think it worked when entering the wrong password.

ecdsa: ah right

PR #9766, opened 2025-04-28
qt: pass Invoice object to transaction dialog when appropriate.

accumulator:
This is purely informational and optional, with the main immediate use to provide the invoice description/message/label to the transaction dialog, so it can be stored when saving the tx in history, or passed along with PSBTs sent to cosigners.

Before, the tx description was not saved in history when an invoice was not saved before signing and saving the tx for sending later.


accumulator:
closing, included in #9793

PR #9773, opened 2025-04-30
qt: validate and deduplicate relay config input in qt gui

f321x:
Adds validation and deduplication of the relay urls entered in the QT settings dialog. This is supposed to prevent malformed or duplicated relay entries.
Also resets the relays to the default value if no (valid) url is entered. This prevents the user from getting stuck without relays (otherwise the user would have to research for relay urls manually if they don't know any).


electrum/gui/qt/settings_dialog.py
```diff
@@ -181,7 +181,16 @@ def lnfee_slider_released():
...
-            self.config.NOSTR_RELAYS = str(self.nostr_relays_e.text())
+            relays: Dict[str, None] = dict()  # dicts keep insertion order
+            for url in self.nostr_relays_e.text().split(','):
+                url = url.strip()
+                if url and is_valid_websocket_url(url):
+                    relays[url] = None
+            if relays.keys():
+                self.config.NOSTR_RELAYS = ",".join(relays.keys())
+            else:
+                self.config.NOSTR_RELAYS = self.config.cv.NOSTR_RELAYS.get_default_value()
```
accumulator, line 192: You can do a `self.config.NOSTR_RELAYS = None` to let it fall back to the default

accumulator (approved):

electrum/gui/qt/settings_dialog.py
f321x: Ok, doing this now in https://github.com/spesmilo/electrum/pull/9773/commits/ee7d2ee17daeca9ee4af08208d39ba720a344672

PR #9780, opened 2025-05-01
QML send change to lightning option

accumulator:


electrum/gui/common_qt/swaps.py
```diff
@@ -0,0 +1,209 @@
...
+                    )
+                elif change_amount > max_amount:
+                    msg = _('Change amount exceeds the swap providers maximum value of {}.').format(
+                        self.config.format_amount_and_units(max_amount)
+                    )
+            else:
+                msg = _('Will not send change to Lightning')
+        return msg
+
+    # --- Shared functionality for submarine swaps (change to ln and submarine payments) ---
```
ecdsa, line 108: The methods related to the swap transport should be ported to the `submarine_swaps.py` in the backend, rather than to `gui/common_qt`. If possible, maybe also do the same with `get_message_for_swap_change`: in that case this `common_qt/swaps.py` file will not be needed anymore.

I suggest to make this refactoring (including the changes to qt/confirm_tx_dialog.py) as a separate pull request.

accumulator:
> The methods related to the swap transport should be ported to the `submarine_swaps.py` in the backend, rather than to `gui/common_qt`. If possible, maybe also do the same with `get_message_for_swap_change`: in that case this `common_qt/swaps.py` file will not be needed anymore.
> 
> I suggest to make this refactoring (including the changes to qt/confirm_tx_dialog.py) as a separate pull request.

closing this PR, I will add a new PR as a followup to https://github.com/spesmilo/electrum/pull/10694

Issue #9700, opened 2025-04-06
Submarine swaps don't work: stuck at "Awaiting swap payment"


ln2max:
### Description

Now that #9698 is fixed, submarine swaps get stuck a little bit further:

When I try to use submarine swaps to convert on-chain into Lightning Bitcoins, I get a message "Awaiting swap payment...". Eventually it times out and the swap fails due to lack of payment.

Using Electrum 4.5.8 appImage.

VeneerBrad:
Hello @ln2max 
Send an email to; elm76609@gmail.com to assist on this

ln2max:
> Hello [@ln2max](https://github.com/ln2max) Send an email to; 

This is a scam / spam post @ecdsa

ecdsa:
> When I try to use submarine swaps to convert on-chain into Lightning Bitcoins, I get a message "Awaiting swap payment...". Eventually it times out and the swap fails due to lack of payment.

this looks like a path finding/liquidity issue.
is your node well connected?

ln2max:
> > When I try to use submarine swaps to convert on-chain into Lightning Bitcoins, I get a message "Awaiting swap payment...". Eventually it times out and the swap fails due to lack of payment.
> 
> this looks like a path finding/liquidity issue. is your node well connected?

I'm using the Electrum trampoline so I hope the answer is "yes" :)

ecdsa:
Checking the Eclair log, I see some errors:

[...]

PR #9781, opened 2025-05-02
lightning: verify single payment limits for swaps and invoices

accumulator:
fix for #9700

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1099,29 +1099,41 @@ class Channel(AbstractChannel):
             if amount_msat < chan_config.htlc_minimum_msat:
                 raise PaymentFailure(f'HTLC value too small: {amount_msat} msat')
 
+        if self.too_many_htlcs(htlc_proposer, strict):
+            raise PaymentFailure('Too many HTLCs already in channel')
+
+        if amount_msat > self.remaining_max_inflight(htlc_receiver):
+            raise PaymentFailure(
+                f'HTLC value sum (sum of pending htlcs plus new htlc) '
+                f'would exceed max allowed: {chan_config.max_htlc_value_in_flight_msat/1000} sat')
+
         # check proposer can afford htlc
         max_can_send_msat = self.available_to_spend(htlc_proposer, strict=strict)
         if max_can_send_msat < amount_msat:
             raise PaymentFailure(f'Not enough balance. can send: {max_can_send_msat}, tried: {amount_msat}')
 
+    def too_many_htlcs(self, htlc_proposer: HTLCOwner, strict:bool) -> bool:
         # check "max_accepted_htlcs"
+        htlc_receiver = htlc_proposer.inverted()
+        ctn = self.get_next_ctn(htlc_receiver)
+        chan_config = self.config[htlc_receiver]
         # this is the loose check BOLT-02 specifies:
         if len(self.hm.htlcs_by_direction(htlc_receiver, direction=RECEIVED, ctn=ctn)) + 1 > chan_config.max_accepted_htlcs:
-            raise PaymentFailure('Too many HTLCs already in channel')
+            return True
         # however, c-lightning is a lot stricter, so extra checks:
         # https://github.com/ElementsProject/lightning/blob/4dcd4ca1556b13b6964a10040ba1d5ef82de4788/channeld/full_channel.c#L581
         if strict:
             max_concurrent_htlcs = min(self.config[htlc_proposer].max_accepted_htlcs,
                                        self.config[htlc_receiver].max_accepted_htlcs)
             if len(self.hm.htlcs(htlc_receiver, ctn=ctn)) + 1 > max_concurrent_htlcs:
-                raise PaymentFailure('Too many HTLCs already in channel')
+                return True
+        return False
 
+    def remaining_max_inflight(self, htlc_receiver: HTLCOwner) -> int:
         # check "max_htlc_value_in_flight_msat"
+        ctn = self.get_next_ctn(htlc_receiver)
         current_htlc_sum = htlcsum(self.hm.htlcs_by_direction(htlc_receiver, direction=RECEIVED, ctn=ctn).values())
-        if current_htlc_sum + amount_msat > chan_config.max_htlc_value_in_flight_msat:
-            raise PaymentFailure(f'HTLC value sum (sum of pending htlcs: {current_htlc_sum/1000} sat '
-                                 f'plus new htlc: {amount_msat/1000} sat) '
-                                 f'would exceed max allowed: {chan_config.max_htlc_value_in_flight_msat/1000} sat')
+        return self.config[htlc_receiver].max_htlc_value_in_flight_msat - current_htlc_sum
 
     def can_pay(self, amount_msat: int, *, check_frozen=False) -> bool:
         """Returns whether we can add an HTLC of given value."""
@@ -1541,10 +1553,7 @@ class Channel(AbstractChannel):
                             ),
         )
 
-        ctn = self.get_next_ctn(receiver)
-        current_htlc_sum = htlcsum(self.hm.htlcs_by_direction(receiver, direction=RECEIVED, ctn=ctn).values())
-        remaining_max_inflight = self.config[receiver].max_htlc_value_in_flight_msat - current_htlc_sum
-        max_send_msat = min(max_send_msat, remaining_max_inflight)
+        max_send_msat = min(max_send_msat, self.remaining_max_inflight(receiver))
 
         max_send_msat = max(max_send_msat, 0)
         return max_send_msat
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1540,6 +1540,12 @@ class Channel(AbstractChannel):
                                 consider_ctx(ctx_owner=sender, is_htlc_dust=False),
                             ),
         )
+
+        ctn = self.get_next_ctn(receiver)
+        current_htlc_sum = htlcsum(self.hm.htlcs_by_direction(receiver, direction=RECEIVED, ctn=ctn).values())
+        remaining_max_inflight = self.config[receiver].max_htlc_value_in_flight_msat - current_htlc_sum
+        max_send_msat = min(max_send_msat, remaining_max_inflight)
+
         max_send_msat = max(max_send_msat, 0)
         return max_send_msat
 
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1088,10 +1088,6 @@ class Channel(AbstractChannel):
             if not self.can_send_update_add_htlc():
                 raise PaymentFailure('Channel cannot add htlc')
 
-        # If proposer is LOCAL we apply stricter checks as that is behaviour we can control.
-        # This should lead to fewer disagreements (i.e. channels failing).
-        strict = (htlc_proposer == LOCAL)
-
         # check htlc raw value
         if not ignore_min_htlc_value:
             if amount_msat <= 0:
@@ -1099,7 +1095,7 @@ class Channel(AbstractChannel):
             if amount_msat < chan_config.htlc_minimum_msat:
                 raise PaymentFailure(f'HTLC value too small: {amount_msat} msat')
 
-        if self.too_many_htlcs(htlc_proposer, strict):
+        if self.too_many_htlcs(htlc_proposer):
             raise PaymentFailure('Too many HTLCs already in channel')
 
         if amount_msat > self.remaining_max_inflight(htlc_receiver):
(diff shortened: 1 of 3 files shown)


ecdsa:
can you provide a description? how is this related to #9700?

accumulator:
> can you provide a description? how is this related to #9700?

This split is to allow `lnworker.num_sats_can_{receive,send}` to use `lnchannel.determine_htlc_constraints` to calculate the available balance for a single payment.

The idea is to extend `lnworker.num_sats_can_{receive,send}` with extra kwargs so it can determine limits on a single payment, e.g. by adding the kwargs `single_payment:bool` and `mpp:bool`

electrum/lnchannel.py
```diff
@@ -1091,32 +1095,71 @@ def _assert_can_add_htlc(self, *, htlc_proposer: HTLCOwner, amount_msat: int,
...
+        htlc_receiver = htlc_proposer.inverted()
+        ctn = self.get_next_ctn(htlc_receiver)
+        chan_config = self.config[htlc_receiver]
+        if self.get_state() != ChannelState.OPEN:
+            raise PaymentFailure('Channel not open', self.get_state())
+        if htlc_proposer == LOCAL:
+            if not self.can_send_ctx_updates():
+                raise PaymentFailure('Channel cannot send ctx updates')
+            if not self.can_send_update_add_htlc():
+                raise PaymentFailure('Channel cannot add htlc')
```
ecdsa, line 1126: I would not raise `PaymentFailure` inside `determine_payment_constraints`, because that method is called by `num_sats_can_send/receive`. (those `PaymentFailures` are triggered by network conditions, while `num_sat_can_send/receive` should return stable results). It also simplifies  the upstream code if you do not have to catch `PaymentFailure`

accumulator: Yes, I left this in to keep the behavior of `num_sats_can_xxx` the same as I took the constraints part out, and I need the error path using exceptions to propagate the error message out.

I think it feels more appropriate if I raise e.g. `ChannelConstraintFailure` or something, catch these in `num_sats_can_xxx`, and re-raise there with the same error message, but in a `PaymentFailed` class.

accumulator: > It also simplifies the upstream code if you do not have to catch `PaymentFailure`

Agreed. The exceptions are useful within the context of `num_sats_can_xxx`, but not so much in other places. I can add a convenience method that just returns `None`

ecdsa: The way the current code works is that `get_channels_for_sending/receiving` filters the channels that cannot send due to transient network conditions. If you need to add new checks, please them where the existing checks are.

electrum/lnworker.py
```diff
@@ -2717,7 +2717,13 @@ def send_capacity(chan):
...
             else:
                 delta_msat = 0
-            return chan.available_to_spend(LOCAL) + delta_msat
+            if single_payment:
+                try:
+                    return chan.determine_payment_constraints(LOCAL).htlc_maximum_msat + delta_msat
+                except PaymentFailure:
+                    return 0
+            else:
+                return chan.available_to_spend(LOCAL) + delta_msat
```
ecdsa, line 2726: why not pass single_payment to `chan.available_to_spend`, and let that method do the if branching?

That way, you could write a unit test directly in test_lnchannel. Currently, a unit test would need to be added to test_lnpeer. However, the property we are testing does not really belong in lnpeer, it is lower level.

accumulator: Done in e83055c4339024ac52e4cf02a9cedef577632906

ecdsa: ok, I see you have removed `determine_payment_constraints`, that looks good. However, in `_assert_can_add_htlc`, you are calling `available_to_spend` without passing `single_payment=True.` I guess that is oversight?

You should really add a unit test in test_lnchannel, that tests adding an htlc above max_inflight. It would have detected that.

accumulator: Yeah, sorry, I should've put this to draft while working on this, it was not finished yet :)

accumulator:
Note: in the `channel_update` message peers can send `htlc_maximum_msat`, which is the max amount a single HTLC can have. This is currently never used AFAICS

tests/test_lnchannel.py
```diff
@@ -721,6 +730,59 @@ def test_DesyncHTLCs(self):
         self.assertEqual(500000000000, bob_channel.available_to_spend(LOCAL))
         alice_channel.add_htlc(htlc)
 
+    def test_single_payment(self):
+        alice_channel, bob_channel = create_test_channels(
+            anchor_outputs=self.TEST_ANCHOR_CHANNELS,
+            local_msat=bitcoin.COIN * 40,
+            remote_msat=bitcoin.COIN * 40,
+            local_max_inflight=bitcoin.COIN * 10,
+            remote_max_inflight=bitcoin.COIN * 20)
```
ecdsa, line 739: This is very confusing. 
If we had proper types, bitcoin.COIN would be defined as sats, not as an integer. 
It what you want is 40 mBTC expressed in msats, it would be less confusing to write it as 4000000000.

electrum/lnchannel.py
```diff
@@ -1513,19 +1513,25 @@ def consider_ctx(*, ctx_owner: HTLCOwner, is_htlc_dust: bool) -> int:
...
                     if new_receiver_balance < 0:
-                        return 0
-                    return max_send_msat
+                        max_send_msat = 0
+
+            if single_payment:  # check max_htlc_value_in_flight_msat
+                current_htlc_sum = htlcsum(self.hm.htlcs_by_direction(receiver, direction=RECEIVED, ctn=ctn).values())
+                remaining_max_inflight = self.config[receiver].max_htlc_value_in_flight_msat - current_htlc_sum
+                max_send_msat = min(max_send_msat, remaining_max_inflight)
+
```
ecdsa, line 1533: why is this inside the consider_ctx function?

accumulator: indeed, moved.

tests/test_lnchannel.py
```diff
@@ -138,6 +141,8 @@ def create_test_channels(*, feerate=6000, local_msat=None, remote_msat=None,
     funding_sat = ((local_msat + remote_msat) // 1000) if local_msat is not None and remote_msat is not None else (bitcoin.COIN * 10)
     local_amount = local_msat if local_msat is not None else (funding_sat * 1000 // 2)
     remote_amount = remote_msat if remote_msat is not None else (funding_sat * 1000 // 2)
+    local_max_inflight = local_amount if local_max_inflight is None else local_max_inflight
+    remote_max_inflight = remote_amount if remote_max_inflight is None else remote_max_inflight
```
ecdsa, line 145: why use local_amount and remote_amount as default? one of these values can get arbitrarily low

accumulator: indeed, should've been `funding_sat`

electrum/lnworker.py
ecdsa: > I can remove that part and let available_to_spend handle this, but then we lose the specific error message.

I created helper functions to avoid code duplication, pushed in your repo

electrum/lnworker.py
```diff
@@ -1925,7 +1925,7 @@ def suggest_splits(
         r_tags,
     ) -> List['SplitConfigRating']:
         channels_with_funds = {
-            (chan.channel_id, chan.node_id): int(chan.available_to_spend(HTLCOwner.LOCAL))
+            (chan.channel_id, chan.node_id): int(chan.available_to_spend(HTLCOwner.LOCAL, single_payment=True))
```
SomberNight, line 1928: I appreciate the intent and this is going in the right direction but note that suggest_splits might decide to add multiple htlcs onto a given channel, not just one, and this test only checks whether a single htlc can be added. That is, if it decides to split the payment amount within a single channel, the check is not sufficient.

electrum/lnworker.py
```diff
@@ -2043,7 +2043,7 @@ async def create_routes_for_payment(
                         self.logger.info(f'per trampoline fees: {per_trampoline_fees}')
                         for chan_id, part_amount_msat in trampoline_parts:
                             chan = self.channels[chan_id]
-                            margin = chan.available_to_spend(LOCAL, strict=True) - part_amount_msat
+                            margin = chan.available_to_spend(LOCAL) - part_amount_msat
```
SomberNight, line 2046: In relation to my other comment, we could specify `single_payment=True` here, and perhaps in this case it might be sufficient. (if we don't split a payment into multiple parts within a given channel)

electrum/lnchannel.py
```diff
@@ -1461,7 +1472,7 @@ def balance_tied_up_in_htlcs_by_direction(self, ctx_owner: HTLCOwner = LOCAL, *,
     def has_unsettled_htlcs(self) -> bool:
         return len(self.hm.htlcs(LOCAL)) + len(self.hm.htlcs(REMOTE)) > 0
 
-    def available_to_spend(self, subject: HTLCOwner, *, strict: bool = True) -> int:
+    def available_to_spend(self, subject: HTLCOwner, *, single_payment: bool = False) -> int:
```
SomberNight, line 1475: > @SomberNight I am not sure what the (unused) `strict` parameter of available_to_spend was supposed to do. I am thinking that the newly introduced `single_payment` parameter should be named `strict`, that would be less confusing.

As explained in a comment in `_assert_can_add_htlc`, the reasoning and usage for the "strict" param was that we can be more strict regarding our own behaviour than what we tolerate from others. In that sense I don't think it's a good idea to rename `single_payment` to `strict`, as it is doing something different.
Though whatever the name, I think the param's behaviour should be briefly explained in the docstring just below.

electrum/lnworker.py
ecdsa: Indeed I  think it should be specified here. This branch is for the trampoline case, where we do not split a payment within a given channel.

electrum/lnworker.py
ecdsa: @accumulator I believe @SomberNight's comment was about fixing #8857, by checking `max_accepted_htlcs`

electrum/lnchannel.py
accumulator: docstring added in 310e4fe05ff11363c836664041de7c48df23b655

accumulator: > I just find the name very obscure. of course we can add a docstring, but that should not be an excuse for using misleading names.

I think `strict` is vague, like 'strict in what sense?'.

> We should also question whether this parameter is useful at all. @accumulator, are there really cases where `single_payment` should not be `True`?

There are a few things the user might want to know;
- what can I pay in a single invoice
- what can I pay over multiple invoices
- what's the balance I actually own

From a technical perspective, we mostly are interested in single invoice as that would avoid payment errors etc.
The other two are with and without reservations

> I had a quick review of the code, and I do not seem to find any. (If there are, they seem to be the exception rather than the rule, and I would suggest that True be the default.)

I'm fine with that.

electrum/lnworker.py
accumulator: Ah yes, you can read that in multiple ways :)

Anyway, that would be for `suggest_splits` to check I think. 

My initial PR had `determine_payment_constraints` for that reason, to have available the remaining # of htlcs for a channel, and the minimum msat amount per htlc, not just max amount.

electrum/lnchannel.py
ecdsa: `num_sats_can_send` is not used in the computation of lightning balance;we use `channel.balance` for that.
I do not think the current codebase currently shows what you can send under the constraint of using multiple invoices. Even if we wanted to do that, passing `single_payment=False` would not achieve that.

What I am asking is, in the current codebase, are there specific instances of `num_sats_can_send` where `single_payment` should be `False`. If there are none, we should not introduce that parameter.

ecdsa: > we could inform the user why (s)he cannot pay an invoice while the available to spend sats might suggest otherwise.

Why would `available_to_spend` suggest otherwise, if we remove the `single_payment` parameter? 
The introduction of that parameter is what could lead to inconsistencies.  I am asking whether we can remove it. 

In this commit you added the parameter to some of the calls, and not to others: [use single_payment in lnworker.num_sats_can_{receive,send} calls where applicable](https://github.com/spesmilo/electrum/pull/9781/commits/0e9c0405ebbab0c0351a8c77744775a72338a008)

I believe `single_payment=True` is applicable everywhere. 
So, again, I am asking if there are specific instances where it should not be applied.

accumulator: https://github.com/spesmilo/electrum/blob/33ea89c94b2ca3a42a6ac3b01be97f9877868545/electrum/gui/qml/qechannellistmodel.py#L108-L109

here it is used for rendering the channel balance/spendable ratios

accumulator: > The user visible fields 'can send' and 'can receive' should really show what we can send and receive, so I think the single_payment=True checks should apply there.

IMO there's no binary good/bad choice here. I'll let you decide what is the proper way.

(As I described above, I see the use of `single_payment=False` in certain situations, e.g. for channels like ACINQ that have a max of 45% of the channel capacity in-flight the information can be useful for users when they want to pay multiple invoices and they want to know the total available funds, even beyond the in-flight limitations of the channels)

> Regarding the blue/yellow ChannelBar, I am not sure whether it is useful to show canSend instead of localCapacity. 

ChannelBar is showing both, with canSend as a fraction of localCapacity. In high-fee situations the difference is significant. It might be confusing without a legend though.

> It seems to me that it would be better to just show the balance. However, if we show CanSend, why should it not include the htlc checks?

I'm fine with either choice, see above.

ecdsa: > and they want to know the total available funds, even beyond the in-flight limitations of the channels)

That is what the channel balance is for. I think it is enough to show them two numbers. Nobody would understand the meaning of a third one. 

Most importantly, I do not want to add a parameter that is not really needed, especially when it impacts so many lines of code.

> I expect a lot of user questions if we display 45% of capacity as "can send" for a large channel

That is because QML does not display the channel balance at all. On desktop Qt we display the channel balance in the channel details, and we show the aggregate can_send/can_receive values over all active channels. On QML, the channel details should at least display the channel balance, and maybe available_to_spend, but that is less useful, because we what really matters is the aggregate value, which is already displayed in the channels list.

ecdsa: of course we could also add some gui explanation as to why you cannot send your whole balance. but I guess this out of scope for this PR.

ecdsa:
ok, thanks for removing that parameter. I had to force push in order to fix two remaining issues:
 - the `strict` parameter to available_to_spend was removed in the first commit (this makes the commit fail flake8)
 - with the single_payment parameter removed, the checks performed by the helper functions in `_assert_can_add_htlc` must be performed before "check proposer can afford htlc", not after, or we will not raise the correct exception

PR #9782, opened 2025-05-02
qt: add qr reading from file to ScanQRTextEdit and SendTab

f321x:
There was no ability to read qr codes contained in image files. This is confusing in some contexts, as `on_file_input()` of ScanQRTextEdit will read the whole content of the file (instead of looking for qr codes). 
The revealer plugin for example generates png files containing qr codes and uses the `ScanQRTextEdit` to get user input, for the user it would seem logical to click on 'Read from file' to load the generated file, however this will result in the wrong data getting loaded. Having the option to explicitly load a QR from file makes this clear. 
Besides that it seems useful, considering reading QR from screenshots doesn't work on wayland.


accumulator:
Nothing wrong with the code per se, but it doesn't find a QR code in my testing (screenshot saved as png from spectacle, which itself detects a QR code)

accumulator:
Screenshot_20250502_160807

accumulator:
> That's weird, works fine for me with the following screenshots:

Yeah, your screenshots work fine here too

f321x:
The zbar library we use here for detecting the qr codes doesn't seem to like dense qr codes that are a large portion of the image. So i added a function `_reduce_qr_code_density()` to create a new image of which the original image is only a smaller part (it pads the image with white space), so it can attempt the scan multiple times with the qr code being different proportions of the image. This seems to make it more reliable, also with your screenshot.

electrum/gui/qt/qrreader/__init__.py
```diff
@@ -59,16 +60,37 @@ def scan_qrcode(
...
+        if res:
+            break
+        # zbar doesn't like qr codes that are too large in relation to the whole image
+        image = _reduce_qr_code_density(image)
     return res
 
+def _reduce_qr_code_density(image: QImage) -> QImage:
+    """ Reduces the size of the qr code relative to the whole image. """
+    new_width = image.width() * 2
+    new_height = image.height() * 2
```
accumulator, line 82: This can really get large. Maybe `+ 200` suffices, as you offset the source image (100,100) below?

accumulator (changes requested):

electrum/gui/qt/qrreader/__init__.py
f321x: yeah right, + 200 didn't work, but instead i changed the function to make the passed image a smaller part of a new (white) image with the same size, this also is faster and seems as reliable. I also removed the offset as it could happen to cut parts of the original image if it is very small.

accumulator: > instead i changed the function to make the passed image a smaller part of a new (white) image with the same size, this is faster too and seems as reliable.

That's.. unexpected :)

PR #9784, opened 2025-05-02
qt: allow lightning addresses in contacts

f321x:
Lightning addresses are used very widely so it should be possible to create contacts with a lightning address.
This would later allow us to import the contact list of a nostr npub and all the lightning addresses of these contacts.

### DIFF
--- a/electrum/gui/qt/send_tab.py
+++ b/electrum/gui/qt/send_tab.py
@@ -803,7 +803,7 @@ class SendTab(QWidget, MessageBoxMixin, Logger):
     def payto_contacts(self, labels):
         paytos = [self.window.get_contact_payto(label) for label in labels]
         self.window.show_send_tab()
-        self.payto_e.do_clear()
+        self.do_clear()
         if len(paytos) == 1:
             self.logger.debug('payto_e setText 1')
             self.payto_e.setText(paytos[0])
--- a/electrum/payment_identifier.py
+++ b/electrum/payment_identifier.py
@@ -289,6 +289,10 @@ class PaymentIdentifier(Logger):
                 self._type = PaymentIdentifierType.EMAILLIKE
                 self.emaillike = contact['address']
                 self.set_state(PaymentIdentifierState.NEED_RESOLVE)
+            elif contact['type'] == 'lnaddress':
+                self._type = PaymentIdentifierType.LNADDR
+                self.lnurl = contact['address']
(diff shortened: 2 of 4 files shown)


electrum/payment_identifier.py
```diff
@@ -648,7 +648,7 @@ def _get_bolt11_fields(self):
         amount = lnaddr.get_amount_sat()
         return pubkey, amount, description
 
-    async def resolve_openalias(self, key: str) -> Optional[dict]:
```
ecdsa, line 651: so, this was potentially blocking the asyncio thread?

electrum/payment_identifier.py
```diff
@@ -648,7 +648,7 @@ def _get_bolt11_fields(self):
...
+    def resolve_openalias(self, key: str) -> Optional[dict]:
         # TODO: below check needed? we already matched RE_EMAIL/RE_DOMAIN
         # if not (('.' in key) and ('<' not in key) and (' ' not in key)):
         #     return None
```
accumulator, line 654: might as well remove the TODO..

electrum/gui/qt/send_tab.py
```diff
@@ -801,7 +801,7 @@ def toggle_paytomany(self):
     def payto_contacts(self, labels):
         paytos = [self.window.get_contact_payto(label) for label in labels]
         self.window.show_send_tab()
-        self.payto_e.do_clear()
+        self.do_clear()
```
ecdsa, line 804: why this change? it will clear amount and description fields

electrum/payment_identifier.py
f321x: changed this back to `async def ` in this pr, to prevent conflicts with the other pr.

electrum/gui/qt/send_tab.py
f321x: Because `payto_e.do_clear()` wasn't clearing all fields, if you select a lnaddress contact it will resolve its description, if you then go back to contacts and select another contact it will not resolve and update the description but keep the one of the previous lnaddress. Using `self.do_clear()` solves this by clearing all fields.

electrum/payment_identifier.py
ecdsa: ok, I will merge this. the asyncio blocking issue will be fixed in amother pr

PR #9792, opened 2025-05-06
qml: fix storing override amount too soon, while user is allowed to set a new override amount which was silently ignored.

accumulator:
- lightning: implicit save of lightning invoice with override amount deferred until actual payment is done. 
- on-chain: for now, invoice is implicitly saved when override amount is passed to `ConfirmTxDialog`. (this is sooner than current desktop client, which saves when `OK` is clicked from `ConfirmTxDialog`, or `Broadcast` is clicked from `Preview`)

### DIFF
--- a/electrum/gui/qml/components/InvoiceDialog.qml
+++ b/electrum/gui/qml/components/InvoiceDialog.qml
@@ -487,10 +487,6 @@ ElDialog {
                         if (amountMax.checked)
                             invoice.amountOverride.isMax = true
                     }
-                    if (!invoice.isSaved) {
-                        // save invoice if newly parsed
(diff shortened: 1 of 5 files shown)


accumulator:
Marking as ready for review:
- it fixes the LN case
- it improves the onchain case, however it should be further refined to work as the desktop client. 
- The desktop client is inconsistent here too, as it only saves the invoice implicitly if `OK` is clicked from `ConfirmTxDialog`, or `Broadcast` is clicked from `Preview`. It doesn't save when `Add to History` is clicked or if `Send to Cosigner` is clicked, where you'd expect it to be saved as well. It should also implicitly save if any of the `Share` options is used IMO.

Issue #9742, opened 2025-04-16
[Security Suggestion] Increase PBKDF2-HMAC Iterations for Stronger Key Derivation


choseogyeong:
### Description

Hi, while reviewing the code I noticed that the PBKDF2-HMAC implementation currently uses:

`return to_hexstr(hashlib.pbkdf2_hmac('sha512', key, b'Digital Bitbox', iterations = 20480))
`

According to [OWASP 2023 recommendations](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2), the recommended minimum for PBKDF2-HMAC-SHA512 is 210,000 iterations.

This should significantly improve security and protect against brute-force attacks.
Would it be possible to update this in future releases?

Thank you!

0xopahc:
+1 lthis 
looking into this .  im a electurm user and think this is good idea.  havnt read the code but am jr dev ...

PR #9799, opened 2025-05-09
looking for feedback updated itterations

bhappy-buidl:
what else in the codebase is using this function ? What else needs to be updated ? 
@spesmilo a point in the right dirrection on  how to further address the issue #9742 would be much apreicated I just read this code for the first time a few min ago


ecdsa:
this would break existing keys and passwords

PR #9820, opened 2025-05-12
qml: add terms of use to setup wizard

f321x:
Adds terms of use to qml setup wizard. Flow is similar to qt. Tested on a Pixel 9, and Pixel 4a in the Emulator.

### DIFF
--- a/electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
+++ b/electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
@@ -1,0 +1,48 @@
+import QtQuick
+import QtQuick.Layouts
+import QtQuick.Controls
+
+import "../controls"
+
+WizardComponent {
+    valid: true
+    last: true
+
+    Flickable {
+        anchors.fill: parent
+        contentHeight: mainLayout.height
+        clip: true
+        interactive: height < contentHeight
+
+        ColumnLayout {
+            id: mainLayout
+            width: parent.width
+            spacing: constants.paddingLarge
+
+            Image {
+                Layout.fillWidth: true
+                fillMode: Image.PreserveAspectFit
+                source: Qt.resolvedUrl('../../../icons/electrum_presplash.png')
+                // reduce spacing a bit
+                Layout.topMargin: -100
+                Layout.bottomMargin: -200
+            }
+
+            Label {
+                Layout.fillWidth: true
+                text: qsTr("Terms of Use")
+                font.pixelSize: constants.fontSizeLarge
+                font.bold: true
+                horizontalAlignment: Text.AlignHCenter
+            }
+
+            Label {
+                Layout.fillWidth: true
+                text: wiz.termsOfUseText
+                wrapMode: Text.WordWrap
+                font.pixelSize: constants.fontSizeMedium
+                padding: constants.paddingSmall
+            }
+        }
+    }
+}
--- a/electrum/gui/qml/components/wizard/WCWelcome.qml
+++ b/electrum/gui/qml/components/wizard/WCWelcome.qml
@@ -6,41 +6,29 @@ import "../controls"
 
 WizardComponent {
     valid: true
-    wizard_title: qsTr('Electrum Bitcoin Wallet')
+    wizard_title: qsTr('Network Configuration')
 
     function apply() {
-        wizard_data['use_defaults'] = !config_advanced.checked
-        wizard_data['want_proxy'] = config_advanced.checked && config_proxy.checked
-        wizard_data['autoconnect'] = !config_server.checked || !config_advanced.checked
+        wizard_data['use_defaults'] = !config_proxy.checked && !config_server.checked
+        wizard_data['want_proxy'] = config_proxy.checked
+        wizard_data['autoconnect'] = !config_server.checked
     }
 
     ColumnLayout {
         width: parent.width
 
-        Image {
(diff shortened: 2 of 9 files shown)


electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
```diff
@@ -0,0 +1,52 @@
...
+
+        Label {
+            Layout.fillWidth: true
+            text: qsTr("Terms of Use")
+            font.pixelSize: 18
+            font.bold: true
+            horizontalAlignment: Text.AlignHCenter
+        }
+
+        ScrollView {
```
accumulator, line 32: Instead of a `ScrollView`, you should use a `Flickable` around the `ColumnLayout` instead. (see e.g. https://github.com/spesmilo/electrum/blob/master/electrum/gui/qml/components/wizard/WCScriptAndDerivation.qml)

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
```diff
@@ -0,0 +1,52 @@
...
+
+            // Prevent horizontal scrolling
+            ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
+            ScrollBar.vertical.policy: ScrollBar.AsNeeded
+
+            Label {
+                width: scrollView.width
+                text: wiz.termsOfUseText
+                wrapMode: Text.WordWrap
+                font.pixelSize: 14
```
accumulator, line 47: same here, use `constants.fontSizeMedium`

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
```diff
@@ -0,0 +1,52 @@
...
+                padding: 8
```
accumulator, line 48: same here, use `constants.paddingSmall`

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
```diff
@@ -0,0 +1,52 @@
...
+            source: Qt.resolvedUrl('../../../icons/electrum_presplash.png')
+            // reduce spacing a bit
+            Layout.topMargin: -100
+            Layout.bottomMargin: -200
+        }
+
+        Label {
+            Layout.fillWidth: true
+            text: qsTr("Terms of Use")
+            font.pixelSize: 18
```
accumulator, line 27: use `constants.fontSizeLarge` (or one of the other font size symbols in `constants`) instead of numeric literal. This allows us to change theming centrally (in `Constants.qml`), and at runtime.

accumulator (changes requested):

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
f321x: Changed, thanks for the hint. 
Would it make sense to set `spacing:  constants.paddingSomething` (or introduce spacing constants) or should i just leave it at 16?
```qml
       ColumnLayout {
            id: mainLayout
            width: parent.width
            spacing: 16
```

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
f321x: changed

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
f321x: changed

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
f321x: Changed in https://github.com/spesmilo/electrum/pull/9820/commits/c32b7d35a229cfada06b70ce5368963c6617147b.

electrum/gui/qml/components/wizard/WCTermsOfUseRequest.qml
accumulator: Yes please replace 16. BTW, spacing and padding use the same `constants.paddingXYZ`, so no need to add spacing* symbols.

f321x: Ok, changed the spacing to PaddingLarge

electrum/gui/qml/components/wizard/WCWelcome.qml
```diff
@@ -54,6 +42,17 @@ WizardComponent {
                 checked: false
                 onCheckedChanged: checkIsLast()
             }
+
+            Layout.bottomMargin: 40
```
accumulator, line 46: PR looks good currently, but a few small nitpicks:

- I usually try to keep `Layout` properties together at the beginning of the component. 
- You could move both margins into the `ColumnLayout` as `Layout.topMargin: 40; Layout.bottomMargin: 40`
- You could also de-literal the value 40 to `2*constants.paddingLarge`

accumulator (approved):

electrum/gui/qml/components/wizard/WCWelcome.qml
f321x: ok changed it like this and used `2*paddingXLarge (20)` as margin. Thanks for the hints.

PR #9826, opened 2025-05-14
fix: reduce update_fee target for anchor channels

f321x:
the update_fee logic for lightning channels is not adapted to anchor channels causing us to send update_fee with a eta target of 2 blocks. This causes force closes when there are mempool spikes as the fees we try to update to are a lot higher than e.g. eclair uses. Eclair will send errors if our attempted fee update is 10x > than their fee.
This now uses the 1008 block estimate + 25% buffer as the anchor channel fee target. This is the lowest value available to fetch from electrum servers / bitcoin core estimatesmartfee limit, which also (guarantees)[https://github.com/bitcoin/bitcoin/blob/33dfbbdff69dee2f1a27f69f5f7a123d0d47ef42/src/rpc/fees.cpp#L85] acceptance into the mempool.

Potentially an addition to the electrum protocol allowing to explicitly fetch the minimal mempool acceptance fee would be useful for the future.

Example of such a force close:
```
20250512T153015.040871Z |     INFO | network | fee_estimates {25: 1000, 10: 1198, 5: 1624, 2: 25239}
20250512T153015.042300Z |     INFO | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | (889664x860x1) oncha  
in fees have changed considerably. updating fee.  
20250512T153015.042453Z |     INFO | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | (chan: 889664x860x1)  
current pending feerate 353. new feerate 6309  
20250512T153015.042765Z |    DEBUG | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | Sending UPDATE_FEE  
20250512T153015.043246Z |     INFO | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | send_commitment. cha  
n 889664x860x1. ctn: 608.  
20250512T153015.043339Z |     INFO | lnchannel.Channel.[889664x860x1] | sign_next_commitment. ctn=608  
20250512T153015.046253Z |     INFO | transaction | adding signature for 02cbc2ce4bb807bed254258422d2318  
76862e92280d5aa1c1ec71bc11a51ab702c. spending utxo 954703ec4d6044a4d38e98b44116d062c64b8c2798b5be3c186f  
d73df4fe146d:1  
20250512T153015.046905Z |    DEBUG | transaction | tx.sign() finished. is_complete=False  
20250512T153015.047170Z |    DEBUG | lnchannel.Channel.[889664x860x1] | sign_next_commitment. pending_r  
emote_commitment.serialize()='cHNidP8BAN8CAAAAAW0U/vQ9128YPL61mCeMS8Zi0BZBtJiO06REYE3sA0eVAQAAAADe5EOAB  
EoBAAAAAAAAIgAgBNYj6SGWhmgpLuy5t+wOZydMd5w9AXtiFi5/iphcFPZKAQAAAAAAACIAIAl4sd/cEJ0F9Ur3FDu2cCqPwXWVjoNU  
83mwx198Q8EXMzwBAAAAAAAiACAjfV4QIrbAk3IE9shrBvDwAlsIxMBcN5XI8fBpf2FXTBWrDgAAAAAAIgAgm4c18e0wfwPFI+1yRCH  
dpg2u13HOZ+ZULvH2jtMZ75HHn0MgACICAsvCzku4B77SVCWEItIxh2hi6SKA1aocHscbwRpRq3AsRzBEAiBHtguWPqa6Szgo/zlpNV  
YHt5+gT8xX4AoVhqOYqSyxrwIgBIRmrm7As0uI+FgcEfPwJ5UU6D7zylhUxmFrvZTyRTABAQVHUiECy8LOS7gHvtJUJYQi0jGHaGLpI  
oDVqhwexxvBGlGrcCwhA1j1uhAUjVhQUHpjY8/CnYWFYqui22k0nNevgSMfFtcTUq4AAAAAAA=='. sig_64.hex()='47b60b963ea  
6ba4b3828ff3969355607b79fa04fcc57e00a1586a398a92cb1af048466ae6ec0b34b88f8581c11f3f0279514e83ef3ca5854c6  
616bbd94f24530'  
20250512T153015.047759Z |    DEBUG | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | Sending COMMITMENT_S  
IGNED  
20250512T153015.048056Z |     INFO | wallet_db.WalletDB | appending 83 pending changes  
20250512T153015.049813Z |    DEBUG | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | Received ERROR  
20250512T153015.050042Z |     INFO | lnpeer.Peer.[LNWallet, 03ecef675b-1efa36c8] | remote peer sent err  
or [DO NOT TRUST THIS MESSAGE]: 'local/remote feerates are too different: remoteFeeratePerKw=6309 local  
FeeratePerKw=562'. chan_id=6d14fef43dd76f183cbeb598278c4bc662d01641b4988ed3a444604dec034794. is_known_c  
han_id=True  
20250512T153015.053411Z |     INFO | transaction | adding signature for 02cbc2ce4bb807bed254258422d2318  
76862e92280d5aa1c1ec71bc11a51ab702c. spending utxo 954703ec4d6044a4d38e98b44116d062c64b8c2798b5be3c186f  
d73df4fe146d:1  
20250512T153015.053948Z |    DEBUG | transaction | tx.sign() finished. is_complete=False  
20250512T153015.054171Z |    DEBUG | lnchannel.Channel.[889664x860x1] | Setting channel state: OPEN ->  
FORCE_CLOSING  
```


electrum/lnpeer.py
```diff
@@ -2605,51 +2608,64 @@ def maybe_update_fee(self, chan: Channel):
...
-            if feerate_per_kw == chan_fee:
-                feerate_per_kw += 1
+            if current_feerate_per_kw == chan_fee:
+                update_feerate_per_kw += 1
         else:
             return
         self.logger.info(f"({chan.get_id_for_log()}) current pending feerate {chan_fee}. "
-                         f"new feerate {feerate_per_kw}")
-        chan.update_fee(feerate_per_kw, True)
+                         f"new feerate {current_feerate_per_kw}")
```
SomberNight, line 2662: I find the `current_feerate_per_kw` / `update_feerate_per_kw` split somewhat confusing.
For example, here, we should probably log the value we are actually updating the fee to.

electrum/lnpeer.py
```diff
@@ -2605,51 +2608,64 @@ def maybe_update_fee(self, chan: Channel):
...
-                ratio = chan_feerate / feerate_per_kw
+                ratio = chan_feerate / current_feerate_per_kw
                 if ratio < 0.5:
                     # Note that we trust the Electrum server about fee rates
                     # Thus, automated force-closing might not be a good idea
                     # Maybe we should display something in the GUI instead
                     self.logger.warning(
                         f"({chan.get_id_for_log()}) feerate is {chan_feerate} sat/kw, "
-                        f"current recommended feerate is {feerate_per_kw} sat/kw, consider force closing!")
+                        f"current recommended feerate is {current_feerate_per_kw} sat/kw, consider force closing!")
```
SomberNight, line 2648: here also, we probably want `update_feerate_per_kw` instead.
(what we "recommend" should be what we would want to update the fee to, no? :P)

electrum/lnpeer.py
```diff
@@ -2605,51 +2608,64 @@ def maybe_update_fee(self, chan: Channel):
...
-        feerate_per_kw: Optional[int] = self.lnworker.current_target_feerate_per_kw()
-        if feerate_per_kw is None:
+        current_feerate_per_kw: Optional[int] = self.lnworker.current_target_feerate_per_kw(
+            has_anchors=chan.has_anchors()
+        )
+        if current_feerate_per_kw is None:
             return
+        # add some buffer to anchor chan fees as we always act at the lower end and don't
+        # want to get kicked out of the mempool immediately if it grows
+        fee_buffer = current_feerate_per_kw * 0.25 if chan.has_anchors() else 0
```
SomberNight, line 2618: note: the +25% comes from eclair: https://github.com/ACINQ/eclair/blob/f14b92d7df8727d6362d86251d717a93f37df251/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/OnChainFeeConf.scala#L127
but they have additional logic giving it more buffer

I think we could just use +50%, to be safer, but not have said additional logic.

SomberNight: on second thought, let's just add a 5 sat/vbyte floor, to reduce the chance of not getting into the mempool after coming back from a long offline period

electrum/lnpeer.py
f321x: Yeah missed that log, changed it.
I found it useful to have these two variables as it is hard to do the calculations otherwise, would you suggest different naming or do you see any cleaner approach?

electrum/lnpeer.py
f321x: have added a floor in `current_target_feerate_per_kw()` and increased the buffer to 50%

electrum/lnpeer.py
f321x: Switched to `update_feerate_per_kw`, i thought it may is better to be "less aggressive" here but it's just a warning anyways, so just using ` update_feerate_per_kw` for the ratio and the log seems more consistent yes.

electrum/lnworker.py
```diff
@@ -3034,16 +3035,22 @@ async def reestablish_peers_and_channels(self):
...
-    def current_target_feerate_per_kw(self) -> Optional[int]:
+    def current_target_feerate_per_kw(self, *, has_anchors: bool) -> Optional[int]:
         if self.network.fee_estimates.has_data():
-            feerate_per_kvbyte = self.network.fee_estimates.eta_target_to_fee(FEE_LN_ETA_TARGET)
+            target: int = FEE_LN_MINIMUM_ETA_TARGET if has_anchors else FEE_LN_ETA_TARGET
+            feerate_per_kvbyte = self.network.fee_estimates.eta_target_to_fee(target)
+            if has_anchors:
+                # set a floor of 5 sat/vb to have some safety margin in case the mempool
+                # grows quickly
+                feerate_per_kvbyte = max(feerate_per_kvbyte, 5000)
```
SomberNight, line 3045: self-note: we add the +50% buffer on top of this later, so the effective floor is 7.5 s/b

Issue #9829, opened 2025-05-15
bug: qml: function called on wrong thread when saving psbt to wallet


f321x:
### Description

The following exception occurred to me when saving a PSBT received as cosigner to the wallet history:
```
 107.60 | I | plugins.psbt_nostr.qml.QmlCosignerWallet.[psbt_multisig] | new event e3629f975e776d539fa194fed958fc302064ea0450a029341f5271cb321064ed
107.60 | I | plugins.psbt_nostr.qml.QmlCosignerWallet.[psbt_multisig] | received PSBT from 8970e204fb5f2d7461104d38167efe3620b1c44849107cea80e6139e76ad2e34
[...]

PR #9832, opened 2025-05-16
qml: handle callbacks via ui thread.

accumulator:


f321x:
Tested it, fixes the exception in https://github.com/spesmilo/electrum/issues/9829 for me.

ecdsa:
This will indeed fix #9829 by ensuring `update()` runs in the qt thread

However, why are `add_info_from_wallet` and `add_info_from_network` called from there?
I think `add_info_from_wallet` only needs to be called once, not on every update. 
In qt/transaction_dialog.py these methods are called from the constructor.

PR #9833, opened 2025-05-16
dns: use async dnspython interface 

f321x:
dnspython has [added](https://github.com/rthalley/dnspython/commit/7f197681162b420c8b3fd5b5339628378cf80d62) an async interface in version 2.0 which we already require in our requirements. 
The api is the same as the sync version and is stable between 2.0 and the current 2.7. 
So i switched the dnspython calls in dnssec (used for openalias resolving) and the lightning dns seed fetching to use the async interface instead of threads.

### DIFF
--- a/electrum/contacts.py
+++ b/electrum/contacts.py
@@ -22,16 +22,15 @@
 # SOFTWARE.
 import re
 from typing import Optional, Tuple, Dict, Any, TYPE_CHECKING
-
+import asyncio
 import dns
-import threading
 from dns.exception import DNSException
 
 from . import bitcoin
 from . import dnssec
 from .util import read_json_file, write_json_file, to_string, is_valid_email
 from .logging import Logger, get_logger
-from .util import trigger_callback
+from .util import trigger_callback, get_asyncio_loop
 
 if TYPE_CHECKING:
     from .wallet_db import WalletDB
@@ -85,7 +84,7 @@ class Contacts(dict, Logger):
             return res
         return None
 
-    def resolve(self, k):
+    async def resolve(self, k) -> dict:
         if bitcoin.is_address(k):
             return {
                 'address': k,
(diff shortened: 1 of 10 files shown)


electrum/contacts.py
```diff
@@ -159,6 +157,8 @@ def _resolve_openalias(cls, url: str) -> Optional[Tuple[str, str, bool]]:
                 if not address:
                     continue
                 return address, name, validated
+            return None
+        return None
```
SomberNight, line 161: Note: the return inside the for loop seems to change the behaviour.
(though this change is already on master coming from a different merged PR?)

f321x: Fixed in https://github.com/spesmilo/electrum/commit/b3fa2a9d9cd40276994b02dbedfa65dc63fa8057

electrum/lnworker.py
```diff
@@ -427,9 +427,9 @@ async def _get_next_peers_to_try(self) -> Sequence[LNPeerAddr]:
             return [random.choice(fallback_list)]
 
         # last resort: try dns seeds (BOLT-10)
-        return await run_in_thread(self._get_peers_from_dns_seeds)
+        return await self._get_peers_from_dns_seeds()
 
-    def _get_peers_from_dns_seeds(self) -> Sequence[LNPeerAddr]:
+    async def _get_peers_from_dns_seeds(self) -> Sequence[LNPeerAddr]:
         # NOTE: potentially long blocking call, do not run directly on asyncio event loop.
```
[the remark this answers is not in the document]
f321x: yes, removed

electrum/dns_hacks.py
```diff
@@ -69,8 +62,15 @@ def resolve_with_dnspython(host):
...
         expected_errors = (dns.resolver.NXDOMAIN, dns.resolver.NoAnswer,
                            concurrent.futures.CancelledError, concurrent.futures.TimeoutError)
-        ipv6_fut = _dns_threads_executor.submit(dns.resolver.resolve, host, dns.rdatatype.AAAA)
-        ipv4_fut = _dns_threads_executor.submit(dns.resolver.resolve, host, dns.rdatatype.A)
+        loop = get_asyncio_loop()
+        ipv6_fut = asyncio.run_coroutine_threadsafe(
+            dns.asyncresolver.resolve(host, dns.rdatatype.AAAA),
+            loop,
+        )
+        ipv4_fut = asyncio.run_coroutine_threadsafe(
```
SomberNight, line 70: Note: calling `.result()` on these futures will deadlock if we are running on the asyncio thread - which was not the case before.
Though socket.getaddrinfo is unlikely to run on the asyncio thread I think.

SomberNight: I added an assert in https://github.com/spesmilo/electrum/commit/e75476430ca50a3c572b7cb55fcf8a83e668e98a

PR #9838, opened 2025-05-18
CLI: add preimage to invoices

ecdsa:
This should make regtest extract_preimage more reliable


ecdsa:
note: alternatively, we could add the preimage in the fields returned by `list_invvoices`

PR #9842, opened 2025-05-19
txbatcher: sleep on NotEnoughFunds

ecdsa:
In some edge cases, this favors waiting instead of creating a child transaction (see unit test)


ecdsa:
@SomberNight I pushed this PR in order to show that txbatcher could use a different behavior:
 - Currently, we create a new batch if there are not enough funds. this creates a child transaction.
 - In this PR, we wait until funds are available, which favors batching
 
I am not sure what the best choice is. This is a design decision.

PR #9845, opened 2025-05-20
fix: suggest_splits exclude single channel splits

f321x:
passing `exclude_single_channel_splits` to `suggest_splits` did not reliably remove single channel splits from configurations if the same configuration also contained a valid split on another channel

This [comment](https://github.com/spesmilo/electrum/blob/f908a34b0b100a9b81ed9a5c76e128eb73776bdc/electrum/lnworker.py#L1942C47-L1942C51) explains `exclude_single_channel_splits` as:

> we don't split within a channel when sending to a trampoline node, 
the trampoline node will split for us

### DIFF
--- a/electrum/mpp_split.py
+++ b/electrum/mpp_split.py
@@ -82,14 +82,10 @@ def remove_single_part_configs(configs: List[SplitConfig]) -> List[SplitConfig]:
 
 
 def remove_single_channel_splits(configs: List[SplitConfig]) -> List[SplitConfig]:
-    filtered = []
-    for config in configs:
-        for v in config.values():
-            if len(v) > 1:
-                continue
-            filtered.append(config)
-    return filtered
-
+    return [
+        config for config in configs
+        if all(len(channel_splits) <= 1 for channel_splits in config.values())
+    ]
 
 def rate_config(
         config: SplitConfig,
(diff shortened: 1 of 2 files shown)


electrum/mpp_split.py
```diff
@@ -84,9 +84,11 @@ def remove_single_part_configs(configs: List[SplitConfig]) -> List[SplitConfig]:
...
-        for v in config.values():
-            if len(v) > 1:
-                continue
+        for channel_splits in config.values():
+            if len(channel_splits) > 1:
+                # e.g. [225902904, 464361733]
+                break
+        else:
             filtered.append(config)
     return filtered
```
SomberNight, line 93: Nice catch!
how about this?
```py
    return [
        config for config in configs
        if all(len(channel_splits) == 1 for channel_splits in config.values())
    ]
```
(can `channel_splits` have `len 0`?)

f321x: Yes `channel_splits` can have `len 0` as `suggest_splits` adds channels to the config but don't necessarily also adds amounts to it:
```python
                for c in channel_keys:
                    if c not in config:
          --------->    config[c] = []          
                    channel_funds, channel_slots = channels_with_funds[c]
                    if sum(config[c]) + amount <= channel_funds and len(config[c]) < channel_slots:
                        config[c].append(amount)
                        break
```
 
Changed it to your version, but i check for <= 1 instead of == 1 so it doesn't sort out 'valid' configs according the goal of this function.

SomberNight (approved):
Looks good!

PR #9847, opened 2025-05-20
wip: implement release channels

f321x:
**work in progress**
Implements release channels so the user can select either a stable or beta release channel in the SettingsDialog for update notifications. 
Adapts the release_www.sh script to handle alpha and beta versions which are detected by an 'a' or 'b' in the version name. 
The version fields are added into the `extradata` field of the version file as 'version_alpha' and 'version_beta' and the `extradata` field is being signed too in the `extradata_signature` field.
The update checker is supposed to return the newest version of those contained in the version file which is
allowed according to the user configured option.


contrib/make_download
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/python3
-import re
 import os
 import sys
-import importlib
+import json
+import importlib.util
 from collections import defaultdict
+from packaging.version import Version
```
f321x, line 7: I use packaging here which is already in our requirements and provides a `Version` class similar to the `StrictVersion` used in `update_checker.py` which cannot be properly imported here (?). If we don't want to use it in this script we could also manually parse the version.

contrib/release_www.sh
```diff
@@ -34,29 +34,71 @@ if [ -z "$ELECTRUM_SIGNING_WALLET" ] || [ -z "$ELECTRUM_SIGNING_ADDRESS" ]; then
...
+
+# serialize extradata msg using the same method as in update_checker.py to prevent differences
+extradata=$(python3 -c "import json, sys; print(json.dumps({
+    'android_versioncode_nullarch': int(sys.argv[1]),
+    'version_alpha': sys.argv[2],
+    'version_beta': sys.argv[3]
+}, sort_keys=True, separators=(',', ':')))" "$ANDROID_VERSIONCODE_NULLARCH" "$VERSION_ALPHA" "$VERSION_BETA")
+
+# signing the hash of extradata instead of passing extradata directly,
+# so it doesn't get parsed as dict when passing it into signmessage,
```
SomberNight, line 90: I started looking into this but it is turning into a quite rabbit hole...
The root cause is here:
https://github.com/spesmilo/electrum/blob/9b243169151ed18a06c3c02bebc508207d4b0456/run_electrum#L247-L249

I would say it's a bug that e.g. `$ ./run_electrum -v --testnet -w ~/.electrum/testnet/wallets/9dk -o signmessage tb1qq2tmmcngng78nllq2pvrkchcdukemtj5s6l0zu "5"` does not work, due to the message getting json-decoded to an int.
Anyway, I am still looking into this.

ecdsa: maybe these 3 lines can be deleted? if something needs to be converted to json, it will be done in parser.parse_args. Example: The following command works after deleting these 3 lines


```diff
@@ -256,9 +256,6 @@ async def run_offline_command(config: 'SimpleConfig', config_options: dict, wall
         print_msg("Warning: running command offline")
     # arguments passed to function
     args = [config.get(x) for x in cmd.params]
-    # decode json arguments
-    if cmdname not in ('setconfig',):
-        args = list(map(json_decode, args))
     # options
     kwargs = {}
     for x in cmd.options:
```

```
/run_electrum --testnet -o serialize '{                                                                                                                                                     
            "inputs": [                                                                                                                                                                        
                {"prevout_hash": "9d221a69ca3997cbeaf5624d723e7dc5f829b1023078c177d37bdae95f37c539", "prevout_n": 1,                                                                           
                 "value_sats": 1000000, "privkey": "p2wpkh:cVDXzzQg6RoCTfiKpe8MBvmm5d5cJc6JLuFApsFDKwWa6F5TVHpD"}                                                                              
            ],                                                                                                                                                                                 
            "outputs": [                                                                                                                                                                       
                {"address": "tb1q4s8z6g5jqzllkgt8a4har94wl8tg0k9m8kv5zd", "value_sats": 990000}                                                                                                
            ]                                                                                                                                                                                  
        }'

```

ecdsa: ah, I forgot that this will not work with curl/jsonrpc

SomberNight: yes, it's not so simple. (1) jsonrpc vs (2) online CLI vs (3) offline CLI all work very subtly different. We need explicit unit tests, as otherwise the whole thing is extremely fragile.
I have some WIP changes from when I last commented above:
https://github.com/SomberNight/electrum/commit/ef57336b8634559c52c8466632856c6648c13f7b

PR #9848, opened 2025-05-20
lnchannel: refactor can_send_ctx_updates

SomberNight:
This adds a lot more sanity checks to lnpeer/lnchannel for chan update operations, and more logging in case they are violated.


electrum/lnpeer.py
```diff
@@ -1720,6 +1725,11 @@ def on_channel_ready(self, chan: Channel, payload):
         self.logger.info(f"on_channel_ready. channel: {chan.channel_id.hex()}")
         if chan.peer_state != PeerState.GOOD:  # should never happen
             raise Exception(f"received channel_ready in unexpected {chan.peer_state=!r}")
```
ecdsa, line 1727: This test that chan.peer_state is GOOD has been removed from various places (see below). Why keep it in `on_channel_ready`?

ecdsa: ah, I see, you are not calling `can_update_ctx` in that specific case, I guess that's why

ecdsa:
maybe there could be a decorator for all the methods that call `can_update_ctx`, log an error and early return. that would save a few lines of code.

PR #9852, opened 2025-05-21
qml: fix assumption history item is onchain when history item has no timestamp

accumulator:


electrum/gui/qml/qetransactionlistmodel.py
```diff
@@ -149,12 +149,13 @@ def tx_to_model(self, tx_item):
...
-            item['section'] = 'local' if tx_mined_info.is_local_like() else 'mempool'
-            status, status_str = self.wallet.get_tx_status(txid, tx_mined_info=tx_mined_info)
-            item['date'] = status_str
+            if not item['lightning']:
+                txid = item['txid']
+                assert txid
+                tx_mined_info = self._tx_mined_info_from_tx_item(tx_item)
+                item['section'] = 'local' if tx_mined_info.is_local_like() else 'mempool'
+                status, status_str = self.wallet.get_tx_status(txid, tx_mined_info=tx_mined_info)
+                item['date'] = status_str
```
ecdsa, line 158: Why is this a PR?
The change looks trivial, but note that 'section' and 'date' are no longer guaranteed to exist after this commit.

PR #9857, opened 2025-05-22
fee_policy: add fallback_feepolicy to FeePolicy

accumulator:
fee_policy: add fallback_feepolicy to FeePolicy constructor and use the fallback
when estimate_fee() cannot determine the (dynamic) fee.

This generalizes static fee estimation fallback to a feepolicy estimator chain.

### DIFF
--- a/electrum/fee_policy.py
+++ b/electrum/fee_policy.py
@@ -80,7 +80,11 @@ class FeePolicy(Logger):
 
     def __init__(self, descriptor: str, *, fallback_feepolicy: 'FeePolicy' = None):
         Logger.__init__(self)
-        self._fallback_feepolicy = fallback_feepolicy
+        if len(chain := descriptor.split(',', 1)) > 1:
+            self._fallback_feepolicy = FeePolicy(chain[1], fallback_feepolicy=fallback_feepolicy)
+            descriptor = chain[0]
+        else:
+            self._fallback_feepolicy = fallback_feepolicy
         try:
             name, value = descriptor.split(':')
             self.method = FeeMethod[name.upper()]
@@ -94,7 +98,8 @@ class FeePolicy(Logger):
         return self.get_descriptor()
 
     def get_descriptor(self) -> str:
-        return self.method.name.lower() + ':' + str(self.value)
+        return self.method.name.lower() + ':' + str(self.value) + \
+            (f',{self._fallback_feepolicy.get_descriptor()}' if self._fallback_feepolicy else '')
 
     def set_method(self, method: FeeMethod):
         assert isinstance(method, FeeMethod)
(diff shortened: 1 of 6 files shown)


accumulator:
As discussed on IRC, to add a mempool fee estimator when ETA fee estimator fails, and a static feerate if also the mempool estimator fails, would be implemented as follows:
`
policy = FeePolicy('eta:2', fallback_policy=FeePolicy('mempool:1000000', fallback_policy=FixedFeeRatePolicy(FEERATE_FALLBACK_STATIC_FEE)))
`

electrum/fee_policy.py
```diff
@@ -78,8 +78,13 @@ def slider_index_of_method(cls, method):
...
     # object associated to a fee slider
 
-    def __init__(self, descriptor: str):
+    def __init__(self, descriptor: str, *, fallback_feepolicy: 'FeePolicy' = None):
         Logger.__init__(self)
+        if len(chain := descriptor.split(',', 1)) > 1:
+            self._fallback_feepolicy = FeePolicy(chain[1], fallback_feepolicy=fallback_feepolicy)
+            descriptor = chain[0]
+        else:
+            self._fallback_feepolicy = fallback_feepolicy
```
ecdsa, line 87: The idea of a policy chain is nice, but this API is dangerous. If the caller reads a policy descriptor from disk, calls this constructor with a `fallback_feepolicy` parameter, and then saves it to disk it again, this will make the policy chain grow indefinitely. 

Note that `TxBatcher` does store fee policies on disk. Fortunately, this uncontrolled growth did not happen here, because this PR does not add a fallback policy to `TxBatcher`. The `fallback_feepolicy` parameter is added everywhere except in `TxBatcher`, thus leaving the original issue unsolved.

If the descriptor can be a chain, there is no point having a `fallback_feepolicy` parameter. It is sufficient to pass the fallback in the descriptor. In general, try to avoid APIs that offer two redundant ways of doing the same thing.

accumulator:
> It would be wise not to use a comma separated list in the descriptor, but something specific to fallbacks. We will probably want to create other types complex policies, and to use descriptors that look like function calls. For example, the maximum of two policies would be described as `max(descriptor1,descriptor2)` It would be better to define a similar functional operator for fallbacks, such as `or(policy1,policy2)`.

Hmm, I was planning to keep it simple. We can add this later if there turns out to be a use for it?

electrum/fee_policy.py
ecdsa: > I did add a fallback in TxBatcher where allow_fallback_to_static_rates was added before.

Ok, I am getting a bit lost here. what is the goal of this PR?
 - I was under the impression you wanted to fix the lack of static fallback in `run_iteration`. 
 - The method that your PR is touching is `is_dust`. That is not going to fix the issue.

PR #9859, opened 2025-05-22
Qt: let user enable/disable keystores with seed or hw wallet

ecdsa:


ecdsa:
> Not really happy with inheriting from `NewWalletWizard` here.
> 
> This is somewhat fragile, as flow might change and `wizard_data` assumptions are made, e.g. `WCScriptAndDerivation` limits script choice options for multisig wallets, that are not checked in this new wizard.

I agree. they should inherit from a common ancestor.

PR #9862, opened 2025-05-24
Fix Timelock Recovery plugin for watch wallets

oren-z0:
In watch-wallets (i.e. air-gapped solutions), the transactions can be signed by clicking the View button,
exporting the PSBT, loading and signing it on a different device, and then merging the signature.

Added assertions that all the transactions are completed before moving to the final window (for downloading PDFs and JSONs).


accumulator:
Is there a reason for the extra check on `parse_output`? This looks to be already covered by the check right above it.

https://github.com/spesmilo/electrum/blob/e30392cd161de8dfcf5981c4d528f82efd4645c3/electrum/plugins/timelock_recovery/qt.py#L408-L412

oren-z0:
@accumulator You're right, it's redundant.
I think I wrote it before I wrote the `or pi.type != PaymentIdentifierType.SPK or not pi.spk_is_address` part.
I'll replace it with a simple `assert` just in case.

accumulator:
I cherry-picked a slightly changed version of your last commit to master, in 2c7afacbe1913241e6cc319131c6913ada06a06d

oren-z0:
@accumulator Great, I've removed my commit and rebased the branch.

electrum/plugins/timelock_recovery/qt.py
```diff
@@ -231,20 +234,30 @@ def update_transactions():
...
+                alert_tx_complete_label.setText(_("✓ Signed" if context.alert_tx.is_complete() else ""))
                 alert_tx_fee_label.setText(_("Fee: {}").format(self.config.format_amount_and_units(context.alert_tx.get_fee())))
                 context.recovery_tx = context.make_unsigned_recovery_tx(fee_policy)
                 assert all(tx_input.is_segwit() for tx_input in context.recovery_tx.inputs())
+                recovery_tx_complete_label.setText(_("✓ Signed" if context.recovery_tx.is_complete() else ""))
                 recovery_tx_fee_label.setText(_("Fee: {}").format(self.config.format_amount_and_units(context.recovery_tx.get_fee())))
                 if create_cancel_cb.isChecked():
                     context.cancellation_tx = context.make_unsigned_cancellation_tx(fee_policy)
                     assert all(tx_input.is_segwit() for tx_input in context.cancellation_tx.inputs())
+                    cancellation_tx_complete_label.setText(_("✓ Signed" if context.cancellation_tx.is_complete() else ""))
```
f321x: Can you separate the logic from the translation in these labels?
e.g. `cancellation_tx_complete_label.setText(_("✓ Signed") if context.alert_tx.is_complete() else "")`
Its more readable and prevents strings from not getting registered for translation.

f321x:
Tested, works fine for me with a regular singlesig and a watchonly wallet. One comment on the code (see above).

Two things i noticed could may be done to improve UX a bit (unrelated to this PR), adding a MIN/MAX button to the `Cancellation time-window (days)` field as the help text is a bit hidden, this would make it more intuitive to figure out the limits instead of brute forcing the input field until it turns red.
And wrapping the PDF saving call into a `WaitingDialog` as it blocks the GUI for a couple of seconds which would feel better if it gave some feedback (especially on slow computers).

electrum/plugins/timelock_recovery/qt.py
oren-z0: ✓ Done 😄

oren-z0:
@f321x Thanks. Fixed and rebased from master branch.

SomberNight:
Thx.

PR #9867, opened 2025-05-26
txbatcher: pass fee policy names instead of fee descriptors

ecdsa:
- this allows to change the fee policy dynamically
 - the fee policy that applies is set in config
 - the fee policy name is stored in txbatch storage


ecdsa:
> AttributeError: 'str' object has no attribute 'get'

ah, thanks.

Issue #9528, opened 2025-02-04
Option to server fallback to next bookmarked server.


nostitos:
Everytime the server we use goes down for just 1h it causes a disproportionate amount of confusion for the users.

This is a situation where the end user is not the same as the person who's chosen the server,

This invariably lead to them switching to Automatic selection and compromise any privacy gain that was being achieved by using trusted servers.

SomberNight:
Note: conceptually there are two different options here:
- only restrict history server to bookmarked set, other connected nodes can be any
- restrict all connected nodes to bookmarked set

nostitos:
https://github.com/spesmilo/electrum/issues/10374

PR #9873, opened 2025-05-27
gui: oneserver option: make text more verbose

SomberNight:
some cosmetic followups to https://github.com/spesmilo/electrum/pull/9243

I think the checkbox text needs to be more descriptive.
I also added a longer explanation for the option.  However TBH to properly explain the tradeoffs it would need an even longer explanation, which would also touch on auto-connect. 

-----

In general I think the auto-connect and oneserver options are inherently intertwined (already showcased by the fact that the `auto_connect and oneserver` combination is disallowed).  Maybe we could have a slider instead of two checkboxes. The slider of "trust" :D
```
(A1. auto_connect) --> (A2. no auto_connect) --> (A3. oneserver)
```

Also consider https://github.com/spesmilo/electrum/issues/9528, which also kind of fits into the slider.  Then we could have:
```
(B1. auto_connect to random) --> (B2. history_server must be bookmarked, other_nodes can be any) --> (B3. all nodes must be bookmarked before we connect) --> (B4. oneserver)
```
yep... it gets kind of complicated. and not easy to explain :(
My thinking with the slider approach is that if we implemented https://github.com/spesmilo/electrum/issues/9528, `A2` would turn into `B2`, and `B3` would be an entirely new state.
So that option, instead of adding another checkbox and increasing the state space from 2x2 to 2x2x2, would increase it from 3 to 4.

### DIFF
--- a/electrum/gui/qml/qeconfig.py
+++ b/electrum/gui/qml/qeconfig.py
@@ -36,7 +36,10 @@ class QEConfig(AuthMixin, QObject):
     @pyqtSlot(str, result=str)
     def longDescFor(self, key) -> str:
         cv = getattr(self.config.cv, key)
-        return cv.get_long_desc() if cv else ''
+        if not cv:
+            return ""
+        desc = cv.get_long_desc()
+        return messages.to_rtf(desc)
 
     @pyqtSlot(str, result=str)
     def getTranslatedMessage(self, key) -> str:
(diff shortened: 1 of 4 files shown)


electrum/gui/qml/qeconfig.py
```diff
@@ -36,7 +36,10 @@ def shortDescFor(self, key) -> str:
     @pyqtSlot(str, result=str)
     def longDescFor(self, key) -> str:
         cv = getattr(self.config.cv, key)
-        return cv.get_long_desc() if cv else ''
+        if not cv:
+            return ""
+        desc = cv.get_long_desc()
+        return messages.to_rtf(desc)
```
SomberNight, line 42: note: I think it is harmless to always convert to rich text here like this.

PR #9885, opened 2025-06-01
Replace config GUI_LAST_WALLET with CURRENT_WALLET

ecdsa:
- CURRENT_WALLET is set when a single wallet is loaded in memory.
 - if several wallets are loaded, RPCs must specify which wallet they apply to, using the --wallet option.
 - the fallback to 'default_wallet' essentially only applies when creating a new wallet file

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -342,7 +342,7 @@ class CommandsServer(AuthenticatedServer):
                 if config_options.get(SimpleConfig.NETWORK_OFFLINE.key()) and not self.config.NETWORK_OFFLINE:
                     raise UserFacingException(
                         "error: current GUI is running online, so it cannot open a new wallet offline.")
-                path = config_options.get('wallet_path') or self.config.get_wallet_path(use_gui_last_wallet=True)
+                path = config_options.get('wallet_path') or self.config.get_wallet_path()
                 self.daemon.gui_object.new_window(path, config_options.get('url'))
                 return True
             else:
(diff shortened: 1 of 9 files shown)


ecdsa:
This replaces #8903 and #7296
An explicit wallet path when several wallets are loaded in memory is also what `bitcoind` requires.

electrum/daemon.py
```diff
@@ -540,6 +546,7 @@ async def _stop_wallet(self, path: str) -> bool:
         wallet = self._wallets.pop(wallet_key, None)
         if not wallet:
             return False
+        self.update_current_wallet()
```
SomberNight, line 549: One consequence of this flow is that if the GUI program segfaults or gets SIGKILLED at a time when there are multiple wallets open, config.CURRENT_WALLET is left as None and on next launch we try to open `default_wallet`.

Previously in such a case the GUI opened to a reasonable "last" wallet.

Note that in case of a SIGINT, as the program can do a  graceful shutdown, the highlighted line gets a chance to run for each open wallet, and hence config.CURRENT_WALLET will get set to the last one being stopped - which is reasonable, though it might differ from the previous behaviour of "last opened".

SomberNight: > One consequence of this flow is that if the GUI program segfaults or gets SIGKILLED at a time when there are multiple wallets open, config.CURRENT_WALLET is left as None and on next launch we try to open default_wallet.

This was fixed in https://github.com/spesmilo/electrum/commit/972ebb04203534b34ad1121358b25f767893cce2

PR #9888, opened 2025-06-02
Qml fixes

### LINKED ISSUE #9882: QML GUI minor issues
### Description

I have been reviewing the QML GUI for the next release. 
Here are a few things I think need to be improved:

- [x] when Electrum receives a payment, the "Receive Payment" dialog is turned into an almost empty screen, with this green "paid" icon and nothing else to show. This is not useful. We should close that dialog instead, and open the "Lightning payment details" dialog, or the corresponding 'on-chain transaction' dialog.
- [x] The "Lightning payment details" dialog could use some green color when an invoice has been paid, instead of just showing 'Status: settled'. Just show the same widget as in the Invoice dialog of the sender.
- [x] The requests that have been paid should be removed from the list of requests as soon as a request has been paid. Currently this is only done when we start a new session. Currently, if we open a paid request from the list, Electrum displays that almost empty screen with the green paid icon.
- [x] When we receive an on-chain payment, the 'on-chain transaction' dialog shows 'You can bump its fee to speed up its confirmation'. However, in that case, the only method available is CPFP, not RBF. I think we should not display that phrase in that case, because paying transaction fees is usually the responsibility of the sender, not of the receiver. (of course, we should keep the bump fee button, not remove it).
- [x] In the network dialog, the Nostr server list is not useful. I think it is sufficient to display that list in the 'nostr relays' dialog. 
- [x] The Nostr relays dialog has plenty of space, so the help could be displayed at the top of the screen, instead of requiring the user to press a button. The help text needs to be adapted to the QML gui, because it does not have Nostr wallet connect plugin. More importantly, the help text should explain that Electrum only connects to Nostr when required, and ephemerally. Otherwise people are going to think that we enforce a permanent connection to Nostr by default, and they will want to remove all the relays from the list (or even uninstall the app).
- [x] The server settings could merge the oneserver and auto-connect options using a combo box, like Qt. This is not just about saving space, it also better explains how these options work

accumulator:
#9882

### DIFF
--- a/electrum/gui/qml/components/NostrConfigDialog.qml
+++ b/electrum/gui/qml/components/NostrConfigDialog.qml
@@ -51,19 +51,15 @@ ElDialog {
             Layout.leftMargin: constants.paddingLarge
             Layout.rightMargin: constants.paddingLarge
 
-            RowLayout {
+            TextHighlightPane {
                 Layout.fillWidth: true
-                TextHighlightPane {
-                    Layout.fillWidth: true
-                    Label {
-                        text: qsTr('Enter the list of Nostr relays')
-                        width: parent.width
-                        wrapMode: Text.Wrap
-                    }
-                }
-                HelpButton {
-                    heading: Config.shortDescFor('NOSTR_RELAYS')
-                    helptext: Config.longDescFor('NOSTR_RELAYS')
+                Label {
+                    text: Config.shortDescFor('NOSTR_RELAYS') + '<br/><br/>' +
+                        qsTr('Nostr relays are used to send and receive submarine swap offers.') +
+                        ' ' + qsTr('For multisig wallets, nostr is also used to relay transactions to your co-signers.') +
+                        ' ' + qsTr('Connections to nostr are only made when required, and ephemerally.')
+                    width: parent.width
+                    wrapMode: Text.Wrap
                 }
             }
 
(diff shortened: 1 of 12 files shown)


electrum/gui/qml/components/NostrConfigDialog.qml
```diff
@@ -51,19 +51,15 @@ ElDialog {
...
-                        text: qsTr('Enter the list of Nostr relays')
-                        width: parent.width
-                        wrapMode: Text.Wrap
-                    }
-                }
-                HelpButton {
-                    heading: Config.shortDescFor('NOSTR_RELAYS')
-                    helptext: Config.longDescFor('NOSTR_RELAYS')
+                Label {
+                    text: Config.shortDescFor('NOSTR_RELAYS') + '<br/><br/>' +
```
f321x: The "Nostr relays" in the label seems redundant as the title of the window already contains the same:

Screenshot_20250603_101641-1

f321x:
I think it would be good to have a "Reset" button for the relay list, like in the Qt GUI, so the user doesn't get lost when accidentally deleting all relays and not knowing where to get relay urls from again.

accumulator:
> I think it would be good to have a "Reset" button for the relay list, like in the Qt GUI, so the user doesn't get lost when accidentally deleting all relays and not knowing where to get relay urls from again.

I think it's simpler to revert to the default list when the list is made empty.

ecdsa:
thanks

Issue #9893, opened 2025-06-03
qml: server dialog doesn't correctly change state


f321x:
### Description

i just noticed two (probably related) issues while reviewing https://github.com/spesmilo/electrum/pull/9888

When switching from "Manual server selection" to "Connect only to a single server", selecting another server and exiting with "Ok" the Network will still connect to multiple servers (instead of the expected single server). Only after a full restart the config comes into effect.

Also when switching modes from auto connect to single server, and directly going back into the `ServerConfigDialog`, the `ServerConnectModeComboBox` still shows the previous state (Auto Connect) instead of the new, previously selected single server mode.

accumulator:
ref #9895

PR #9895, opened 2025-06-03
qml: fix updating network settings

accumulator:
Previously the server parameters were each handled differently, e.g. auto-connect was only applied when updating Network.server and not when Config.autoConnect was updated. Similarly, updating Network.server did not restart the network, leading to >1 connection when Network.oneServer was set to True before updating Network.server.

Consolidate server parameter updates into a single call, remove the individual setters, and move Config.autoConnect and Config.autoConnectDefined to Network.


f321x:
Tested, fixes the issues described in https://github.com/spesmilo/electrum/issues/9893 for me

PR #9898, opened 2025-06-03
qt: windows: protect against screenshots and screen recordings

### LINKED ISSUE #9858: windows 11 "Microsoft Recall" might screenshot seed words. could we by default set some DRM flags to black out the seed?
### Description

Should we try to add safeguards against Windows 11 "Recall" taking automatic screenshots of the seed dialog?
There might be some DRM flags we could set (which the user could disable via the config) that blacks out the window in screenshots.

See https://signal.org/blog/signal-doesnt-recall/ :

> Signal Desktop now includes support for a new “Screen security” setting that is designed to help prevent your own computer from capturing screenshots of your Signal chats on Windows. This setting is automatically enabled by default in Signal Desktop on Windows 11.
>
> If you’re wondering why we’re only implementing this on Windows right now, it’s because the purpose of this setting is to protect your Signal messages from [Microsoft Recall](https://support.microsoft.com/en-us/windows/retrace-your-steps-with-recall-aa03f8a0-a78b-4b3e-b0a1-2eb8ac48701c).
> [...]
> 
> Fade to Black
>
> If you attempt to take a screenshot of Signal Desktop when screen security is enabled, nothing will appear. This limitation can be frustrating, but it might look familiar to you if you’ve ever had the audacity to try and take a screenshot of a movie or TV show on Windows. According to Microsoft’s [official developer documentation](https://learn.microsoft.com/en-us/windows/client-management/manage-recall#information-for-developers), setting the correct [Digital Rights Management](https://en.wikipedia.org/wiki/Digital_rights_management) (DRM) flag on the application window will ensure that “content won’t show up in Recall or any other screenshot application.” So that’s exactly what Signal Desktop is now doing on Windows 11 by default.

f321x:
fixes #9858 
Protects the main window, wallet wizard and `WindowModalDialog` instances from screenshots and screen recordings on Windows OS (including windows recall) by setting the `WDA_EXCLUDEFROMCAPTURE` flag on the windows (see https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowdisplayaffinity). 
This should cover all critical windows like the seed dialog and balances.
Introduces a config variable `GUI_QT_SCREENSHOT_PROTECTION` which is enabled by default and can (only on windows) be set from the 'Misc' tab of `SettingsDialog`.

### DIFF
--- a/electrum/gui/qt/__init__.py
+++ b/electrum/gui/qt/__init__.py
@@ -78,7 +78,8 @@ from electrum import constants
 from electrum.gui.common_qt.i18n import ElectrumTranslator
 from electrum.gui.messages import TERMS_OF_USE_LATEST_VERSION
 
-from .util import read_QIcon, ColorScheme, custom_message_box, MessageBoxMixin, WWLabel
+from .util import (read_QIcon, ColorScheme, custom_message_box, MessageBoxMixin, WWLabel,
+                   set_windows_os_screenshot_protection_drm_flag)
 from .main_window import ElectrumWindow
 from .network_dialog import NetworkDialog
 from .stylesheet_patcher import patch_qt_stylesheet
@@ -102,6 +103,20 @@ class OpenFileEventFilter(QObject):
             if len(self.windows) >= 1:
                 self.windows[0].set_payment_identifier(event.url().toString())
                 return True
+        return False
+
+
+class ScreenshotProtectionEventFilter(QObject):
+    def __init__(self):
+        super().__init__()
+
+    def eventFilter(self, obj, event):
+        if (
+            event.type() == QtCore.QEvent.Type.Show
+            and isinstance(obj, QWidget)
+            and obj.isWindow()
+        ):
+            set_windows_os_screenshot_protection_drm_flag(obj)
         return False
 
 
@@ -136,9 +151,12 @@ class ElectrumGui(BaseElectrumGui, Logger):
         QGuiApplication.setApplicationName("Electrum")
         self.gui_thread = threading.current_thread()
         self.windows = []  # type: List[ElectrumWindow]
-        self.efilter = OpenFileEventFilter(self.windows)
+        self.open_file_efilter = OpenFileEventFilter(self.windows)
         self.app = QElectrumApplication(sys.argv)
-        self.app.installEventFilter(self.efilter)
+        self.app.installEventFilter(self.open_file_efilter)
+        self.screenshot_protection_efilter = ScreenshotProtectionEventFilter()
+        if sys.platform in ['win32', 'windows'] and self.config.GUI_QT_SCREENSHOT_PROTECTION:
+            self.app.installEventFilter(self.screenshot_protection_efilter)
         self.app.setWindowIcon(read_QIcon("electrum.png"))
         self.translator = ElectrumTranslator()
         self.app.installTranslator(self.translator)
@@ -247,8 +265,11 @@ class ElectrumGui(BaseElectrumGui, Logger):
             return
         self._cleaned_up = True
         self.app.new_window_signal.disconnect()
-        self.app.removeEventFilter(self.efilter)
-        self.efilter = None
+        self.app.removeEventFilter(self.open_file_efilter)
+        self.open_file_efilter = None
+        # it is save to remove the filter, even if it has not been installed
+        self.app.removeEventFilter(self.screenshot_protection_efilter)
+        self.screenshot_protection_efilter = None
         # If there are still some open windows, try to clean them up.
         for window in list(self.windows):
             window.close()
--- a/electrum/gui/qt/util.py
+++ b/electrum/gui/qt/util.py
@@ -6,6 +6,7 @@ import platform
 import queue
 import os
 import webbrowser
+import ctypes
 from functools import partial, lru_cache, wraps
 from typing import (NamedTuple, Callable, Optional, TYPE_CHECKING, List, Any, Sequence, Tuple, Union)
 
@@ -1530,6 +1531,21 @@ def insert_spaces(text: str, every_chars: int) -> str:
     '''Insert spaces at every Nth character to allow for WordWrap'''
     return ' '.join(text[i:i+every_chars] for i in range(0, len(text), every_chars))
 
+
+def set_windows_os_screenshot_protection_drm_flag(window: QWidget) -> None:
+    """
+    sets the windows WDA_MONITOR flag on the window so windows prevents capturing
+    screenshots and microsoft recall will not be able to record the window
+    https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowdisplayaffinity
+    """
+    if sys.platform not in ('win32', 'windows'):
+        return
+    try:
+        window_id = int(window.winId())
+        WDA_MONITOR = 0x01
+        ctypes.windll.user32.SetWindowDisplayAffinity(window_id, WDA_MONITOR)
+    except Exception:
+        _logger.exception(f"failed to set windows screenshot protection flag")
 
 class _ABCQObjectMeta(type(QObject), ABCMeta): pass
 class _ABCQWidgetMeta(type(QWidget), ABCMeta): pass
--- a/electrum/gui/qt/settings_dialog.py
+++ b/electrum/gui/qt/settings_dialog.py
@@ -24,6 +24,7 @@
 # SOFTWARE.
 
 import ast
+import sys
 from typing import TYPE_CHECKING, Dict
 
 from PyQt6.QtCore import Qt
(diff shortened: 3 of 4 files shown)


electrum/gui/qt/util.py
```diff
@@ -1531,6 +1536,20 @@ def insert_spaces(text: str, every_chars: int) -> str:
...
+    """
+    sets the windows WDA_EXCLUDEFROMCAPTURE flag on the window so windows prevents capturing
+    screenshots and microsoft recall will not be able to record the window
+    https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowdisplayaffinity
+    """
+    if sys.platform not in ('win32', 'windows'):
+        return
+    try:
+        window_id = int(window.winId())
+        ctypes.windll.user32.SetWindowDisplayAffinity(window_id, 11)
```
SomberNight: note that the linked document uses hex numbers, so `WDA_EXCLUDEFROMCAPTURE` would be `0x11`.
Actually what we want is `WDA_MONITOR`, so `0x01`.

`WDA_EXCLUDEFROMCAPTURE` results in the window being invisible and the screenshot showing the window behind it.
`WDA_MONITOR` blacks out the window and the screenshot contains the black rectangle(s).
I think the black rectangle, so `0x01` is the desirable outcome, to make it obvious to the user what is going on.

I assume you tested with `11`, and it just happens to work and behave the same as `0x01` as the other bits are presumably currently unused.

```suggestion
        WDA_MONITOR = 0x01
        ctypes.windll.user32.SetWindowDisplayAffinity(window_id, WDA_MONITOR)
```

if you agree, note your docstring also mentions the flag by name

electrum/gui/qt/wizard/wallet.py
```diff
@@ -163,6 +164,9 @@ def __init__(self, config: 'SimpleConfig', app: 'QElectrumApplication', plugins:
             },
         })
 
+        if self.config.GUI_QT_SCREENSHOT_PROTECTION:
+            set_windows_os_screenshot_protection_drm_flag(self)
```
SomberNight: there is also e.g. `QEKeystoreWizard`, which is not covered by this

electrum/gui/qt/util.py
```diff
@@ -407,6 +408,10 @@ def __init__(self, parent, title=None):
         if title:
             self.setWindowTitle(title)
 
+        if hasattr(parent, 'config') and isinstance(parent.config, SimpleConfig):
+            if parent.config.GUI_QT_SCREENSHOT_PROTECTION:
+                set_windows_os_screenshot_protection_drm_flag(self)
```
SomberNight: This seems fragile but ok. If we wanted, we could also check if `self` has a `config`.

SomberNight: In general, my main complaint is that the most critical thing to protect, the seed (and xprv), is only implicitly protected. The protection is very far away, code-wise, from e.g. the SeedWidget. This seems fragile. Either the protection should be "close" to it, or if it's "far", then we should just find a way to black out everything.

As in, seeds usually appear in a SeedDialog, which gets protected here as it inherits WindowModalDialog.
(Seeds also appear in the wizard, and the wizard as a whole is protected explicitly.)

Ultimately it is a design decision what we want to "protect", e.g. the current code does not cover TxDialogs. I don't really care if TxDialogs and similar are covered. But seeds, master private keys, and other secret key material should be protected.

Seeds and xprvs we could easily protect "close":
<details>
 <summary>patch</summary>

```patch
diff --git a/electrum/gui/qt/seed_dialog.py b/electrum/gui/qt/seed_dialog.py
index 3a93363e09..d1a064223f 100644
--- a/electrum/gui/qt/seed_dialog.py
+++ b/electrum/gui/qt/seed_dialog.py
@@ -25,7 +25,7 @@
 
 from typing import TYPE_CHECKING
 
-from PyQt6.QtCore import Qt, pyqtSignal
+from PyQt6.QtCore import Qt, pyqtSignal, QEvent
 from PyQt6.QtGui import QPixmap
 from PyQt6.QtWidgets import (QVBoxLayout, QCheckBox, QHBoxLayout, QLineEdit,
                              QLabel, QCompleter, QDialog, QStyledItemDelegate,
@@ -40,6 +40,7 @@ from electrum.util import ChoiceItem
 from .util import (
     Buttons, OkButton, WWLabel, ButtonsTextEdit, icon_path, EnterButton,
     CloseButton, WindowModalDialog, ColorScheme, font_height, ChoiceWidget,
+    set_windows_os_screenshot_protection_drm_flag_recursive,
 )
 from .qrtextedit import ShowQRTextEdit, ScanQRTextEdit
 from .completion_text_edit import CompletionTextEdit
@@ -99,6 +100,9 @@ class SeedWidget(QWidget):
         self.config = config
         self.msg = msg
 
+        if self.config.GUI_QT_SCREENSHOT_PROTECTION:  # make sure seeds cannot be screenshot
+            set_windows_os_screenshot_protection_drm_flag_recursive(self)
+
         if options:
             self.seed_types = [
                 ChoiceItem(key=stype, label=label) for stype, label in (
@@ -189,6 +193,12 @@ class SeedWidget(QWidget):
 
         vbox.addWidget(self.seed_warning)
 
+    def changeEvent(self, event: QEvent):
+        if event.type() == QEvent.Type.ParentChange:
+            if self.config.GUI_QT_SCREENSHOT_PROTECTION:
+                set_windows_os_screenshot_protection_drm_flag_recursive(self)
+
+
     def seed_options(self):
         dialog = QDialog()
         dialog.setWindowTitle(_("Seed Options"))
@@ -397,6 +407,7 @@ class KeysWidget(QWidget):
         vbox = QVBoxLayout()
         self.setLayout(vbox)
 
+        self.config = config
         self.is_valid = is_valid
         self.text_e = ScanQRTextEdit(allow_multi=allow_multi, config=config)
         self.text_e.textChanged.connect(self.on_edit)
@@ -406,6 +417,9 @@ class KeysWidget(QWidget):
             vbox.addLayout(header_layout)
         vbox.addWidget(self.text_e)
 
+        if self.config.GUI_QT_SCREENSHOT_PROTECTION:  # make sure xprvs cannot be screenshot
+            set_windows_os_screenshot_protection_drm_flag_recursive(self)
+
     def get_text(self):
         return self.text_e.text()
 
@@ -416,6 +430,11 @@ class KeysWidget(QWidget):
             valid = False
         self.validChanged.emit(valid)
 
+    def changeEvent(self, event: QEvent):
+        if event.type() == QEvent.Type.ParentChange:
+            if self.config.GUI_QT_SCREENSHOT_PROTECTION:
+                set_windows_os_screenshot_protection_drm_flag_recursive(self)
+
 
 class SeedDialog(WindowModalDialog):
 
diff --git a/electrum/gui/qt/util.py b/electrum/gui/qt/util.py
index e5d3aa49e9..a40043d7f5 100644
--- a/electrum/gui/qt/util.py
+++ b/electrum/gui/qt/util.py
@@ -1550,6 +1550,13 @@ def set_windows_os_screenshot_protection_drm_flag(window: QWidget) -> None:
     except Exception:
         _logger.exception(f"failed to set windows screenshot protection flag")
 
+
+def set_windows_os_screenshot_protection_drm_flag_recursive(widget: QWidget) -> None:
+    while widget:
+        set_windows_os_screenshot_protection_drm_flag(widget)
+        widget = widget.parent()
+
+
 class _ABCQObjectMeta(type(QObject), ABCMeta): pass
 class _ABCQWidgetMeta(type(QWidget), ABCMeta): pass
 class AbstractQObject(QObject, ABC, metaclass=_ABCQObjectMeta): pass
```
</details>

However there are other instances of secret key stuff, e.g. WIF keys, and there might be even more that I can't recall atm:
https://github.com/spesmilo/electrum/blob/2024fa4507c73469fc3a33bc617c7332cd827aa8/electrum/gui/qt/main_window.py#L2101
https://github.com/spesmilo/electrum/blob/2024fa4507c73469fc3a33bc617c7332cd827aa8/electrum/gui/qt/main_window.py#L2419

So perhaps doing it from "far", with an approach of just blacking out every window would be better (?). E.g. maybe there is a way to get an event every time a window gets created. Not sure.

electrum/gui/qt/util.py
f321x: Yeah i agree showing a black screen seems better, i used `WDA_EXCLUDEFROMCAPTURE` as it was proposed in this microsoft blog https://learn.microsoft.com/en-us/windows/client-management/manage-recall#information-for-developers. However when i was testing windows showed black as i didn't pass the proper hex value.

Changed it to `WDA_MONITOR` now.

electrum/gui/qt/util.py
f321x: I now found a nice approach by installing an EventFilter on the qt `QApplication` instance which does black out every window created. This seems like the cleanest and most effective approach.

electrum/gui/qt/__init__.py
```diff
@@ -105,6 +106,18 @@ def eventFilter(self, obj, event):
...
 
+class ScreenshotProtectionEventFilter(QObject):
+    def __init__(self):
+        super().__init__()
+
+    def eventFilter(self, obj, event):
+        if event.type() == QtCore.QEvent.Type.Show \
+                and isinstance(obj, QWidget) \
+                and obj.isWindow():
+                set_windows_os_screenshot_protection_drm_flag(obj)
```
SomberNight: formatting: try to avoid having the same indentation level for both the condition and the inner code block, unless there is another line between. It is hard to see where the header ends and the body start.

E.g. this is more verbose but much more readable IMO:
```suggestion
        if (
            event.type() == QtCore.QEvent.Type.Show
            and isinstance(obj, QWidget)
            and obj.isWindow()
        ):
            set_windows_os_screenshot_protection_drm_flag(obj)
```

SomberNight (commented):
Nice! I like this global approach much better than special-casing certain windows.

(haven't tested yet)

electrum/gui/qt/__init__.py
f321x: Ok, changed it to your version: https://github.com/spesmilo/electrum/pull/9898/commits/ca9c4777d83381050dd426b7a10785d66b735426

accumulator:
Maybe guard the screenshot protection behind a window attribute (similar to QML) and check with `hasattr(obj, '__secureWindow')and getattr(obj, '__secureWindow', False)`?

e.g.
https://github.com/spesmilo/electrum/blob/6711b65b8c15516fd3e53d75f70dbd1a49fb7f2c/electrum/gui/qml/components/wizard/WCCreateSeed.qml#L10

Otherwise nothing can be screenshotted, which is a bit drastic.

f321x:
If the user wants to do screenshots there is a config option available in the SettingsDialog -> Misc tab. 
I think if the user enables the "Prevent screenshots" option it makes sense to protect as much as possible (best effort), that's what i would expect as user when enabling this, also protecting only specific windows it seems easy to forget something considering the large amount of possible windows in Qt. 
It also seems hard to decide what to protect and what not when not limiting the scope just to secrets but also trying to protect the users privacy, as Recall is recording the screen continuously.

SomberNight (approved):
Looks good. Thanks!

Issue #9728, opened 2025-04-11
supporting silent payments (bip352)


ecdsa:
### Description

@MorenoProg asked about how to implement silent payments in https://github.com/spesmilo/electrum/issues/7408#issuecomment-2796696251

Some preliminary remarks:
 - the "send to" logic does not require taproot, but the "spend from" logic does.
 - actually, even if we do not expose it to users, we already have limited support for taproot key-spend (see #9082)
 - `electrum-ecc` does not support musig yet, but recent `libsecp256k1` does, so I guess musig could be added to electrum-ecc "easily".

Note that for a feature to be merged in Electrum, it must not create a situation where user funds are not recoverable from seed. AFAICS, receiving silent payments would not only need a taproot wallet, but they would also break that assumption. Thus, if you want to implement the full "spend from" logic, I think it should be a plugin. 

Also, the receiver will need to scan the blockchain in a way that is currently not indexed by Electrum servers. So you need an extra channel to request that information. That's another reason to implement this as a plugin (maybe talking to a local bitcoind)

SomberNight:
ref https://github.com/spesmilo/electrum/issues/8847

MorenoProg:
## Planning Sender Silent Payment Support in Electrum
I'm currently working on implementing sender-side silent payments in Electrum (as described in [BIP-352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki)). In doing so, I've come across several points that need clarification regarding Electrum's wallet types, supported sighash flags, and how script types are handled.

Please correct me if I've misunderstood anything or if you spot inaccuracies in the assumptions below.

### Silent Payment Requirements from BIP-352

- At least one input MUST be from the (IFSSD) [Inputs For Shared Secret Derivation](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki#inputs-for-shared-secret-derivation) list. TL;DR: IFSSD = [p2tr, p2pkh, p2wpkh, p2sh-p2wpkh]
- No Input with SegWit version > 1 can be used
- The sender should sign with one of the sighash flags DEFAULT, ALL, SINGLE, NONE (ANYONECANPAY is unsafe). It is strongly recommended implementations use SIGHASH_ALL (SIGHASH_DEFAULT for taproot inputs) when possible

### Determining Wallet Compatibility with Silent Payments

[...]

Issue #8847, opened 2024-01-24
[ Feature Request ] Implement BIP352 sending (send to silent payment address)


cygnet3:
BIP352 (silent payments) is entering its last revision stage and probably won't change from now, so it's time to start a discussion about possibly integrating it with wallets such as electrum.

The BIP is split in a 'sending' and a 'receiving' section which can be integrated by wallets independently. Adding 'sending' support to electrum can probably be done quite easily and with minimal impact to the internals of electrum. Receiving is more sophisticated, and is outside the scope of this issue.

@Sosthene00 and I have been working on a rust library ([cygnet3/rust-silentpayments](https://github.com/cygnet3/rust-silentpayments)) that is intended to be used by wallets to add BIP352 support. We can use this library along with Python bindings (e.g. with [mozilla/uniffi-rs](https://github.com/mozilla/uniffi-rs)) to add sending support. Another option is to write a python native implementation.

edwargix:
Since libsecp256k1 is already a required dependency, I assume electrum can use the send/receive functionality implemented in https://github.com/bitcoin-core/secp256k1/pull/1519 (once it's merged) (?)

SomberNight:
Note: naively RBF-ing a tx (in a way the changes the input set) that pays to a silent payment address will ~burn the coins.

This means that when having the same Electrum wallet on both a laptop and a phone (or after restoring from seed), the device that did not create the tx - at the very least - would need to know not to allow RBF-ing it.

We have a similar problem for LN funding txs (https://github.com/spesmilo/electrum/issues/7072), where we decided to store this data in the input nSequence by not opting into RBF.

siwatanejo:
> This means that when having the same Electrum wallet on both a laptop and a phone (or after restoring from seed), the device that did not create the tx - at the very least - would need to know not to allow RBF-ing it.
> 
> We have a similar problem for LN funding txs (#7072), where we decided to store this data in the input nSequence by not opting into RBF.

What happens if fullRBF gets activated?

sethforprivacy:
A bounty has been opened on silentpayments.xyz from a trusted, anonymous source to implement sending support in Electrum to Silent Payment addresses.

I wanted to open an issue here to raise awareness and ensure that prospective contributors are aware of open bounties for Electrum from outside sources.

You can learn more about the bounty at https://silentpayments.xyz/docs/bounties/ and find more development resources focused on Silent Payments at https://silentpayments.xyz/docs/developers/.

[...]

PR #9900, opened 2025-06-03
Feature: Add Silent Payment sending support

MorenoProg:
This PR introduces support for sending silent payments (BIP-0352), integrated into the desktop Qt interface.

**Key Notes**

- Silent payments can be sent from standard wallets with BIP32 keystores only. This ensures all UTXOs in the wallet are eligible for the shared secret derivation.  Legacy or imported wallets may include uncompressed keys (for P2PKH), which are not compatible.
- Invoices involving silent payments are not persisted, as Electrum’s current system tracks invoice status via the scriptpubkeys of the invoice outputs. This approach doesn’t work for silent payments, since their final scriptpubkeys are only known at the point of transaction creation, not at invoice creation. Supporting this would have required intrusive changes to Electrum’s code base, with little practical benefit.
- ~RBF is disabled for silent payment transactions to prevent potential fund loss. If a user shares the same seed across two devices, broadcasts the transaction from one, and later bumps the fee on the other with added or changed inputs, the shared secret used in silent payments is broken, causing the recipient to miss the payment. If a transaction gets stuck in the mempool, users can still use CPFP as an alternative.~
RBF is allowed for silent payment transactions only when the fee bump is performed by the same wallet instance that created the transaction.


electrum/gui/qt/transaction_dialog.py
```diff
@@ -403,7 +403,7 @@ def on_context_menu_for_outputs(self, pos: QPoint):
         if addr := out.address:
             if self.wallet.is_mine(addr):
                 show_list += [(_("Address Details"), lambda: self.main_window.show_address(addr, parent=self))]
-            copy_list += [(_(f"Copy {"On Chain " if out.is_silent_payment() else ""}Address"),
+            copy_list += [(_(f"Copy {'On Chain ' if out.is_silent_payment() else ''}Address"),
```
f321x: `_()` translation of f-strings doesn't work, see https://github.com/spesmilo/electrum/blob/cae71222d29babec9387f7a05e261dab68376241/electrum/i18n.py#L52
You can use `.format()` instead.

electrum/wallet.py
```diff
@@ -2671,13 +2735,21 @@ def sign_transaction(
         if any(DummyAddress.is_dummy_address(txout.address) for txout in tx.outputs()):
             raise DummyAddressUsedInTxException("tried to sign tx with dummy address!")
 
+        if any(o.scriptpubkey == SILENT_PAYMENT_DUMMY_SPK for o in tx.outputs()):
+            raise SilentPaymentException("tried to sign tx with silent payment dummy SPK!")
+
```
SomberNight: is it not possible to just add a new `DummyAddress.SILENT_PAYMENT` property into the existing DummyAddress class? then this special-casing would not be needed here

accumulator:
According to updated [BIP21 specification](https://github.com/bitcoin/bips/blob/master/bip-0321.mediawiki) silent payment addresses should be put in a `sp` query attribute. I don't see that supported in this PR.

In fact, it is even disallowed to put a silent payment address in the BIP21 uri root:

_The bitcoinaddress body MUST be either a base58 P2SH or P2PKH address, bech32 Segwit version 0 address, bech32m Segwit address, or empty. Future address formats SHOULD instead be placed in query keys as optional payment instructions to provide backwards compatibility during upgrade cycles. The bitcoinaddress part of the URI MAY be left empty if there is at least one payment instruction provided in a query parameter, allowing for recipients wishing to avoid a standard on-chain fallback._

electrum/wallet.py
MorenoProg: Yes. The SILENT_PAYMENT_DUMMY_SPK is actually not needed, I can use a DummyAddress.SILENT_PAYMENT instead. I will apply these changes as soon as I can.

MorenoProg:
@accumulator Yes thanks! I completely missed that bip21 had specifications for bip352. I addressed this in commit [b1d4144](https://github.com/spesmilo/electrum/pull/9900/commits/b1d41447ebf8917c2e71da2358535b01029438ed). It was more complicated than I thought, especially dealing with the absence of fallback addresses in wallets that can't send silent payments.
For example 'bitcoin:?sp=validspaddr' is a valid bip32-URI and therefore a valid payment identifier imo, but it is unusable in non-silent-payment-wallets. I saw no other way but to inject wallet-silent-payment-capability into the relevant functions inside payment_identifier.py.
This could be simplified by marking the payment identifier as invalid during BIP21 URI parsing, but that would be misleading — the URI itself is valid, even if the current wallet cannot process it.

electrum/wallet.py
MorenoProg: I introduced DummyAddress.SILENT_PAYMENT in commit [fdb4836](https://github.com/spesmilo/electrum/pull/9900/commits/fdb48368291a1070e8dc6b62ef4faea93508f436). 
I ran into a small issue around a rare edge case.
If someone manually enters DummyAddress.SILENT_PAYMENT (intentionally or not), it leads to exceptions that currently fall through to the global handler.
I wasn’t sure whether to block this earlier by marking the payment_identifier as invalid — even though the address is technically valid.
Is it worth guarding against this?

electrum/payment_identifier.py
```diff
@@ -279,12 +283,13 @@ def parse(self, text: str):
             self.spk_is_address = is_address
             self.set_state(PaymentIdentifierState.AVAILABLE)
         elif self.contacts and (contact := self.contacts.by_name(text)):
-            if contact['type'] == 'address':
+            if contact['type'] in ('address', 'sp_address'):
```
accumulator, line 303: I don't think we need a separate contact type. We can determine the address type here and set the corresponding `bip21` dict kv.

electrum/payment_identifier.py
```diff
@@ -463,18 +468,35 @@ async def _do_notify_merchant(
...
             return self.bip70_data.get_outputs()
         elif self.multiline_outputs:
             return self.multiline_outputs
         elif self.spk:
-            return [PartialTxOutput(scriptpubkey=self.spk, value=amount)]
+            output = PartialTxOutput(scriptpubkey=self.spk, value=amount)
+            if output.address == DummyAddress.SILENT_PAYMENT:
+                sp_addr = self.parse_address(self.text) # parse again if e.g. a contact was entered
+                output.sp_addr = SilentPaymentAddress(sp_addr)
+            return [output]
```
accumulator: `self.spk` should not be (ab)used here to store a dummy value. Instead, make silent payment addresses a first class citizen and create a `PaymentIdentifierType` for it.

electrum/payment_identifier.py
```diff
@@ -463,18 +468,35 @@ async def _do_notify_merchant(
             if on_finished:
                 on_finished(self)
 
-    def get_onchain_outputs(self, amount):
+    def get_onchain_outputs(self, amount, bip21_use_fallback=False):
```
accumulator: Instead of `bip21_use_fallback` I'd prefer a more explicit `allow_silent_payment' kwarg.

electrum/payment_identifier.py
```diff
@@ -569,7 +595,7 @@ def _get_error_from_invoiceerror(self, e: 'InvoiceError') -> str:
                 error = _("Invoice requires unknown or incompatible Lightning feature") + f":\n{e!r}"
         return error
 
-    def get_fields_for_GUI(self) -> FieldsForGUI:
+    def get_fields_for_GUI(self, *, bip21_prefer_fallback=False) -> FieldsForGUI:
```
accumulator: Don't parameterize `get_fields_for_GUI`. It is fine to return a silent payment address here, even if the final payment is sent to a fallback address

accumulator (changes requested):

electrum/wallet.py
SomberNight: It is ok not letting the user manually create txs that pay to these DummyAddress(es).

Regarding what type of exception is raised to the user though, and from where, not sure, I haven't checked, sorry. I suppose the same issue is present for the existing DummyAddress(es)? If so, we can consider it a pre-existing issue unrelated to the PR.

accumulator:
> @accumulator Yes thanks! I completely missed that bip21 had specifications for bip352. I addressed this in commit [b1d4144](https://github.com/spesmilo/electrum/pull/9900/commits/b1d41447ebf8917c2e71da2358535b01029438ed). It was more complicated than I thought, especially dealing with the absence of fallback addresses in wallets that can't send silent payments. For example 'bitcoin:?sp=validspaddr' is a valid bip32-URI and therefore a valid payment identifier imo, but it is unusable in non-silent-payment-wallets. I saw no other way but to inject wallet-silent-payment-capability into the relevant functions inside payment_identifier.py. This could be simplified by marking the payment identifier as invalid during BIP21 URI parsing, but that would be misleading — the URI itself is valid, even if the current wallet cannot process it.

Yes, we currently do the same for lightning-only payment identifiers; The PI is valid, even if the wallet does not support lightning.

This leads to silent payment specific parameters and code in quite a few places.. I've added some review comments.

electrum/wallet.py
accumulator: We can check for `DummyAddress`es in `PaymentIdentifier` and mark as `INVALID` if found. This would let the GUI do the right thing automatically.

electrum/payment_identifier.py
MorenoProg: I created an icon to signal when an address is a silent payment (like this is done for lightning and on-chain addresses). I added a separate contact_type so that contact_list.py, and potentially other code, don’t need to re-parse the address just to decide which icon to display. Should this be dismissed?
Code in question:
https://github.com/spesmilo/electrum/blob/f4f050832931f8720de3e4da9eda6cb8a4ae0b30/electrum/gui/qt/contact_list.py#L122-L127

electrum/wallet.py
MorenoProg: I've added guards for DummyAddress.SILENT_PAYMENT in [1fc7ee5](https://github.com/spesmilo/electrum/pull/9900/commits/1fc7ee5861d25c687400c88247a86e4c3ec5bc46). I checked the behavior for the other two Dummys: CHANNEL is handled gracefully when the user tries to sign the tx. SWAP leads to an unhandled exception after being asked to enable Nostr.

electrum/payment_identifier.py
```diff
@@ -275,12 +278,23 @@ def parse(self, text: str):
...
+                    return # user manually entered DummyAddress.SILENT_PAYMENT
                 self.set_state(PaymentIdentifierState.AVAILABLE)
         elif self.parse_output(text)[0]:
             scriptpubkey, is_address = self.parse_output(text)
-            self._type = PaymentIdentifierType.SPK
-            self.spk = scriptpubkey
-            self.spk_is_address = is_address
+            if bitcoin.script_to_address(scriptpubkey) == DummyAddress.SILENT_PAYMENT:
+                sp_address = self.parse_address(text)
+                if sp_address == DummyAddress.SILENT_PAYMENT:
```
accumulator: You can check for all dummy addresses by using `DummyAddress.is_dummy_address(addr)`

electrum/payment_identifier.py
```diff
@@ -275,12 +278,23 @@ def parse(self, text: str):
...
+                    self.set_state(PaymentIdentifierState.INVALID)
+                    return # user manually entered DummyAddress.SILENT_PAYMENT
                 self.set_state(PaymentIdentifierState.AVAILABLE)
         elif self.parse_output(text)[0]:
             scriptpubkey, is_address = self.parse_output(text)
-            self._type = PaymentIdentifierType.SPK
-            self.spk = scriptpubkey
-            self.spk_is_address = is_address
+            if bitcoin.script_to_address(scriptpubkey) == DummyAddress.SILENT_PAYMENT:
+                sp_address = self.parse_address(text)
```
accumulator: Feel free to adapt `self.parse_output` to return also the silent payment address, so we don't need to re-parse the same string.

Note: there is one more use of `parse_output` outside of `payment_identifier.py`, used in the timelock plugin, see

https://github.com/spesmilo/electrum/blob/e30392cd161de8dfcf5981c4d528f82efd4645c3/electrum/plugins/timelock_recovery/qt.py#L404-L412

electrum/payment_identifier.py
accumulator: Hmm.. We could re-parse here, but `contact_list.py` is not consistent here either with openalias vs lnaddress..

Ok, leave as is for now..

accumulator (changes requested):

accumulator:
Have you considered OpenAlias resolving to a silent payment address? This would seem like an ideal combination..

https://github.com/MorenoProg/electrum/blob/1fc7ee5861d25c687400c88247a86e4c3ec5bc46/electrum/payment_identifier.py#L345-L358

accumulator:
Getting an exception when address_synchronizer emits `remove_transaction` (e.g. replaced by fee tx)

```
  File "/home/sander/projects/electrum/electrum/electrum/util.py", line 1206, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/electrum/interface.py", line 644, in wrapper_func
    return await func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/electrum/interface.py", line 670, in run
    await self.open_session(ssl_context=ssl_context)
  File "/home/sander/projects/electrum/electrum/electrum/interface.py", line 840, in open_session
               ^^^^^^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/packages/aiorpcx/curio.py", line 304, in __aexit__
    await self.join()
  File "/home/sander/projects/electrum/electrum/electrum/util.py", line 1381, in join
    task.result()
  File "/home/sander/projects/electrum/electrum/electrum/util.py", line 1511, in run_tasks_wrapper
    await self._run_tasks(taskgroup=taskgroup)
  File "/home/sander/projects/electrum/electrum/electrum/synchronizer.py", line 79, in _run_tasks
    async with taskgroup as group:
               ^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/packages/aiorpcx/curio.py", line 304, in __aexit__
    await self.join()
  File "/home/sander/projects/electrum/electrum/electrum/util.py", line 1381, in join
    task.result()
  File "/home/sander/projects/electrum/electrum/electrum/synchronizer.py", line 203, in _on_address_status
    await self._request_missing_txs(hist)
  File "/home/sander/projects/electrum/electrum/electrum/synchronizer.py", line 222, in _request_missing_txs
    async with OldTaskGroup() as group:
               ^^^^^^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/packages/aiorpcx/curio.py", line 304, in __aexit__
    await self.join()
  File "/home/sander/projects/electrum/electrum/electrum/util.py", line 1381, in join
    task.result()
  File "/home/sander/projects/electrum/electrum/electrum/synchronizer.py", line 244, in _get_transaction
    self.adb.receive_tx_callback(tx)
  File "/home/sander/projects/electrum/electrum/electrum/util.py", line 1235, in func_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/electrum/address_synchronizer.py", line 440, in receive_tx_callback
    self.add_transaction(tx, allow_unrelated=True)
  File "/home/sander/projects/electrum/electrum/electrum/address_synchronizer.py", line 325, in add_transaction
    self.remove_transaction(tx_hash2)
  File "/home/sander/projects/electrum/electrum/electrum/address_synchronizer.py", line 379, in remove_transaction
    self._remove_transaction(txid)
  File "/home/sander/projects/electrum/electrum/electrum/address_synchronizer.py", line 420, in _remove_transaction
    self.db.remove_silent_payment_address(txo.address)
  File "/home/sander/projects/electrum/electrum/electrum/json_db.py", line 42, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sander/projects/electrum/electrum/electrum/wallet_db.py", line 1484, in remove_silent_payment_address
    assert isinstance(onchain_addr, str)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

```

MorenoProg:
@accumulator Oh.. yes, this is very likely because remove_silent_payment_address is called with txo.address as argument, which of course is None when there is a non-address scriptpubkey. Then the assertion makes no sense... I haven't thought about this, thx! I will get to it and your other comments tonight.

electrum/payment_identifier.py
MorenoProg: I have `self.parse_output` return a 3-tuple now, with `sp_addr` as last element to avoid re-parsing.

electrum/payment_identifier.py
MorenoProg: I added guards for manually entering DummyAddress for all types (at least for spk, bip21 and multiline). If we care we could also guard against it when creating a contact, which is not handled yet.

MorenoProg:
> Have you considered OpenAlias resolving to a silent payment address? This would seem like an ideal combination..
> 
> https://github.com/MorenoProg/electrum/blob/1fc7ee5861d25c687400c88247a86e4c3ec5bc46/electrum/payment_identifier.py#L345-L358

I looked into it. Do you think the silent payment address can just be placed into the `recipient_address `-field? I haven't found any specs about handling different address types (like in bip21), all I found was this on the official OpenAlias Website:

> As this is an open standard that is meant to be extensible, defining additional pairs is up to you. Your client-side application may require certain key-value pairs as a minimum, and you should make that information easily available. If you have a use-case where you feel certain key-value pairs may provide widespread use and benefit, please reach out to us and we can include it in the standard if appropriate.

If we just place the sp_address into the `recipient_address `-field it would be quite easy to implement I guess. I checked the code, and only a few places would need adjustments. That said, there's a 'TODO' note about refactoring in one of the relevant sections, so I waited with making any changes for now.

https://github.com/MorenoProg/electrum/blob/1fc7ee5861d25c687400c88247a86e4c3ec5bc46/electrum/payment_identifier.py#L687-L694

accumulator:
> > Have you considered OpenAlias resolving to a silent payment address? This would seem like an ideal combination..
> > https://github.com/MorenoProg/electrum/blob/1fc7ee5861d25c687400c88247a86e4c3ec5bc46/electrum/payment_identifier.py#L345-L358
> 
> I looked into it. Do you think the silent payment address can just be placed into the `recipient_address `-field? I haven't found any specs about handling different address types (like in bip21), all I found was this on the official OpenAlias Website:
> 
> > As this is an open standard that is meant to be extensible, defining additional pairs is up to you. Your client-side application may require certain key-value pairs as a minimum, and you should make that information easily available. If you have a use-case where you feel certain key-value pairs may provide widespread use and benefit, please reach out to us and we can include it in the standard if appropriate.
> 

The unreleased [OA2](https://github.com/openalias/openalias-spec/blob/main/oa2.md) spec is a bit more formalized than [OA1](https://github.com/openalias/openalias-spec/blob/main/oa1.md), and even explicitly mentions BIP352. However we only implement OA1 currently.
 
> If we just place the sp_address into the `recipient_address `-field it would be quite easy to implement I guess. 

I think that would indeed be sufficient.

> I checked the code, and only a few places would need adjustments. That said, there's a 'TODO' note about refactoring in one of the relevant sections, so I waited with making any changes for now.

That TODO is orthogonal to this change, and there's currently no PR or branch changing this code, so feel free to make changes there.

MorenoProg:
@accumulator I added openalias recognizing sp_addresses in [0c5e83d](https://github.com/spesmilo/electrum/pull/9900/commits/0c5e83d3af7a8a69ed32ad28ab6b923886950a57). It was almost nothing to change, which makes me think I missed something. I tested it by mocking the call to the openalias-server in `resolve_openalias` in `contacts.py`:

```
    @classmethod
    async def resolve_openalias(cls, url: str) -> Dict[str, Any]:
         if url == "sp@test.com":
            await asyncio.sleep(1)
            return {
                'address': "valid_sp",
                'name': "Silent Payment Test",
                'type': 'openalias',
                'validated': True
            }
```
This worked fine, also in non-sp-wallets.

accumulator:
> It was almost nothing to change, which makes me think I missed something.

No, you didn't miss anything, that's sufficient.

You can even remove https://github.com/MorenoProg/electrum/blob/0c5e83d3af7a8a69ed32ad28ab6b923886950a57/electrum/payment_identifier.py#L701-L703 as that is stale code

MorenoProg:
There are still two design points where feedback is appreciated:

1. Password prompt: Since creating a silent payment requires access to the input private keys, the user is currently prompted for the password before building the transaction. They’ll get prompted again when signing. To avoid this double prompt, we could pass the password to confirm_tx_dialog (and also to transaction_dialog for previews), so it's reused. Would that be ok?

2. Disabling export/save for PSBTs involving silent payments: If a user exports a PSBT with silent payment outputs and edits it on another device, funds could be lost. We could block PSBT export in that case (informing why it's disabled), and only allow exporting `.txn` files where the transaction is already finalized. I’m aware that BIP375 addresses the issue of PSBTs involving silent payments. However, since the BIP is still in progress (open TODOs), and users who don’t interact with PSBTs aren’t affected, I think integrating this BIP can be deferred to a later stage.

EDIT: Both concerns addressed and changes implemented in [8b39e15](https://github.com/spesmilo/electrum/pull/9900/commits/8b39e15f324135329ced2eeacfbc55fc7526e95b).

macgyver13:
Tested ACK - tests/test_silent_payment.py 

Please consider implementing [sending test vectors](https://github.com/bitcoin/bips/blob/master/bip-0352/send_and_receive_test_vectors.json) as well
Additional References:
- [BDP-SP](https://github.com/bitcoindevkit/bdk-sp/blob/master/silentpayments/tests/functional_tests/send_bip352_vectors.rs)
- [BlueWallet](https://github.com/BlueWallet/SilentPayments/blob/c2e64ece7475bb7a962d7e8d14261007299639d5/tests/silent-payment.test.ts#L49)


Big thank you to everyone that has reviewed this PR to date.  
Gentle reminder @accumulator to review requested changes.

MorenoProg:
@macgyver13 Thanks for testing. I am using a slightly modified version of the test vectors, in which the mapping of `sp_address` -> `taproot_outputs` is preserved. Also, I cleaned the [test file](https://github.com/MorenoProg/electrum/blob/feature/sender-silent-payment/tests/bip-0352/sp_send_test_vectors.json) to remove irrelevant cases like receiving or taproot inputs.

macgyver13:
I missed the implementation of those test cases 😢. 
Perhaps printing each subtest would help future reviewers.

The attached patch has the following output:
```
collected 15 items

tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Simple_send_two_inputs] PASSED [  6%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Simple_send_two_inputs,_order_reversed] PASSED [ 13%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Simple_send_two_inputs_from_the_same_transaction] PASSED [ 20%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Simple_send_two_inputs_from_the_same_transaction,_order_reversed] PASSED [ 26%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Outpoint_ordering_byte-lexicographically_vs._vout-integer] PASSED [ 33%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Single_recipient_multiple_UTXOs_from_the_same_public_key] PASSED [ 40%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Multiple_outputs_multiple_outputs,_same_recipient] PASSED [ 46%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Multiple_outputs_multiple_outputs,_multiple_recipients] PASSED [ 53%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Multiple_outputs_with_labels_un-labeled_and_labeled_address;_same_recipient] PASSED [ 60%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Multiple_outputs_with_labels_multiple_outputs_for_labeled_address;_same_recipient] PASSED [ 66%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_silent_payment_create_outputs[Multiple_outputs_with_labels_un-labeled,_labeled,_and_multiple_outputs_for_labeled_address;_same_recipients] PASSED [ 73%]
tests/test_silent_payment.py::TestSilentPaymentCreateOutputs::test_private_keys_sum_to_zero PASSED                            [ 80%]
tests/test_silent_payment.py::TestSilentPaymentParseAddress::test_malformed_silent_payment_address PASSED                     [ 86%]
tests/test_silent_payment.py::TestSilentPaymentParseAddress::test_silent_payment_parse_address_success PASSED                 [ 93%]
tests/test_silent_payment.py::TestSilentPaymentTxCreation::test_merge_duplicate_tx_outputs PASSED                             [100%]
```
[subtest.patch](https://github.com/user-attachments/files/21352314/subtest.patch)

accumulator:
> There are still two design points where feedback is appreciated:
> 
>     1. Password prompt: Since creating a silent payment requires access to the input private keys, the user is currently prompted for the password before building the transaction. They’ll get prompted again when signing. To avoid this double prompt, we could pass the password to confirm_tx_dialog (and also to transaction_dialog for previews), so it's reused. Would that be ok?

Yes, that would be better. The current branch immediately signs the transaction as soon as the dialog is opened, which is different behavior than normal payments, and might be confusing.

>     2. Disabling export/save for PSBTs involving silent payments: If a user exports a PSBT with silent payment outputs and edits it on another device, funds could be lost. We could block PSBT export in that case (informing why it's disabled), and only allow exporting `.txn` files where the transaction is already finalized. I’m aware that BIP375 addresses the issue of PSBTs involving silent payments. However, since the BIP is still in progress (open TODOs), and users who don’t interact with PSBTs aren’t affected, I think integrating this BIP can be deferred to a later stage.
> 
> 
> EDIT: Both concerns addressed and changes implemented in [8b39e15](https://github.com/spesmilo/electrum/pull/9900/commits/8b39e15f324135329ced2eeacfbc55fc7526e95b).

Instead of passing `show_combine_menu` from outside `TxDialog` it's better to encapsulate it within the dialog, to avoid accidentily enabling these options when called from other locations (e.g. from the history tab)

See my commit https://github.com/accumulator/electrum/commit/7334e77ab3069fbe172293e19cf148209cc5aacb

electrum/gui/qt/send_tab.py
```diff
@@ -319,6 +328,8 @@ def make_tx(fee_policy, *, confirmed_only=False, base_tx=None):
                 is_sweep=is_sweep,
                 send_change_to_lightning=self.config.WALLET_SEND_CHANGE_TO_LIGHTNING,
                 merge_duplicate_outputs=self.config.WALLET_MERGE_DUPLICATE_OUTPUTS,
+                password=password,
+                mind_silent_payments=True # always mind silent payments to detect sending to a previous sp-onchain addr
```
accumulator, line 335: Shouldn't `mind_silent_payments` be conditional on `any(o.is_silent_payment() for o in outputs)`?

accumulator:
I think what's lacking still is a check to verify a `Tx` for correctness in the context of silent payments in `Abstract_Wallet.sign_transaction`. If for whatever reason a `Tx` is messing with its inputs right before signing, this has the potential to send funds into the void. Currently there is only a `SIGHASH` check.

electrum/gui/qt/send_tab.py
MorenoProg: At this point in the code, the outputs haven't been given to `wallet.add_silent_payment_output_info(output)` yet, so `any(o.is_silent_payment() for o in outputs)` will not detect a reused sp-onchain address. We could catch the address reuse already in `send_tab.py`, but I think its better to have all the edge-case-handling centralized in `wallet.make_unsigned_transaction`.

MorenoProg:
> 
> Yes, that would be better. The current branch immediately signs the transaction as soon as the dialog is opened, which is different behavior than normal payments, and might be confusing.

Agreed. I addressed this in [ab8486a](https://github.com/spesmilo/electrum/pull/9900/commits/ab8486ab57d23cc1681725fc99ffc2cc4b34ecfc). I re-added the sign button (disabled) and added a tool tip that informs the user about the automatic signing. My motivation for immediate signing is to never expose an incomplete sp-transaction to the user (not even for adding to history).


> Instead of passing `show_combine_menu` from outside `TxDialog` it's better to encapsulate it within the dialog, to avoid accidentily enabling these options when called from other locations (e.g. from the history tab)

Applied.

MorenoProg:
> I think what's lacking still is a check to verify a `Tx` for correctness in the context of silent payments in `Abstract_Wallet.sign_transaction`. If for whatever reason a `Tx` is messing with its inputs right before signing, this has the potential to send funds into the void. Currently there is only a `SIGHASH` check.

@accumulator I added an integrity check in [5486dd0](https://github.com/spesmilo/electrum/pull/9900/commits/5486dd0482edff33ff6d8ed7867892e98a1ab821). Right after tx-creation all the relevant silent payment elements (inputs and sp-outputs) are hashed and bound to the tx. This binding can happen only once. The integrity hash is verified against the actual transaction at the time of signing to ensure it hasn't been tampered with. Actually, we could be even more careful and also add this check again when broadcasting...

electrum/wallet.py
```diff
@@ -2104,6 +2126,44 @@ def distribute_amount(amount):
...
+            spks_by_sp_addr = create_silent_payment_outputs(input_privkeys, outpoints, sp_recipients)
+
+            # replace dummy spks with calculated taproot spks
+            sp_outputs = [o for o in tx.outputs() if o.is_silent_payment()]
+            for sp_output in sp_outputs:
+                tr_spk = spks_by_sp_addr.get(sp_output.sp_addr).pop()
+                sp_output.scriptpubkey = tr_spk
+            assert all(not lst for lst in spks_by_sp_addr.values()), "Found remaining sp_spks, but no more outputs left"
+
+            tx.set_rbf(False) # Transaction should be final
```
SomberNight: Not allowing bumping the fee sucks.
Could we discuss the rationale here, and condense it into a code comment / small paragraph so that it is recorded?

-----

What exactly must remain invariant when RBF-modifying a tx paying to a silent payment address?
AFAIU, roughly,
- the set of input prevouts must not change
- the set of output addresses must not change

So in particular, bumping the fee by decreasing a change output (if present) should be possible.
but e.g. bumping the fee by adding a new input burns the coins. (ref https://github.com/spesmilo/electrum/issues/8847#issuecomment-2154822196)
Is this correct?

MorenoProg: > What exactly must remain invariant when RBF-modifying a tx paying to a silent payment address? AFAIU, roughly,
> 
> * the set of input prevouts must not change

The set of silent payment *eligible* input prevouts must not change. Ineligible inputs (like uncompressed p2pkh) can be added without breaking the shared secret.

> * the set of output addresses must not change

The set of output addresses derived from a silent payment address must not change. It is no problem to add or remove outputs that are not sending silent payments.

> So in particular, bumping the fee by decreasing a change output (if present) should be possible. but e.g. bumping the fee by adding a new input burns the coins. (ref [#8847 (comment)](https://github.com/spesmilo/electrum/issues/8847#issuecomment-2154822196)) Is this correct?

Yes. Both decreasing change and decreasing payment is safe, but adding an eligible input would ~burn the coins. Under the assumption that the fee bumping happens on the same device, there would be no concerns at all, because the sp-outputs could be recalculated easily if an eligible input gets added.

I have thought about this quite a bit and I don't see a way where RBF can be considered safe. Some ideas:
* If the tx has change that is above a safe threshold, and we can be sure that the strategy on the fee-bumping-device is decreasing change first, RBF could be set. However, Electrum currently is adding new inputs under 'BumpFeeStrategy.PRESERVE_PAYMENT' before trying to decrease change (though theres a FIXME for this).
* If the fee-bumping-device is only allowed to add inputs when the tx is sp-ineligible (e.g. contains no sp-eligible input or no taproot output), RBF could be set. However, this would prevent many transactions from using RBF, as they appear silent-payment-eligible from the outside.
* If the fee-bumping-device only adds an ineligible input for sp-eligible tx's, RBF could be set. However, the txin-type is fixed for deterministic wallets in electrum, so there will be no access to such utxo's anyway (let me know if you can think of a trick here).

That being said, any of these points would require the user to be forced to perform an update, once sending silent payment is available. Also, all of this assumes that both devices run Electrum; users using other wallet software for fee bumping aren’t protected.

Unfortunately, [BIP-352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki) doesn't address this multi-device-problem, and I haven’t found any other resources discussing it either.

RBF would definitely be nice, but I don't think its feasible. At least CPFP is possible. Also, this PR informs the user explicitly that fee bumping will not be possible and enough fee should be paid.

Let me know if you have other ideas.

SomberNight: > If the tx has change that is above a safe threshold, and we can be sure that the strategy on the fee-bumping-device is decreasing change first, RBF could be set. However, Electrum currently is adding new inputs under 'BumpFeeStrategy.PRESERVE_PAYMENT' before trying to decrease change (though theres a FIXME for this).

https://github.com/spesmilo/electrum/blob/01694a363b3f1ae33d319de90fda8912d14948e1/electrum/wallet.py#L2294-L2305
Well, it's a bit complicated. `_bump_fee_through_coinchooser` does not necessarily add a new input, but yes, it might. Usually it just decreases the change, but it does it in a round-about way: it removes the change output(s), and then tries to tweak the tx to reach the desired fee: usually by re-adding a change output with a smaller value. (but the new change output *could* even end up using a different address) Also note that if the `multiple_change` option is enabled, it can get even messier: first all change outputs are removed, and then when re-adding change, it might decide to create fewer/more change outputs.

Also note that even `_bump_fee_through_decreasing_change` can end up changing the output set: if needed to reach the target fee level, it can decide to completely remove the change output.

-----

> Unfortunately, [BIP-352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki) doesn't address this multi-device-problem, and I haven’t found any other resources discussing it either.
> Let me know if you have other ideas.

Well, there is one general approach that we could use, that works both for RBF-ing silent payments and RBF-ing lightning funding transactions.

(1): So, the current approach is that we use the opt-in-RBF flag to store a single bit of information in a tx re whether it is *safe* to RBF it. If the flag is set, we allow all devices to RBF it. If it is not set, we don't allow any device to RBF it.

Note that historically, before `mempoolfullrbf=1`, this flag was even enforced by network policy. But I guess not anymore. Now we purely use it for our own wallet logic.

(2): We could extend (1), so that the original device that created the tx disregards the flag. That is, every time we create an interesting tx (e.g. one that pays to a silent payment addr), we store the txid (or similar) to a data structure in the wallet file (persisted). This could even be cleared up after the tx gets deeply mined.

When the user instructs the wallet to RBF a tx,
- we first check if the tx opted-in to RBF
  - if yes, early return: we can RBF it
- then we check if the tx was originally created by this current wallet instance, by looking it up in the persisted list
  - if no, early return: we cannot RBF it
    - question: do we allow `dscancel`? I guess we could always allow that.
- now we can apply arbitrary restrictions:
  - e.g. `dscancel` is allowed for lightning funding
  - for pay-to-silent-payment-addr, bump_fee is restricted to only touching amounts

I think `dscancel` we could allow unconditionally for every tx. (but that might be orthogonal to this PR)

Basically, if you restore from seed while there is a tx in the mempool, you lose the ability to RBF it.
You can only RBF "special" txs if they were created in the same wallet instance.

MorenoProg: > question: do we allow dscancel? I guess we could always allow that.

Yes, dscancel can safely be allowed.

---

Basically, in terms of silent payment, we can allow any of electrums current fee-bumping strategies except `_bump_fee_through_coinchooser`, as this might adds inputs. I checked both `_bump_fee_through_decreasing_change` and `_bump_fee_through_decreasing_payment`; they don't affect anything regarding silent payments.

So, you're suggestion sounds feasible and we actually could restrict RBF with the following guard:

```python
allow_adding_inputs = True
if wallet.can_send_silent_payment() and tx.is_silent_payment_eligible() and not wallet.db.tx_originated_from_here(tx):
      allow_adding_inputs = False
```

If adding inputs is disallowed, we can fallback to the other two strategies. If these are not applicable, we decline RBF and could suggest `dscancel` (`cpfp` will unfortunately not be possible at this point, since there is no or too little change).

---

Still, for this to be safe, the fee-bumping device must be running a version of Electrum that supports this guarding, otherwise the user is not protected... How could we approach this?

If you think it is still safe enough, I would give it a go now and start implementing this on top of this branch.

SomberNight: > Still, for this to be safe, the fee-bumping device must be running a version of Electrum that supports this guarding, otherwise the user is not protected... How could we approach this?

No released version of Electrum ever supported RBF-ing txs that do not signal RBF.
I am suggesting that txs that pay to a silent payment addr, should not signal RBF. (as is done currently in this branch) Hence existing Electrum versions will not allow RBF-ing them. New versions of Electrum however could sometimes decide to ignore the lack of signalling, and allow RBF-ing for txs that it originated from the current instance, and they can apply arbitrary rules in that case, such as not allowing adding inputs.

All this only works because `mempoolfullrbf=1` is now widespread in bitcoin node policy.

MorenoProg: Ok. I implemented this in [0938fd3](https://github.com/spesmilo/electrum/pull/9900/commits/0938fd3d317d2faca181e68328f0a713acaa778d). This supports all fee-bumping methods, provided the bumping is done on the same wallet instance that created the tx.

There was no need to persist more information than this branch already did. A silent payment tx will not signal RBF externally, though internally, RBF is considered allowed if `tx.is_rbf_enabled() or tx.contains_silent_payment()`. And only the wallet instance that created the tx will know about `tx.contains_silent_payment()`.

SomberNight: Nice. Thanks.
Could you also add a unit test for fee bumping a tx that pays to a silent payment addr?
And maybe even one case that pays to two different silent payments addrs in the same tx?

MorenoProg: Sure, but can you elaborate a bit on what you want to be tested exactly? Because the code I added to `bump_fee` doesn't touch any RBF logic (like strategy-methods); it just updates sp-addresses if necessary, otherwise it treats silent payment tx's as 'normal' tx's, and this is already covered in large in `test_wallet_vertical.py`.

Without writing redundant tests, I can only think of the following new meaningful test cases that:
* test if silent payment outputs are recalculated if the input set changes
* simulate a two device scenario and test if rbf is refused on the fee bumping device (though this seems trivial, since I'd just test if CannotBumpFee is raised with a tx that doesn't signal RBF).

SomberNight: Yes, that sounds good:

> simulate a two device scenario and test if rbf is refused on the fee bumping device

1. test that the originating wallet instance *can* bump the fee
    - then still in same test, close the wallet, and load it back again from disk, and see it can still bump the fee
      (see e.g. https://github.com/spesmilo/electrum/blob/d6c300ebcc122d635fc40f987e2a4a2abfe21a2e/tests/test_wallet.py#L108-L112)
2. test that another wallet instance *cannot* bump the fee
    _(this can be combined with the previous, ofc)_

> test if silent payment outputs are recalculated if the input set changes

3. create a tx that pays to e.g. two silent-payment-addrs and another simple address, maybe without creating a change output,  and then bump its fee by adding a new input and a change output.  the two SPAs should have the outputs recalculated but the simple address should remain untouched.

MorenoProg: Done in [d3ed311](https://github.com/spesmilo/electrum/pull/9900/commits/d3ed3110301bb337b25c8022a150dd0146d35800). Let me know if this suffices.

accumulator:
I propose to consider this PR for merging, as rebases are hard to re-review with this large delta, and the PR IMO looks pretty solid. If there's still remaining doubts w.r.t. loss of coins, we could add an 'experimental feature' warning when pasting/scanning a silent payment identifier.

MorenoProg:
Just in case @accumulator's comment has gone unnoticed, tagging @ecdsa and @SomberNight.

ecdsa:
@MorenoProg I am not keen to merge this feature into master. 

At the moment, there does not seem to be a realistic way to implement both sending and receiving silent payments in a light client. This implies that silent payments will probably remain a fringe feature, not something that will see mass adoption. However, this feature also comes at a significant cost, in terms of complexity and maintenance burden for developers. It touches a bunch of files, it has its own logic, and it  requires new developers to understand that logic in order not to break it. 

We recently opened Electrum to third-party plugins. Why not add this as a plugin instead?

MorenoProg:
> At the moment, there does not seem to be a realistic way to implement both sending and receiving silent payments in a light client.

@ecdsa Just last week, a PoC, [Frigate](https://github.com/sparrowwallet/frigate), was released that demonstrates a practical path to receiving silent payments in light clients. It uses an optimized tweak index and extends the Electrum protocol with methods that let a light client subscribe to Silent Payment addresses, similar to normal addresses. With this, a light client can also recover from seed.

> This implies that silent payments will probably remain a fringe feature, not something that will see mass adoption.

Mass adoption remains to be seen, but there is quite some [WIP](https://github.com/orgs/silent-payments/projects/2/views/4) across the silent payment ecosystem.

> We recently opened Electrum to third-party plugins. Why not add this as a plugin instead?

Your concerns listed in opening issue #9728 affect receiving only and you would prefer the receiving part as a plugin. Since sending and receiving are independent, it was my understanding that sending should belong in the core codebase.

craigraw:
> I am suggesting that txs that pay to a silent payment addr, should not signal RBF... New versions of Electrum however could sometimes decide to ignore the signalling

An observation here: any onchain wallet that ignores RBF signalling (due to the prevalence of `mempoolfullrbf=1`) risks ~burning user funds should the original transaction have included a silent payments derived address. This applies regardless of whether silent payments support has been added to the wallet, since the original transaction may have been created in a wallet that does support it.

SomberNight:
> > I am suggesting that txs that pay to a silent payment addr, should not signal RBF... New versions of Electrum however could sometimes decide to ignore the signalling
> 
> An observation here: any onchain wallet that ignores RBF signalling (due to the prevalence of `mempoolfullrbf=1`) risks ~burning user funds should the original transaction have included a silent payments derived address. This applies regardless of whether silent payments support has been added to the wallet, since the original transaction may have been created in a wallet that does support it.

Yes.  My idea was that a wallet could decide to allow RBF-ing a non-signalling tx *if* that tx was originally created by *that* wallet. In that restricted setting your concern does not apply.

However if there is a wallet out there that allows unrestricted RBF, even for txs not created by itself, that can result in burning coins, yes. This seems to be an inherent design issue with silent payments. E.g. it is not so safe anymore to move your seed between wallets.

ecdsa:
Sorry but I do not want new plugins to be merged, unless they have shown to be extremely popular. Please distribute your plugin as an external plugin, and submit it to https://plugins.electrum.org

The reason is, we need to focus and prioritize our own code development. Support for external plugins has been added in order to reduce the workload on developers, so that they do not spend their time serving requests.

MorenoProg:
@ecdsa Thanks for the response.

Quite a bit of time went into this PR, not just from my side, but also from contributors who reviewed it and asked to add unit tests, probably under the assumption that this was something you were seriously considering merging. That impression didn’t come out of nowhere: in Electrum issue #9728, the discussion was framed around how this could be implemented, not whether it should be supported at all.  

Given that, it’s a bit frustrating to reach this stage and find that there isn’t alignment on the feature direction itself.

More generally, this doesn’t seem like just a question of waiting for community demand. Electrum is in a position to shape that demand. If silent payments are something you want to see adopted, supporting them here would directly contribute to that.

If the current stance is that this isn’t a direction you want to take, that’s fine. but having that clarity earlier would have avoided a fair amount of unnecessary work.

Still happy and eager to revisit this if your position changes in the future.

PR #9906, opened 2025-06-04
CLI: complete wallet_path

ecdsa:
If the wallet_path passed to the RPC is a simple filename, interpret it as relative to the user wallets directory, rather than the current working directory.

This is a breaking change, might affect existing scripts


tests/test_daemon.py
```diff
@@ -229,10 +226,7 @@ async def test_wp_command_with_diskfile_wallet_no_password(self):
         await cmds.load_wallet(wallet_path=wpath, password=None)
         wallet = self.daemon.get_wallet(wpath)
         self.assertIsInstance(wallet, Abstract_Wallet)
-
-        # when using the CLI/RPC to run commands, the "wallet" param is a path:
         self.assertEqual("bitter grass shiver impose acquire brush forget axis eager alone wine silver",
-                         await cmds.getseed(wallet=wpath))
-        # in unit tests or custom code, the "wallet" param is often an Abstract_Wallet:
+                         await cmds.getseed(wallet_path=wpath))
```
SomberNight: > If the wallet_path passed to the RPC is a simple filename, interpret it as relative to the user wallets directory, rather than the current working directory.

You could add a test for this new behaviour. Would look very similar to these tests touched here.

PR #9908, opened 2025-06-04
additions to RELEASE-NOTES

accumulator:
feel free to rephrase


RELEASE-NOTES
```diff
@@ -102,6 +102,20 @@
    - support for hold invoices
    - support for onion messages
 
+ * Security:
+   - Mitigate against dust attacks; avoid spending from used addresses
```
SomberNight: if you mean `config.WALLET_FREEZE_REUSED_ADDRESS_UTXOS`, note that it is disabled by default

accumulator: clarified more that it is an option.

PR #9909, opened 2025-06-04
fix: remove unused anchor sweep local tx from history

f321x:
removes the local anchor sweeping tx from the transaction history if it is not required anymore, so it doesn't linger in the history and block the "Max" spend function if spending unconfirmed utxos is enabled.

### DIFF
--- a/electrum/txbatcher.py
+++ b/electrum/txbatcher.py
@@ -325,20 +325,33 @@ class TxBatch(Logger):
             prev_txid, index = prevout.to_str().split(':')
             if not self.wallet.adb.db.get_transaction(prev_txid):
                 continue
+            spender_txid = self.wallet.adb.db.get_spent_outpoint(prev_txid, index)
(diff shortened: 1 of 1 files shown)


ecdsa:
can you try this instead?

```diff --git a/electrum/txbatcher.py b/electrum/txbatcher.py
index a7fb8aca5..20b321ca4 100644
--- a/electrum/txbatcher.py
+++ b/electrum/txbatcher.py
@@ -445,7 +445,7 @@ class TxBatch(Logger):
             can_broadcast, wanted_height = self._can_broadcast(v, base_tx)
             if can_broadcast:
                 to_sweep_now[k] = v
-            else:
+            elif not v.is_anchor():
                 self.wallet.add_future_tx(v, wanted_height)
         while True:
             if not to_pay and not to_sweep_now and not self._should_bump_fee(base_tx):

```

EDIT: no, probably not going to solve that

ecdsa:
replaced by #9917

PR #9916, opened 2025-06-05
fix: use single long lived swap transport in qeswaphelper

f321x:
changes qeswaphelper to share a single, long lived transport instance instead of opening new transports to do swaps and fetch offers. 
This allows to continuosly fetch offers, so events which get returned later by slow relays don't get lost and the fee values stay updated. 
Also fixes a race causing the list to miss swap servers sometimes, as the current implementation fetches only until `swap_manager.is_initialized()` is set, which will get set as soon as an event of the configured swap server is received. So if the event of the configured swap server is received as first, all server events coming in after it would get ignored.
Also replaces the threads with asyncio coroutines.

### DIFF
--- a/electrum/gui/qml/components/NostrSwapServersDialog.qml
+++ b/electrum/gui/qml/components/NostrSwapServersDialog.qml
@@ -51,6 +51,19 @@ ElDialog {
                     clip: true
                     model: swaphelper.availableSwapServers
 
+                    Connections {
+                        target: swaphelper
+                        function onOffersUpdated() {
+                            listview.model = null
+                            listview.model = swaphelper.availableSwapServers
+                            listview.forceLayout()
+                            if (dialog.selectedPubkey) {
+                                listview.currentIndex = swaphelper.availableSwapServers.indexFor(dialog.selectedPubkey)
+                            }
+                            console.log("swapserver list refreshed")
+                        }
+                    }
+
                     delegate: ItemDelegate {
                         width: ListView.view.width
                         height: itemLayout.height
--- a/electrum/gui/qml/components/NostrSwapServersDialog.qml
+++ b/electrum/gui/qml/components/NostrSwapServersDialog.qml
@@ -54,9 +54,6 @@ ElDialog {
                     Connections {
                         target: swaphelper
                         function onOffersUpdated() {
-                            listview.model = null
-                            listview.model = swaphelper.availableSwapServers
-                            listview.forceLayout()
                             if (dialog.selectedPubkey) {
                                 listview.currentIndex = swaphelper.availableSwapServers.indexFor(dialog.selectedPubkey)
                             }
--- a/electrum/gui/qml/qeswaphelper.py
+++ b/electrum/gui/qml/qeswaphelper.py
@@ -66,9 +66,8 @@ class QESwapServerNPubListModel(QAbstractListModel):
         self._services = []
         self.endResetModel()
 
-    def initModel(self, items: Sequence['SwapOffer']):
-        self.beginInsertRows(QModelIndex(), len(items), len(items))
-        self._services = [{
+    def offer_to_model(self, x: 'SwapOffer'):
+        return {
             'npub': x.server_npub,
             'percentage_fee': x.pairs.percentage,
             'mining_fee': x.pairs.mining_fee,
@@ -76,9 +75,39 @@ class QESwapServerNPubListModel(QAbstractListModel):
             'max_forward_amount': x.pairs.max_forward,
             'max_reverse_amount': x.pairs.max_reverse,
             'timestamp': age(x.timestamp),
-        } for x in items]
-        self.endInsertRows()
-        self.countChanged.emit()
+        }
+
+    def updateModel(self, items: Sequence['SwapOffer']):
+        offers = items.copy()
+
+        remove = []
+
+        for i, x in enumerate(self._services):
+            if matches := list(filter(lambda offer: offer.server_npub == x['npub'], offers)):
+                # update
+                self._services[i] = self.offer_to_model(matches[0])
+                index = self.index(i, 0)
+                self.dataChanged.emit(index, index, self._ROLE_KEYS)
+                offers.remove(matches[0])
+            else:
+                # add offer to remove items
+                remove.append(i)
+
+        # # remove offers from model
+        for ri in reversed(remove):
+            self.beginRemoveRows(QModelIndex(), ri, ri)
+            self._services.pop(ri)
+            self.endRemoveRows()
+
+        # add new offers
+        if offers:
+            self.beginInsertRows(QModelIndex(), len(self._services), len(self._services) + len(offers) - 1)
+            for offer in offers:
+                self._services.append(self.offer_to_model(offer))
+            self.endInsertRows()
+
+        if offers or remove:
+            self.countChanged.emit()
 
     @pyqtSlot(str, result=int)
     def indexFor(self, npub: str):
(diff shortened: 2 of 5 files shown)


electrum/gui/qml/components/NostrSwapServersDialog.qml
```diff
@@ -51,6 +51,19 @@ ElDialog {
                     clip: true
                     model: swaphelper.availableSwapServers
 
+                    Connections {
+                        target: swaphelper
+                        function onOffersUpdated() {
+                            listview.model = null
+                            listview.model = swaphelper.availableSwapServers
+                            listview.forceLayout()
```
accumulator: Reassigning the model in the UI is not needed, `QESwapServerNPubListModel` should be updated correctly instead.

electrum/gui/qml/qeswaphelper.py
```diff
@@ -459,6 +498,7 @@ def init_swap_slider_range(self):
         self.leftVoidChanged.emit()
         self.rightVoidChanged.emit()
 
+        self.sliderPos = 0
```
accumulator: Reset to 0 is a bit drastic. Instead it should be clamped to the new limits.

electrum/gui/qml/qeswaphelper.py
```diff
@@ -356,62 +360,78 @@ def init_swap_manager(self):
...
+
+                        return QESwapHelper.State.Initializing
 
                     if not swap_manager.is_initialized.is_set():
                         self.userinfo = _('Initializing...')
                         fut = asyncio.run_coroutine_threadsafe(wait_initialized(), get_asyncio_loop())
-                        fut.result()
+                        state = self.state = fut.result()
+                        if state in (QESwapHelper.State.NoService, None):
+                            return
```
accumulator: `self.state` should never be `None`

electrum/gui/qml/qeswaphelper.py
```diff
@@ -356,62 +360,78 @@ def init_swap_manager(self):
...
+
+                while not self.stop_swap_manager.is_set():
+                    # keep fetching new incoming offer events
+                    # the slider range will not get updated continuously as it would irritate the user
+                    if isinstance(transport, NostrTransport):
+                        if self.recent_offers != transport.get_recent_offers():
+                            self._logger.debug(f"received new swap offer")
+                            self.recent_offers = transport.get_recent_offers()
+                            self.offersUpdated.emit()
+                    time.sleep(1)
```
accumulator: 1s is a bit tight, I think 10s makes more sense here.

electrum/gui/qml/qeswaphelper.py
```diff
@@ -356,62 +360,78 @@ def init_swap_manager(self):
...
+                    self.setReadyState()
+
+                while not self.stop_swap_manager.is_set():
+                    # keep fetching new incoming offer events
+                    # the slider range will not get updated continuously as it would irritate the user
+                    if isinstance(transport, NostrTransport):
+                        if self.recent_offers != transport.get_recent_offers():
+                            self._logger.debug(f"received new swap offer")
+                            self.recent_offers = transport.get_recent_offers()
+                            self.offersUpdated.emit()
```
accumulator, line 445: Instead of emitting a signal here, and letting the GUI handle reset & update the model (see above), you can instead update the model directly. This should be done smartly (i.e. don't reset and refill the model, but compare the list with the model and update/insert/remove individual offers)

electrum/gui/qml/qeswaphelper.py
```diff
@@ -420,9 +440,28 @@ def npubSelectionCancelled(self):
...
+        if isinstance(self.swap_transport, NostrTransport):
+            swap_manager = self._wallet.wallet.lnworker.swap_manager
+            pair = self.swap_transport.get_offer(self._wallet.wallet.config.SWAPSERVER_NPUB)
+            swap_manager.update_pairs(pair.pairs)
+
+    @pyqtSlot()
+    def initSwapSliderRange(self):
         lnworker = self._wallet.wallet.lnworker
         swap_manager = lnworker.swap_manager
+        # update swap_manager to the newest pair so the slider represents the latest available data
```
accumulator: This comment's wording implies the slider is modified.

accumulator:
I've replaced `QESwapServerNPubListModel.initModel()` with `QESwapServerNPubListModel.updateModel()` that keeps the model in sync without reset+repopulate

Note to self: this should be used in `QEServerListModel` as well..

accumulator:
There is an issue when no npub or offline npub is selected.
After the provider list is opened, then cancelled, the 'Choose swap provider' button gets disabled

electrum/gui/qml/qeswaphelper.py
f321x: Ok, i clamp it in the range now

electrum/gui/qml/qeswaphelper.py
f321x: right, that was a leftover. Thanks, removed it.

electrum/gui/qml/qeswaphelper.py
f321x: I'd prefer to keep it at 1 sec as it improves the UX and doesn't really cost anything. Imagine the user opens the swap window and after initialization clicks on the swapserver list to see what is available. If now a slow relay returns an additional server it would be shown directly, while if we wait 10 sec the user might close the dialog again and miss the additional event coming in later.
Do you see any serious downside to using a 1 sec sleep here?

electrum/gui/qml/qeswaphelper.py
f321x: changed the comment to better represent why the update is done (so the newest available values are used)

f321x:
Thanks for the improvements. I could not reproduce your issue, for me the button didn't get disabled when not selecting a swap provider. Does the issue still exist with these new changes?

electrum/gui/qml/qeswaphelper.py
accumulator: Not really, it only lives as long as the swaphelper, so it's not a constant drain.
:+1:

accumulator:
> Thanks for the improvements. I could not reproduce your issue, for me the button didn't get disabled when not selecting a swap provider. Does the issue still exist with these new changes?

Yes. If `undefinedNPub` is emitted from 
https://github.com/f321x/electrum-fork/blob/beaf8d4f81a36199e2fabd9af9f6d41bf658796d/electrum/gui/qml/qeswaphelper.py#L434

.. and no provider is selected (click outside window to close), the button stays disabled.

I think we need one more `QESwapHelper.State` to indicate we have providers to choose from, but no npub selected.

f321x:
Thats weird, it doesn't stay disabled for me, see:
[Screencast_20250606_130508.webm](https://github.com/user-attachments/assets/85f4faf4-be40-4baa-8ae5-7861cfd91d48)
also works on Android.

Which seems to make sense, as if the npub selection is cancelled `npubSelectionCancelled()` gets called and will set `self.state = QESwapHelper.State.NoService` enabling the 'Choose swap provider' button as it's condition is:
```
                enabled: _swaphelper.state != SwapHelper.Initializing
                    && _swaphelper.state != SwapHelper.Success
                    && _swaphelper.availableSwapServers.count
``` 
what am i missing?

accumulator:
I see correct behavior if there is no `swapserver_npub` in config, but if there is an existing `swapserver_npub` in config that doesn't respond or has no offers it triggers this.

f321x:
Ah yeah now i see the issue too. The problem was `npubSelectionCancelled()` did only check if the swapserver npub is None, if an (invalid/unavaliable) npub was already configured it did not change the state to `NoService`.
I changed it now to check if the configured npub is in the list of available swapservers so it also handles unavailable/invalid npubs.

PR #9925, opened 2025-06-09
change plugins keyfile path to XDG Base Directory spec

### LINKED ISSUE #9922: linux plugin keyfile path may not work on some distros
### Description

On linux we currently store the root owned plugin keyfile in `/etc/electrum/plugins_key`, however the `/etc` directory may not work on some distributions like NixOS and Fedora Silverblue as there the `/etc` directory is read only or its content can get deleted on system rebuilds/updates.
Is there a reason why we can't just store the keyfile in `config.electrum_path()` or alternatively at `$HOME/.config/electrum/`?

f321x:
changes the plugins keyfile to either
XDG_CONFIG_HOME/electrum/plugins_key, 
or if empty to HOME/.config/electrum/plugins_key 
or if HOME is not set either (which should not happen in practice), to /etc/electrum/plugins_key.

Also adds a `bring_to_front()` call after installing external plugin so the PluginsDialog is brought back in front.

see: https://specifications.freedesktop.org/basedir-spec/latest/
fixes #9922


electrum/plugin.py
```diff
@@ -355,9 +378,14 @@ def _write_key_to_root_file_linux(self, key_hex: str) -> None:
...
-
         sh_command: str = self._posix_plugin_key_creation_command(key_hex)
+        # create base directory
+        base_dir_path = os.path.dirname(self.keyfile_posix)
+        try:
+            os.makedirs(base_dir_path, exist_ok=True)
+        except OSError:
+            # fallback to creating the dir with root permissions too
+            sh_command = f"mkdir -p {base_dir_path} && " + sh_command
         commands = ['pkexec', 'sh', '-c', sh_command]
```
SomberNight, line 389: What if `base_dir_path` contains whitespaces or quotes? Previously we used a hardcoded path but now with changing to reading env vars (`XDG_CONFIG_HOME` and `HOME), anything is game.
See e.g. https://docs.python.org/3/library/shlex.html for escaping stuff.

ecdsa:
I am not expert enough to tell whether this is safe. It seems to me that the permissions/owner of a file located in a user-owned directory might be less restricted than in a root owned directory.

I think it is safer to postpone this for now. There might be other ways to address #9922

PR #9933, opened 2025-06-10
fix: prevent KeyError if _fail_swap gets called multiple times

f321x:
If `_fail_swap()` gets called multiple times (e.g. from callbacks) it raises a `KeyError` as the swap got already popped from `self._swaps`.
In theory `_fail_swap` unregisters itself from the lnwatcher callback but the callback may is scheduled multiple times before it has the chance to unregister itself.
Lost the traceback, but `LNWatcher.on_event_adb_added_tx()` called `_claim_swap()` multiple times on a swap after startup which caused this to show up in my logs.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -361,7 +361,8 @@ class SwapManager(Logger):
         self.lnwatcher.remove_callback(swap.lockup_address)
         if not swap.is_funded():
             with self.swaps_lock:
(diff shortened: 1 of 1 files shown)


accumulator:
Maybe log in case the key is not found, so we do have some indication in cases where we do expect a key to be present?

f321x:
@accumulator yeah good idea, i added a debug log now.

PR #9939, opened 2025-06-11
allow testnet4 and signet in crash reporter

f321x:
allows to submit crash reports from testnet 4 and signet too in addition to testnet3 and mainnet.


SomberNight (approved):
Oh, good catch!

PR #9945, opened 2025-06-12
build-osx: add docs for testing with self-signed certificate

f321x:
Adds docs to osc README on how to create a self-signed certificate and use it to sign the dmg. Also adds a check in `extract_sig.sh` to exclude the notarization files from the comparison if `IS_NOTARIZED` is explicitly set `false`.


SomberNight:
Thanks!

PR #9950, opened 2025-06-13
release.sh: better support for 3rd party signers

f321x:
adds binary downloading from website if the specified signer is not one of the two hardcoded signers with sftp access to the airlock server. 
This makes it easier for other signers to run the script and create signatures for spesmilo/electrum-signatures.

### DIFF
--- a/contrib/release.sh
+++ b/contrib/release.sh
@@ -6,8 +6,9 @@
 # - assumes all keys are available, and signs everything
 # This script, for other builders:
 # - builds all reproducible binaries,
-# - downloads binaries built by the release manager (from airlock), compares and signs them,
-# - and then uploads sigs
+# - downloads binaries built by the release manager (from airlock if SFTPUSER, else from website),
+#   compares and signs them,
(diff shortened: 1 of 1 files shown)


contrib/release.sh
```diff
@@ -254,8 +283,14 @@ if [ -z "$RELEASEMANAGER" ] ; then
         signame="$fname.$GPGUSER.asc"
         gpg --sign --armor --detach $PUBKEY --output "$PROJECT_ROOT/dist/sigs/$signame" "$fname"
     done
+
+    if [ "$SSHUSER" != "$GPGUSER" ]; then
     # upload sigs
     ELECBUILD_UPLOADFROM="$PROJECT_ROOT/dist/sigs/" "$CONTRIB/upload.sh"
+    else
```
SomberNight: Could you indent the inner body of the `if` so that it's deeper than the header? just like the other branch (`else`) below

SomberNight:
Thanks!

PR #9958, opened 2025-06-15
address_synchronizer: add a cache in front of get_utxos()

nabijaczleweli:
`get_utxos()` is called pretty often, both spuriously and on focus change, on tab switch, &c.

It blocks as it iterates, functionally, *every* address the wallet knows of.

On large wallets (like testnet `vpub5VfkVzoT7qgd5gUKjxgGE2oMJU4zKSktusfLx2NaQCTfSeeSY3S723qXKUZZaJzaF6YaF8nwQgbMTWx54Ugkf4NZvSxdzicENHoLJh96EKg` from #6625 with 11k TXes and 10.5k addresses), this takes 1.3s of 100%ed CPU usage, basically in a loop from the UI thread.

`get_utxos()` is 50-70% of the flame-graph when sampling a synced wallet process. Here's a few:
2968630-2025-06-15T16:07:37+02:00
3022580-2025-06-15T16:25:33+02:00
3637852-2025-06-15T19:37:53+02:00
(these are interactive if you save them and open them in your browser locally).

This data is a function of the block-chain state, and we have hooks that notify us of when the block-chain state changes: we can just cache the result and only re-compute it then.

For example, here's a trace log where `get_utxos()` has
```python
print(end - start, len(domain), block_height)
```
and a transaction is clearing:
```
1.3775344607420266 10540 4507192
0.0010390589013695717 10540 4507192 cached!
0.001393263228237629 10540 4507192 cached!
0.0009001069702208042 10540 4507192 cached!
0.0010241391137242317 10540 4507192 cached!
...
0.00207632128149271 10540 4507192 cached!
0.001397700048983097 10540 4507192 cached!
invalidate_cache
1.4686454269103706 10540 4507192
0.0012429207563400269 10540 4507192 cached!
0.0015075239352881908 10540 4507192 cached!
0.0010459059849381447 10540 4507192 cached!
0.0009669591672718525 10540 4507192 cached!
...
on_event_blockchain_updated
invalidate_cache
1.3897203942760825 10540 4507193
0.0010689008049666882 10540 4507193 cached!
0.0010420521721243858 10540 4507193 cached!
...
invalidate_cache
1.408584670163691 10540 4507193
0.001336586195975542 10540 4507193 cached!
0.0009196233004331589 10540 4507193 cached!
0.0009176661260426044 10540 4507193 cached!
...
```
about 30s of low activity.

Without this patch, the UI is prone to freezing, running behind, and I wouldn't be surprised if UI thread blocking on Windows ends up crashing to some extent as the issue notes. In the log, this manifests as a much slower but consistent stream of full 1.3-1.4s updates during use, and every time the window is focused.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -22,6 +22,7 @@
 # SOFTWARE.
 
 import asyncio
+import copy
 import threading
 import itertools
 from collections import defaultdict
@@ -99,8 +100,14 @@ class AddressSynchronizer(Logger, EventListener):
         self.threadlocal_cache = threading.local()
 
         self._get_balance_cache = {}
+        self._get_utxos_cache = {}
 
         self.load_and_cleanup()
+
+    @with_lock
+    def invalidate_cache(self):
+        self._get_balance_cache.clear()
+        self._get_utxos_cache.clear()
 
     def diagnostic_name(self):
         return self.name or ""
@@ -203,7 +210,7 @@ class AddressSynchronizer(Logger, EventListener):
     @event_listener
     @with_lock
     def on_event_blockchain_updated(self, *args):
-        self._get_balance_cache = {}  # invalidate cache
+        self.invalidate_cache()
         self.db.put('stored_height', self.get_local_height())
 
     async def stop(self):
@@ -335,7 +342,7 @@ class AddressSynchronizer(Logger, EventListener):
                         pass
                     else:
                         self.db.add_txi_addr(tx_hash, addr, ser, v)
-                        self._get_balance_cache.clear()  # invalidate cache
+                        self.invalidate_cache()
             for txi in tx.inputs():
                 if txi.is_coinbase_input():
                     continue
@@ -353,7 +360,7 @@ class AddressSynchronizer(Logger, EventListener):
                 addr = txo.address
                 if addr and self.is_mine(addr):
                     self.db.add_txo_addr(tx_hash, addr, n, v, is_coinbase)
-                    self._get_balance_cache.clear()  # invalidate cache
+                    self.invalidate_cache()
                     # give v to txi that spends me
                     next_tx = self.db.get_spent_outpoint(tx_hash, n)
                     if next_tx is not None:
@@ -405,7 +412,7 @@ class AddressSynchronizer(Logger, EventListener):
             remove_from_spent_outpoints()
             self._remove_tx_from_local_history(tx_hash)
             for addr in itertools.chain(self.db.get_txi_addresses(tx_hash), self.db.get_txo_addresses(tx_hash)):
-                self._get_balance_cache.clear()  # invalidate cache
+                self.invalidate_cache()
             self.db.remove_txi(tx_hash)
             self.db.remove_txo(tx_hash)
             self.db.remove_tx_fee(tx_hash)
@@ -503,7 +510,7 @@ class AddressSynchronizer(Logger, EventListener):
     def clear_history(self):
         self.db.clear_history()
         self._history_local.clear()
-        self._get_balance_cache.clear()  # invalidate cache
+        self.invalidate_cache()
 
     @with_lock
     def _get_tx_sort_key(self, tx_hash: str) -> Tuple[int, int]:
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
I think this may be cacheable even more strongly: since the query and the cache key include the block-chain height, we may be able to *never* clear this cache, assuming we will only be queried about block heights that exist and not future ones we'd return effectively-incomplete data for (or if we can reject those queries from the cache by finding they're, like, `>= local_height`)? This feels reasonable to me, but I'm not 100% sure.

electrum/address_synchronizer.py
```diff
@@ -200,11 +200,11 @@ def start_network(self, network: Optional['Network']) -> None:
...
             self.register_callbacks()
 
+    @with_lock
     @event_listener
     def on_event_blockchain_updated(self, *args):
-        with self.lock:
-            self._get_balance_cache = {}  # invalidate cache
-            self.db.put('stored_height', self.get_local_height())
+        self._get_balance_cache = {}  # invalidate cache
+        self.db.put('stored_height', self.get_local_height())
```
SomberNight: This technically accidentally works, but it would make much more sense to chain the function decorators in the other order:

```py
    @event_listener
    @with_lock
    def on_event_blockchain_updated(self, *args):
```
The only reason it works as-is, is because `event_listener` refers to functions by `name`.

However, the order I recommend fails for another reason, that I now fixed in https://github.com/spesmilo/electrum/commit/4887fb3e7ff8d842648b8477903bc5ce24941894, so you would need to rebase.

nabijaczleweli: Yes, I initially wrote it in that order (spuriously) and it exploded, so I changed it. Rebased + applied.

SomberNight:
> I think this may be cacheable even more strongly: since the query and the cache key include the block-chain height, we may be able to never clear this cache, assuming we will only be queried about block heights that exist and not future ones we'd return effectively-incomplete data for (or if we can reject those queries from the cache by finding they're, like, >= local_height)? This feels reasonable to me, but I'm not 100% sure.

Yeah, there could never exist two blocks at the same height, right? Chain splits and reorgs are myths. :P
Also, what about unbounded memory growth? No point in keeping data for old blocks in memory.
Also, addresses can be added and removed, e.g. to an Imported Wallet, which can change the set of utxos without blockchain events. (this is handled indirectly in `add_transaction` and `_remove_transaction`)

There's a reason the existing invalidate cache calls are there for the balance cache. :)

electrum/address_synchronizer.py
```diff
@@ -970,6 +976,10 @@ def get_utxos(
         if excluded_addresses:
             domain = set(domain) - set(excluded_addresses)
         mempool_height = block_height + 1  # height of next block
+        cache_key = ''.join(domain) + f"{mature_only}{confirmed_funding_only}{confirmed_spending_only}{nonlocal_only}{block_height}"
```
SomberNight: I would prefer clear separators for clarity.
While it cannot happen in today's Bitcoin, imagine if "aaa" and "bbb" and "aaabbb" were all valid addresses. You should really specify if you are caching "aaa,bbb" or "aaabbb".
This is why the cache_key in get_balance was designed as such.
```suggestion
        cache_key = sha256(
            ','.join(sorted(domain))
            + f";{mature_only};{confirmed_funding_only};{confirmed_spending_only};{nonlocal_only};{block_height}"
        )
```

nabijaczleweli:
I tend to agree here.

Also, I've been clicking around and broadly using the wallet in this state and I can't really tell all that well when the real 1.3s recomputes happen anyway (except when clicking Pay for the first time, but that's over-shadowed ~3x by the layout(?) delay which I'm also tracking and hoping to correct), so I don't think there would be much point to cranking this even more, and this seems sufficient.

electrum/address_synchronizer.py
nabijaczleweli: Yeah, fair cop, the full cache key is apparently just under 500k here so the SHA256 (~0.1ms) is worth it as well; applied

electrum/address_synchronizer.py
```diff
@@ -970,6 +976,13 @@ def get_utxos(
         if excluded_addresses:
             domain = set(domain) - set(excluded_addresses)
         mempool_height = block_height + 1  # height of next block
+        cache_key = sha256(
+            ','.join(sorted(domain))
+            + f";{mature_only};{confirmed_funding_only};{confirmed_spending_only};{nonlocal_only};{block_height}"
+        )
+        cached = self._get_utxos_cache.get(cache_key)
+        if cached is not None:
+            return cached
```
SomberNight: Note that callers might mutate the return value, which would affect the cache and hence subsequent calls. That is an unexpected behaviour change compared to current master.

Instead we should take copies.
```suggestion
            return copy.deepcopy(cached)
```

electrum/address_synchronizer.py
```diff
@@ -987,6 +1000,7 @@ def get_utxos(
                     continue
                 coins.append(txo)
                 continue
+        self._get_utxos_cache[cache_key] = coins
```
SomberNight: note that a deepcopy should be taken here as well, otherwise the first caller gets a chance to mutate the value in the cache.

electrum/address_synchronizer.py
nabijaczleweli: Both applied

SomberNight:
Thanks.

PR #9963, opened 2025-06-16
fix: build-wine/unsign.sh fail if osslsigncode is missing

f321x:
If `osslsigncode` is missing `build-wine/unsign.sh` fails silently with status code 0 causing the `release.sh` script to interpret the result as valid and skipping the actual comparison of the windows binaries.


contrib/build-wine/unsign.sh
```diff
@@ -8,7 +8,7 @@ cd $here
 
 if ! which osslsigncode > /dev/null 2>&1; then
     echo "Please install osslsigncode"
-    exit
+    exit 1
```
SomberNight, line 14: o.O

contrib/build-wine/unsign.sh
```diff
@@ -8,7 +8,7 @@ cd $here
...
 fi
 
 # exit if command fails
```
SomberNight: in a similar spirit, can we put `set -e` at the top of the file?
most of our other scripts do that iirc

f321x: Ok i have added `set -e` in https://github.com/spesmilo/electrum/pull/9963/commits/ee03df3a62eb10f4229f2f9c04cf2d077329d944

SomberNight: oh but I mean the `set -e` was already there but not at the top. I should have said "move", not "put".

There is a `set -e` just below this line. I blame github for not letting me comment on lines too far away from the current diff :P

f321x: Oh right, i did not see that. *Moved* it to the top now in https://github.com/spesmilo/electrum/pull/9963/commits/eba21231c9b70094a2987622eda3fa1f2fc87865

SomberNight:
Thanks. Good find.

PR #9966, opened 2025-06-16
block windows installer if Electrum is still running

### LINKED ISSUE #6748: 4.0.5 always shows "No FX rate available"
After upgrading to 4.0.5, I see "No FX rate available" in the status bar. Switching back to 4.0.4 shows the correct value. This impacts all currencies for me.

I am not using tor or a VPN or anything like that.

f321x:
Adds a check to the Windows nsis installer and uninstaller that will show an error popup if Electrum is still running to prevent a broken installation that happen if the installer only partly overwrites the existing installation (fixes #6748).
This works by trying to open the .exe files in the Electrum installation directory in append mode, if Electrum is still running the installer will not be able to open the .exe and the check fails. Based on https://github.com/Electron-Cash/Electron-Cash/pull/2185

### DIFF
--- a/contrib/build-wine/electrum.nsi
+++ b/contrib/build-wine/electrum.nsi
@@ -88,6 +88,55 @@
   !insertmacro MUI_LANGUAGE "English"
 
 ;--------------------------------
+;Functions
+
+!macro CreateEnsureNotRunning prefix operation
+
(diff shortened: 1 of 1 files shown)


accumulator:
tested, works beautifully!

PR #9967, opened 2025-06-16
Populate tx_status_cache before it's used, not after. Only compute short_id when requested 

nabijaczleweli:
The endInsertRows() call triggers the main calls to get_data_for_role(): in the current order, the try/except goes to the exception every time

In the updated order, it doesn't do it ever

We have full control over what's in the cache, and we always fully populate it, so we can get rid of the fallback

### DIFF
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -317,6 +317,13 @@ class HistoryModel(CustomModel, Logger):
             balance += node._data['value'].value
             node._data['balance'] = Satoshis(balance)
 
+        # update tx_status_cache
+        self.tx_status_cache.clear()
+        for txid, tx_item in transactions.items():
+            if not tx_item.get('lightning', False):
+                tx_mined_info = self._tx_mined_info_from_tx_item(tx_item)
+                self.tx_status_cache[txid] = self.window.wallet.get_tx_status(txid, tx_mined_info)
+
         new_length = self._root.childCount()
         self.beginInsertRows(QModelIndex(), 0, new_length-1)
         self.transactions = transactions
@@ -336,12 +343,6 @@ class HistoryModel(CustomModel, Logger):
                 end_date = self.transactions.value_from_pos(len(self.transactions) - 1).get('date') or end_date
             self.view.years = [str(i) for i in range(start_date.year, end_date.year + 1)]
             self.view.period_combo.insertItems(1, self.view.years)
-        # update tx_status_cache
-        self.tx_status_cache.clear()
-        for txid, tx_item in self.transactions.items():
-            if not tx_item.get('lightning', False):
-                tx_mined_info = self._tx_mined_info_from_tx_item(tx_item)
-                self.tx_status_cache[txid] = self.window.wallet.get_tx_status(txid, tx_mined_info)
         # update counter
         num_tx = len(self.transactions)
         if self.view:
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -124,9 +124,10 @@ class HistoryNode(CustomNode):
                 status_str = format_time(int(timestamp))
         else:
             tx_hash = tx_item['txid']
-            txpos_in_block = tx_item.get('txpos_in_block')
-            if txpos_in_block is not None and txpos_in_block >= 0:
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
On the wallet from #9958/#6625, the overall time to pre-compute this is about 90ms. When instrumented as
```python
from timeit import default_timer as timer
GLOBULE = 0
GLOBULE_COUNT = 0
class HistoryNode(CustomNode):

    model: 'HistoryModel'

    def get_data_for_role(self, index: QModelIndex, role: Qt.ItemDataRole) -> QVariant:
        start = timer()
        ret = self.get_data_for_role_(index, role)
        end = timer()
        global GLOBULE
        global GLOBULE_COUNT
        GLOBULE += end - start
        GLOBULE_COUNT += 1
        return ret

    def get_data_for_role_(self, index: QModelIndex, role: Qt.ItemDataRole) -> QVariant:
        # real impl
```
with the data zeroed at top of refresh() and read out at bottom, I see, before:
```
1.4012  132348
1.4551  132348
1.4161  132348
```
after:
```
3.9735  132348
3.8653  132348
4.1097  132348
```

Just measuring refresh e2e, without the instrumentation above (but *excluding* the ~2.4s wallet.get_full_history() takes, so measuring from after `if transactions == self.transactions: return`):
before
```
6.713636453729123
6.914032700937241
6.382696805987507
```
after
```
3.733156899921596
3.6829894930124283
3.5451462320052087
```

So this is a ~2x speed-up.

nabijaczleweli:
The short_id patch wins another 4% on average:
```
3.512978475075215
3.5296032559126616
3.4849150171503425
```

ecdsa:
why is this cache not part of `wallet.py `?

nabijaczleweli:
One has to assume it's because the model views a snapshot of the data, so the cache is bundled with the snapshot and also for performance?

nabijaczleweli:
Is this missing anything?

f321x:
For me this throw tons of exceptions on startup (seemingly multiple for each txid):
```
  File "/home/user/code/electrum-fork/electrum/gui/qt/history_list.py", line 83, in lessThan
    item1 = self.sourceModel().data(source_left, ROLE_SORT_ORDER)
  File "/home/user/code/electrum-fork/electrum/gui/qt/custom_model.py", line 94, in data
    return node.get_data_for_role(index, role)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/history_list.py", line 132, in get_data_for_role
    status, status_str = self.model.tx_status_cache[tx_hash]
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'd4decca1cc2f2c110056864021a49c861b121e0dc52a7e9e36fd1d532bc2a9e7'
```
Seems like there is some situation in which the cache is not properly populated before calling `get_data_for_role()`.

nabijaczleweli:
I can't repro this but fall-back re-added, performance appears unaffected

SomberNight (approved):
Looks good. Thanks for looking into this.

electrum/gui/qt/history_list.py
```diff
@@ -124,9 +124,10 @@ def get_data_for_role(self, index: QModelIndex, role: Qt.ItemDataRole) -> QVaria
                 status_str = format_time(int(timestamp))
         else:
             tx_hash = tx_item['txid']
-            txpos_in_block = tx_item.get('txpos_in_block')
-            if txpos_in_block is not None and txpos_in_block >= 0:
-                short_id = f"{tx_item['height']}x{txpos_in_block}"
+            if col == HistoryColumns.SHORT_ID:
+                txpos_in_block = tx_item.get('txpos_in_block', -1)
+                if txpos_in_block >= 0:
```
SomberNight, line 129: actually `tx_item['txpos_in_block']` is explicitly None for unconfirmed/local txs, so this was not an equivalent transformation

we would need `txpos_in_block = tx_item.get('txpos_in_block') or -1`, to convert explicit `None`s to `-1`

fixed in https://github.com/spesmilo/electrum/commit/38f83940a0a1e9a4d324767c952766e47122ec27

PR #9968, opened 2025-06-16
Pre-compute sort keys, don't launder getting them through get_data_for_role()

nabijaczleweli:
endInsertRows() triggers a sort, the history sort/filter proxy model then launders the comparison requests through get_data_for_role() with a custom key, which is then especially looked up in there, but custom-flattened in the proxy model. This is OO hell

Measuring refresh() from after `if transactions == self.transactions: return` to the final return, sorting the wallet from #9958/#6625 takes ~3.5s and 132348 get_data_for_role() calls

Beside that get_data_for_role() is called maybe 10 times per on-hover? It's immaterial

Thus: compute the sorting keys when constructing the HistoryNode and use them directly in the comparison. This decouples get_data_for_role() from the sorter
and speeds the sort up by ~2x, now being mostly made up of the python interpreter overhead(?) for upcalls to lessThan()

A dict wins with a list for the lookup a tiny bit (and is 100x better code-wise):
```
Benchmark 1: dict
  Time (mean ± σ):       1918.039234 ms ±    63.688609 ms
  Range (min … max):     1848.120000 ms …  2046.648838 ms    11 runs

Benchmark 2: list
  Time (mean ± σ):       1945.052027 ms ±   164.019588 ms
  Range (min … max):     1757.185085 ms …  2432.649423 ms    11 runs

Summary
  'dict' ran
    1.014084 ± 0.090082 times faster than 'list'
```

Draft because based on #9967

### DIFF
--- a/electrum/gui/qt/custom_model.py
+++ b/electrum/gui/qt/custom_model.py
@@ -62,12 +62,18 @@ class CustomModel(QtCore.QAbstractItemModel):
         parent.addChild(self, node)
 
     def index(self, row, column, _parent=None):
+        # Performance-critical function
+
         if not _parent or not _parent.isValid():
             parent = self._root
         else:
             parent = _parent.internalPointer()
 
-        if not QtCore.QAbstractItemModel.hasIndex(self, row, column, _parent):
+        # Open-coded
+        #   if not QtCore.QAbstractItemModel.hasIndex(self, row, column, _parent):
+        # the implementation is equivalent but it's in C++,
+        # so VM entries take up inordinate amounts of time (up to 25% of refresh()):
+        if row < 0 or column < 0 or row >= self.rowCount(_parent) or column >= self._columncount:
             return QtCore.QModelIndex()
 
         child = parent.child(row)


nabijaczleweli:
> mostly made up of the python interpreter overhead(?)

that's not true actually, appx. 25% is spent in CustomModel.index()!
full refresh:
full-refresh

just endInsertRows:
just-endInsertRows
(save & open for interactive)

nabijaczleweli:
Open-coding hasIndex() buys another 500ms/40% overall, down to ~1.3s

nabijaczleweli:
Rebased, should be g2g

electrum/gui/qt/history_list.py
```diff
@@ -449,14 +440,12 @@ class Columns(MyTreeView.BaseColumnsEnum):
...
         FIAT_CAP_GAINS = enum.auto()
         TXID = enum.auto()
-        SHORT_ID = enum.auto()  # ~SCID
 
     filter_columns = [
         Columns.STATUS,
         Columns.DESCRIPTION,
         Columns.AMOUNT,
         Columns.TXID,
-        Columns.SHORT_ID,
```
SomberNight: Indeed the SHORT_ID column is never shown. It was added for this filtering functionality exclusively, in https://github.com/spesmilo/electrum/commit/f2dc651c9b0176a06dcaa53e354784a963ebcf65.

E.g. if you are testing with the wallet from https://github.com/spesmilo/electrum/issues/6625#issuecomment-724912330, press Ctrl+F and type `1326030x59`. That filters the list to the tx at block height `1326030`, pos in block `59`.

If it's not that expensive to compute, I think it's a nice (though certainly niche) feature.

electrum/gui/qt/history_list.py
```diff
@@ -129,11 +129,7 @@ def get_data_for_role(self, index: QModelIndex, role: Qt.ItemDataRole) -> QVaria
                 if txpos_in_block >= 0:
                     short_id = f"{tx_item['height']}x{txpos_in_block}"
             conf = tx_item['confirmations']
-            try:
-                status, status_str = self.model.tx_status_cache[tx_hash]
-            except KeyError:
-                tx_mined_info = self.model._tx_mined_info_from_tx_item(tx_item)
-                status, status_str = window.wallet.get_tx_status(tx_hash, tx_mined_info)
```
SomberNight: bad rebase?
in https://github.com/spesmilo/electrum/pull/9967#issuecomment-3005021888, @f321x said this was raising KeyError for him, after which you readded the fallback?

nabijaczleweli: bad rebase

electrum/gui/qt/history_list.py
nabijaczleweli: Indeed, even after all these years I didn't even realise there's search functionality, so I didn't see any way to actually get at the column. Re-added, it only got computed when it's requested so it's effectively free, and storing it in `sort_keys` makes it fully free.

nabijaczleweli:
Rebased, all applied.

SomberNight:
Thanks. Looks good.

Issue #9974, opened 2025-06-23
qml: LnurlPayRequestDialog has "Pay" button disabled at first


SomberNight:
After scanning an lnurl-pay qr code (or pasting from clipboard), the "Pay" button at the bottom is disabled by default.
If I change the amount away and back, the button becomes enabled.

SomberNight:
It looks like this might be a regression from https://github.com/spesmilo/electrum/commit/376d88115445c918935d43ae92e0a52bfe3da42b, which sets `amountOverride` to 0  (invisible to the UI).

```patch
diff --git a/electrum/gui/qml/components/LnurlPayRequestDialog.qml b/electrum/gui/qml/components/LnurlPayRequestDialog.qml
index ce6b85f5f2..46549ef2a0 100644
--- a/electrum/gui/qml/components/LnurlPayRequestDialog.qml
+++ b/electrum/gui/qml/components/LnurlPayRequestDialog.qml
@@ -17,10 +17,33 @@ ElDialog {
 
     padding: 0
 
+    property bool _trigger_update_valid: false
     property bool commentValid: comment.text.length <= invoiceParser.lnurlData['comment_allowed']
     property bool amountValid: amountBtc.textAsSats.satsInt >= parseInt(invoiceParser.lnurlData['min_sendable_sat'])
         && amountBtc.textAsSats.satsInt <= parseInt(invoiceParser.lnurlData['max_sendable_sat'])
-    property bool valid: commentValid && amountValid
+    property bool valid: commentValid && amountValid && _trigger_update_valid
+
+    Component.onCompleted: {
+        console.log('heyheyhey11. valid=' + valid)
+        console.log('heyheyhey12. commentValid=' + commentValid)
+        console.log('heyheyhey13. amountValid=' + amountValid)
+        console.log('heyheyhey14. _trigger_update_valid=' + _trigger_update_valid)
+        _trigger_update_valid = true
+        console.log('heyheyhey21. valid=' + valid)
+        console.log('heyheyhey22. commentValid=' + commentValid)
+        console.log('heyheyhey23. amountValid=' + amountValid)
+        console.log('heyheyhey24. _trigger_update_valid=' + _trigger_update_valid)
+    }
+
+    Connections {
+        target: invoiceParser
+        function onAmountOverrideChanged() {
+            console.log('heyheyhey31. valid=' + valid)
+            console.log('heyheyhey32. _trigger_update_valid=' + _trigger_update_valid)
[...]

PR #9975, opened 2025-06-23
qml: set BtcField.textAsSats initial value to match its text

SomberNight:
textAsSats was being set to 0 initially, regardless of actual text, only healing after onTextChanged.

fixes https://github.com/spesmilo/electrum/issues/9974

-----

@accumulator could you review if this looks ok?

- not sure exactly what https://github.com/spesmilo/electrum/commit/376d88115445c918935d43ae92e0a52bfe3da42b was fixing
- e.g. do we need to explicitly initialise the `text` field to `""` or similar?

### DIFF
--- a/electrum/gui/qml/components/controls/BtcField.qml
+++ b/electrum/gui/qml/components/controls/BtcField.qml
@@ -32,5 +32,5 @@ TextField {
         }
     }
 
-    Component.onCompleted: textAsSats = Config.unitsToSats('')
+    Component.onCompleted: amount.textChanged()
 }
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -155,7 +155,7 @@ class QEInvoice(QObject, QtEventListener):
         return self._amountOverride
 
     @amountOverride.setter
-    def amountOverride(self, new_amount):
+    def amountOverride(self, new_amount: QEAmount):
         self._logger.debug(f'set new override amount {repr(new_amount)}')
         self._amountOverride.copyFrom(new_amount)
         self.amountOverrideChanged.emit()


accumulator:
> textAsSats was being set to 0 initially, regardless of actual text, only healing after onTextChanged.
> 
> fixes #9974
> 
> @accumulator could you review if this looks ok?
> 
>     * not sure exactly what [376d881](https://github.com/spesmilo/electrum/commit/376d88115445c918935d43ae92e0a52bfe3da42b) was fixing
> 
>     * e.g. do we need to explicitly initialise the `text` field to `""` or similar?

I couldn't find the discussion leading up to `376d881`, but I think I found the reason; Without `376d881`, when trying to create an amountless LN invoice, the button for LN enables, but does nothing due to `amountBtc.textAsSats` being `null` in `ReceiveDetailsDialog`. Annoyingly, quirky qml logs no error, but also seems to skip evaluating the `else` branches.

The fix you propose works (by making sure `textAsSats` is properly initialized) when passing a `text` parameter to `BtcField` at construction time, but it skips updating the associated `FiatField` via the `onTextChanged` handler.

I think we can simply trigger the signal: 

```
    Component.onCompleted: amount.textChanged()
```

SomberNight:
> but it skips updating the associated `FiatField` via the `onTextChanged` handler.

but then that must have already been the case even before this PR, mustn't it?

> I couldn't find the discussion leading up to 376d881

We should all remember to add references in the commit message :P
Backlinks are the next best thing (github comment linking to commit), but they are not tracked well by the web interface.

SomberNight:
> I think we can simply trigger the signal:
> ```
>    Component.onCompleted: amount.textChanged()
> ```

Thanks. That looks nicer and seems to work. I force-pushed to that.

accumulator:
> but then that must have already been the case even before this PR, mustn't it?

Yes, but this only leads to `FiatField` to stay empty in case the `BtcField.text` is pre-set, which is not a functional problem (just an UI inconsistency)

PR #9978, opened 2025-06-23
fix: sort qml swap provider list by PoW

### LINKED ISSUE #9976: qml: swap providers are not sorted by PoW amount
### Description

The swap provider list in qml is not ordered by PoW amount. In Qt servers are sorted by PoW amount to show the providers with the highest PoW amount in their announcement first. See `SwapOffer.pow_bits`.

f321x:
This change orders the list of swapproviders in the qml gui by the `pow_bits` value in their announcement by inserting the offers at the correct index in the list instead of just appending them.
Fixes #9976

### DIFF
--- a/electrum/gui/qml/qeswaphelper.py
+++ b/electrum/gui/qml/qeswaphelper.py
@@ -1,4 +1,5 @@
 import asyncio
+import bisect
 from enum import IntEnum
 from typing import Union, Optional, TYPE_CHECKING, Sequence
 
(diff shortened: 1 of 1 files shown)


electrum/gui/qml/qeswaphelper.py
```diff
@@ -105,7 +107,13 @@ def updateModel(self, items: Sequence['SwapOffer']):
         if offers:
             self.beginInsertRows(QModelIndex(), len(self._services), len(self._services) + len(offers) - 1)
```
SomberNight: https://doc.qt.io/qt-6/qabstractitemmodel.html#beginInsertRows
`void QAbstractItemModel::beginInsertRows(const QModelIndex &parent, int first, int last)`

the `first` and `last` args look suspicious

SomberNight: Perhaps you could call `begin...` and `end...` inside the for loop, each time for a single index, similar to how it's done in the "remove offers from model" loop above

f321x: Yeah right this doesn't seem correct according to the qt docs, thanks for catching. I changed it in https://github.com/spesmilo/electrum/pull/9978/commits/f1837ef58d9b257024b6531ef93c19cce4143a70 as you described.

SomberNight:
Looks good. Thx!

PR #9983, opened 2025-06-24
android: replace qr code scanning library

### LINKED ISSUE #9982: android: bad qr code scanner performance
### Description

The qr code scanner we use on android is so weak that it becomes unusable in certain situations (scanning qr from bright screen in dark environment) and leads to unpleasant experiences like having to use another qr scanner and then pasting the invoice in Electrum while having a queue of other waiting customers behind one.
As the qr scanner is a key feature of a mobile wallet this should be improved imo.
On android we currently use a very old qr scanning library which is essentially unmaintained since 2017 (https://github.com/dm77/barcodescanner). Replacing it with a modern library like the [Google ML Kit QR scanner](https://developers.google.com/ml-kit/vision/barcode-scanning/) or similar will probably improve the reliability.

f321x:
Replaces the unmaintained and unreliable
`me.dm7.barcodescanner:zxing:1.9.8` qr code scanning library used only in the android build with the `com.github.markusfisch:BarcodeScannerView:1.6.0` (https://github.com/markusfisch/BarcodeScannerView) library which seems actively maintained.
The `BarcodeScannerView` library is super fast and makes scanning qr codes fun again :)

Tested on Pixel 9 (Android 15),  x86_64 in Android Studio Device Emulator (Pixel 9 Pro XL, Android 16) and arm v7 on Galaxy Tab A (Android 8.1)

Fixes #9982


contrib/android/fetch_barcode_scanner.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ ! -d "$CONTRIB_ANDROID/aars" ]; then
+    mkdir -p "$CONTRIB_ANDROID/aars"
+fi
+
+if [ ! -f "$CONTRIB_ANDROID/aars/BarcodeScannerView.aar" ]; then
+    info "Fetching BarcodeScannerView..."
+    curl -L "${DOWNLOAD_REPOSITORY_ROOT}/BarcodeScannerView/${BARCODE_SCANNER_VIEW_VERSION}/BarcodeScannerView-${BARCODE_SCANNER_VIEW_VERSION}.aar" \
+        -o "$CONTRIB_ANDROID/aars/BarcodeScannerView.aar"
+fi
+echo "$BARCODE_SCANNER_VIEW_AAR_SHA256  $CONTRIB_ANDROID/aars/BarcodeScannerView.aar" | sha256sum -c -
```
SomberNight: Could this be replaced with `download_if_not_exist` and `verify_hash` calls?
See e.g.
https://github.com/spesmilo/electrum/blob/26e6c0900ba4f3bc10a25e0df814141fe937cfef/contrib/build-linux/appimage/make_appimage.sh#L46-L47

contrib/android/fetch_barcode_scanner.sh
```diff
@@ -0,0 +1,56 @@
...
+CONTRIB_ANDROID="$(dirname "$(readlink -e "$0")")"
+CONTRIB="$CONTRIB_ANDROID"/..
+
+. "$CONTRIB"/build_tools_util.sh
+
+# check if contrib/android/aars exists and contains BarcodeScannerView.aar, CameraView.aar and zxing-cpp.aar already,
+# if not, create and fetch them
+if [ ! -d "$CONTRIB_ANDROID/aars" ]; then
+    mkdir -p "$CONTRIB_ANDROID/aars"
+fi
```
SomberNight: Let's create a `.cache/` folder (or perhaps `build/`), and either use that directly or put this new folder inside that, e.g.: `contrib/android/.cache/aars`.
Then list `contrib/android/.cache` in the top-level `.gitignore`.
See how other build scripts do this:

https://github.com/spesmilo/electrum/blob/26e6c0900ba4f3bc10a25e0df814141fe937cfef/contrib/osx/make_osx.sh#L21
https://github.com/spesmilo/electrum/blob/26e6c0900ba4f3bc10a25e0df814141fe937cfef/contrib/build-linux/appimage/make_appimage.sh#L11

https://github.com/spesmilo/electrum/blob/26e6c0900ba4f3bc10a25e0df814141fe937cfef/.gitignore#L31-L40

contrib/android/fetch_barcode_scanner.sh
f321x: yes, that seems cleaner. Changed in https://github.com/spesmilo/electrum/pull/9983/commits/5ae2deb7048fec246cde7e97ac4cf640dbf7d0e9

contrib/android/fetch_barcode_scanner.sh
f321x: Ok, made it put the files in `android/.cache/aars/.` in https://github.com/spesmilo/electrum/pull/9983/commits/5ae2deb7048fec246cde7e97ac4cf640dbf7d0e9

contrib/android/fetch_barcode_scanner.sh
```diff
@@ -0,0 +1,50 @@
...
+
+info "Fetching CameraView..."
+download_if_not_exist "$CACHEDIR/aars/CameraView.aar" \
+    "${DOWNLOAD_REPOSITORY_ROOT}/CameraView/${CAMERA_VIEW_VERSION}/CameraView-${CAMERA_VIEW_VERSION}.aar"
+verify_hash "$CACHEDIR/aars/CameraView.aar" "$CAMERA_VIEW_AAR_SHA256"
+
+info "Fetching zxing-cpp..."
+download_if_not_exist "$CACHEDIR/aars/zxing-cpp.aar" \
+    "${DOWNLOAD_REPOSITORY_ROOT}/zxing-cpp/${ZXING_CPP_VERSION}/zxing-cpp-${ZXING_CPP_VERSION}.aar"
+verify_hash "$CACHEDIR/aars/zxing-cpp.aar" "$ZXING_CPP_AAR_SHA256"
```
SomberNight: Ok, so `BarcodeScannerView.aar` and `CameraView.aar` are ~10 KB, and contain platform-agnostic JVM bytecode.
`zxing-cpp.aar` is 3 MB, and contains some JVM bytecode and also contain compiled C++ shared libs for x86, x86_64, armv8, and armv7.

Would it be realistic for us to build these as part of the build process?

f321x: Ok yes, so gradlew accepts a `-Pandroid.injected.build.abi="$TARGET_ARCH"` argument, allowing zxing-cpp to be built for specific targets. This results in a binary size reduction of ~2mb (as expected).

So i wrote a script which builds all three libs from source, allowing us to build arch specific and removing the dependency on this jitpack.io service. The builds seem reproducible between two machines for me.
See https://github.com/spesmilo/electrum/pull/9983/commits/217221728ebe5c9a8cfd4379d7e08d874630bf3a

SomberNight: Very nice! I will have a proper look a bit later, but this looks great!

> This results in a binary size reduction of ~2mb (as expected).

Yeah, I don't care that much about that, but it's not bad ofc :)  The point of building it is to reduce the trust in that website and to some degree in the maintainer.
The reason I was listing sizes in my previous comment was not really about trying to minimise the size of the apk, but rather "there is lots of opportunity to hide weird shit in a 3 MB .so".

contrib/build_tools_util.sh
```diff
@@ -50,6 +50,28 @@ function download_if_not_exist() {
...
+    
+    if [ -z "$repo_url" ] || [ -z "$commit_hash" ] || [ -z "$repo_dir" ]; then
+        fail "clone_or_update_repo: invalid arguments: repo_url='$repo_url', commit_hash='$commit_hash', repo_dir='$repo_dir'"
+    fi
+
+    if [ -d "$repo_dir" ]; then
+        info "Repository $repo_url exists in $repo_dir, updating..."
+        git -C "$repo_dir" clean -ffxd >/dev/null 2>&1 || fail "Failed to clean repository $repo_dir"
+        git -C "$repo_dir" fetch --all >/dev/null 2>&1 || fail "Failed to fetch from repository"
+        git -C "$repo_dir" reset --hard "$commit_hash" >/dev/null 2>&1 || fail "Failed to reset to commit $commit_hash"
```
SomberNight: I have this OCD tick, to always add a `^{commit}` suffix.
see https://github.com/spesmilo/electrum/commit/f5f3394552c4efa46b71e13d1ad3ab171936bd88
```suggestion
        git -C "$repo_dir" reset --hard "$commit_hash^{commit}" >/dev/null 2>&1 || fail "Failed to reset to commit $commit_hash"
```

contrib/build_tools_util.sh
```diff
@@ -50,6 +50,28 @@ function download_if_not_exist() {
...
+    else
+        info "Cloning repository: $repo_url to $repo_dir"
+        git clone "$repo_url" "$repo_dir" >/dev/null 2>&1 || fail "Failed to clone repository $repo_url"
+        git -C "$repo_dir" checkout "$commit_hash" >/dev/null 2>&1 || fail "Failed to checkout commit $commit_hash"
```
SomberNight:
```suggestion
        git -C "$repo_dir" checkout "$commit_hash^{commit}" >/dev/null 2>&1 || fail "Failed to checkout commit $commit_hash"
```

contrib/build_tools_util.sh
f321x: Good to know👀 Changed it in https://github.com/spesmilo/electrum/pull/9983/commits/10b75ebed7e8d0594ded4709cc40e8c840fce6a3

contrib/build_tools_util.sh
f321x: Changed in https://github.com/spesmilo/electrum/pull/9983/commits/10b75ebed7e8d0594ded4709cc40e8c840fce6a3

SomberNight (approved):
Very nice. Great work!

The difference in terms of being able to scan qr codes is huge. This library seems to insta-scan ~max size qr codes at mediocre lighting that were practically ~impossible to scan using the old approach. Thanks a lot for looking into this.

The library looks decent. Looks to be well maintained, especially when also looking at https://github.com/markusfisch/BinaryEye. Great.

I built the full apk for one arch twice and it still reproduces.

contrib/android/make_barcode_scanner.sh
```diff
@@ -0,0 +1,144 @@
...
+
+    CAMERA_AAR_SOURCE="$CAMERA_VIEW_DIR/cameraview/build/outputs/aar/cameraview-release.aar"
+    CAMERA_AAR_DEST_GENERIC="$CACHEDIR/aars/CameraView.aar"
+    CAMERA_AAR_DEST_SPECIFIC="$CACHEDIR/aars/CameraView-$CAMERA_VIEW_BUILD_ID.aar"
+    if [ ! -f "$CAMERA_AAR_SOURCE" ]; then
+        fail "CameraView AAR not found at $CAMERA_AAR_SOURCE"
+    fi
+    cp "$CAMERA_AAR_SOURCE" "$CAMERA_AAR_DEST_GENERIC"
+    cp "$CAMERA_AAR_SOURCE" "$CAMERA_AAR_DEST_SPECIFIC"
+    info "CameraView AAR copied to $CAMERA_AAR_DEST"
```
SomberNight, line 113: is this meant to be `$CAMERA_AAR_DEST_GENERIC`?

contrib/android/make_barcode_scanner.sh
```diff
@@ -0,0 +1,144 @@
...
+
+    BARCODE_AAR_SOURCE="$BARCODE_SCANNER_VIEW_DIR/barcodescannerview/build/outputs/aar/barcodescannerview-release.aar"
+    BARCODE_AAR_DEST_GENERIC="$CACHEDIR/aars/BarcodeScannerView.aar"
+    BARCODE_AAR_DEST_SPECIFIC="$CACHEDIR/aars/BarcodeScannerView-$BARCODE_SCANNER_VIEW_BUILD_ID.aar"
+    if [ ! -f "$BARCODE_AAR_SOURCE" ]; then
+        fail "BarcodeScannerView AAR not found at $BARCODE_AAR_SOURCE"
+    fi
+    cp "$BARCODE_AAR_SOURCE" "$BARCODE_AAR_DEST_GENERIC"
+    cp "$BARCODE_AAR_SOURCE" "$BARCODE_AAR_DEST_SPECIFIC"
+    info "BarcodeScannerView AAR copied to $BARCODE_AAR_DEST"
```
SomberNight, line 140: is this meant to be `$BARCODE_AAR_DEST_GENERIC`?

contrib/android/make_barcode_scanner.sh
f321x: Yes, that must be a stale variable from a previous change. Fixed in https://github.com/spesmilo/electrum/pull/10066

contrib/android/make_barcode_scanner.sh
f321x: Yes, see https://github.com/spesmilo/electrum/pull/10066

PR #9992, opened 2025-06-27
qml: let user finalize forward swap onchain tx before initiating swap

accumulator:
Let users finalize the onchain tx when doing forward swaps.

An additional `deadline` parameter is added to `QETxFinalizer` that if set, limits fee policy type selection to ETA, and guard against fee selection that's too low priority.

### DIFF
--- a/electrum/gui/qml/components/SwapDialog.qml
+++ b/electrum/gui/qml/components/SwapDialog.qml
@@ -290,7 +290,19 @@ ElDialog {
                 enabled: swaphelper.valid && (swaphelper.state == SwapHelper.ServiceReady || swaphelper.state == SwapHelper.Failed)
 
                 onClicked: {
-                    swaphelper.executeSwap()
+                    if (swaphelper.isReverse) {
+                        swaphelper.executeSwap()
+                    } else {
+                        swaphelper.prepNormalSwap()
+                        var dialog = forwardSwapTxDialog.createObject(app, {
+                            finalizer: swaphelper.finalizer,
+                            satoshis: swaphelper.finalizer.amount
+                        })
+                        dialog.accepted.connect(function() {
+                            swaphelper.executeSwap()
+                        })
+                        dialog.open()
+                    }
                 }
             }
             FlatButton {
@@ -325,6 +337,15 @@ ElDialog {
                 })
                 dialog.open()
             }
+        }
+    }
+
+    Component {
+        id: forwardSwapTxDialog
+        ConfirmTxDialog {
+            amountLabelText: qsTr('Amount to swap')
+            sendButtonText: qsTr('Swap')
+            finalizer: swaphelper.finalizer
         }
     }
 
--- a/electrum/gui/qml/components/SwapDialog.qml
+++ b/electrum/gui/qml/components/SwapDialog.qml
@@ -293,7 +293,7 @@ ElDialog {
                     if (swaphelper.isReverse) {
                         swaphelper.executeSwap()
                     } else {
-                        swaphelper.prepNormalSwap()
+                        swaphelper.prepareNormalSwap()
                         var dialog = forwardSwapTxDialog.createObject(app, {
                             finalizer: swaphelper.finalizer,
                             satoshis: swaphelper.finalizer.amount
--- a/electrum/gui/qml/components/SwapDialog.qml
+++ b/electrum/gui/qml/components/SwapDialog.qml
@@ -290,7 +290,19 @@ ElDialog {
                 enabled: swaphelper.valid && (swaphelper.state == SwapHelper.ServiceReady || swaphelper.state == SwapHelper.Failed)
 
                 onClicked: {
-                    swaphelper.executeSwap()
+                    if (swaphelper.isReverse) {
+                        swaphelper.executeSwap()
+                    } else {
+                        swaphelper.prepNormalSwap()
+                        var dialog = forwardSwapTxDialog.createObject(app, {
+                            finalizer: swaphelper.finalizer,
+                            satoshis: swaphelper.finalizer.amount
+                        })
+                        dialog.accepted.connect(function() {
+                            swaphelper.executeSwap()
+                        })
+                        dialog.open()
+                    }
                 }
             }
             FlatButton {
@@ -325,6 +337,14 @@ ElDialog {
                 })
                 dialog.open()
             }
+        }
+    }
+
+    Component {
+        id: forwardSwapTxDialog
+        ConfirmTxDialog {
+            amountLabelText: qsTr('Amount to swap')
+            sendButtonText: qsTr('Swap')
         }
     }
 
--- a/electrum/gui/qml/qeswaphelper.py
+++ b/electrum/gui/qml/qeswaphelper.py
@@ -13,7 +13,9 @@ from electrum.logging import get_logger
 from electrum.transaction import PartialTxOutput, PartialTransaction
 from electrum.util import (NotEnoughFunds, NoDynamicFeeEstimates, profiler, get_asyncio_loop, age,
                            wait_for2, send_exception_to_crash_reporter)
-from electrum.submarine_swaps import NostrTransport, SwapServerTransport, pubkey_to_rgb_color
+from electrum.submarine_swaps import (
+    NostrTransport, SwapServerTransport, pubkey_to_rgb_color, LOCKTIME_DELTA_REFUND, LOCKTIME_DELTA_REFUND_BUFFER
+)
 from electrum.fee_policy import FeePolicy
 
 from electrum.gui.common_qt.util import QtEventListener, qt_event_listener
@@ -705,6 +707,7 @@ class QESwapHelper(AuthMixin, QObject, QtEventListener):
         self._finalizer.canRbf = False
         self._finalizer.amount = QEAmount(amount_sat=self._send_amount)
         self._finalizer.wallet = self._wallet
+        self._finalizer.deadline = LOCKTIME_DELTA_REFUND - LOCKTIME_DELTA_REFUND_BUFFER  # 10-block buffer before refund deadline
         self.finalizerChanged.emit()
 
     def do_reverse_swap(self, lightning_amount, onchain_amount):
--- a/electrum/gui/qml/qeswaphelper.py
+++ b/electrum/gui/qml/qeswaphelper.py
@@ -7,6 +7,7 @@ from PyQt6.QtCore import (pyqtProperty, pyqtSignal, pyqtSlot, QObject, QTimer, p
                           QModelIndex, QVariant)
 from PyQt6.QtGui import QColor
 
+from electrum.gui.qml.qetxfinalizer import QETxFinalizer
 from electrum.i18n import _
 from electrum.bitcoin import DummyAddress
 from electrum.logging import get_logger
@@ -27,9 +28,6 @@ from .qewallet import QEWallet
 if TYPE_CHECKING:
     import concurrent.futures
     from electrum.submarine_swaps import SwapOffer
-
-
-class InvalidSwapParameters(Exception): pass
 
 
 class QESwapServerNPubListModel(QAbstractListModel):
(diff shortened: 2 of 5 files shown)


f321x:
Tested, works fine for me. Maybe the minimal allowed tx fee should be limited in the tx finalizer to an ETA limit < `LOCKTIME_DELTA_REFUND` (+ buffer), otherwise it seems easy to lose money by cheaping out on the tx fees. 
E.g. the user selects 1 sat/vb for a non urgent swap in the hope it will confirm in the next weeks (or months) during a large mempool, not knowing there are time limitations to swaps. Then he will have to broadcast a refund tx later and waste fees for nothing (ignoring the case of not opening the wallet at all, as we show a warning for this). 

Also noticed these two qml logs while testing:
After confirming the swap, while receiving the htlcs:
`331.59 | W | gui.qml.qeapp | file:///home/user/code/electrum-fork/electrum/gui/qml/components/Channels.qml:124:13: QML FlatButton: Binding loop detected for property "enabled"`

When clicking on the input/output dropdown in the tx finalization dialog:
`147.06 | W | gui.qml.qeapp | Qt Quick Layouts: Detected recursive rearrange. Aborting after two iterations.`
This seems to be an issue with the this dialog, also happens when i do a regular onchain tx.

I can also open separate issues if they are out of scope of this PR.

accumulator:
> Tested, works fine for me. Maybe the minimal allowed tx fee should be limited in the tx finalizer to an ETA limit < `LOCKTIME_DELTA_REFUND` (+ buffer), otherwise it seems easy to lose money by cheaping out on the tx fees. E.g. the user selects 1 sat/vb for a non urgent swap in the hope it will confirm in the next weeks (or months) during a large mempool, not knowing there are time limitations to swaps. Then he will have to broadcast a refund tx later and waste fees for nothing (ignoring the case of not opening the wallet at all, as we show a warning for this).

Yes, good idea. I'm leaning towards disabling static fees in these cases, only allowing finalizing when dynamic fees (ETA, mempool) are available. What do you think @f321x?

> Also noticed these two qml logs while testing: After confirming the swap, while receiving the htlcs: `331.59 | W | gui.qml.qeapp | file:///home/user/code/electrum-fork/electrum/gui/qml/components/Channels.qml:124:13: QML FlatButton: Binding loop detected for property "enabled"`
> 
> When clicking on the input/output dropdown in the tx finalization dialog: `147.06 | W | gui.qml.qeapp | Qt Quick Layouts: Detected recursive rearrange. Aborting after two iterations.` This seems to be an issue with the this dialog, also happens when i do a regular onchain tx.
> 
> I can also open separate issues if they are out of scope of this PR.

There are a few places where these frontend issues are still logged. For the binding loops I don't really see how they are getting triggered. They seem to be harmless. The layout issues are sometimes logged when anchoring an Item outside a layout to items within a Layout. Also seems mostly harmless.

f321x:
@accumulator i think only allowing ETA and capping the max allowed ETA would be the safest way. Static rates and mempool seem easy to get wrong for the funding tx. It seems nice to have the flexibility to save some fees for non-urgent swaps but the risk of a failing (refunded) swap by setting too low fees should be minimal imo.

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -423,6 +439,17 @@ def update(self):
             self.validChanged.emit()
             return
 
+        if self._deadline:
+            deadline_fee_policy = FeePolicy(f'eta:{self._deadline}')
+            deadline_fee = deadline_fee_policy.fee_per_kb(self._wallet.wallet.network)
```
f321x: We have these targets available `FEE_ETA_TARGETS = [1008, 144, 25, 10, 5, 2, 1]` and `deadline_fee_policy.fee_per_kb` will use the next closest available eta to `self._deadline` through `closest_index()`. For the 60 blocks deadline in this PR it would be the 25 block eta which would be fine. But what if the deadline is e.g. 85, then the allowed ETA would be <= 144 for the deadline before a warning is shown which could be unsafe as the allowed eta setting would be 59 blocks larger than the intended deadline.

A direct comparison of the ETA values would not have this issue:
```if self._deadline < self._fee_policy.value:```

Comment (f321x) [reply electrum/gui/qml/qetxfinalizer.py]: Also the limit only gets activated when moving the slider. When the eta is already at e.g. 1008 blocks when the finalizer shows up it is possible to finalize by just clicking "Swap" without moving the slider.

Comment (accumulator) [reply electrum/gui/qml/qetxfinalizer.py]: > We have these targets available `FEE_ETA_TARGETS = [1008, 144, 25, 10, 5, 2, 1]` and `deadline_fee_policy.fee_per_kb` will use the next closest available eta to `self._deadline` through `closest_index()`. For the 60 blocks deadline in this PR it would be the 25 block eta which would be fine. But what if the deadline is e.g. 85, then the allowed ETA would be <= 144 for the deadline before a warning is shown which could be unsafe as the allowed eta setting would be 59 blocks larger than the intended deadline.

Ah yes, sharp. Assumptions are deadly. This behavior of backend `FeePolicy` is dangerous in the context of deadlines :) But why is a frontend concept of slider position used in a backend object? This doesn't make sense. Also the fact that any ETA value gets mapped to a closest index in a very coarse limited set of allowed values should be clearly documented.

> A direct comparison of the ETA values would not have this issue: `if self._deadline < self._fee_policy.value:`

Yes, that seems clear now :)

Comment (accumulator) [reply electrum/gui/qml/qetxfinalizer.py]: > Also the limit only gets activated when moving the slider. When the eta is already at e.g. 1008 blocks when the finalizer shows up it is possible to finalize by just clicking "Swap" without moving the slider.

Yes this depends on the order in which QML populates `QETXFinalizer`, if `wallet` is set first, it calls `update()` without a deadline set.

Comment (f321x) [reply electrum/gui/qml/qetxfinalizer.py]: Tested https://github.com/spesmilo/electrum/pull/9992/commits/72b16ebaaa4665b4c0ff371bb748c6f3616e0832, works well now

**Comment by accumulator** [comment]
rebased on master with two minor nits resolved

File: electrum/gui/qml/qetxfinalizer.py [hunk 2]
```diff
@@ -462,8 +463,24 @@ def canRbf(self, canRbf):
...
+
+    @deadline.setter
+    def deadline(self, relative_num_blocks: int) -> None:
+        """if set, limits the finalizer to ETA fee policies that meet the deadline.
+           deadline is in relative blocks"""
+        if self._deadline != relative_num_blocks:
+            self._deadline = relative_num_blocks
+            self.deadlineChanged.emit()
+            if self._deadline > 0:
+                self.method = FeeSlider.FSMethod.ETA
```
f321x, line 479: Assigning to `self.method` will persist the `ETA` fee estimation method to the config, even though the user might have set mempool etc., so they have to change it back on the next regular onchain tx.
Maybe `_deadline` could become an attribute of `FeeSlider`, then `FeeSlider.save_config()` could be gated on `if not self._deadline` to skip saving if a deadline is set?

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -462,8 +463,24 @@ def canRbf(self, canRbf):
             self.canRbfChanged.emit()
         self.rbf = self._canRbf  # if we can RbF, we do RbF
 
+    deadlineChanged = pyqtSignal()
+    @pyqtProperty(int, notify=deadlineChanged)
+    def deadline(self):
+        return self._deadline
+
+    @deadline.setter
```
f321x, line 471: It might also be nice to initialize the fee slider with an allowed default value if a deadline gets set and the current ETA is out of the allowed range. Currently if the user has e.g. 1 week ETA set from a previous tx and enters the forward swap finalizer, they immediately get shown a warning instead of a usable default ETA.

electrum/gui/qml/qeswaphelper.py
```diff
@@ -683,32 +692,23 @@ async def swap_task():
...
-        return tx
+    @pyqtSlot()
+    def prepareNormalSwap(self):
+        """prepare for normal swap by instantiating a finalizer for the requested amount.
+           self._finalized_tx will contain the finalized tx using the fees set by the user"""
+        def mktx(amt, fee_policy: FeePolicy):
+            try:
+                self._finalized_tx = self._create_swap_tx(amt, fee_policy)
+            except (NotEnoughFunds, NoDynamicFeeEstimates):
+                self._finalized_tx = None
```
f321x, line 703: `QETxFinalizer.make_tx()` relies on the `mktx` returning either a transaction or raising an exception, returning `self._finalized_tx = None` would throw an `AttributeError` in `make_tx().
Maybe this would work?

```suggestion
                self._finalized_tx = None
                raise
```

electrum/gui/qml/components/SwapDialog.qml
f321x, line 147: The mining fee shown here is now not accurate anymore as the user will later set its own fee. 
Maybe adding a `~` in front of the amount (for forward swaps) would indicate the user that this is just a rough estimate but not what they will pay exactly? In the end it might be useful to see an estimated value, even if we don't know what the user will set later, to judge the current fee environment before continuing.

Image

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -508,6 +525,14 @@ def update(self):
             self.validChanged.emit()
             return
 
+        if self._deadline:
+            if self._deadline < self._fee_policy.value:
```
f321x, line 529: Its a bit confusing that the target feerate says "Within 25 blocks" when moving the slider to the left but the warning says that the target exceeds 60 blocks deadline. 

Image

This works for me:
```diff
diff --git a/electrum/gui/qml/qetxfinalizer.py b/electrum/gui/qml/qetxfinalizer.py
index 0f9c84f0a..cfe402002 100644
--- a/electrum/gui/qml/qetxfinalizer.py
+++ b/electrum/gui/qml/qetxfinalizer.py
@@ -504,6 +504,15 @@ class QETxFinalizer(TxFeeSlider):
             self._logger.debug('wallet not set, ignoring update()')
             return
 
+        if self._deadline and self._deadline < self._fee_policy.value:
+            # reject the policy before calling make_tx so a invalid tx is not retained (e.g. in swap helper)
+            self._logger.info(f"current fee '{str(self._fee_policy)}' below deadline {str(self._deadline)}")
+            self.update_target()
+            self.warning = _("Current fee doesn't meet deadline of {} blocks").format(self._deadline)
+            self._valid = False
+            self.validChanged.emit()
+            return
+
         try:
             # make unsigned transaction
             amount = '!' if self._amount.isMax else self._amount.satsInt
@@ -525,14 +534,6 @@ class QETxFinalizer(TxFeeSlider):
             self.validChanged.emit()
             return
 
-        if self._deadline:
-            if self._deadline < self._fee_policy.value:
-                self._logger.info(f"current fee '{str(self._fee_policy)}' below deadline {str(self._deadline)}")
-                self.warning = _("Current fee doesn't meet deadline of {} blocks").format(self._deadline)
-                self._valid = False
-                self.validChanged.emit()
-                return
-
         self._tx = tx
 
         amount = self._amount.satsInt if not self._amount.isMax else tx.output_value()

```

electrum/gui/qml/components/SwapDialog.qml
```diff
@@ -328,6 +340,14 @@ ElDialog {
         }
     }
 
+    Component {
+        id: forwardSwapTxDialog
+        ConfirmTxDialog {
+            amountLabelText: qsTr('Amount to swap')
+            sendButtonText: qsTr('Swap')
+        }
```
f321x, line 348: https://github.com/spesmilo/electrum/blob/8b177394b99fcbbd3685deac2cf08d006843c409/electrum/gui/qml/components/WalletMainView.qml#L708-L711

Could we call `destroy()` here `onClosed` as there is no finalizer that lives in the `ConfirmTxDialog`? (maybe same for `confirmOpenChannelDialog` too then).

```suggestion
        ConfirmTxDialog {
            amountLabelText: qsTr('Amount to swap')
            sendButtonText: qsTr('Swap')
            onClosed: destroy()
        }
```

Issue #9980, opened 2025-06-24
Wallet failed to claim "Our CTX to Local" transaction


nc50lc:
### Description

After a failed Reverse Swap attempt, there had been a "_Pending_" payment in my channel.  (_shown as "(-0.x BTC)"_)
Even after one successful swap using a different Nostr Swap Provider and one lightning payment, it stayed pending in my channel's balance
So I tried to request the remote node to force-close the channel and it managed to send the 0.00074569 BTC to my on-chain funds.
The pending 0.00019676 BTC was in a week timelock.
But after waiting for a week, it turned into a normal local transaction and my client failed to sign the claim transaction to broadcast.

Swap Provider used in my failed Reverse Swap Attempt: e58c6cad84311c20f0c3e82345bd9...

Channel's Remote node is: Electrum Trampoline
Channel ID (Closed): a0a0f26147022370183eda195c7b2d56b6f026bb091414af23faeb129e7c7f92
Short Channel ID: 896833x3338x0

Local transaction PSBT:
[...]

PR #9994, opened 2025-06-28
lnwatcher: keep_watching until fees go down

SomberNight:
- adds a regtest test to reproduce https://github.com/spesmilo/electrum/issues/9980
- fixes https://github.com/spesmilo/electrum/issues/9980

(see individual commits)

### DIFF
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -99,6 +99,7 @@ class LNWatcher(Logger, EventListener):
         if not self.adb.is_mine(address):
             return
         # inspect_tx_candidate might have added new addresses, in which case we return early
+        # note: maybe we should wait until adb.is_up_to_date... (?)
         funding_txid = funding_outpoint.split(':')[0]
         funding_height = self.adb.get_tx_height(funding_txid)
         closing_txid = self.adb.get_spender(funding_outpoint)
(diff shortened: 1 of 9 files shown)


electrum/txbatcher.py
```diff
@@ -263,20 +264,25 @@ def add_payment_output(self, output: 'PartialTxOutput') -> None:
         self.batch_payments.append(output)
 
     def is_dust(self, sweep_info: SweepInfo) -> bool:
+        """Can raise BelowDustLimit or NoDynamicFeeEstimates."""
```
SomberNight: fixme: can only raise NoDynamicFeeEstimates

electrum/lnwatcher.py
```diff
@@ -256,10 +265,18 @@ def maybe_add_pending_forceclose(
...
+        (we are waiting for ctx to be confirmed and there are received htlcs)
+        """
+        if is_local_ctx and sweep_info.name == 'received-htlc':
+            cltv = sweep_info.cltv_abs
+            assert cltv is not None, f"missing cltv for {sweep_info}"
+            if self.adb.get_local_height() > cltv + REDEEM_AFTER_DOUBLE_SPENT_DELAY:
+                # We had plenty of time to sweep. The remote also had time to time out the htlc.
+                # Maybe its value has been ~dust at current and past fee levels (every time we checked).
+                # We should not keep warning the user forever.
+                return
```
SomberNight, line 279: note: this is a functional change here.
Previously we only warned the user if the htlc was higher-than-dust value at *current* fee levels. Now that information is not readily available, so instead we always warn if there is an unspent received htlc. However, so that we don't keep warning forever, we now consider the cltv of the htlc. If it's past the cltv, the remote party could time it out at any point. Past the cltv, it's questionable if it makes sense to warn the user not to close the wallet.

PR #9995, opened 2025-06-28
chore: fix some minor issues in the comments

cloudclaim:
fix some minor issues in the comments


electrum/plugins/jade/jadepy/jade.py
```diff
@@ -1027,7 +1027,7 @@ def get_registered_descriptor(self, descriptor_name):
         dict
             Description of registered descriptor wallet identified by registration name.
             Contains keys:
-                descriptor_name - str, name of descritpor registration
+                descriptor_name - str, name of descriptor registration
```
SomberNight: This should be fixed upstream.
See https://github.com/spesmilo/electrum/blob/master/electrum/plugins/jade/jadepy/README.md

cloudclaim: Thanks. Reverted.

PR #9999, opened 2025-06-30
history: fix displayed STATUS sort order

### LINKED ISSUE #9969: History sorts of Description/Amount/Balance are backwards
### Description

Image
Image
Image
Image
Image
Image
Image
Image

Date sorts normal, everything else sorts backwards. fdaafd5abf197be0e7fcbc34132357a4919ccaad on bookworm. I can't repro on 4.1.5 on Win32.

nabijaczleweli:
Draft because on top of #9968

Closes #9969

### DIFF
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -81,7 +81,7 @@ class HistorySortModel(QSortFilterProxyModel):
         col = index.column()
         if col == HistoryColumns.STATUS:
             # respect sort order of self.transactions (wallet.get_full_history)
-            return -index.row()
+            return index.row()
         else:
             node = index.internalPointer()
             return node.sort_keys[col]
@@ -507,7 +507,7 @@ class HistoryList(MyTreeView, AcceptFileDragDrop):
         self.period_combo.addItems([_('All'), _('Custom')])
         self.period_combo.activated.connect(self.on_combo)
         self.wallet = self.main_window.wallet  # type: Abstract_Wallet
-        self.sortByColumn(HistoryColumns.STATUS, Qt.SortOrder.AscendingOrder)
+        self.sortByColumn(HistoryColumns.STATUS, Qt.SortOrder.DescendingOrder)
         self.setRootIsDecorated(True)
         self.header().setStretchLastSection(False)
         for col in HistoryColumns:


nabijaczleweli:
all sorts correctly now

SomberNight:
Could you also test with the qdarkstyle "dark" theme? Preferences>Appearance>"Color theme">Dark

It looks to me the arrows point the other way there :D
Also, on Windows, on the non-dark theme, the arrow points the other way.
I am starting to wonder which way is the "right" way >.<  (but in any case I agree that before this PR the date column ("status)" is not consistent with the rest)

electrum/gui/qt/history_list.py
```diff
@@ -81,13 +81,13 @@ def data_for(self, index: QModelIndex):
...
             # respect sort order of self.transactions (wallet.get_full_history)
-            return -index.row()
+            return index.row()
         else:
             node = index.internalPointer()
             return node.sort_keys[col]
 
     def lessThan(self, source_left: QModelIndex, source_right: QModelIndex):
-        return self.data_for(source_left) < self.data_for(source_right)
+        return self.data_for(source_left) > self.data_for(source_right)
```
SomberNight: Are you sure this sign should be flipped?

from https://doc.qt.io/qt-6/qsortfilterproxymodel.html#lessThan :
> Returns true if the value of the item referred to by the given index source_left is less than the value of the item referred to by the given index source_right, otherwise returns false.

nabijaczleweli:
Yes... the icons being wrong would certainly explain why lessThan is correct with > instead of < :skull: I hadn't considered that

nabijaczleweli:
With `qdarkstyle` I see what I rather expected to see: with the patch with everything inverted, the sort order is backwards:




The correct order/icon matching corresponds geometrically: the big side goes on the side with the big values and vice versa. Thus, ▲ is small-to-big top-to-bottom and ▼ is big-to-small top-to-bottom.

So, the default theme's icons are wrong and the date column's sort order was inverted to compensate for the default sort order being small-to-big when in reality you want it to be (and it is) big-to-small.

nabijaczleweli:
I made the displayed sort order for STATUS consistent instead of double-inverting it as was done originally:






(the first in each style is the default).

AFAICT the default style is the system's/Qt's default style. These screen-shots were taken on bookworm, and the current electrum win32 exe doesn't exhibit this:



so that ought to be remedied upstream.

nabijaczleweli:
Yeah, I've confirmed the icons being wrong with a trivial Qt application.

nabijaczleweli:
https://bugs.debian.org/1108632

SomberNight (approved):
Looks good. Thx.

Issue #10007, opened 2025-07-04
qt: revealer plugin ValueError


f321x:
### Description

The revealer plugin throws the following exception when clicking `Create a new Revealer noise file` in the Appimage. It works fine when running from source. The exception is swallowed and there is no `CrashReporter` dialog.
```
QPainter::setRenderHint: Painter must be active to set rendering hints
QImage::scaleWidth: Image is a null image
 19.59 | E | plugins.revealer.qt.Plugin | 
Traceback (most recent call last):
  File "/tmp/.mount_electrjPl0bE/usr/lib/python3.12/site-packages/electrum/plugins/revealer/qt.py", line 195, in create_noise_file
    self.make_digital(self.d)
  File "/tmp/.mount_electrjPl0bE/usr/lib/python3.12/site-packages/electrum/plugins/revealer/qt.py", line 243, in make_digital
    self.bdone(dialog)
  File "/tmp/.mount_electrjPl0bE/usr/lib/python3.12/site-packages/electrum/plugins/revealer/qt.py", line 276, in bdone
    dialog.show_message(''.join([_("Digital Revealer ({}_{}) saved as PNG and PDF at:").format(version, code_id),
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: unexpected '{' in field name
```

Edit: Caused by broken translation, temporary workaround is to switch Language to English.

SomberNight:
> Edit: Caused by broken translation, temporary workaround is to switch Language to English.

We could try to come up with a more general and intellectually more satisfying solution.
Something along these lines might work, though I haven't tested, and ideally we should also add unit tests:

```diff
diff --git a/electrum/i18n.py b/electrum/i18n.py
index cc76e915ab..4768d2d0c7 100644
--- a/electrum/i18n.py
+++ b/electrum/i18n.py
@@ -23,6 +23,7 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import os
+import string
 from typing import Optional
 
 import gettext
@@ -44,6 +45,29 @@ def _get_null_translations():
 _language = _get_null_translations()
 
 
+def _ensure_translation_keeps_format_string_syntax_similar(translator):
+    """This checks that the source string (msg) is syntactically similar to the translated string.
[...]

PR #10011, opened 2025-07-05
i18n: syntax-check translations at runtime

SomberNight:
We often call str.format() on translated strings.
E.g. `_("time left: {} seconds").format(t1)`
If the translated string has a different format syntax, this can raise at runtime.

This PR adds some runtime checks that try to ensure the source string and the translated string have a similar format syntax. If the checks fail, `_()` will "reject" the translation by returning the source string.

fixes https://github.com/spesmilo/electrum/issues/10010
ref https://github.com/spesmilo/electrum/issues/10007#issue-3203378250

### DIFF
--- a/electrum/i18n.py
+++ b/electrum/i18n.py
@@ -22,7 +22,9 @@
 # ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import functools
 import os
+import string
 from typing import Optional
 
 import gettext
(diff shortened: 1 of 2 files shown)


accumulator:
Why use a decorator when the only func that uses it is `_()`?

electrum/i18n.py
```diff
@@ -44,6 +46,32 @@ def _get_null_translations():
...
+    """
+    sf = string.Formatter()
+    @functools.wraps(translator)
+    def safe_translator(msg: str, **kwargs):
+        translation = translator(msg, **kwargs)
+        parsed1 = list(sf.parse(msg))  # iterable of tuples (literal_text, field_name, format_spec, conversion)
+        try:
+            parsed2 = list(sf.parse(translation))
+        except ValueError:  # malformed format string in translation
+            return msg
```
f321x, line 62: Maybe a log like `_logger.warning(f"malformed translation string: {translation=}")` would be useful so the incorrect string can be searched and fixed on crowdin.

SomberNight: Ok, added in https://github.com/spesmilo/electrum/pull/10011/commits/d16c6250193fd2a189a62ff1360967118fced8df.

Note that this line will get logged every time `_()` is evaluated for the string. If it is in a hot path, we might end up spamming the log. I will do a separate follow-up for that. For now, decided to set it to INFO level, so that at least we can't spam the stderr of non `-v` users.

SomberNight:
> Why use a decorator when the only func that uses it is `_()`?

The new logic is quite long so it makes sense to put it into a separate function.
As for why I used a decorator instead of a normal function: there are two return paths in `_()` and the new logic should be applied in both. I did not want to refactor `_()` - it is not tested and its logic is already not trivial. Also, I think nesting the function calls (with or without a decorator) makes some sense here: there is an existing translator function, and the new logic is something higher-level which uses the translator and decides whether its output is good or garbage.
but yeah, it's subjective.

Issue #10012, opened 2025-07-07
TypeError during `gettransaction`


theborakompanioni:
### Description

Running and testing v4.6.0b1 here.
This scenario is part of an integration test which successfully runs with v4.5.8, but fails in v4.6.0b1.
Getting an error with `gettransaction` for txid `eadde85706765d801dfe9a4807f627754bdab7a2635841b699a76d8e167df8d9`:

```
~ $ electrum --regtest gettransaction eadde85706765d801dfe9a4807f627754bdab7a2635841b699a76d8e167df8d9
(inside daemon): Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/electrum/daemon.py", line 260, in handle
    response['result'] = await f(*params)
  File "/usr/local/lib/python3.10/site-packages/electrum/daemon.py", line 373, in run_cmdline
    result = await func(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/electrum/commands.py", line 172, in func_wrapper
    wallet = daemon.get_wallet(wallet_path)
  File "/usr/local/lib/python3.10/site-packages/electrum/daemon.py", line 527, in get_wallet
    wallet_key = self._wallet_key_from_path(path)
  File "/usr/local/lib/python3.10/site-packages/electrum/daemon.py", line 454, in _wallet_key_from_path
    path = standardize_path(path)
  File "/usr/local/lib/python3.10/site-packages/electrum/util.py", line 602, in standardize_path
    os.path.expanduser(
  File "/usr/local/lib/python3.10/posixpath.py", line 232, in expanduser
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

```
~ $ electrum --regtest onchain_history
[
    {
        "amount_sat": 5000000000,
        "bc_balance": "50.",
        "bc_value": "50.",
        "confirmations": 101,
        "date": "2025-07-07 13:42",
        "fee_sat": null,
        "group_id": null,
        "height": 2,
        "incoming": true,
        "label": "",
        "monotonic_timestamp": 1751258466,
        "timestamp": 1751258466,
        "txid": "251bad076b1f4cab110a1bdffddd57daa60f087f5bceff9433ecd228b3968ed9",
        "txpos_in_block": 0,
        "wanted_height": null
    },
    {
        "amount_sat": -200000,
        "bc_balance": "49.998",
        "bc_value": "-0.002",
        "confirmations": 0,
        "date": null,
[...]

PR #10013, opened 2025-07-07
commands: fix: only try to get wallet if wallet_path

f321x:
Only try to get wallet from daemon in the `command` decorator if the wallet_path is available to prevent raising `TypeError` when `daemon.get_wallet(path=None)` gets called.
Fixes https://github.com/spesmilo/electrum/issues/10012


SomberNight (approved):
Looks good. Thx.

PR #10015, opened 2025-07-07
logging: add "only_once=False" param to logger.info/warning/etc calls

SomberNight:
adds very basic form of opt-in rate-limiting of log spam

related https://github.com/spesmilo/electrum/pull/10011#discussion_r2190209204


ecdsa:
instead of overloading the `Logger` class, it is possible to use `addFilter/removeFilter`
see https://stackoverflow.com/questions/31953272/logging-print-message-only-once

SomberNight:
I saw `Logger.addFilter` but I wanted an API that allows opting-into this behaviour on a per-logline-call level.
addFilter is not a good fit for that.

PR #10016, opened 2025-07-09
wizard: qml: separate passphrase input from seed input

### LINKED ISSUE #10008: qml: "Extend seed with custom words" checkbox is too hidden in wizard
### Description

When trying to recover a wallet from seed on a Oneplus 6 Android 11 the passphrase extension (`Extend seed with custom words`) checkbox is not shown in the wizard. It is not affected by the font size accessibility settings, even with the smallest font size the checkbox is not visible, so it is not possible for the user to restore a wallet from seed with additional passphrase.

Edit: ok just noticed that it gets enabled after entering some words and then is hidden and becomes  visible by scrolling down. So recovery is possible but its quite unintuitive to find the checkbox. We should probably show it somewhere else so its not missed as easily.

Image

f321x:
* Separates the passphrase input from the seed input in the qml gui (similar to the Qt GUI)
* Only show the infotext in `WCHaveSeed` if seed format is not electrum, and remove a sentence from the bip39 infotext to make the screen less filled
* move the passphrase input flow from the Qt wizard to the Abstract Wizard so it is shared between Qt and QML

This intends to make the passphrase input more usable, especially on small phone screens, and prevent confusion like in https://github.com/spesmilo/electrum/issues/10008.
This will always show an additional passphrase step in the qml wizard, alternatively it would also be possible e.g. to re-add just a checkbox (but no input field) above the seed input field and only show the separate passphrase step if this checkbox has been enabled. However this would again fill the seed input step with more stuff than theoretically required. 

closes #10008


electrum/gui/qml/components/wizard/WCConfirmExt.qml
```diff
@@ -0,0 +1,85 @@
...
+    Flickable {
+        anchors.fill: parent
+        contentHeight: mainLayout.height
+        clip: true
+        interactive: height < contentHeight
+
+        GridLayout {
+            id: mainLayout
+            width: parent.width
+            columns: 2
```
accumulator: This can be simplified to a `ColumnLayout` since the TextField below is spanning over the full width anyway.

electrum/gui/qml/components/wizard/WCConfirmExt.qml
```diff
@@ -0,0 +1,85 @@
...
+                wrapMode: Text.Wrap
+                text: qsTr('Please enter your custom word(s) a second time:')
+            }
+
+            TextField {
+                id: customwordstext
+                Layout.fillWidth: true
+                Layout.columnSpan: 2
+                placeholderText: qsTr('Enter your custom word(s) here')
+                inputMethodHints: Qt.ImhNoPredictiveText
```
accumulator: This should be further restricted, see e.g. `WCHaveMasterKey.qml`

`inputMethodHints: Qt.ImhSensitiveData | Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase`

electrum/gui/qml/components/wizard/WCConfirmExt.qml
```diff
@@ -0,0 +1,85 @@
...
+        valid = false
+        validationTimer.restart()
+    }
+
+    Timer {
+        id: validationTimer
+        interval: 150
+        repeat: false
+        onTriggered: checkValid()
+    }
```
accumulator: `checkValid()` is low-cost, do we really need a pushback timer here?

electrum/gui/qml/components/wizard/WCCreateSeed.qml
```diff
@@ -9,13 +9,12 @@ import "../controls"
...
 
-    valid: seedtext.text != '' && extendcb.checked ? customwordstext.text != '' : true
+    valid: seedtext.text != ''
 
     function apply() {
         wizard_data['seed'] = seedtext.text
         wizard_data['seed_variant'] = 'electrum' // generated seed always electrum variant
-        wizard_data['seed_extend'] = extendcb.checked
-        wizard_data['seed_extra_words'] = extendcb.checked ? customwordstext.text : ''
+        wizard_data['seed_extend'] = true
```
accumulator: This is slightly different from the desktop client.

If we unconditionally go to the seed extension page, we don't strictly need to set it here. Might be useful to add a comment here that `seed_extend` is definitively set in the next `WCEnterExt` page.

electrum/gui/qml/components/wizard/WCEnterExt.qml
```diff
@@ -0,0 +1,120 @@
...
+    Flickable {
+        anchors.fill: parent
+        contentHeight: mainLayout.height
+        clip: true
+        interactive: height < contentHeight
+
+        GridLayout {
+            id: mainLayout
+            width: parent.width
+            columns: 2
```
accumulator: This is effectively a `ColumnLayout`

electrum/gui/qml/components/wizard/WCEnterExt.qml
```diff
@@ -0,0 +1,120 @@
...
+                onCheckedChanged: checkValid()
+            }
+
+            TextField {
+                id: customwordstext
+                visible: extendcb.checked
+                Layout.fillWidth: true
+                Layout.columnSpan: 2
+                placeholderText: qsTr('Enter your custom word(s)')
+                inputMethodHints: Qt.ImhNoPredictiveText
```
accumulator: see `inputMethodHints` comment above

electrum/gui/qml/components/wizard/WCEnterExt.qml
```diff
@@ -0,0 +1,120 @@
...
+                id: extendcb
+                Layout.columnSpan: 2
+                Layout.fillWidth: true
+                text: qsTr('Extend seed with custom words')
+                onCheckedChanged: checkValid()
+            }
+
+            TextField {
+                id: customwordstext
+                visible: extendcb.checked
```
accumulator: maybe change to `enabled: extendcb.checked`, as we have plenty of screen space, and this page is dedicated to the seed extension/passphrase now.

electrum/gui/qml/components/wizard/WCConfirmExt.qml
f321x: Ok, done in https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

electrum/gui/qml/components/wizard/WCConfirmExt.qml
f321x: Done here https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

electrum/gui/qml/components/wizard/WCConfirmExt.qml
f321x: Yeah its probably overkill, i already lowered the interval but lets just remove it. Removed in https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

electrum/gui/qml/components/wizard/WCEnterExt.qml
f321x: Changed in https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

electrum/gui/qml/components/wizard/WCEnterExt.qml
f321x: Changed this too in https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

electrum/gui/qml/components/wizard/WCEnterExt.qml
f321x: Yes good idea, that looks better, changed in https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

electrum/gui/qml/components/wizard/WCCreateSeed.qml
f321x: Yeah right its a bit different that in the the desktop we only get forwarded when the var is set true with the checkbox (which we want to remove from the seed creation/input page in qml as it takes too much space).

We need to set it unconditionally true so we always resolve to the passphrase input page next, the alternative would be to have a custom wizard flow again for qml, but i think this is a bit cleaner as its just a single line?

I have added a comment to the line in https://github.com/spesmilo/electrum/pull/10016/commits/11de760a04556e17d3d44f53a5cef80878cad247

accumulator:
LGTM

electrum/gui/qml/components/wizard/WCEnterExt.qml
```diff
@@ -0,0 +1,119 @@
...
+                visible: text
+                iconStyle: InfoTextArea.IconStyle.Error
+            }
+
+            Label {
+                Layout.fillWidth: true
+                wrapMode: Text.Wrap
+                text: [
+                    qsTr('You may extend your seed with custom words.'),
+                    qsTr('Your seed extension must be treated as sensible as your seed.'),
```
SomberNight: that's a weird sentence. just re grammar, would it not be "sensibly" instead?

note: qt gui text here: https://github.com/spesmilo/electrum/blob/9e752d2c672cba6cc363b38e853cb7950a159826/electrum/gui/qt/wizard/wallet.py#L558-L565

Could just use same `Your seed extension must be saved together with your seed.`?

f321x: Ok, changed it to the Qt sentence. 
But I do think its a bit weird to explicitly recommend storing both seed and passphrase together, isn't the main use-case of a passphrase (besides using it to create 'accounts' etc.) to have some form of geographic separation for increased security? Nevertheless it's probably best to use the same sentences in both guis yes.

SomberNight: Right... Well, the important point to convey to the user is that both secrets will be needed, and that they are both *secret*. If they understand the concept, hopefully they can make the jump to decide for themselves if they actually store them in separate locations. I think it's hard to be terse and concise here. I guess we basically want to say something like:

> Both the seed and this seed extension will be needed to restore your wallet, and must be kept secret.

Btw, note that "together" here is ambiguous, because ["together with"](https://www.merriam-webster.com/dictionary/together%20with) is a preposition meaning "in addition to" / "as well as". Maybe that's the meaning the original writer meant.

tests/test_wizard.py
```diff
@@ -226,10 +289,12 @@ async def test_2fa(self):
         self.assertEqual('trustedcoin_have_seed', v.view)
         d.update({
             'seed': 'oblige basket safe educate whale bacon celery demand novel slice various awkward',
-            'seed_type': '2fa', 'seed_extend': False, 'seed_variant': 'electrum',
+            'seed_type': '2fa', 'seed_extend': True, 'seed_variant': 'electrum',
             'seed_extra_words': False
```
SomberNight, line 293: This does not make sense. Where is the seed extension itself defined? Neither are we testing if it gets used.

SomberNight (commented):
Thanks for this. It looks good.

I will merge it *now* because we desperately need to improve test_wizard.py and I don't want to conflict with this PR.

tests/test_wizard.py
f321x: The point of always setting `seed_extend` to `True` in the qml `...have_seed` and `...create_seed` views is so that they always resolve to the `have_ext` or `create_ext` screen. The next screen then either adds `seed_extra_words` or sets `seed_extend` to `False` again.

See also https://github.com/spesmilo/electrum/pull/10016#discussion_r2200228564

Issue #9962, opened 2025-06-16
Reproducible build of appimage fails on Fedora 42 host


f321x:
### Description

There seems to be some issue making the `4.6.0b1` Appimage build non reproducible on my Fedora 42 host, either caused by Fedora or some issue with my OS, nevertheless some information of the host seems to leak into the build script altering it's output. Building on a clean Debian 12 VM produces the same outputs as the release. 

This is the Appimage created on my Fedora host:
https://drive.proton.me/urls/GZWW8NWVHW#2DcnVW15TfDF

Docker version:
```
user@hp:~$ docker version
Client: Docker Engine - Community
 Version:           28.2.2
 API version:       1.50
 Go version:        go1.24.3
 Git commit:        e6534b4
 Built:             Fri May 30 12:09:10 2025
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          28.2.2
  API version:      1.50 (minimum version 1.24)
  Go version:       go1.24.3
  Git commit:       45873be
  Built:            Fri May 30 12:07:24 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.27
  GitCommit:        05044ec0a9a75232cad458027ca83437aae3f4da
 runc:
  Version:          1.2.5
  GitCommit:        v1.2.5-0-g59923ef
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

System information:
```
Operating System: Fedora Linux 42
KDE Plasma Version: 6.3.5
KDE Frameworks Version: 6.14.0
Qt Version: 6.9.1
Kernel Version: 6.14.9-300.fc42.x86_64 (64-bit)
Graphics Platform: Wayland
```

This doesn't seem super urgent as it's still possible to reproduce on a Debian host, however opening this issue so we can investigate if more issues arise.

EDIT: output of diffoscope:
```diff
user@hp:~/code/trash$ diffoscope electrum-30e80f6d8-x86_64-fedora.AppImage electrum-30e80f6d8-x86_64-ubuntu.AppImage 
--- electrum-30e80f6d8-x86_64-fedora.AppImage
+++ electrum-30e80f6d8-x86_64-ubuntu.AppImage
├── strings --all --bytes=8 {}
│ @@ -33644,43 +33644,40 @@
│  0stcsfn4 
│  unQ_TI>3(
│  j|l\L<,h
│  \XUQMIEB?<8
[...]

PR #10019, opened 2025-07-11
appimage: bump appimagetool to new version/repo

### LINKED ISSUE #8105: AppImage fails to load libfuse.so.2
When running the appimage in Ubuntu (since jammy), currently kinetic 22.10), I get the erorr "error loading libfuse2.so.2".

To resolve, I must install libfuse2.so.2 with "apt install libfuse2".

libfuse2 has been deprecated and the default now is libfuse3.

f321x:
Updates the appimage build scripts to use the newer https://github.com/AppImage/appimagetool tool to bundle the appimage instead of the discontinued https://github.com/AppImage/AppImageKit. 
To prevent the new appimagetool from downloading a random "latest" `type2-runtime` binary (see [appimagetool_fetch_runtime.cpp](https://github.com/AppImage/appimagetool/blob/7cfafc45e5a8e64ad0755870b1001e5d7ffb4e85/src/appimagetool_fetch_runtime.cpp)) this PR also adds functionality to clone and build https://github.com/AppImage/type2-runtime from source. 
This is done using the build scripts provided in the `type2-runtime` repository, however to make the build of the runtime binary reproducible a patch is required which is applied on the repo before building the runtime binary. With this patch i try to reduce randomness in the build process to a minimum, though the most important part seems to be to not add the gnu debuglinks to the binary.

I tested the reproducibility on 3 systems, Ubuntu 24.04.2 LTS (VM), Debian 12 (VM) and on again on Ubuntu 24.04.2 but another machine. Fedora is not reproducible but that is a preexisting issue, potentially related to SELinux, see #9962. 

This fixes the issue of missing libfuse2 which users of the appimage have on "modern" distributions (see #8105).
The new `type2-runtime` is statically linked and includes the required dependencies now instead of relying on the host to provide them.

Fixes #8105


contrib/build-linux/appimage/build.sh
```diff
@@ -11,11 +11,41 @@ PROJECT_ROOT_OR_FRESHCLONE_ROOT="$PROJECT_ROOT"
...
+    info "building type2-runtime in build container..."
+    cd "$TYPE2_RUNTIME_REPO_DIR/scripts/docker"
+    env ARCH=x86_64 ./build-with-docker.sh
+    mv "./runtime-x86_64" "$TYPE2_RUNTIME_REPO_DIR/"
+
+    # clean up the empty created 'out' dir to prevent permission issues
+    rm -rf "$TYPE2_RUNTIME_REPO_DIR/out"
+
+    info "runtime build successful: $(sha256sum "$TYPE2_RUNTIME_REPO_DIR/runtime-x86_64")"
+) || fail "Failed to build type2-runtime"
```
SomberNight: Why is this not inside `make_appimage.sh` instead?
Is it to avoid docker-inside-docker? If so, we should at least put a short comment about that at the top (line 25).

> I tested the reproducibility on 3 systems, Ubuntu 24.04.2 LTS (VM), Debian 12 (VM) and again on Ubuntu 24.04.2 but another machine.

Have you tried to reproduce the full AppImage, or only the runtime? Because for the full AppImage, where we use the fresh_clone thing (e.g. `ELECBUILD_COMMIT=HEAD`), I don't see how the runtime would be found in the fresh clone?

(also, the uid/umask manipulations we do later in this script, so I wonder if running build.sh as a different userid (than 1000) or different umask causes issues)

f321x: > Why is this not inside make_appimage.sh instead?
Is it to avoid docker-inside-docker? If so, we should at least put a short comment about that at the top (line 25).

Yes its done so we can use the provided alpine (musl) based docker build pipeline of the runtime project, it seemed more elegant than doing some docker-in-docker hack. I have added a more detailed comment.

> Have you tried to reproduce the full AppImage, or only the runtime? Because for the full AppImage, where we use the fresh_clone thing (e.g. ELECBUILD_COMMIT=HEAD), I don't see how the runtime would be found in the fresh clone?

I built the full AppImages, which were reproducible, but I missed the `ELECBUILD_COMMIT` variable, i just manually did fresh clones. In https://github.com/spesmilo/electrum/pull/10019/commits/96751054f162df5b46ad3c4ae9022f2f1430dbbb i moved the runtime build part further down in the script to use `PROJECT_ROOT_OR_FRESHCLONE_ROOT` as base for the runtime build once it is set, so now it also works with the `ELECBUILD_COMMIT` env var.

> (also, the uid/umask manipulations we do later in this script, so I wonder if running build.sh as a different userid (than 1000) or different umask causes issues)

I just did some builds with UID 1001 and it all seems to work well for me. In https://github.com/AppImage/type2-runtime/blob/5e7217b7cfeecee1491c2d251e355c3cf8ba6e4d/scripts/docker/build-with-docker.sh#L9 they also use the host UID for the container (`-u "$(id -u):$(id -g)"`) so it should behave similar to our scripts iiuc. If we build with `ELECBUILD_COMMIT` in another UID the `chown` in https://github.com/spesmilo/electrum/blob/62a30c56882efe0f0dddb7b95b60657f20225505/contrib/build-linux/appimage/build.sh#L51 also changes the `runtime-x86_64` binary to UID 1000.

contrib/build-linux/appimage/patches/type2-runtime-reproducible-build.patch
```diff
@@ -0,0 +1,149 @@
+From 0c54d91dd1d33235ae97566600e692edfb613642 Mon Sep 17 00:00:00 2001
```
SomberNight, line 1: I am a bit a concerned re maintaining this patch, but I guess it is ok.
We probably don't need to bump the runtime commit often.
We will see.

PR #10027, opened 2025-07-14
wallet_db: add configvar for partial_writes, disable by default

SomberNight:
Adds a new configvar `WALLET_PARTIAL_WRITES` to enable/disable partial writes for the walletDB. This is a further restriction on top of the existing restrictions, e.g. wallet files still need to have file encryption disabled for partial writes. It defaults to off, so even for unencrypted wallets we disable partial writes for now.

This is used as a stopgap measure until we fix the issues found with the partial writes impl (see https://github.com/spesmilo/electrum/issues/10000).

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -515,7 +515,7 @@ class Daemon(Logger):
                 raise InvalidPassword('No password given')
             storage.decrypt(password)
         # read data, pass it to db
-        db = WalletDB(storage.read(), storage=storage, upgrade=upgrade)
+        db = WalletDB(storage.read(), storage=storage, upgrade=upgrade, config=config)
         if db.get_action():
             raise WalletUnfinished(db)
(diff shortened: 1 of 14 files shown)


ecdsa:
the diff could be made a lot smaller if we passed allow_partial_writes to the constructor, instead of the config object

SomberNight:
> the diff could be made a lot smaller if we passed allow_partial_writes to the constructor, instead of the config object

With a default value? (e.g. False?)
Without a default value, I don't see how it would be non-trivially smaller.

SomberNight:
I cheated a bit to decrease the diff by splitting off https://github.com/spesmilo/electrum/commit/3c82b00c5e71b1f69b0f2ea1a48ab061031c1200 (but that even contains a bug fix).

SomberNight:
ok, let's do https://github.com/spesmilo/electrum/pull/10031 instead

PR #10032, opened 2025-07-15
qt ConfirmTxDialog: also save custom fixed feerate fee_policies

SomberNight:
- save custom fixed feerate fee_policies, not just slider-restricted ones
- dynamically update fee_target text, to follow feerate_e
  - otherwise when the slider is set to "feerate", users would be shown two conflicting values

~regressions from https://github.com/spesmilo/electrum/pull/9597

### DIFF
--- a/electrum/gui/qt/confirm_tx_dialog.py
+++ b/electrum/gui/qt/confirm_tx_dialog.py
@@ -137,7 +137,7 @@ class TxEditor(WindowModalDialog):
         raise NotImplementedError()
 
     def update_fee_target(self):
-        text = self.fee_slider.fee_policy.get_target_text()
+        text = self.fee_policy.get_target_text()
         self.fee_target.setText(text)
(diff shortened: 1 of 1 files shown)


accumulator:
tested, looks good.

Some notes:
- Editing the absolute fee doesn't save the fee policy to the config, which is I presume is as intended.
- Editing the relative fee should probably update the fee policy combobox to 'Feerate'?
- Editing the relative fee empties the Fee policy text (should prob be similar to slider based feerate)
- Editing the absolute fee also empties the Fee policy text

f321x:
The changes make sense and work fine for me too. Noticed one minor exception related to `ConfirmTxDialog` fixed here: https://github.com/spesmilo/electrum/pull/10034

SomberNight:
> Editing the absolute fee doesn't save the fee policy to the config, which is I presume is as intended.

Yes, that is intended.

> Editing the relative fee should probably update the fee policy combobox to 'Feerate'?

Right, I thought about that. I think it's fine either way, so I won't touch it now. It worked this current way also in older versions, I checked.

Issue #10035, opened 2025-07-16
output of 'tap to thow seed extension' on android is truncated


jay-zk:
## description
seed field wraps text correctly but seed extension field truncates without scrolling option **after six words**

## steps to reproduce
in wallet details
1. tap to show seed (and extension)
2. compare text display in both fields

## impact 
**Medium**
users cannot fully corroborate or restore what they wrote down

## environment
- app version: 4.5.7
- protocol version: 1.4
- qt version: 6.4.3 
- pyqt version 6.4.2

accumulator:
thx, fixed in b9ea17674d4688986dbc9ec0f82f8242771bed03

PR #10036, opened 2025-07-16
qml: fix passphrase/seed extension wordwrap

accumulator:


SomberNight:
(ref https://github.com/spesmilo/electrum/issues/10035)

PR #10037, opened 2025-07-16
run_electrum: relax aiorpcx version check

schildbach:
Debian 13 (trixie) has aiorpcx 0.24, so this fix makes the Electrum 4.5.x branch compatible.

### DIFF
--- a/run_electrum
+++ b/run_electrum
@@ -74,8 +74,8 @@ def check_imports():
         import aiorpcx
     except ImportError as e:
         sys.exit(f"Error: {str(e)}. Try 'sudo python3 -m pip install <module-name>'")
-    if not ((0, 22, 0) <= aiorpcx._version <= (0, 24, 0)):
-        raise RuntimeError(f'aiorpcX version {aiorpcx._version} does not match required: 0.22.0<=ver<=0.24.0')
+    if not ((0, 22, 0) <= aiorpcx._version < (0, 26, 0)):
+        raise RuntimeError(f'aiorpcX version {aiorpcx._version} does not match required: 0.22.0<=ver<0.26.0')
     # the following imports are for pyinstaller
     from google.protobuf import descriptor
     from google.protobuf import message


run_electrum
```diff
@@ -74,8 +74,8 @@ def check_imports():
         import aiorpcx
     except ImportError as e:
         sys.exit(f"Error: {str(e)}. Try 'sudo python3 -m pip install <module-name>'")
-    if not ((0, 22, 0) <= aiorpcx._version < (0, 24)):
-        raise RuntimeError(f'aiorpcX version {aiorpcx._version} does not match required: 0.22.0<=ver<0.24')
+    if not ((0, 22, 0) <= aiorpcx._version <= (0, 24, 0)):
+        raise RuntimeError(f'aiorpcX version {aiorpcx._version} does not match required: 0.22.0<=ver<=0.24.0')
```
SomberNight: might as well relax it even more then, these versions should also work:
```suggestion
    if not ((0, 22, 0) <= aiorpcx._version < (0, 26, 0)):
        raise RuntimeError(f'aiorpcX version {aiorpcx._version} does not match required: 0.22.0<=ver<0.26.0')
```

schildbach: Sure, I'm all for it. In fact, I would not define an upper limit at all, except if there is a confirmed incompatibility. Upper limits tend to cause bit rot.

schildbach:
Feel free to squash when merging this. Or I can tidy it up as well, if you prefer.

SomberNight:
Thanks.

SomberNight:
As I mentioned out-of-band earlier, perhaps see if you find the [`electrum-env`](https://github.com/spesmilo/electrum/blob/48aa454786df8e2dc9d4aa20e857146656347686/electrum-env) script useful for your use case. With that, you would not rely on Debian packaging every dependency, instead you would get them from PyPI. What that script downloads is the same pinned dependencies that we bundle when building the binaries, so not arbitrary latest ones. (It was changed recently to this new behaviour in https://github.com/spesmilo/electrum/commit/01a1eacdb91222efc63dcb4eefa0f73ca9ab7978, but I think it makes much more sense than what it was doing previously)

OTOH I also understand the desire to simply use what is packaged by Debian. :)

PR #10048, opened 2025-07-18
fix: exception when clicking "New Channel"

f321x:
Fixes exception occuring when `wallet.lnworker` is `None` and the user clicks on `New Channel` in the channels list tab. The `New Channel` button is enabled when the wallet *can* have lightning, not depending on the wallet actually having lightning enabled.
With this patch the `init_lightning_dialog` will show up if the user clicks on `New Channel` but lightning is not yet enabled. I noticed this by restoring from a funded vpub and then loading the keystore from seed afterwards.
```
  File "/home/user/code/electrum-fork/electrum/gui/qt/main_window.py", line 2001, in new_channel_dialog
    if not lnworker.channels and not lnworker.channel_backups:
           ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'channels'

```


SomberNight:
> restoring from a funded vpub and then loading the keystore from seed afterwards.

alternatively if you restore from a vpub and have `config.ENABLE_ANCHOR_CHANNELS = False`, then wallet.can_have_lightning is True  (with anchors enabled it's false) and wallet.has_lightning is False

PR #10051, opened 2025-07-18
qml: ExceptionDialog: put user_text inside Flickable

SomberNight:
This tries to make the dialog better behave if the user types a lot of text.

note: I have not tested on Android.

### DIFF
--- a/electrum/gui/qml/components/ExceptionDialog.qml
+++ b/electrum/gui/qml/components/ExceptionDialog.qml
@@ -61,12 +61,21 @@ ElDialog
             Layout.fillWidth: true
             text: qsTr('Please briefly describe what led to the error (optional):')
         }
-        TextArea {
-            id: user_text
-            Layout.fillWidth: true
+        Flickable {
+            width: parent.width
             Layout.fillHeight: true
-            background: Rectangle {
-                color: Qt.darker(Material.background, 1.25)
+            contentHeight: user_text.height
+            interactive: height < contentHeight
+            clip: true
+
+            TextArea {
+                id: user_text
+                width: parent.width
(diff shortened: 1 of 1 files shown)


f321x:
> note: I have not tested on Android.

tested on android, works as expected

electrum/gui/qml/components/ExceptionDialog.qml
```diff
@@ -61,12 +61,21 @@ ElDialog
             Layout.fillWidth: true
             text: qsTr('Please briefly describe what led to the error (optional):')
         }
-        TextArea {
-            id: user_text
-            Layout.fillWidth: true
+        Flickable {
+            width: parent.width
```
accumulator, line 65: Avoid binding `width` (or `height`, `x`, `y`) in direct children of `Layout`. Direct children should as a rule of thumb always use `Layout` attached properties, e.g. `Layout.fillWidth: true`, or `Layout.preferredWidth: 100`, as `Layout` takes control of these properties in its children.

electrum/gui/qml/components/ExceptionDialog.qml
```diff
@@ -61,12 +61,21 @@ ElDialog
             Layout.fillWidth: true
             text: qsTr('Please briefly describe what led to the error (optional):')
         }
-        TextArea {
-            id: user_text
-            Layout.fillWidth: true
+        Flickable {
```
accumulator, line 64: Note that there is `controls/ElTextArea` that aims to solve the scrolling issue as well (albeit still a bit glitchy in its current form). It is used in e.g. `SignVerifyMessageDialog.qml`

PR #10049, opened 2025-07-18
fix: handle UserCancelled in show_transaction

f321x:
`TxDialog` may has to fetch tx information from the network on construction (`self.set_tx()`) and is showing a `RunCoroutineDialog`. If the user cancels this dialog the `UserCancelled` exception is not caught in `show_transaction()` and can even lead to Electrum crashing. 

Fixes #10041

Can be tested with this diff:
```diff
diff --git a/electrum/gui/qt/transaction_dialog.py b/electrum/gui/qt/transaction_dialog.py
index d8e708a95..3fa6bd36a 100644
--- a/electrum/gui/qt/transaction_dialog.py
+++ b/electrum/gui/qt/transaction_dialog.py
@@ -614,7 +614,7 @@ class TxDialog(QDialog, MessageBoxMixin):
         tx.add_info_from_wallet(self.wallet)
         # FIXME for PSBTs, we do a blocking fetch, as the missing data might be needed for e.g. signing
         # - otherwise, the missing data is for display-completeness only, e.g. fee, input addresses (we do it async)
-        if not tx.is_complete() and tx.is_missing_info_from_network():
+        if True:  # not tx.is_complete() and tx.is_missing_info_from_network():
             self.main_window.run_coroutine_dialog(
                 tx.add_info_from_network(self.wallet.network, timeout=10),
                 _("Adding info to tx, from network..."),
diff --git a/electrum/transaction.py b/electrum/transaction.py
index 8bbae09bf..e62c9d9a2 100644
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -1264,6 +1264,7 @@ class Transaction:
         timeout=None,
     ) -> None:
         """note: it is recommended to call add_info_from_wallet first, as this can save some network requests"""
+        import asyncio; await asyncio.sleep(10)
         if not self.is_missing_info_from_network():
             return
         if progress_cb is None:
```

PR #10052, opened 2025-07-18
crash reporter: rm `Never`, add confirm to `Send`, and only show window once per exc group

SomberNight:
This changes the crash reporter in the Qt/QML guis a bit.

- on clicking "Send", there is now an additional small confirmation popup
  - this makes it harder to accidentally send a report
- the "Never" button and the corresponding `config.SHOW_CRASH_REPORTER` option is removed
  -  we now always hook into `sys.excepthook` to show the crash reporter
- don't show crash reporter multiple times for the "same" exception
  - at least until the user restarts the program (there is no persistence)
  - exceptions are grouped the same way the crash reporter server is grouping them (I think)
  - in case there's e.g. a thread with timer spamming an exception erroneously,
    we won't spam the user with the crash reporter window

The motivation is https://github.com/spesmilo/electrum/pull/10049#issuecomment-3089278083 :
- it turns out @f321x accidentally clicked "Never" and set `config.SHOW_CRASH_REPORTER=False` some time in the past without realising it
- the GUI does not expose `config.SHOW_CRASH_REPORTER` any further, there is no way to reenable it after "Never" was clicked
- in that particular crash example, the whole Qt program was aborting (if `config.SHOW_CRASH_REPORTER is False`), as without hooking into sys.excepthook, a Qt thread died
  - by always hooking into `sys.excepthook`, the testing matrix becomes smaller
- I see two usecases for the (now-removed) "Never" option: 
  - (1) user never wants to send reports to us
    - this PR does not properly handle this, however at least it makes it harder to accidentally send a report
  - (2) an exception is getting spammed, e.g. from a thread on a timer, but for non-critical functionality, so that the program would be operational with the exception silenced
    - has happened in the past
    - this is now handled differently, by grouping the exceptions and only showing one reporter window for any given groupid

### DIFF
--- a/electrum/base_crash_reporter.py
+++ b/electrum/base_crash_reporter.py
@@ -25,7 +25,8 @@ import locale
 import traceback
 import sys
 import queue
-from typing import TYPE_CHECKING, NamedTuple, Optional
+from typing import TYPE_CHECKING, NamedTuple, Optional, TypedDict
+from types import TracebackType
 
 from .version import ELECTRUM_VERSION
 from . import constants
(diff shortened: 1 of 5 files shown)


f321x:
tested this with qt and qml, diff looks logical and works as expected for me.

faskalajob:
@SomberNight

PR #10062, opened 2025-07-20
qml: catch exceptions in QESwapHelper transport_task

f321x:
Adds an on_done_callback to the QESwapHelper.transport_task future to prevent the task from swallowing exceptions by retrieving exceptions and calling `util.send_exception_to_crash_reporter` with them.


accumulator:
Isn't this also catching transient I/O errors?

f321x:
I don't think so (just verified it by placing exceptions in the `NostrTransport`), as they as well seem to be swallowed in the task created by the context manager. 
However not sure if there maybe is a more suitable/generic solutions than my patch? I was just wondering that raising exceptions in the swap logic doesn't seem to trigger any crash reporter, which could lead to unnoticed exceptions.

SomberNight (approved):
I haven't tested but this looks reasonable.

PR #10078, opened 2025-07-23
simplify QR scanner QML interface, remove properties, add results to signals

accumulator:
android: pass on QR binary scan result data as well
qml: add signals for QEQRScanner and fallbacks, add QEBytes container type so we can pass along byte arrays between QML and python, port qr scan occurrences to new signals.


f321x:
Tested, looks good to me

PR #10091, opened 2025-07-29
wallet: don't spend reserve utxo to create new reserve utxo

f321x:
Right now if a lightning reserve utxo is required and there is already a reserve sized utxo available, `make_unsigned_transaction()` will still add the reserve as input to a 'spend max' transaction and add a reserve change
output. This seems wasteful, this patch instead just removes the input from the tx so it is not spent at all.
The `PartialTxOutput.is_utxo_reserve` flag is kept as it is also used in the qml gui.

### DIFF
--- a/tests/test_wallet_vertical.py
+++ b/tests/test_wallet_vertical.py
@@ -2060,9 +2060,10 @@ class TestWalletSending(ElectrumTestCase):
         fee_sats = 1000
         outputs = [PartialTxOutput.from_address_and_value(outgoing_address, balance - fee_sats)]
         def make_tx(b):
+            wallet.lnworker = mock.Mock()
+            wallet.lnworker.has_anchor_channels.return_value = b
             return wallet.make_unsigned_transaction(
                 outputs = outputs,
-                is_anchor_channel_opening = b,
                 fee_policy = FixedFeePolicy(fee_sats),
             )
         tx = make_tx(False)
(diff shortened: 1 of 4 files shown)


tests/test_wallet_vertical.py
```diff
@@ -2085,6 +2085,42 @@ def make_tx(address):
...
+        wallet2.sign_transaction(tx, password=None)
+        wallet1.adb.receive_tx_callback(tx, tx_height=1000100)
+
+        # now there is a reserve UTXO of config.LN_UTXO_RESERVE sat in wallet1, so wallet1 should
+        # not add it as input to the tx
+        assert len(wallet1.get_spendable_coins()) > 1, f"{len(wallet1.get_spendable_coins())=}"
+        tx = make_tx(outgoing_address1, wallet1)
+        self.assertEqual(1, len(tx.outputs()))
+        wallet1.adb.receive_tx_callback(tx, tx_height=1000200)
+        assert len(wallet1.get_spendable_coins()) == 1, f"{len(wallet1.get_spendable_coins())=}"
```
SomberNight, line 2125: maybe also repeat the assert from above, for completeness:
```
assert sum(utxo.value_sats() for utxo in wallet1.get_spendable_coins()) == self.config.LN_UTXO_RESERVE
```
or just e.g.
```
assert wallet1.get_spendable_coins()[0].value_sats() == self.config.LN_UTXO_RESERVE
```

tests/test_wallet_vertical.py
```diff
@@ -2085,6 +2085,42 @@ def make_tx(address):
...
+        instead of creating a new one
+        """
+        wallet1, outgoing_address1, to_self_address1 = self._create_cause_carbon_wallet()
+        wallet2 = self.create_standard_wallet_from_seed('cycle rocket west magnet parrot shuffle foot correct salt library feed song')
+        wallet2_addr = wallet2.get_receiving_addresses()[0]
+        def make_tx(address, wallet):
+            outputs = [PartialTxOutput.from_address_and_value(address, '!')]
+            return wallet.make_unsigned_transaction(
+                outputs = outputs,
+                is_anchor_channel_opening = True,
```
SomberNight: Could we instead of passing `is_anchor_channel_opening=True`, mock `lnworker.has_anchor_channels()`?
I think that might be less likely to break silently in the future.
https://github.com/spesmilo/electrum/blob/4a370af64a8df5041dd5837dfc0c366a6a6796a9/electrum/wallet.py#L1889

f321x: Ok, done in https://github.com/spesmilo/electrum/pull/10091/commits/23a48e7cfd5d742a33b21d086bcc59253c627283
Also mocked it for the two other tests, `test_ln_reserve_send_everything` and `test_ln_reserve_spend_max`

tests/test_wallet_vertical.py
f321x: Added in https://github.com/spesmilo/electrum/pull/10091/commits/23a48e7cfd5d742a33b21d086bcc59253c627283

SomberNight:
Looks good. Thanks!

PR #10092, opened 2025-07-30
clarified unconfirmed txn fee (sat/vB), mempool depth (vMB)

Geremia:
"Unconfirmed [1.8 sat/vB, 4.13 vMB]" is confusing. It should say: "Unconfirmed [fee: 1.8 sat/vB, mempool depth: 4.13 vMB]".

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -1749,12 +1749,12 @@ class Abstract_Wallet(ABC, Logger, EventListener):
             if fee is not None:
                 size = tx.estimated_size()
                 fee_per_byte = fee / size
-                extra.append(format_fee_satoshis(fee_per_byte) + f" {util.UI_UNIT_NAME_FEERATE_SAT_PER_VB}")
+                extra.append("fee: " + format_fee_satoshis(fee_per_byte) + f" {util.UI_UNIT_NAME_FEERATE_SAT_PER_VB}")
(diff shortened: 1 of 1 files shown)


SomberNight:
I think we need to be terse here, space is limited. This string is also displayed on mobile.
Which part exactly is the confusing part?
Also note that if the user inspects the details of the tx, they can learn/intuit what the string means, and from that point onwards the verbosity is not needed.

Geremia:
@SomberNight

> Which part exactly is the confusing part?

The second number (depth of the txn in fee-sorted mempool?), measured in vMB, confuses me. It starts and 0 and increases while the transaction is waiting for a confirmation?
There should at least be some documentation on it, or a descriptive tool-tip.

SomberNight:
In the transaction details dialog (right-click>`Details` or just double-click), the same number is shown with more context:
`Position in mempool: 0.23 vMB from tip`

<details>
 <summary>pic</summary>

depth
</details>

We could add even more description there, in the tx details dialog. That would be preferable.
e.g. "?" help button at the end of that line, or make the "Position in mempool" label clickable, or with a tooltip, etc.

Geremia:
@SomberNight 

> We could add even more description there, in the tx details dialog. That would be preferable.
> e.g. "?" help button at the end of that line, or make the "Position in mempool" label clickable, or with a tooltip, etc.

That would be good.

ecdsa:
closing: the PR is inactive.

PR #10093, opened 2025-07-31
txbatcher: don't spend anchors if ctx fee is already sufficient

f321x:
If the tx fee of the ctx is already higher than the required target it is not useful to spend the anchor so instead it is skipped.

As a side effect the closing tx is not shown in the history anymore until any output got claimed, only the lightning part of the closing group is shown, but the closing tx can be found in the channel details:
image
This is because `sweep_commitment_transaction` only adds the anchor and to_local outputs to the accounting addresses once there is a (nonlocal) spending tx, so it is missed by `get_onchain_history()`, same issue(?) for SRK channels. However only declaring them "accounting address" after the sweep seems intentional:
https://github.com/spesmilo/electrum/blob/dfd480d5dd86b3a5c358a0b4605fd88438fed93e/electrum/wallet.py#L420

### DIFF
--- a/tests/test_txbatcher.py
+++ b/tests/test_txbatcher.py
@@ -1,4 +1,3 @@
-import unittest
 import logging
 from unittest import mock
 import asyncio
@@ -6,14 +5,15 @@ import dataclasses
 
 from aiorpcx import timeout_after
 
-from electrum import storage, bitcoin, keystore, wallet
+import electrum.fee_policy
+from electrum import keystore, wallet, lnutil
 from electrum import SimpleConfig
 from electrum import util
-from electrum.address_synchronizer import TX_HEIGHT_UNCONFIRMED, TX_HEIGHT_UNCONF_PARENT, TX_HEIGHT_LOCAL
+from electrum.address_synchronizer import TX_HEIGHT_UNCONFIRMED
 from electrum.transaction import Transaction, PartialTxInput, PartialTxOutput, TxOutpoint
 from electrum.logging import console_stderr_handler, Logger
 from electrum.submarine_swaps import SwapManager, SwapData
-from electrum.lnsweep import SweepInfo
+from electrum.lnsweep import SweepInfo, sweep_ctx_anchor
 from electrum.fee_policy import FeeTimeEstimates
 
 from . import ElectrumTestCase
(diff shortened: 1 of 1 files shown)


electrum/txbatcher.py
```diff
@@ -321,14 +321,29 @@ def _to_sweep_after(self, tx: Optional[PartialTransaction]) -> Dict[TxOutpoint,
...
-            if not self.wallet.adb.db.get_transaction(prev_txid):
+            if not (prev_tx := self.wallet.adb.db.get_transaction(prev_txid)):
                 continue
             if sweep_info.is_anchor():
                 prev_tx_mined_status = self.wallet.adb.get_tx_height(prev_txid)
                 if prev_tx_mined_status.conf > 0:
                     self.logger.info(f"anchor not needed {prevout}")
                     self.batch_inputs.pop(prevout)  # note: if the input is already in a batch tx, this will trigger assert error
                     continue
+                prev_tx_current_fee = self.wallet.adb.get_tx_fee(prev_txid)
```
SomberNight, line 340: Just a note: I believe currently it just so happens that this does not occur here, but adb.get_tx_fee in some cases trusts the electrum server for the fee instead of calculating it itself.
Atm this does not seem to be the case as all the addresses watched by lnwatcher count as adb.is_mine and this includes the anchor outputs.

tests/test_txbatcher.py
```diff
@@ -87,6 +87,28 @@ def is_connected(self):
...
+    pubkey="034f986fd43561d52a96b19fbdd0c296f0442034d3f3f63fc394320a95750d42a8",
+)
+anchor_txin = sweep_ctx_anchor(ctx=anchor_chan_ctx, multisig_key=chan_multisig_key)
+ANCHOR_SWEEP_INFO = SweepInfo(
+    name='local_anchor',
+    cltv_abs=None,
+    txin=anchor_txin,
+    txout=None,
+    can_be_batched=True,
+)
```
SomberNight, line 113: could you rebase? the tests are failing

```
tests/test_txbatcher.py:None (tests/test_txbatcher.py)
tests/test_txbatcher.py:106: in <module>
    ANCHOR_SWEEP_INFO = SweepInfo(
E   TypeError: SweepInfo.__new__() missing 1 required positional argument: 'dust_override'

```

f321x: Rebased and fixed in https://github.com/spesmilo/electrum/pull/10093/commits/04a034e6bae168020b78e5f57197dc67b9c62a97

Issue #10030, opened 2025-07-15
Sub 1sat/vB PSBT can not be broadcasted


junderw:
### Description

https://github.com/spesmilo/electrum/blob/4.5.8/electrum/bitcoin.py#L350

Currently, over 25% of mining power supports sub 1sat/vB mempool admission.

This `max()` call is preventing this from working.

junderw:
Searching through the code there's a lot of repeated instanced of `FEERATE_DEFAULT_RELAY` and `FEERATE_DEFAULT_RELAY/1000` being used to clamp things.

junderw:
There's an obvious monkeypatch fix where we add something like a FEERATE_MIN_RELAY for clamping in places that actually deal with transactions... the metrics stuff can still clamp at FEERATE_DEFAULT_RELAY/1000 etc.

But this seems more like a judgement call. Right now mempool's Electrum RPC server is reporting sub-1sat/vB feerate, so this hard coded max() is preventing users from utilizing the Electrum RPC.

SomberNight:
> Currently, over 25% of mining power supports sub 1sat/vB mempool admission.

source for claim?

> mempool's Electrum RPC server is reporting sub-1sat/vB feerate

which server is that?

-----

electrum servers:

<details>
 <summary>script</summary>

```py
#!/usr/bin/env python3
import sys
import asyncio

from electrum.network import filter_protocol, Network
from electrum.util import create_and_start_event_loop, log_exceptions
from electrum.simple_config import SimpleConfig
from electrum.bitcoin import COIN


config = SimpleConfig()

loop, stopping_fut, loop_thread = create_and_start_event_loop()
network = Network(config)
network.start()

@log_exceptions
async def f():
    try:
        peers = await network.get_peers()
        peers = filter_protocol(peers)
        results = await network.send_multiple_requests(peers, 'blockchain.relayfee', [])
        for k, v in results.items():
            relayfee = round(COIN * v)
            print(f"{k=}, {relayfee=}. {'<<< lookie here <<<' if relayfee != 1000 else ''}")
        print(f"Received {len(results)} answers")
    finally:
        stopping_fut.set_result(1)

fut = asyncio.run_coroutine_threadsafe(f(), loop)
fut.result()
```
</details>

```
$ python3 ./electrum/scripts/relayfee.py
k=<ServerAddr host=164.92.148.39 port=50002 protocol=s>, relayfee=1000.
[...]

PR #10096, opened 2025-08-01
change minrelayfee clamps from `[1, 50]` to `[0.1, 50]` sat/vbyte

SomberNight:
Review should be done commit-by-commit. First few commits are conceptual clean-up.

- split dust_threshold from minrelayfee, make it hardcoded instead
  - the dust threshold in bitcoin core is calculated with a 3 sat/vbyte feerate,
    which we were interpreting as 3*minrelayfee
  - and now bitcoin core is considering changing the minrelayfee (1->0.1 s/b),
    independently from the dust threshold
- fee_policy: lower min relay fee from 1 sat/byte to 0.1 s/b
  - minrelayfee is still server-reported, but the clamps are changed from [1, 50] to [0.1, 50] sat/vbyte
  - dynamic feerates coming from the estimator are still clamped to [1, 1500] sat/vbyte
- ~~util: increase feerate quanta from 0.1 sat/vbyte to 0.01 sat/vbyte~~

closes https://github.com/spesmilo/electrum/issues/10030
ref https://github.com/bitcoin/bitcoin/pull/33106

### DIFF
--- a/electrum/bitcoin.py
+++ b/electrum/bitcoin.py
@@ -325,14 +325,14 @@ def construct_script(
 
 def relayfee(network: 'Network' = None) -> int:
     """Returns feerate in sat/kbyte."""
-    from .fee_policy import FEERATE_DEFAULT_RELAY, FEERATE_MAX_RELAY
+    from .fee_policy import FEERATE_MIN_RELAY, FEERATE_DEFAULT_RELAY, FEERATE_MAX_RELAY
     if network and network.relay_fee is not None:
         fee = network.relay_fee
     else:
         fee = FEERATE_DEFAULT_RELAY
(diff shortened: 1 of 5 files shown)


yonijuerga:
iit's neccesary! thxs!

accumulator:
LGTM

SomberNight:
> - util: increase feerate quanta from 0.1 sat/vbyte to 0.01 sat/vbyte

@ecdsa thinks the extra resolution is not worth it, too much mental load for the typical, higher feerates. I don't have a strong opinion so I am nuking that commit from here. https://github.com/spesmilo/electrum/commit/4dfc06a23c23ffc53e16c7ca97f4b5f45373f9f5

PR #10105, opened 2025-08-04
rpc: handle passing wallet path as 'wallet' argument

f321x:
Handles passing the wallet path as `wallet` command argument instead of `wallet_path`. This did work in 4.5.8 and fails in 4.6.0 as the wallet path string will get passed directly to the command instead of the required wallet object, causing an exception when methods are called on the str.


SomberNight:
This was intentionally removed in https://github.com/spesmilo/electrum/commit/37914d5af0d23d010ad66eabbb5fb5ad3ea89966, perhaps due to misunderstanding why it was there.
Some unit tests even got adapted so that they pass.

If we change it back, we need some kind of test that explicitly tests this behaviour, and explains we want it, otherwise we are just going back-and-forth :/

f321x:
will undraft after implementing unittests based on the command testing framework mentioned in https://github.com/spesmilo/electrum/pull/9847#discussion_r2251574609 once its on master.

PR #10114, opened 2025-08-08
wallet: add internal support for changing gap_limit_for_change, and use it in tests

SomberNight:
- this uses parts of https://github.com/spesmilo/electrum/pull/8726, to allow changing the gap limit for the change addresses
    - however the functionality is not exposed to the CLI/RPC
- instead we only use this to speed up the unit tests, by setting minimal gap limits

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -4,12 +4,14 @@ import unittest
 import threading
 import tempfile
 import shutil
+import functools
 
 import electrum
 import electrum.logging
 from electrum import constants
 from electrum import util
 from electrum.logging import Logger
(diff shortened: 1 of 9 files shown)


SomberNight:
measured on my laptop, with python 3.10, running all the unit tests take:
```
tests before:
	04:41  - 281s
	03:56  - 236s
	03:42  - 222s
	03:44  - 224s
	03:45  - 225s
tests after:
	03:04  - 184s
	02:56  - 176s
	02:55  - 175s
	02:54  - 174s
	02:59  - 179s
```
So the new median time is 0.78x the old median time.  (176s/225s)

PR #10118, opened 2025-08-09
fix(gui): avoid KeyError in QELnPaymentDetails.update when key missing issue:#10116

harsh1504660:
revents crash when Lightning payment key is missing in QELnPaymentDetails.update by using dict.get() and logging a warning instead of raising a KeyError

Issue #10116

### DIFF
--- a/electrum/gui/qml/qelnpaymentdetails.py
+++ b/electrum/gui/qml/qelnpaymentdetails.py
@@ -95,17 +95,20 @@ class QELnPaymentDetails(QObject):
             self._logger.error('wallet undefined')
             return
 
-        # TODO this is horribly inefficient. need a payment getter/query method
-        tx = self._wallet.wallet.lnworker.get_lightning_history()[self._key]
-        self._logger.debug(str(tx))
+        history = self._wallet.wallet.lnworker.get_lightning_history()
+        tx = history.get(self._key)
+        if tx is None:
+            self._logger.warning(f"Payment key {self._key} not found in history.")
+            return  # Or set fields to empty/default
 
         self._fee.msatsInt = 0 if not tx.fee_msat else int(tx.fee_msat)
         self._amount.msatsInt = int(tx.amount_msat)
         self._label = tx.label
         self._date = format_time(tx.timestamp)
         self._timestamp = tx.timestamp
-        self._status = 'settled'  # TODO: other states? get_lightning_history is deciding the filter for us :(
+        self._status = 'settled'
         self._phash = tx.payment_hash
         self._preimage = tx.preimage
 
         self.detailsChanged.emit()
+


electrum/gui/qml/qelnpaymentdetails.py
```diff
@@ -95,17 +95,20 @@ def update(self):
             self._logger.error('wallet undefined')
             return
 
-        # TODO this is horribly inefficient. need a payment getter/query method
-        tx = self._wallet.wallet.lnworker.get_lightning_history()[self._key]
-        self._logger.debug(str(tx))
+        history = self._wallet.wallet.lnworker.get_lightning_history()
+        tx = history.get(self._key)
+        if tx is None:
+            self._logger.warning(f"Payment key {self._key} not found in history.")
```
SomberNight, line 102: How can it happen that the key is missing?
How do you enter the this dialog? Is it directly after making a payment, or clicking an item in the history list? (those are the two entry points)

electrum/gui/qml/qelnpaymentdetails.py
```diff
@@ -95,17 +95,20 @@ def update(self):
             self._logger.error('wallet undefined')
             return
 
-        # TODO this is horribly inefficient. need a payment getter/query method
```
SomberNight, line 98: why delete this comment?

electrum/gui/qml/qelnpaymentdetails.py
```diff
@@ -95,17 +95,20 @@ def update(self):
...
+        if tx is None:
+            self._logger.warning(f"Payment key {self._key} not found in history.")
+            return  # Or set fields to empty/default
 
         self._fee.msatsInt = 0 if not tx.fee_msat else int(tx.fee_msat)
         self._amount.msatsInt = int(tx.amount_msat)
         self._label = tx.label
         self._date = format_time(tx.timestamp)
         self._timestamp = tx.timestamp
-        self._status = 'settled'  # TODO: other states? get_lightning_history is deciding the filter for us :(
```
SomberNight: why delete this comment?  seems unrelated

electrum/gui/qml/qelnpaymentdetails.py
harsh1504660: Hi @SomberNight this is what I am thinking,

- lnworker.get_lightning_history() is populated asynchronously (callbacks from lnwatcher/lntransport etc).

- The UI can request to show details in two main ways:

- Right after a payment event (e.g. send flow opens details / or notification allows “show details”), and

- From a user clicking a history item in the GUI list.

- If the GUI constructs the dialog with a key (payment hash) and the lnworker hasn’t yet inserted that key into the history, a dict[key] lookup raises KeyError.

- Similarly, if the history was refreshed such that the item was removed (cleanup, expiry, or replacement), the GUI may have a stale key when the user clicks.

Issue #10117, opened 2025-08-09
Android: Restoring a just-deleted wallet with the same seed fails with Error 'seed_extra_words' until app restart


value96:
### Description

**Environment**

- Electrum for Android: Version 4.6.1; Protocol version 1.4; Qt Version 6.4.3; PyQt Version 6.4.2
- Device: Galaxy A13, Android 13
- Install source: Google Play
- Wallet type: Standard wallet (no seed extension / no 2FA) 

**Steps to reproduce**

1. Create a Standard wallet in Electrum (no custom passphrase/seed extension).
2. Delete that wallet (Wallet → Delete). 
3. Create a new wallet → I already have a seed → enter the same seed you just deleted. 
4. Tap Next.

**Expected**
Wizard proceeds and restores the wallet.

**Actual**
A dialog appears:
`Error
'seed_extra_words'
`
The wizard cannot proceed. 

**Frequency**
Reproducible for me [e.g., 3/3 times] after deleting and immediately trying to restore the same seed.

**Workaround**
Restarting the app lets the same seed restore successfully.

f321x:
@value96 thanks for testing and reporting this. The issue should be fixed by https://github.com/spesmilo/electrum/pull/10120

PR #10121, opened 2025-08-12
wizard: raise more specific exc in create_storage() and pass some to crash reporter

f321x:
Currently the exceptions raised by `create_storage()` are all plain `Exception` and shown as warning in the GUI. This causes us to miss bugs like https://github.com/spesmilo/electrum/issues/10117 because they never get forwarded to the crash reporter.

This PR makes the exception types of `create_storage()` more specific, `UserFacingException` are still shown to the user as before, but other exception types are caught in the calling GUI and forwarded to the crash reporter instead.

### DIFF
--- a/electrum/gui/qt/__init__.py
+++ b/electrum/gui/qt/__init__.py
@@ -418,8 +418,7 @@ class ElectrumGui(BaseElectrumGui, Logger):
                                    parent=None,
                                    title=_('Error'),
                                    text=_('Cannot load wallet') + '(2) :\n' + err_text)
-            elif isinstance(e, WalletFileException) and e.should_report_crash \
-                    or not isinstance(e, WalletFileException):
+            else:
                 send_exception_to_crash_reporter(e)
             if app_is_starting:
                 # If we raise in this context, there are no more fallbacks, we will shut down.


electrum/gui/qt/__init__.py
```diff
@@ -411,12 +411,15 @@ def start_new_window(
...
+            if isinstance(e, UserFacingException) \
+                    or isinstance(e, WalletFileException) and not e.should_report_crash:
+                err_text = str(e) if isinstance(e, WalletFileException) else repr(e)
+                custom_message_box(icon=QMessageBox.Icon.Warning,
+                                   parent=None,
+                                   title=_('Error'),
+                                   text=_('Cannot load wallet') + '(2) :\n' + err_text)
+            elif isinstance(e, WalletFileException) and e.should_report_crash \
+                    or not isinstance(e, WalletFileException):
                 send_exception_to_crash_reporter(e)
```
SomberNight:
```suggestion
            else:
                send_exception_to_crash_reporter(e)
```

SomberNight: done in followup.

PR #10128, opened 2025-08-13
fix: psbt_nostr: don't save tx without txid

f321x:
Stops the psbt nostr plugin from trying to save transactions without txid to the wallet history and doesn't give the user the option to do so if the txid is not available.

### DIFF
--- a/electrum/plugins/psbt_nostr/qml.py
+++ b/electrum/plugins/psbt_nostr/qml.py
@@ -48,7 +48,7 @@ class QReceiveSignalObject(QObject):
         QObject.__init__(self)
         self._plugin = plugin
 
-    cosignerReceivedPsbt = pyqtSignal(str, str, str, str)
+    cosignerReceivedPsbt = pyqtSignal(str, str, str, str, bool)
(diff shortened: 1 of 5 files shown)


electrum/plugins/psbt_nostr/qml.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, plugin: 'Plugin'):
         QObject.__init__(self)
         self._plugin = plugin
 
-    cosignerReceivedPsbt = pyqtSignal(str, str, str, str)
+    cosignerReceivedPsbt = pyqtSignal(str, str, str, str, bool)
```
SomberNight, line 51: would be nicer if this had the same signature as in the desktop GUI:
```
cosignerReceivedPsbt = pyqtSignal(str, str, object, str)
```
so instead of passing `tx.serialize()`, we would pass the `Transaction` object and the recipients could easily calculate tx.txid() closer to where it's needed
OTOH it is not a simple change as `tx.txid` might have to be marked with `@pyqtSlot` for that to work >.<

PR #10133, opened 2025-08-14
qml: reduce tx history timestamp precision

f321x:
Reduces the precision of the date field in the qml transaction history
list to minutes. Seconds don't seem very useful in practice and add
clutter to the UI. This adapts the behavior to the Qt GUI.


SomberNight:
Great. Thanks!

PR #10149, opened 2025-08-19
swaps: make NostrTransport.send_direct_message() more robust

f321x:
Adds logic to retry sending a direct message in
`NostrTransport.send_direct_message()` on `TimeoutError`. Handles `TimeoutError` exception more gracefully by catching it and returning `None`.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -1567,25 +1567,37 @@ class NostrTransport(SwapServerTransport):
         tags = [['d', f'electrum-swapserver-{self.NOSTR_EVENT_VERSION}'],
                 ['r', 'net:' + constants.net.NET_NAME],
                 ['expiration', str(int(time.time() + self.OFFER_UPDATE_INTERVAL_SEC + 10))]]
-        event_id = await aionostr._add_event(
-            self.relay_manager,
-            kind=self.USER_STATUS_NIP38,
-            tags=tags,
-            content=json.dumps(offer),
-            private_key=self.nostr_private_key)
-        self.logger.info(f"published offer {event_id}")
+        try:
+            event_id = await aionostr._add_event(
+                self.relay_manager,
+                kind=self.USER_STATUS_NIP38,
+                tags=tags,
+                content=json.dumps(offer),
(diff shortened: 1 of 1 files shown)


electrum/submarine_swaps.py
```diff
@@ -1567,33 +1567,47 @@ async def publish_offer(self, sm: 'SwapManager') -> None:
...
-    async def send_direct_message(self, pubkey: str, content: str) -> str:
+        try:
+            event_id = await aionostr._add_event(
+                self.relay_manager,
+                kind=self.USER_STATUS_NIP38,
+                tags=tags,
+                content=json.dumps(offer),
+                private_key=self.nostr_private_key)
+            self.logger.info(f"published offer {event_id}")
+        except TimeoutError as e:
```
SomberNight:
```py
>>> import asyncio
>>> TimeoutError == asyncio.TimeoutError
False
```

f321x: Ah they changed that in Python 3.11 -> https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for.
Changed it to catch `asyncio.TimeoutError` in https://github.com/spesmilo/electrum/pull/10149/commits/37614e90924ebe0d38f26c3f7bfcd0b2d762e03c

electrum/submarine_swaps.py
```diff
@@ -1567,33 +1567,47 @@ async def publish_offer(self, sm: 'SwapManager') -> None:
...
+                tags=tags,
+                content=json.dumps(offer),
+                private_key=self.nostr_private_key)
+            self.logger.info(f"published offer {event_id}")
+        except asyncio.TimeoutError as e:
+            self.logger.warning(f"failed to publish swap offer: {str(e)}")
+
+    @ignore_exceptions
+    @log_exceptions
+    async def send_direct_message(self, pubkey: str, content: str, retries: int = 0) -> Optional[str]:
```
SomberNight, line 1583: nit: I think `retries` should be kwarg-only.

electrum/submarine_swaps.py
SomberNight: ah right. well okay.

PR #10152, opened 2025-08-19
swaps: nostr: add sanity checks for replies

SomberNight:
I am a bit fuzzy on how this stuff works exactly -- could you review, @f321x ?

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -1480,7 +1480,7 @@ class NostrTransport(SwapServerTransport):
         self.private_key = keypair.privkey
         self.nostr_private_key = to_nip19('nsec', keypair.privkey.hex())
         self.nostr_pubkey = keypair.pubkey.hex()[2:]
-        self.dm_replies = {}  # type: Dict[str, asyncio.Future[dict]]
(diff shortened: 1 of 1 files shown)


electrum/submarine_swaps.py
```diff
@@ -1636,11 +1636,12 @@ async def send_request_to_server(self, method: str, request_data: dict) -> dict:
         self.logger.debug(f"swapserver req: method: {method} relays: {self.relays}")
         request_data['method'] = method
         server_npub = self.config.SWAPSERVER_NPUB
-        event_id = await self.send_direct_message(server_npub, json.dumps(request_data), retries=1)
+        server_pubkey = aionostr.util.from_nip19(server_npub)['object'].hex()
```
f321x, line 1639: `send_direct_message()` already converts npubs to hex internally if an npub is passed as `pubkey` argument, however for expressiveness it might not harm to do it here explicitly?

f321x: ah noticed you use it for the dm_replies dict key

f321x:
Looks reasonable and works as expected.

PR #10153, opened 2025-08-19
lnpeer: add some rate-limits

SomberNight:


f321x:
FYI because somewhat related in https://github.com/f321x/lntransport-testing/blob/cbdb890f9319c6856a2fa19ac40156a33e5b3985/src/electrum_lntransport/lntransport.py#L386 i also implemented some way of preventing the peer from spamming us by using the cost calculation of aiorpcx (increasing when receiving data, decreasing over time) even before the received data is decrypted and potentially disconnecting if the cost has risen above a limit. 
However not clear yet if that will be used anytime soon and both limits could be useful to have.

SomberNight:
> i also implemented some way of preventing the peer from spamming us by using the cost calculation of aiorpcx (increasing when receiving data, decreasing over time) even before the received data is decrypted 

Nice.  I think these are complementary.

PR #10156, opened 2025-08-20
swaps: add sanity check to reverse swap mining fee

f321x:


electrum/submarine_swaps.py
```diff
@@ -920,6 +921,8 @@ async def reverse_swap(
         - Server fulfills HTLC using preimage.
 
         Note: expected_onchain_amount_sat is BEFORE deducting the on-chain claim tx fee.
+        Note: server_mining_fee_sat is passed as argument instead of accessing self.mining_fee to ensure
+        the mining fees the user sees in the GUI are also the values used for the checks performed here
```
SomberNight: Also say sth like:
> We commit to server_mining_fee_sat as it limits the max fee pre-payment amt, which the server is trusted with.

SomberNight:
Looks good. Thx!
I pushed a trivial follow-up.

PR #10159, opened 2025-08-20
logging: add config.LOGS_MAX_TOTAL_SIZE_BYTES: to limit size on disk

SomberNight:
This is about logging to disk, which happens if `config.WRITE_LOGS_TO_DISK` is enabled - which is not the default.

Currently on master:
- on every startup, a new logfile is created
  - the size of this logfile is unbounded
- old log files get deleted on startup, with only the newest few being kept
  - the number of files kept is configured via `config.LOGS_NUM_FILES_KEEP=30`

New:
- a new configvar is added: `config.LOGS_MAX_TOTAL_SIZE_BYTES=200_000_000`
- the logfile during a single process lifecycle is limited to `config.LOGS_MAX_TOTAL_SIZE_BYTES`
- old log files get deleted on startup, and the total size of those that are kept is limited to `config.LOGS_MAX_TOTAL_SIZE_BYTES`
- hence the max disk usage of log files is capped at `2 * config.LOGS_MAX_TOTAL_SIZE_BYTES`
  - however due to `config.LOGS_NUM_FILES_KEEP`, disk usage is typically not expected to get near this
- the log file is created using `logging.handlers.RotatingFileHandler(backupCount=4)`, which will result in multiple log files, each suffixed with `.1`, `.2`, etc. Each of these log files can grow up to `config.LOGS_MAX_TOTAL_SIZE_BYTES/(backupCount+1)`, and only the latest `backupCount+1` of them are kept.
  - so by default, each log file chunk created will be at most 40 MB, and only the last 5 for the current session is kept. The logger keeps deleting the oldest chunk whenever it creates a new one.
  - typical usage is expected to not result in any log rotation. So I expect approx 1 log file to be created per process launch.

### DIFF
--- a/electrum/logging.py
+++ b/electrum/logging.py
@@ -116,24 +116,52 @@ class TruncatingMemoryHandler(logging.handlers.MemoryHandler):
         super().close()
 
 
-def _delete_old_logs(path, *, num_files_keep: int):
-    files = sorted(list(pathlib.Path(path).glob("electrum_log_*.log")), reverse=True)
-    for f in files[num_files_keep:]:
+def _delete_old_logs(path, *, num_files_keep: int, max_total_size: int):
+    """Delete old logfiles, only keeping the latest few."""
+    def sortkey_oldest_first(p: pathlib.PurePath):
+        fname = p.name
+        basename, ext, counter = str(fname).partition(".log")
+        # - each time electrum is launched, there will be a new basename, ordered by date
+        # - for any given basename, there might be multiple log files, differing by counter
+        #   - empty counter is newest, then .1 is older, .2 is even older, etc
         try:
-            os.remove(str(f))
+            counter = int(counter[1:]) if counter else 0  # convert ".2" -> 2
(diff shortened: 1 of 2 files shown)


electrum/logging.py
```diff
@@ -116,24 +116,52 @@ def close(self) -> None:
...
+def _delete_old_logs(path, *, num_files_keep: int, max_total_size: int):
+    """Delete old logfiles, only keeping the latest few."""
+    def sortkey_oldest_first(p: pathlib.PurePath):
+        fname = p.name
+        basename, ext, counter = str(fname).partition(".log")
+        # - each time electrum is launched, there will be a new basename, ordered by date
+        # - for any given basename, there might be multiple log files, differing by counter
+        #   - empty counter is newest, then .1 is older, .2 is even older, etc
         try:
-            os.remove(str(f))
```
CasPerUD, line 123: 9900

CasPerUD, line 123: 9901

PR #10162, opened 2025-08-21
swaps: rate limit swapserver requests

f321x:
limits the swapserver to process at most one request every 5 seconds and drops incoming requests if a queue size of 5 is reached


SomberNight:
Thanks

PR #10164, opened 2025-08-21
swaps: cleanup data after swap gets failed

f321x:
Removes the persisted payment info from lnworker once a swap got failed. 
Stops persisting the OnionRoutingFailure as it is sufficient to delete the payment info to fail potential incoming htlcs. 
Deletes stored swap leftovers in lnworker and SwapManager

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2318,6 +2318,13 @@ class LNWallet(LNWorker):
                 return key_list
         return []
 
+    def delete_payment_bundle(self, payment_hash: bytes) -> None:
+        payment_key = self._get_payment_key(payment_hash)
+        for key_list in self.payment_bundles:
+            if payment_key in key_list:
+                self.payment_bundles.remove(key_list)
+                return
+
     def save_preimage(self, payment_hash: bytes, preimage: bytes, *, write_to_disk: bool = True):
         if sha256(preimage) != payment_hash:
             raise Exception("tried to save incorrect preimage for payment_hash")
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -42,7 +42,7 @@ from .json_db import StoredObject, stored_in
 from . import constants
 from .address_synchronizer import TX_HEIGHT_LOCAL, TX_HEIGHT_FUTURE
 from .fee_policy import FeePolicy
-from .invoices import Invoice
+from .invoices import Invoice, PR_PAID
 from .lnonion import OnionRoutingFailure, OnionFailureCode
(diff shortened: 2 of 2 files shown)


electrum/lnworker.py
```diff
@@ -2312,10 +2312,18 @@ def bundle_payments(self, hash_list):
         payment_keys = [self._get_payment_key(x) for x in hash_list]
         self.payment_bundles.append(payment_keys)
 
-    def get_payment_bundle(self, payment_key: bytes) -> Sequence[bytes]:
+    def get_payment_bundle(self, payment_key: bytes) -> Optional[Sequence[bytes]]:
         for key_list in self.payment_bundles:
             if payment_key in key_list:
                 return key_list
+        return None
```
SomberNight: note: I accidentally conflicted with this in https://github.com/spesmilo/electrum/commit/c1d2fc2b42ab68c36096283868c22f80395e5ece.  was pure chance, sorry

f321x: Fixed in https://github.com/spesmilo/electrum/commit/c623cca65468a4196ba88a0f65707bf55fcb1dca, thanks @ecdsa

electrum/submarine_swaps.py
```diff
@@ -370,16 +370,22 @@ def _fail_swap(self, swap: SwapData, reason: str):
...
                     self.logger.debug(f"swap {swap.payment_hash.hex()} has already been deleted.")
-                # TODO clean-up other swaps dicts, i.e. undo _add_or_reindex_swap()
+                if swap._funding_prevout is not None:
+                    self._swaps_by_funding_outpoint.pop(swap._funding_prevout, None)
+                self._swaps_by_lockup_address.pop(swap.lockup_address, None)
+                if swap.prepay_hash is not None:
+                    self._prepayments.pop(swap.prepay_hash, None)
+                    if self.lnworker.get_payment_status(swap.prepay_hash) != PR_PAID:
+                        self.lnworker.delete_payment_info(swap.prepay_hash.hex())
+                        self.lnworker.delete_payment_bundle(swap.payment_hash)
```
ecdsa, line 388: Shouldn't `delete_payment_bundle` also be called in case of success? Actually, bundles are conceptually independent from swaps, although they are currentlx only used in that context.

f321x: Yes they can be safely deleted after the swap is redeemed. I have added a commit https://github.com/spesmilo/electrum/pull/10164/commits/97a0d27fa6c50703262ceb5560846bb9207cf572 to delete the payment bundle and unregister the hold invoice callback after the swap has been set redeemed.

PR #10165, opened 2025-08-21
reverse swaps: in the CLI, replace 'server_mining_fee' with

ecdsa:
'prepayment', which corresponds to the trusted part of the lightning payment.

We use 2*sm.mining_fee, where 'mining_fee' is the flat part of the server fee. However, future protocol should probably allow to set a value that does not depend on 'mining_fee'. (note that LND uses a hardcoded amount).

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -1989,7 +1989,7 @@ class Commands(Logger):
                 "max_forward_sat": offer.pairs.max_forward,
                 "max_reverse_sat": offer.pairs.max_reverse,
                 "min_amount_sat": offer.pairs.min_amount,
-                "provider_mining_fee": offer.pairs.mining_fee,
+                "prepayment": 2 * offer.pairs.mining_fee,
             }
         return result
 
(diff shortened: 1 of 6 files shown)


electrum/commands.py
```diff
@@ -1989,7 +1989,7 @@ async def get_submarine_swap_providers(self, query_time=15, wallet: Abstract_Wal
                 "max_forward_sat": offer.pairs.max_forward,
                 "max_reverse_sat": offer.pairs.max_reverse,
                 "min_amount_sat": offer.pairs.min_amount,
-                "provider_mining_fee": offer.pairs.mining_fee,
+                "prepayment": 2 * offer.pairs.mining_fee,
```
SomberNight, line 1992: also need to update this in the unit tests

ecdsa: right

SomberNight (approved):
Looks good. Can merge after tests pass.

electrum/commands.py
```diff
@@ -1989,7 +1989,7 @@ async def get_submarine_swap_providers(self, query_time=15, wallet: Abstract_Wal
                 "max_forward_sat": offer.pairs.max_forward,
                 "max_reverse_sat": offer.pairs.max_reverse,
                 "min_amount_sat": offer.pairs.min_amount,
-                "provider_mining_fee": offer.pairs.mining_fee,
```
CasPerUD, line 1992: 8001

CasPerUD, line 1992: 9900

PR #10167, opened 2025-08-21
commands: use format_satoshis consistently. don't use sci-notation

SomberNight:
old behaviour:
```
>>> from electrum.commands import format_satoshis
>>> format_satoshis(1)
'1E-8'
```

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -36,7 +36,7 @@ import inspect
 from collections import defaultdict
 from functools import wraps
 from decimal import Decimal, InvalidOperation
-from typing import Optional, TYPE_CHECKING, Dict, List, Any
+from typing import Optional, TYPE_CHECKING, Dict, List, Any, Union
 import os
 import re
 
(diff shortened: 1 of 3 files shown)


electrum/commands.py
```diff
@@ -719,13 +725,13 @@ async def getbalance(self, wallet: Abstract_Wallet = None):
...
+        out = {"confirmed": format_satoshis(c)}
         if u:
-            out["unconfirmed"] = str(to_decimal(u)/COIN)
+            out["unconfirmed"] = format_satoshis(u)
         if x:
-            out["unmatured"] = str(to_decimal(x)/COIN)
+            out["unmatured"] = format_satoshis(x)
         if l:
-            out["lightning"] = str(to_decimal(l)/COIN)
+            out["lightning"] = format_satoshis(l)
```
SomberNight, line 734: self-note: need to do a follow-up as floats behave weird here.
util.format_satoshis* should use `to_decimal` internally for float conversion, just like the old code was doing that explicitly here
```py
>>> Decimal(20999999.12345678501)
Decimal('20999999.123456783592700958251953125')
```
or in sats, e.g.
```py
>>> Decimal(20999999_12345678.375)
Decimal('2099999912345678.5')
```

SomberNight: Ok, maybe this can't actually be triggered here. Anyway, https://github.com/spesmilo/electrum/commit/8d8d1dba0f204a42451741c7e7d937f76bb455ca is related

PR #10169, opened 2025-08-22
swaps: more cleanup

SomberNight:


electrum/submarine_swaps.py
```diff
@@ -79,46 +79,73 @@
 
 WITNESS_TEMPLATE_REVERSE_SWAP = [
```
f321x: Should the template maybe be renamed to `WITNESS_TEMPLATE_SWAP` as it is used for both types of swaps?
Edit: ok it is always a reverse swap for one side, still the name would make one expect there is also a non reverse swap template.

f321x:
looks reasonable to me. also did some test swaps on signet and everything works well.

electrum/submarine_swaps.py
SomberNight: Sure, we can rename it.
The name is for historical reasons, the forward and the reverse swaps used to have distinct witness scripts. The old flow for the forward swaps did not use hold invoices: we were using the Boltz backend and they designed it so that clients would not need to have support for hold invoices to be able to do a swap. Servers were assumed to be hold-invoice-capable, so reverse swaps used hold invoices. Hence the scripts were asymmetric.

note that the name is confusing for another reason, but that reason is recurring in many places in the codebase: this is not a template for a *witness* but for a *witness script*. The *witness script* (in p2wsh) is the script code that gets executed, *witness* is the items on the stack, used to satisfy the execution. but anyway

Done in https://github.com/spesmilo/electrum/pull/10169/commits/7989d6467becb608ca02cd96cf491db9d0239e0a

PR #10175, opened 2025-08-26
swaps: add explicit check that (onchain_locktime < LN_locktime)

SomberNight:
This was already implicitly checked. This diff makes the check explicit, and serves as an additional sanity-check.
- for client-forward-swaps, we have
  - "cltv safety requirement: (onchain_locktime < LN_locktime),   otherwise client is vulnerable"
  - server chooses onchain locktime delta = 70
    https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/submarine_swaps.py#L701
  - client checks that onchain locktime delta is <100
    https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/submarine_swaps.py#L887
  - client chooses LN locktime delta = 432
    https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/submarine_swaps.py#L907
- for client-reverse-swaps, we have
  - "cltv safety requirement: (onchain_locktime < LN_locktime),   otherwise server is vulnerable"
  - server chooses onchain locktime delta = 70
    https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/submarine_swaps.py#L598
  - server chooses LN locktime delta: unset, i.e. our default of 147
    https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/submarine_swaps.py#L612
    https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/lnworker.py#L2273

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2545,6 +2545,8 @@ class Peer(Logger, EventListener):
             return None, None
 
         # TODO check against actual min_final_cltv_expiry_delta from invoice (and give 2-3 blocks of leeway?)
+        # note: payment_bundles might get split here, e.g. one payment is "already forwarded" and the other is not.
+        #       In practice, for the swap prepayment use case, this does not matter.
         if local_height + MIN_FINAL_CLTV_DELTA_ACCEPTED > htlc.cltv_abs and not already_forwarded:
             log_fail_reason(f"htlc.cltv_abs is unreasonably close")
             raise exc_incorrect_or_unknown_pd
@@ -2580,7 +2582,9 @@ class Peer(Logger, EventListener):
                 return None, (payment_key, callback)
 
         # TODO don't accept payments twice for same invoice
-        # TODO check invoice expiry
+        # note: we don't check invoice expiry (bolt11 'x' field) on the receiver-side.
+        #       - semantics are weird: would make sense for simple-payment-receives, but not
+        #         if htlc is expected to be pending for a while, e.g. for a hold-invoice.
         info = self.lnworker.get_payment_info(payment_hash)
         if info is None:
             log_fail_reason(f"no payment_info found for RHASH {htlc.payment_hash.hex()}")
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -190,7 +190,7 @@ class SwapOffer:
 @attr.s
 class SwapData(StoredObject):
     is_reverse = attr.ib(type=bool)  # for whoever is running code (PoV of client or server)
-    locktime = attr.ib(type=int)
+    locktime = attr.ib(type=int)  # onchain, abs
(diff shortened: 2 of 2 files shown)


electrum/lnpeer.py
```diff
@@ -2545,6 +2545,8 @@ def log_fail_reason(reason: str):
             return None, None
 
         # TODO check against actual min_final_cltv_expiry_delta from invoice (and give 2-3 blocks of leeway?)
+        # note: payment_bundles might get split here, e.g. one payment is "already forwarded" and the other is not.
+        #       In practice, for the swap prepayment use case, this does not matter.
```
SomberNight, line 2549: btw this made me realise that the description I wrote for bundled payments is incorrect:
https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/lnworker.py#L2312-L2316
It is hard to explain exactly what it means to bundle some RHASHes...

electrum/submarine_swaps.py
```diff
@@ -653,12 +653,17 @@ def add_normal_swap(
             channels=channels,
             min_final_cltv_expiry_delta=min_final_cltv_expiry_delta,
         )
+        margin_to_get_refund_tx_mined = MIN_LOCKTIME_DELTA
+        if not (locktime + margin_to_get_refund_tx_mined < self.network.get_local_height() + lnaddr1.get_min_final_cltv_delta()):
+            raise Exception(
+                f"onchain locktime ({locktime}+{margin_to_get_refund_tx_mined}) "
+                f"too close to LN-htlc-expiry ({self.network.get_local_height()+lnaddr1.get_min_final_cltv_delta()})")
```
f321x, line 660: Only partly related to this, but i think we should probably also enforce that the sender actually sends us a htlc with the final cltv delta that we requested and not just accept any htlc with final cltv delta > 144 and hold it?

Swaps seems safe as in:
https://github.com/spesmilo/electrum/blob/71255c1e735cdfcd881e045b9f2f6bc6b599f459/electrum/lnpeer.py#L2548
we will fail htlcs that have `cltv_abs < local_height + 144`, as for the first iteration of `maybe_fulfill_htlc` `already_forwarded` is False and will then be true for following iterations due to the hold invoice callback. So we have at least 144 blocks until the htlc expires. After that `_claim_swap()` will indirectly fail the htlcs before they time out fully through `_fail_swap()` which works as the swap timeout is only 70 blocks, but that seems a bit indirect, maybe we should check that the htlc is correct directly when receiving it and fail it if it is bogus.

PR #10177, opened 2025-08-27
perform filesystem R/W checks at startup

accumulator:
perform filesystem R/W checks at startup. This avoids raising Exceptions later when actual writes are performed (e.g. #4151).

Not sure yet how portable this is, needs testing on win and mac

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -740,3 +740,18 @@ class Daemon(Logger):
             recent = recent[:5]
         self.config.RECENTLY_OPEN_WALLET_FILES = recent
         util.trigger_callback('recently_opened_wallets_update')
+
+
+def check_fs_permissions(config: 'SimpleConfig'):
+    electrum_path = config.electrum_path()
+    if not os.access(electrum_path, os.R_OK | os.W_OK):
+        raise Exception('can not read/write root folder at ' + electrum_path)
+    for f in ['config', 'blockchain_headers', 'recent_servers']:
+        fpath = os.path.join(electrum_path, f)
+        if os.path.exists(fpath):
+            if not os.access(fpath, os.R_OK | os.W_OK):
+                raise Exception(f'can not read/write {f} file at ' + electrum_path)
+    fpath = os.path.join(electrum_path, 'wallets')
+    if not os.access(fpath, os.R_OK | os.W_OK):
+        raise Exception('can not read/write wallet folder at ' + fpath)
+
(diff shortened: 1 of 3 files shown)


electrum/daemon.py
```diff
@@ -740,3 +740,18 @@ def update_recently_opened_wallets(self, wallet_path, *, remove: bool = False):
...
+
+
+def check_fs_permissions(config: 'SimpleConfig'):
+    electrum_path = config.electrum_path()
+    if not os.access(electrum_path, os.R_OK | os.W_OK):
+        raise Exception('can not read/write root folder at ' + electrum_path)
+    for f in ['config', 'blockchain_headers', 'recent_servers']:
+        fpath = os.path.join(electrum_path, f)
+        if os.path.exists(fpath):
+            if not os.access(fpath, os.R_OK | os.W_OK):
```
zrawsig, line 752: The os.access() check is repeated a few times with the exception message.  
This could be factored out to improve readability, for example:

```python
def check_permissions(path: str):
    if not os.access(path, os.R_OK | os.W_OK):
        raise Exception(f'Cannot read/write {path}')
```

electrum/daemon.py
```diff
@@ -740,3 +740,18 @@ def update_recently_opened_wallets(self, wallet_path, *, remove: bool = False):
             recent = recent[:5]
         self.config.RECENTLY_OPEN_WALLET_FILES = recent
         util.trigger_callback('recently_opened_wallets_update')
+
+
+def check_fs_permissions(config: 'SimpleConfig'):
+    electrum_path = config.electrum_path()
+    if not os.access(electrum_path, os.R_OK | os.W_OK):
+        raise Exception('can not read/write root folder at ' + electrum_path)
```
zrawsig, line 748: "can not"  is grammatically incorrect, please change to "Cannot". See the factored out example above.

electrum/daemon.py
```diff
@@ -740,3 +740,18 @@ def update_recently_opened_wallets(self, wallet_path, *, remove: bool = False):
             recent = recent[:5]
         self.config.RECENTLY_OPEN_WALLET_FILES = recent
         util.trigger_callback('recently_opened_wallets_update')
+
+
+def check_fs_permissions(config: 'SimpleConfig'):
```
zrawsig, line 745: This is the only function below the classes, while other generic functions such as this one lives in the beginning of the file.  Maybe it could be moved there to keep the file organized that way.

zrawsig (commented):
Suggested a refactor to improve the readability, small grammar nitpick and suggested to move the new function to the first half of the file to group all similar functions together and not mix with the classes.

PR #10188, opened 2025-09-01
qt: receive dialog: also propose swaps for 0 amount invoices

f321x:
Proposes a swap to the user even for 0 amount invoices if there is no available inbound liquidity at all. If there is no inbound liquidity it's not possible to receive anything, so the user should do a swap in this case too.

Also allows the user to modify the values in the swap dialog when it gets instantiated with a `recv_amount_sat` arg by receive tab, as the user might want to do a larger swap than the minimum required to receive the payment they wanted to request.


electrum/gui/qt/main_window.py
```diff
@@ -1276,7 +1276,7 @@ def run_swap_dialog(self, is_reverse=None, recv_amount_sat=None, channels=None):
         with transport:
             if not self.initialize_swap_manager(transport):
                 return
-            d = SwapDialog(self, transport, is_reverse=is_reverse, recv_amount_sat=recv_amount_sat, channels=channels)
+            d = SwapDialog(self, transport, is_reverse=is_reverse, recv_amount_sat=int(recv_amount_sat), channels=channels)
```
SomberNight: why the int cast?

e.g.
```
 10.31 | E | gui.qt.exception_window.Exception_Hook | exception caught by crash reporter
  File "/home/user/wspace/electrum/electrum/gui/qt/balance_dialog.py", line 264, in <lambda>
    swap_button.clicked.connect(lambda: self.window.run_swap_dialog())
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/wspace/electrum/electrum/gui/qt/main_window.py", line 1279, in run_swap_dialog
    d = SwapDialog(self, transport, is_reverse=is_reverse, recv_amount_sat=int(recv_amount_sat), channels=channels)
                                                                           ~~~^^^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

f321x: Originally I noticed some float values being passed to `SwapDialog`, but this cast is a leftover i forgot from testing, also had added another guarded cast in `SwapDialog` but this would have failed with `!`.

I renamed the var from `recv_amount_sat` to `recv_amount_sat_or_max` and added proper type hints and an assert so this is more robust in https://github.com/spesmilo/electrum/pull/10188/commits/45bdd6a827593949dbf4ad0e53c9abcf1e897328

PR #10178, opened 2025-08-27
android: make app compatible with edge-to-edge layout

f321x:
This PR makes the QML gui and the java android barcode scanner view compatible with the android edge-to-edge layout which is enforced from sdk version 35. The first commit also bumps the sdk version to 35.
It works by fetching the size of the status- and navigation bar from the android API and applying padding to the gui elements so that nothing is overlapped by the system bars.
I tested on Pixel 9, Android 16; Pixel 9 XL Android 16 emulator; Pixel 4a Android 13 emulator and Galaxy Tab A (2016) Android 8.1 (armv7).
Also tested with the 16kb patch commit of https://github.com/spesmilo/electrum/pull/10148.
If the edge-to-edge layout is not enforced (Android < 15) it doesn't opt-in and the UI stays as before.

Fixes #10139
Part of #9899

PR #10192, opened 2025-09-02
qml: android: add padding to report dialog in ExcDialog

f321x:
Adds padding to the report content dialog in the QML ExceptionDialog if required on android.
Followup https://github.com/spesmilo/electrum/pull/10178

Screenshot_20250902_103809


SomberNight:
Thanks.

PR #10198, opened 2025-09-05
swapserver: add commands for stats and swap history

f321x:
Adds two command to the swapserver plugin to retrieve some stats so swap providers can see more easily what is going on with their swapserver.


ecdsa:
how does this work with batching?

f321x:
@ecdsa single elements of the `swapserver_get_history` command could cover multiple swaps if they have been batched. This should then be visible through the label.

ecdsa:
Note: I am not sure if this is the right approach. A command that returns the history with grouping would be useful in commands.py, not only in this plugin. The reason why we do not have one is that I did not finish the history refactoring: while both `get_onchain_history` and `get_lightning_history` have been refactored and return sorted dicts of `HistoryItem`, `get_full_history` still returns a dict of dicts. I think it would be better to have `get_full_history` return stuctured data before exposing it to the CLI.

A work in progress is in https://github.com/spesmilo/electrum/tree/history_cleanup_part2

That being said, I do not mind merging that pull request, because it is a plugin, it does not touch the main codebase.

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,91 @@
...
+async def get_history(self: 'Commands', wallet: 'Abstract_Wallet' = None, plugin = None) -> List[dict]:
+    """
+Get a list of all swaps provided by this swapserver.
+Single elements can potentially cover multiple swaps if transactions have been batched.
+
+    Example result:
+
+        [
+            {
+                "date": "September 04, 2025",
```
SomberNight: please use iso 8601
e.g. 2025-09-04

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,91 @@
...
 SimpleConfig.SWAPSERVER_ANN_POW_NONCE = ConfigVar('plugins.swapserver.ann_pow_nonce', default=0, type_=int, plugin=__name__)
+
+
+@plugin_command('wl', plugin_name)
+async def get_history(self: 'Commands', wallet: 'Abstract_Wallet' = None, plugin = None) -> List[dict]:
+    """
+Get a list of all swaps provided by this swapserver.
+Single elements can potentially cover multiple swaps if transactions have been batched.
+
+    Example result:
```
SomberNight: I think docstrings should be indented so that they don't use the top/outermost indent.
e.g.
```suggestion
    """
    Get a list of all swaps provided by this swapserver.
    Single elements can potentially cover multiple swaps if transactions have been batched.

    Example result:
```
The only reason you would want to do this kind of indenting, is if the string is sensitive to whitespaces. E.g. consider: https://github.com/spesmilo/electrum/blob/8abe4a1dd44727cae532eadcec97cf72c1bb2cb4/electrum/simple_config.py#L691-L697
There the whitespaces would end up being in the string, and displayed in the GUI in an ugly way.

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,91 @@
...
+    swap_history_items = []
+    for swap_group_id in swap_group_ids:
+        if swap_history_item := full_history.get('group:' + swap_group_id):
+            swap_history_items.append(swap_history_item)
+
+    result = []
+    for swap in swap_history_items:
+        result.append({
+            'label': swap['label'],
+            'return': int(swap['value'].value),
```
SomberNight: label this with the unit, e.g. `return_sat`

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,91 @@
...
+@plugin_command('wl', plugin_name)
+async def get_summary(self: 'Commands', wallet: 'Abstract_Wallet' = None, plugin = None) -> dict:
+    """
+Get a summary of all swaps provided by this swapserver.
+Can become incorrect if closed lightning channels have been deleted in this wallet.
+
+    Example result:
+    {
+        "num_swaps": 160,
+        "overall_return": 159052,  # value earned or lost in satoshi
```
SomberNight: same: label this with the unit, e.g. `overall_return_sat`

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,91 @@
...
+        "swaps_per_day": 0.78  # between first swap and last swap
+    }
+
+    """
+    swap_history = await get_history(self)
+    profit_loss_sum = sum(swap['return'] for swap in swap_history)
+    first_swap = min(swap['timestamp'] for swap in swap_history)
+    last_swap = max(swap['timestamp'] for swap in swap_history)
+    swaps_per_day = len(swap_history) / ((last_swap - first_swap) // 86400)
```
SomberNight: divide by zero.

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,91 @@
...
+        "num_swaps": 160,
+        "overall_return": 159052,  # value earned or lost in satoshi
+        "swaps_per_day": 0.78  # between first swap and last swap
+    }
+
+    """
+    swap_history = await get_history(self)
+    profit_loss_sum = sum(swap['return'] for swap in swap_history)
+    first_swap = min(swap['timestamp'] for swap in swap_history)
+    last_swap = max(swap['timestamp'] for swap in swap_history)
```
SomberNight:
```
>>> min([])
  File "<python-input-1>", line 1, in <module>
    min([])
    ~~~^^^^
ValueError: min() iterable argument is empty
>>> max([])
  File "<python-input-2>", line 1, in <module>
    max([])
    ~~~^^^^
ValueError: max() iterable argument is empty
>>> sum([])
0
```

electrum/plugins/swapserver/__init__.py
f321x: The reason i did this was because it seems like the only option to get the first two lines of the docstring to align. Otherwise the second line is intended while the first one isn't. 
However this behavior is similar with other commands too so i just changed it now according to your suggestion.
```
usage: run_electrum swapserver_get_summary [-h]

Get a summary of all swaps provided by this swapserver.
    Can become incorrect if closed lightning channels have been deleted in this wallet.

    Example result:
    {
        "num_swaps": 160,
        "overall_return_sat": 159052,  # value earned or lost in satoshi
        "swaps_per_day": 0.78  # between first swap and last swap
    }

options:
  -h, --help  show this help message and exit

Run 'electrum -h' to see the list of global options
```

electrum/plugins/swapserver/__init__.py
f321x: Changed in https://github.com/spesmilo/electrum/pull/10198/commits/708aefb036a5522887245420ca95766082294f8b

electrum/plugins/swapserver/__init__.py
f321x: Changed in https://github.com/spesmilo/electrum/pull/10198/commits/708aefb036a5522887245420ca95766082294f8b

electrum/plugins/swapserver/__init__.py
f321x: Changed in https://github.com/spesmilo/electrum/pull/10198/commits/708aefb036a5522887245420ca95766082294f8b

electrum/plugins/swapserver/__init__.py
f321x: Oh yes, didn't consider this. Fixed in https://github.com/spesmilo/electrum/pull/10198/commits/708aefb036a5522887245420ca95766082294f8b

electrum/plugins/swapserver/__init__.py
f321x: Fixed https://github.com/spesmilo/electrum/pull/10198/commits/708aefb036a5522887245420ca95766082294f8b

SomberNight:
self-note: to test:
```
$ python3 -m unittest tests.regtest.TestLightningSwapserver.test_swapserver_success
$ ./run_electrum --regtest --dir /tmp/bob daemon -d
$ ./run_electrum --regtest --dir /tmp/bob load_wallet -w /tmp/bob/regtest/wallets/default_wallet
$ ./run_electrum --regtest --dir /tmp/bob swapserver_get_history
```

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,90 @@
...
+        "num_swaps": 160,
+        "overall_return_sat": 159052,  # value earned or lost in satoshi
+        "swaps_per_day": 0.78  # between first swap and last swap
+    }
+    """
+    swap_history = await get_history(self)
+    profit_loss_sum = sum(swap['return_sat'] for swap in swap_history) if swap_history else 0
+    first_swap = min(swap['timestamp'] for swap in swap_history) if swap_history else 0
+    last_swap = max(swap['timestamp'] for swap in swap_history) if swap_history else 0
+    days_in_operation = (last_swap - first_swap) // 86400
```
SomberNight, line 83: I think it would be more reasonable to round up, especially considering the first day of operation. E.g.:
```py
days_in_operation = (last_swap - first_swap) // 86400 + ((last_swap - first_swap) % 86400 > 0)
```
anyway, it's fine either way. We can keep it as-is.

electrum/plugins/swapserver/__init__.py
```diff
@@ -1,5 +1,90 @@
...
+plugin_name = "swapserver"
+
 
 SimpleConfig.SWAPSERVER_PORT = ConfigVar('plugins.swapserver.port', default=None, type_=int, plugin=__name__)
 SimpleConfig.SWAPSERVER_FEE_MILLIONTHS = ConfigVar('plugins.swapserver.fee_millionths', default=5000, type_=int, plugin=__name__)
 SimpleConfig.SWAPSERVER_ANN_POW_NONCE = ConfigVar('plugins.swapserver.ann_pow_nonce', default=0, type_=int, plugin=__name__)
+
+
+@plugin_command('wl', plugin_name)
+async def get_history(self: 'Commands', wallet: 'Abstract_Wallet' = None, plugin = None) -> List[dict]:
```
SomberNight, line 20: Perhaps we should have a very basic unit test for these, even though they are plugin-stuff.
But we can skip it this time.

electrum/plugins/swapserver/__init__.py
SomberNight: Weird. That's not how it looks like for me:
```
$ ./run_electrum --regtest --dir /tmp/bob swapserver_get_summary -h
usage: run_electrum swapserver_get_summary [-h]

Get a summary of all swaps provided by this swapserver.
Can become incorrect if closed lightning channels have been deleted in this wallet.

Example result:
{
    "num_swaps": 160,
    "overall_return_sat": 159052,  # value earned or lost in satoshi
    "swaps_per_day": 0.78  # between first swap and last swap
}

options:
  -h, --help  show this help message and exit

Run 'electrum -h' to see the list of global options
```

Do you see this behaviour for all commands? E.g. I see:
```
$ ./run_electrum lnpay -h
usage: run_electrum lnpay [-h] [--timeout TIMEOUT] [--max_cltv MAX_CLTV] [--max_fee_msat MAX_FEE_MSAT]
                          [--password PASSWORD]
                          invoice

Pay a lightning invoice
Note: it is *not* safe to try paying the same invoice multiple times with a timeout.
      It is only safe to retry paying the same invoice if there are no more pending HTLCs
      with the same payment_hash.  # FIXME should there even be a default timeout? just block forever.

positional arguments:
  invoice               Lightning invoice (bolt 11)

options:
  -h, --help            show this help message and exit
  --timeout TIMEOUT     Timeout in seconds (default=120)
  --max_cltv MAX_CLTV   Maximum total time lock for the route (default=4032+invoice_final_cltv_delta)
  --max_fee_msat MAX_FEE_MSAT
                        Maximum absolute fee budget for the payment (if unset, the default is a percentage fee
                        based on config.LIGHTNING_PAYMENT_FEE_MAX_MILLIONTHS)
  --password PASSWORD   Wallet password. Use '--password :' if you want a prompt.

Run 'electrum -h' to see the list of global options
```

f321x: I did yes, but now i cannot reproduce anymore. Maybe it was some terminal bug.

PR #10211, opened 2025-09-08
swaps: only set swap redeemed if preimage is available

f321x:
For forward swaps this will ensure that the swap only gets set redeemed and cleaned up after the preimage has been extracted, as it could happen that `current_height - spent_height > REDEEM_AFTER_DOUBLE_SPENT_DELAY` is true even if the preimage has not been extracted yet. This would make the client fail the pending htlcs if the following preimage extraction step is not successful due to any reason or if an async context switch happens in between (currently there is no await in the code between those steps).

For reverse swaps this changes nothing as they have the preimage stored from the beginning.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -464,7 +464,7 @@ class SwapManager(Logger):
             if spent_height in [TX_HEIGHT_LOCAL, TX_HEIGHT_FUTURE]:
                 spent_height = None
             if spent_height is not None:
-                if spent_height > 0:
+                if spent_height > 0 and swap.preimage:
                     if current_height - spent_height > REDEEM_AFTER_DOUBLE_SPENT_DELAY:
                         self.logger.info(f'stop watching swap {swap.lockup_address}')
                         swap.is_redeemed = True


electrum/submarine_swaps.py
```diff
@@ -464,7 +464,7 @@ async def _claim_swap(self, swap: SwapData) -> None:
             if spent_height in [TX_HEIGHT_LOCAL, TX_HEIGHT_FUTURE]:
                 spent_height = None
             if spent_height is not None:
-                if spent_height > 0:
+                if spent_height > 0 and swap.preimage:
```
SomberNight, line 467: What if we never learn the preimage?  Is it ok that this cleanup logic will never run?
For example, user starts a forward-swap, receives and holds LN-htlcs, funds the onchain script (only in mempool) and goes offline for long time. When he comes back again thousands of blocks later, he will timeout onchain the htlc. He will never learn the preimage.

f321x: If a forward swap fails to extract the preimage and `spent_height > 0` it will get failed assuming the spent is its own refund:
https://github.com/spesmilo/electrum/blob/61aeffd26f54653dfdb8b242da7c11ef34e19300/electrum/submarine_swaps.py#L488

`_fail_swap()` will then do the cleanup.

SomberNight: I see. Well okay.

PR #10216, opened 2025-09-10
adb: change API of util.TxMinedInfo: height() is now always SPV-ed

SomberNight:
Conceptually this is quite an invasive change that transforms unverified txs to unconfirmed txs.

I think it is a code-safety improvement that might prevent security-bugs.
In particular the TxInput.block_height and TxInput.spent_height not being SPV-ed was really bothering me. This fixes that.

### DIFF
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -140,6 +140,7 @@ class AddressSynchronizer(Logger, EventListener):
     def get_address_history(self, addr: str) -> Dict[str, int]:
         """Returns the history for the address, as a txid->height dict.
         In addition to what we have from the server, this includes local and future txns.
+        Note: heights are SPV-verified.
 
         Also see related method db.get_addr_history, which stores the response from the server,
         so that only includes txns the server sees.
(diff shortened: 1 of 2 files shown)


angelsmia777-milionz (commented):
Changes are complete everything is in order

electrum/util.py
```diff
@@ -1264,26 +1264,39 @@ def func_wrapper(self, *args, **kwargs):
...
+@dataclass(frozen=True, kw_only=True)
+class TxMinedInfo:
+    _height: int                       # height of block that mined tx
     conf: Optional[int] = None         # number of confirmations, SPV verified. >=0, or None (None means unknown)
     timestamp: Optional[int] = None    # timestamp of block that mined tx
     txpos: Optional[int] = None        # position of tx in serialized block
     header_hash: Optional[str] = None  # hash of block that mined tx
     wanted_height: Optional[int] = None  # in case of timelock, min abs block height
 
+    def height(self, *, requires_spv: bool = True) -> int:
```
ecdsa: `require_spv` is unused, do we really need to add this option?

SomberNight: You're right. I had anticipated that to be needed but ended up not using it anywhere.
Force-pushed in https://github.com/spesmilo/electrum/compare/4d0a082e1e3b636af8e6b9dd955dee6e67eaa5ac..a4293c483e3ac3d83b501d8de1fe11799d9f818e

PR #10222, opened 2025-09-13
docs: add note for zsh users to quote extras during pip install

PeytonAllworth:
Adds a short note for zsh users explaining that they must quote extras
(e.g. `pip install '.[fast]'`) when installing to avoid "no matches found"
errors caused by shell globbing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,17 @@ so make sure that is on your `PATH` variable.
 
 ### Development version (git clone)
 
+$ python3 -m pip install --user -e .
+
+> **Note for zsh users:** If your shell is `zsh`, you might see  
+> `zsh: no matches found: .[fast]`  
+> when installing. To fix this, quote the extras like so:
+>
+> ```bash
+> python3 -m pip install --user '.[fast]'
+> ```
+
+
 _(For OS-specific instructions, see [here for Windows](contrib/build-wine/README_windows.md),
 and [for macOS](contrib/osx/README_macos.md))_
 


PeytonAllworth:
This PR only modifies README.md, so the failing test_payment_multipart_trampoline_e2e in test_lnpeer.py seems unrelated. Could we re-run the CI job?

README.md
```diff
@@ -92,6 +92,17 @@ so make sure that is on your `PATH` variable.
...
 ### Development version (git clone)
 
+$ python3 -m pip install --user -e .
+
+> **Note for zsh users:** If your shell is `zsh`, you might see  
+> `zsh: no matches found: .[fast]`  
+> when installing. To fix this, quote the extras like so:
+>
+> ```bash
+> python3 -m pip install --user '.[fast]'
```
SomberNight, line 102: AFAICT all mentions of "pip install ..." with extras already has it quoted.
For example:
https://github.com/spesmilo/electrum/blob/c1ab6cda534bfe5bdb2530a1672fddba68a3f179/README.md#L26
https://github.com/spesmilo/electrum/blob/c1ab6cda534bfe5bdb2530a1672fddba68a3f179/contrib/build-wine/README_windows.md#L19
https://github.com/spesmilo/electrum/blob/c1ab6cda534bfe5bdb2530a1672fddba68a3f179/contrib/osx/README_macos.md#L31

If you find one that does not, you can add quotes at there.

README.md
```diff
@@ -92,6 +92,17 @@ so make sure that is on your `PATH` variable.
 
 ### Development version (git clone)
 
+$ python3 -m pip install --user -e .
```
SomberNight, line 95: why add this command here? In terms of ordering, this does not make sense for this section.
The command already appears on line 107:
https://github.com/spesmilo/electrum/blob/c1ab6cda534bfe5bdb2530a1672fddba68a3f179/README.md#L107

README.md
SomberNight: I guess if you think that helps, we could quote even the lone dots. e.g.
`$ python3 -m pip install --user -e "."`

SomberNight:
Thanks but I think this is overkill :)

Issue #10220, opened 2025-09-12
Windows zip plugins readfile and manifest icon


copronista:
### Description
trying to use: 
https://github.com/spesmilo/electrum-plugins 
Exception shown in debug terminal:
```
26.67 | E | plugin | Plugin error. plugin: virtualkeyboard, hook: password_dialog
Traceback (most recent call last):
File "electrum\plugin.py", line 825, in run_hook
File "C:\Users\paolo\AppData\Roaming\Electrum\plugins\virtualkeyboard-0.0.1.zip\virtualkeyboard\qt.py", line 17, in password_dialog
vkb_button.setIcon(read_QIcon_from_bytes(self.read_file("keyboard-icon.svg")))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "electrum\plugin.py", line 898, in read_file
File "electrum\plugin.py", line 798, in read_file
File "zipfile_init_.py", line 1608, in open
File "zipfile_init_.py", line 1536, in getinfo
KeyError: "There is no item named 'virtualkeyboard\\keyboard-icon.svg' in the archive"

```
this is worst when trying to add an icon in manifest.json

Non zipped plugins are not affected.
Linux not affected.
Problem only affect ZIP plugins on windows.

ecdsa:
I do not have a windows machine at the moment.
Could you test if `zipfile` expects a path that differs from what `os.path.join` returns?
For example, you could print the output of  `myzip.namelist()` before calling `myzip.open`

ecdsa:
fixed by #10224

PR #10224, opened 2025-09-14
Update plugin.py

copronista:
https://github.com/spesmilo/electrum/issues/10220


SomberNight:
This is the kind of change that without a unit test needs at least a comment in the code, or at the very least rationale in the commit message.

Please write better commit messages. Please don't merge commits to master that just say "update file". At least there is a link to an issue but what if github disappears or deletes our project, or the user deletes their comment from the thread?

PR #10227, opened 2025-09-18
Add Guardian Address plugin for physical coercion resistance

bitcoinguardian:
- Implements BIP draft for Guardian Address signaling protocol
- Provides lock/unlock wallet functionality via on-chain signals
- Includes mempool monitoring and automatic protection application
- Supports both mempool.space and blockstream.info APIs
- Adds Tools > Guardian Settings menu for configuration


bitcoinguardian:
Wrong repo.

PR #10234, opened 2025-09-23
wallet: anchor reserve: handle wallet.use_change being off

SomberNight:
fixes https://github.com/spesmilo/electrum/issues/10231

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -1850,6 +1850,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
     def get_change_addresses_for_new_transaction(
             self, preferred_change_addr=None, *, allow_reusing_used_change_addrs: bool = True,
     ) -> List[str]:
+        """note: might return an empty list! (e.g. if use_change is disabled, or allow_reuse is False)"""
         change_addrs = []
         if preferred_change_addr:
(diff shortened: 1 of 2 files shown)


electrum/wallet.py
```diff
@@ -2111,7 +2111,8 @@ def distribute_amount(amount):
                     to_distribute -= reserve_sized_input.value_sats()
                 else:
                     self.logger.info(f'Adding change output to meet utxo reserve requirements')
-                    change_addr = self.get_change_addresses_for_new_transaction(change_addr)[0]
+                    change_addrs = self.get_change_addresses_for_new_transaction(change_addr) or [tx_inputs[0].address]
+                    change_addr = change_addrs[0]
```
ecdsa: I think it would be easier to read if we do one thing at a time:
```python
change_addrs = self.get_change_addresses_for_new_transaction(change_addr)
change_addr = change_addrs[0] if change_addrs else tx_inputs[0].address

```

SomberNight: ok. I force-pushed that in https://github.com/spesmilo/electrum/compare/ff30e7c81b833b46093ee2e6ae12208898c81ec6..6a9c94ff0b8d9c91bb15b4689948b6fab49f4e28

PR #10238, opened 2025-09-26
ci: only run coveralls on one of the unittest tasks

SomberNight:
coveralls has been quite flaky recently -- let's try to lower the number of CI tasks where it spreads the flakiness.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -55,7 +55,7 @@ task:
                -m pytest tests -v
     - coverage report
   coveralls_script:
-    - if [ ! -z "$COVERALLS_REPO_TOKEN" ] ; then coveralls ; fi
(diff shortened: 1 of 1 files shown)


ecdsa:
would it make sense to run coverall as a separate test?

SomberNight:
Not really. I think it depends on the prior `coverage run` and `coverage report` calls.
As in, it cannot be easily separated from running the unit tests.

AFAIU,
- `coverage run` runs the unit tests with pytest while keeping track of which lines are hit.
- `coverage report` prints the stats to stdout
- `coveralls` uploads the stats to an HTTP endpoint at `https://coveralls.io`

-----

As coveralls.io has been unreliable recently, it raises the question what value it even brings to us.
For example, when it works, you can view nice historical graphs re how our test coverage changed over time.
It also gives us that "badge" in the readme:
https://github.com/spesmilo/electrum/blob/14d7c964195ce40b8bb8074b0601ef387967bd8d/README.md#L11
but I guess that's it?

SomberNight:
ref https://github.com/spesmilo/electrum/pull/3245

PR #10239, opened 2025-09-26
tests: clear util.callback_mgr between test cases

SomberNight:
`util.callback_mgr.callbacks` was not getting properly cleared between tests. Every time an `Abstract_Wallet` or an `LNWorker` (or many other subclasses of `EventListener`) is instantiated, `self.register_callbacks()` is called in `__init__`, which puts callbacks into `util.callback_mgr.callbacks`. These are only cleaned up if we explicitly call `Abstract_Wallet.stop()` or `LNWorker.stop()` later, which we usually do not do in the tests.

As a result, when running multiple unit tests in a row, lots of objects created in a given testcase are never GC-ed and leak into subsequent tests. This is not only a memory leak, but wastes compute too: when events are triggered and cbs get called, these old objects also have their cbs called.

After running all (~1061) unit tests, I observe `util.callback_mgr.callbacks` had 30 events with a total of 3156 callbacks stored.

On my laptop, running all unit tests previously took ~115 sec, and now it takes ~73 sec.

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1953,20 +1953,24 @@ class CallbackManager(Logger):
     def __init__(self):
         Logger.__init__(self)
         self.callback_lock = threading.Lock()
-        self.callbacks = defaultdict(list)      # note: needs self.callback_lock
+        self.callbacks = defaultdict(list)  # type: Dict[str, List[Callable]]  # note: needs self.callback_lock
 
-    def register_callback(self, func, events):
+    def register_callback(self, func: Callable, events: Sequence[str]) -> None:
         with self.callback_lock:
             for event in events:
                 self.callbacks[event].append(func)
 
-    def unregister_callback(self, callback):
+    def unregister_callback(self, callback: Callable) -> None:
         with self.callback_lock:
             for callbacks in self.callbacks.values():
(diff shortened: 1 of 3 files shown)


SomberNight:
Compare e.g. [this run](https://github.com/spesmilo/electrum/pull/10239/checks?check_run_id=51346019848) of `unittests: py3.10`, finishing with `02:22 pytest`,
and these prior runs for recent commits:
- [04:23 pytest](https://github.com/spesmilo/electrum/runs/51339102690)
- [04:16 pytest](https://github.com/spesmilo/electrum/runs/51338467306)
- [05:37 pytest](https://github.com/spesmilo/electrum/runs/51138195978)
- [04:42 pytest](https://github.com/spesmilo/electrum/runs/50324405100)

(Notice the significant variance though. Perhaps due to difference in CPU / load of the CI runners)

SomberNight:
The python "debug mode" CI task seems to be the most affected:

going from
[34:21 pytest](https://github.com/spesmilo/electrum/runs/51339104793)
to
[04:56 pytest](https://github.com/spesmilo/electrum/pull/10239/checks?check_run_id=51346020773)

Issue #10163, opened 2025-08-21
swaps: potential send change to lightning option improvements 


f321x:
### Description

There are multiple rough edges when using the send change to lightning option, in Qt I noticed:
* The (onchain) tx fee shown for the transaction is misleading as it doesn't contain the swap fees. I think it would be nice if the tx fee field would simply show the whole expected fee, including the swap fee. So there is only a single fee value and it represents the true fee that will be paid when initiating the transaction.
* coin selection doesn't respect the min/max swap amounts and has no knowledge of them (as they are only fetched after constructing the tx). This results in unreliable tx flow as the tx will often fail if the coin selection happens to select utxos that are outside of the supported swap limits. If no UTXO is available in the possible swap range the user should either not be able to initiate the tx, by being forced to disable send change to lightning or select a different swap provider if available, or alternatively the transaction could just create change as fallback.
* It is possible to set a very low onchain fee which could result in the swap output being refunded and other unpleasant UX, the minimal allowed ETA/tx fee should be limited (same as in the QML forward swap finalizing screen PR, see https://github.com/spesmilo/electrum/pull/9992).

PR #10247, opened 2025-09-30
qt: improve send_change_to_lightning feedback

f321x:
Improves the GUI feedback given to the user when sending change to lightning by showing different failure messages and the additional fee charged for the swap.
Also considers the min/max swap amount announced by the provider when deciding to add the swap dummy output in `make_tx` for improved reliability.
Mostly fixes [github.com/spesmilo/electrum/issues/10163](https://github.com/spesmilo/electrum/issues/10163)

Screenshot_20260122_101531

### DIFF
--- a/electrum/gui/qt/confirm_tx_dialog.py
+++ b/electrum/gui/qt/confirm_tx_dialog.py
@@ -1043,14 +1043,16 @@ class TxEditor(WindowModalDialog, QtEventListener, Logger):
         if self.expected_onchain_amount_sat < self.swap_manager.get_min_amount():
             return '\n'.join([
                 _("Payment amount below the minimum possible swap amount."),
+                _("Minimum amount: {}").format(f(self.swap_manager.get_min_amount())), "",
                 _("You need to send a higher amount to be able to do a Submarine Payment."),
-                _("Minimum amount: {}").format(f(self.swap_manager.get_min_amount())),
             ])
 
-        too_low_outbound_liquidity_msg = '\n'.join([
-            _("You don't have enough outgoing capacity in your lightning channels."),
+        too_low_outbound_liquidity_msg = ''.join([
+            _("You don't have enough outgoing capacity in your lightning channels."), '\n',
+            _("Your lightning channels can send: {}").format(f(ln_can_send)), '\n',
+            _("For this transaction you need: {}").format(f(self.lightning_send_amount_sat)) if self.lightning_send_amount_sat else '',
+            '\n\n' if self.lightning_send_amount_sat else '\n',
             _("To add outgoing capacity you can open a new lightning channel or do a submarine swap."),
-            _("Your lightning channels can send: {}").format(f(ln_can_send)),
         ])
 
         # prioritize showing the swap provider liquidity warning before the channel liquidity warning
@@ -1062,8 +1064,8 @@ class TxEditor(WindowModalDialog, QtEventListener, Logger):
                 provider_liquidity = 0
             msg = [
                 _("The selected swap provider is unable to offer a forward swap of this value."),
+                _("Available liquidity") + f": {f(provider_liquidity)}", "",
                 _("In order to continue select a different provider or try to send a smaller amount."),
-                _("Available liquidity") + f": {f(provider_liquidity)}",
             ]
             # we don't know exactly how much we need to send on ln yet, so we can assume 0 provider fees
             probably_too_low_outbound_liquidity = self.expected_onchain_amount_sat > ln_can_send


ecdsa:
I do not like that we are doing a nostr `get_pairs` request ahead of transaction finalization. 
This can be very slow, and the user may want to disable the option before any request is made.

Also, this logic leads to closing the dialog if we toggle the option, which is quite unfriendly.

f321x:
If we want to show the user how many fees they are going to pay (`Additional fees:`) and make the flow reliable we need to fetch the pairs before the user initiates the transaction by clicking `OK`.

But i can use the code from https://github.com/spesmilo/electrum/pull/10303 to fetch the pairs concurrently while already showing the Dialog. 
The `Additional fees:` label could show a `loading...` value and the `OK` button stays disabled until we got the pairs and know either the fees or can show an according warning label.

Then we could also show the swap provider selection button similar to https://github.com/spesmilo/electrum/pull/10303.

ecdsa:
yes, that sounds much better. 
if the option is enabled, we should display the swap provider selection button in the left bottom corner.
if no pair has been fetched yet, this button should just show "fetching swap offers", and the fee labels should also show a similar message.

I think this logic could be extended to the SwapDialog as well

f321x:
This is now updated to reuse the swap transport of the submarine payments tab and fetch the offers in the background.
The Onchain Payment tab now also shows the swapserver button when send change to lightning is enabled.

In case a swap would not be possible it will create change instead of preventing the transaction. I think this is less intrusive than blocking the creation of the tx and having the user toggle the config each time they try to create change that cannot be swapped. The dialog message clearly says that change will be created if the send change to ln option is enabled.

electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -1158,9 +1223,22 @@ def create_grid(self):
         return grid
 
     def _update_extra_fees(self):
```
ecdsa, line 1236: The 'additional fee' line is very unspecific. The GUI shows "additional fees" and the info line about a swap, but there is nothing that causally links them. I think this widget should not be used for that.

electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -712,13 +751,114 @@ def _update_message(self):
...
+            return
+        if self.ongoing_swap_transport_connection_attempt:
+            # another task is currently trying to connect
+            return
+
+        # there should only be a connected transport.
+        # a useless transport should get cleaned up and not stored.
+        assert self.swap_transport is None, "swap transport wasn't cleaned up properly"
+
+        new_swap_transport = self.main_window.create_sm_transport()
```
ecdsa, line 787: raises with non-ln wallet
```
  File "/opt/electrum/electrum/gui/qt/send_tab.py", line 592, in do_pay_or_get_invoice
    self.do_pay_invoice(self.pending_invoice)
  File "/opt/electrum/electrum/gui/qt/send_tab.py", line 619, in do_pay_invoice
    self.pay_onchain_dialog(invoice.outputs, invoice=invoice)
  File "/opt/electrum/electrum/gui/qt/send_tab.py", line 342, in pay_onchain_dialog
    tx, is_preview, paid_with_swap = self.window.confirm_tx_dialog(
  File "/opt/electrum/electrum/gui/qt/main_window.py", line 1528, in confirm_tx_dialog
    return d.run(), d.is_preview, d.did_swap
  File "/opt/electrum/electrum/gui/qt/confirm_tx_dialog.py", line 612, in run
    self.prepare_swap_transport()
  File "/opt/electrum/electrum/gui/qt/confirm_tx_dialog.py", line 778, in prepare_swap_transport
    new_swap_transport = self.main_window.create_sm_transport()
  File "/opt/electrum/electrum/gui/qt/main_window.py", line 1302, in create_sm_transport
    sm = self.wallet.lnworker.swap_manager
AttributeError: 'NoneType' object has no attribute 'swap_manager'

```

f321x: Fixed in https://github.com/spesmilo/electrum/pull/10247/commits/41b1a5bbd80feb4c68aab3d0d3cfd816b7f7fcb3

electrum/gui/qt/confirm_tx_dialog.py
f321x: Changed this in https://github.com/spesmilo/electrum/pull/10247/commits/21cc9532684ac6594867cb837c65fd4293ee192e. The additional fee is now shown in the message:
Screenshot_20260121_111002

electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -667,8 +682,38 @@ def get_messages(self):
...
             self.error = _('Fee estimates not available. Please set a fixed fee or feerate.')
-        if self.tx.get_dummy_output(DummyAddress.SWAP):
-            messages.append(_('This transaction will send funds to a submarine swap.'))
+        if dummy_output := self.tx.get_dummy_output(DummyAddress.SWAP):
+            swap_msg = _('This transaction will send funds to a submarine swap')
+            swap_fee_msg = "."
+            if self.swap_manager and self.swap_manager.is_initialized.is_set():
+                ln_amount_we_recv = self.swap_manager.get_recv_amount(send_amount=dummy_output.value, is_reverse=False)
+                if ln_amount_we_recv:
+                    swap_fee_msg = " [" + _("Additional fees:") + " " + self.main_window.format_amount_and_units(ln_amount_we_recv) + "]."
```
ecdsa: this is wrong: the LN amount we receive is not the fee.

f321x: Yes, fixed in https://github.com/spesmilo/electrum/compare/9ef2f3af1470a87f35bce633ea33878df11ffd57..2d97717e992dea4846bc08460ca896eabd62f7b8

PR #10266, opened 2025-10-09
Timelock Recovery Plugin - Prefer change-address for unused-address

### LINKED ISSUE #10263: Timelock Recovery Plugin using Ledger unnecessarily shows the address of the funds that remain on the same wallet.
### Description

Hi, I've tested the Timelock Recovery plugin using Ledger and noticed a strange behavior. The first transaction that the plugin generates, which is called Alert\Initiate transaction, sends 600 sats to the external addresses (to support CPFP in case of need), and the rest of the funds back to an address on the same wallet, labeled as "Timelock Recovery Alert Address". This address is derived from the main derivation path: m/84'/0'/a'/0/i.
When you sign this transaction, Ledger displays this address as if it was an external address, even though the funds remain on the same wallet. Ledger only "hides" an output address when it is derived from a change path: m/84'/0'/a'/1/i.
The risk is that users won't verify that the address really belongs to them, and some virus will replace the address with a similar one.

A simple fix would be to choose the "Alert Address" (and also the "Cancellation Address") from the change-path instead of the main-path. It's a one line fix in timelock_recovery.py of the line that calls `self.wallet.get_unused_addresses()`. but I'm not sure what I should replace it with. Should it be `self.wallet.get_change_addresses_for_new_transaction()`? With or without "allow_reusing_used_change_addrs"? Or maybe I should call `self.wallet.calc_unused_change_addresses()`?

oren-z0:
When sending funds back to the same wallet (in Alert Transaction and Cancellation Transaction), the destination address should preferabbly be derived from the change derivation path.

This makes things look better on hardware wallets like Ledger, which display main-derivation path addresses as if they where external addresses (thus forces the user to verify them).

Closes #10263

### DIFF
--- a/tests/test_timelock_recovery.py
+++ b/tests/test_timelock_recovery.py
@@ -44,7 +44,7 @@ class TestTimelockRecovery(ElectrumTestCase):
 
         context = TimelockRecoveryContext(wallet)
         alert_address = context.get_alert_address()
-        self.assertEqual(alert_address, 'tb1qchyc02y9mv4xths4je9puc4yzuxt8rfm26ef07')
+        self.assertEqual(alert_address, 'tb1q02g5nde0heaed0y24rztkedh9nvswknw50h7fx')
 
     async def test_get_cancellation_address(self):
         wallet = self._create_default_wallet()
@@ -52,7 +52,7 @@ class TestTimelockRecovery(ElectrumTestCase):
         context = TimelockRecoveryContext(wallet)
         context.get_alert_address()
         cancellation_address = context.get_cancellation_address()
-        self.assertEqual(cancellation_address, 'tb1q6k5h4cz6ra8nzhg90xm9wldvadgh0fpttfthcg')
+        self.assertEqual(cancellation_address, 'tb1qzyaz308030saay93zqma0at032vfqa9y0gfge3')
 
     async def test_make_unsigned_alert_tx(self):
         wallet = self._create_default_wallet()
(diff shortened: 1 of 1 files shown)


electrum/plugins/timelock_recovery/timelock_recovery.py
```diff
@@ -61,7 +65,7 @@ def _get_address_by_label(self, label: str) -> str:
                 self.wallet.set_label(addr, label)
                 return addr
         if self.wallet.is_deterministic():
-            addr = self.wallet.create_new_address(False)
+            addr = self.wallet.create_new_address(True) or self.wallet.create_new_address(False)
```
SomberNight, line 68: What is this trying to do? When exactly does the right-hand-side of the `or` execute?

electrum/plugins/timelock_recovery/timelock_recovery.py
```diff
@@ -52,7 +52,11 @@ def __init__(self, wallet: 'Abstract_Wallet'):
         self.wallet_name = str(self.wallet)
 
     def _get_address_by_label(self, label: str) -> str:
-        unused_addresses = list(self.wallet.get_unused_addresses())
+        unused_change_addresses = list(self.wallet.get_change_addresses_for_new_transaction(
+            allow_reusing_used_change_addrs=False,
+            max_change=self.wallet.gap_limit_for_change if self.wallet.gap_limit_for_change is not None else 2,
+        ))
+        unused_addresses = unused_change_addresses + list(self.wallet.get_unused_addresses())
```
SomberNight, line 59: This whole method is getting convoluted.
Would this work instead?

```suggestion
        unused_addresses = list(self.wallet._get_change_addresses_we_can_use_now()) + list(self.wallet.get_unused_addresses())
```

oren-z0: Since `_get_change_addresses_we_can_use_now` starts with a `_` I didn't consider the option to call it from outside.
Do you want me to rename it? Or maybe add a new function to Abstract_Wallet that will also do the safety checks:
```
def get_change_addresses_we_can_use_now(self, *, allow_reuse=True):
    if not self.use_change:
        return []
    change_addrs = self._get_change_addresses_we_can_use_now(allow_reuse=allow_reuse)
    for addr in change_addrs:
        assert is_address(addr), f"not valid bitcoin address: {addr}"
        self.check_address_for_corruption(addr)
    return change_addrs        
```

electrum/plugins/timelock_recovery/timelock_recovery.py
oren-z0: This line is just an edge case for when all the previous addresses are reserved or labelled for other purposes.
In this case I prefer to create a new change-address, but if `create_new_address(True)` returns `None` then it's fine to create a non-change address.

So yes, in `Deterministic_Wallet` the right-hand-side is never really executed because the left-hand-side always return a truthful value - but maybe in other wallet implementations (with `is_deterministic()` true) it will?

I can change this to:
```
addr = self.wallet.create_new_address(True)
if not addr:
    addr = self.wallet.create_new_address(False)
```

Or if you say that only `Deterministc_Wallet` will ever return true for `is_deterministic()`, then we can have only `addr = self.wallet.create_new_address(True)` as you suggested.

oren-z0:
Passing change-addresses isn't always solving the issue.
It seems Ledger requires to explicitly tell it what the change outputs are and what their exact derivation path.
Closing the PR to investigate.

PR #10272, opened 2025-10-20
timelock_recovery: recovery destination checks for address is_mine or script output

accumulator:
disallow recovery address(es) to be scripts, or address(es) in this wallet.


f321x:
tested, lgtm
I noticed a bunch of the (tooltip-)strings aren't translated, i can do a pr once this is merged.

PR #10277, opened 2025-10-22
qt: add text to toolbar button

f321x:
The toolbar button is very small and some users probably don't even notice it. As we have lots of space in the toolbar anyways i think it makes sense to add some text to it to make it more visible and easier to click. Maybe this is also useful for screen readers.


ecdsa:
the problem is, there is no text that will fit all contexts here

f321x:
Changed the button text to "Tools" in https://github.com/spesmilo/electrum/pull/10277/commits/c2068a4e4b7a081a68d09aa2912f32609a23fcce.

PR #10280, opened 2025-10-27
qt: fix ServerWidget (Network Tab)

f321x:
The ServerWidget was not working properly, when switching from "Manual Mode" to "Auto Connect" with an empty server field the change wouldn't get applied (but shown until the next restart) as it depended on having a correct server string entered (which isn't necessary for Auto Connect).
Also makes the widget behave more sane and predictable by cleaning the server input if Auto Connect is enabled and switching to Manual Mode if the user manually selects a server.

Updates the ServerWidget every time it is shown (on initialization and also when the user opens it again or switches between network dialog tabs).
This will clean it up if the user has entered some invalid server and closes it, otherwise this server would stay in the input field until the application is restarted and the widget might show something that doesn't actually represent the current configuration.

### DIFF
--- a/electrum/gui/qt/network_dialog.py
+++ b/electrum/gui/qt/network_dialog.py
@@ -435,7 +435,10 @@ class ServerWidget(QWidget, QtEventListener):
 
         def do_set_server(server):
             self.server_e.setText(server)
-            self.set_server()
+            if self.is_auto_connect():
+                # switch to manual mode as the user manually selected a server
+                self.set_connect_mode(ConnectMode.MANUAL, block_signals=True)
+            self.on_server_settings_changed()
         self.nodes_list_widget.setServer.connect(do_set_server)
 
         self.layout().addWidget(self.nodes_list_widget)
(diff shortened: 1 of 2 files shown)


electrum/gui/qt/network_dialog.py
```diff
@@ -458,17 +468,50 @@ def is_auto_connect(self):
...
+        if self.is_one_server() != current_net_params.oneserver:
+            settings_changed = True
+
+        if settings_changed:
+            _logger.debug(
+                f"ServerWidget.on_server_settings_changed:\n"
+                f"[{new_server=} ./. {current_net_params.server=}]\n"
+                f"[{self.is_auto_connect()=} ./. {current_net_params.auto_connect=}]\n"
+                f"[{self.is_one_server()=} ./. {current_net_params.oneserver=}]"
+            )
```
SomberNight, line 501: Somewhat strange notation. :P
I find something like this more readable (when logged):
```py
_logger.debug(
    f"ServerWidget.on_server_settings_changed:\n"
    f"[server: {current_net_params.server} -> {new_server}]\n"
    f"[auto_connect: {current_net_params.auto_connect} -> {self.is_auto_connect()}]\n"
    f"[oneserver: {current_net_params.oneserver} -> {self.is_one_server()}]"
)
```

f321x: ok, replaced in https://github.com/spesmilo/electrum/pull/10280/commits/f2aa4404ad580f4f095f6c296b2fbe1c9ccc06df

PR #10283, opened 2025-10-28
cli: fix add_peer command

f321x:
The timeout parameter of `add_peer` is unused and peer initialization was never awaited in the `add_peer` method, it just instantly returned True which is misleading.
This change marks `timeout` as deprecated as we always use `lnworker.LN_P2P_NETWORK_TIMEOUT` 
in LNPeer and awaits the initialization of the peer so that the caller actually knows if connection succeeded or
timed out. 
We could also keep the timeout parameter and wait only for this timeout but that would be potentially misleading too as then there would be two independent timeouts for each peer connection through the cli.

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1,3 +1,4 @@
+import asyncio
 import binascii
 import datetime
 import os.path
@@ -740,3 +741,38 @@ class TestCommandsTestnet(ElectrumTestCase):
         assert await cmds.export_lightning_preimage(payment_hash=payment_hash.hex(), wallet=w) == preimage.hex()
         assert await cmds.export_lightning_preimage(payment_hash=os.urandom(32).hex(), wallet=w) is None
 
+    @mock.patch.object(wallet.Abstract_Wallet, 'save_db')
+    @mock.patch('electrum.commands.LN_P2P_NETWORK_TIMEOUT', 0.001)
+    async def test_add_peer(self, *mock_args):
+        w = restore_wallet_from_text__for_unittest(
+            'disagree rug lemon bean unaware square alone beach tennis exhibit fix mimic',
+            path='if_this_exists_mocking_failed_648151893',
+            config=self.config)['wallet']
+        cmds = Commands(config=self.config)
+
+        # Mock the network and lnworker
+        mock_lnworker = mock.Mock()
+        w.lnworker = mock_lnworker
+        mock_peer = mock.Mock()
+        mock_peer.initialized = asyncio.Future()
+        connection_string = "test_node_id@127.0.0.1:9735"
+        called = False
+        async def lnworker_add_peer(*args, **kwargs):
+            assert args[0] == connection_string
+            nonlocal called
+            called += 1
+            return mock_peer
+        mock_lnworker.add_peer = lnworker_add_peer
+
+        # check if add_peer times out if peer doesn't initialize (LN_P2P_NETWORK_TIMEOUT is 0.001s)
+        with self.assertRaises(UserFacingException):
+            await cmds.add_peer(connection_string=connection_string, wallet=w)
+        # check if add_peer called lnworker.add_peer
+        assert called == 1
+
+        mock_peer.initialized = asyncio.Future()
+        mock_peer.initialized.set_result(True)
+        # check if add_peer returns True if peer is initialized
+        result = await cmds.add_peer(connection_string=connection_string, wallet=w)
+        assert called == 2
+        self.assertTrue(result)


electrum/commands.py
```diff
@@ -1681,10 +1682,12 @@ async def add_peer(self, connection_string, timeout=20, gossip=False, wallet: Ab
 
         arg:str:connection_string:Lightning network node ID or network address
         arg:bool:gossip:Apply command to your gossip node instead of wallet node
-        arg:int:timeout:Timeout in seconds (default=20)
+        arg:int:timeout:DEPRECATED [Timeout in seconds (default=20)]
```
SomberNight: > The timeout parameter of add_peer is unused and peer initialization was never awaited in the add_peer method, it just instantly returned True which is misleading.

The `timeout` parameter is not unused. It has some effect in some cases >.<
```py
  1.70 | E | __main__ | error running command (with daemon)
  File "/home/user/wspace/electrum/./run_electrum", line 548, in handle_cmd
    result = daemon.request(config, 'run_cmdline', (config_options,), timeout)
  File "/home/user/wspace/electrum/electrum/daemon.py", line 151, in request
    return fut.result(timeout=timeout)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/concurrent/futures/_base.py", line 458, in result
    raise TimeoutError()
TimeoutError
```

See comment for `lnpay`:
https://github.com/spesmilo/electrum/blob/d379a66a8f0bf8ad09ada6bfc5284ca3235b027d/electrum/commands.py#L1783-L1787

Anyway, I agree it "works" in a weird way atm.

electrum/commands.py
```diff
@@ -1681,10 +1682,12 @@ async def add_peer(self, connection_string, timeout=20, gossip=False, wallet: Ab
...
         """
+        # todo: remove timeout parameter in future version
         lnworker = self.network.lngossip if gossip else wallet.lnworker
-        await lnworker.add_peer(connection_string)
+        peer = await lnworker.add_peer(connection_string)
+        await util.wait_for2(peer.initialized, timeout=LN_P2P_NETWORK_TIMEOUT)
```
SomberNight: At least some exceptions are not really handled well.

Terminal 1, start daemon:
```
$ ./run_electrum --testnet daemon -v
```
Terminal 2:
```
$ ./run_electrum --testnet -v load_wallet -w /home/user/.electrum/testnet/wallets/9dk
$ ./run_electrum --testnet add_peer "02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa@localhost:1111" --timeout 100 --gossip
(inside daemon): Traceback (most recent call last):
  File "/home/user/wspace/electrum/electrum/daemon.py", line 268, in handle
    response['result'] = await f(*params)
                         ^^^^^^^^^^^^^^^^
  File "/home/user/wspace/electrum/electrum/daemon.py", line 381, in run_cmdline
    result = await func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/wspace/electrum/electrum/commands.py", line 203, in func_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/wspace/electrum/electrum/commands.py", line 1690, in add_peer
    await util.wait_for2(peer.initialized, timeout=LN_P2P_NETWORK_TIMEOUT)
  File "/home/user/wspace/electrum/electrum/util.py", line 1494, in wait_for2
    return await asyncio.wait_for(fut, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/asyncio/tasks.py", line 507, in wait_for
    return await fut
           ^^^^^^^^^
asyncio.exceptions.CancelledError

internal error while executing RPC
```

Log from daemon:
```
122.80 | I | lnworker.LNGossip | adding peer 02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa@localhost:1111
122.80 | E | daemon.CommandsServer | internal error while executing RPC
  File "/home/user/wspace/electrum/electrum/daemon.py", line 268, in handle
    response['result'] = await f(*params)
                         ^^^^^^^^^^^^^^^^
  File "/home/user/wspace/electrum/electrum/daemon.py", line 381, in run_cmdline
    result = await func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/wspace/electrum/electrum/commands.py", line 203, in func_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/wspace/electrum/electrum/commands.py", line 1690, in add_peer
    await util.wait_for2(peer.initialized, timeout=LN_P2P_NETWORK_TIMEOUT)
  File "/home/user/wspace/electrum/electrum/util.py", line 1494, in wait_for2
    return await asyncio.wait_for(fut, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/asyncio/tasks.py", line 507, in wait_for
    return await fut
           ^^^^^^^^^
asyncio.exceptions.CancelledError
122.80 | I | lnpeer.Peer.[LNGossip, 02aaaaaaaa-c12755eb] | Disconnecting: GracefulDisconnect('initialize failed: OSError("Multiple exceptions: [Errno 111] Connect call failed (\'::1\', 1111, 0, 0), [Errno 111] Connect call failed (\'127.0.0.1\', 1111)")')
```

electrum/commands.py
f321x: I misunderstood this, thought it is some leftover parameter. Reverted the change back in https://github.com/spesmilo/electrum/pull/10283/commits/d83314c0756e5e8f2a1ece8eb01237a1407b5b47.

electrum/commands.py
f321x: Changed it in https://github.com/spesmilo/electrum/pull/10283/commits/d83314c0756e5e8f2a1ece8eb01237a1407b5b47 so it always returns a `UserFacingException` instead of letting the original exception go out uncaught.

electrum/commands.py
SomberNight: Ok. Nevertheless, fixing that timeout param to work properly in all cases or to make it more obvious how it works is on the (low-priority?) wishlist :D

electrum/commands.py
```diff
@@ -1684,7 +1685,11 @@ async def add_peer(self, connection_string, timeout=20, gossip=False, wallet: Ab
         arg:int:timeout:Timeout in seconds (default=20)
         """
         lnworker = self.network.lngossip if gossip else wallet.lnworker
-        await lnworker.add_peer(connection_string)
+        peer = await lnworker.add_peer(connection_string)
+        try:
+            await util.wait_for2(peer.initialized, timeout=LN_P2P_NETWORK_TIMEOUT)
+        except BaseException as e:
```
SomberNight: `BaseException` is too extreme I think, should be used sparingly.
I guess we could just catch `Exception`.  Unless you have a scenario in mind that would not be covered.
(see https://docs.python.org/3/library/exceptions.html#exception-hierarchy )

SomberNight: maybe add a comment:
```suggestion
        except BaseException as e:  # FIXME often simply CancelledError and real cause (e.g. timeout) remains hidden
```

f321x: I used `BaseException` because `CancelledError` does inherit directly from `BaseException`. However yeah it seems cleaner to just catch both `CancelledError` and `Exception` explicitly. Changed that in https://github.com/spesmilo/electrum/pull/10283/commits/95ba7e7547eb7d1a016793d363dd71410eacab76

f321x: Added in https://github.com/spesmilo/electrum/pull/10283/commits/95ba7e7547eb7d1a016793d363dd71410eacab76

PR #10284, opened 2025-10-28
android: qr scanner: add emoji to clipboard string

f321x:
The string is not translated and might not be accessible for non-english speakers, even though its relatively self-explanatory i think adding this emoji makes it more accessible while still looking clean.

signal-2025-10-28-143345


SomberNight:
Sounds reasonable.

PR #10287, opened 2025-10-29
qt: allow opening BalanceDialog if warning is set and balance is 0

f321x:
Allow to open the BalanceDialog, by clicking on the warning icon in the bottom left corner of the main window, if there is an active warning, even if the wallets balance is 0. 
Right now the user can see the warning icon for the lighting channel reserve, but cannot click on the icon to read the actual warning if there is no balance in the wallet.

Screenshot_20251029_105025


SomberNight:
note: kinda hard to end up in a situation where the user has 0 balance but has a warning about anchor channels. Means no UTXOs and a freshly opened incoming channel.

PR #10300, opened 2025-11-04
qt: followup #10277: change icons depending on context

f321x:
Changes the icons of the "Tools" button to the same icon as the tab in which the button is located to make it visible that the available tools are dependent on the buttons context.

Screenshot_20251104_142408
Screenshot_20251104_142350

### DIFF
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -125,7 +125,7 @@ class AddressList(MyTreeView):
         self.main_window.show_address(addr)
 
     def create_toolbar(self, config: 'SimpleConfig'):
-        toolbar, menu = self.create_toolbar_with_menu('')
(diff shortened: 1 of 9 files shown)


SomberNight:
I guess it is subjective but I mildly prefer the "preferences" icon. Though perhaps it's just what I got "used to".

> to make it visible that the available tools are dependent on the buttons context

Well, yes, good point...

idk, no strong preference either way.

f321x:
Yeah i also feel like the "preferences" icon approach looks a bit cleaner. 
@ecdsa you wanted to test this style, do you have any preferences?

ecdsa:
I wanted a visual hint that the content of the drop-down is not the same in every tab.
I think I prefer the way it is in this PR, although it is not perfect.

ecdsa:
Note: ideally, I want the toolbar to be configurable: It would be possible to drag an icon from the drop-down to the toolbar, and it would add a direct button to the toolbar.

PR #10230, opened 2025-09-19
wip: refactor htlc switch

f321x:
See commit messages

wip, todo: 
* more unittests
* testing wallet db changes

PR #10303, opened 2025-11-11
swaps/qt: expose swaps to external address

f321x:
Allows to do reverse submarine swaps to an external address and exposes this as "Submarine Payment" tab in the `ConfirmTxDialog`.

Does a wallet upgrade conflicting with https://github.com/spesmilo/electrum/pull/10230, if switching between these branches the db needs to be manually edited.

### DIFF
--- a/electrum/gui/qt/confirm_tx_dialog.py
+++ b/electrum/gui/qt/confirm_tx_dialog.py
@@ -47,7 +47,8 @@ from electrum.submarine_swaps import NostrTransport, HttpTransport
 from .seed_dialog import seed_warning_msg
 
 from .util import (WindowModalDialog, ColorScheme, HelpLabel, Buttons, CancelButton, WWLabel,
-                   read_QIcon, debug_widget_layouts, qt_event_listener, QtEventListener, IconLabel)
+                   read_QIcon, debug_widget_layouts, qt_event_listener, QtEventListener, IconLabel,
+                   pubkey_to_q_icon)
 from .transaction_dialog import TxSizeLabel, TxFiatLabel, TxInOutWidget
 from .fee_slider import FeeSlider, FeeComboBox
 from .amountedit import FeerateEdit, BTCAmountEdit
@@ -885,7 +886,7 @@ class TxEditor(WindowModalDialog, QtEventListener, Logger):
             self.server_button.setEnabled(True)
             if self.config.SWAPSERVER_NPUB:
                 pubkey = from_nip19(self.config.SWAPSERVER_NPUB)['object'].hex()
-                self.server_button.setIcon(SwapServerDialog.pubkey_to_q_icon(pubkey))
+                self.server_button.setIcon(pubkey_to_q_icon(pubkey))
             self.server_button.setText(
                 f' {len(self.swap_transport.get_recent_offers())} ' +
                 (_('providers') if len(self.swap_transport.get_recent_offers()) != 1 else _('provider'))
--- a/electrum/gui/qt/swap_dialog.py
+++ b/electrum/gui/qt/swap_dialog.py
@@ -13,12 +13,13 @@ from electrum.util import NotEnoughFunds, NoDynamicFeeEstimates, UserCancelled
 from electrum.bitcoin import DummyAddress
 from electrum.transaction import PartialTxOutput, PartialTransaction
 from electrum.fee_policy import FeePolicy
-from electrum.submarine_swaps import NostrTransport, pubkey_to_rgb_color
+from electrum.submarine_swaps import NostrTransport
 
 from electrum.gui import messages
 from . import util
 from .util import (WindowModalDialog, Buttons, OkButton, CancelButton,
-                   EnterButton, ColorScheme, WWLabel, read_QIcon, IconLabel, char_width_in_lineedit)
+                   EnterButton, ColorScheme, WWLabel, read_QIcon, IconLabel, char_width_in_lineedit,
+                   pubkey_to_q_icon)
 from .util import qt_event_listener, QtEventListener
 from .amountedit import BTCAmountEdit
 from .fee_slider import FeeSlider, FeeComboBox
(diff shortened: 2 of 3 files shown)


f321x:
Rebased, now shows the onchain and lightning icons in the dialog:

image

f321x:
Now showing the pay-to-many warning before fetching the swap pairs and showing the channel liquidity warning simultaneously with the swap provider liquidity warning if it is clear that the channels have too low liquidity.

electrum/gui/qt/send_tab.py
```diff
@@ -338,9 +338,24 @@ def make_tx(fee_policy, *, confirmed_only=False, base_tx=None):
...
         candidates = self.wallet.get_candidates_for_batching(outputs, coins=coins_conservative)
 
-        tx, is_preview = self.window.confirm_tx_dialog(make_tx, output_value, batching_candidates=candidates)
+        submarine_payment_payee_outputs = [o for o in outputs if not o.is_change]
+        if is_max and submarine_payment_payee_outputs:
+            # replace '!' max value with amount_e int value as the swap logic should use the same
+            # amount that is shown here in the send tab. It can be set on the first value as
+            # submarine payments are only possible with a single TxOutput and will err if
+            # there are more outputs anyway.
+            submarine_payment_payee_outputs[0].value = int(self.amount_e.get_amount() or 0)
```
ecdsa: maybe this should be disabled? I don't see why it makes sense

electrum/gui/qt/swap_dialog.py
```diff
@@ -531,12 +531,12 @@ def update_servers_list(self, servers: Sequence['SwapOffer']):
...
+            item.setIcon(self.Columns.PUBKEY, self.pubkey_to_q_icon(x.server_pubkey))
             items.append(item)
         self.servers_list.insertTopLevelItems(0, items)
 
     @staticmethod
-    def _pubkey_to_q_icon(server_pubkey: str) -> QIcon:
+    def pubkey_to_q_icon(server_pubkey: str) -> QIcon:
         color = QColor(*pubkey_to_rgb_color(server_pubkey))
         color_pixmap = QPixmap(100, 100)
         color_pixmap.fill(color)
```
ecdsa: if this becomes public, this method could be moved to lnutil

electrum/submarine_swaps.py
```diff
@@ -199,7 +200,7 @@ class SwapData(StoredObject):
     prepay_hash = attr.ib(type=Optional[bytes], converter=hex_to_bytes)
     privkey = attr.ib(type=bytes, converter=hex_to_bytes)
     lockup_address = attr.ib(type=str)
-    receive_address = attr.ib(type=str)
```
ecdsa, line 202: so that was not used? :-)

ecdsa: probably a leftover from before adding txbatcher

electrum/gui/qt/swap_dialog.py
f321x: I moved it to qt/utils.py in  https://github.com/spesmilo/electrum/pull/10303/commits/3b80795e9c8adffc3b24a128ccb2fe6d00a5d239

electrum/gui/qt/send_tab.py
f321x: Yeah it's not especially useful, instead i now disable the tab if it is a Max spend or Pay-to-many and show an according tooltip

electrum/submarine_swaps.py
```diff
@@ -540,6 +544,42 @@ async def _claim_swap(self, swap: SwapData) -> None:
...
+        """
+        Construct claim tx that spends exactly the funding utxo to the swap output, independent of the
+        current fee environment to guarantee the correct amount is being sent to the claim output which
+        might be an external address.
+        """
+        assert swap.claim_to_output, swap
+        txout = PartialTxOutput.from_address_and_value(swap.claim_to_output[0], swap.claim_to_output[1])
+        tx = PartialTransaction.from_io([claim_txin], [txout])
+        can_be_broadcast = self.wallet.adb.get_tx_height(swap.funding_txid).height() > 0
+        already_broadcast = self.wallet.adb.get_tx_height(tx.txid()).height() >= 0
```
SomberNight, line 557: This comparison is a bit confusing. `0` and `-1` should probably be treated the same.
https://github.com/spesmilo/electrum/blob/bc2bd485167ab5335bc577034f688f5a8d4770da/electrum/address_synchronizer.py#L50-L53

f321x: Ok, added a followup pr https://github.com/spesmilo/electrum/pull/10327 to make this a bit clearer

electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -593,17 +686,343 @@ def _update_send_button(self):
...
+        normal_page = QWidget()
+        h = QGridLayout(normal_page)
+        self.submarine_lightning_send_amount_label = QLabel()
+        self.submarine_onchain_send_amount_label = QLabel()
+        self.submarine_claim_mining_fee_label = QLabel()
+        self.submarine_server_fee_label = QLabel()
+        self.submarine_we_send_label = IconLabel(text=_('You send')+':')
+        self.submarine_we_send_label.setIcon(read_QIcon('lightning.png'))
+        self.submarine_they_receive_label = IconLabel(text=_('They receive')+':')
+        self.submarine_they_receive_label.setIcon(read_QIcon('bitcoin.png'))
```
SomberNight, line 709: I almost tried to use a submarine payment to pay a btcpayserver onchain invoice that had a 15 minute expiry.
btcpayserver expects to see a tx paying to that onchain address, even if in the mempool, before that expiry.
I don't know if 15 minutes is the default (?).

When using a submarine payment, the claim tx that the merchant is waiting for only gets broadcast after the funding tx gets mined, right? So it is quite risky to try to use it for short expiry invoices.
I guess if the expiry is missed but the tx gets mined, the user might have to talk to human customer support.

Should we add some short text about how these submarine payments work, with all this in mind, perhaps to this layout? We could tell the user that the tx the merchant is waiting for will only be broadcast after the swap funding tx is mined, which might take 10-30+ mins.

f321x: > When using a submarine payment, the claim tx that the merchant is waiting for only gets broadcast after the funding tx gets mined, right?

yes

I added a commit to https://github.com/spesmilo/electrum/pull/10366 which adds a `HelpButton` to the tab so the user can better understand this function and the risk you mentioned.

SomberNight: Great! Thank you.

Issue #10315, opened 2025-11-19
Feerate flag of payto command broken on 4.6.x


Keeqler:
### Description

When using:

```
electrum payto <address> <amount> --feerate "feerate:1234"
```

the serialized tx is returned, but the specified fee rate is not used since the fee policy descriptor fails to be parsed, as shown in the daemon logs:

```
36.01 | W | fee_policy.FeePolicy | Could not parse fee policy descriptor 'feerate:feerate:1234feerate:1234feerate:1234feerate:1234... Falling back to 'eta:2'
```

Specifying `eta:3` for example fails to parse as well:

```
674.55 | W | fee_policy.FeePolicy | Could not parse fee policy descriptor 'feerate:eta:3eta:3eta:3eta:3eta... Falling back to 'eta:2'
```

f321x:
Thanks for the report, there was a bug that will be fixed with https://github.com/spesmilo/electrum/pull/10317.

However please note the way you are using `--feerate` is not supported either way. `--feerate` is supposed to receive a sat/vbyte decimal value, not a eta target or text string like `feerate:` in your example.

So if you want to pay with e.g. 5.5 sat/vb this will work:
```bash
$ electrum payto <address> <amount> --feerate "5.5"
```

If you want to use estimates, like e.g. `eta:25` (~next 25 blocks/250 minutes), you can set the default fee policy using setconfig:
```bash
$ electrum setconfig fee_policy.default eta:25
```
and then use `payto` without any fee argument, it will automatically use the default fee policy:
```
$ electrum payto <address> <amount>
```

Keeqler:
Awesome! Thanks for working on the fix so quickly. And thanks for the heads up.

PR #10317, opened 2025-11-19
cli: payto: fix feerate parsing

f321x:
Feerate is passed to `Commands._get_fee_policy()` as str which then tried to multiply the string by 1000. Now it first casts the string to `Decimal` and multiplies the decimal.
Also changes the `fee` and `feerate` types in the docstrings of `payto`, `paytomany` and `sweep` from str to decimal, so entering something like `eta:...` will fail right away.

Fixes https://github.com/spesmilo/electrum/issues/10315


SomberNight:
Thanks!

PR #10319, opened 2025-11-24
lnworker: use config lightning fee for estimate

f321x:
I was unable to do a "Max" amount submarine swap because the
`fee_estimate` method used by `LNWallet.num_sats_can_send()` uses a
hardcoded `fee_proportional_millionths` to estimate the fee for the
lightning payment.
When the actual fee determined later is higher
than the estimated fee the payment fails as the channel is unable to add
the htlc sum including the real fees as the amount exceeds the balance of
the channel.
Using the fees the maximum fees user has configured and estimate the
potential fee as inverse of PaymentFeeBudget is more
reliable/conservative as we definitely aren't going to pay more fees
than this amount.

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -2720,6 +2720,10 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger, QtEventListener):
         run_hook('close_settings_dialog')
         if d.need_restart:
             self.show_warning(_('Please restart Electrum to activate the new GUI settings'), title=_('Success'))
+        else:
+            # Some values might need to be updated if settings have changed.
+            # For example 'Can send' in the lightning tab will change if the fees config is changed.
+            self.refresh_tabs()
 
     def _show_closing_warnings(self) -> bool:
         """Show any closing warnings and return True if the user chose to quit anyway."""


electrum/lnworker.py
```diff
@@ -2881,7 +2881,7 @@ def fee_estimate(self, amount_sat):
         # Here we have to guess a fee, because some callers (submarine swaps)
         # use this method to initiate a payment, which would otherwise fail.
         fee_base_msat = 5000               # FIXME ehh.. there ought to be a better way...
-        fee_proportional_millionths = 500  # FIXME
+        fee_proportional_millionths = self.config.LIGHTNING_PAYMENT_FEE_MAX_MILLIONTHS
         # inverse of fee_for_edge_msat
```
SomberNight: >  Using SimpleConfig.LIGHTNING_PAYMENT_FEE_MAX_MILLIONTHS seems more reliable here as it will always use the possible maximum fee the user is willing to pay for the fee estimate.

Well, with your current diff, you are estimating with `fee_base_msat = 5000` plus `config.LIGHTNING_PAYMENT_FEE_MAX_MILLIONTHS`. So for large payments, because of that 5 sat base fee, you are overestimating a bit.
For small payments, because of the cutoff, you are underestimating.

What I mean is that `lnworker.fee_estimate` is currently inverting:
https://github.com/spesmilo/electrum/blob/530260638a48b94e89786720291b964c7288aed6/electrum/lnrouter.py#L59-L61

but maybe what you are trying to do - if you just want to assume max fees - is to instead invert:
https://github.com/spesmilo/electrum/blob/530260638a48b94e89786720291b964c7288aed6/electrum/lnutil.py#L2020-L2023

electrum/lnutil.py
```diff
@@ -2022,3 +2023,28 @@ def _calculate_fee_msat(
...
+        )
+        cutoff_clamped = min(
+            max(0, config.LIGHTNING_PAYMENT_FEE_CUTOFF_MSAT),
+            cls.PAYMENT_FEE_CUTOFF_CLAMP,
+        )
+
+        # inverse of _calculate_fee_msat
+        amount_minus_fees = (total_amount_msat * 1_000_000) // (1_000_000 + millionths_clamped)
+        fees_msat = max(total_amount_msat - amount_minus_fees, cutoff_clamped)
+        return fees_msat
```
SomberNight, line 2111: note that
`10000 == PaymentFeeBudget.reverse_from_total_amount(total_amount_msat=1, config=config)`
```suggestion
        amount_minus_fees = (total_amount_msat * 1_000_000) // (1_000_000 + millionths_clamped)
        fees_msat = max(total_amount_msat - amount_minus_fees, cutoff_clamped)
        fees_msat = min(fees_msat, total_amount_msat)  # to handle (invalid?) inputs below cutoff_clamped
        return fees_msat
```

electrum/lnutil.py
```diff
@@ -2022,3 +2023,28 @@ def _calculate_fee_msat(
...
+    @classmethod
+    def reverse_from_total_amount(cls, total_amount_msat: int, config: 'SimpleConfig') -> int:
+        """
+        Given the total amount (including fees) return the amount of fees
+        included assuming highest allowed from config fees are being used.
+
+        This allows to guess a fee that has to be reserved to reliably allow
+        doing a "Max" amount lightning send (e.g. for submarine swaps).
+        """
+        assert isinstance(total_amount_msat, int), type(total_amount_msat)
```
SomberNight: I guess  in the same spirit, we could do `assert total_amount_msat >= 0`

electrum/lnutil.py
```diff
@@ -2022,3 +2023,28 @@ def _calculate_fee_msat(
         fee_msat = invoice_amount_msat * millionths_clamped // 1_000_000
         fee_msat = max(fee_msat, cutoff_clamped)
         return fee_msat
+
+    @classmethod
+    def reverse_from_total_amount(cls, total_amount_msat: int, config: 'SimpleConfig') -> int:
```
SomberNight: nit: at least `config` should be a kwarg.
In my view, every arg to a function - that is not obvious just from reading the name of the function - should be a kwarg.

electrum/lnworker.py
```diff
@@ -2884,15 +2885,10 @@ def get_channels_for_sending(self):
...
-    def fee_estimate(self, amount_sat):
-        # Here we have to guess a fee, because some callers (submarine swaps)
-        # use this method to initiate a payment, which would otherwise fail.
-        fee_base_msat = 5000               # FIXME ehh.. there ought to be a better way...
-        fee_proportional_millionths = 500  # FIXME
-        # inverse of fee_for_edge_msat
-        amount_msat = amount_sat * 1000
-        amount_minus_fees = (amount_msat - fee_base_msat) * 1_000_000 // ( 1_000_000 + fee_proportional_millionths)
-        return Decimal(amount_msat - amount_minus_fees) / 1000
+    def fee_estimate(self, amount_sat: int | Decimal) -> int:
```
SomberNight: btw I think this function is badly named. Hard to guess from the name what it is doing.
but that's a pre-existing issue, so nvm.

f321x: Yeah I agree it was a bit opaque what it is actually doing, I renamed it to `estimate_fee_reserve_for_total_amount`.

electrum/lnutil.py
f321x: Made `reverse_from_total_amount` kwarg only so its interface is similar to the other methods of `PaymentFeeBudget`.

electrum/lnutil.py
f321x: Good idea, added this check

electrum/lnutil.py
f321x: Yes but not sure why we wouldn't want this? When we later try to pay 1 msat the fee budget will also be the cutoff value.
Edit: yeah ok i see the point, probably doesn't make sense to allocate more fees than what we can actually send.
Added this line in https://github.com/spesmilo/electrum/pull/10319/commits/4c53dbeb99fef68d3e2e1bc78055d2db69354b04

electrum/lnworker.py
f321x: > but maybe what you are trying to do - if you just want to assume max fees - is to instead invert:

Doing this now in the latest commits

electrum/lnworker.py
```diff
@@ -2950,15 +2951,17 @@ def get_channels_for_sending(self):
...
+        """
+        Estimate how much of the given amount needs to be reserved for
+        ln payment fees to reliably pay the remaining amount.
+        """
+        amount_msat = ceil(amount_sat * 1000)  # round up to the next sat
+        fee_msat = PaymentFeeBudget.reverse_from_total_amount(
+            total_amount_msat=amount_msat,
+            config=self.config,
+        )
+        return ceil(fee_msat / 1000)
```
SomberNight: floatsss
```py
>>> math.ceil((2*10**16+1)/1000)
20000000000000
```
EDIT:
```py
>>> from decimal import Decimal
>>> math.ceil(Decimal(2*10**16+1)/1000)
20000000000001
```

SomberNight:
```py
(a // b) + (a % b > 0)
```

f321x: Changed it to use Decimal in https://github.com/spesmilo/electrum/pull/10319/commits/af4dc24d87424f97e0661f1c10911d8ef2ea4ce5

SomberNight:
Thanks!

PR #10321, opened 2025-11-25
qml millisat precision fix

### LINKED ISSUE #10320: QML GUI cannot pay lightning invoices with millisat precision
### Description

Phoenix creates invoices with millisat precision.
The following checks make it impossible to pay:

```diff --git a/electrum/gui/qml/qeinvoice.py b/electrum/gui/qml/qeinvoice.py
index acb80bb81..77d8cccae 100644
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -331,7 +331,7 @@ class QEInvoice(QObject, QtEventListener):
             if status in [PR_UNPAID, PR_FAILED]:
                 if self.get_max_spendable_lightning() >= amount.satsInt:
                     lnaddr = self._effectiveInvoice._lnaddr
-                    if lnaddr.amount and amount.satsInt < lnaddr.amount * COIN:
+                    if lnaddr.amount and amount.satsInt < int(lnaddr.amount * COIN):
                         self.userinfo = _('Cannot pay less than the amount specified in the invoice')
                 elif not self.address or self.get_max_spendable_onchain() < amount.satsInt:
                     # TODO: for onchain: validate address? subtract fee?
@@ -368,7 +368,7 @@ class QEInvoice(QObject, QtEventListener):
             if status in [PR_UNPAID, PR_FAILED]:
                 if self.get_max_spendable_lightning() >= amount.satsInt:
                     lnaddr = self._effectiveInvoice._lnaddr
-                    if not (lnaddr.amount and amount.satsInt < lnaddr.amount * COIN):
+                    if not (lnaddr.amount and amount.satsInt < int(lnaddr.amount * COIN)):
                         self.canPay = True
                 elif self.address and self.get_max_spendable_onchain() > amount.satsInt:
                     # TODO: validate address?

```

what annoys me is that the check is replicated in two places. this needs to be refactored.

accumulator:
should fix #10320

### DIFF
--- a/electrum/gui/qml/components/InvoiceDialog.qml
+++ b/electrum/gui/qml/components/InvoiceDialog.qml
@@ -181,7 +181,9 @@ ElDialog {
                                 font.pixelSize: constants.fontSizeXLarge
                                 font.family: FixedFont
                                 font.bold: true
-                                text: Config.formatSats(invoice.amount, false)
+                                text: invoice.invoiceType == Invoice.LightningInvoice
+                                    ? Config.formatMilliSats(invoice.amount, false)
+                                    : Config.formatSats(invoice.amount, false)
                             }
 
                             Label {
@@ -223,12 +225,13 @@ ElDialog {
                                 Layout.preferredWidth: amountFontMetrics.advanceWidth('0') * 14 + leftPadding + rightPadding
                                 fiatfield: amountFiat
                                 readOnly: amountMax.checked
+                                msatPrecision: invoice.invoiceType == Invoice.LightningInvoice
                                 color: readOnly
                                     ? Material.accentColor
                                     : Material.foreground
                                 onTextAsSatsChanged: {
                                     if (!amountMax.checked)
-                                        invoice.amountOverride.satsInt = textAsSats.satsInt
+                                        invoice.amountOverride.copyFrom(textAsSats)
                                 }
                                 Connections {
                                     target: invoice.amountOverride
--- a/electrum/gui/qml/components/controls/BtcField.qml
+++ b/electrum/gui/qml/components/controls/BtcField.qml
@@ -7,12 +7,13 @@ TextField {
     id: amount
 
     required property TextField fiatfield
+    property bool msatPrecision: false
 
     font.family: FixedFont
     placeholderText: qsTr('Amount')
     inputMethodHints: Qt.ImhDigitsOnly
     validator: RegularExpressionValidator {
-        regularExpression: Config.btcAmountRegex
+        regularExpression: msatPrecision ? Config.btcAmountRegexMsat : Config.btcAmountRegex
     }
 
     property Amount textAsSats
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -383,7 +383,7 @@ class QEInvoice(QObject, QtEventListener):
         if self.amount.isEmpty:
             if self.amountOverride.isEmpty:
                 raise Exception('can not pay 0 amount')
-            amount_msat = self.amountOverride.satsInt * 1000
+            amount_msat = self.amountOverride.msatsInt
 
         self._wallet.pay_lightning_invoice(self._effectiveInvoice, amount_msat)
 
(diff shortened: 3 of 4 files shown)


electrum/gui/qml/qeconfig.py
```diff
@@ -373,18 +374,13 @@ def unitsToSats(self, unitAmount):
...
-            return self._amount
-        self._logger.debug('fallthrough')
-        # otherwise, scale it back to the expected unit
-        #amount = Decimal(max_prec_amount) / Decimal(pow(10, self.max_precision()-self.decimal_point()))
-        #return int(amount) #Decimal(amount) if not self.is_int else int(amount)
+        sat_max_precision = self.config.BTC_AMOUNTS_DECIMAL_POINT
+        msat_max_precision = self.config.BTC_AMOUNTS_DECIMAL_POINT + 3
+        sat_max_prec_amount = int(pow(10, sat_max_precision) * x)
+        msat_max_prec_amount = int(pow(10, msat_max_precision) * x)
+        self._amount = QEAmount(amount_sat=sat_max_prec_amount, amount_msat=msat_max_prec_amount)
```
SomberNight, line 381: note: I find it quite annoying and error-prone that QEAmount's amount_sat and amount_msat are not guaranteed to be in sync. Internally it should only have a single integer to store the amount, and getters could expose sat vs msat.
But ok, this is a pre-existing issue on master.

electrum/gui/qml/qeconfig.py
```diff
@@ -358,13 +366,6 @@ def formatMilliSats(self, amount, with_unit=False):
         else:
```
SomberNight, line 366: side-note: what's up with this fallback return value?
https://github.com/spesmilo/electrum/blob/d7bc617034377119523709c504edb1205eaae34c/electrum/gui/qml/qeconfig.py#L353-L354

electrum/gui/qml/components/InvoiceDialog.qml
```diff
@@ -223,12 +223,13 @@ ElDialog {
                                 Layout.preferredWidth: amountFontMetrics.advanceWidth('0') * 14 + leftPadding + rightPadding
                                 fiatfield: amountFiat
                                 readOnly: amountMax.checked
+                                msatPrecision: invoice.invoiceType == Invoice.LightningInvoice
```
SomberNight, line 228: The web ui does not let me comment on the line I actually want to:
https://github.com/spesmilo/electrum/blob/4a63916d5ee28392f74f2bcac5b3c27f8a51b59c/electrum/gui/qml/components/InvoiceDialog.qml#L184

Shouldn't that show text at max precision? e.g. `text: Config.formatMilliSats(invoice.amount, false)`

As-is, if you scan an invoice that has msat precision, we currently don't show the user the msat part, but we do pay it if they click pay. It feels wrong to send out more money than what we display beforehand.
but it's only sub-satoshi stuff, so ultimately we could ignore it

SomberNight (commented):
Looks good. Some minor comments.

electrum/gui/qml/components/InvoiceDialog.qml
accumulator: You're right. fixed.

electrum/gui/qml/qeconfig.py
accumulator: > note: I find it quite annoying and error-prone that QEAmount's amount_sat and amount_msat are not guaranteed to be in sync. Internally it should only have a single integer to store the amount, and getters could expose sat vs msat. But ok, this is a pre-existing issue on master.

It is very annoying. I don't remember exactly why, but early on (Qt5) it was causing problems with QML property update events, so this was the 'temporary' solution. Might be good to revisit this.

electrum/gui/qml/qeconfig.py
accumulator: This was for visual confirmation that everywhere a `QEAmount` is passed to this func. I'm quite sure this can be removed :)

SomberNight:
Thanks for this.

electrum/gui/qml/qeconfig.py
SomberNight: done in https://github.com/spesmilo/electrum/commit/bc2bd485167ab5335bc577034f688f5a8d4770da

PR #10325, opened 2025-11-27
lightning: fix race when doing concurrent payments

f321x:
There is a race when initiating multiple lightning payments concurrently
(e.g. when doing a reverse swap with prepayment + swap payment).
suggest_splits might overallocate
split amounts for a channel as the splitting of both invoice amounts runs
concurrently and before acutal htlcs that reduce the channels balance
have been added to the channel yet. This results in a "not
enough balance" PaymentFailure once we try to send the htlcs and the
other payment attempt already reduced the available balance of the
channel.

This fix takes a lock from splitting the amount until the htlcs are
put on the channel, so suggest_splits always acts on the correct channel
balance.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -914,6 +914,7 @@ class LNWallet(LNWorker):
         self._paysessions = dict()                      # type: Dict[bytes, PaySession]
         self.sent_htlcs_info = dict()                   # type: Dict[SentHtlcKey, SentHtlcInfo]
         self.received_mpp_htlcs = self.db.get_dict('received_mpp_htlcs')   # type: Dict[str, ReceivedMPPStatus]  # payment_key -> ReceivedMPPStatus
+        self._channel_sending_capacity_lock = asyncio.Lock()
 
         # detect inflight payments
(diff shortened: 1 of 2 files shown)


f321x:
Rebased, now uses an async lock instead of the sleep.

electrum/lnworker.py
```diff
@@ -1698,27 +1699,33 @@ async def pay_to_node(
...
-                        budget=budget._replace(fee_msat=remaining_fee_budget_msat),
-                    )
-                    # 2. send htlcs
-                    async for sent_htlc_info, cltv_delta, trampoline_onion in routes:
-                        await self.pay_to_route(
+                    # splitting the amount of the payment between our channels requires the correct
+                    # available channel balance. to prevent concurrent splitting attempts from
+                    # using stale channel balances for the split calculation a lock needs to be
+                    # taken until the htlcs are added to the channel so the next splitting attempt
+                    # acts on a correct channel balance.
```
SomberNight, line 1707: btw it's not just the balance, a race here also e.g. affects the `max_accepted_htlcs` (number of htlc slots).
but this comment is already long.

SomberNight:
Thanks for the fix.

PR #10327, opened 2025-11-27
swaps: followup 10303

f321x:
small followup replacing tx height integers with const variables and considering claim tx broadcast too if there is an unconfirmed parent (height -1).


SomberNight:
thx

PR #10328, opened 2025-11-28
qt: SettingsDialog: update ln fee config on slider moved

f321x:
When moving the lightning fee slider in the SettingsDialog by clicking on the range instead of pulling the slider the new fee value wouldn't get stored in the config as the sliderRelased signal is only emitted when the slider is pulled.
This change updates it on valueChanged.


ecdsa:
note: this pattern of saving only on release dates back from an old continuous slider, that caused a lot of disk writes.

PR #10329, opened 2025-11-28
SwapProvidersButton class

ecdsa:
factorize code that is used in both SwapDialog and Submarine payments

### DIFF
--- a/electrum/gui/qt/swap_dialog.py
+++ b/electrum/gui/qt/swap_dialog.py
@@ -9,7 +9,7 @@ from PyQt6.QtWidgets import QTreeWidget, QTreeWidgetItem, QHeaderView
 from electrum_aionostr.util import from_nip19
 
 from electrum.i18n import _
-from electrum.util import NotEnoughFunds, NoDynamicFeeEstimates, UserCancelled
+from electrum.util import NotEnoughFunds, NoDynamicFeeEstimates, UserCancelled, trigger_callback
 from electrum.bitcoin import DummyAddress
 from electrum.transaction import PartialTxOutput, PartialTransaction
 from electrum.fee_policy import FeePolicy
@@ -43,6 +43,43 @@ ROLE_NPUB = Qt.ItemDataRole.UserRole + 1000
 class InvalidSwapParameters(Exception): pass
 
 
+class SwapProvidersButton(QPushButton):
+
+    def __init__(self, transport_getter, config, main_window):
+        """parent must have a transport() method"""
+        QPushButton.__init__(self)
+        self.config = config
+        self.transport_getter = transport_getter
+        self.main_window = main_window
+        self.clicked.connect(self.choose_swap_server)
+        self.update()
+
+    def update(self):
+        transport = self.transport_getter()
+        if not isinstance(transport, NostrTransport):
+            # HTTPTransport or no Network, not showing server selection button
+            self.setEnabled(False)
+            self.setVisible(False)
+            return
+        self.setEnabled(True)
+        self.setVisible(True)
+        offer_count = len(transport.get_recent_offers())
+        button_text = f' {offer_count} ' + (_('swap providers') if offer_count != 1 else _('swap provider'))
+        self.setText(button_text)
+        # update icon
+        if self.config.SWAPSERVER_NPUB:
+            pubkey = from_nip19(self.config.SWAPSERVER_NPUB)['object'].hex()
+            self.setIcon(pubkey_to_q_icon(pubkey))
+
+    def choose_swap_server(self) -> None:
+        transport = self.transport_getter()
+        assert isinstance(transport, NostrTransport), transport
+        self.main_window.choose_swapserver_dialog(transport)  # type: ignore
+        self.update()
+        trigger_callback('swap_provider_changed')
+
+
+
 class SwapDialog(WindowModalDialog, QtEventListener):
 
     def __init__(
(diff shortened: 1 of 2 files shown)


electrum/gui/qt/swap_dialog.py
```diff
@@ -157,18 +197,18 @@ def on_event_fee(self, *args):
         self.update_send_receive()
 
     @qt_event_listener
-    def on_event_swap_offers_changed(self, recent_offers: Sequence['SwapOffer']):
-        self.set_server_button_text(len(recent_offers))
+    def on_event_swap_offers_changed(self):
```
f321x: here you remove the `recent_offers` argument which causes this exception:
```
  File "/home/user/code/electrum-fork/electrum/gui/qt/util.py", line 1500, in on_qt_callback_signal
    return func(self, *args[1:])
TypeError: SwapDialog.on_event_swap_offers_changed() takes 1 positional argument but 2 were given
```

electrum/gui/qt/swap_dialog.py
```diff
@@ -43,6 +43,50 @@
...
+            self.setVisible(False)
+            return
+        self.setEnabled(True)
+        self.setVisible(True)
+        offer_count = len(transport.get_recent_offers())
+        button_text = f' {offer_count} ' + (_('swap providers') if offer_count != 1 else _('swap provider'))
+        self.setText(button_text)
+        # update icon
+        pubkey = from_nip19(self.config.SWAPSERVER_NPUB)['object'].hex() if self.config.SWAPSERVER_NPUB else ''
+        self.setIcon(pubkey_to_q_icon(pubkey))
```
f321x: This raises if `pubkey` is `''`:
```
 18.21 | E | gui.qt.confirm_tx_dialog.ConfirmTxDialog | failed to create swap transport
  File "/home/user/code/electrum-fork/electrum/gui/qt/confirm_tx_dialog.py", line 772, in connect_and_update_tab
    self.update_submarine_tab()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/confirm_tx_dialog.py", line 878, in update_submarine_tab
    self.server_button.update()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/swap_dialog.py", line 72, in update
    self.setIcon(pubkey_to_q_icon(pubkey))
                 ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/gui/qt/util.py", line 731, in pubkey_to_q_icon
    color = QColor(*pubkey_to_rgb_color(server_pubkey))
                    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/submarine_swaps.py", line 222, in pubkey_to_rgb_color
    assert len(swapserver_pubkey) == 64, len(swapserver_pubkey)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 0
```

electrum/gui/qt/swap_dialog.py
ecdsa: thanks, fixed

electrum/gui/qt/swap_dialog.py
ecdsa: fixed too

electrum/gui/qt/swap_dialog.py
```diff
@@ -43,6 +43,51 @@
...
+class SwapProvidersButton(QPushButton, QtEventListener):
+
+    def __init__(self, transport_getter, config, main_window):
+        """parent must have a transport() method"""
+        QPushButton.__init__(self)
+        self.config = config
+        self.transport_getter = transport_getter
+        self.main_window = main_window
+        self.clicked.connect(self.choose_swap_server)
+        self.setEnabled(not self.config.SWAPSERVER_URL)
```
f321x: this `setEnabled` call seems redundant as the following call to `self.update()` will set it again

ecdsa: indeed

electrum/gui/qt/swap_dialog.py
```diff
@@ -43,6 +43,50 @@
...
+        self.setVisible(True)
+        offer_count = len(transport.get_recent_offers())
+        button_text = f' {offer_count} ' + (_('swap providers') if offer_count != 1 else _('swap provider'))
+        self.setText(button_text)
+        # update icon
+        if self.config.SWAPSERVER_NPUB:
+            pubkey = from_nip19(self.config.SWAPSERVER_NPUB)['object'].hex()
+            self.setIcon(pubkey_to_q_icon(pubkey))
+
+    @qt_event_listener
```
f321x: You aren't calling `self.register_callbacks()` and `self.unregister_callbacks()`, so the callback doesn't get called.

electrum/gui/qt/swap_dialog.py
```diff
@@ -43,6 +43,50 @@
...
+    def on_event_swap_offers_changed(self, recent_offers: Sequence['SwapOffer']):
+        self.update()
+
+    def choose_swap_server(self) -> None:
+        transport = self.transport_getter()
+        if transport is None:
```
f321x: This could probably be an assert too as the button shouldn't be visible if the transport is None

ecdsa: ok, removed

electrum/gui/qt/swap_dialog.py
ecdsa: calling update() from parent now

Issue #10331, opened 2025-11-28
android: the perils of upgrading Qt and rebasing p4a


SomberNight:
We should rebase our fork of python-for-android.
We are stuck on old versions of python (~~3.10~~, now 3.11) ~~and openssl (1.1, already EOL)~~.
Relatedly, we should also update Qt (from 6.4).
And the android NDK (from r23).

But everything is interwoven and tangled together.

Currently our [p4a fork](https://github.com/spesmilo/python-for-android/commit/003451604321deb1f10d879b63450ec76db4a758) is based on the [v2023.05.21](https://github.com/kivy/python-for-android/releases/tag/v2023.05.21) tag of upstream. Atm that's 2.5 years old. Though most of what we would want from p4a upstream has not been released yet - they [haven't tagged](https://github.com/kivy/python-for-android/issues/3248) a new release since v2024.01.21. I think that's not a serious problem -- we could just pick a recent commit.

Issues:

- we need to adapt the GUI to styling changes in newer Qt (already in 6.5)
    - we currently use Qt 6.4

- @accumulator [says](https://github.com/spesmilo/electrum/issues/10139#issuecomment-3214195424), in QML, assignment of custom property types has issues since Qt>=6.6
    - This would mean at least changing all types in the QML<->QObject boundary to a generic type.
    - see https://www.riverbankcomputing.com/pipermail/pyqt/2024-June/045922.html
        ```
        $ pip install --upgrade 'pyqt6==6.10.0' 'pyqt6-qt6==6.10.1'
        $ python3 ./assign-bug.py
        <Unknown File>:13:9: Unable to assign Simple to Simple
        ```
        ```
        $ pip install --upgrade 'pyqt6==6.5.3' 'pyqt6-qt6==6.5.3'
        $ python3 ./assign-bug.py
        // no issue
        ```

- Qt restricts the android NDK version:
    - Qt 6.4 -> NDK r23
        - https://doc.qt.io/archives/qt-6.4/android.html
    - Qt 6.5 -> NDK r25 or r26 or r27
        - https://doc.qt.io/qt-6.5/android.html
    - Qt 6.8 -> NDK r26 or r27
        - https://doc.qt.io/qt-6.8/android.html

- p4a restricts the NDK version:
    - current min of [upstream is r25](https://github.com/kivy/python-for-android/blob/ad8f90296ed16831c4768d93988ac53fe50fb1de/pythonforandroid/recommendations.py#L12)
    - by default they use r28c
[...]

PR #10332, opened 2025-11-30
android build: update openssl, and switch cryptography->pycryptodomex

SomberNight:
- re android build:
  - bumps OpenSSL 1.1.1w -> 3.0.18
    - the 1.1.1 branch has been EOL since 2023-09
    - the 3.0 branch is supported until 2026-09
  - replaces `cryptography` with `pycryptodomex`
    - we had been using an old `cryptography==2.8` in the Android build
      - newer versions require Rust to build, which is hard with p4a (though upstream p4a has some kind of solution)
      - this old version of `cryptography` is not compatible with new OpenSSL however
      - `pycryptodomex` is easy to build. it does not even depend on OpenSSL.

-----

- p4a changes:
  - update openssl recipe: https://github.com/SomberNight/python-for-android/commit/e91310fbf88857b2fbd9f9a5f591300ae5da90b5
- electrum-aionostr changes:
  - rm hard dependency on `cryptography`, allow using `pycryptodomex` as alternative
    - this needs https://github.com/spesmilo/electrum-aionostr/pull/18 merged, and released
      - done as `electrum-aionostr==0.1.0`
    - and then the pin needs to be updated in `contrib/deterministic-build/requirements.txt`
      - done in https://github.com/spesmilo/electrum/commit/e188102eb3b30279c8efbd7ff484d904dbd70ac0 
 
note: tested and builds are still reproducible

### DIFF
--- a/run_electrum
+++ b/run_electrum
@@ -100,6 +100,13 @@ if not is_android:
     check_imports()
 
 
+if is_android:
+    # hack to make pycryptodomex work on Android
+    # from https://github.com/kivy/python-for-android/issues/1866#issuecomment-927157780
+    import ctypes
+    ctypes.pythonapi = ctypes.PyDLL("libpython%d.%d.so" % sys.version_info[:2])  # replaces ctypes.PyDLL(None)
+
+
 sys._ELECTRUM_RUNNING_VIA_RUNELECTRUM = True  # used by logging.py
 
 from electrum.logging import get_logger, configure_logging  # import logging submodule first


SomberNight:
note: I forgot about https://github.com/spesmilo/electrum/pull/7192

SomberNight:
> note: I forgot about #7192

Ok, I propose we don't care about openalias. For now we switch to using pycryptodomex instead of cryptography, on Android only. That way we can unblock upgrading OpenSSL.

Besides DNSSEC, everything else works with pycryptodomex. We only use DNSSEC for openalias atm. The switch means we break openalias on Android.
Note: a few hardware wallet plugins also require cryptography :/

---

In the future we might want to add support for [bip-0353](https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki), which also requires DNSSEC. By then we will need to figure something out for Android.

Ideally we will rebase p4a (see https://github.com/spesmilo/electrum/issues/10331) and be able to build cryptography for Android. In that case we will be able to switch back.

PR #10336, opened 2025-12-01
json_db: fix some type hints, and bugfix for `StoredDict.__delitem__`

SomberNight:
follow-ups to https://github.com/spesmilo/electrum/pull/10233
review commit-by-commit

- add/fix some type hints
- add some asserts
- fix a bug re `StoredDict.__delitem__`

### DIFF
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -222,8 +222,11 @@ class StoredDict(dict, BaseStoredObject):
     @locked
     def __delitem__(self, key: _FLEX_KEY) -> None:
         assert isinstance(key, _FLEX_KEY), repr(key)
+        r  = self.get(key, None)
         dict.__delitem__(self, key)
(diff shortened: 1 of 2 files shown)


electrum/json_db.py
```diff
@@ -222,8 +222,11 @@ def __setitem__(self, key: _FLEX_KEY, v) -> None:
     @locked
     def __delitem__(self, key: _FLEX_KEY) -> None:
         assert isinstance(key, _FLEX_KEY), repr(key)
+        r  = self.get(key, None)
         dict.__delitem__(self, key)
         self.db_remove(key)
+        if isinstance(r, StoredDict):
+            r._parent = None
```
ecdsa, line 229: Good catch!
Actually, this should be `if isinstance(r, BaseStoredObject)`. (in both methods)

Issue #5999, opened 2020-02-27
Add proper header to storage (to expose some data before decryption)


SomberNight:
Currently, encrypted wallet files don't really have a header.
The file contents are:
https://github.com/spesmilo/electrum/blob/ce81957d25e66296c0baaf48683adcc623a3dae4/electrum/ecc.py#L503-L506
So it's just a 4 byte magic, and data needed for the decryption itself.

The magic itself is used to distinguish between two wallet types.
`BIE1` wallets are encrypted with a user provided password ("userpw"), and
`BIE2` wallets are encrypted with a pubkey used as password that comes from a hardware wallet (along a hardcoded bip32 derivation path) ("xpubpw").
https://github.com/spesmilo/electrum/blob/ce81957d25e66296c0baaf48683adcc623a3dae4/electrum/storage.py#L171-L176

It would be useful to have a proper header. For example, the header could include
- for xpubpw, the derivation path (related: https://github.com/spesmilo/electrum/pull/5993)
- a random salt, to go in the hmac with the password (useful for both userpw and xpubpw)
- maybe fancy things such as hmac iteration count, or even hmac algorithm
- IIRC some metadata would need to be stored to allow partial writes (https://github.com/spesmilo/electrum/issues/4823)

SomberNight:
Note: if we make an incompatible change to the storage file format, we should also change the KDF to something stronger. Maybe [scrypt](https://docs.python.org/3/library/hashlib.html#hashlib.scrypt) (and introduce a random salt at the same time).

SomberNight:
somewhat related: https://github.com/spesmilo/electrum/issues/5561

SomberNight:
@ln2max take a look

ln2max:
See my comment here: https://github.com/spesmilo/electrum/issues/4823#issuecomment-919348130

ecdsa:
To allow password-encryption for hardware wallets (#5561), we need to have multiple ways to decrypt the same wallet. 
We can encrypt the wallet with a random "master key", and add several fields, each containing master key, encrypted with a different passwords. This also allows us to change the password without rewriting the entire file.

Regarding #5993, I am not sure why the derivation path should be part of the header. Since the derivation path depends on the hardware wallet, I believe it should be defined in the corresponding plugin.

Since the master key we are encrypting is random, no rainbow table can be built. Thus, I believe a salt is unnecessary.

The header would consist of:
- Magic: b"Electrum"
- Version (1 byte)
[...]

PR #10339, opened 2025-12-02
New encrypted wallet file header:

ecdsa:
- the file is encrypted with a master key
 - the master key is encrypted with one or several passwords
 - passwords can be added to or removed from the list
 - a hmac of the file is written at the beginning of the file
 - passwords can be checked/updated without having to read/write the whole file
 - the GUI assumes a single password
 - partial writes: encrypted blobs are appended at the end of the file each blob is prefixed by its length can recover from incomplete write: hmac is updated after appending


SomberNight:
related https://github.com/spesmilo/electrum/issues/5999

PR #10340, opened 2025-12-02
android: implement biometric authentication (fingerprint)

### LINKED ISSUE #7560: Feature Request: Android Biometric & Numeric Keypad Support
I would like to see Electrum on Android unlock wallets using fingerprint scanning.

Also would like to see the option for a 10 key numeric keyboard when entering the wallet password.

f321x:
Allows to unlock the android app with the android biometric api (e.g. fingerprint). Can be enabled in the settings.
The initial version was LLM generated but i reviewed it and rewrote a couple things.

TODO: checkbox to enable it in the wizard during wallet setup once one password is enforced for creation of new wallets

closes https://github.com/spesmilo/electrum/issues/7560


electrum/simple_config.py
```diff
@@ -676,6 +676,11 @@ def __setattr__(self, name, value):
     WALLET_UNCONF_UTXO_FREEZE_THRESHOLD_SAT = ConfigVar('unconf_utxo_freeze_threshold', default=5_000, type_=int)
     WALLET_PAYREQ_EXPIRY_SECONDS = ConfigVar('request_expiry', default=invoices.PR_DEFAULT_EXPIRATION_WHEN_CREATING, type_=int)
     WALLET_USE_SINGLE_PASSWORD = ConfigVar('single_password', default=False, type_=bool)
+    WALLET_ANDROID_USE_BIOMETRIC_AUTHENTICATION = ConfigVar('android_use_biometrics', default=False, type_=bool)
+    WALLET_ANDROID_BIOMETRIC_DATA = ConfigVar(
+        'android_biometric_data', default='', type_=str,
+        short_desc=lambda: 'Wallet password encrypted with android biometric authentication key'
```
SomberNight: i18n localization missing

electrum/gui/qml/qebiometrics.py
```diff
@@ -0,0 +1,129 @@
...
+            intent.putExtra(jString("data"), jString(encrypted_password))
+        else:
+            raise ValueError(f"unsupported {action=}")
+
+        activity.bind(on_activity_result=self._on_activity_result)
+        jPythonActivity.startActivityForResult(intent, 12345) # Request code
+
+    def _on_activity_result(self, requestCode: int, resultCode: int, intent):
+        if requestCode != 12345:
+            return
```
SomberNight: No magic numbers. Create a named constant for it.

E.g. (though this example is in java)
https://github.com/spesmilo/electrum/blob/cb7e550abbbbcaebab8087ad4f562fd78fb5fc02/electrum/gui/qml/java_classes/org/electrum/qr/SimpleScannerActivity.java#L33

----

Also, just noticed we don't really do this for the qr scanner?? weird. I guess we should.
https://github.com/spesmilo/electrum/blob/cb7e550abbbbcaebab8087ad4f562fd78fb5fc02/electrum/gui/qml/qeqrscanner.py#L57-L60

SomberNight (commented):
just some quick comments, I haven't really looked at this

electrum/simple_config.py
f321x: Yeah I intentionally didn't add it as the description is not being used, but added the description as i thought of it as a more productive use of the space than adding a simple comment, probably the worst of both worlds. 
I removed the description and use a more descriptive variable name now (`*BIOMETRIC_DATA` sounded a bit ambiguous).

electrum/gui/qml/qebiometrics.py
f321x: Created constants and added a commit https://github.com/spesmilo/electrum/pull/10340/commits/3d722758e0e2562d8a946ec84a9c55d4b7122754 to also use a `requestCode` in the QR code scanner.

accumulator:
I assume this will be combined with #10339? (as handing over the wallet password to some biometric backend is a risk)

f321x:
@accumulator no, the PRs are independent. 
The password is encrypted with a secret retrieved from the secure element inside the new Java activity, we then store the encrypted password in our config file, so the wallet password is not stored anywhere outside of Electrum.

ecdsa:
this looks good to me, although I did not build an apk,

f321x:
running this apk for 2 weeks now, no issues so far

electrum/gui/qml/qebiometrics.py
```diff
@@ -0,0 +1,133 @@
...
+    jString = autoclass('java.lang.String')
+
+
+class BiometricAction(str, Enum):
+    ENCRYPT = "ENCRYPT"
+    DECRYPT = "DECRYPT"
+
+
+class QEBiometrics(QObject):
+    REQUEST_CODE_BIOMETRIC_ACTIVITY = 24553
```
SomberNight: could use same `# random 16 bit int` comment

electrum/gui/qml/qebiometrics.py
```diff
@@ -0,0 +1,133 @@
...
+    RESULT_CODE_SETUP_FAILED = 101
+    RESULT_CODE_POPUP_CANCELLED = 102
```
SomberNight:
```suggestion
    RESULT_CODE_SETUP_FAILED = 101  # duplicated from BiometricActivity.java
    RESULT_CODE_POPUP_CANCELLED = 102
```

electrum/gui/qml/qebiometrics.py
```diff
@@ -0,0 +1,133 @@
...
+    def decrypt(self, encrypted_bundle: str):
+        """
+        Called when the user needs to authenticate. We pass the encrypted password, it will get
+        decrypted on java side with the decryption key from the OS.
+        encrypted_bundle format: "iv:ciphertext"
+        """
+        self._start_activity(BiometricAction.DECRYPT, data=encrypted_bundle)
+
+    def _start_activity(self, action: BiometricAction, data: str):
+        self._current_action = action
```
SomberNight, line 128: Is it deliberate that the action is not passed around? You already put it into the request intent, you could just also put it into `resultIntent`. `_on_activity_result` could then just read it out from the response instead of being stateful in this way.

electrum/gui/qml/qebiometrics.py
```diff
@@ -0,0 +1,133 @@
...
+        elif action == BiometricAction.DECRYPT:
+            assert ':' in data, f"malformed encrypted_bundle: {data=}"
+            iv, encrypted_password = data.split(':')
+            intent.putExtra(jString("iv"), jString(iv))
+            intent.putExtra(jString("data"), jString(encrypted_password))
+        else:
+            raise ValueError(f"unsupported {action=}")
+
+        activity.bind(on_activity_result=self._on_activity_result)
+        jPythonActivity.startActivityForResult(intent, self.REQUEST_CODE_BIOMETRIC_ACTIVITY)
```
SomberNight, line 145: How do intents work?

- e.g. can another android application register an intent filter so that their app would get this intent, and e.g. obtain the plaintext unified password?
  I mean this kind of stuff: https://developer.android.com/privacy-and-security/risks/implicit-intent-hijacking#java
  Are we protected from this by using this `intent = jIntent(jPythonActivity, jBiometricActivity)` [constructor](https://developer.android.com/reference/android/content/Intent#Intent(android.content.Context,%20java.lang.Class%3C?%3E))?

- similarly but in other direction, can some other application send *us* this intent, and e.g. trigger our `_on_activity_result`?
  Or is that only possible if we explicitly set an `<intent-filter>` tag in our `AndroidManifest.xml`?

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
```diff
@@ -0,0 +1,168 @@
...
+            Cipher cipher = cryptoObject.getCipher();
+            Intent intent = getIntent();
+            String action = intent.getStringExtra("action");
+            Intent resultIntent = new Intent();
+
+            if ("ENCRYPT".equals(action)) {
+                String data = intent.getStringExtra("data"); // Wallet password to encrypt
+                byte[] encrypted = cipher.doFinal(data.getBytes(Charset.forName("UTF-8")));
+                resultIntent.putExtra("data", Base64.encodeToString(encrypted, Base64.NO_WRAP));
+                resultIntent.putExtra("iv", Base64.encodeToString(cipher.getIV(), Base64.NO_WRAP));
```
SomberNight: So @accumulator's [point](
https://github.com/spesmilo/electrum/pull/10340#issuecomment-3612682330) stands that we do pass around the wallet password across process boundaries, and to some crypto backend (running in some system process to carry out the AndroidKeystore stuff?), even if temporarily?


Note the wallet password already travels through the QML->python boundaries, though I think that is the same unix process. Crossing into Java (using an intent) is cross-process though (but re security of intents, see my other comment). And then even when in Java land, this cipher stuff makes the password travel to yet another process to do AES?

See https://developer.android.com/privacy-and-security/keystore#ExtractionPrevention

> Key material never enters the application process. When an app performs cryptographic operations using an Android Keystore key, behind the scenes plaintext, ciphertext, and messages to be signed or verified are fed to a system process that carries out the cryptographic operations.

although I am a bit confused - does this `cipher.doFinal()` call out to that external process then?

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
```diff
@@ -0,0 +1,168 @@
...
+            setResult(RESULT_CANCELED);
+        }
+        finish();
+    }
+
+    private SecretKey getSecretKey() throws Exception {
+        KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
+        keyStore.load(null);
+        return (SecretKey) keyStore.getKey(KEY_NAME, null);
+    }
```
SomberNight, line 152: Can another android application get our secret key?

Do we get protection automatically by using [`AndroidKeyStore`](https://developer.android.com/privacy-and-security/keystore#WhichShouldIUse)?

electrum/simple_config.py
SomberNight: I see. I think some sort of explanation there is useful, so you had the right line of thought. You could add a code comment perhaps. Something like:

> `# this is the "unified wallet password", encrypted with a random token (token stored in AndroidKeyStore)`

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
SomberNight: To be clear, I am not trying to imply that these are the wrong choices, I just want to make sure we understand what we are doing.

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
f321x: Using `AndroidKeyStore` prevents other apps from using our secret by assigning it to the Electrum apps specific UID. 
So unless the phone is rooted or compromised it shouldn't be possible for other apps to request any cryptographic operation on the Electrum AndroidKeyStore entry. And each operation requires user authentication as the key has been generated with `setUserAuthenticationRequired(true)`

> Use the Android Keystore provider to let an individual app store its own credentials, which only that app can access. - https://developer.android.com/privacy-and-security/keystore#WhichShouldIUse

> Access control
Android Keystore provides a central component for the storage and use of hardware-backed cryptographic keys, both for apps and for other system components. As such, access to any individual key is normally limited to the app or system component that created the key.

> Keystore domains
> To support this access control, keys are identified to Keystore with a key descriptor. This key descriptor indicates a domain that the descriptor belongs to, together with an identity within that domain.
> Android apps access Keystore using the standard Java Cryptography Architecture, which identifies keys with a string alias. This method of identification maps to the Keystore APP domain internally; the UID of the caller is also included to disambiguate keys from different apps, preventing one app from accessing another's keys.
> Internally, the frameworks code also receives a unique numeric key ID after a key has been loaded. This numeric ID is used as the identifier for key descriptors within the KEY_ID domain. However, access control is still performed: even if one app discovers a key ID for another app's key, it can't use it in normal circumstances. - https://source.android.com/docs/security/features/keystore#access-control

This is the source code of the keystore, you can grep for `caller_uid` to find the permission checks:
https://android.googlesource.com/platform/system/security/+/main/keystore2/src/database.rs

electrum/gui/qml/qebiometrics.py
f321x: > e.g. can another android application register an intent filter so that their app would get this intent, and e.g. obtain the plaintext unified password?

Your link states that:

> Unless the application requires it, make intents explicit by calling setPackage(). This allows the intent to be interpreted only by a specific component (either in-app or from other applications), preventing untrusted applications from intercepting the data sent along with the intent.

So using an an explicit Intent should prevent hijacking.

As we construct the jIntent with the target class `jBiometricActivity` this is the same as calling setComponent, according to the docs of the Intent constructor you linked.
Setting the component with setComponent makes an Intent explicit:

> There are two primary forms of intents you will use. Explicit Intents have specified a component (via [setComponent(ComponentName)](https://developer.android.com/reference/android/content/Intent#setComponent(android.content.ComponentName)) or [setClass(Context, Class)](https://developer.android.com/reference/android/content/Intent#setClass(android.content.Context,%20java.lang.Class%3C?%3E))), which provides the exact class to be run...  - https://developer.android.com/reference/android/content/Intent#intent-resolution

So our biometric intent should be safe from hijacking as it explicitly specifies the receiver of the intent, jBiometricActivity.

> similarly but in other direction, can some other application send us this intent, and e.g. trigger our _on_activity_result?

For another app to be able to launch our intent we would have to allow this by setting android:exported true in the AndroidManifest.xml or register an intent filter which would also set android:exported true. Looking at the generated AndroidManifest.xml we don't do this:
```
cat ./.buildozer_qml/android/platform/build-arm64-v8a/dists/Electrum/src/main/AndroidManifest.xml
...
    <activity android:name="org.electrum.qr.SimpleScannerActivity"></activity>
    
    <activity android:name="org.electrum.biometry.BiometricActivity"></activity>
    
    </application>
...
```
https://developer.android.com/guide/topics/manifest/activity-element#exported

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
f321x: > Crossing into Java (using an intent) is cross-process though

https://developer.android.com/guide/components/processes-and-threads says:

> By default, all components of the same application run in the same process and thread, called the main 

Is this a python for android specific thing to start java activities in a separate process? If i understand the docs correctly "all components" should run in the same process unless explicitly specified to run in another process with an android:process attribute, which i don't find us specifying anywhere.

> does this cipher.doFinal() call out to that external process then?

I think so, yes. Looking at https://android.googlesource.com/platform/frameworks/base/+/508e665/keystore/java/android/security/keystore/AndroidKeyStoreCipherSpiBase.java you can see that they send the data to the keystore through some IPC handler `mMainDataStreamer`. Grep `createMainDataStreamer`, `mMainDataStreamer.doFinal`

electrum/simple_config.py
f321x: Added this comment in https://github.com/spesmilo/electrum/compare/0cf8b3ca8c9ca831af5dc65b02af49aad284ef0a..b0800d62232376ff61717c01e23f1a70189fd5b9

electrum/gui/qml/qebiometrics.py
f321x: Added in https://github.com/spesmilo/electrum/compare/0cf8b3ca8c9ca831af5dc65b02af49aad284ef0a..b0800d62232376ff61717c01e23f1a70189fd5b9

electrum/gui/qml/qebiometrics.py
f321x: Added in https://github.com/spesmilo/electrum/compare/0cf8b3ca8c9ca831af5dc65b02af49aad284ef0a..b0800d62232376ff61717c01e23f1a70189fd5b9

electrum/gui/qml/qebiometrics.py
f321x: Yeah good idea, doing this now in https://github.com/spesmilo/electrum/compare/0cf8b3ca8c9ca831af5dc65b02af49aad284ef0a..b0800d62232376ff61717c01e23f1a70189fd5b9

electrum/gui/qml/qebiometrics.py
SomberNight: > As we construct the jIntent with the target class `jBiometricActivity` this is the same as calling setComponent, according to the docs of the Intent constructor you linked. [..]
> So our biometric intent should be safe from hijacking as it explicitly specifies the receiver of the intent, jBiometricActivity.

I was also wondering if another application could force a collision by creating a Java package also named `org.electrum.biometry.BiometricActivity` and how that would be handled.

Is your reading of the docs that by using the `intent = jIntent(jPythonActivity, jBiometricActivity)` constructor (`public Intent (Context packageContext, Class<?> cls)`), `packageContext=jPythonActivity` will get set to something that effectively limits the search to the `org.electrum.electrum` (i.e. the current) android package?

> For another app to be able to launch our intent we would have to allow this by setting android:exported true in the AndroidManifest.xml or register an intent filter which would also set android:exported true.
> https://developer.android.com/guide/topics/manifest/activity-element#exported

Are you assuming the default for `android:exported` is `false`? That doc does not say what the default is :/
Looks like [the default is unclear](https://developer.android.com/privacy-and-security/risks/android-exported)... >.<

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
SomberNight: >> Crossing into Java (using an intent) is cross-process though
>
> Is this a python for android specific thing to start java activities in a separate process?

Our python/qml/java code is all running in the same process. (What I don't know is whether intent resolution is done inside this same process.)

```
$ adb logcat | grep -F "`adb shell ps | grep org.electrum.electrum | cut -c14-19`"
01-11 06:07:10.530 18591 18591 W zygote64: Activation of cgroup controller +memory failed in path /sys/fs/cgroup/apps/uid_1110227: Invalid argument
01-11 06:07:10.533 18591 18591 I libc    : SetHeapTaggingLevel: tag level set to 0
01-11 06:07:10.537 18591 18591 I Zygote  : Process 18591 created for org.electrum.electrum
[...]
01-11 06:07:10.792 18591 18591 V PythonActivity: My onCreate running
01-11 06:07:10.792 18591 18591 V PythonActivity: loading native lib
[...]
01-11 06:07:11.087 18591 18615 I python  : Initializing Python for Android
[...]
01-11 06:07:11.895 18591 18615 I python  :   0.69 | I | logging | Electrum version: 4.6.2 - https://electrum.org - https://github.com/spesmilo/electrum
```
The two columns after the timestamp are the process id and the thread id.
- `01-11 06:07:10.792 18591 18591 V PythonActivity: My onCreate running`
    - https://github.com/spesmilo/python-for-android/blob/a01269f7799587ad74ee40e0b642d917b8db7d4e/pythonforandroid/bootstraps/qt6/build/src/main/java/org/kivy/android/PythonActivity.java#L100
- `01-11 06:07:10.792 18591 18591 V PythonActivity: loading native lib`
    - https://github.com/spesmilo/python-for-android/blob/a01269f7799587ad74ee40e0b642d917b8db7d4e/pythonforandroid/bootstraps/qt6/build/src/main/java/org/kivy/android/PythonActivity.java#L86-L88
- `01-11 06:07:11.087 18591 18615 I python  : Initializing Python for Android`
    - https://github.com/spesmilo/python-for-android/blob/a01269f7799587ad74ee40e0b642d917b8db7d4e/pythonforandroid/bootstraps/common/build/jni/application/src/Android.mk#L5
    - https://github.com/spesmilo/python-for-android/blob/a01269f7799587ad74ee40e0b642d917b8db7d4e/pythonforandroid/bootstraps/common/build/jni/application/src/Android.mk#L12-L13
    - https://github.com/spesmilo/python-for-android/blob/a01269f7799587ad74ee40e0b642d917b8db7d4e/pythonforandroid/bootstraps/common/build/jni/application/src/start.c#L83
    - https://github.com/spesmilo/python-for-android/blob/a01269f7799587ad74ee40e0b642d917b8db7d4e/pythonforandroid/bootstraps/common/build/jni/application/src/start.c#L309-L311
- `01-11 06:07:11.895 18591 18615 I python  :   0.69 | I | logging | Electrum version: 4.6.2 - https://electrum.org - https://github.com/spesmilo/electrum`
    - https://github.com/spesmilo/electrum/blob/45e1b0da8d3461234db213cf255e5a1f49ff54c0/electrum/logging.py#L336

-----

As intents can be used to communicate between different apps, I was thinking the intent goes out to an external process to decide what to do. But maybe that only happens for implicit intents, and perhaps explicit intents are handled in-process. Not sure.

electrum/gui/qml/qebiometrics.py
f321x: I reverted this change in https://github.com/spesmilo/electrum/compare/b0800d62232376ff61717c01e23f1a70189fd5b9..147e4185b28f8317a6b4dd7b31e11daebfc6aaa0.

It introduced a regression because i didn't return the action string in the result intent when the fingerprint dialog gets cancelled by the user, causing the app to crash because pyjnius couldn't find the `action` string.

I think its easier to reason about and more maintainable to just keep this little state, allowing us to just `setResult(RESULT_CANCELED)` anywhere in the Java activity, than having to think about returning the action string in every possible result intent which adds quite some boilerplate to the Java activity as we even have to retrieve the action string first in some exit paths where we haven't retrieved it yet.
The risk of keeping the state seems low as we cannot fire multiple fingerprint requests concurrently (they block the screen), so everything should happen in order.

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
f321x: Okay, yes, the Activity runs in our process, however the Intent goes to an Android system process through IPC, which resolves the Activity for us and then schedules it to run it in our process. So this Android process can see the content (e.g. password) of our Intent. This blog describes the Activity flow with snippets of the source: https://devpress.csdn.net/cloud/62f658ee7e6682346618b142.html

ecdsa: Note: if the concern is that the wallet password may be exposed to external processes, this is easy to address with an extra level of indirection:
 - choose a random token T
 - use T to encrypt/decrypt the wallet password
 - ask the android system to encrypt/decrypt T
 - store encrypted T and encrypted wallet password in config

@SomberNight points out that T should be updated whenever the user updates the password.

electrum/gui/qml/qebiometrics.py
f321x: > Is your reading of the docs that by using the intent = jIntent(jPythonActivity, jBiometricActivity) constructor (public Intent (Context packageContext, Class<?> cls)), packageContext=jPythonActivity will get set to something that effectively limits the search to the org.electrum.electrum (i.e. the current) android package?

Yes, the constructor sets `mComponent` to a `ComponentName`:
```
    /**
     * Create an intent for a specific component.  All other fields (action, data,
     * type, class) are null, though they can be modified later with explicit
     * calls.  This provides a convenient way to create an intent that is
     * intended to execute a hard-coded class name, rather than relying on the
     * system to find an appropriate class for you; see {@link #setComponent}
     * for more information on the repercussions of this.
     *
     * @param packageContext A Context of the application package implementing
     * this class.
     * @param cls The component class that is to be used for the intent.
     *
     * @see #setClass
     * @see #setComponent
     * @see #Intent(String, android.net.Uri , Context, Class)
     */
    public Intent(Context packageContext, Class<?> cls) {
        mComponent = new ComponentName(packageContext, cls);
    }
```
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/content/Intent.java

The component `ComponentName` contains both, package name and class name:

>  Two pieces of information, encapsulated here, are required to identify a component: the package (a String) it exists in, and the class (a String) name inside of that package.

https://developer.android.com/reference/android/content/ComponentName

The `mComponent` is then used to resolve the Activity:
```
    public ComponentName resolveActivity(@NonNull PackageManager pm) {
        if (mComponent != null) {
            return mComponent;
        }
```
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/content/Intent.java


> Are you assuming the default for android:exported is false? That doc does not say what the default is :/
Looks like [the default is unclear](https://developer.android.com/privacy-and-security/risks/android-exported)... >.<

I would like to set exported false, however python-for-android/buildozer doesn't allow it in a straightforward way. We would probably have to modify https://github.com/spesmilo/python-for-android/blob/003451604321deb1f10d879b63450ec76db4a758/pythonforandroid/bootstraps/qt6/build/templates/AndroidManifest.tmpl.xml#L150 to allow setting custom properties on the Activities or use some other hack to inject it in the AndroidManifest.xml.
However as we don't register any filters for this Activity it seems like android defaults it to false, looking at the Service docs: https://developer.android.com/guide/topics/manifest/service-element?hl=en#exported

> The default value depends on whether the service contains intent filters. The absence of any filters means that it can be invoked only by specifying its exact class name. This implies that the service is intended only for application-internal use, since others don't know the class name. So, in this case, the default value is "false". On the other hand, the presence of at least one filter implies that the service is intended for external use, so the default value is "true". 

Note these are the docs for Services, not Activities, in the Activities docs this isn't specified as explicitly, however the API looks very similar. Also if you register filters Android [forces you to explicitly set exported](https://stackoverflow.com/questions/68554294/androidexported-needs-to-be-explicitly-specified-for-activity-apps-targeting), another indication that the default behavior when not specifying filters is to reject requests.

SomberNight: > The component ComponentName contains both, package name and class name

All right, this approach looks perfectly safe then.

> Also if you register filters Android [forces you to explicitly set exported](https://stackoverflow.com/questions/68554294/androidexported-needs-to-be-explicitly-specified-for-activity-apps-targeting), another indication that the default behavior when not specifying filters is to reject requests.

Ok, that looks reassuring.

Thank you for looking into these.

electrum/gui/qml/qebiometrics.py
SomberNight: >  The risk of keeping the state seems low as we cannot fire multiple fingerprint requests concurrently (they block the screen), so everything should happen in order.

Well, okay. Let's keep it like this then.

f321x:
Force pushed from 147e4185b28f8317a6b4dd7b31e11daebfc6aaa0 to dac9e6a028fc3ab28a22292d5b2de692d99df733.
Implements the password encryption indirection described by @ecdsa in https://github.com/spesmilo/electrum/pull/10340#discussion_r2682532270 and simplifies the `QEBiometrics` api by keeping all config changes on the Python side.

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
f321x: Implemented this in https://github.com/spesmilo/electrum/pull/10340/commits/d4c2123db6dccdb0c53995f58e4effe09874fad9

electrum/gui/qml/qedaemon.py
```diff
@@ -232,6 +232,12 @@ def load_wallet_task():
 
                 if self.daemon.config.WALLET_SHOULD_USE_SINGLE_PASSWORD:
                     self._use_single_password = self.daemon.update_password_for_directory(old_password=local_password, new_password=local_password)
+                    if not self._use_single_password and self.daemon.config.WALLET_ANDROID_USE_BIOMETRIC_AUTHENTICATION:
+                        # we need to disable biometric auth if the user creates wallets with different passwords as
+                        # we only store one encrypted password which is not associated to a specific wallet
+                        self._logger.warning(f"biometric authentication disabled, not in single password mode")
+                        self.daemon.config.WALLET_ANDROID_USE_BIOMETRIC_AUTHENTICATION = False
+                        self.daemon.config.WALLET_ANDROID_BIOMETRIC_ENCRYPTED_DATA = ''
```
SomberNight: configvar does not exist

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
```diff
@@ -0,0 +1,168 @@
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) return;
+        try {
+            BiometricPrompt.CryptoObject cryptoObject = result.getCryptoObject();
+            Cipher cipher = cryptoObject.getCipher();
+            Intent intent = getIntent();
+            String action = intent.getStringExtra("action");
+            Intent resultIntent = new Intent();
+
+            if ("ENCRYPT".equals(action)) {
+                String data = intent.getStringExtra("data"); // Wallet password to encrypt
```
SomberNight: stale comment

electrum/gui/qml/qedaemon.py
```diff
@@ -232,6 +232,12 @@ def load_wallet_task():
 
                 if self.daemon.config.WALLET_SHOULD_USE_SINGLE_PASSWORD:
                     self._use_single_password = self.daemon.update_password_for_directory(old_password=local_password, new_password=local_password)
+                    if not self._use_single_password and self.daemon.config.WALLET_ANDROID_USE_BIOMETRIC_AUTHENTICATION:
+                        # we need to disable biometric auth if the user creates wallets with different passwords as
+                        # we only store one encrypted password which is not associated to a specific wallet
+                        self._logger.warning(f"biometric authentication disabled, not in single password mode")
```
SomberNight:
```suggestion
                        self._logger.warning(f"disabling biometric authentication, not in single password mode")
```

SomberNight (commented):
Very nice!

electrum/simple_config.py
```diff
@@ -678,6 +678,11 @@ def __setattr__(self, name, value):
     WALLET_SHOULD_USE_SINGLE_PASSWORD = ConfigVar('should_use_single_password', default=False, type_=bool)
     # TODO: consider removing WALLET_DID_USE_SINGLE_PASSWORD once encrypted wallet file headers are available
     WALLET_DID_USE_SINGLE_PASSWORD = ConfigVar('did_use_single_password', default=False, type_=bool)
+    WALLET_ANDROID_USE_BIOMETRIC_AUTHENTICATION = ConfigVar('android_use_biometrics', default=False, type_=bool)
+    # this is the wrap key encrypted with a secret stored in AndroidKeyStore
+    WALLET_ANDROID_BIOMETRIC_AUTH_ENCRYPTED_WRAP_KEY = ConfigVar('android_biometrics_encrypted_wrap_key', default='', type_=str)
+    # this is the "unified wallet password", encrypted with a the wrap key
```
SomberNight:
```suggestion
    # this is the "unified wallet password", encrypted with the wrap key
```

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
SomberNight: Great. Thanks. I think this is a good approach.

electrum/simple_config.py
f321x: Fixed in https://github.com/spesmilo/electrum/compare/dac9e6a028fc3ab28a22292d5b2de692d99df733..6450187902f0802626894f5728d2980cb1bdc21b

electrum/gui/qml/qedaemon.py
f321x: Changed in https://github.com/spesmilo/electrum/compare/dac9e6a028fc3ab28a22292d5b2de692d99df733..6450187902f0802626894f5728d2980cb1bdc21b

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
f321x: Fixed in https://github.com/spesmilo/electrum/compare/dac9e6a028fc3ab28a22292d5b2de692d99df733..6450187902f0802626894f5728d2980cb1bdc21b

electrum/gui/qml/qedaemon.py
f321x: Fixed in https://github.com/spesmilo/electrum/compare/dac9e6a028fc3ab28a22292d5b2de692d99df733..6450187902f0802626894f5728d2980cb1bdc21b

electrum/gui/qml/components/main.qml
```diff
@@ -771,43 +828,57 @@ ApplicationWindow
...
-                dialog.open()
-            }
+                }
+            })
+            dialog.rejected.connect(function() {
+                qtobject.authCancel()
+            })
+            dialog.open()
+        } else if (method == 'pin') {
+            var dialog = app.pinDialog.createObject(app, {
```
ecdsa: After testing the APK, I think it would be better to remove the PIN code feature completely.

We currently have 3 methods of authentication: password, biometric and PIN. this is too much. What's even more confusing is that the user still needs to choose a PIN code in order to authenticate payments using biometric, as a fallback.

I propose to remove the PIN code completely. We use the password as fallback if biometric authentication does not work. We keep a boolean toggle for authenticating all payments. If biometrics is not available, this setting will ask for the password.

SomberNight: Before this PR, I have been using a wallet with a password that takes like 10-15 seconds to enter, and with a PIN.
The PIN I can enter in less than a second.

I want some kind of auth before sending a payment, otherwise it is not safe to pay IRL at POS terminals/etc as someone could just grab your phone and empty your hot wallet. It is not practical at all to have a strong password for the wallet but also have to enter that every time you *attempt* to pay for a beer.

It is not just about people with shitty phones that don't have biometric capabilities. It is also about people who intentionally choose not to enable them. If you don't use the fingerprint scanner of your phone for the lockscreen, you also can't use it for Electrum.

EDIT: So I think it is worth it to keep all the complexity.

SomberNight: Alternatively, I propose using the Android API that allows using the lockscreen unlock, instead of the purely biometric API. BlueWallet uses an Android API that allows unlocking using *any* lockscreen unlock method. That is,
- if I have PIN code as primary unlock, and fingerprint as secondary, that API allows using either
- if I have PIN code as primary, no secondary, that API allows using PIN code

That would satisfy my concerns.

ecdsa: > EDIT: So I think it is worth it to keep all the complexity.

I disagree. With this PR, we not only have 3 authentication methods, but they are also interacting:
 - the password is needed to disable the PIN if you forgot your PIN
 - you need to setup a PIN  in order to have an authentication fallback in case biometrics does not work, even you NEVER use a pin in normal conditions.
 
There is no way a normal person can infer how this works, unless they have a degree in computer engineering. This is unacceptable, We should not add features upon feature without cleaning things up from time to time.

ecdsa: > Alternatively, I propose using the Android API that allows using the lockscreen unlock

not sure how that works, but I suppose the PIN code you are referring to is not a Electrum setting then?

f321x: > Alternatively, I propose using the Android API that allows using the lockscreen unlock, instead of the purely biometric API

Note this would make payment authentication fully rely on Android, if someone is running QML on some other platform like Ubuntu Touch(?) they could not have any fallback method to authenticate payments (except we fall back to the password).

Though 99% of the users are probably running Android.

Also we would have to handle different Android API versions, it would be easy to enable the lock screen authentication for Android phones that are also supported by this PR (API >=29) as the API is the same, we could just add something like this to the `BiometricPrompt.PromptInfo.Builder()`: 
```
        .setAllowedAuthenticators(BIOMETRIC_STRONG or DEVICE_CREDENTIAL)
```
however for older phones we also support we'd probably have to implement some older Android APIs in addition.

So not sure if this ends up less complex if we want to allow all users to have some form of payment authentication without password and remove the built in pin.

SomberNight: Ok, so
- we remove config PIN
- we do this:
  > `BiometricPrompt.PromptInfo.Builder()`
  > `        .setAllowedAuthenticators(BIOMETRIC_STRONG or DEVICE_CREDENTIAL)`
- then without relying on Android, there is only the concept of the wallet password: it is needed to decrypt the wallet, and if the "PIN protect" config option (name TBD) is enabled, it is also needed before each payment
- if the app is running on Android 10+, they can enable the "biometric security" config option (name TBD), even if they don't have a fingerprint set up. We leave it to the system to decide how this works, e.g. allow either system PIN or fingerprint.
  - if the user cancels the system unlock prompt, we fall back to the wallet password
  - we reuse the same flow for decrypting wallets and re-authenticating before payments

f321x: Implemented this in https://github.com/spesmilo/electrum/pull/10340/commits/23ee523ba70ec361ff401ba2cb5865db77bcc8d6, the pin code / pin dialog logic is fully removed and the password dialog is used as fallback.

Also increases the minimum required Android API version to use the biometric functionality from 29 (Android 10, 2019) to 30 (Android 11, 2020) to simplify the Java code, otherwise we would need to maintain two different paths for a couple things as many API calls have been introduced or changed in 30.

electrum/gui/qml/components/Preferences.qml
```diff
@@ -158,59 +158,83 @@ Pane {
...
+
+                        Switch {
+                            id: paymentAuthentication
+                            // showing the toggle as checked even if the wallet has no password would be misleading
+                            checked: Config.paymentAuthentication && !(Daemon.currentWallet && parent.noWalletPassword)
+                            onToggled: {
+                                console.log('paymentAuthentication: ' + checked)
+                                if (activeFocus) {
+                                    // will request authentication when checked = false
+                                    Config.paymentAuthentication = !!checked;
```
SomberNight: is the double-negation intended? If so, is it done to type-cast?

Also, I notice other Switches are using [`onCheckedChanged`](https://github.com/spesmilo/electrum/blob/9857534e18eaf0af38cf51b91b921ff46b4447db/electrum/gui/qml/components/Preferences.qml#L83-L89) instead of `onToggled`. Is that because here the value of `checked` might change programmatically due to e.g. `parent.noWalletPassword` changing, and you want not to get triggered in that case?

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
```diff
@@ -0,0 +1,175 @@
...
+
+        Intent intent = getIntent();
+        String action = intent.getStringExtra("action");
+        String authMessage = intent.getStringExtra("auth_message");
+
+        Executor executor = getMainExecutor();
+        BiometricPrompt biometricPrompt = new BiometricPrompt.Builder(this)
+                .setTitle("Electrum Wallet")
+                .setAllowedAuthenticators(BiometricManager.Authenticators.BIOMETRIC_STRONG | BiometricManager.Authenticators.DEVICE_CREDENTIAL)
+                .setSubtitle(authMessage)
```
SomberNight, line 59: nice!
Showing a short string re what the prompt is for might help avoid "accidentally paying" as a last resort against a bug like https://github.com/spesmilo/electrum/issues/10406

electrum/gui/qml/components/WalletDetails.qml
```diff
@@ -475,6 +475,15 @@ Pane {
             })
             dialog.accepted.connect(function() {
                 var success = Daemon.setPassword(dialog.password)
+                if (success && Biometrics.isEnabled) {
+                    if (Biometrics.isAvailable) {
+                        // also update the biometric authentication
+                        Biometrics.enable(dialog.password)
```
SomberNight, line 481: note: if the user cancels the biometric system prompt here, biometrics will get disabled, the wallet password will get changed, and we will only display a "Password changed" success message. The user likely will not realise that biometrics got disabled.

but maybe it's ok. They will notice next time they try to authenticate and can just re-enable easily then.

f321x: I added a dialog to make it less confusing in case this happens:
Bildschirmfoto_20260116_111833

electrum/gui/qml/components/Preferences.qml
f321x: The double !! is a javascript pattern to ensure checked is a bool: https://stackoverflow.com/a/784946 (like bool(something) in Python). Though in hindsight it seems overkill for `checked` so i removed it to make it more readable.

I used onToggled because it seems to do what we want, only trigger the callback/connection code when the user actively toggles the switch (i think the activeFocus check is therefore redundant), however `onCheckedChanged` behaves similar with the `activeFocus` check as `activeFocus` is also only true if the user clicks on the switch. 
I changed the connection to use `onCheckedChanged` like the other elements so it is more consistent across the Preferences and causes no confusion on what to use.

electrum/gui/qml/components/WalletDetails.qml
ecdsa: two remarks:
 - if biometrics is enabled, in Walllet Details, the 'change password' dialog asks for fingerprint OR system password. This means that the user can type their system password in order to modify their Electrum password. I think this is going to confuse a lot of people. A legitimate expectation is that the next screen is going to modify the password that was entered to access it. I think this "change password" button should be treated as an exception, and never be hooked to biometrics.
 - if biometrics and authenticate payments are enabled, imagine that the user starts a payment, but then decides not to pay, because they changed their mind. When they click cancel, the Electrum password dialog pops up as a fallback. In most scenarios, this is going to be annoying: if I cancel an action, I would like it to be cancelled. We should probably not offer that fallback. I think the case where the user actually wants to revert to their Electrum password is rare enough to deserve visiting the preferences.

f321x: Okay, yes I see the point, added these changes in https://github.com/spesmilo/electrum/compare/bb8f9772b00d5fcb77d725910b64f279ff0d09b9..9a76759ce330541e6281fb3fb171778585a1c72d.

It will not fall back to the password dialog for signing, sign&broadcast and lightning payments (which are probably the most authentication cases in regular use).

SomberNight: https://github.com/spesmilo/electrum/blob/9a76759ce330541e6281fb3fb171778585a1c72d/electrum/gui/qml/auth.py#L10-L21

^ This is too much. Let's just remove the fallbacks from everywhere I think.
We could make do with three auth methods (named TBD?):

- `wallet` ("wallet_no_fallback"):
    - if biometrics enabled: do biometrics
        - if user cancels biometrics, no fallback, auth fails
    - if biometrics disabled: wallet password dialog
- `payment_auth` ("if_config_no_fallback"):
    - if 'Payment authentication' config disabled: auth succeeds
    - if 'Payment authentication' config enabled: go to `wallet` flow
- `wallet_password_only`: wallet password dialog

(When I say "do biometrics", I mean `BIOMETRIC_STRONG | DEVICE_CREDENTIAL`.)

Disabling biometrics in the Preferences dialog can be done without any auth. This is good I think. It means the user can effectively convert any auth_method to wallet_password_only.

So, I think, re `@auth_protect` methods, compared to master,
- `wallet_else_pin` can be killed. It was only used to remove the config PIN.
- `wallet` becomes above described `wallet` ("wallet_no_fallback")
- `pin` becomes `payment_auth`
- QEDaemon.startChangePassword uses `wallet_password_only`

One drawback is that the user can easily forget the wallet password if they have biometrics enabled: they basically never need it. I say we should accept this.

f321x: Okay, yes that's simpler, done this in https://github.com/spesmilo/electrum/compare/9a76759ce330541e6281fb3fb171778585a1c72d..9e42c1215be210ca47428d1d11c38b3d84357090.

> One drawback is that the user can easily forget the wallet password if they have biometrics enabled: they basically never need it.

We could simply prevent the user from falling for this footgun by requiring `wallet_password_only` to disable biometrics. Then they notice "oh what was my password again?" and still have the chance to access the seed phrase or send funds out using the biometrics to do a backup in case they don't have one. Otherwise if they disable biometrics and restart the app they got locked out and cannot re-enable biometrics (only possible after unlock).

SomberNight: > We could simply prevent the user from falling for this footgun by requiring `wallet_password_only` to disable biometrics.

Ok, that sounds reasonable.
If you do that, please add a code comment near that briefly explaining that the auth is there to prevent the user locking themselves out if they forgot the password.

Though note that if the system decides to clear the AndroidKeystore tokens [for whatever reason](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationRequired(boolean)), and the user forgot the wallet password, they can still get locked out:

> The key will become irreversibly invalidated once the secure lock screen is disabled (reconfigured to None, Swipe or other mode which does not authenticate the user) or when the secure lock screen is forcibly reset (e.g., by a Device Administrator). Additionally, if the key requires that user authentication takes place for every use of the key, it is also irreversibly invalidated once a new biometric is enrolled or once\ no more biometrics are enrolled, unless [setInvalidatedByBiometricEnrollment(boolean)](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setInvalidatedByBiometricEnrollment(boolean)) is used to allow validity after enrollment, or KeyProperties.AUTH_DEVICE_CREDENTIAL is specified as part of the parameters to [setUserAuthenticationParameters(int, int)](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationParameters(int,%20int)).

f321x: Doing this now in https://github.com/spesmilo/electrum/compare/9e42c1215be210ca47428d1d11c38b3d84357090..47efb8b108e0a0bd6c161b527d0af185bad12826

>  If you do that, please add a code comment near that briefly explaining that the auth is there to prevent the user locking themselves out if they forgot the password.

I added a docstring to the `QEBiometrics.disableProtected()` func.

> Though note that if the system decides to clear the AndroidKeystore tokens [for whatever reason](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationRequired(boolean)), and the user forgot the wallet password, they can still get locked out:

true, adding this doesn't make it safe to forget the password, but I think it's still worth adding as it might prevent trouble for some at a low cost

electrum/gui/qml/components/main.qml
```diff
@@ -761,53 +817,52 @@ ApplicationWindow
...
-                handleAuthConfirmationOnly(qtobject, authMessage)
-            } else {
-                var dialog = app.pinDialog.createObject(app, {
-                    mode: 'check',
-                    pincode: Config.pinCode,
+        if (Daemon.currentWallet.verifyPassword('')) {
+            // wallet has no password
+            qtobject.authProceed()
+            return
+        }
```
SomberNight, line 834: I guess this is a pre-existing issue, but shouldn't we call `handleAuthConfirmationOnly` instead?

SomberNight (approved):
Looks good. Thanks!

electrum/gui/qml/components/main.qml
f321x: If the original call method was `payment_auth` yes we probably should (though wallets without password should be rare(?)). For `wallet` or `wallet_password_only` I don't think it would be useful.

SomberNight: > though wallets without password should be rare(?)

IIRC for several years, older versions of kivy did not allow setting a password for a watch-only wallet. So old installs, without password unification, could have those wallets still around.
Then again, if it's watch-only, not prompting for auth is not that critical.

PapaMarc:
just found a notification on this, and updated to 4.8.0.
Superb gents. kudos
Thanks!

Issue #9075, opened 2024-05-30
Wallet password unification on Android: enforce irreversibility


SomberNight:
On Android, in the past, each wallet could have a separate wallet password.
Then https://github.com/spesmilo/electrum/pull/6805 introduced the wallet password unification.
With that:
- every time a wallet is opened, or a wallet password is changed, we check if the password is or can be unified
- it is considered a legacy state to have multiple wallets that have different passwords
- if the password is unified, when the user creates a new wallet, we automatically use it for the new wallet
    - otherwise, the wizard lets the user type a new password for the new wallet

Note that before any wallet is opened, currently we don't know if the password is unified.
The kivy gui did not allow creation of new wallets until an existing wallet was opened (https://github.com/spesmilo/electrum/pull/6805#issuecomment-759370628). The qml gui however allows this atm. This lets the user reverse password unification by creating a new wallet with a different password.

Maybe we could revert to the kivy behaviour and disable the create wallet button (or show popup when it is clicked) until a wallet is opened.

---

Some problematic/impossible use cases/scenarios:
- user who forgot their wallet password will not be able to create a new wallet (only recourse is to reinstall the application)
- multiple people wanting to use the same phone and each have their own wallet with different passwords

SomberNight:
> Maybe we could revert to the kivy behaviour and disable the create wallet button (or show popup when it is clicked) until a wallet is opened.

Note: @ecdsa had an idea that we could even go a step further, and disallow creating new wallets at all if the password is not unified: show an error to the user telling them that they need to manually change all their passwords to be the same. This sounds a bit extreme though.

Sentini14:
I need Bitcoin address to open my account

nc50lc:
> Some problematic/impossible use cases/scenarios:
> 
> * user who forgot their wallet password will not be able to create a new wallet (only recourse is to reinstall the application)
> * multiple people wanting to use the same phone and each have their own wallet with different passwords

In QML, the user can tap outside of the password prompt (on the black space) and the "Open/Create Wallet" menu will be available.
The user(s) can select or create wallets there without unlocking the auto-loaded wallet.

SomberNight:
If the password is currently not unified, @f321x said during new wallet creation we could limit the choice of password to the set of already used passwords.

> Maybe we could revert to the kivy behaviour and disable the create wallet button (or show popup when it is clicked) until a wallet is opened.

[...]

PR #10345, opened 2025-12-05
qml: limit creation of new wallets to existing password

f321x:
Closes the loophole to create wallets with a different password than that of other existing wallets. This guides the user towards a single password for all wallets.

Fixes https://github.com/spesmilo/electrum/issues/9075

### DIFF
--- a/electrum/daemon.py
+++ b/electrum/daemon.py
@@ -562,6 +562,8 @@ class Daemon(Logger):
         if os.path.exists(path):
             os.unlink(path)
             self.update_recently_opened_wallets(path, remove=True)
+            if self.config.CURRENT_WALLET == path:
+                self.config.CURRENT_WALLET = None
             return True
         return False
 


electrum/gui/qml/components/Wallets.qml
```diff
@@ -121,7 +121,23 @@ Pane {
...
+            onClicked: {
+                if (Daemon.availableWallets.rowCount() > 0 && !Daemon.currentWallet && Config.walletUseSinglePassword) {
+                    // if the user has wallets but hasn't unlocked any wallet yet force them to do so.
+                    // this ensures they know at least one wallets password and can complete the wizard
+                    // where they will need to enter the password of an existing wallet.
+                    var dialog = app.messageDialog.createObject(app, {
+                        title: qsTr('Wallet unlock required'),
+                        text: [
+                            qsTr("You have to unlock any existing wallet first before creating a new wallet."),
+                            qsTr("If you forgot all passwords you can reinstall Electrum.")
```
accumulator: Reinstalling deletes the wallet files, maybe add a warning that to recover funds from lightning channels you need channel backups.

f321x: But its not possible to create a backup if the passwords are all forgotten, so this warning wouldn't be so useful as it is already too late to create a backup?

I added this reinstall string to show lost users that forgot all passwords a solution to get out of this situation.
I could remove the string completely and let the users who forgot all their passwords figure out the reinstalling step themselves to prevent motivating some users to reinstall even if not necessary?

accumulator: > But its not possible to create a backup if the passwords are all forgotten, so this warning wouldn't be so useful as it is already too late to create a backup?

Yes, but there's a grey zone, casual users that don't use the wallet often or know how lightning works exactly should not lightly decide to reinstall under the assumption a reinstall will assuredly recover all funds. 

> I added this reinstall string to show lost users that forgot all passwords a solution to get out of this situation. I could remove the string completely and let the users who forgot all their passwords figure out the reinstalling step themselves to prevent motivating some users to reinstall even if not necessary?

Maybe add a short 'Help I lost my password' button that opens a more elaborate explanation how to proceed, so there's no doom suggestion of reinstalling on the regular unlock screen?

electrum/gui/qml/qedaemon.py
```diff
@@ -321,12 +321,31 @@ def fx(self):
...
+            d.) we are on desktop where different passwords are allowed
+        """
         return self._use_single_password
 
     @pyqtProperty(str, notify=singlePasswordChanged)
     def singlePassword(self):
         return self._password
 
+    @pyqtSlot(str, result=int)
+    def walletsWithPassword(self, password: str) -> int:
```
SomberNight: maybe prefix name with `num` or `count` or `n`:
`numWalletsWithPassword`

electrum/gui/qml/qedaemon.py
```diff
@@ -321,12 +321,31 @@ def fx(self):
...
+        """Returns the number of wallets that can be unlocked with the given password"""
+        wallet_dir = os.path.dirname(self.daemon.config.get_wallet_path())
+        _, _, count_wallets_unlocked = self.daemon.check_password_for_directory(
```
SomberNight: `num_can_unlock`

electrum/daemon.py
```diff
@@ -708,7 +709,8 @@ def _check_password_for_directory(self, *, old_password, new_password=None, wall
                 wallet.update_password(old_password_real, new_password, encrypt_storage=True)
         can_be_unified = failed == []
         is_unified = can_be_unified and is_unified
-        return can_be_unified, is_unified
+        could_be_unlocked = len(wallet_filenames) - len(failed)
+        return can_be_unified, is_unified, could_be_unlocked
```
SomberNight: `could_be_unlocked` vs `can_be_unlocked` vs `count_wallets_unlocked`
You have three different names for this single thing already in the PR that introduces it :)

Also, if the name starts with `can` or `could` or `is`, etc, by convention it is usually a boolean, which is not the case here.

I suggest changing the name to e.g. `num_can_unlock`, but in any case, be consistent.

```suggestion
        num_can_unlock = len(wallet_filenames) - len(failed)
        return can_be_unified, is_unified, num_can_unlock
```

electrum/gui/qml/components/wizard/WCWalletPassword.qml
```diff
@@ -5,37 +5,70 @@ import QtQuick.Controls.Material
 
 import "../controls"
 
+// We will only end up here if Daemon.singlePasswordEnabled == False.
+// If there are existing wallets the user must not use a password different from at least
+// one of the other wallets. This way they are guided towards password unification.
```
SomberNight: This first sentence is really hard to parse and/or ambiguous.
If the existing passwords are A and B, then every possible password is different from at least one of the existing passwords, as A != B and B != A  :P

Just say:

> If there are existing wallets, the user must reuse the password of one of them.

SomberNight (commented):
I just skimmed this, but here are some nits to annoy you. :P  Sorry.

electrum/gui/qml/components/Wallets.qml
f321x: I removed now, i tried to add a info button to the message dialog but that kind of bloats the message dialog and on second thought this sentence doesn't seem very useful anyways. If the user really forgot all their passwords they will quickly notice that a reinstall is the only way to use the app again. All other users are not affected by this and would potentially only get confused by the message/info box, so i think just removing it is the simplest approach.

electrum/gui/qml/qedaemon.py
f321x: added in https://github.com/spesmilo/electrum/pull/10345/commits/ea35fe2452e1d00f60f78dfbd7c818b7a047fb6d

electrum/gui/qml/qedaemon.py
f321x: changed in https://github.com/spesmilo/electrum/pull/10345/commits/ea35fe2452e1d00f60f78dfbd7c818b7a047fb6d

electrum/daemon.py
f321x: Yeah right that got out of hand, changed it in https://github.com/spesmilo/electrum/pull/10345/commits/ea35fe2452e1d00f60f78dfbd7c818b7a047fb6d

electrum/gui/qml/components/wizard/WCWalletPassword.qml
f321x: Yeah true, changed the sentence.

ecdsa:
- In the initial password dialog, instead of failing if the password is does not correspond to the expected wallet, why not fallback to the first wallet that can be open with this password? this would be more user friendly, IMO.
- If the user clicks outside the dialog, we are taken to this "no wallet loaded" screen. This screen is useless (not blaming this PR for that, of course). If no wallet is open, we should be taken directly to the wallet list. This wallet list is useful if one desires to see the list and select a wallet before entering the password.

f321x:
@ecdsa :
> In the initial password dialog, instead of failing if the password is does not correspond to the expected wallet, why not fallback to the first wallet that can be open with this password? this would be more user friendly, IMO.

Doing this now in https://github.com/spesmilo/electrum/pull/10345/commits/890bd078a5bb4f751e2f02a6af0abbff5125cfb3

> If the user clicks outside the dialog, we are taken to this "no wallet loaded" screen. This screen is useless (not blaming this PR for that, of course). If no wallet is open, we should be taken directly to the wallet list. This wallet list is useful if one desires to see the list and select a wallet before entering the password.

Doing this now in https://github.com/spesmilo/electrum/pull/10345/commits/54c80d67bcf572731cde5a73ea32011168f92e09

f321x:
Now also removes the "No wallet loaded" logic from `WalletMainView.qml` and shows the `Wallets.qml` list directly as root view on startup.

electrum/gui/qml/components/WalletMainView.qml
f321x: note/todo: https://github.com/spesmilo/electrum/pull/10345/commits/75165423a20b2c5cdf3cd3c5a58ce98c7a7e7027 seems to break starting the app with an intent (e.g. scanning invoice qr code in another app and sharing the invoice with the Electrum app).

ecdsa: Maybe this Connection has to be declared in Wallets.qml, since this now is the first screen in the stack?
The uri needs to be saved where onWalletLoaded can access it

f321x: We can listen in main.qml and then just access a property of app in the `WalletMainView` once a wallet is loaded like this, however need to build an apk and verify this actually works:
```diff
diff --git a/electrum/gui/qml/components/WalletMainView.qml b/electrum/gui/qml/components/WalletMainView.qml
index a7d3e5558..bf1412ee8 100644
--- a/electrum/gui/qml/components/WalletMainView.qml
+++ b/electrum/gui/qml/components/WalletMainView.qml
@@ -14,7 +14,6 @@ Item {
     property string title: Daemon.currentWallet.name
 
     property var _sendDialog
-    property string _intentUri
 
     property string _request_amount
     property string _request_description
@@ -441,19 +440,6 @@ Item {
         id: bitcoin
     }
 
-    Connections {
-        target: AppController
-        function onUriReceived(uri) {
-            console.log('uri received: ' + uri)
-            if (!Daemon.currentWallet) {
-                console.log('No wallet open, deferring')
-                _intentUri = uri
-                return
-            }
-            piResolver.recipient = uri
-        }
-    }
-
     Connections {
         target: Daemon
         function onWalletLoaded() {
@@ -462,9 +448,19 @@ Item {
                 return
             }
             infobanner.hide() // start hidden when switching wallets
-            if (_intentUri) {
-                piResolver.recipient = _intentUri
-                _intentUri = ''
+            if (app.pendingIntent) {
+                piResolver.recipient = app.pendingIntent
+                app.pendingIntent = ""
+            }
+        }
+    }
+
+    Connections {
+        target: app
+        function onPendingIntentChanged() {
+            if (app.pendingIntent) {
+                piResolver.recipient = app.pendingIntent
+                app.pendingIntent = ""
             }
         }
     }
diff --git a/electrum/gui/qml/components/main.qml b/electrum/gui/qml/components/main.qml
index 2b18f2aa4..c8d364525 100644
--- a/electrum/gui/qml/components/main.qml
+++ b/electrum/gui/qml/components/main.qml
@@ -38,6 +38,8 @@ ApplicationWindow
     property alias keyboardFreeZone: _keyboardFreeZone
     property alias infobanner: _infobanner
 
+    property string pendingIntent: ""
+
     property variant activeDialogs: []
 
     property var _exceptionDialog
@@ -708,6 +710,10 @@ ApplicationWindow
             if (obj != null)
                 app.pluginobjects[name] = obj
         }
+        function onUriReceived(uri) {
+            console.log('uri received (main): ' + uri)
+            app.pendingIntent = uri
+        }
     }
 
     function pluginsComponentsByName(comp_name) {
```

f321x: Resolved in https://github.com/spesmilo/electrum/pull/10345/commits/70084750ef81c5ed29edc3e42bb6ec6e829db9a7

ecdsa:
thanks

PR #10346, opened 2025-12-05
lnchannel: allow deleting unfunded incoming channels

f321x:
We tried to delete incoming channels that didn't get funded after `lnutil.CHANNEL_OPENING_TIMEOUT`, however an assert prevented this. 
This implements the [2016 block limit of bolt 2](https://github.com/lightning/bolts/commit/ba00bf8f4cd85f21bacfc03adcafd4acc7d68382) after which an unconfirmed incoming channel can be dropped, and removes the old timestamp based timeout of 24h.

```
  3.63 | E | lnwatcher.LNWatcher.[default_wallet-LNW] | Exception in check_onchain_situation: AssertionError()
  File "/home/user/code/electrum-fork/electrum/util.py", line 1233, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/lnwatcher.py", line 117, in check_onchain_situation
    await self.update_channel_state(
    ...<5 lines>...
        keep_watching=keep_watching)
  File "/home/user/code/electrum-fork/electrum/lnwatcher.py", line 135, in update_channel_state
    chan.update_onchain_state(
    ~~~~~~~~~~~~~~~~~~~~~~~~~^
        funding_txid=funding_txid,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        closing_height=closing_height,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        keep_watching=keep_watching)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/lnchannel.py", line 341, in update_onchain_state
    self.update_unfunded_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/code/electrum-fork/electrum/lnchannel.py", line 382, in update_unfunded_state
    self.lnworker.remove_channel(self.channel_id)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/user/code/electrum-fork/electrum/lnworker.py", line 3244, in remove_channel
    assert chan.can_be_deleted()
           ~~~~~~~~~~~~~~~~~~~^^
AssertionError
```

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -60,7 +60,7 @@ from .lnsweep import sweep_their_ctx_to_remote_backup
 from .lnhtlc import HTLCManager
 from .lnmsg import encode_msg, decode_msg
 from .address_synchronizer import TX_HEIGHT_LOCAL
-from .lnutil import CHANNEL_OPENING_TIMEOUT
+from .lnutil import CHANNEL_OPENING_TIMEOUT_BLOCKS, CHANNEL_OPENING_TIMEOUT_SEC
 from .lnutil import ChannelBackupStorage, ImportedChannelBackupStorage, OnchainChannelBackupStorage
 from .lnutil import format_short_channel_id
 from .fee_policy import FEERATE_PER_KW_MIN_RELAY_LIGHTNING
@@ -356,7 +356,6 @@ class AbstractChannel(Logger, ABC):
         self.delete_closing_height()
         if not self.lnworker:
             return
-        chan_age = now() - self.storage.get('init_timestamp', 0)
         state = self.get_state()
         if state in [ChannelState.PREOPENING, ChannelState.OPENING, ChannelState.FORCE_CLOSING]:
             if self.is_initiator():
@@ -377,11 +376,11 @@ class AbstractChannel(Logger, ABC):
                             self.logger.info(f'channel is double spent {inputs}')
                             self.set_state(ChannelState.REDEEMED)
                             break
-            else:
-                if chan_age > CHANNEL_OPENING_TIMEOUT:
-                    self.lnworker.remove_channel(self.channel_id)
+            elif self.has_funding_timed_out():
+                self.logger.warning(f"dropping incoming channel, funding tx not found in mempool")
+                self.lnworker.remove_channel(self.channel_id)
         elif self.is_zeroconf() and state in [ChannelState.OPEN, ChannelState.CLOSING, ChannelState.FORCE_CLOSING]:
-            assert self.storage.get('init_timestamp') is not None, "init_timestamp not set for zeroconf channel"
+            chan_age = now() - self.storage['init_timestamp']
             # handling zeroconf channels with no funding tx, can happen if broadcasting fails on LSP side
             # or if the LSP did double spent the funding tx/never published it intentionally
             # only remove a timed out OPEN channel if we are connected to the network to prevent removing it if we went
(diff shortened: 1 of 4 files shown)


f321x:
Shouldn't `lnutil.CHANNEL_OPENING_TIMEOUT = 24*60*60
` be higher, like 2 weeks?

[BOLT-2 says](https://github.com/lightning/bolts/blob/0cf21511a781c295b9374aeaef37cf9c4d193502/02-peer-protocol.md?plain=1#L1039-L1041):

> A non-funding node (fundee):
     - SHOULD forget the channel if it does not see the correct funding
       transaction after a timeout of 2016 blocks.
       
EDIT:
seems like the timeout was added afterwards:
https://github.com/lightning/bolts/commit/ba00bf8f4cd85f21bacfc03adcafd4acc7d68382
Will work on block based timeout.

electrum/lnchannel.py
```diff
@@ -830,12 +842,52 @@ def get_remote_scid_alias(self) -> Optional[bytes]:
...
+        if 'init_height' in self.storage:
+            return self.storage['init_height']
+        init_timestamp = self.storage.get('init_timestamp')
+        if init_timestamp is None:
+            return None
+        blockchain = self.lnworker.network.blockchain()
+        if blockchain.is_tip_stale():
+            return None
+
+        # binary search through headers for old channels that have been established before init_height
```
f321x: This could also be a wallet upgrade or simple lambda like: 
`guess_height = lambda: max(0, 920000 + (self.storage['init_timestamp'] - 1760995477) // 580))`
Not sure if its worth to have this, though it might be useful if we rely on `init_height` for something else in the future.

SomberNight: > This could also be a wallet upgrade

When the DB upgrade runs, the block headers might not be available. E.g. moved wallet file to fresh electrum dir.

---

IMO this binary search here is overkill :D

In `chan.has_funding_timed_out()`, we could wait until both 2016 blocks AND 2 weeks' time has passed, whichever is longer,
and refer to init_height as `storage.get('init_height', 0)` (or do a DB upgrade setting `init_height` to 0).

> it might be useful if we rely on init_height for something else in the future.

If we don't do the DB upgrade or do it but set `init_height` to 0, we could always do another DB upgrade later and correct the zeroes to guessed heights based on the timestamp. Though I don't foresee a usecase atm.

f321x: > When the DB upgrade runs, the block headers might not be available. E.g. moved wallet file to fresh electrum dir.

Yeah i thought of some extrapolation like in the example before.

> In chan.has_funding_timed_out(), we could wait until both 2016 blocks AND 2 weeks' time has passed, whichever is longer

Ok, doing this now in https://github.com/spesmilo/electrum/pull/10346/commits/c34efce984433a627e9833edb460bbe4d4ab6389, it's much simpler. Originally i tried to avoid 'violating' the spec, but that is probably too pedantic.

SomberNight: >>   - SHOULD forget the channel if it does not see the correct funding
>>      transaction after a timeout of 2016 blocks.
>
> Originally i tried to avoid 'violating' the spec, but that is probably too pedantic.

My interpretation of the spec here is that "we should wait at least 2016 blocks before forgetting the channel, waiting more is fine, however at some point after a reasonable time we really should forget it".

PR #10347, opened 2025-12-05
electrum refuses to exchange gossip with CLN nodes, due to missing feature OPTION_CHANNEL_TYPE_OPT

accumulator:
Not sure if we should advertise this feature or if CLN should relax gossip-only connections, but this PR is a 
workaround, adds CHANNEL_TYPE to BASE_FEATURES.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -162,7 +162,7 @@ class ErrorAddingPeer(Exception): pass
 
 
 # set some feature flags as baseline for both LNWallet and LNGossip
-# note that e.g. DATA_LOSS_PROTECT is needed for LNGossip as many peers require it
+# note that e.g. DATA_LOSS_PROTECT and OPTION_CHANNEL_TYPE_OPT are needed for LNGossip as many peers require it
 BASE_FEATURES = (
     LnFeatures(0)
     | LnFeatures.OPTION_DATA_LOSS_PROTECT_OPT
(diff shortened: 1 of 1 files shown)


electrum/lnworker.py
```diff
@@ -193,6 +193,7 @@ class ErrorAddingPeer(Exception): pass
     # that peers satisfy our queries
     | LnFeatures.GOSSIP_QUERIES_REQ
     | LnFeatures.GOSSIP_QUERIES_OPT
+    | LnFeatures.OPTION_CHANNEL_TYPE_OPT
```
SomberNight: It would make more sense to set this in BASE_FEATURES, instead of setting it in both `LNWALLET_FEATURES` and `LNGOSSIP_FEATURES`. Just remove it from both, and add it to BASE.

Pretty much all the features in base features are there with the same motivation:
`DATA_LOSS_PROTECT` / `VAR_ONION` / `PAYMENT_SECRET`: if you don't set them, you can't even connect to modern peers.
https://github.com/spesmilo/electrum/blob/55cc27d27e0e3e0f54d329c8c3f04567458c9c21/electrum/lnworker.py#L164-L172

SomberNight: btw you can make these code comments more descriptive if you want/can

accumulator: not sure what I can add now to comments except adding a reason for `OPTION_CHANNEL_TYPE_OPT`

ecdsa:
LGTM

SomberNight:
Thanks

PR #10351, opened 2025-12-08
lnpeer/lnutil: fail mpp if we didn't signal mpp in invoice

f321x:
Followup of  #10348
Fails incoming htlcs if we requested a single part payment by disabling the mpp flag in the invoice features but the sender sent us multiple htlcs anyways. This is done by storing the invoice features in the PaymentInfo class and checking against them when receiving the htlcs.

Also changes the channel reference in `ReceivedMPPHtlc` from `scid` to the full `channel_id` to simplify it and prevent issues with just in time channels where the scid can be None as the channel is used before a funding tx has been mined. (Alternatively we could use the local scid alias, but that seems more error prone).

### DIFF
--- a/electrum/wallet_db.py
+++ b/electrum/wallet_db.py
@@ -1325,7 +1325,7 @@ class WalletDBUpgrader(Logger):
         old_payment_infos = self.data.get('lightning_payments', {})
         for key, old_v in old_payment_infos.items():
             amount_msat, status, min_final_cltv_expiry, expiry, creation_ts = old_v
-            invoice_features = 147712  # <VAR_ONION_REQ|PAYMENT_SECRET_REQ|BASIC_MPP_OPT: 0x24100>
+            invoice_features = 0x24100  # <VAR_ONION_REQ|PAYMENT_SECRET_REQ|BASIC_MPP_OPT>
             new_v = (amount_msat, status, min_final_cltv_expiry, expiry, creation_ts, invoice_features)
             new_payment_infos[key] = new_v
 


electrum/lnpeer.py
```diff
@@ -3074,8 +3075,11 @@ def _check_unfulfilled_htlc_set(
 
         if mpp_set.resolution == RecvMPPResolution.WAITING:
             # calculate the sum of just in time channel opening fees
-            htlc_channels = [self.lnworker.get_channel_by_short_id(scid) for scid in set(h.scid for h in mpp_set.htlcs)]
+            htlc_channels = [self.lnworker.channels[channel_id] for channel_id in set(h.channel_id for h in mpp_set.htlcs)]
             jit_opening_fees_msat = sum((c.jit_opening_fee or 0) for c in htlc_channels)
+            if jit_opening_fees_msat and len(mpp_set.htlcs) > 1:
+                # we don't support mpp with just-in-time channel openings, the payer has to send a single htlc
+                return OnionFailureCode.INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS, None, None
```
SomberNight: Ideally we should know whether the invoice we created signalled MPP, and just check against that.
That would work if e.g. `invoice_features` were part of `payment_info`. In that case the check could even be done on the individual htlc in `_check_unfulfilled_htlc`.
but ok, not sure if we want to do that now

electrum/lnutil.py
```diff
@@ -1965,18 +1965,18 @@ class RecvMPPResolution(IntEnum):
...
 
 class ReceivedMPPHtlc(NamedTuple):
-    scid: ShortChannelID
+    channel_id: bytes
     htlc: UpdateAddHtlc
     unprocessed_onion: str
 
     def __repr__(self):
-        return f"{self.scid}, {self.htlc=}, {self.unprocessed_onion[:15]=}..."
+        return f"{self.channel_id}, {self.htlc=}, {self.unprocessed_onion[:15]=}..."
```
SomberNight: I guess, in repr, we should use hex.

```suggestion
        return f"chan_id={self.channel_id.hex()}, {self.htlc=}, {self.unprocessed_onion[:15]=}..."
```

f321x: Fixed in https://github.com/spesmilo/electrum/pull/10351/commits/5be598b808457ad917242053a081551a1261521d

electrum/lnpeer.py
f321x: Doing that now in https://github.com/spesmilo/electrum/pull/10351/commits/0ce2fa830b6baf3975c4d4fa911c0c788b174779. 
It seems cleaner and more generic.
The persisted features could get stale if the user changes the config (e.g. from gossip to trampoline), then shows the invoice again and tries to pay it, though i guess thats acceptable

SomberNight (approved):
Looks good. Thanks!

PR #10353, opened 2025-12-10
lnurlw: catch UserCancelled and handle 'null' minWithdrawable

f321x:
Accept a `null` value as response for the `minWithdrawable` field in the
lnurlw response. Some servers seem to set this to `null` instead of 0
when having no minimum withdrawal amount.

Catch the UserCancelled exception if the user cancels the lnurlw
coroutine dialog during the withdrawal.


SomberNight:
Thanks!

PR #10354, opened 2025-12-10
lnworker: prevent creation of PaymentInfo with 0 exp

### LINKED ISSUE #10350: regression: cannot create lightning invoice that never expires
### Description

if I try to create a LN invoice without expiry, I get this:

```
  File "/opt/electrum/electrum/lnworker.py", line 140, in validate
    assert isinstance(self.expiry_delay, int) and self.expiry_delay > 0
AssertionError

```
this regression was introduced in 2b0cab6
not sure how we should handle it.

f321x:
In some parts of the application 0 (sec) == no expiry, however we use `LN_EXPIRY_NEVER` (100 years) instead of 0 for lightning invoices.

This replaces a 0 second expiry with `LN_EXPIRY_NEVER` in `LNWallet.create_payment_info()` to fix the assertion error in #10350 and prevent htlcs for no-expiry invoices from getting failed incorrectly (which the assert prevented).

Fixes #10350.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -146,7 +146,7 @@ class PaymentInfo:
         assert isinstance(self.direction, int)
         assert isinstance(self.status, int)
         assert isinstance(self.min_final_cltv_delta, int)
-        assert isinstance(self.expiry_delay, int) and self.expiry_delay > 0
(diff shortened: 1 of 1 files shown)


SomberNight:
needs rebase

PR #10356, opened 2025-12-10
tx: replace all whitespace chars in raw tx (convert_raw_tx_to_hex)

f321x:
Removes all whitespace characters from a raw transaction string in `convert_raw_tx_to_hex()`. This is useful for example when loading raw transactions from text input as it happens that there are some newline characters in the text. I noticed this when copying-pasting from a timelock recovery pdf.

E.g. this is a single tx, removing the newlines manually is quite annoying:
```
02000000000104AA2F554D78B5B785B52CC396B20DC1259A63438D9B9851B717C232EA4CD7A41B0000000000FDFFFFFF5B0F29
67003923F338AA0CEB628C0012329881DD98DBB4A17D8F798A1FA7FE330000000000FDFFFFFFD164BEB8792B7871DBFB17FAE1
FB8C103F566A6DA53B8D137A87EE13AB029AB50100000000FDFFFFFF667821EE01F47A9F833B6298A2817AF82D4767901C42CD
9885179E1D84F493E80100000000FDFFFFFF0358020000000000001600140D10180CA80C936E8060229FB4754F4174F5B06E58
020000000000001600144FC93671899237C42AE13A0DEA50D1838822FCAE65981F00000000001600146E777C49135536FA4F40
B6D15CF0D2E5389B90D302473044022059EA1286E440A94E086334192B0DC8FA3B48183D87D91987230F7DDBECCA794402207A
1A88A4146A9E46A486720FD849D861490E6F267527A9D1DB65D11B1A42F8BA012102D1589A49839AD6C9E88770B6E6B9069FC4
04AEF6F5F27DF92C3C336843ECFDAC02473044022009CBC46F47F7F27F6BED007D4DD38FF861DDDF4537322CE8E5705B73D71A
AB27022052DAF39F31A72CA705C7C75CCC6C005FB77074A309892D093F662A55769CB98A012103E92383B64B878F55210EDBD2
CD2FBF54760E6B02A180605CC981438D8BE4EF1E02473044022059DA0981CC3349325B80510B74D7B3D8029A3E74ED81307379
542121FAC4AFF20220084A8B998FF492A1BB85CF2CF4F829465AE4C377C1E4E8BCA36ED18777093BD0012102481C80E2785E84
6C47B2008F1B977C24931447021CC119C59B08FFFDDA09363A0247304402203110947852D71A2E743F4136D65E594EAE22C8BB
55145F2E70D3E0B8EBA6985B022022C4DF0C8FD5C9CA41A94427FA8888A0FC3693C3DFBADF7540733DB088EC4ED701210263F2
46322CB912A056634890AC2142C1CDDC7BEEBBFD4D37ABCA7B5CDB60ABC22C032900
```


electrum/transaction.py
```diff
@@ -1482,6 +1482,8 @@ def convert_raw_tx_to_hex(raw: Union[str, bytes]) -> str:
     if not raw:
         raise ValueError("empty string")
     raw_unstripped = raw
+    if isinstance(raw, str):
+        raw = raw.replace('\n', '').replace('\r', '')
```
SomberNight: can you add a test? maybe extend this one:
https://github.com/spesmilo/electrum/blob/8e5780b98cc03438df3a35d1352ba36b282d790e/tests/test_transaction.py#L250

f321x: In https://github.com/spesmilo/electrum/pull/10356/commits/7d307048a07220006c75f90b2d636d2111e7571d I changed my patch to remove all whitespace chars from the input, independent of the position and adapted the `test_tx_from_any` to also insert whitespace into the transaction input.

electrum/transaction.py
```diff
@@ -1482,7 +1483,11 @@ def convert_raw_tx_to_hex(raw: Union[str, bytes]) -> str:
     if not raw:
         raise ValueError("empty string")
     raw_unstripped = raw
-    raw = raw.strip()
+    # remove all whitespace characters
+    if isinstance(raw, str):
+        raw = re.sub(r'\s', '', raw)
+    else:
+        raw = re.sub(rb'\s', b'', raw)
```
SomberNight: it is not safe to remove bytes that look like whitespace characters from a bytes-like raw tx input.
However note we were already doing this for leading/trailing bytes. That is not safe either. E.g. the last byte of the nLocktime might look like a whitespace character - it should not be stripped.
I added follow-up commits to address this.

PR #10363, opened 2025-12-12
tests: add test_lnwallet to test lnwallet utils

f321x:
It would be nice to have some way to test the (utility) methods of `LNWallet` to allow for more coverage and detect regressions like https://github.com/spesmilo/electrum/issues/10350 before they reach the GUI etc.
So i added test_lnwallet.py as a place to test `LNWallet` things like the creation of `PaymentInfo`.

### DIFF
--- a/tests/test_lnwallet.py
+++ b/tests/test_lnwallet.py
@@ -1,0 +1,45 @@
+import logging
+import os
+
+from . import ElectrumTestCase
+
+from electrum.lnutil import RECEIVED, MIN_FINAL_CLTV_DELTA_ACCEPTED
+from electrum.logging import console_stderr_handler
+from electrum.invoices import LN_EXPIRY_NEVER, PR_UNPAID
+
+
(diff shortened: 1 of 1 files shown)


ecdsa:
I would very much prefer if we stopped mocking up LNWallet. 
For example, we could separate LNWallet from LNWorker. 
LNWorker should manage peer-to-peer communications, and be mocked up in unit tests. LNWallet would not be mocked up.

Of course what I am saying is not conflicting with this pull request. 
But this PR builds more stuff upon MockLNWallet...

f321x:
>  LNWorker should manage peer-to-peer communications, and be mocked up in unit tests. LNWallet would not be mocked up.

Ok, then lets do this instead of extending dependency on the mock.

f321x:
This now uses the new `LNWallet` architecture introduced in https://github.com/spesmilo/electrum/pull/10372

Issue #10273, opened 2025-10-21
GUI to always allow signing low/zero-fee txs


utxocrane:
### Description
Users cannot set a fee rate below 1 sat/vB, even though the Electrum server(eg. blockstream.info:700:s) accepts values as low as 0.1 sat/vB.

As the minimum fee rate in major mempools (e.g., mempool.space) has dropped to 0.1 sat/vB at 15.Jul,2025 , it's strongly recommended to add a "Manual" FeeRate mode that allows users to specify custom values(stepsize=1sat/tx rather than 1sat/vb) , including precise rates like 0.xxxxxxxx sat/vB.
DO NOT report low fee error until the electrum server reject the tx, and just output the tx hex to let user treat it instead !

For those who need a simple tool to fully control transaction fees, check out [PsbtCreator](https://utxocrane.github.io/PsbtToolsHtml/creator.html) and broadcast signed tx via mempool's website ui.

Electrum server like mempool.8333.mobi:50002:s contains more mempool txs, this is helpful to view tx's correct TIP ,check https://1209k.com/bitcoin-eye/ele.php for more electrum server.

SomberNight:
> add a "Manual" FeeRate mode that allows users to specify custom values(stepsize=1sat/tx rather than 1sat/vb) , including precise rates like 0.xxxxxxxx sat/vB.

The Qt GUI already allows specifying custom absolute fees in satoshis. Can't get higher resolution than that.

>  DO NOT report low fee error until the electrum server reject the tx

The GUI works in a reasonable way atm I think. What specifically is your scenario? If you give more details re how/what you are doing, perhaps we can improve the UX flow.

utxocrane:
> > add a "Manual" FeeRate mode that allows users to specify custom values(stepsize=1sat/tx rather than 1sat/vb) , including precise rates like 0.xxxxxxxx sat/vB.
> 
> The Qt GUI already allows specifying custom absolute fees in satoshis. Can't get higher resolution than that.
> 
> > DO NOT report low fee error until the electrum server reject the tx
> 
> The GUI works in a reasonable way atm I think. What specifically is your scenario? If you give more details re how/what you are doing, perhaps we can improve the UX flow.

The UI is not showing the TX Preview(Button is Disabled) with low fee error, but I'd like to broadcast the tx raw via other mempool's websites, so just show the tx hex(base64), and users can view/copy the tx script even if electrumx server reject !

PR #10371, opened 2025-12-17
qml: allow manual editing of fee/feerate

### LINKED ISSUE #6710: Android: Allow manually editing fee rates
On Android (Kivy) it is not possible to edit fees manually.  It would be nice to have a simple text box above the slider with sat / byte description.

Currently on Kivy you can only use the slider and see what fee rate it pre-filled for you, you cannot manually edit the fee rate manually for any of the fee methods available (Static, ETA, Mempool).

#5507 related to Android / Kivy fees.

accumulator:
also improve warning box styling consistency across finalizers, 
add CPFP new feerate > old feerate check,
add relayfee checks for CPFP, DSCancel,
proper warning string for no dynamic fee estimates.

fixes #6710 #10273

### DIFF
--- a/electrum/gui/qml/components/controls/FeePicker.qml
+++ b/electrum/gui/qml/components/controls/FeePicker.qml
@@ -18,6 +18,8 @@ Item {
 
     property bool showTxInfo: true
     property bool showPicker: true
+    property bool allowPickerAbsFees: true
+    property bool allowPickerFeeRates: true
 
     property bool manualFeeEntry: finalizer.method == FeeSlider.FSMethod.MANUAL
 
@@ -121,24 +123,22 @@ Item {
             }
         }
 
-        Label {
+        RowLayout {
+            Layout.columnSpan: 2
             Layout.fillWidth: true
-            Layout.preferredWidth: 1
-            text: qsTr('Rate')
-            color: Material.accentColor
-            visible: showPicker && manualFeeEntry
-        }
+            visible: showPicker && manualFeeEntry && allowPickerFeeRates
 
-        GridLayout {
-            Layout.preferredWidth: 2
-            Layout.rowSpan: 2
-            visible: showPicker && manualFeeEntry
-            columns: 2
-            columnSpacing: constants.paddingMedium
+            Label {
+                Layout.fillWidth: true
+                Layout.preferredWidth: 1
+                text: qsTr('Rate')
+                color: Material.accentColor
+            }
 
             TextField {
                 id: rate
                 Layout.fillWidth: true
+                Layout.preferredWidth: 2
                 text: finalizer.userFeerate
                 color: finalizer.isUserFeerateLast ? Material.foreground : Material.accentColor
                 inputMethodHints: Qt.ImhDigitsOnly
@@ -152,13 +152,28 @@ Item {
 
             Label {
                 Layout.fillWidth: true
+                Layout.preferredWidth: 1
                 color: Material.accentColor
                 text: UI_UNIT_NAME.FEERATE_SAT_PER_VBYTE
+            }
+        }
+
+        RowLayout {
+            Layout.columnSpan: 2
+            Layout.fillWidth: true
+            visible: showPicker && manualFeeEntry && allowPickerAbsFees
+
+            Label {
+                Layout.fillWidth: true
+                Layout.preferredWidth: 1
+                color: Material.accentColor
+                text: qsTr('Total')
             }
 
             TextField {
                 id: absolute
                 Layout.fillWidth: true
+                Layout.preferredWidth: 2
                 text: finalizer.userFee
                 color: finalizer.isUserFeerateLast ? Material.accentColor : Material.foreground
                 inputMethodHints: Qt.ImhDigitsOnly
@@ -172,17 +187,10 @@ Item {
 
             Label {
                 Layout.fillWidth: true
+                Layout.preferredWidth: 1
                 color: Material.accentColor
                 text: UI_UNIT_NAME.FIXED_SAT
             }
-        }
-
-        Label {
-            Layout.fillWidth: true
-            Layout.preferredWidth: 1
-            visible: showPicker && manualFeeEntry
-            color: Material.accentColor
-            text: qsTr('Total')
         }
 
     }
--- a/electrum/gui/qml/components/CpfpBumpFeeDialog.qml
+++ b/electrum/gui/qml/components/CpfpBumpFeeDialog.qml
@@ -147,6 +147,7 @@ ElDialog {
                             Layout.fillWidth: true
                             finalizer: dialog.cpfpfeebumper
                             showTxInfo: false
+                            allowPickerFeeRates: false
                         }
                     }
                 }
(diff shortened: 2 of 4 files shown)


accumulator:
marked as draft, missing filtering invalid alphanumeric input

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -661,12 +725,16 @@ def update(self):
             # not initialized yet
             return
 
-        fee_per_kb = self._fee_policy.fee_per_kb(self._wallet.wallet.network)
-        if fee_per_kb is None:
-            # dynamic method and no network
-            self._logger.debug('no fee_per_kb')
-            self.warning = _('Cannot determine dynamic fees, not connected')
-            return
+        if self._fee_policy.method == FeeMethod.FIXED:
```
f321x, line 742: this block is nearly duplicated 3 times, maybe it can be moved to the `TxFeeSlider` as `TxFeeSlider.get_fee_per_kb(tx_size)`?

accumulator:
> we seem to allow broadcasting a tx on QML when showing a warning (e.g. too low relayfee), which then fails with a server error, while disabling the Broadcast/Pay button in Qt. Should this be unified?

This sounds like a bug

electrum/gui/qml/components/controls/FeePicker.qml
```diff
@@ -19,6 +19,8 @@ Item {
     property bool showTxInfo: true
     property bool showPicker: true
 
+    property bool manualFeeEntry: finalizer.method == FeeSlider.FSMethod.MANUAL
```
SomberNight, line 24: I think we could hide the line containing `targetLabel` when in `manualFeeEntry` mode.

For reference, this is how it looks atm without hiding that line:
<details>
 <summary>pic</summary>

Screenshot_20260112_184407
</details>

electrum/gui/qml/components/controls/FeePicker.qml
```diff
@@ -117,5 +120,66 @@ Item {
...
+                }
+                onTextEdited: {
+                    finalizer.userFeerate = text
+                }
+            }
+
+            Label {
+                Layout.fillWidth: true
+                color: Material.accentColor
+                text: qsTr('sat/vbyte')
```
SomberNight: should use the existing named constants (UI_UNIT_NAME.FEERATE_SAT_PER_VBYTE)

electrum/gui/qml/components/controls/FeePicker.qml
```diff
@@ -117,5 +120,66 @@ Item {
...
+                text: qsTr('sat')
+            }
+        }
+
+        Label {
+            Layout.preferredWidth: 1
+            visible: showPicker && manualFeeEntry
+            color: Material.accentColor
+            text: qsTr('Total')
+        }
```
SomberNight: This layout seems very brittle and *magic* o.O
So there is a grid with 2 columns: we put the "Rate" label into it. Then we put a sub-grid into it which spans two rows. Then we put the "Total" label into it, expecting it to be just under "Rate" and to the left of the sub-grid. o.O
It works atm but if I try to do the subtlest change, it breaks :/

electrum/gui/qml/components/controls/FeePicker.qml
```diff
@@ -117,5 +120,66 @@ Item {
...
+                Layout.fillWidth: true
+                text: finalizer.userFeerate
+                inputMethodHints: Qt.ImhDigitsOnly
+                validator: RegularExpressionValidator {
+                    regularExpression: /^[0-9]*\.[0-9]?$/
+                }
+                onTextEdited: {
+                    finalizer.userFeerate = text
+                }
+            }
```
SomberNight, line 151: Note that the desktop Qt GUI does not always expose editing both the feerate and abs fee.
- for simple-sends, it exposes both
- for RBF bump-fee and dscancel, it only exposes feerate
- for CPFP, it only exposes abs fee

This is because the underlying wallet.py methods are limited accordingly. E.g. `wallet.bump_fee` wants a feerate.

The problem with exposing both as done by this PR currently is that it is lying to the user in a kind of unexpected way. Try the RbfBumpFeeDialog: in manual fee edit mode, start increasing the absolute fee by 1 satoshi a time: +1 sat, +1 sat, +1 sat. Observe that the actual abs fee does not increase, only occasionally. The calculation relies on the feerate with a 0.1 sat/vbyte precision to be an intermediate value: so the actual abs fee can only be changed with a granularity of ~15 sats (depending on tx size -- it is only the feerate that the user can truly change)

somewhat related: https://github.com/spesmilo/electrum/commit/ce8f564fb48d2aa0d22a1d6162ed165e4ea9f4f3

electrum/gui/qml/components/controls/FeePicker.qml
SomberNight: I pushed a commit to your branch to fix this: https://github.com/spesmilo/electrum/pull/10371/commits/f387300ab21fab37374b4b1e48eb13c285abecbc

electrum/gui/qml/components/controls/FeePicker.qml
SomberNight: Pushed a commit to your branch with this: https://github.com/spesmilo/electrum/pull/10371/commits/65f245f475eaeb9884bf7b32fab5d9a2f638a951

I think there are too many fields with some redundant values, so with this commit I am trying to remove at least one of them. Re the alternating coloring, it's the best I could come up with. We use a similar approach in the desktop GUI (coloring derived fields vs active target fields differently).

electrum/gui/qml/components/controls/FeePicker.qml
SomberNight: I have a patch that tries to tackle this, but it breaks the layout (hence my other comment that the layout is brittle :D)
See https://github.com/spesmilo/electrum/commit/154b7047efa5ad35daceb58ccd3850d048a6e547

electrum/gui/qml/components/controls/FeePicker.qml
accumulator: Yeah I was in doubt whether to leave it in or remove it. Fine with removing it.

electrum/gui/qml/components/controls/FeePicker.qml
SomberNight: done in https://github.com/spesmilo/electrum/pull/10371/commits/8a3d9fd75869be974a429c8be28c873f9b5c8a82

electrum/gui/qml/components/controls/FeePicker.qml
SomberNight: I changed the layout in https://github.com/spesmilo/electrum/pull/10371/commits/8a3d9fd75869be974a429c8be28c873f9b5c8a82

it looks a bit worse as it is not column-aligned :(  but at least it behaves predictably >.<

electrum/gui/qml/qetxfinalizer.py
SomberNight: note: I had a brief look and it is not trivial to deduplicate

SomberNight (approved):
Thanks a lot for the work here.

It only took us 10 years to add precise fee selection to the Android app! :D
I think a lot of users will be happy :)

PR #10372, opened 2025-12-17
lnworker: split LNWallet and LNWorker: LNWallet "has an" LNWorker

SomberNight:
This is a major simplification re how we are mocking things in `test_lnpeer.py` and friends.

- LNWallet no longer "is-an" LNWorker, instead LNWallet "has-an" LNWorker
- the motivation is to make the unit tests nicer, and allow writing unit tests for more things
  - I hope this makes it possible to e.g. test lnsweep in the unit tests
  - some stuff we would previously have to write a regtest for, maybe we can write a unit test for, now
- in unit tests,
  - MockLNWallet is now almost a real LNWallet: it simply inherits LNWallet
  - MockWallet is now almost a real Abstract_Wallet: it simply inherits Abstract_Wallet

---

not done:
- we could rename `{wallet,chan}.lnworker` to `*.lnwallet`
- but e.g. `peer.lnworker` is polymorphic between `LNWallet` and `LNGossip`
  - `LNPeerManager._lnwallet_or_lngossip` has the same polymorphism
- could split `lnworker.py` (LNWallet is 3000 lines out of 4000 of that file) but that would be really annoying for some branches/PRs

### DIFF
--- a/tests/test_lnpeer.py
+++ b/tests/test_lnpeer.py
@@ -58,14 +58,6 @@ from .test_bitcoin import needs_test_with_all_chacha20_implementations
 from . import ElectrumTestCase, restore_wallet_from_text__for_unittest
 
 
-def keypair():
-    priv = ECPrivkey.generate_random_key().get_secret_bytes()
-    k1 = Keypair(
-            pubkey=privkey_to_pubkey(priv),
-            privkey=priv)
-    return k1
-
-
 class MockNetwork:
     def __init__(self, *, config: SimpleConfig):
         self.lnwatcher = None


SomberNight:
This is ready for review. There is potential for follow-ups ofc, but I think it is in a good state we could merge for incremental progress.

SomberNight:
rebased from 2ee0c55a25007e82a4e160ac2db36d6d6f5500f8 to e0284e83ba6ee143e346c840cbf95b2c9f569a6d

ecdsa:
I would prefer to merge this asap, because of possible conflicts

electrum/lnworker.py
```diff
@@ -600,15 +665,44 @@ def __init__(self, config: 'SimpleConfig'):
         self._num_chan_upd = 0
         self._num_chan_upd_good = 0
 
+    @property
+    def features(self) -> 'LnFeatures':
+        return self.lnpeermgr.features
+
+    @property
+    def network(self) -> 'Network':
```
f321x: Isn't this Optional? same for `.channel_db`
```suggestion
    def network(self) -> Optional['Network']:
```

SomberNight: you are correct.
fixed in rebase https://github.com/spesmilo/electrum/compare/e0284e83ba6ee143e346c840cbf95b2c9f569a6d..927724145a0c9328e5950ae7dc2344e0e809fa58

PR #10381, opened 2025-12-19
interface: remove prefix from donation addresses

f321x:
Some servers seem to serve their donation address with `bitcoin:` prefix, preventing them from getting donations. This removes the prefix.

E.g.:
```
262.63 | I | interface.[2ex.digitaleveryware.com:50002] | invalid donation address from server: 'bitcoin:bc1q0q6lcajak7r8h8c8luzp0zj35v0usqd4pdzqjg'
```

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -1560,12 +1560,15 @@ class Interface(Logger):
         # check response
         if not res:  # ignore empty string
             return ''
-        if not bitcoin.is_address(res):
+        if not isinstance(res, str):
+            raise RequestCorrupted(f'{res!r} should be a str')
+        address = res.removeprefix('bitcoin:')
+        if not bitcoin.is_address(address):
             # note: do not hard-fail -- allow server to use future-type
             #       bitcoin address we do not recognize
             self.logger.info(f"invalid donation address from server: {repr(res)}")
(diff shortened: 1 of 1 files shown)


electrum/interface.py
```diff
@@ -1560,12 +1560,13 @@ async def get_donation_address(self) -> str:
         # check response
         if not res:  # ignore empty string
             return ''
-        if not bitcoin.is_address(res):
+        address = res.removeprefix('bitcoin:')
+        if not bitcoin.is_address(address):
```
SomberNight, line 1566: If `res` was not a str, previously `bitcoin.is_address()` handled that gracefully and returned False.
Now `.removeprefix()` would raise `AttributeError`.
Let's just check explicitly for type and raise the proper thing:
```suggestion
        if not isinstance(res, str):
            raise RequestCorrupted(f'{res!r} should be a str')
        address = res.removeprefix('bitcoin:')
        if not bitcoin.is_address(address):
```

f321x: ah right, added this check in https://github.com/spesmilo/electrum/pull/10381/commits/c56c27966a38260b42e2baaf11697a638c89f18e

PR #10384, opened 2025-12-22
pi: fix incorrectly parsing emaillike with 'ln' prefix as bech32 ln pi

f321x:
On master we incorrectly attempt to parse email like payment identifiers beginning with a 'ln' prefix as bech32 lightning payment identifiers. This obviously fails and the pi is reported to be invalid. E.g. `lntest@lnaddress.com` cannot be paid.

Also allows email like payment identifiers starting with `lightning:` uri prefix. I encountered this on a website and it makes sense to put a `lightning:` in front of a lightning address as the device somehow needs to differentiate between a email address and a lightning address, so we should allow this construction as well.

### DIFF
--- a/electrum/payment_identifier.py
+++ b/electrum/payment_identifier.py
@@ -293,9 +293,9 @@ class PaymentIdentifier(Logger):
                 self._type = PaymentIdentifierType.EMAILLIKE
                 self.emaillike = contact['address']
                 self.set_state(PaymentIdentifierState.NEED_RESOLVE)
-        elif re.match(RE_EMAIL, text):
+        elif re.match(RE_EMAIL, (maybe_emaillike := remove_uri_prefix(text, prefix=LIGHTNING_URI_SCHEME))):
             self._type = PaymentIdentifierType.EMAILLIKE
-            self.emaillike = text
+            self.emaillike = maybe_emaillike
             self.set_state(PaymentIdentifierState.NEED_RESOLVE)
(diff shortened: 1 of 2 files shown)


electrum/payment_identifier.py
```diff
@@ -21,30 +21,32 @@
...
-        return data
-    return None
+def maybe_extract_bech32_lightning_payment_identifier(data: str) -> Optional[str]:
+    data = remove_uri_prefix(data, prefix=LIGHTNING_URI_SCHEME)
+    if not data.startswith('ln'):
+        return None
+    try:
+        decoded_bech32 = bech32_decode(data, ignore_long_length=True)
+    except Exception:
+        return None
```
SomberNight: Can/should this ever raise? What are you trying to catch here?

SomberNight:
Besides above comment, eb45edb6631ec8fbc9fd188a3d5ef25762599457 looks good.

electrum/payment_identifier.py
f321x: I don't see a obvious reason how it could raise, it just felt safer to catch it as it might be fed weird stuff (user input). But thats probably overly cautious. Removed it in https://github.com/spesmilo/electrum/pull/10384/commits/2cd1de4f21cf807906004ba68d021a521032cb3c

SomberNight: I think if `bech32_decode` raises, it's a bug that needs to be fixed.
(assuming the args are of the correct type)

Issue #10383, opened 2025-12-20
Electrum mobile app does not focus camera in Waydroid


Talkless:
### Description

When using Electrum 4.6.2 on Waydroid LineageOS 20-20250809-VANILLA-waydroid_arm65, which in turn runs inside Ubuntu Touch 24.04 (24.04-1.0) 😂 on Volla Phone X (Gigaset GX290), I get this kind of image when scanning QR code:

<img width="332" height="720" alt="Image" src="https://github.com/user-attachments/assets/610de152-dbc3-45dc-9b52-6d4523797704" />

It's blurry.

In this case, QR Code is rather big, and bringing it closer makes it scan it successfully, but I did have experience where it just did not scan now matter what.

Touching screen does not engage autofocus.

Meanwhile Breez, Phoenix apps does focus camera on touch without issue *in the same system*.

PR #10148, opened 2025-08-19
android: 16kb alignment updates

accumulator:
p4a ref: electrum_20240930_android_16kb_page_alignment
Dockerfile: obtain 16kb aligned NDK r23 from google CI (dl-ndk-ci.sh)
barcode, zxing-cpp: add 16kb align patch
build_tools_util.sh: add function to apply a patch

verify produced shared objects using:
`for i in $(find . -name \*.so); do sh -c "echo $i;objdump -p $i|grep LOAD" ; done`
and check the `LOAD` entries show `align 2**14`

PR #10385, opened 2025-12-22
android: bump barcodescannerview, allow tap to focus, rm 16kb patch

f321x:
Bumps the BarcodeScannerView and zxing-cpp versions, this allows us to remove [the 16kb page alignment patch](https://github.com/spesmilo/electrum/commit/87d82f38c559ef27c7f749dac565a92d984148da) as this is [now done upstream as well](https://github.com/markusfisch/BarcodeScannerView/commit/76209b12b85804efb46ee0c5f25961805e14c01f).

Enables touch-to-focus in the Barcode Scanner activity, this allows the user to manually refocus by touching the screen. Might fixes https://github.com/spesmilo/electrum/issues/10383 and makes the scanner usable for users without autofocus. See https://github.com/markusfisch/CameraView#tap-to-focus

### DIFF
--- a/electrum/gui/qml/java_classes/org/electrum/qr/SimpleScannerActivity.java
+++ b/electrum/gui/qml/java_classes/org/electrum/qr/SimpleScannerActivity.java
@@ -95,7 +95,10 @@ public class SimpleScannerActivity extends Activity {
     private void startCamera() {
         if (mScannerView == null) {
             mScannerView = new BarcodeScannerView(this);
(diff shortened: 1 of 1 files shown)


SomberNight:
note: I have verified that the magic incantation from https://github.com/spesmilo/electrum/pull/10148#issue-3333967355 still works:

> verify produced shared objects using:
> `for i in $(find . -name \*.so); do sh -c "echo $i;objdump -p $i|grep LOAD" ; done`
> and check the `LOAD` entries show `align 2**14`

(for the .so inside `zxing-cpp-arm64-v8a-79f5adc6250e90de0bd635eb9181c5f8a18affda.aar`)

PR #10400, opened 2026-01-08
qml: add close no-op to QEQRScanner to fix type error

f321x:
Adds close() no-op method to `QEQRScanner` class to prevent this type error (and similar ones calling close on a QEQRScanner dialog):

```
01-02 17:28:09.645 10543 10565 I python  : 162.27 | W | gui.qml.qeapp | file:///data/data/org.electrum.electrum/files/app/electrum/gui/qml/components/SweepDialog.qml:123: TypeError: Property 'close' of object QEQRScanner(0xdd32f66fb600)
is not a function
```

The call to close() is only required for the desktop qml scanner dialog, on android it can be a no-op as the scanner is the java activity view.


SomberNight (approved):
Thanks!

PR #10401, opened 2026-01-08
add support for LevelDB

ecdsa:
This is still a draft, but it passes all the tests :-). See #4823 for context.

Note: I am NOT suggesting to use LevelDB for end-user wallets. The goal of this PR is to have scalable storage for lightning forwarding nodes. 

Also, this PR is not really about using LevelDB. It is about making our code able to use multiple types of storage (99% of the work here is about making our code database agnostic). Now that the logic is in place, we could use any other kind of database. I used levelDB because it is simple enough :-).

Main changes:
 - The `StoredDict` logic has been separated from `json_db` and moved to its own module. 
 - The wallet code interacts with a `StoredDict` object, that is database agnostic.

As a consequence:
- `WalletDB` no longer inherits from `JsonDB`
- `StoredDict` has a database field, that inherits from `BaseDB`
- `JsonDB` and `LevelDB` are child classes of `BaseDB`
- `WalletDB` upgrades are database agnostic.

Todo:
 - The levelDB storage needs a version number, in order to describes changes in encoding
 - We could use cbor for encoding, but that's another dependency (currently using json)
 - LevelDB should have unit tests
 - unit tests for stored_dict and json_db are currently mixed, should be separated.

Note that JsonDB requires to have the whole database in memory. In contrast, LevelDB does not, and uses iterators. That's really the main advantage of LevelDB. We could, in theory, rewrite `JsonDB` so that it does not load the whole database in memory either, with some sort of low level parsing (counting braces). Not sure if we could get acceptable performance from that approach.

PS: I hate sqlite. please do not suggest adding that.

### DIFF
--- a/.github/workflows/security-review.yml
+++ b/.github/workflows/security-review.yml
@@ -76,7 +76,7 @@ jobs:
       - name: Install Qt/QML runtime deps
         run: |
           sudo apt-get update
-          sudo apt-get -y install libgl1 libegl1 libxkbcommon0 libdbus-1-3
+          sudo apt-get -y install libgl1 libegl1 libxkbcommon0 libdbus-1-3 libleveldb-dev
 
       - name: Install Python dependencies
         run: |
--- a/electrum/wallet_db.py
+++ b/electrum/wallet_db.py
@@ -27,7 +27,7 @@ import json
 import copy
 from collections import defaultdict
 from typing import (Dict, Optional, List, Tuple, Set, Iterable, NamedTuple, Sequence, TYPE_CHECKING,
-                    Union, AbstractSet)
+                    Union, AbstractSet, Any)
 import time
 from functools import partial
 
@@ -35,21 +35,18 @@ import attr
 
 from . import bitcoin
 from . import constants
-from .util import profiler, WalletFileException, multisig_type, TxMinedInfo, MyEncoder
+from .util import with_lock as locked
+from .util import profiler, WalletFileException, multisig_type, TxMinedInfo
 from .keystore import bip44_derivation
 from .transaction import Transaction, TxOutpoint, tx_from_any, PartialTransaction, PartialTxOutput, BadHeaderMagic
 from .logging import Logger
 
 from .lnutil import HTLCOwner, ChannelType, RecvMPPResolution
-from .json_db import JsonDB, locked, modifier
-from . import stored_dict
-from .stored_dict import StoredObject, stored_at, register_key, register_name
+from .stored_dict import register_name, register_key
+from .stored_dict import StoredObject, StoredDict, StoredList, stored_at
 from .plugin import run_hook, plugin_loaders
 from .version import ELECTRUM_VERSION
 from .i18n import _
-
-if TYPE_CHECKING:
-    from .storage import WalletStorage
 
 
 class WalletRequiresUpgrade(WalletFileException):
@@ -685,6 +682,7 @@ class WalletDBUpgrader(Logger):
         invoices = self.data.get('invoices', {})
         for d in [invoices, requests]:
             for key, r in list(d.items()):
+                r = r.dump() # convert StoredDict/List to dict/list
                 _type = r.get('type', 0)
                 item = {
                     'type': _type,
@@ -928,6 +926,7 @@ class WalletDBUpgrader(Logger):
         for name in ['invoices', 'payment_requests']:
             invoices = self.data.get(name, {})
             for key, item in invoices.items():
+                item = item.dump() # convert StoredDict/List to dict/list
                 is_lightning = item['type'] == 2
                 lightning_invoice = item['invoice'] if is_lightning else None
                 outputs = item['outputs'] if not is_lightning else None
@@ -967,6 +966,7 @@ class WalletDBUpgrader(Logger):
             outputs_str = "\n".join(f"{txout.scriptpubkey.hex()}, {txout.value}" for txout in outputs)
             return sha256d(outputs_str + "%d" % timestamp).hex()[0:10]
         for key, item in list(invoices.items()):
+            item = item.dump() # convert StoredDict/List to dict/list
             is_lightning = item['lightning_invoice'] is not None
             if is_lightning:
                 continue
@@ -1532,7 +1532,7 @@ class WalletDBUpgrader(Logger):
         raise WalletFileException(msg)
 
 
-def upgrade_wallet_db(data: dict, do_upgrade: bool) -> Tuple[dict, bool]:
+def upgrade_wallet_db(data: dict, do_upgrade: bool=True) -> Tuple[dict, bool]:
     was_upgraded = False
 
     if len(data) == 0:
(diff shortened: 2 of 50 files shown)


brain-zhang:
I tested the new LevelDB backend support in Electrum on macOS 15.6.1, and it works very well. Thank you for this improvement 👍
I have two small suggestions that could further improve usability and completeness:

Dependency declaration
Currently, the plyvel dependency is only added to the testing requirements.
Since LevelDB is now a supported backend, it would be better to add plyvel to all relevant requirements files (e.g. default/runtime requirements), so users do not need to install it manually when using LevelDB in production.

GUI wallet creation option
In the wallet creation GUI, it would be helpful to add an option to choose the wallet storage backend (e.g. JSONDB vs LevelDB).
This could be implemented as a dropdown or selection box, with a sensible default value to preserve current behavior and avoid confusion for new users.

I believe these two small changes would make the LevelDB support more complete and user-friendly.

brain-zhang:
## Bug Report: LevelDB lock not released after [restore](cci:1:electrum/electrum/commands.py:329:4-359:21) or [create](cci:1:electrum/electrum/commands.py:299:4-327:21) command with `--level_db` flag

### Summary

When using the [restore](cci:1:electrum/electrum/commands.py:329:4-359:21) or [create](cci:1:electrum/electrum/commands.py:299:4-327:21) command with the `--level_db` flag to create a wallet, the LevelDB database lock is not released. This causes subsequent [load_wallet](cci:1:electrum/electrum/commands.py:283:4-292:26) calls to fail with an "already held by process" error until the electrum daemon is restarted.

### Steps to Reproduce

1. Start electrum in daemon mode:
   ```bash
   ./run_electrum daemon -d
   ```

2. Create/restore a wallet with LevelDB storage:
   ```bash
   ./run_electrum restore "your seed phrase here" --level_db
   ```

3. Try to load the wallet:
   ```bash
   ./run_electrum load_wallet
   ```

### Expected Behavior

The [load_wallet](cci:1:electrum/electrum/commands.py:283:4-292:26) command should successfully load the wallet.

### Actual Behavior

The command fails with the following error:

```
plyvel._plyvel.IOError: b'IO error: lock /path/to/wallet/LOCK: already held by process'
```

The wallet cannot be loaded until the daemon is stopped and restarted.

### Root Cause Analysis

In [electrum/commands.py](cci:7:electrum/electrum/commands.py:0:0-0:0), the [restore()](cci:1:electrum/electrum/commands.py:329:4-359:21) and [create()](cci:1:electrum/electrum/commands.py:299:4-327:21) commands call [restore_wallet_from_text()](cci:1:electrum/electrum/wallet.py:4320:0-4397:41) or `create_new_wallet()` respectively, which create a wallet with an open LevelDB connection. However, when these commands return, they only extract the path and message from the result dict, discarding the wallet object without closing its storage database.

Since LevelDB only allows a single process to hold the lock on a database at a time, the orphaned connection keeps the lock, preventing any subsequent attempt to open the same database.

**Problematic code in [commands.py](cci:7:electrum/electrum/commands.py:0:0-0:0):**

```python
@command('')
async def restore(self, text, passphrase=None, password=None, encrypt_file=True, wallet_path=None, use_levelDB=False):
    # ...
    d = restore_wallet_from_text(...)
    return {
        'path': d['wallet'].storage.path,  # wallet object is accessed but never closed
        'msg': d['msg'],
    }
```

### Proposed Solution

Close the storage database before returning from the [restore()](cci:1:electrum/electrum/commands.py:329:4-359:21) and [create()](cci:1:electrum/electrum/commands.py:299:4-327:21) commands:

```python
@command('')
async def restore(self, text, passphrase=None, password=None, encrypt_file=True, wallet_path=None, use_levelDB=False):
    # ...
    d = restore_wallet_from_text(...)
    wallet = d['wallet']
    result = {
        'path': wallet.storage.path,
        'msg': d['msg'],
    }
    # Close the storage DB to release the LevelDB lock (if any),
    # so that subsequent load_wallet calls can open the database.
    if wallet.storage and wallet.storage._db:
        wallet.storage._db.close()
    return result
```

The same fix should be applied to the [create()](cci:1:electrum/electrum/commands.py:299:4-327:21) command.

### Environment

- OS: macOS (MacOS15.6.2)
- Electrum version: 49848a2a7e766eca2a576923bf5bb828dc3f81ac
- Python version: [CPython 3.12.0]

ecdsa:
> Bug Report: LevelDB lock not released after restore or create command with `--level_db` flag

Thanks for the report. Your suggested fix makes sense, I will add it. 

Please note that if you find an issue in a pull request, you can add your report/suggestion as a comment to the code. This allows the comment to be later "resolved" once it is addressed, instead of being visible permanenty in this thread.

electrum/wallet_db.py
```diff
@@ -1551,67 +1552,95 @@ def upgrade_wallet_db(data: dict, do_upgrade: bool) -> Tuple[dict, bool]:
...
+
+
+class WalletDB(Logger):
+
+    def __init__(self, data: 'StoredDict', upgrade: bool = True):
+        Logger.__init__(self)
+        self.data = data
+        self.lock = self.data.lock
+        # we must perform db upgrades on the storeddict
+        was_upgraded = upgrade_wallet_db(self.data, upgrade)
```
SomberNight: Previously the (1) DB upgrades were run on the json data, before (2) converting json to python objects ([see here](https://github.com/spesmilo/electrum/blob/f60cdb0f405cb4ae8511d2bc9cfd4beab8a894bb/electrum/json_db.py#L294-L301)).

In this PR, the order is reversed: the WalletDB already gets a StoredDict, i.e. accessing its elements will already return python objects, and only after that we do the DB upgrades.

How would we do a DB upgrade that changes the structure of a StoredObject?
With the old flow, it was nothing special, but with the new flow it seems like a real headache.

e.g. consider we wanted to change `LocalConfig`: say we wanted to split `current_htlc_signatures` from a single long concatenated hex string to a list of shorter hex strings. How would we do that?
https://github.com/spesmilo/electrum/blob/f60cdb0f405cb4ae8511d2bc9cfd4beab8a894bb/electrum/lnutil.py#L99-L116
https://github.com/spesmilo/electrum/blob/f60cdb0f405cb4ae8511d2bc9cfd4beab8a894bb/electrum/lnutil.py#L218-L225

electrum/wallet_db.py
```diff
@@ -103,27 +100,24 @@ class WalletFileExceptionVersion51(WalletFileException): pass
...
 
 
 class WalletDBUpgrader(Logger):
-    def __init__(self, data: dict):
+    def __init__(self, data: StoredDict):
         Logger.__init__(self)
         self.data = data
-        # self.data must be in-memory dict (not a StoredDict or similar),
-        # so a failed, partial upgrade won't get commited to disk
-        assert type(self.data) == dict, type(self.data)
```
SomberNight, line 138: We should at least add a unit test that checks that a partial upgrade won't get committed to disk.
By partial upgrade I mean a DB upgrade that gets aborted in-between two well-defined numbered versions.

Say the user opens a wallet with DB version 30, and at the time the latest is 100. It is okay if we upgrade it to version 50 but then error out. And then on disk the DB is at version 50.

What would NOT be okay is if a single specific upgrade e.g. 55->56 fails mid-function, and what ends up on the disk is consistent with neither version 55 nor 56.

i.e. the `version += 1` upgrades should be atomic

electrum/wallet_db.py
ecdsa: No, the db upgrader still gets json-compatible objects. see `StoredDict.should_convert()`

electrum/wallet_db.py
ecdsa: this should be possible with levelDB batch writes

electrum/wallet_db.py
SomberNight: Ahh, hmm. Ok.

.github/workflows/security-review.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Install Qt/QML runtime deps
         run: |
           sudo apt-get update
-          sudo apt-get -y install libgl1 libegl1 libxkbcommon0 libdbus-1-3
+          sudo apt-get -y install libgl1 libegl1 libxkbcommon0 libdbus-1-3 libleveldb-dev
```
SomberNight, line 79: just a note re the CI failures:
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

> `pull_request_target`
> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the pull_request event does

i.e. it will not install `libleveldb-dev` as it uses the yml file from master, not from the PR, when running the CI

ecdsa: I see, thanks. how do you suggest to fix this?

SomberNight: One workaround that you are not going to like is that as the yml also has a `workflow_dispatch` trigger, you can run it manually using that.

On this page:
https://github.com/spesmilo/electrum/actions/workflows/security-review.yml
Select the `levelDB` branch (to specify that you want the .yml file from that branch), and then enter the PR number.

I started one run now: https://github.com/spesmilo/electrum/actions/runs/26505326899/job/78056093888


workflow_dispatch2

PR #10410, opened 2026-01-14
qml: fix invalid QEInvoiceParser state

### LINKED ISSUE #10406: qml: starting lnurl flow corrupts "Send" state
Steps: 
1. in qml, open a wallet, be on history screen
2. click Send    - scan QR code of an lnurl-pay ([see example](https://jochen-hoenicke.de/queue/#BTC,24h,weight) at the bottom)
3. "LNURL Payment request" popup appears (where user could e.g. specify exact amount or enter message)
    - cancel popup by clicking outside or using system Back button
4. now on history screen again, click Send
    - scan QR code of a bolt11 invoice
5. BUG: observe that the app enters the state of *final* confirmation for paying the bolt11 (e.g. PIN prompt)
    - instead we should just show the decoded bolt11 invoice
    - step 3 must be leaving some side-effect around

I noticed this when testing https://github.com/spesmilo/electrum/pull/10340 (with fingerprint enabled and config PIN set). In that case this is a critical bug: after scanning a bolt11 QR code, the user is prompted for the fingerprint and if they touch their finger, the payment is instantly sent -- without showing even the payment amount to the user.

f321x:
Fixes #10406  .
When scanning a lightning invoice we would pass it to `QEInvoiceParser.fromResolvedPaymentIdentifier()`, however `fromResolvedPaymentIdentifier()` doesn't reset the state of `QEInvoiceParser._lnurlData` which is used in QML to evaluate `payImmediately: invoiceParser.isLnurlPay` in the `onValidationSuccess` connection.

This change calls `clear()` in `fromResolvedPaymentIdentifier()` to ensure that `QEInvoiceParser` state gets reset when loading a new invoice. However when retrieving a bolt11 from a lnurl-pay callback we don't wan't to reset `QEInvoiceParser._lnurlData` so that `payImmediately` is true when confirming the lnurl pay dialog, for that I skip calling `fromResolvedPaymentIdentifier()` and instead call `validateRecipient()` directly so the `QEInvoiceParser` state doesn't get reset in this case.

### DIFF
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -451,7 +451,7 @@ class QEInvoiceParser(QEInvoice):
 
     @pyqtSlot(object)
     def fromResolvedPaymentIdentifier(self, resolved_pi: PaymentIdentifier) -> None:
-        self.canPay = False
+        self.clear()
         self.amountOverride = QEAmount()
         if resolved_pi:
             assert not resolved_pi.need_resolve()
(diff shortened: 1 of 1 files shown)


SomberNight (approved):
Looks good, tested. Thanks!

PR #10395, opened 2026-01-05
lnpeer: save received_mpp_htlcs after modifying ReceivedMPPStatue.htlcs

ecdsa:
ReceivedMPPStatus is a Namedtuple, which is immutable, but it contains a mutable field. Since ReceivedMPPStatus is not a StoredObject, no patch will be created when the htlcs list is modified, and we may end up not saving the change to disk if partial writes are enabled.

PR #10413, opened 2026-01-15
lnutil: change ReceivedMPPStatus.htlcs to frozenset, i.e. immutable

SomberNight:
As ThomasV says:

> ReceivedMPPStatus is a Namedtuple, which is immutable, but it contains
> a mutable field. Since ReceivedMPPStatus is not a StoredObject,
> no patch will be created when the htlcs list is modified, and we may
> end up not saving the change to disk if partial writes are enabled.

patch taken from https://github.com/spesmilo/electrum/pull/10395#pullrequestreview-3634244541
closes https://github.com/spesmilo/electrum/pull/10395

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2200,7 +2200,7 @@ class Peer(Logger, EventListener):
         # get payment hash of any htlc in the set (they are all the same)
         payment_hash = htlc_set.get_payment_hash()
         assert payment_hash is not None, htlc_set
-        assert payment_hash not in self.lnworker.dont_settle_htlcs
+        assert payment_hash.hex() not in self.lnworker.dont_settle_htlcs
         self.lnworker.dont_expire_htlcs.pop(payment_hash.hex(), None)  # htlcs wont get expired anymore
         for mpp_htlc in list(htlc_set.htlcs):
             htlc_id = mpp_htlc.htlc.htlc_id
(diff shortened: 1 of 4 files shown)


electrum/lnpeer.py
```diff
@@ -2214,10 +2214,12 @@ def _fulfill_htlc_set(self, payment_key: str, preimage: bytes):
...
                 # this check is intended to gracefully handle stale htlcs in the set, e.g. after a crash
                 self.logger.debug(f"{mpp_htlc=} was already settled before, dropping it.")
-                htlc_set.htlcs.remove(mpp_htlc)
+                htlc_set = htlc_set._replace(htlcs=htlc_set.htlcs - {mpp_htlc})
+                self.lnworker.received_mpp_htlcs[payment_key] = htlc_set
                 continue
             self._fulfill_htlc(chan, htlc_id, preimage)
-            htlc_set.htlcs.remove(mpp_htlc)
+            htlc_set = htlc_set._replace(htlcs=htlc_set.htlcs - {mpp_htlc})
+            self.lnworker.received_mpp_htlcs[payment_key] = htlc_set
```
ecdsa, line 2222: Note that saving to storage was performed outside the for loop in #10395.
This will now write to storage for every htlc in the set. 
I guess the goal is to make the code safer, in case an await is later introduced in the loop?

f321x: Yeah right, my intention was to keep it simple to reason about.
But yes this seems unnecessary, we already write to db when failing/fulfilling the htlc, and handle stale htlcs in the set (e.g. `chan.hm.was_htlc_preimage_released`), so this doesn't provide additional safety guarantees. I opened a follow up PR https://github.com/spesmilo/electrum/pull/10417 to move the saving outside of the loop.

PR #10418, opened 2026-01-16
wallet: update _password_in_memory on password change

### LINKED ISSUE #10415: qml: temporary corruption of QEDaemon.singlePassword
### Description

When running the qml gui with `config.WALLET_SHOULD_USE_SINGLE_PASSWORD is True` (i.e. on Android) and with single_password active, changing the wallet password won't change the in-memory `Daemon.singlePassword`.
This will leave the in-memory state inconsistent and weird things can happen.
Such as
- trying to display the seed results in displaying garbage
- a new wallet can be created, and then when trying to switch back to the older wallet, no password would be accepted (neither old pw, nor new pw)
- trying to sign transactions fails with invalid password

AFAICT, after restarting the program, the situation self-heals (?)

So there is [`QEDaemon.requestNewPassword`](https://github.com/spesmilo/electrum/blob/9857534e18eaf0af38cf51b91b921ff46b4447db/electrum/gui/qml/components/WalletDetails.qml#L470), and there is [`QEWallet.requestNewPassword`](https://github.com/spesmilo/electrum/blob/9857534e18eaf0af38cf51b91b921ff46b4447db/electrum/gui/qml/components/WalletDetails.qml#L523).
But only the latter considers changing `Daemon.singlePassword` -- and even that was only introduced in https://github.com/spesmilo/electrum/pull/10345.
However the bug affects 4.6.2 as well.

f321x:
Updates `Abstract_Wallet._password_in_memory` when the wallet password is being changed and the wallet was unlocked before instead of locking the wallet again. 
So that code that previously unlocked the wallet, and then changed the password will not have to re-unlock the wallet. 

Fixes #10415

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -382,3 +382,33 @@ class TestWalletPassword(WalletTestCase):
         with self.assertRaises(InvalidPassword):
             wallet.check_password("wrong password")
         wallet.check_password("1234")
+
+    async def test_update_password_unlocked_wallet(self):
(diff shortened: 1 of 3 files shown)


tests/test_wallet.py
```diff
@@ -382,3 +382,33 @@ async def test_update_password_with_app_restarts(self):
...
+        storage = WalletStorage(self.wallet_path)
+        db = WalletDB(wallet_str, storage=storage, upgrade=True)
+        wallet = Wallet(db, config=self.config)
+
+        wallet.check_password(None)
+        wallet.unlock(None)
+
+        # check that password_in_memory gets updated
+        wallet.update_password(None, "1234")
+        self.assertEqual(wallet.get_unlocked_password(), "1234")
```
SomberNight, line 397: Arghhh.. Thanks a lot for writing this test!!
It made me realise this is perhaps the wrong approach. :/

If there was no password set, and unlock was called, and then now the password is changed, should it really still be "unlocked"? It feels like a huge gotcha in the API (e.g. imagine a CLI/RPC user) that they need to know they might want to re-lock the wallet after changing the password. The wallet would remain "silently" unlocked - e.g. if scripts then call RPCs with the expectation that they need to provide the password, and they do, they might believe the wallet is locked as every call would just work.

In contrast, if wallet.update_password() re-locks the wallet, if a user expects it to keep the wallet unlocked, they will see on the next command that that's not the case and realise they need to unlock it again. Seems safer.

Hmhm...

SomberNight: Let me think about this a bit

f321x: Maybe we should only keep the wallet unlocked and update the password if it already had a password previously, otherwise lock it as before.
This would still fix the QML issue except for the case where someone has QML wallets without password and changes their password which seems niche but we still should handle this too somehow i guess.

f321x:
Closed in favor of https://github.com/spesmilo/electrum/pull/10419

PR #10421, opened 2026-01-18
Skip write on read-only config file

0x85C:
**Motive:** When you open electrum, it opens your last used wallet **immediately**, which may be undesirable if that happened to be a hidden / large balance wallet, from people who are near you.

**Current behaviour:** On application startup, a read-only config file results in a "Cannot load wallet" exception, as it tries to write the config key/value for "recently_open" and "current_wallet".

**Improvement:** Separate the PermissionError exception from the more generic OSError exception, and continue.

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -455,9 +455,12 @@ class SimpleConfig(Logger):
             with open(path, "w", encoding='utf-8') as f:
                 os_chmod(path, stat.S_IREAD | stat.S_IWRITE)  # set restrictive perms *before* we write data
                 f.write(s)
+        except PermissionError:
+            # Skip write if we don't have permission... e.g. read-only config file
+            return
         except OSError:
             # datadir probably deleted while running... e.g. portable exe running on ejected USB drive
-            # (in which case it is typically either FileNotFoundError or PermissionError,
+            # (in which case it is typically FileNotFoundError,
             #  but let's just catch the more generic OSError and test explicitly)
             if os.path.exists(self.path):  # or maybe not?
                 raise


SomberNight:
> Motive: When you open electrum, it opens your last used wallet immediately, which may be undesirable if that happened to be a hidden / large balance wallet, from people who are near you.

So your suggested workaround is to have the user change the file permissions of the config to be read-only, and then have Electrum tolerate the config being read-only.

Note that there is already a [`--forgetconfig`](https://github.com/spesmilo/electrum/pull/5898) option that you can set to instruct Electrum not to write anything to the config file.

> When you open electrum, it opens your last used wallet

In fact this was also the exact same motivation for adding the `--forgetconfig` option.

Alternatively, we could also add a config option, even exposed in the GUI, to disable storing the [last used](https://github.com/spesmilo/electrum/blob/99e173e50410f4d1bc29f11c0bdbd20550d1f83c/electrum/simple_config.py#L791) / [recently used](https://github.com/spesmilo/electrum/blob/99e173e50410f4d1bc29f11c0bdbd20550d1f83c/electrum/simple_config.py#L919) wallet files. (or perhaps some other config vars(?))
(EDIT: maybe we could even remove forgetconfig in favour of that approach!)

I am not sure silently ignoring failures updating the config file is desirable.

---

EDIT2: note that this PR, as-is, is completely redundant with `--forgetconfig`. This codepath does not get hit if `--forgetconfig` is set, as that results in the early-exit 10 lines up.

0x85C:
@SomberNight 

>In fact this was also the exact same motivation for adding the --forgetconfig option.

I saw that line above too and did the same early return for permission error, I didn't realise it was a command line option only and not a config option, I tried guessing if I could have "forget_config" in the config file, but this didn't work, perhaps it can also be added as an option in the config file so the application can be started without accidentally forgetting that command line option?

>Alternatively, we could also add a config option, even exposed in the GUI

If that is something likely to be accepted, then I could look into it.

SomberNight:
> I tried guessing if I could have "forget_config" in the config file, but this didn't work

Ok, that was unintentional. fixed in https://github.com/spesmilo/electrum/commit/21248817c9952965f8957743eeef45b07a53db91
You can add it manually to config file by editing the json. (adding it by calling `setconfig` would still not work)

> so the application can be started without accidentally forgetting that command line option?

You can set the CLI flag as part of the shortcut or bash alias, etc, that you use to start electrum.

Further, if you just consider this present diff re what you are trying to change and how --forgetconfig works, you must realise that you could set the config file read-only and start electrum with --forgetconfig. If you forget to set --forgetconfig, it will just crash as the config file is read-only. If you remember to set --forgetconfig, it will launch and run normally (but never change the config). Isn't that what you want? :)

0x85C:
> Ok, that was unintentional. fixed in [2124881](https://github.com/spesmilo/electrum/commit/21248817c9952965f8957743eeef45b07a53db91) You can add it manually to config file by editing the json. (adding it by calling `setconfig` would still not work)

Great, thanks for fixing.

Closing PR, no longer needed.

PR #10423, opened 2026-01-19
qml: fix: allow opening passwordless wallets

SomberNight:
This must be an old regression.
The GUI was not allowing to open a wallet that did not have a password set: it prompted for a password and did not accept any string (should at least accept empty "").

Without this, it was only possible to open a passwordless wallet if that was the first wallet the user opened (as otherwise we would overwrite the empty pw with the pw of the current wallet).


ecdsa:
good catch

PR #10424, opened 2026-01-19
util: use FileNotFoundError for dangling symlink

vishwas-droid:
Use FileNotFoundError instead of a generic Exception when a dangling
symlink is encountered.

This makes the error type more precise without changing behavior.

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1172,7 +1172,7 @@ def make_dir(path, *, allow_symlink=True):
     """
     if not os.path.exists(path):
         if not allow_symlink and os.path.islink(path):
-            raise Exception('Dangling link: ' + path)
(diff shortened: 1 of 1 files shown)


SomberNight:
Is there any specific codepath where this improves things, or is this purely theoretical?
Ultimately someone will have to manually fix or remove the dangling symlink anyway.
(is this LLM slop?)

vishwas-droid:
@SomberNight Fair question. This isn’t theoretical — it affects the filesystem path when a dangling symlink is encountered, and FileNotFoundError better matches Python’s semantics.
I agree the symlink still needs to be fixed manually; this just avoids misclassifying the error. If this looks fine, I’d be happy to have it merged.

vishwas-droid:
cc @SomberNight

PR #10425, opened 2026-01-20
qml: PasswordDialog: show error on invalid password

f321x:
Currently the PasswordDialog on QML would just close if the user enters an incorrect password. This is confusing as the user doesn't know why the dialog closed and if it initiated any action or not.

With the change the PasswordDialog will get the ability to show an error message and will show "Invalid Password" if an incorrect password is entered.
I also used it for the password unification warning ("Need to enter similar password ...") instead of showing a separate popup.

Screenshot_20260120_123858


ecdsa:
tested, looks good.

PR #10426, opened 2026-01-20
qml: restructure preferences, security improvements

f321x:
* Adds a "Security" section  to the Preferences that contains Biometric auth, Payment auth and screenshot protection
* Disables the screenshot protection toggle if the user is not on Android (it is misleading)
* Protects address private keys from screenshots
* Requires authentication for signing a message

Screenshot_20260121_103928

### DIFF
--- a/electrum/gui/qml/components/Preferences.qml
+++ b/electrum/gui/qml/components/Preferences.qml
@@ -197,6 +197,7 @@ Pane {
                         Layout.columnSpan: 2
                         Layout.fillWidth: true
                         spacing: 0
+                        enabled: AppController.isAndroid()
                         Switch {
                             id: setMaxBrightnessOnQrDisplay
                             onCheckedChanged: {
@@ -206,7 +207,7 @@ Pane {
                         }
                         Label {
                             Layout.fillWidth: true
-                            text: qsTr('Set display to max brightness when displaying QR codes')
+                            text: qsTr('Increase brightness when displaying QR codes')
                             wrapMode: Text.Wrap
                         }
                     }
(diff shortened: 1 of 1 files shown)


electrum/gui/qml/components/Preferences.qml
```diff
@@ -161,55 +161,62 @@ Pane {
...
-                                    }
-                                }
+                                if (activeFocus)
+                                    Config.setMaxBrightnessOnQrDisplay = checked
                             }
                         }
                         Label {
                             Layout.fillWidth: true
-                            text: qsTr('Biometric authentication')
+                            text: qsTr('Increase brightness when displaying QR codes')
```
SomberNight, line 210: Note: some lines are now not properly aligned when running on Android (but they are on desktop QML -- perhaps the difference is due to differing Qt versions).

I mean the rows that contain a switch and a label: looks like if the label is ~longish, it formats it into a two-line label but only actually uses the first line of that??

<details>
 <summary>pic</summary>

Screenshot_20260122-221059
</details>

f321x: I ran fa575ab1f (pre fingerprint commit) on Android to confirm that this is a preexisting issue. 
It seems to occur for longer strings that still fit on a single line. Seems like a bug in Qt. I tried running PyQt 6.4.2 on desktop but it segfaults immediately. Best approach is maybe to re-investigate this when we have a newer Qt version on Android?

signal-2026-01-23-105630

SomberNight: > preexisting issue

ah ok, nvm then

Issue #10427, opened 2026-01-20
util.CallbackManager memory leak of wallet objects


SomberNight:
In many subclasses of [`EventListener`](https://github.com/spesmilo/electrum/blob/ac99463e4f0868738f41c3f8529aed8625ad4a0d/electrum/util.py#L2004), such as `Abstract_Wallet`, `LNWatcher`,
`LNPeerManager`, we call `register_callbacks()` in `__init__`.
`unregister_callbacks()` is usually called in the `stop()` method.

Example - consider the wallet object:
- `Abstract_Wallet.__init__()` calls `register_callbacks()`
- there is a `start_network()` method
- there is a `stop()` method, which calls `unregister_callbacks()`
- typically the wallet API user only calls `stop()` if they also called
  `start_network()`.

This means the callbacks are often left registered, leading to the wallet
objects not getting GC-ed. The GC won't clean them up as
`util.callback_mgr.callbacks` stores strong refs to instance methods
of `Abstract_Wallet`, hence strong refs to the `Abstract_Wallet` objects.

An annoying example is `daemon.check_password_for_directory`, which
potentially creates wallet objects for all wallet files in the datadir.
It simply constructs the wallets, does not call `start_network()` and
neither does it call `stop()`.

PR #10428, opened 2026-01-20
fix util.CallbackManager memory leak

SomberNight:

fixes https://github.com/spesmilo/electrum/issues/10427

-----

note: I think we should split out the CallbackManager from util.py into its own module, but did not want to do that now.

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -1954,7 +1954,7 @@ class CallbackManager(Logger):
 
     def __init__(self):
         Logger.__init__(self)
-        self.callback_lock = threading.Lock()
+        self.callback_lock = threading.RLock()
         self._wcallbacks = defaultdict(set)  # type: Dict[str, Set[weakref.ref[Callable]]]  # note: needs self.callback_lock
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


tests/test_callbackmgr.py
SomberNight, line 1: Hmm. Seems some (random?) unit tests hang probabilistically. Not sure what's going on. :/

SomberNight: fixed in https://github.com/spesmilo/electrum/commit/cc4611f43d726ba61acbc9e8d83989405709b574

electrum/util.py
```diff
@@ -2005,11 +2023,28 @@ def on_done(fut_: concurrent.futures.Future):
 _event_listeners = defaultdict(set)  # type: Dict[str, Set[str]]
```
ecdsa, line 2007: could we set a boolean and raise an exception if it is called more than once?

SomberNight: which function do you mean?

ecdsa: oh sorry this comment was attached to the line "register_callbacks() should be called exactly once ", but apparently it has moved :-)

SomberNight: It is now harmless to call `register_callbacks()` multiple times, due to the change from `list` to `set` in CallbackManager. There is a unittest named `test_multiple_calls_to_register_callbacks` that tries to validate this.

but I think either approach is good, and a strict improvement over the status quo (where calling it multiple times would result in duplicate registrations).

PR #10429, opened 2026-01-21
plugin: make DeviceMgr.run non-blocking, fix lock

f321x:
If something in the `_hwd_comms_executor` thread is waiting for user input, e.g. when setting up a hww in the wizard the user needs to unlock the hww for `HardwareClientBase.get_xpub()` to return, the `_hwd_comms_executor` is blocked. If then `DeviceMgr.run()` gets called by the `Plugins` `DaemonThread` concurrently and tries to check the hww timeout on the `_hwd_comms_executor` as well, the `DaemonThread` is blocked too until the `_hwd_comms_executor` continues (and the `DaemonThread.job_lock` is taken.

Now if something tries to take the `DaemonThread.job_lock` it blocks as well, as it is taken while the jobs are being executed, so if a user e.g. tries to load a new plugin from the plugins dialog the whole gui thread will freeze until the hww gets unlocked.

To fix this I schedule the timeout check to run on the thread instead of awaiting its result, this way the `Plugins` `DaemonThread` doesn't get blocked.

There was a similar issue in https://github.com/Electron-Cash/Electron-Cash/pull/3173, but their approach is different. I think its better if `DeviceMgr.run()` doesn't block the `Plugins` `DaemonThread` as other jobs might also want to run.

### DIFF
--- a/electrum/plugin.py
+++ b/electrum/plugin.py
@@ -36,6 +36,7 @@ from urllib.parse import urlparse
 from typing import (NamedTuple, Any, Union, TYPE_CHECKING, Optional, Tuple,
                     Dict, Iterable, List, Sequence, Callable, TypeVar, Mapping)
 import concurrent
+from concurrent.futures import Future
 import zipimport
 from functools import wraps, partial
 from itertools import chain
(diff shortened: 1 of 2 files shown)


electrum/plugin.py
```diff
@@ -1053,10 +1055,16 @@ def run(self):
...
-        for client in clients:
-            client.timeout(cutoff)
+        for client, client_id in clients:
+            if fut := self._ongoing_timeout_checks.get(client_id):
+                if not fut.done():
+                    continue
+            # scheduling the timeout check prevents blocking the Plugins DaemonThread if the
+            # _hwd_comms_executor Thread is blocked (e.g. due to it awaiting user input).
+            fut = _hwd_comms_executor.submit(client.timeout, cutoff)
+            self._ongoing_timeout_checks[client_id] = fut
```
SomberNight, line 1067: Maybe there's too many threads and locks here. :/
Besides the GUI thread and the asyncio thread, the threads in question are for the Plugins and for HWD comms. Wonder if we could simplify. Well, later.. :)

PR #10433, opened 2026-01-22
qt: add Changelog to Help menu in toolbar

f321x:
Adds a `Changelog` item to the Help menu of the toolbar. Clicking on the item opens the changelog in the browser (github text-only form). 

Screenshot_20260122_155817

### DIFF
--- a/electrum/constants.py
+++ b/electrum/constants.py
@@ -57,6 +57,7 @@ def create_fallback_node_list(fallback_nodes_dict: dict[str, dict]) -> List[LNPe
 
 GIT_REPO_URL = "https://github.com/spesmilo/electrum"
 GIT_REPO_ISSUES_URL = "https://github.com/spesmilo/electrum/issues"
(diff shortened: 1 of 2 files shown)


electrum/constants.py
```diff
@@ -57,6 +57,7 @@ def create_fallback_node_list(fallback_nodes_dict: dict[str, dict]) -> List[LNPe
 
 GIT_REPO_URL = "https://github.com/spesmilo/electrum"
 GIT_REPO_ISSUES_URL = "https://github.com/spesmilo/electrum/issues"
+RELEASE_NOTES_URL = "https://raw.githubusercontent.com/spesmilo/electrum/refs/heads/master/RELEASE-NOTES"
```
SomberNight, line 60: note: one consequence of pointing to the master branch is that we (kind of) will never be able to cleanly rename this file in the git repo

but ok, I think it's fine.

PR #10318, opened 2025-11-20
lnsweep: lnwatcher needs to keep_waiting for pending hold_invoice

SomberNight:
If RHASH is in lnworker.dont_settle_htlcs, we should not reveal
the preimage. But also, we should not disregard the htlc either.
lnwatcher needs to keep waiting until either the user cancels/settles
the hold invoice, or until the hold_invoice's CLTV expires.

TODO: I need to write a test for this!

PR #10442, opened 2026-01-26
lnsweep: simplify maybe_reveal_preimage_for_htlc

SomberNight:
"When should we reveal preimages onchain?"
This commit tries to simplify the thinking to:
- we can reveal preimages (actually in any context) if they are already public
- a preimage is public if any other lightning node knows it besides us
  - if we learn the preimage from another LN node, it is public
  - if we send update_fulfill_htlc, it becomes public
  - if we see a preimage onchain, it is public

- we can't restrict revealing preimages to only already public ones both onchain and offchain: there needs to be a state transition out of non-public for our own invoices
  - we allow the state transition when we send update_fulfill_htlc (offchain)
    - this logic is unchanged
  - ~then onchain we can simplify: enforce preimage already being public~
    - EDIT: several of the regtest tests expect the code to make the other choice, as that's what we've been doing so far. It is simpler for now not to change this logic.

- in lnsweep._maybe_reveal_preimage_for_htlc:
  - partial mpp check is not relevant if preimage is already public
  - let's just always do KeepWatchingTXO, for sanity/safety

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2705,17 +2705,14 @@ class LNWallet(Logger):
         preimage: bytes,
         *,
         write_to_disk: bool = True,
-        mark_as_public: Optional[bool] = None,  # see is_preimage_public
+        mark_as_public: bool = False,  # see is_preimage_public
     ):
         assert isinstance(payment_hash, bytes), f"expected bytes, but got {type(payment_hash)}"
         assert isinstance(preimage, bytes), f"expected bytes, but got {type(preimage)}"
         if sha256(preimage) != payment_hash:
             raise Exception("tried to save incorrect preimage for payment_hash")
-        old_tuple = _, old_is_public = self._preimages.get(payment_hash.hex(), (None, None))
-        if mark_as_public is None:  # if unset, keep current DB value
-            mark_as_public = old_is_public or False
-        if old_is_public and not mark_as_public:
-            raise Exception("preimage mark_as_public: True->False transition is forbidden")
+        old_tuple = _, old_is_public = self._preimages.get(payment_hash.hex(), (None, False))
+        mark_as_public |= old_is_public  # disallow True->False transition
         # sanity checks and conversions done.
         new_tuple = preimage.hex(), mark_as_public
         if old_tuple == new_tuple:  # no change


electrum/lnsweep.py
```diff
@@ -481,12 +481,19 @@ def _maybe_reveal_preimage_for_htlc(
...
         #   for this MPP set. So the MPP set might still transition to complete!
         #   The MPP_TIMEOUT is only around 2 minutes, so this window is short.
         #   The default keep_watching logic in lnwatcher is sufficient to call us again.
+        # - if this is an incoming payment, where
+        #   - all the HTLCs have already been received (covering invoice amt) and are pending currently, and
+        #   - no one knows the preimage besides us, and
+        #   - this channel was just force-closed by either party
+        #   -> then we can freely decide whether to claim the HTLCs or let them time out.
+        #      Both choices are valid and safe. *Here* it is simpler to implement
+        #      letting the HTLC time out, so that's what we do.
```
SomberNight: note: several of the regtest tests expect the code to make the other choice, as that's what we've been doing so far

SomberNight: if we are forwarding, and the outgoing HTLC has already been irrevocably committed, we need to *keep watching* this incoming HTLC until cltv_abs and be prepared to still fulfill it if we learn the preimage from downstream

f321x: Besides the simplicity aspect you mention i can also think of these tradeoffs:
Claiming may consume a large amount of the htlcs value, esp during high mining fees, so it might be preferable to retry another lightning payment to pay less fees and let the sender pay the mining fees to time out the htlcs.
OTOH not claiming them would cause the payment to be stuck until the htlcs timed out, which is also quite bad experience for both sides of the payment.
Though in practice i guess it should not happen often, so simplicity might outweigh the cost of any potential UX improvement that gets triggered only rarely.

ecdsa:
the concept looks good to me, I would prefer a db upgrade.
also, `lnworker.save_preimage` could take an additional parameter for this boolean.

electrum/lnchannel.py
```diff
@@ -1466,6 +1466,7 @@ def extract_preimage_from_htlc_txin(self, txin: TxInput, *, is_deeply_mined: boo
         if preimage:
             assert ripemd(sha256(preimage)) == ripemd_payment_hash
             payment_hash = sha256(preimage)
+            self.lnworker.mark_preimage_as_public(payment_hash)
```
ecdsa: This call seems unnecessary; if we have already extracted the `preimage` in line 1484, then it should already be marked as public.

Even if adding this line may feel 'safer', I would prefer if reviewers did not have to spend brain cycles to figure this out. In the long run, making the code easy to review makes it safer.

SomberNight: Note  there is no `mark_preimage_as_public` to correspond to the codepath where we first reveal the preimage *onchain*. This call here is the only path that might handle that scenario and atm I am not fully certain it always gets to run.
That is, for example, if we are the original creator of a preimage, and we already revealed it through e.g. lnsweep+txbatcher, this line here is the only one that might mark it as public. but yes, maybe there should be a better place to add the `mark_preimage_as_public` call.

-----

EDIT: 
> we first reveal the preimage onchain

oh hmm, well, if we [decide never to first reveal preimages onchain](https://github.com/spesmilo/electrum/pull/10442#discussion_r2728804911), then this should never happen.  but for encapsulation purposes not sure that decision should leak here

ecdsa: I do not understand. This `extract_preimage` method is mostly for the case where the remote reveals the preimage of an htlc we proposed. In the case where we reveal the preimage of an htlc we received, it feels extremely convoluted to mark it as public once we have extracted it from the witness of a transaction we received from the Electrum server. We should mark it as public as soon as we broadcast our htlc sweep transaction.

SomberNight: > We should mark it as public as soon as we broadcast our htlc sweep transaction.

Yes, of course you are correct that it should work like that. However in the current code, where we broadcast is far away from where we make the decision to maybe reveal the preimage (perhaps some data could be saved in `SweepInfo`?).

Anyway, I started to try to write unit tests for lnsweep, and am bogged down in that.

electrum/lnsweep.py
SomberNight: note: this is currently not handled and we might lose money when forwarding.
https://github.com/spesmilo/electrum/blob/4f7b6e897710338e8a4cfb8fa97c305218bdff88/electrum/lnwatcher.py#L163-L168
If `chan.get_ctx_sweep_info` does not include an item for the incoming HTLC, e.g. because we don't YET know the preimage for it, lnwatcher might stop watching the channel and never redeem the incoming HTLC onchain, even after the outgoing HTLC gets fulfilled.

ecdsa: This looks like another use case for `KeepWatchingTXO`, 7e1fd008f032ca56a3727e1a225e6f2a2f95ad82

electrum/lnchannel.py
ecdsa: > perhaps some data could be saved in SweepInfo ?

how about marking the preimage as public at the end of `_maybe_reveal_preimage_for_htlc` ?

SomberNight: Well, yes, I think maybe we could do that.
Note that this might result in us marking the preimage public way too early: after `_maybe_reveal_preimage_for_htlc` runs, we might still run into issues/events that lead us not to broadcast the HTLC-claim tx. However that seems okay - it would be the other direction that is dangerous: if we marked it too late.

electrum/wallet_db.py
```diff
@@ -1345,6 +1346,16 @@ def _convert_version_67(self):
         self.data['channels'] = channels
         self.data['seed_version'] = 67
 
+    def _convert_version_68(self):
+        if not self._is_upgrade_method_needed(67, 67):
+            return
+        old_preimages = self.data.get('lightning_preimages', {})
+        new_preimages = {}
+        for _hash, preimage in old_preimages.items():
+            new_preimages[_hash] = (preimage, False)
```
SomberNight, line 1364: Are we okay with the weird edge cases that happen around upgrade time?

For example, consider Alice received an HTLC, they revealed the preimage offchain, but the HTLC is not yet irrevocably removed. -- Bob did not revack and he disappeared.
Now Alice upgrades Electrum, db upgrade happens, preimage gets marked as public=False, and now either Alice or Bob force-closes, with the HTLC still being there. Alice then should but will not reveal the preimage onchain.

Note that blanket setting public=True during the upgrade would introduce its own issues: partial MPP, etc.

Anyway this seems hard to fix and unlikely to happen, so perhaps we can say we consider it out of scope.

electrum/lnsweep.py
```diff
@@ -503,6 +507,8 @@ def _maybe_reveal_preimage_for_htlc(
         )
         return None, keep_watching_txo
     preimage = chan.lnworker.get_preimage(htlc.payment_hash)
+    # this preimage will be revealed
+    chan.lnworker.mark_preimage_as_public(htlc.payment_hash)
     return preimage, None
```
SomberNight, line 516: can preimage be None here? if so, is it okay to mark it as public?

SomberNight: I think the safe thing might be to always do "KeepWatchingTXO" until cltv_abs, if we decide not to reveal the preimage for any reason. E.g. if we return None because we don't have the preimage, we should keep watching...

---

E.g. what if the HTLC is a hold invoice?

Hold-invoices and JIT channels were the two original usecases for adding the KeepWatchingTXO logic in https://github.com/spesmilo/electrum/pull/10318, however `lnworker.dont_settle_htlcs` changed in the meantime and it is now only used for JIT channels (not hold invoices).

Though maybe `lnworker.is_complete_mpp` is False then and so execution would take the first branch?

electrum/wallet_db.py
ecdsa: that's fine with me.

electrum/lnsweep.py
SomberNight: > I think the safe thing might be to always do "KeepWatchingTXO" until cltv_abs, if we decide not to reveal the preimage for any reason

done in https://github.com/spesmilo/electrum/commit/0a081695b2c33799e28b3863b3b8dcd8efe76d64

SomberNight:
I squashed and force-pushed. There is a copy of the [old branch here](https://github.com/SomberNight/electrum/tree/202601_lnworker_is_preimage_public_bak20260224b) ([commit](https://github.com/SomberNight/electrum/commit/d0eb87de0cc475d5d0cc3eb611e864a3221f45c2)).

electrum/lnsweep.py
SomberNight: note: decided not to change this behaviour for now

electrum/lnworker.py
```diff
@@ -2699,19 +2699,35 @@ def delete_payment_bundle(
...
+        if mark_as_public is None:  # if unset, keep current DB value
+            mark_as_public = old_is_public or False
+        if old_is_public and not mark_as_public:
+            raise Exception("preimage mark_as_public: True->False transition is forbidden")
+        # sanity checks and conversions done.
+        new_tuple = preimage.hex(), mark_as_public
+        if old_tuple == new_tuple:  # no change
+            return
+        self.logger.debug(f"saving preimage for {payment_hash.hex()} (public={mark_as_public})")
+        self._preimages[payment_hash.hex()] = new_tuple
```
SomberNight, line 2721: >  also, lnworker.save_preimage could take an additional parameter for this boolean.

I changed to this, as the previous API was too error-prone - there was already a bug in this branch due to that. (ordering of calls to `lnworker.save_preimage` vs `lnworker.mark_preimage_as_public` mattered)

Unfortunately, to keep the API (for callers) simple, I had to put quite a bit of complexity inside the implementation of `save_preimage`. Still, I think it's ~okay.

electrum/lnsweep.py
```diff
@@ -480,14 +480,22 @@ def _maybe_reveal_preimage_for_htlc(
     htlc: 'UpdateAddHtlc',
     sweep_info_name: str,
 ) -> Tuple[Optional[bytes], Optional[KeepWatchingTXO]]:
-    """Given a Remote-added-HTLC, return the preimage if it's okay to reveal it on-chain."""
-    if not chan.lnworker.is_complete_mpp(htlc.payment_hash):
+    """Given a Remote-added-HTLC, return the preimage if it's okay to reveal it on-chain.
```
SomberNight, line 483: lnsweep desperately needs unit tests. I have a WIP branch to build the architecture for it but we can merge this before I get to finish that.

ecdsa: I have tested the current branch using regtest (using the 'timebomb' approach).
It behaves well.

electrum/lnworker.py
ecdsa: regarding complexity, why does `save_preimage` require the payment_hash as parameter?
I guess we could kill that parameter, although the caller would probably want to check the hash in some cases.

SomberNight: No substantial reason. Indeed it is a final sanity check that tests the preimage is correct.
Anyway, I would keep it as-is for now.

Issue #10445, opened 2026-01-27
Transaction export - fee in satoshi


OEsmir:
### Description

The fee in the exported csv file is now (or maybe for a while) reported in satoshis, even though the value of the transaction is reported in BTC. I have tried to search for any configuration inside the electrum client to change this but nothing seems to be available. 

As someone who uses Electrum for (multiple) organizations, I rely on the export to maintain the bookkeeping of the wallet, and the discrepancy between the units is making it a bit difficult. Can we have an option developed to allow selecting the exported units, or at least keep the units consistent between the different columns. 

Thanks

pranitaurlam:
Hi  @OEsmir 
I’m interested in working on this issue and would love to take it up if it’s available.
I’m happy to investigate and propose a fix.

If that sounds okay, could you please assign this issue to me?

Thanks!

f321x:
Hey, thanks for reporting the issue you encountered. I agree it is inconsistent to show the fees in sats while the other values are shown in bitcoin. There is now an open pull request https://github.com/spesmilo/electrum/pull/10451 that changes the fees to be exported in the bitcoin unit as well.

OEsmir:
Thank you for solving this. Do you know when we can expect an update for the client?

f321x:
@OEsmir there is no fixed release schedule, but a release (4.7.1) in the next 1-2 months seems likely.

SomberNight:
note: we are in the process of releasing 4.7.1 right now, but decided not to include this change in the 4.7.x branch. It will only be in 4.8.
This is because it is a breaking a change in the export format and some people might have already adapted their scripts/workflow to the current format (which was last changed in 4.6.0, in https://github.com/spesmilo/electrum/commit/86183797dbcc1eaf1c2dba4783f7808c202ad5d9). We try not to include breaking changes in bugfix versions.

PR #10446, opened 2026-01-27
Fix consistent fee units in history CSV export (#10445)

pranitaurlam:
This PR ensures that transaction fees in the exported CSV file use the same units (BTC) as the transaction amounts, addressing issue #10445.

### Changes:
- **Consistent Units**: Wrapped `fees_sat` in the `Satoshis` class in `electrum/gui/qt/history_list.py`. This ensures that fees are formatted as BTC strings (e.g., `0.00005000`) instead of raw satoshi integers (e.g., `5000`), matching the formatting of the amount columns.
- **Header Update**: Renamed the CSV column header from `network_fee_satoshi` to `network_fee` to reflect the consistent unit usage.

### DIFF
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -886,7 +886,7 @@ class HistoryList(MyTreeView, AcceptFileDragDrop):
                     item['bc_value'],
                     item['ln_value'],
                     item.get('fiat_value', ''),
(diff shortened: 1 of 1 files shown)


pranitaurlam:
Hi @SomberNight ,
Could you please review this when you have time?
Thanks!

Issue #10420, opened 2026-01-17
Signet Lightning Error


vittoridavide:
### Description

Hello, I opened a channel on the Signet version of the app and the channel shows as status "Funded" but I'm unable to make any payments. When I click on send it tells me "You cannot pay that invoice using Lightning.
Your channels can send 0. mBTC." 

here below is the information of the channel I opened

<img width="803" height="739" alt="Image" src="https://github.com/user-attachments/assets/cb988d71-c4c5-411f-9ec5-390262a3d074" />

SomberNight:
> channel shows as status "Funded"

You can only use a channel to make payments when it is in the ["OPEN" state](https://github.com/spesmilo/electrum/blob/99e173e50410f4d1bc29f11c0bdbd20550d1f83c/electrum/lnchannel.py#L76-L95).

Is the channel somehow stuck in FUNDED?
Is the remote peer online?
What if you restart the program?
Do you have logs?

pranitaurlam:
Hi  @VittoriDavide 
I’m interested in working on this issue and would love to take it up if it’s available.
I’m happy to investigate and propose a fix.

If that sounds okay, could you please assign this issue to me?

Thanks!

PR #10447, opened 2026-01-27
Fix Signet Lightning usability: remove restrictive sync and trampolin…

pranitaurlam:
checks (#10420)
This PR addresses several usability issues with Lightning channels, particularly on Signet, where channels could get stuck in the "Funded" state or report 0 mBTC capacity even when active.

### Changes:
- **Improved State Transitions**: In `electrum/lnpeer.py`, I removed the `is_up_to_date()` check in `reestablish_channel`. This ensures that channels can complete the Lightning handshake as soon as their funding transaction is confirmed, without waiting for the entire wallet history to sync (which is problematic on slow Signet nodes).
- **Inclusive Capacity Reporting**: In `electrum/lnworker.py`, I removed a restrictive filter in `get_channels_for_sending` that excluded active channels if they weren't to recognized trampoline nodes or if the local gossip database was empty. This fixes the confusing "Your channels can send 0. mBTC" error for users with private or direct peer channels.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -1597,7 +1597,6 @@ class Peer(Logger, EventListener):
             self._send_channel_reestablish(chan)
             return
         if self.network.blockchain().is_tip_stale() \
-                or not self.lnworker.wallet.is_up_to_date() \
                 or self.lnworker.current_target_feerate_per_kw(has_anchors=chan.has_anchors()) \
             is None:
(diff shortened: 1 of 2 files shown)


pranitaurlam:
Hi @SomberNight,
I’ve looked into this issue and proposed a change. I’d really appreciate your feedback and review when possible.
Thanks!

SomberNight:
LLM slop

PR #10451, opened 2026-01-28
history export: make fees bitcoin, add unittest, rm unconf tx

### LINKED ISSUE #10445: Transaction export - fee in satoshi

f321x:
* Makes fees in history export bitcoin again (instead of sat). This is more consistent with the other values which are also exported as bitcoin. Fixes #10445
* Moves the history export function from Qt gui to electrum.util. This allows it to get unittested and as a side effect makes it more available (e.g. for the CLI or QML).
* Fixes inconsistency in export where we would show "No Data" for the fiat value, but 0 for the fiat fee.
* Removes unconfirmed/local tx from the history export, they are unordered and make the output non-deterministic, also they don't seem helpful for accounting purposes.
* Adds a new unittest to test_util for the history export, comparing the output against reference files.


tests/test_util.py
```diff
@@ -517,3 +524,53 @@ async def foo():
         # refs should be cleaned up by now:
         self.assertEqual(0, len(util._running_asyncio_tasks))
 
+
+class TestUtilTestnet(ElectrumTestCase):
+    TESTNET = True
```
SomberNight: this is fine but just wanted to note that we also have an `@as_testnet` function decorator:
https://github.com/spesmilo/electrum/blob/9aa2898c1510217eca58fe70228b0a5015273ac7/tests/__init__.py#L114-L115

electrum/util.py
```diff
@@ -2462,6 +2464,76 @@ def to_dict(self):
         }
 
 
+def create_wallet_history_export(*, wallet: 'Abstract_Wallet', fx: Optional['FxThread'], file_path: str, is_csv: bool):
```
SomberNight: wouldn't this make more sense in wallet.py?
like an `export_history_to_file` or similar method for `Abstract_Wallet`?
well, not sure.

tests/test_util.py
```diff
@@ -517,3 +524,53 @@ async def foo():
...
+        # prepare wallet with realistic history
+        c = SimpleConfig({'electrum_path': self.electrum_path})
+        c.NETWORK_OFFLINE = True
+        c.FX_EXCHANGE, c.FX_CURRENCY, c.FX_USE_EXCHANGE_RATE, c.FX_HISTORY_RATES = "BitFinex", "EUR", True, True
+        shutil.copytree(Path(__file__).parent / "fiat_fx_data", Path(self.electrum_path) / "cache")
+        daemon = Daemon(config=c, listen_jsonrpc=False)
+        test_wallets = (
+            "client_3_3_8_xpub_with_realistic_history",
+            "client_4_5_2_9dk_with_ln",  # has labels, local tx, ln tx
+        )
```
SomberNight: it is worth it to test both?

8 files seems a bit overkill:
{wallets} x {fx, no-fx} x {csv, json}
we could reduce it to 4 files if we only keep e.g. `client_4_5_2_9dk_with_ln`

f321x: I thought the overhead is negligible (like 80ms runtime on my machine), but yeah might be overkill to add that many files for a single test. I removed 4 and kept the client_4_5_2_9dk_with_ln export references.

electrum/util.py
f321x: I considered this but was hesitant at first because it seemed somewhat unrelated to the other wallet logic like transaction creation etc. But yeah its also a functionality of the wallet, so i added it to `Abstract_Wallet` in https://github.com/spesmilo/electrum/pull/10451/commits/e85b0e29c0eb8ecc766316f0f15430420e7ae1e0.

tests/test_util.py
f321x: Oh this is much nicer, but now that i moved the test to test_wallet.py it became irrelevant. I added a comment to `ElectrumTestCase`  in https://github.com/spesmilo/electrum/pull/10451/commits/31a2f515eb067db26b131c262f90056302169b9c that this exists so it is more obvious the next time.

electrum/wallet.py
```diff
@@ -3614,6 +3614,80 @@ def add_future_tx(self, sweep_info: 'SweepInfo', wanted_height: int):
...
+                    else (child.get('fee_msat', 0) or 0) // 1000
+                if fees_fiat is not None and (child_fiat_fee := child.get('fiat_fee')):
+                    fees_fiat += child_fiat_fee
+
+            fees_sat += h_item['fee_sat'] or 0 if 'fee_sat' in h_item \
+                else (h_item.get('fee_msat', 0) or 0) // 1000
+            if fees_fiat is not None and (h_item_fiat_fee := h_item.get('fiat_fee')):
+                fees_fiat += h_item_fiat_fee
+
+            if fees_fiat is not None and str(h_item.get('fiat_value')) == 'No Data':
```
SomberNight: This looks fragile. Isn't this string translated?
https://github.com/spesmilo/electrum/blob/dcfef1111fc861ebbaaa6e1313dd7d88db61003d/electrum/util.py#L289-L293

electrum/wallet.py
```diff
@@ -3614,6 +3614,80 @@ def add_future_tx(self, sweep_info: 'SweepInfo', wanted_height: int):
...
+                assert str(fees_fiat) == 'No Data' if str(item.get('fiat_value')) == 'No Data' else True
+                line = [
+                    item.get('txid', ''),
+                    item.get('payment_hash', ''),
+                    item.get('label', ''),
+                    item.get('confirmations', ''),
+                    item['bc_value'],
+                    item['ln_value'],
+                    item.get('fiat_value', ''),
+                    util.format_satoshis(fees_sat),
```
SomberNight, line 3668: note: this loses msat precision.
ah but I guess it was already lost by the `// 1000` stuff above.
If it's really for "accounting", it might matter.

Well but anyway, it is a pre-existing issue.

SomberNight: maybe just add a short comment on this line, as it is not obvious.
```suggestion
                    util.format_satoshis(fees_sat),  # note: truncates msat prec but already did "// 1000" stuff above
```

SomberNight:
(self-note: reviewed 31a2f515eb067db26b131c262f90056302169b9c)

electrum/wallet.py
f321x: Good catch, indeed, found a more robust solution in https://github.com/spesmilo/electrum/compare/31a2f515eb067db26b131c262f90056302169b9c..a6a57f0fec5cb4b2f3b238130de68b189730bc45

electrum/wallet.py
f321x: I added fixme comments in https://github.com/spesmilo/electrum/compare/31a2f515eb067db26b131c262f90056302169b9c..a6a57f0fec5cb4b2f3b238130de68b189730bc45

electrum/wallet.py
SomberNight: you have the same test in other places:
https://github.com/f321x/electrum-fork/blob/a6a57f0fec5cb4b2f3b238130de68b189730bc45/electrum/wallet.py#L3659

f321x: In that place I kept it intentionally as its independent of the translation. 
It will only raise if one string is 'No Data' while the other isn't, which I think in this context makes sense, the assert will just become useless if translation is enabled. 
But i can also do a value comparison like above, that's just more verbose.

SomberNight:
Thanks!

PR #10452, opened 2026-01-29
build: android: upgrade docker base img to debian 13

SomberNight:
This should make the F-Droid build simpler.

ref https://gitlab.com/fdroid/fdroiddata/-/merge_requests/32358#note_3038803195

----

- upgrades the android builder docker image base debian 12 -> 13 (to current stable)
- issues (for which I have shitty workarounds)
    - debian 12 ships python 3.11, debian 13 ships python 3.13
        - we can't use python 3.13 as cython < 3.0 cannot be built with it, and p4a requires cython < 3.0
        - so now we manually build python 3.12
    - debian 12 ships Java 17, debian 13 only ships Java 21 and 25
        - we can't upgrade to Java 21, as then we would also have to upgrade Gradle
        - the QR scanner lib we use ([markusfisch/zxing-cpp](https://github.com/markusfisch/zxing-cpp) and friends) pins an old version of Gradle
            - [seemingly](https://github.com/markusfisch/zxing-cpp/commit/d98ed5d0be513c4077b4c13d1f7873f141949839) because they want to support ancient versions of Android

### DIFF
--- a/contrib/android/Dockerfile
+++ b/contrib/android/Dockerfile
@@ -91,11 +91,42 @@ RUN mkdir --parents "${ANDROID_SDK_HOME}/.android/" \
     && echo '### User Sources for Android SDK Manager' \
         > "${ANDROID_SDK_HOME}/.android/repositories.cfg"
 
-# accept Android licenses (JDK necessary!)
+# download Java-17 (debian 13 only packages Java-21 and Java-25)
+# - we download the amd64 binaries from debian 12 repos
+# - we should try to upgrade to Java-21...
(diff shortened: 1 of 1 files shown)


SomberNight:
Ok, so I built apks from 4da7b7f55d85f11d84cfb5995bd6ad1d9d9cb311, and it works at runtime.
Felix also built one and we confirmed it is still reproducible.

However, it is extremely annoying how we seem to have to amass more and more tech debt. :/
Deploying a python project on android keeps being an epic challenge still.

PR #10453, opened 2026-01-29
wallet: estimate base tx feerate based on original base tx size

f321x:
I noticed `TxBatcher` tried to create a new batch transaction that got rejected because the fee rate of the replacement transaction was lower than the fee rate of the transaction it was supposed to replace.

In `wallet.py` we first stripped the base transaction and then estimated its weight for the calculation of the fee rate. Because `Transaction.estimated_size()` potentially slightly overestimates the size, we slightly underestimated the fee rate of the base transaction. 
As we have access to the original base tx i suggest we use this information and estimate the size of the transaction before stripping it to get a more accurate estimate. This seems to fix the issue.
 
The unittests in this PR reproduce the issue on master.

```
INFO | txbatcher.TxBatch | to_sweep: []
INFO | txbatcher.TxBatch | to_pay: [<TxOutput script=..... address=..... value=.....>]
INFO | txbatcher.TxBatch | locktime: .....
INFO | coinchooser.CoinChooserPrivacy | Total number of buckets: 11
INFO | coinchooser.CoinChooserPrivacy | Num candidates considered: 24. Winning penalty: 0
INFO | coinchooser.CoinChooserPrivacy | using 3 inputs
INFO | coinchooser.CoinChooserPrivacy | using buckets: ['....']
DEBUG | wallet.Standard_Wallet.[default_wallet] | coinchooser returned tx with 3 inputs and 4 outputs
INFO | txbatcher.TxBatch | created tx .... with 3 inputs and 4 outputs
INFO | transaction | adding signature for ..... spending utxo ....:1
INFO | transaction | adding signature for ..... spending utxo ....:1
INFO | transaction | adding signature for ..... spending utxo ....:0
DEBUG | transaction | tx.sign() finished. is_complete=True
INFO | interface.[......onion:50001] | broadcast_transaction error [DO NOT TRUST THIS MESSAGE]: "RPCError(1, 'the transaction was rejected by network rules.\\n\\ninsufficient fee, rejecting replacement .....; new feerate 0.00001109 BTC/kvB <= old feerate 0.00001110 BTC/kvB\\n[........ (truncated. orig_len=1476). tx=......
INFO | network | error: could not broadcast batch ...., insufficient fee
Your transaction is trying to replace another one in the mempool but it does not meet the rules to do so. Try to increase the fee.
INFO | txbatcher.TxBatch | cannot broadcast tx ....
INFO | address_synchronizer.AddressSynchronizer.[default_wallet] | removing tx from history ....
INFO | txbatcher.TxBatch | starting new batch because could not broadcast
```

### DIFF
--- a/electrum/transaction.py
+++ b/electrum/transaction.py
@@ -1349,7 +1349,14 @@ class Transaction:
         BIP-0141 defines 'Virtual transaction size' to be weight/4 rounded up.
         This definition is only for humans, and has little meaning otherwise.
         If we wanted sub-byte precision, fee calculation should use transaction
-        weights, but for simplicity we approximate that with (virtual_size)x4
+        weights, but for simplicity we approximate that with (virtual_size)x4.
+        note: while we try to estimate as close to the true value as possible,
+            whenever that's not possible, we should over-estimate. E.g. ecdsa DER sig
+            sizes can be 71 or 72 bytes (even 73 though that is non-standard).
+            Over-estimating is preferred as the typical use-case is the user selecting
+            a target_feerate, and the code calculating abs fees as target_feerate*est_size.
+            If we over-estimate est_size there, that means the final true_feerate is going to
+            be higher than target_feerate, which is desirable especially near the min_relay_fee.
         """
         weight = self.estimated_weight()
         return self.virtual_size_from_weight(weight)


electrum/wallet.py
```diff
@@ -2019,16 +2019,16 @@ def make_unsigned_transaction(
...
                 if not isinstance(base_tx, PartialTransaction):
                     base_tx = PartialTransaction.from_tx(base_tx)
                     base_tx.add_info_from_wallet(self)
                 else:
                     # don't cast PartialTransaction, because it removes make_witness
                     base_tx.remove_signatures()
-                base_tx_fee = base_tx.get_fee()
-                base_feerate = Decimal(base_tx_fee)/base_tx.estimated_size()
-                relayfeerate = Decimal(self.relayfee()) / 1000
-                original_fee_estimator = fee_estimator
```
ecdsa, line 2031: can you add a comment explaining why base_feerate should be estimated before stripping signatures? (like in line 2305)

f321x: Added in https://github.com/spesmilo/electrum/compare/180947ea972d089dcfe189d7dd995f50df6f975b..3133148acdfd6d38c3443f165b037c15e079ad43

PR #10461, opened 2026-02-03
Support TS5 in `TrezorClientBase.device_model_name()`

romanz:
Tested locally with an emulator and a physical TS5:

image

image


SomberNight:
Thanks!

PR #10463, opened 2026-02-05
lnwallet: zeroconf/just-in-time improvements and tests

f321x:
Successor of https://github.com/spesmilo/electrum/pull/9584

Some incremental improvements and fixes of the existing just-in-time channel logic and new unittests to prevent code rot:

* client: add fee sanity check to just-in-time client, will reject incoming channels if fee > 10% of channel size
* client: fixes the zeroconf timeout path of `AbstractChannel.update_unfunded_state()`, unittests it
* LSP: check if zeroconf is enabled both for trampoline and regular forwarding, clarify config var
* LSP: stop setting the static scid alias when opening zeroconf channels, extend regtest 
* LSP: add cleanup logic for failed channel openings, add funding tx rebroadcasting logic
* LSP: make channel size and fee configurable, add mining fees to opening fees
* client: stop signaling `OPTION_ZEROCONF_OPT` to untrusted peers, unittest it
* client: fix flickering zeroconf message in ReceiveTab

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -41,7 +41,7 @@ from .bitcoin import redeem_script_to_address
 from .crypto import sha256, sha256d
 from .transaction import Transaction, PartialTransaction, TxInput, Sighash
 from .logging import Logger
-from .lntransport import LNPeerAddr
+from .lntransport import LNPeerAddr, extract_nodeid, ConnStringFormatError
 from .lnonion import OnionRoutingFailure
 from . import lnutil
 from .lnutil import (Outpoint, LocalConfig, RemoteConfig, Keypair, OnlyPubkeyKeypair, ChannelConstraints,
@@ -59,7 +59,7 @@ from .lnsweep import sweep_their_htlctx_justice, sweep_our_htlctx, SweepInfo, Ma
 from .lnsweep import sweep_their_ctx_to_remote_backup
 from .lnhtlc import HTLCManager
 from .lnmsg import encode_msg, decode_msg
-from .address_synchronizer import TX_HEIGHT_LOCAL
+from .address_synchronizer import TX_HEIGHT_LOCAL, TX_HEIGHT_UNCONFIRMED
 from .lnutil import CHANNEL_OPENING_TIMEOUT_BLOCKS, CHANNEL_OPENING_TIMEOUT_SEC
 from .lnutil import ChannelBackupStorage, ImportedChannelBackupStorage, OnchainChannelBackupStorage
 from .lnutil import format_short_channel_id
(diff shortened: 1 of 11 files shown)


electrum/lnchannel.py
```diff
@@ -375,12 +375,14 @@ def update_unfunded_state(self) -> None:
...
             # handling zeroconf channels with no funding tx, can happen if broadcasting fails on LSP side
-            # or if the LSP did double spent the funding tx/never published it intentionally
-            # only remove a timed out OPEN channel if we are connected to the network to prevent removing it if we went
-            # offline before seeing the funding tx
-            if state != ChannelState.OPEN or chan_age > ZEROCONF_TIMEOUT and self.lnworker.network.is_connected():
+            # or if the LSP did double spent the funding tx/never published it intentionally.
+            if not self.lnworker.wallet.is_up_to_date() or not self.lnworker.network \
+                    or self.lnworker.network.blockchain().is_tip_stale():
+                # ensure we are up to date to prevent accidentally dropping a channel that is funded
+                return
```
ecdsa, line 384: what if the electrum server is lying by omission? maybe we should not delete the channel...

f321x: right, maybe we should only freeze it for receiving and then delete it after `has_funding_timed_out` becomes true. This would leave much more time to connect to a honest server and find the funding tx if it really has been omitted.

f321x: Ok https://github.com/spesmilo/electrum/pull/10463/commits/1f17574dfad1cb2b177c3de22a1fac1c3b868909 now freezes the channel after 10 minutes without funding transaction and deletes the channel after `has_funding_timed_out` (~2 weeks). If the channel gets funded it is unfrozen.

ecdsa: why wait 10 minutes? should we not freeze it by default, and unfreeze it when we detect the funding tx?

f321x: I agree this would be simpler. 

As freezing the channel doesn't prevent the channel peer from sending htlcs through it we could simply not release the preimage for any htlc(set) that arrived on a zeroconf channel without funding tx (except for the first one which pays for the channel).
In practice the funding tx should become visible in the mempool quickly, so we wouldn't have to fail any htlcs that already made it to us during this time period.

If the electrum server omits the transaction, we could either fail after some timeout, or the user restarts the wallet, we connect to a different server that tells us about it and can still settle.
On first thought this seems like an acceptable tradeoff, what do you think?

ecdsa: I think this is acceptable for now. In the future, we should implement fraud proofs, so this will need to be updated.

PR #10476, opened 2026-02-11
docs: clarify TL;DR installation steps

yashithac618:
This PR adds a short clarification explaining that the TL;DR
installation commands are Linux-specific and describes the
purpose of the ELECTRUM_ECC_DONT_COMPILE=1 environment variable,
to reduce confusion for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@ but not everything. The following sections describe how to run from source, but 
 is a TL;DR:
 
 ```
+Note:
+- The following commands are intended for Linux systems.
+- The environment variable  `ELECTRUM_ECC_DONT_COMPILE=1` tells Electrum
+  to use the system-installed `libsecp256k1` instead of compiling it locally.
+
 $ sudo apt-get install libsecp256k1-dev
 $ ELECTRUM_ECC_DONT_COMPILE=1 python3 -m pip install --user ".[gui,crypto]"
 ```


README.md
```diff
@@ -22,6 +22,11 @@ but not everything. The following sections describe how to run from source, but
 is a TL;DR:
 
 ```
+Note:
+- The following commands are intended for Linux systems.
+- The environment variable  `ELECTRUM_ECC_DONT_COMPILE=1` tells Electrum
+  to use the system-installed `libsecp256k1` instead of compiling it locally.
```
SomberNight, line 28: this is already explained in the `libsecp256k1` section, ~15 lines below

README.md
```diff
@@ -22,6 +22,11 @@ but not everything. The following sections describe how to run from source, but
 is a TL;DR:
 
 ```
+Note:
```
SomberNight, line 25: Don't put comments this way inside a code block. *code* goes in a code block. If really needed, standard practice is to prefix comments with `#` (as this is a shell code block).

README.md
```diff
@@ -22,6 +22,11 @@ but not everything. The following sections describe how to run from source, but
...
+- The following commands are intended for Linux systems.
```
SomberNight, line 26: I think this is obvious from context and the intent is for the reader to learn what they would need and try to generalise. E.g. when wanting to run on macOS, even though you don't have apt-get, an apt-get line tells you roughly what libraries you need (e.g. from homebrew). Also note that `apt-get` is typically only used on debian-based systems, Fedora has their own thing, Arch has their own thing, etc. So it's not even generic to "Linux systems".

yashithac618: Thanks for the detailed explanation and review, I understand the reasoning now.
Appreciate you taking the time to explain this and for maintaining the project!

PR #10479, opened 2026-02-16
locale: gui: show translation completion percentage in language names

SomberNight:
In the GUIs, on the language-select screen, show e.g.
```
  Czech (100%), Danish (13%), Dutch (54%)
```
instead of
```
  Czech, Danish, Dutch
```

-----

image


electrum/i18n.py
```diff
@@ -168,3 +168,53 @@ def set_language(x: Optional[str]) -> None:
...
 }
 assert '' in languages
+
+
+def get_gui_lang_names(*, show_completion_percent: bool = False) -> dict[str, str]:
+    """Returns a  lang_code -> lang_name  mapping, sorted.
+
+    If show_completion_percent is True, lang_name includes a % estimate for translation completeness.
+    The completion percentage calculation is cached, however the first call is still expensive:
+    on my decentish laptop it takes ~40 ms to calculate, so on an old phone, it might be noticeable.
```
SomberNight: My proposed compromise is to show the fancy stuff in the Qt GUI, but not on Android. Given it takes 40 ms on my laptop, I guess it could easily take 10x that on an old phone.

ecdsa: why is it computed at runtime? we could perform this calculation in build_locale

SomberNight: Do you mean build_locale could generate e.g. a json file with the results, which we would bundle into the binaries/distributables? That could work, yes.

EDIT: I calculate it at runtime as that seemed like the easy thing to do. Given it is more expensive than expected, indeed doing it at build-time might be better.

SomberNight: done in https://github.com/spesmilo/electrum/pull/10479/changes/895923c0dfd2cdeb93b30325137e05183a45bd1a
but I have not tested it enough yet

SomberNight:
note: At one point I was considering keeping all commits in this branch, to show how previous implementations worked, to show rationale of why the final impl works as-is, however I now changed my mind and squashed the commits. I expanded the commit message to try to add the rationale there.

In any case, backup of old branch prior to squash: https://github.com/SomberNight/electrum/tree/202602_locale_fancy_names_bak20260221 (https://github.com/SomberNight/electrum/commit/4b6c15c346ba24dc7971452d1082482620cf13ad)

PR #10482, opened 2026-02-16
Remove non-existing electrum server on owner's request

Sanksari12:
Remove lavahost.org from servers.json as the host does not host a public electrum instance.

As the owner of the domain I ask you to please remove the entry as I get bombarded with requests to the non-existing electrum server.
I have published a DNS TXT record on the domain lavahost.org confirming that request and authenticating it.

### DIFF
--- a/electrum/chains/mainnet/servers.json
+++ b/electrum/chains/mainnet/servers.json
@@ -366,11 +366,6 @@
         "t": "50001",
         "version": "1.4.2"
     },
-    "lavahost.org": {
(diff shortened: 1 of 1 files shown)


SomberNight:
>  I have published a DNS TXT record on the domain lavahost.org confirming that request and authenticating it.

Checks out, thx.

> as the host does not host a public electrum instance

There used to be a server there when it was added to the list, but perhaps not anymore :)

Anyway, we will honour your request of course.

PR #10485, opened 2026-02-19
p4a rebase, use ndk28 and qt6.10

accumulator:
This updates P4A to a much more recent branch point (~2025-10), right before upstream moves to python3.14.

- NDK upgraded to 28c. This is more recent than the Qt6.10 [supported](https://doc.qt.io/qt-6/android.html) NDK27c, but this seems to work without issues, and allows us to drop the 16kb alignment patches.
- Qt upgraded to 6.10.2
- p4a minapi increased to 26 (Android 8.0)
- pyqt6sip upgraded to 13.10.3
- pyqt_builder upgraded to 1.19.1
- pyqt6 upgraded to 6.10.2
- sip upgraded to 6.15.1
- openssl (implicitly) upgraded to 3.3.1

patches dropped:
- 16 kb alignment patches for qt6, cffi, sqlite, pyqt6, qt6-bootstrap
- recipes: bump openssl to 3.0.x (e91310fbf88857b2fbd9f9a5f591300ae5da90b5)
- python: update to 3.11.5 (a01269f7799587ad74ee40e0b642d917b8db7d4e)
- patches for reproducibility (8e3bad863d52ae537066d222d9cfe177b868ec9d, e04b87585714114a626a34b558066144a6c91217, 39cc65c1812a9937b0b0e4afda18fb40ab0fa1b0)
- Update `cffi` recipe for Python 3.10 (41fee3e938d13b6701b205da8c17d0d444188e63)
- recipes: add patches for python3 recipes, for reproducible .pyc generation (0ab0d872e6c6b88ddc05b9c4ba6fcd3aa7921242)

patches added:
- compat-windowinsets.patch to avoid crashing on requesting window insets on Android 8 (375a05de21b538d704174b1efeb3fc85d151f94e)

other changes:
- squashed all of Qt5 history to a single commit
- squashed all of Qt6.4 history to a single commit
- qt6 compile option `-no-feature-getentropy`, this feature requires API28 and has a fallback reading `/dev/urandom`. (Note: we don't use any cryptographic primitives from Qt, so this should not be potentially problematic)

changes in QML GUI:
- changed all `pyqtProperty` accessors that also have a setter to pass `QVariant` instead of the actual type, to work around [this](https://www.riverbankcomputing.com/pipermail/pyqt/2024-June/045922.html) PyQt issue.
- changed QML typed properties declarations that use custom types as `var` instead.
- to restore Android8 compatibility, direct imports of `QtMultimedia` are removed. (`QtMultimedia` is currently not needed on android, but is used for camera/scanning on desktop)
- styling updates due to new Material 3 theme


accumulator:
~~Note: we need to check if the build is still reproducible!~~
~~Note2: styling is probably inconsistent in many places~~
~~Note3: this should only be merged *after* the next release, as currently android 7 and 8 are not working due to biometric java classes crash, and this branch will NOT work anymore on android 7. The next release therefore should indicate current minapi (23) to get electrum working on current android7 devices before moving to a new minapi (26), otherwise android7 will get stuck with a non-working version without the option to downgrade.~~

SomberNight:
hey, sorry for lack of feedback. It is most certainly *not* due to lack of interest! :)

Did you close this just to make it clear we should not include it in 4.7.1 (which I agree with - and now 4.7.x is branched off, and a huge change like this I would not want to cherry-pick to 4.7.x), or is it for some other reason?
Is it ready for testing/review?

accumulator:
oops, I don't recall closing this.. reopening

accumulator:
> Is it ready for testing/review?

Yes

f321x:
When running this as a debug apk on my Pixel 9 (Android 16) i noticed two issues:

This `Unexpected nullptr parameter` warning shows up when stopping the app:
```
03-04 10:08:04.941 19730 19764 I python  :  11.68 | I | plugin.Plugins | jnius detach
03-04 10:08:04.942 19730 19764 I python  :  11.69 | I | plugin.Plugins | stopped
03-04 10:08:04.945 19730 19763 I python  :  11.69 | I | daemon.Daemon | removing lockfile
03-04 10:08:04.946 19730 19763 I python  :  11.69 | I | daemon.Daemon | stopped
03-04 10:08:05.022 19730 19753 I python  :  11.77 | W | gui.qml.qeapp | QObject::disconnect: Unexpected nullptr parameter
```

The app crashes most of the time when starting it (like 4 out of 5 times), when cancelling the following Biometric authentication it crashes almost certainly. However when i manage to open it it runs reliably and doesn't crash anymore. Seems to be some kind of race during startup? I noticed the following log lines during these crashes:
```
03-04 10:08:46.762 19813 19846 I python  :   2.92 | D | gui.qml.qenetwork | fee histogram updated
03-04 10:08:47.238 19813 19813 I python  :   3.40 | D | gui.qml.qebiometrics | biometric auth cancelled by user
03-04 10:08:47.238 19813 19835 I python  :   3.40 | W | gui.qml.qeapp | Biometric auth failed: CANCELLED
03-04 10:08:47.287 19813 19867 I python  :   3.45 | W | gui.qml.qeapp | Failed to acquire deadlock protector for 'QAndroidPlatformOpenGLWindow::eglSurface()' while already locked by 'QtAndroidAccessibility::runInObjectContext()'.
03-04 10:08:47.289 19813 19867 I python  :   3.45 | W | gui.qml.qeapp | Failed to acquire deadlock protector for QAndroidPlatformOpenGLWindow::eglSurface().
```

```
03-04 10:09:09.983 19895 19917 I python  :   1.08 | I | gui.qml.ElectrumGui | Entering main loop
03-04 10:09:09.997 19895 19917 I python  :   1.09 | W | gui.qml.qeapp | wallet requires password
03-04 10:09:10.001 19895 19917 I python  :   1.10 | D | gui.qml.qebiometrics | _start_activity: DECRYPT, len(data)=133
03-04 10:09:10.015 19895 19917 I python  :   1.11 | W | gui.qml.qeapp | daemon loading false
03-04 10:09:10.070 19895 19942 I python  :   1.17 | W | gui.qml.qeapp | Failed to acquire deadlock protector for 'QAndroidPlatformOpenGLWindow::eglSurface()' while already locked by 'QtAndroidAccessibility::runInObjectContext()'.
03-04 10:09:10.077 19895 19942 I python  :   1.17 | W | gui.qml.qeapp | Failed to acquire deadlock protector for QAndroidPlatformOpenGLWindow::eglSurface().
03-04 10:09:10.086 19895 19928 I python  :   1.18 | I | lnworker.LNGossip | starting taskgroup.
03-04 10:09:10.087 19895 19928 I python  :   1.18 | I | lnworker.LNPeerManager.[LNGossip] | starting taskgroup.
03-04 10:09:10.470 19895 19928 I python  :   1.56 | I | interface.[5.9.83.108:50002] | connection established. version: ['ElectrumX 1.18.0', '1.6'], handshake duration: 138.00 ms
```

This is the Android crash report:
```
type: crash
osVersion: google/tokay/tokay:16/BP4A.260205.002/2026021201:user/release-keys
flags: dev options enabled
package: org.electrum.electrum:45407013, targetSdk 35
process: org.electrum.electrum
installer: com.android.packageinstaller

signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------

backtrace:
      #00 pc 000000000007d9c0  /apex/com.android.runtime/lib64/bionic/libc.so (abort+160) (BuildId: b2e2593ea9af5cb426017f2c32a8fcf5)
      #01 pc 0000000000304aa0  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Core_arm64-v8a.so (BuildId: 8931ee3d5926136691ad4138d5a045c25d029803)
      #02 pc 000000000031acb4  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Core_arm64-v8a.so (BuildId: 8931ee3d5926136691ad4138d5a045c25d029803)
      #03 pc 000000000031b550  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Core_arm64-v8a.so (QMessageLogger::fatal(char const*, ...) const+152) (BuildId: 8931ee3d5926136691ad4138d5a045c25d029803)
      #04 pc 000000000009a848  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libplugins_platforms_qtforandroid_arm64-v8a.so (BuildId: 273879cc841d14aae157ecda0dfc0530629054c7)
      #05 pc 000000000009a318  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libplugins_platforms_qtforandroid_arm64-v8a.so (BuildId: 273879cc841d14aae157ecda0dfc0530629054c7)
      #06 pc 0000000000581788  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Gui_arm64-v8a.so (QOpenGLContext::makeCurrent(QSurface*)+144) (BuildId: a56b29ee58a4afe709da073c19350f00467b1f4f)
      #07 pc 000000000058dfd8  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Gui_arm64-v8a.so (BuildId: a56b29ee58a4afe709da073c19350f00467b1f4f)
      #08 pc 000000000058e264  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Gui_arm64-v8a.so (BuildId: a56b29ee58a4afe709da073c19350f00467b1f4f)
      #09 pc 000000000049fdb8  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Gui_arm64-v8a.so (QRhi::create(QRhi::Implementation, QRhiInitParams*, QFlags<QRhi::Flag>, QRhiNativeHandles*, QRhiAdapter*)+152) (BuildId: a56b29ee58a4afe709da073c19350f00467b1f4f)
      #10 pc 00000000004f1014  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Quick_arm64-v8a.so (QSGRhiSupport::createRhi(QQuickWindow*, QSurface*, bool)+1036) (BuildId: 06be7c46a481e005d36bba0d2ee08484b99c24d7)
      #11 pc 00000000005f2238  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Quick_arm64-v8a.so (BuildId: 06be7c46a481e005d36bba0d2ee08484b99c24d7)
      #12 pc 00000000005f2448  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Quick_arm64-v8a.so (BuildId: 06be7c46a481e005d36bba0d2ee08484b99c24d7)
      #13 pc 000000000048ce24  /data/app/~~Ugakqqk7pSuNKMd_DGYNLA==/org.electrum.electrum-xUuIdVbQ5QFI0x95-tbQig==/lib/arm64/libQt6Core_arm64-v8a.so (BuildId: 8931ee3d5926136691ad4138d5a045c25d029803)
      #14 pc 0000000000091584  /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_start(void*) (.__uniq.67847048707805468364044055584648682506)+180) (BuildId: b2e2593ea9af5cb426017f2c32a8fcf5)
      #15 pc 00000000000813d4  /apex/com.android.runtime/lib64/bionic/libc.so (__start_thread+68) (BuildId: b2e2593ea9af5cb426017f2c32a8fcf5)
```

accumulator:
> When running this as a debug apk on my Pixel 9 (Android 16) i noticed two issues:
> 
> This `Unexpected nullptr parameter` warning shows up when stopping the app:
> 
> ```
> 03-04 10:08:04.941 19730 19764 I python  :  11.68 | I | plugin.Plugins | jnius detach
> 03-04 10:08:04.942 19730 19764 I python  :  11.69 | I | plugin.Plugins | stopped
> 03-04 10:08:04.945 19730 19763 I python  :  11.69 | I | daemon.Daemon | removing lockfile
> 03-04 10:08:04.946 19730 19763 I python  :  11.69 | I | daemon.Daemon | stopped
> 03-04 10:08:05.022 19730 19753 I python  :  11.77 | W | gui.qml.qeapp | QObject::disconnect: Unexpected nullptr parameter
> ```

Yes I see the same.


> The app crashes most of the time when starting it (like 4 out of 5 times), when cancelling the following Biometric authentication it crashes almost certainly. However when i manage to open it it runs reliably and doesn't crash anymore. Seems to be some kind of race during startup? 

I haven't seen this at all. I'm also using debug apk on Pixel 9. I assume you fully cleaned the buildozer build tree first?

> I noticed the following log lines during these crashes:
> 
> ```
> 03-04 10:08:46.762 19813 19846 I python  :   2.92 | D | gui.qml.qenetwork | fee histogram updated
> 03-04 10:08:47.238 19813 19813 I python  :   3.40 | D | gui.qml.qebiometrics | biometric auth cancelled by user
> 03-04 10:08:47.238 19813 19835 I python  :   3.40 | W | gui.qml.qeapp | Biometric auth failed: CANCELLED
> 03-04 10:08:47.287 19813 19867 I python  :   3.45 | W | gui.qml.qeapp | Failed to acquire deadlock protector for 'QAndroidPlatformOpenGLWindow::eglSurface()' while already locked by 'QtAndroidAccessibility::runInObjectContext()'.
> 03-04 10:08:47.289 19813 19867 I python  :   3.45 | W | gui.qml.qeapp | Failed to acquire deadlock protector for QAndroidPlatformOpenGLWindow::eglSurface().
> ```
> 
> ```
> 03-04 10:09:09.983 19895 19917 I python  :   1.08 | I | gui.qml.ElectrumGui | Entering main loop
> 03-04 10:09:09.997 19895 19917 I python  :   1.09 | W | gui.qml.qeapp | wallet requires password
> 03-04 10:09:10.001 19895 19917 I python  :   1.10 | D | gui.qml.qebiometrics | _start_activity: DECRYPT, len(data)=133
> 03-04 10:09:10.015 19895 19917 I python  :   1.11 | W | gui.qml.qeapp | daemon loading false
> 03-04 10:09:10.070 19895 19942 I python  :   1.17 | W | gui.qml.qeapp | Failed to acquire deadlock protector for 'QAndroidPlatformOpenGLWindow::eglSurface()' while already locked by 'QtAndroidAccessibility::runInObjectContext()'.
> 03-04 10:09:10.077 19895 19942 I python  :   1.17 | W | gui.qml.qeapp | Failed to acquire deadlock protector for QAndroidPlatformOpenGLWindow::eglSurface().
> 03-04 10:09:10.086 19895 19928 I python  :   1.18 | I | lnworker.LNGossip | starting taskgroup.
> 03-04 10:09:10.087 19895 19928 I python  :   1.18 | I | lnworker.LNPeerManager.[LNGossip] | starting taskgroup.
> 03-04 10:09:10.470 19895 19928 I python  :   1.56 | I | interface.[5.9.83.108:50002] | connection established. version: ['ElectrumX 1.18.0', '1.6'], handshake duration: 138.00 ms
> ```

This is my log. It has some warnings and errors, but never crashes and seems to run fine:
```
03-04 12:00:11.388 20865 20887 I python  :   1.09 | I | gui.qml.ElectrumGui | Entering main loop
03-04 12:00:11.399 20865 20887 I python  :   1.10 | W | gui.qml.qeapp | wallet requires password
03-04 12:00:11.403 20865 20887 I python  :   1.11 | D | gui.qml.qebiometrics | _start_activity: DECRYPT, len(data)=133
03-04 12:00:11.416 20865 20887 I python  :   1.12 | W | gui.qml.qeapp | daemon loading false
03-04 12:00:11.974 20865 20909 I python  :   1.68 | W | gui.qml.qeapp | makeCurrent(): no EGLSurface, likely Surface destroyed by Android.
03-04 12:00:11.974 20865 20909 I python  :   1.68 | W | gui.qml.qeapp | QRhiGles2: Failed to make context current. Expect bad things to happen.
03-04 12:00:11.974 20865 20909 I python  :   1.68 | W | gui.qml.qeapp | Failed to start frame
03-04 12:00:12.713 20865 20897 I python  :   2.42 | D | exchange_rate.CoinGecko | received fx quotes
03-04 12:00:12.714 20865 20897 I python  :   2.42 | D | gui.qml.qefx | new quotes
03-04 12:00:12.807 20865 20897 I python  :   2.51 | I | interface.[skbxmit.coinjoined.com:50002] | connection established. version: ['ElectrumX 1.19.0', '1.6'], handshake duration: 192.64 ms
03-04 12:00:12.877 20865 20897 I python  :   2.58 | I | interface.[skbxmit.coinjoined.com:50002] | set blockchain with height 939251
03-04 12:00:12.881 20865 20897 I python  :   2.59 | I | interface.[skbxmit.coinjoined.com:50002] | new chain tip. height=939252
03-04 12:00:12.881 20865 20897 I python  :   2.59 | D | gui.qml.qenetwork | new height: 939252
03-04 12:00:12.884 20865 20897 I python  :   2.59 | D | gui.qml.qenetwork | network_status updated: Connected to 1 node.
03-04 12:00:12.885 20865 20897 I python  :   2.59 | D | gui.qml.qenetwork | chain tips # changed: 1
03-04 12:00:12.927 20865 20897 I python  :   2.63 | I | network | fee_estimates {}
03-04 12:00:14.341 20865 20865 I python  :   4.05 | D | gui.qml.qebiometrics | biometric auth cancelled by user
03-04 12:00:14.344 20865 20887 I python  :   4.05 | W | gui.qml.qeapp | Biometric auth failed: CANCELLED
03-04 12:00:14.430 20865 20909 I python  :   4.14 | W | gui.qml.qeapp | QEGLPlatformContext: eglSwapBuffers failed: 300d
03-04 12:00:14.460 20865 20909 E Surface : [733f4e6 SurfaceView[org.electrum.electrum/org.kivy.android.PythonActivity]#4(BLAST Consumer)4](id:518100000004) clearBuffersForDisconnectLocked: 1 buffers were freed while being dequeued!
03-04 12:00:14.592  1391  6537 D CoreBackPreview: Window{eae2219 u0 org.electrum.electrum/org.kivy.android.PythonActivity}: Setting back callback OnBackInvokedCallbackInfo{mCallback=android.window.IOnBackInvokedCallback$Stub$Proxy@27d23d6, mPriority=0, mIsAnimationCallback=true, mOverrideBehavior=0}
03-04 12:00:14.640 20865 20865 D VRI[PythonActivity]: WindowInsets changed: ime:[0,0,0,796] 
```

f321x:
weird.
> I assume you fully cleaned the buildozer build tree first?

yes, i started in a clean repo with fresh container.

Also i am using GrapheneOS, but had all optional exploit protections disabled, so this shouldn't be the reason.

f321x:
The recent changes fixed the crash for me, https://github.com/spesmilo/electrum/pull/10485/commits/3ebe8e5aa4ebad8f6652ca0e14a83a82c3427451 now works fine.

contrib/android/p4a_recipes/openssl/__init__.py
```diff
@@ -6,14 +6,13 @@
 util = load_source('util', os.path.join(os.path.dirname(os.path.dirname(__file__)), 'util.py'))
 
 
-assert OpenSSLRecipe._version == "3.0.18"
+assert OpenSSLRecipe._version == "3.3.1"
```
SomberNight: sneaky way to change to an older openssl, lol :)
just kidding, but indeed this larger version number is around 1 year older than the 3.0.18 LTS

I think it's easiest to stay on either the 3.0 or the 3.5 LTS branches.
https://openssl-library.org/source/

contrib/android/p4a_recipes/python3/__init__.py
```diff
@@ -13,7 +13,7 @@
 class Python3RecipePinned(util.InheritedRecipeMixin, Python3Recipe):
     # PYTHON_VERSION=    # < line here so that I can grep the codebase and teleport here
     version = "3.11.14"
-    sha512sum = "41fb3ae22ce4ac0e8bb6b9ae8db88a810af1001d944e3f1abc9e86824ae4be31347e3e3a70425ab12271c6b7eeef552f00164ef23cfffa2551c3c9d1fe5ab91f"
+    sha512sum = "4642f6d59c76c6e5dbd827fdb28694376a9cc76e513146d092b49afb41513b3c9dff2339cfcebfb5b260f5cdc49a59a69906e284e5d478b2189d3374e9e24fd5"
```
SomberNight, line 16: I am confused. Why did this hash change if the corresponding version number did not?

SomberNight: answer: it is because p4a is now downloading archives from github instead of official releases from python.org
https://github.com/kivy/python-for-android/pull/3180/changes#diff-6aaa5c7168576825b0ac43915df1873d45e5594eee8189ec7c717cf89bcfe400R59

electrum/gui/qml/qeaddressdetails.py
```diff
@@ -35,12 +35,13 @@ def __init__(self, parent=None):
         self._historyModel = None
 
     walletChanged = pyqtSignal()
-    @pyqtProperty(QEWallet, notify=walletChanged)
+    @pyqtProperty(QVariant, notify=walletChanged)
     def wallet(self):
         return self._wallet
```
SomberNight, line 40: note: I think we should (also) use python type-hints to signal the types. Previously they were kind of redundant - repeating what was already in the decorator - but now they are pure signal to the reader:
```py
    @pyqtProperty(QVariant, notify=walletChanged)
    def wallet(self) -> QEWallet:
        return self._wallet
```

contrib/android/buildozer_qml.spec
```diff
@@ -111,13 +111,13 @@ android.api = 35
 android.target_sdk_version = 35
 
 # (int) Minimum API required. You will need to set the android.ndk_api to be as low as this value.
-android.minapi = 23
+android.minapi = 26
```
SomberNight, line 114: Are you sure android 8 is actually supported? Did you test in an emulator or similar?

For Qt 6.8, 6.9, 6.10, the Qt docs say the minimum supported is Android 9.
https://doc.qt.io/qt-6.8/android.html#supported-configurations

note: this is not a blocker, just want to clarify. Also it would not be nice to declare lower minimum than what we actually support, to avoid the kind of "users upgrading and getting stuck" situation you also mentioned re 4.7.1

contrib/android/Dockerfile
SomberNight, line 1: > Note: we need to check if the build is still reproducible!

Tested that, looks okay. :)

```
$ git log -n 1
commit 660f2ead20153ed97771a55a6c91a02af4dbe5bb (HEAD, spesmilo/pr/10485)
Author: f321x <f@f321x.com>
Date:   Thu Mar 5 16:57:54 2026 +0100

    android: update Qt6 to 6.10.2, PyQt6 to 6.10.2
$ sha256sum dist/*
f2bb8df7b1de385abcb3abd029cb36acd2e7b5d099d6bdec81a33f1b33cb6d64  dist/Electrum-4.7.1-arm64-v8a-release-unsigned1.apk
f2bb8df7b1de385abcb3abd029cb36acd2e7b5d099d6bdec81a33f1b33cb6d64  dist/Electrum-4.7.1-arm64-v8a-release-unsigned2.apk
```

contrib/android/p4a_recipes/pyjnius/__init__.py
```diff
@@ -6,14 +6,13 @@
 util = load_source('util', os.path.join(os.path.dirname(os.path.dirname(__file__)), 'util.py'))
 
 
-assert PyjniusRecipe._version == "1.5.0"
-assert PyjniusRecipe.depends == [('genericndkbuild', 'sdl2', 'qt6'), 'six', 'python3']
+assert PyjniusRecipe._version == "1.7.0"
+assert PyjniusRecipe.depends == [('genericndkbuild', 'sdl2', 'sdl3', 'qt6'), 'six', 'python3'], PyjniusRecipe.depends
```
SomberNight, line 10: Something fishy is going on.

New pyjnius requires Cython 3.1.x [[ref]](https://github.com/kivy/pyjnius/blob/53390891ac90ab764ffd6e04170567f181456265/pyproject.toml#L5), but ISTM we only install Cython 0.29 [[ref]](https://github.com/spesmilo/electrum/commit/03400977546f7bd39946b3fba7192e86e748be16).

---

Ok, so I believe it is due to the `recipe.hostpython_prerequisites` introduced in https://github.com/kivy/python-for-android/commit/8110fafabc64192de627807da3f4425779c37969.

Uhum... so this is due to this `hostpython_prerequisites`, installing

https://github.com/spesmilo/python-for-android/blob/ee6886ec3048670b182afb18237e0ae69fde1254/pythonforandroid/recipes/pyjnius/__init__.py#L14
https://github.com/spesmilo/python-for-android/blob/ee6886ec3048670b182afb18237e0ae69fde1254/pythonforandroid/recipes/android/__init__.py#L16
https://github.com/spesmilo/python-for-android/blob/ee6886ec3048670b182afb18237e0ae69fde1254/pythonforandroid/recipes/pyqt6/__init__.py#L18
Looks like all kinds of unpinned packages are getting installed at build-time willy-nilly.

Uhh this [`install_hostpython_prerequisites`](https://github.com/spesmilo/python-for-android/blob/ee6886ec3048670b182afb18237e0ae69fde1254/pythonforandroid/recipe.py#L1066) is a real can of worms. (esp with `force_upgrade=True`!)
So it is calling pip with internet and downloading packages without hash-pins from PyPI. Often without even version-pins. We can't have this... everything must be pinned.

Perhaps we should make sure that `install_hostpython_prerequisites` always gets called with a `packages` arg that contains hashes. We could patch the `pip_options` internal var to add `--require-hashes`, as a final sanity check.

Also, this `recipe.hostpython_prerequisites` should probably be something we assert/check in our contrib/android/p4a_recipes, [similar to `depends`](https://github.com/spesmilo/electrum/blob/169227e148b1fe542e8832cb5c6ea86f887dc200/contrib/android/p4a_recipes/pyqt6/__init__.py#L10-L11) and `python_depends`. We really don't want any of these silently changing.

SomberNight (changes requested):
None of my other comments are too important, but this one is a blocker.
We must hash pin everything we pull in.

( self-note: reviewed 660f2ead20153ed97771a55a6c91a02af4dbe5bb )

contrib/android/buildozer_qml.spec
accumulator: Yes, I tested various older android versions (emulator). I've added a few workarounds for API26/27 as it was low-hanging fruit. Unfortunately Android7 was not easily done, due to extensive >API24 usage within Qt.. 

Unfortunately we erroneously marked API23 as the lowest supported version, but Android6 doesn't work with Qt6.4, so it should have been API24 after moving from Qt5 to Qt6 :-/

contrib/android/p4a_recipes/pyjnius/__init__.py
accumulator: Yes, hostpython package handling considerably changed in p4a upstream. Pinning will take some work.

contrib/android/p4a_recipes/openssl/__init__.py
accumulator: huh? ok, interesting. good catch too, I made an assumption here..

I think we can pin on 3.0.18 instead, without much compile troubles

accumulator:
I've added support for hash pinning the `hostpython_prerequisites`. The `hostpython_prerequisites` are not compiled but pulled binary from pypi. I've changed `sip` and `pyqt_builder` from prereq to dep, so that we compile it ourselves. `packaging` is now not compiled but pulled in as binary instead. If we want to compile it instead this could take some work to accomplish (It is now pulled in as transitive dependency for the `build[virtualenv]` prerequisite for `PyProjectRecipe` recipies). The entire dependency tree for `PyProjectRecipe` prereqs is hash pinned in the `hostpython3` recipe.

> Yes, hostpython package handling considerably changed in p4a upstream.

.. and it still sucks

contrib/android/p4a_recipes/setuptools/__init__.py
```diff
@@ -1,13 +1,19 @@
...
 assert SetuptoolsRecipe.depends == ['python3']
 assert SetuptoolsRecipe.python_depends == []
 
 
 class SetuptoolsRecipePinned(SetuptoolsRecipe):
-    sha512sum = "5a3572466a68c6f650111448ce3343f64c62044650bb8635edbff97e2bc7b216b8bbe3b4e3bccf34e6887f3bedc911b27ca5f9a515201cae49cf44fbacf03345"
+    hostpython_prerequisites = [
+        HashPinnedDependency(package='setuptools==80.9.0',
+                             hashes=[]),
+    ]
```
SomberNight: why is the hash of the host setuptools not pinned?

also, shouldn't `PythonRecipe.install_hostpython_prerequisites` raise an exception in this case?
I would have expected `pip install --require-hashes` to error out (?)

contrib/android/p4a_recipes/hostpython3/__init__.py
```diff
@@ -15,5 +15,31 @@ class HostPython3RecipePinned(util.InheritedRecipeMixin, HostPython3Recipe):
...
+                             hashes=['sha256:1bd755b504931164a5a496d217c014d098426cddc79363ad66ac78125f9d908f']),
+        HashPinnedDependency(package="distlib==0.4.0",
+                             hashes=['sha256:9659f7d87e46584a30b5780e43ac7a2143098441670ff0a49d5f9034c54a6c16']),
+        HashPinnedDependency(package="filelock==3.25.2",
+                             hashes=['sha256:ca8afb0da15f229774c9ad1b455ed96e85a81373065fb10446672f64444ddf70']),
+        HashPinnedDependency(package="platformdirs==4.9.4",
+                             hashes=['sha256:68a9a4619a666ea6439f2ff250c12a853cd1cbd5158d258bd824a7df6be2f868']),
+        HashPinnedDependency(package="python_discovery==1.1.3",
+                             hashes=['sha256:90e795f0121bc84572e737c9aa9966311b9fde44ffb88a5953b3ec9b31c6945e']),
+    ]
```
SomberNight: > `packaging` [...] is now pulled in as transitive dependency for the `build[virtualenv]` prerequisite for `PyProjectRecipe` recipies. The entire dependency tree for `PyProjectRecipe` prereqs is hash pinned in the `hostpython3` recipe.


is `virtualenv` actually needed? can't we just use stdlib `venv`?
I see that upstream declared it there but what specifically is needed that `venv` is missing?
that would rid us of all these transitive deps

contrib/android/p4a_recipes/pyjnius/__init__.py
SomberNight: > I've added support for hash pinning the `hostpython_prerequisites`.

Nice. Thanks a lot!

> The `hostpython_prerequisites` are not compiled but pulled binary from pypi. `packaging` is now not compiled but pulled in as binary instead. If we want to compile it instead this could take some work to accomplish

Ok, I think it's acceptable to pull in some pure-python .wheels from PyPI for a small number of popular packages. As long as everything is hash-pinned.

In general we should make sure everything is hash-pinned and instead of logging warnings or silently ignoring issues, we should add belt-and-suspender sanity-check hard-fails, make sure we don't just accidentally pull in the unpinned current latest of version of something.

contrib/android/p4a_recipes/setuptools/__init__.py
accumulator: It does not trigger because apparently it isn't used as a compiled dependency..

contrib/android/p4a_recipes/hostpython3/__init__.py
accumulator: I'll have a look

accumulator: using `venv` now, https://github.com/spesmilo/electrum/pull/10485/commits/ca5e374df5d2bbeab35491769251be859507d5fc

contrib/android/Dockerfile
```diff
@@ -48,31 +44,21 @@ ENV ANDROID_NDK_DL_URL="https://dl.google.com/android/repository/${ANDROID_NDK_A
...
-#    && mkdir --parents "${ANDROID_NDK_HOME_V}" \
-#    && unzip -q "${ANDROID_NDK_ARCHIVE}" -d "${ANDROID_HOME}" \
-#    && ln -sfn "${ANDROID_NDK_HOME_V}" "${ANDROID_NDK_HOME}" \
-#    && rm -rf "${ANDROID_NDK_ARCHIVE}"
-
-# temporary build using NDK from CI
-ENV CI_REV="12186248"
-ENV CI_NDK_FILE="android-ndk-${CI_REV}-linux-x86_64.zip"
-COPY contrib/android/dl-ndk-ci.sh /tmp/
-RUN /tmp/dl-ndk-ci.sh https://ci.android.com/builds/submitted/${CI_REV}/linux/latest/${CI_NDK_FILE} \
```
SomberNight, line 64: can delete `dl-ndk-ci.sh` from the repo

contrib/android/p4a_recipes/pyparsing/__init__.py
```diff
@@ -15,5 +16,12 @@ class PyparsingRecipePinned(PyparsingRecipe):
     #         see "PyProjectRecipe" from https://github.com/kivy/python-for-android/pull/3007
     sha512sum = "1e692f4cdaa6b6e8ca2729d0a3e2ba16d978f1957c538b6de3a4220ec7d996bdbe87c41c43abab851fffa3b0498a05841373e435602917b8c095042e273badb5"
 
+    hostpython_prerequisites = [
+        HashPinnedDependency(package="setuptools==80.9.0",
+                             hashes=['sha256:062d34222ad13e0cc312a4c02d73f059e86a4acbfbdea8f8f76b28c99f306922']),
+        HashPinnedDependency(package="pip==24.0",
+                             hashes=['sha256:ba0d021a166865d2265246961bec0152ff124de910c5cc39f1156ce3fa7c69dc']),
+    ]
```
SomberNight, line 24: > In general we should make sure everything is hash-pinned and instead of logging warnings or silently ignoring issues, we should add belt-and-suspender sanity-check hard-fails, make sure we don't just accidentally pull in the unpinned current latest of version of something.

When I said we should hard fail if the hashpins are missing, I meant it. I really think we must.
Logged warnings we might not notice for months. :/
 
```diff
diff --git a/pythonforandroid/recipe.py b/pythonforandroid/recipe.py
index d3617caa..fe72ea45 100644
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -1082,7 +1082,8 @@ class PythonRecipe(Recipe):
         hash_pin = any(isinstance(package, HashPinnedDependency) for package in packages)
 
         if not hash_pin:
-            warning('hostpython_prerequisites no hash pinning')
+            error(f"hostpython_prerequisites has no hash pinning for recipe={self}")
+            exit(1)
 
         with temp_directory() as tempdir:
             with open(join(tempdir, 'requirements.txt'), 'w') as reqfile:
@@ -1105,7 +1106,7 @@ class PythonRecipe(Recipe):
             ]
 
             if hash_pin:
-                pip_options.append('--require-hashes')
+                pip_options.append('--require-hashes')  # any pkg hash-pinned => *all* must be hash-pinned
 
             if pip_extra_args:
                 pip_options += pip_extra_args

```
Indeed some packages are missing hashpins:
```
[ERROR]:   hostpython_prerequisites has no hash pinning for recipe=<pythonforandroid.recipes.pyqt6sip.PyQt6SipRecipePinned object at 0x7f61b2ea4320>
[ERROR]:   hostpython_prerequisites has no hash pinning for recipe=<pythonforandroid.recipes.sip.SipRecipePinned object at 0x7f10974887d0>
```

accumulator: > > In general we should make sure everything is hash-pinned and instead of logging warnings or silently ignoring issues, we should add belt-and-suspender sanity-check hard-fails, make sure we don't just accidentally pull in the unpinned current latest of version of something.
> 
> When I said we should hard fail if the hashpins are missing, I meant it. I really think we must. Logged warnings we might not notice for months. :/

You're right, it was a warning still, I cherry-picked your commit.

> Indeed some packages are missing hashpins:
> 
> ```
> [ERROR]:   hostpython_prerequisites has no hash pinning for recipe=<pythonforandroid.recipes.pyqt6sip.PyQt6SipRecipePinned object at 0x7f61b2ea4320>
> [ERROR]:   hostpython_prerequisites has no hash pinning for recipe=<pythonforandroid.recipes.sip.SipRecipePinned object at 0x7f10974887d0>
> ```

These were still local, forgot to push the last commit :)

SomberNight, line 24: In general I am a bit worried about how we will be able to maintain hash-pins like these. Luckily there are not that many atm.

contrib/android/Dockerfile
```diff
@@ -250,8 +235,8 @@ RUN cd /opt \
     && /opt/venv/bin/python3 -m pip install --no-build-isolation --no-dependencies -e .
 
 # install python-for-android
-ENV P4A_CHECKOUT_COMMIT="a01269f7799587ad74ee40e0b642d917b8db7d4e"
-# ^ from branch electrum_20251211 (note: careful with force-pushing! see #8162)
+ENV P4A_CHECKOUT_COMMIT="1098be6964cfc2156959e435e81c2c50f8398586"
+# ^ from branch electrum_202602 (note: careful with force-pushing! see #8162)
```
SomberNight, line 239: We should try to upstream whatever we practically can from our branch to upstream p4a. Or at least some stuff! Would be great if it was easier to rebase to newer p4a in the future.

Currently a rebase is clearly a Herculean effort.

SomberNight (approved):
Thanks a lot for your work on this. It is greatly appreciated.

Thanks.

ecdsa:
thanks!

PR #10486, opened 2026-02-19
afiore's android build fixes

SomberNight:
from https://gitea.com/AaronFiore/electrum/commits/branch/android-docker-build-fixes

### DIFF
--- a/contrib/android/Dockerfile
+++ b/contrib/android/Dockerfile
@@ -232,8 +232,8 @@ ENV VIRTUAL_ENV=/opt/venv
 RUN "/opt/cpython/install/bin/python3" -m venv ${VIRTUAL_ENV}
 ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
 
-COPY contrib/deterministic-build/requirements-build-base.txt /opt/deterministic-build/
-COPY contrib/deterministic-build/requirements-build-android.txt /opt/deterministic-build/
+COPY --chown=user contrib/deterministic-build/requirements-build-base.txt /opt/deterministic-build/
+COPY --chown=user contrib/deterministic-build/requirements-build-android.txt /opt/deterministic-build/
 RUN /opt/venv/bin/python3 -m pip install --no-build-isolation --no-dependencies \
     -r /opt/deterministic-build/requirements-build-base.txt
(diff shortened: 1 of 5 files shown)


contrib/android/Dockerfile
```diff
@@ -196,17 +196,16 @@ RUN apt -y update -qq \
...
-    PATH="${HOME_DIR}/.local/bin:${PATH}"
+ENV HOME_DIR=/home/user
+ENV WORK_DIR="${HOME_DIR}/wspace"
+ENV PATH="${HOME_DIR}/.local/bin:${PATH}"
 WORKDIR ${WORK_DIR}
-RUN chown --recursive ${UID} ${WORK_DIR} ${ANDROID_SDK_HOME}
-RUN chown ${UID} /opt
-USER ${UID}
+RUN chown -R user ${WORK_DIR} ${ANDROID_SDK_HOME} /opt
+USER user
```
SomberNight, line 209: most of this code-block is duplicated across many of our Dockerfiles so if the changes make sense here, they should be copy-pasted to the other files also, I suppose.

SomberNight: done in https://github.com/spesmilo/electrum/pull/10486/changes/369ecdd6622b9c002b6f652289515b2a876bd766

contrib/android/build.sh
```diff
@@ -86,6 +86,7 @@ docker run --rm \
     -v "$PROJECT_ROOT_OR_FRESHCLONE_ROOT":/home/user/wspace/electrum \
     -v "$PROJECT_ROOT_OR_FRESHCLONE_ROOT"/.buildozer/.gradle:/home/user/.gradle \
     $DOCKER_RUN_FLAGS \
+    --user user \
```
SomberNight: The Dockerfile already has a `USER user` directive.
What effect would this `docker run --user user` have then? Shouldn't it be a NO-OP?

SomberNight: I have dropped [that commit](https://gitea.com/AaronFiore/electrum/commit/3d7d7ac116d91fd2423d9dbf3e0d3a38e19ffddb) (https://github.com/spesmilo/electrum/commit/3d7d7ac116d91fd2423d9dbf3e0d3a38e19ffddb) for now.

PR #10490, opened 2026-02-20
locale: exclude llm proofreader directory from build

f321x:
Required for https://github.com/spesmilo/electrum-locale/pull/51
Deletes the `llm_proofreader/` directory in `electrum/locale` when building the locales with `build_cleanlocale.sh`.
Also ignores changes inside the locale submodule in the main repository so deleting or changing files in `electrum/locale` doesn't pollute the git status in the main repo.

### DIFF
--- a/contrib/locale/build_cleanlocale.sh
+++ b/contrib/locale/build_cleanlocale.sh
@@ -13,4 +13,5 @@ LOCALE="$PROJECT_ROOT/electrum/locale/"
 cd "$LOCALE"
 git clean -ffxd
 git reset --hard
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "electrum/locale"]
 	path = electrum/locale
-	url = https://github.com/spesmilo/electrum-locale
+    url = https://github.com/spesmilo/electrum-locale
+    ignore = dirty
```
SomberNight, line 4: oooh nice

PR #10494, opened 2026-02-23
qt: console: allow changing font size

f321x:
Allows changing the font size in the qt Console with `Ctrl` + `+` and `Ctrl` + `-`.

image

### DIFF
--- a/electrum/gui/qt/console.py
+++ b/electrum/gui/qt/console.py
@@ -1,4 +1,3 @@
-
 # source: http://stackoverflow.com/questions/2758159/how-to-embed-a-python-interpreter-in-a-pyqt-widget
 
 import sys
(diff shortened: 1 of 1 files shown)


electrum/gui/qt/console.py
```diff
@@ -308,6 +317,12 @@ def keyPressEvent(self, event):
         elif event.key() == Qt.Key.Key_C and event.modifiers() == Qt.KeyboardModifier.ControlModifier:
             if not self.textCursor().selectedText():
                 self.keyboard_interrupt()
+        elif event.key() == Qt.Key.Key_Plus and event.modifiers() == Qt.KeyboardModifier.ControlModifier:
```
SomberNight: I can't press "+" without shift on my keyboard.

```suggestion
        elif event.key() == Qt.Key.Key_Plus and Qt.KeyboardModifier.ControlModifier in event.modifiers():
```

f321x: Changed in https://github.com/spesmilo/electrum/compare/4da6b4c569a72be141ea5ae87f8e8bf4a51d550b..2df68d9249c0c9161616609d971d9a284a46c75f

PR #10500, opened 2026-02-27
build: win: upgrade docker base img to debian 13

SomberNight:
- bump docker base image (debian 12->13)
- bump wine (10->11)
- change from win-iconv to GNU libiconv
    - (zbar [requires](https://github.com/mchehab/zbar/blob/a549566ea11eb03622bd4458a1728ffe3f589163/README-windows.md) iconv)
    - [issues](https://github.com/SomberNight/electrum/commit/cb00cb60cf9a0cb059dcac5e7acfe5186620cabe) compiling [win-iconv](https://github.com/win-iconv/win-iconv):
        - pre-existing issue: lots of "-Wincompatible-pointer-types" warnings when compiling zbar with win-iconv
        - new debian means newer GCC
        - new GCC [changed that](https://gcc.gnu.org/pipermail/gcc-cvs/2023-December/394351.html) warning type to "error"
    - [GNU libiconv](https://www.gnu.org/software/libiconv/) works as an alternative
        - drawback: win-iconv is more minimal, it uses the win32 API to do most of the work
        - still, is a 25+ year old GNU project with one release every ~2 years, so IMO fine

---

notes:
- tested that built windows exes are still reproducible
- tested at runtime I can still scan QR code using windows exe
  - could not figure out how to statically link libiconv into libzbar, so runtime behaviour needs the proper DLLs copied (as done in current PR)

### DIFF
--- a/contrib/build-wine/Dockerfile
+++ b/contrib/build-wine/Dockerfile
@@ -22,6 +22,8 @@ RUN dpkg --add-architecture i386 && \
         git \
         p7zip-full \
         make \
+        cmake \
+        pkgconf \
         mingw-w64 \
         mingw-w64-tools \
         autotools-dev \
--- a/contrib/build-wine/make_win.sh
+++ b/contrib/build-wine/make_win.sh
@@ -53,28 +53,34 @@ if [ -f "$DLL_TARGET_DIR/libzbar-0.dll" ]; then
     info "libzbar already built, skipping"
 else
     (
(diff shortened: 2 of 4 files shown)


SomberNight:
> - change from win-iconv to GNU libiconv

Hmm, I just realised we already pull in GNU libiconv for the Android build (also for zbar). So technically it is not really even a "new" dependency.

https://github.com/spesmilo/electrum/blob/09a09057f6ff7614cfd9d757f82bf891a5f87fe6/contrib/android/p4a_recipes/libiconv/__init__.py#L14-L16

PR #10503, opened 2026-03-03
qt: utxo_list: only enable 'fully spend...' menu if there are unfrozen coins in the selection.

accumulator:
only enable 'fully spend...' menu if there are unfrozen coins in the selection.

otherwise, when selecting only frozen coins, the set of usable coins is empty, and the menu options will instead fall back to using ALL (unfrozen) coins without informing the user.


SomberNight:
Good find!  Thanks.

PR #10504, opened 2026-03-03
common_qt: move QtEventListener and qt_event_listener decorator to common_qt

accumulator:
deduplicate same implementation, enables sharing future code using `qt_event_listener`

### DIFF
--- a/electrum/gui/common_qt/util.py
+++ b/electrum/gui/common_qt/util.py
@@ -1,5 +1,6 @@
 import queue
 import sys
+from functools import wraps
 from typing import Optional, NamedTuple, Callable
 import os.path
 
@@ -10,6 +11,7 @@ import qrcode
 
 from electrum.i18n import _
 from electrum.logging import Logger
+from electrum.util import EventListener, event_listener
 
 _cached_font_ids: dict[str, int] = {}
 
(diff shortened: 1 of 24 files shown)


electrum/gui/common_qt/util.py
```diff
@@ -190,3 +192,33 @@ def stop(self):
...
+    def register_callbacks(self):
+        self.qt_callback_signal.connect(self.on_qt_callback_signal)
+        EventListener.register_callbacks(self)
+
+    def unregister_callbacks(self):
+        try:
+            self.qt_callback_signal.disconnect()
+        except (RuntimeError, TypeError):  # wrapped Qt object might be deleted
+            # "TypeError: disconnect() failed between 'qt_callback_signal' and all its connections"
+            pass
```
SomberNight, line 209: note: so this was already diverged between the two guis -- qml never disconnected the signals?

SomberNight (approved):
Looks good. Thanks!

electrum/gui/common_qt/util.py
accumulator: disconnecting the signal was left commented out for qml, in anticipation of having to uncomment it, but it was never needed (probably due to signals getting properly disconnected on destruction of the Q objects)

PR #10505, opened 2026-03-03
plugins: nwc: sync with spec, fix some bugs

f321x:
Syncs the NWC plugin with changes to the NIP-47 specification and makes the handling of payment budgets more reliable.

### DIFF
--- a/electrum/plugins/nwc/manifest.json
+++ b/electrum/plugins/nwc/manifest.json
@@ -5,5 +5,5 @@
   "author": "The Electrum Developers",
   "available_for": ["cmdline", "qt"],
   "icon":"nwc.png",
-  "version": "0.0.1"
+  "version": "0.0.2"
 }


electrum/plugins/nwc/nwcserver.py
```diff
@@ -823,30 +825,43 @@ async def pay_invoice(self, b11: str, amount_msat: Optional[int], request_pub: s
         except Exception as e:
             self.logger.exception(f"failed to pay nwc invoice")
             return self.get_error_response("PAYMENT_FAILED", str(e))
+        finally:
+            if not success:
+                # if the user goes offline before the payment succeeds this will incorrectly remove the payment from the budget
+                # though for this use case it seems reasonably accurate
+                self.remove_from_budget(request_pub, budget_item)
```
SomberNight, line 837: by the user going offline, you mean e.g. gracefully exiting the python process or the `close_wallet` hook running? (so that e.g. CancelledError gets raised)

Looks to me we could easily decide either way: whether to add such attempts to the budget or not. You could initialise `success = None` instead of False, and then treat `success == None` specially in the finally.
I think the conservative choice would be to treat None as success. but I agree either choice is reasonable

electrum/plugins/nwc/qt.py
```diff
@@ -117,7 +117,8 @@ def update_connections_list():
                 else:
                     budget = self.config.format_amount(conn['daily_limit_sat'])
                     used = self.config.format_amount(
-                        self.nwc_server.get_used_budget(conn['client_pub']))
+                        amount_sat=round(self.nwc_server.get_used_budget_msat(conn['client_pub']) / 1000)
```
SomberNight: just for my sanity, could we avoid floats here?
I mean, it's just for display purposes, but still. Don't want to think about how floats behave.
```suggestion
                        amount_sat=round(Decimal(self.nwc_server.get_used_budget_msat(conn['client_pub'])) / 1000)
```

electrum/plugins/nwc/nwcserver.py
SomberNight: note that if the process gets killed non-gracefully or e.g. the computer halts due to power cut, the pending payment budgeting will not be undone.
So perhaps also for consistency between these two scenarios (graceful vs non-graceful exit), we could treat the `success is None` case same as `sucess is True` - i.e. only undo budgeting for explicit payment failures.

electrum/plugins/nwc/qt.py
f321x: Sure, added `Decimal` in https://github.com/spesmilo/electrum/compare/731e2752c26fcf765378045aae61dbacd0144d85..835ab39d77c7378f6d929a9f8fa70fc7c18bd5ae

electrum/plugins/nwc/nwcserver.py
f321x: > by the user going offline, you mean e.g. gracefully exiting the python process or the close_wallet hook running? (so that e.g. CancelledError gets raised)

yes that is what i meant. I agree, not removing it from the budget seems safer, lets do that instead: https://github.com/spesmilo/electrum/compare/731e2752c26fcf765378045aae61dbacd0144d85..835ab39d77c7378f6d929a9f8fa70fc7c18bd5ae

electrum/plugins/nwc/nwcserver.py
```diff
@@ -832,10 +816,12 @@ async def pay_invoice(self, b11: str, amount_msat: Optional[int], request_pub: s
         elif invoice.get_amount_msat() is None:
             invoice.set_amount_msat(amount_msat)
 
-        if not self.budget_allows_spend(request_pub, invoice.get_amount_sat()):
+        if not self.budget_allows_spend(request_pub, msat_requested=amount_msat or invoice.get_amount_msat()):
             return self.get_error_response("QUOTA_EXCEEDED", "Payment exceeds daily limit")
+        budget_item = self.add_to_budget(request_pub, amount_msat=amount_msat or invoice.get_amount_msat())
```
SomberNight, line 821: note: I think it would be a safer API if `budget_allows_spend` and `add_to_budget` were combined to a single method. could be done in a follow-up if you agree

SomberNight: but as long as there is no `await` between the two calls, and no multi-threading here, it does not functionally matter

f321x: Doing this now in a followup: https://github.com/spesmilo/electrum/pull/10511
Conceptually we also shouldn't context switch in the finally block (in the followup PR), but i think that should be clear.

PR #10506, opened 2026-03-03
qt: perform 'fully spend' action with coin selection, keep separate from coin control when doing action

accumulator:
perform 'fully spend' action with coin selection, keep separate from coin control when doing action.

Also stop timer when dialog is finished, to avoid re-generating txs with the same input coin set, which results in an exception as these coins have signatures when the swap has started.


ecdsa:
looks good, thanks!

PR #10507, opened 2026-03-04
qt, qml: for new transaction notifications, instead of using sign, explicitly say sent/received.

accumulator:
for new transaction notifications, instead of using sign, explicitly say sent/received.
For multiple transactions, split summary in total sent/received and a balance change.

### DIFF
--- a/electrum/gui/qml/qewallet.py
+++ b/electrum/gui/qml/qewallet.py
@@ -198,7 +198,7 @@ class QEWallet(AuthMixin, QObject, QtEventListener):
             self.invoiceStatusChanged.emit(key, status)
 
     @qt_event_listener
-    def on_event_new_transaction(self, wallet: Abstract_Wallet, tx: Transaction):
+    def on_event_new_transaction(self, wallet: 'Abstract_Wallet', tx: Transaction):
         if wallet == self.wallet:
             self._logger.info(f'new transaction {tx.txid()}')
             self.add_tx_notification(tx)


ecdsa:
that's a lot of code replication.
could you move the part that creates the message to common_qt ?

SomberNight:
>  could you move the part that creates the message to common_qt ?

(that part is not really specific to qt at first glance -- if moving it anyway, it could even go to wallet.py)

accumulator:
moved to `Abstract_Wallet.get_user_notifications_for_new_txns()`

electrum/gui/qt/main_window.py
```diff
@@ -930,22 +930,9 @@ def notify_transactions(self):
...
-                    continue
-                total_amount += tx_wallet_delta.delta
-            self.notify(_("{} new transactions: Total amount received in the new transactions {}")
-                        .format(len(txns), self.format_amount_and_units(total_amount)))
-        else:
-            for tx in txns:
-                tx_wallet_delta = self.wallet.get_wallet_delta(tx)
-                if not tx_wallet_delta.is_relevant:
-                    continue
-                self.notify(_("New transaction: {}").format(self.format_amount_and_units(tx_wallet_delta.delta)))
```
SomberNight, line 948: just a note: in this GUI, we were calling `window.format_amount_and_units` instead of `config.format_amount_and_units`. The former also adds fiat prices. So now we don't do that.

Anyway, now both GUIs behave the same.

PR #10509, opened 2026-03-04
lnpeer: simplify where maybe_send_commitment() is called

SomberNight:
The typical flow of an update is:
```
---UPDATE--->
--- SIG  --->
<--REVACK----
<-- SIG  ----
---REVACK--->
```

It makes sense to try to send a sig ("commitment_signed") right after we send an update. It also makes sense right after we send revack.

Besides those times, we could call "maybe_send_commitment" at *any* time, that is safe, and depending on other call locations, it might be an optimisation, however it is not needed.

In particular it is unclear why we had those calls when we *receive* updates (and only for certain types of updates - not consistently).

-----

The second commit implements a very simple batching of updates: we delay sending a commitsig if we already sent one recently.


ecdsa:
LGTM

PR #10511, opened 2026-03-05
plugin: nwc: followup #10505, include fees in budget

f321x:
* Includes payment fees in budget accounting
* Adds timer to the Qt dialog so it refreshes balances after spends
* fixes fee/amount confusion bug in `on_event_payment_succeeded`

### DIFF
--- a/electrum/plugins/nwc/nwcserver.py
+++ b/electrum/plugins/nwc/nwcserver.py
@@ -523,7 +523,7 @@ class NWCServer(Logger, EventListener):
             b11 = invoice.lightning_invoice
         elif self.wallet.get_request(invoice.rhash):
             direction = "incoming"
-            info = self.wallet.lnworker.get_payment_info(invoice.payment_hash, direction=RECEIVED)
+            info = self.wallet.lnworker.get_payment_info(bytes.fromhex(invoice.rhash), direction=RECEIVED)
             _, b11 = self.wallet.lnworker.get_bolt11_invoice(
                 payment_info=info,
                 message=invoice.message,
@@ -539,11 +539,10 @@ class NWCServer(Logger, EventListener):
                 "created_at": invoice.time,
                 "expires_at": invoice.get_expiration_date(),
                 "fees_paid": 0,
+                "invoice": b11,
                 "metadata": {}
             }
         }
-        if payment_hash:  # if client requested by payment hash we add the invoice
-            response['result']['invoice'] = b11
         if nip47_status := self.invoice_status_to_nip47_state(status):
             response['result']['state'] = nip47_status
 


electrum/plugins/nwc/nwcserver.py
```diff
@@ -816,9 +816,18 @@ async def pay_invoice(self, b11: str, amount_msat: Optional[int], request_pub: s
...
-        budget_item = self.add_to_budget(request_pub, msat_requested=amount_msat or invoice.get_amount_msat())
-        if not budget_item:
-            return self.get_error_response("QUOTA_EXCEEDED", "Payment exceeds daily limit")
+        # add the maximum allowed fee to the budget and update it with the actual fee once the payment succeeds
+        # to prevent exceeding the budget through high fees
+        payment_amount_msat = amount_msat or invoice.get_amount_msat()
+        fee_budget = PaymentFeeBudget.from_invoice_amount(
+            config=self.wallet.config,
+            invoice_amount_msat=payment_amount_msat,
+        )
```
SomberNight, line 824: we should pass this budget to the `lnworker.pay_invoice` call by setting `budget=fee_budget`.

(even though you calculate `fee_budget` using the same default logic `lnworker.pay_invoice` uses, it is better to pass it explicitly to make the code more robust against the `lnworker.pay_invoice` default logic changing)

electrum/plugins/nwc/nwcserver.py
```diff
@@ -831,10 +840,16 @@ async def pay_invoice(self, b11: str, amount_msat: Optional[int], request_pub: s
...
-                self.remove_from_budget(request_pub, budget_item)
+            # note: if the user shuts down or the application crashes before the payment ends, it will not
+            # get deducted from the budget, even if the htlcs later get failed on wallet restart.
+            self.remove_from_budget(request_pub, budget_item)
+            if success:
+                # re-add the spend to the budget, but this time with the actual fees that have been paid
+                info = self.get_payment_info(invoice.rhash)
+                assert info, "info should exist after successful payment"
+                _, total_amount_msat, _, _ = info
+                self.add_to_budget(request_pub, msat_requested=total_amount_msat)
```
SomberNight: Note that `lnworker.pay_invoice` returns as soon as the first htlc is fulfilled, others (as part of an MPP) might still be in-flight.

Also note this comment:
https://github.com/spesmilo/electrum/blob/851781fba5e006972718f18bff3a5d3572e890ab/electrum/lnworker.py#L1294-L1299

I think this behaviour "infects" `NWCServer.get_payment_info`, making it behave unexpectedly.

As a consequence, the code here might end up replacing the budget item with a much smaller amount (which is actually attacker controlled -- they can decide when to fulfill which part of the MPP).

----

There are a lot of implicit assumptions here. For example, would be nice to have an
`assert total_amount_msat >= payment_amount_msat` --- which would I believe fail in the case I described.  However it would not help security: the old budget item has already been removed.

Hence this "remove, then run potentially buggy code, then re-add", strategy is inherently error-prone :/

-----

How about something like this instead? (but this does not fix the `NWCServer.get_payment_info` bug)

```diff
diff --git a/electrum/plugins/nwc/nwcserver.py b/electrum/plugins/nwc/nwcserver.py
index 38b96ecc3..11eeb3316 100644
--- a/electrum/plugins/nwc/nwcserver.py
+++ b/electrum/plugins/nwc/nwcserver.py
@@ -842,13 +842,15 @@ class NWCServer(Logger, EventListener):
         finally:
             # note: if the user shuts down or the application crashes before the payment ends, it will not
             # get deducted from the budget, even if the htlcs later get failed on wallet restart.
-            self.remove_from_budget(request_pub, budget_item)
             if success:
-                # re-add the spend to the budget, but this time with the actual fees that have been paid
+                # replace the spend in the budget, this time using the actual (lower) fees that have been paid
                 info = self.get_payment_info(invoice.rhash)
                 assert info, "info should exist after successful payment"
                 _, total_amount_msat, _, _ = info
-                self.add_to_budget(request_pub, msat_requested=total_amount_msat)
+                assert payment_amount_msat <= total_amount_msat <= budget_amount_msat
+                self._update_budget_item_amount(request_pub, old_budget_item=budget_item, new_msat=total_amount_msat)
+            else:
+                self.remove_from_budget(request_pub, budget_item)
 
         preimage: bytes = self.wallet.lnworker.get_preimage(bytes.fromhex(invoice.rhash))
         response = {}
@@ -873,6 +875,18 @@ class NWCServer(Logger, EventListener):
             self.logger.debug(f"failed to remove {budget_item=} from {client_pub[:4]}...{client_pub[-4:]} budget: not found")
             pass
 
+    def _update_budget_item_amount(self, client_pub: str, *, old_budget_item: list[int], new_msat: int) -> None:
+        assert len(old_budget_item) == 2, old_budget_item
+        budget_spends = self.connections[client_pub].get('budget_spends', [])
+        try:
+            budget_spends.remove(old_budget_item)
+        except ValueError:
+            self.logger.debug(f"failed to rm and update {old_budget_item=} from {client_pub[:4]}...{client_pub[-4:]} budget: not found")
+            return
+        # tuples don't work because jsondb converts them to lists on reload
+        new_budget_item = [new_msat, int(time.time())]
+        budget_spends.append(new_budget_item)
+
     def get_used_budget_msat(self, client_pub: str) -> int:
         """
         Returns the used budget for the given client_pubkey in millisatoshi.

```

electrum/plugins/nwc/nwcserver.py
f321x: Done in 8ccdc3f7afa2fbd32c65ce1daa6eef22cc11c747

electrum/plugins/nwc/nwcserver.py
f321x: > Note that lnworker.pay_invoice returns as soon as the first htlc is fulfilled, others (as part of an MPP) might still be in-flight.

Ah okay, i didn't consider this but it makes sense. 
I added your proposed changes. Additionally I also changed `NWCServer.get_payment_info()` in https://github.com/spesmilo/electrum/pull/10511/commits/f53203d171477abd6ee4574bf631f0c2888b95f0 to include inflight htlcs in the calculation as long as any htlc of the payment has been settled. 
This should fix the issue as we can assume that inflight htlcs will get settled as well.

PR #10514, opened 2026-03-09
transaction: re-raise NetworkException in add_info_from_network

f321x:
`Transaction.add_info_from_network()` would swallow a `NetworkException` even when `ignore_network_issues=False` is passed. 
This causes `Transaction.add_info_from_wallet_and_network()` to incorrectly return `True` even if the operation failed. If `QETxRbfFeeBumper` then proceeds assuming the call was successful `Abstract_Wallet.bump_fee()` would raise an `Exception("tx missing info from network")`.

See https://github.com/spesmilo/electrum/issues/5502#issuecomment-40213084270

### DIFF
--- a/electrum/gui/qml/qetxfinalizer.py
+++ b/electrum/gui/qml/qetxfinalizer.py
@@ -659,7 +659,7 @@ class QETxRbfFeeBumper(TxFeeSlider, TxMonMixin):
         super().__init__(parent)
 
         self._oldfee = QEAmount()
-        self._oldfee_rate = 0
+        self._oldfee_rate = '0'
         self._orig_tx = None
         self._rbf = True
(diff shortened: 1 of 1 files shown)


electrum/gui/qml/qetxfinalizer.py
```diff
@@ -758,6 +758,13 @@ def update(self):
             self.validChanged.emit()
             self.warning = _("The new fee rate needs to be higher than the old fee rate.")
             return
+
+        if not self._orig_tx.add_info_from_wallet_and_network(wallet=self._wallet.wallet, show_error=self._logger.error):
```
SomberNight, line 762: This will retry the blocking network request again, again on the GUI thread. :/
well, it is kind of a pre-existing design issue here..

PR #10515, opened 2026-03-10
Update my Electrum server details in servers.json

jhoenicke:
I moved my node to standard port again because the address with port 50006 is apparently hardcoded in some malware as C&C server.  I also updated the domain name (both work at the moment, but I prefer the shorter one).


SomberNight:
Thank you for running a server :)

Issue #7905, opened 2022-07-27
Add UR2.0 QR codes


Jakob-1870:
UR 2.0 is becoming the standard for QR PSBT's. It's implemented in almost all software and hardware wallets that support QR PSBT's.
https://github.com/BlockchainCommons/Research

Electrum should add support for it.

SomberNight:
> https://github.com/BlockchainCommons/Research

There's a lot of stuff there... Which "spec" in particular are you referring to? Is it this one?
https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md

> UR 2.0 is becoming the standard for QR PSBT's. It's implemented in almost all software and hardware wallets that support QR PSBT's.

Which other software/hardware/services support it?

Jakob-1870:
Yes that's it.

Bluewallet, Sparrow, Specter, Keystone, Passport, Gordian, Krux, Seedsigner all use UR2.0.

stepansnigirev:
Two python implementation that might be interesting for you:
From Foundation Devices: https://github.com/Foundation-Devices/foundation-ur-py
Implementation in Specter-DIY: https://github.com/diybitcoinhardware/f469-disco/tree/master/libs/common/microur

massmux:
Sorry for duplicating the request i did not get aware of this one. But yes this would be important i think for all the community

SomberNight:
Some previous discussion re the custom format we have been using since 2015: https://github.com/spesmilo/electrum/issues/5746
In short, when implementing PSBT support in 2019 we decided to keep using the base43 encoding inside alphanumeric-mode QR codes, instead of using either base64 text inside binary-mode QR codes or raw bytes inside binary-mode QR codes, due to maximising payload size and easy-copy-pasteability.

---

Anyway, if this BCR-2020-005 format has traction, we could have it too. PR welcome. It probably needs some GUI-specific work, but getting it to work with the Qt GUI would be a good starting point.

ecdsa:
I think it would be a pity if we were forced to adopt an animated qr code standard just because some spec is feature bloated.
Has someone taken the time to evaluate that?

massmux:
actually sparrow and specter desktop use such a format. maybe it would be important electrum's users also can share same standard. imho

Medgeina:
[...]

PR #10518, opened 2026-03-12
Add UR decoding capability to scanners on desktop & Android

hectorchu:
As title. Only decoding - no encoding was added.

### DIFF
--- a/electrum/gui/qt/qrreader/qtmultimedia/camera_dialog.py
+++ b/electrum/gui/qt/qrreader/qtmultimedia/camera_dialog.py
@@ -38,6 +38,7 @@ from electrum.simple_config import SimpleConfig
 from electrum.i18n import _
 from electrum.qrreader import get_qr_reader, QrCodeResult, MissingQrDetectionLib
 from electrum.logging import Logger
+from electrum.ur.ur_decoder import URDecoder
 
 from electrum.gui.qt.util import MessageBoxMixin, FixedAspectRatioLayout, ImageGraphicsEffect
 
@@ -91,6 +92,7 @@ class QrReaderCameraDialog(Logger, MessageBoxMixin, QDialog):
         self._ok_done: bool = False
         self.camera_sc_conn = None
         self.resolution: QSize = None
+        self.ur_decoder = URDecoder()
 
         self.config = config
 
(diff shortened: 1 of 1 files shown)


SomberNight:
(ref https://github.com/spesmilo/electrum/issues/7905)

electrum/gui/qt/qrreader/qtmultimedia/camera_dialog.py
SomberNight, line 1: Note: the qtmultimedia toolchain is only used on Windows and macOS. Desktop Linux uses zbar.
But perhaps this is something we should reconsider.
See comment here:
https://github.com/spesmilo/electrum/blob/169227e148b1fe542e8832cb5c6ea86f887dc200/electrum/gui/qt/qrreader/__init__.py#L5-L20

electrum/ur/ur_encoder.py
SomberNight, line 1: Thanks a lot for this. It looks like a solid start.
Would you be interested in adding also the encode/display qr code parts to the GUIs?

hectorchu: I have the encoding change for QRCodeWidget, but it currently generates the parts upfront, and then these are passed to the QRCodeWidget which then cycles them. But this was only for compatibility with Cake Wallet, which makes the assumption there are n-fixed parts. UR codes are actually intended to be used as a fountain code. See the description I wrote up [here](https://github.com/cake-tech/cake_wallet/pull/3048).

So I would take this [change](https://github.com/ltc-electrum/electrum-ltc/commit/4d4c638e19b4ec4500a419550a53b9b446ea447e#diff-84f244a9c21b842eaf7077c620cba78515b48e39fc147a04d42524fc2fddf9b3), but instead the QRCodeWidget should accept the data string and an argument encode_ur, if the latter is True then use a UREncoder to generate a new part on every refresh. These changes aren't difficult and I doubt they are difficult for the Android one also, but I haven't looked at the Android code. Could you do it?

hectorchu: Actually, the QRCodeWidget should just accept a UR|str, because a UR also includes a type, and it would be better for all that to be created by the caller. 
The widget creates encoder = UREncoder(ur), then the widget would just call encoder.next_part() per refresh.

pony-montana:
HI! Here just to signal two python projects that implements these feature, hoping it could be useful :)

[SeedQReader](https://github.com/tadeubas/SeedQReader)
[bitcoin-qr-tools](https://github.com/andreasgriffin/bitcoin-qr-tools)

PR #10524, opened 2026-03-13
timelock_recovery plugin: Fix checksum calculation following BIP-128 standard

oren-z0:
Non-Ascii characters should not be converted for
checksum calculation.
This will give consistent hash to BIP-128 and its
Javascript code example.

Timelock-Recovery Plans that contained only ascii
characters are not affected.

### DIFF
--- a/tests/test_timelock_recovery.py
+++ b/tests/test_timelock_recovery.py
@@ -9,6 +9,7 @@ from electrum.storage import WalletStorage
 from electrum.transaction import PartialTxOutput
 from electrum.wallet import Wallet
 from electrum.wallet_db import WalletDB
+from electrum.plugins.timelock_recovery.qt import Plugin as TimelockRecoveryQtPlugin
 
 from . import ElectrumTestCase
 
@@ -119,3 +120,10 @@ class TestTimelockRecovery(ElectrumTestCase):
         self.assertEqual(cancellation_tx_outputs, [
             ('tb1q6k5h4cz6ra8nzhg90xm9wldvadgh0fpttfthcg', 737065),
         ])
+
+    def test_checksum_non_ascii(self):
+        # Non-ASCII characters must be serialized as-is (ensure_ascii=False),
+        # not escaped as \uXXXX sequences, before hashing.
+        json_data = {"wallet_name": "Ωmega Wörld Ñoño 日本語 中文 עברית العربية", "id": "abc-123"}
+        result = TimelockRecoveryQtPlugin._checksum(json_data)
+        self.assertEqual(result, "74674eca")


electrum/plugins/payserver/www
SomberNight, line 1: this submodule was deleted.
maybe try manually deleting the folder and then running `$ git submodule update --init --force` or something like that

oren-z0: Huh, I didn't notice this entered the commit. Let me fix that.

SomberNight:
Which field could contain non-ascii characters? I guess `"wallet_name"`?

>  This will give consistent hash to BIP-128 and its Javascript code example.

Are you saying the current python code does not produce the same hash as the example at the end of the BIP?
Why is that? Where is the non-ASCII character in that example?

A unit test would be nice.

oren-z0:
Yes, "wallet_name" could contain non-ascii character.
The BIP also allows an optional field for a description and optional labels for the destinations, which are not implemented in Electrum, but could be implemented in the future (and contain non-ascii characters).

Yes, if using `ensure_ascii=True` does give a different result than Javascript's `JSON.stringify()`.
With `ensure_ascii=False` the results are the same (tested using Python 3.13 and Node.JS 24.1).

SomberNight:
> Yes, if using `ensure_ascii=True` does give a different result than Javascript's `JSON.stringify()`.
With `ensure_ascii=False` the results are the same (tested using Python 3.13 and Node.JS 24.1).

The fact that this could happen, even without non-ascii characters being present, makes adding some kind of test even more relevant. Please add a test. Otherwise how would we notice if this ever regressed?

oren-z0:
@SomberNight I've added a test.
I also sliced the checksum to the first 8 hex-digits - which is enough for this kind of checksum (and also follows BIP-128).

SomberNight:
>> Are you saying the current python code does not produce the same hash as the example at the end of the BIP?
>>Why is that? Where is the non-ASCII character in that example?
>
> Yes, if using ensure_ascii=True does give a different result than Javascript's JSON.stringify().
> With ensure_ascii=False the results are the same (tested using Python 3.13 and Node.JS 24.1).

Ok, so to be absolutely clear, I tested now because I was still confused, and the checksum we calculate for the example in the BIP matches what is in the BIP, regardless of this change. I feel we miscommunicated. I was really confused why this change would affect that example as I could not see where the non-ascii character is - which is why I specifically asked that.

SomberNight:
merged in https://github.com/spesmilo/electrum/commit/e4ad44c0c35863e4f476de6f8ef39da76b27754b, with follow-ups

PR #10532, opened 2026-03-19
trampoline: allow trampoline onion packets of arbitrary size

ecdsa:
This commit was cherry-picked from the lazy_trampoline branch.
Apparently, variable size trampoline onions are needed for bolt12.

### DIFF
--- a/electrum/trampoline.py
+++ b/electrum/trampoline.py
@@ -7,7 +7,7 @@ from types import MappingProxyType
 
 from .lnutil import LnFeatures, PaymentFeeBudget, FeeBudgetExceeded
 from .lnonion import (
-    calc_hops_data_for_payment, new_onion_packet, OnionPacket, TRAMPOLINE_HOPS_DATA_SIZE, PER_HOP_HMAC_SIZE
+    calc_hops_data_for_payment, new_onion_packet, OnionPacket, PER_HOP_HMAC_SIZE
 )
 from .lnrouter import TrampolineEdge, is_route_within_budget, LNPaymentTRoute
 from .lnutil import NoPathFound
@@ -39,6 +39,8 @@ TRAMPOLINE_NODES_SIGNET = {
 }
 
 _TRAMPOLINE_NODES_UNITTESTS = {}  # used in unit tests
+
+TRAMPOLINE_HOPS_MAX_DATA_SIZE = 500
 
 
 def hardcoded_trampoline_nodes() -> Mapping[str, LNPeerAddr]:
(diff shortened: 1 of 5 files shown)


ecdsa:
Note: I think it would be possible to remove `TRAMPOLINE_HOPS_MAX_DATA_SIZE`, if we first estimated the available remaining space in the outer onion, and passed it to the method building the inner onion.

f321x:
Note: older versions of lightning-kmp (Phoenix) seem to support only fixed size 400 byte trampoline onions, however this is not critical for us as we don't do end-to-end trampoline payments to Phoenix anyway:
https://github.com/ACINQ/lightning-kmp/blob/b67d8406459e0f1adc0703bcde16d8e9ecf3b106/modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentPacket.kt#L51

electrum/trampoline.py
```diff
@@ -332,7 +334,7 @@ def create_trampoline_onion(
         payload = dict(hops_data[index].payload)
         # try different r_tag order on each attempt
         invoice_routing_info = random_shuffled_copy(route[index].invoice_routing_info)
-        remaining_payload_space = TRAMPOLINE_HOPS_DATA_SIZE \
+        remaining_payload_space = TRAMPOLINE_HOPS_MAX_DATA_SIZE \
```
f321x, line 337: Note: Phoenix seems to use a similar approach for limiting the trampoline onion payload size when including r_tags for legacy payments, however they limit to 1000 byte instead of 500.
https://github.com/ACINQ/lightning-kmp/blob/b67d8406459e0f1adc0703bcde16d8e9ecf3b106/modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentPacket.kt#L100-L105

ecdsa: so we have some margin. I'm fine with increasing our limit. estimating the size of the onion beforehand seems complicated, not sure it is worth the effort.

SomberNight:
> older versions of lightning-kmp (Phoenix) seem to support only fixed size 400 byte trampoline onions, however this is not critical for us as we don't do end-to-end trampoline payments to Phoenix anyway

I think we do try to use end-to-end trampoline when paying phoenix (?) but ok probably not worth it to support *older* phoenix
https://github.com/spesmilo/electrum/blob/3cf2c325d55c698c94cf8e202aafead054415a38/electrum/trampoline.py#L105-L111

PR #10533, opened 2026-03-19
adb.get_spender: subscribe to outputs explicitly, instead of as a sid…

ecdsa:
…e effect

This allows to subscribe only if we need it (for channel closing tx and htlcs)


f321x:
Works for my case where an anchor got swept to [bc1pws6pvj75rcsc2eglpp9k570prnjh40nfpyahlyumk8y8smjayvasyhns5c](https://mempool.space/address/bc1pws6pvj75rcsc2eglpp9k570prnjh40nfpyahlyumk8y8smjayvasyhns5c), causing excessive requests

PR #10534, opened 2026-03-20
daemon: forbid "setconfig" command to change rpcserver settings in-flight

SomberNight:
It is much easier to reason about the rpcserver if we don't allow changing its basic settings while it is already running. What does it mean to change the TCP port it is listening on ("rpcport") if it's already running? It is even problematic to change the rpcpassword: care needs to be taken to already update it for the current server.
(ref spesmilo#6762)

This commit disallows changing all of the "rpc*" config variables if the daemon is already running.

---

Simultaneously, it also ensures rpc_password is always set and auth cannot be disabled.

Previously if there was a daemon running, and the user ran
`$ electrum setconfig rpcpassword ""` that would leave the RPC unauthenticated
for the current session. However next time the daemon restarted, get_rpc_credentials would see
the unset password and generate one.

I think this was the worst of both worlds:
- we did not really allow removing the rpc password, except for the current session, and
- perhaps unexpectedly, we would generate a new password on daemon restart

Instead now we explicitly make sure the RPC server can never get into a state where it does not have a password set.

Based on a report by `Zuzana Kotásková <36777@mail.vsfs.cz>`

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -397,10 +397,19 @@ class Commands(Logger):
 
     def _setconfig(self, key, value):
         value = self._setconfig_normalize_value(key, value)
-        if self.daemon and key == SimpleConfig.RPC_USERNAME.key():
-            self.daemon.commands_server.rpc_user = value
-        if self.daemon and key == SimpleConfig.RPC_PASSWORD.key():
-            self.daemon.commands_server.rpc_password = value
+        if self.daemon and key in (
+            SimpleConfig.RPC_USERNAME.key(),
+            SimpleConfig.RPC_PASSWORD.key(),
+            SimpleConfig.RPC_HOST.key(),
+            SimpleConfig.RPC_PORT.key(),
+            SimpleConfig.RPC_SOCKET_TYPE.key(),
+            SimpleConfig.RPC_SOCKET_FILEPATH.key(),
+        ):
+            raise UserFacingException(
+                "error: RPC server settings cannot be changed for already running daemon. "
+                "Stop the daemon first, and run 'setconfig' in --offline mode. "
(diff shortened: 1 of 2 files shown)


electrum/commands.py
```diff
@@ -397,9 +397,16 @@ def _setconfig_normalize_value(cls, key, value):
...
             self.daemon.commands_server.rpc_user = value
-        if self.daemon and key == SimpleConfig.RPC_PASSWORD.key():
+        if self.daemon and key == SimpleConfig.RPC_PASSWORD.key():  # update RPC creds for currently running daemon
+            if not value:
+                # We always want the daemon RPC to be password-protected.
+                # (user can still set a weak password, but not an empty one)
+                # When the daemon starts, finding an empty password, will result in generating a new strong pw.
+                # So it is useful to allow `electrum -o setconfig rpcpassword ""` - to force regenerating
+                # a random password the next time the daemon starts.
+                raise UserFacingException("error: cannot unset RPC password when daemon is running")
```
SomberNight: Alternatively, we could forbid changing `rpcuser`/`rpcpassword`/etc while there is a daemon running. Only allow changing them with `--offline`.
This seems even easier to reason about.

Though note some use cases might get broken by this:
- e.g. currently it is possible to run the Qt GUI and change the rpcpassword of the running RPC server from the Console tab -- that would no longer work

ecdsa: I believe that solution is cleaner.
The console use case does not seem to be realistic; I was mentioning this as a theoretical situation.
In any case, the console does come with a warning

SomberNight: changed now to this approach in https://github.com/spesmilo/electrum/pull/10534/changes/0dcef9780b59805c980247c98d2fe12c6554ce93

PR #10538, opened 2026-03-23
 wizard: catch NotLegacySinglesigScriptType during private key import

f321x:
Catch NetLegacySinglesigScriptType and convert it to a
UserFacingException if the user tries to import a private key for which
it is not possible to get a singlesig descriptor (e.g. p2wsh).

Fixes https://github.com/spesmilo/electrum/issues/10536

### DIFF
--- a/electrum/gui/qt/__init__.py
+++ b/electrum/gui/qt/__init__.py
@@ -413,11 +413,10 @@ class ElectrumGui(BaseElectrumGui, Logger):
             self.logger.exception('')
             if isinstance(e, UserFacingException) \
                     or isinstance(e, WalletFileException) and not e.should_report_crash:
-                err_text = str(e) if isinstance(e, WalletFileException) else repr(e)
                 custom_message_box(icon=QMessageBox.Icon.Warning,
                                    parent=None,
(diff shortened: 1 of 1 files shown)


electrum/wizard.py
```diff
@@ -702,7 +702,12 @@ def create_storage(self, path: str, data: dict):
                 for pk in keys:
                     assert bitcoin.is_private_key(pk)
                     txin_type, pubkey = k.import_privkey(pk, None)
-                    addr = bitcoin.pubkey_to_address(txin_type, pubkey)
+                    try:
+                        addr = bitcoin.pubkey_to_address(txin_type, pubkey)
+                    except descriptor.NotLegacySinglesigScriptType as e:
+                        raise UserFacingException(
+                            _("Importing individual private keys of type '{}' is not supported.").format(txin_type),
+                        ) from e
```
SomberNight, line 710: note: arguably we could detect this much earlier in the wizard: on the same screen where the user enters the private keys, and not let them continue.  but anyway, this is already an improvement.

SomberNight:
Thanks!

PR #10541, opened 2026-03-24
lnwallet: also signal trampoline support in invoice features if trampoline is disabled

f321x:
Changes the behavior of LNWallet to also set the trampoline feature bit in bolt 11 invoice features if all the open channels are to trampoline peers. 
(Note: we could even signal trampoline support if we have at least one trampoline channel that can receive the requested amount, but this is already an enhancement and i wanted to keep it simple).

If the trampoline feature bit is set, it only includes channels with trampoline peers into the r_tags.

This allows a wallet with disabled trampoline payments to still receive end-to-end trampoline payments in the (likely) case where they only have channels open to trampoline nodes.

### DIFF
--- a/tests/test_lnwallet.py
+++ b/tests/test_lnwallet.py
@@ -1,9 +1,12 @@
 import logging
 import os
 
+import electrum.trampoline
 from . import ElectrumTestCase
+from .test_lnchannel import create_test_channels
 
-from electrum.lnutil import RECEIVED, MIN_FINAL_CLTV_DELTA_ACCEPTED
+from electrum.lnutil import RECEIVED, MIN_FINAL_CLTV_DELTA_ACCEPTED, LnFeatures
+from electrum.lntransport import LNPeerAddr
 from electrum.logging import console_stderr_handler
 from electrum.invoices import LN_EXPIRY_NEVER, PR_UNPAID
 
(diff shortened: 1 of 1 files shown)


ecdsa:
> If the trampoline feature bit is set, it only includes channels with trampoline peers into the r_tags.

Note that this is currently done by `channel.is_frozen_for_receiving`. 

However, I think the decision to enable trampoline support in the invoice should not depend on whether we enabled trampoline routing. It should be enabled if and only if all non-frozen channels are with trampoline nodes. 

edit: I mean non-frozen manually by the user. so we should stop automatically freezing channels for receiving if trampoline routing is enabled.

f321x:
> Note that this is currently done by channel.is_frozen_for_receiving.

Because the PaymentInfo containing the invoice features is created independently of the invoice and persisted, while the invoice is not persisted (only cached) it could happen that a user creates invoice features with trampoline support, then creates a new invoice later when trampoline was disabled and would include non-trampoline channels in the rtags. Now with https://github.com/spesmilo/electrum/pull/10541/commits/1c74678ff32c6665b260ee66ad0783df50a4caae they will get filtered out and we cannot include invalid r_tags.

> so we should stop automatically freezing channels for receiving if trampoline routing is enabled.

ok, doing this now in https://github.com/spesmilo/electrum/pull/10541/commits/41672a1f04fac176d7e0149dd83aa24c798e8f51. 
Note this will cause clients with a non-trampoline channel that have trampoline enabled to stop signaling trampoline support in invoices. With the upside that they now will be able to receive on this channel even if trampoline is enabled.

electrum/lnworker.py
```diff
@@ -2572,7 +2573,12 @@ def get_bolt11_invoice(
 
         assert amount_msat is None or amount_msat > 0
         timestamp = int(time.time())
-        routing_hints = self.calc_routing_hints_for_invoice(amount_msat, channels=channels)
+        routing_hints = self.calc_routing_hints_for_invoice(
+            amount_msat,
+            channels=channels,
+            # if the invoice_features signal trampoline support all included r_tags should support trampoline forwarding
```
ecdsa, line 2579: note that we could also set the feature dynamically depending on the liquidity in trampoline channels.

f321x: yes, added a TODO in https://github.com/spesmilo/electrum/pull/10541/commits/0265c7076694dedad9e5571ccea5c280a696bc88 to keep it simple.
We could pass the payment hash to get_bolt11_invoice and then get the payment info locally and modify the invoice features. Or store only a subset of the invoice features required for receiving in PaymentInfo (but this is weird as we use the PaymentInfo also for sending).

PR #10544, opened 2026-03-24
Lazy trampoline

ecdsa:
This pull request allows a trampoline node to forward payments without having a `channel_db`, if it has direct channels to other trampoline nodes.

If we (the trampoline forwarder) receive a trampoline onion and we do not have a direct connection to the next trampoline, we will return an `unknown_next_peer` error, with a list of trampolines we have a channel with in the error data. The client will use that list to create a longer trampoline route.

Note that trampolines are written in the data field of the `unknown_next_peer` message. (we could use custom TLV fields instead, but that would only works if the route from the client to the trampoline is of length 1, so I think it is better not to).

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2401,7 +2401,10 @@ class LNWallet(Logger):
             self.logger.info(f"trying split configuration: {sc.config.values()} rating: {sc.rating}")
             routes = []
             try:
-                if self.uses_trampoline():
+                is_direct_path = all(node_id == paysession.invoice_pubkey for (chan_id, node_id) in sc.config.keys())
+                if self.uses_trampoline() and not is_direct_path:
+                    if fwd_trampoline_onion:
+                        raise NoPathFound()
                     per_trampoline_channel_amounts = defaultdict(list)
                     # categorize by trampoline nodes for trampoline mpp construction
                     for (chan_id, _), part_amounts_msat in sc.config.items():
@@ -2473,19 +2476,28 @@ class LNWallet(Logger):
                     for (chan_id, _), part_amounts_msat in sc.config.items():
                         for part_amount_msat in part_amounts_msat:
                             channel = self._channels[chan_id]
-                            route = await run_in_thread(
-                                partial(
+                            if is_direct_path:
+                                route = self.create_direct_route(
+                                    amount_msat=part_amount_msat,
+                                    channel=channel,
+                                )
+                            else:
+                                assert not self.uses_trampoline()
+                                route = await run_in_thread(partial(
                                     self.create_route_for_single_htlc,
                                     amount_msat=part_amount_msat,
                                     invoice_pubkey=paysession.invoice_pubkey,
-                                    min_final_cltv_delta=paysession.min_final_cltv_delta,
                                     r_tags=paysession.r_tags,
                                     invoice_features=paysession.invoice_features,
                                     my_sending_channels=[channel] if is_multichan_mpp else my_active_channels,
                                     full_path=full_path,
-                                    budget=budget._replace(fee_msat=budget.fee_msat // sc.config.number_parts()),
-                                )
-                            )
+                                ))
+                            if not is_route_within_budget(
+                                    route, budget=budget,
+                                    amount_msat_for_dest=amount_msat,
+                                    cltv_delta_for_dest=paysession.min_final_cltv_delta):
+                                self.logger.info(f"rejecting route (exceeds budget): {route=}. {budget=}")
+                                raise FeeBudgetExceeded()
                             shi = SentHtlcInfo(
                                 route=route,
                                 payment_secret_orig=paysession.payment_secret,
@@ -2509,17 +2521,40 @@ class LNWallet(Logger):
             raise fee_related_error
         raise NoPathFound()
 
+    def create_direct_route(
+            self, *,
+            amount_msat: int,  # that final receiver gets
+            channel: Channel,
+    ) -> LNPaymentRoute:
+        self.logger.info(f'create_direct_route {channel.node_id.hex()}')
+        my_sending_channels = {channel.short_channel_id: channel}
+        channel_policy = get_mychannel_policy(
+            short_channel_id=channel.short_channel_id,
+            node_id=self.node_keypair.pubkey,
+            my_channels=my_sending_channels)
+        fee_base_msat = channel_policy.fee_base_msat
+        fee_proportional_millionths = channel_policy.fee_proportional_millionths
+        cltv_delta = channel_policy.cltv_delta
+        route_edge = RouteEdge(
+            start_node=self.node_keypair.pubkey,
+            end_node=channel.node_id,
+            short_channel_id=channel.short_channel_id,
+            fee_base_msat=fee_base_msat,
+            fee_proportional_millionths=fee_proportional_millionths,
+            cltv_delta=cltv_delta,
+            node_features=0)
+        route = [route_edge]
+        return route
+
     @profiler
     def create_route_for_single_htlc(
             self, *,
             amount_msat: int,  # that final receiver gets
             invoice_pubkey: bytes,
-            min_final_cltv_delta: int,
             r_tags,
             invoice_features: int,
             my_sending_channels: List[Channel],
             full_path: Optional[LNPaymentPath],
-            budget: PaymentFeeBudget,
     ) -> LNPaymentRoute:
 
         my_sending_aliases = set(chan.get_local_scid_alias() for chan in my_sending_channels)
@@ -2579,11 +2614,6 @@ class LNWallet(Logger):
             raise NoPathFound() from e
         if not route:
             raise NoPathFound()
-        if not is_route_within_budget(
-            route, budget=budget, amount_msat_for_dest=amount_msat, cltv_delta_for_dest=min_final_cltv_delta,
-        ):
-            self.logger.info(f"rejecting route (exceeds budget): {route=}. {budget=}")
-            raise FeeBudgetExceeded()
         assert len(route) > 0
         if route[-1].end_node != invoice_pubkey:
             raise LNPathInconsistent("last node_id != invoice pubkey")
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -90,7 +90,7 @@ from .submarine_swaps import SwapManager
 from .mpp_split import suggest_splits, SplitConfigRating
 from .trampoline import (
     create_trampoline_route_and_onion, is_legacy_relay, trampolines_by_id, hardcoded_trampoline_nodes,
-    is_hardcoded_trampoline, decode_routing_info
+    is_hardcoded_trampoline, decode_routing_info, encode_next_trampolines, decode_next_trampolines
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


electrum/lnworker.py
```diff
@@ -2461,6 +2469,34 @@ async def create_routes_for_payment(
             raise fee_related_error
         raise NoPathFound()
 
+    def create_direct_route(
+            self, *,
+            amount_msat: int,  # that final receiver gets
+            min_final_cltv_delta: int,
+            channel: Channel,
+            budget: PaymentFeeBudget,
```
f321x: The `budget` parameter is unused. Did you want to check `route_edge` against the budget?

electrum/lnworker.py
```diff
@@ -2461,6 +2469,34 @@ async def create_routes_for_payment(
...
+    ) -> LNPaymentRoute:
+        self.logger.info(f'create_direct_route {channel.node_id.hex()}')
+        my_sending_aliases = set([channel.get_local_scid_alias()])
```
f321x: `my_sending_aliases` is unused

electrum/trampoline.py
```diff
@@ -41,6 +41,35 @@
 _TRAMPOLINE_NODES_UNITTESTS = {}  # used in unit tests
 
 TRAMPOLINE_HOPS_MAX_DATA_SIZE = 500
+NEXT_TRAMPOLINES_MAGIC = b'lazy'
+
+def encode_next_trampolines(trampoline_channels: list) -> bytes:
```
f321x: `trampoline_channels` is type hinted as `list`, but in `_maybe_forward_trampoline` you pass a `set`:

```diff
diff --git a/electrum/trampoline.py b/electrum/trampoline.py
index 756bb0d13..691f98485 100644
--- a/electrum/trampoline.py
+++ b/electrum/trampoline.py
@@ -2,7 +2,7 @@ import io
 import os
 import random
 import dataclasses
-from typing import Mapping, Tuple, Optional, List, Iterable, Sequence, Set, Any
+from typing import Mapping, Tuple, Optional, List, Iterable, Sequence, Set, Any, TYPE_CHECKING
 from types import MappingProxyType
 
 from .lnutil import LnFeatures, PaymentFeeBudget, FeeBudgetExceeded
@@ -16,6 +16,9 @@ from . import constants
 from .logging import get_logger
 from .util import random_shuffled_copy
 
+if TYPE_CHECKING:
+    from .lnchannel import Channel
+
 
 _logger = get_logger(__name__)
 
@@ -43,7 +46,7 @@ _TRAMPOLINE_NODES_UNITTESTS = {}  # used in unit tests
 TRAMPOLINE_HOPS_MAX_DATA_SIZE = 500
 NEXT_TRAMPOLINES_MAGIC = b'lazy'
 
-def encode_next_trampolines(trampoline_channels: list) -> bytes:
+def encode_next_trampolines(trampoline_channels: Iterable['Channel']) -> bytes:
     s = NEXT_TRAMPOLINES_MAGIC
     for chan in trampoline_channels:
         s += chan.node_id

```

electrum/lnworker.py
```diff
@@ -4045,7 +4085,17 @@ async def _maybe_forward_trampoline(
         except PaymentFailure as e:
             self.logger.debug(
                 f"maybe_forward_trampoline. PaymentFailure for {payment_hash.hex()=}, {payment_secret.hex()=}: {e!r}")
-            raise OnionRoutingFailure(code=OnionFailureCode.UNKNOWN_NEXT_PEER, data=b'')
+            if self.uses_trampoline():
+                # todo: use max fee & cltv if I have more than 1 channel to the same node
+                trampoline_channels = set(
+                    [chan for chan in self.channels.values()
+                     if chan.is_public() and chan.is_open()
```
f321x: Might be more conservative to filter for `is_active()` instead of `is_open()`?

Also, what about additionally filtering with `can_pay(amt_to_forward)`? It might be more reliable if each channel that is reported back can forward the amount, otherwise the sender needs to split between them.

electrum/trampoline.py
```diff
@@ -41,6 +41,35 @@
 _TRAMPOLINE_NODES_UNITTESTS = {}  # used in unit tests
 
 TRAMPOLINE_HOPS_MAX_DATA_SIZE = 500
+NEXT_TRAMPOLINES_MAGIC = b'lazy'
```
f321x: Should we add a version number so future changes to the lazy trampoline are simpler to handle?

electrum/trampoline.py
```diff
@@ -198,11 +226,6 @@ def _choose_second_trampoline(
     trampolines = set(trampolines)
     if my_trampoline in trampolines:
         trampolines.discard(my_trampoline)
-    for r in failed_routes:
```
f321x, line 236: Is this `failed_routes` check not required anymore?
I assume you removed it because the `UNKNOWN_NEXT_PEER` containing the `next_trampolines` also got added to the `PaySession.failed_trampoline_routes` in `handle_failed_trampoline_htlc` on master?

electrum/lnworker.py
```diff
@@ -860,7 +860,7 @@ def __init__(
             amount_to_pay: int,  # total payment amount final receiver will get
             invoice_pubkey: bytes,
             uses_trampoline: bool,  # whether sender uses trampoline or gossip
-            use_two_trampolines: bool,  # whether legacy payments will try to use two trampolines
+            next_trampolines: Sequence[bytes],
```
f321x: stale type hint
```suggestion
            next_trampolines: dict[bytes, tuple],
```

electrum/lnworker.py
```diff
@@ -931,6 +929,9 @@ def handle_failed_trampoline_htlc(self, *, htlc_log: HtlcLog, failure_msg: Onion
                 self.failed_trampoline_routes.append(r)
             else:
                 pass  # maybe the route was reused between different MPP parts
+            if failure_msg.code == OnionFailureCode.UNKNOWN_NEXT_PEER:
+                self.next_trampolines = decode_next_trampolines(failure_msg.data)
```
f321x: If the sender splits a payment e.g. between two lazy trampoline forwarders, both fail back, and only one of them returns `next_trampolines` it could happen that the sender immediately overwrites the `next_trampolines` received in the first error with the empty dict returned by `decode_next_trampolines` for the second, empty `failure_msg.data`. This could cause the sender to never try to use them?

```suggestion
            if failure_msg.code == OnionFailureCode.UNKNOWN_NEXT_PEER:
                self.next_trampolines.update(decode_next_trampolines(failure_msg.data))
                # only keeping 10 to prevent unbounded growth
                self.next_trampolines = {id: self.next_trampolines[id] for id in list(self.next_trampolines)[:10]}
```

or replacing them but only if new ones are available:

```suggestion
            if failure_msg.code == OnionFailureCode.UNKNOWN_NEXT_PEER and failure_msg.data:
                self.next_trampolines = decode_next_trampolines(failure_msg.data)
```

electrum/trampoline.py
f321x, line 361: On the maximum trampoline fee level `_allocate_fee_along_route` will allocate the whole budget on all edges with placeholder fees. 
If an additional `next_trampoline` gets added, the sum of the allocated fees will exceed the budget, so `is_route_within_budget` will fail. 
You could deduct the `next_trampoline` fee from the `budget` so that the following `_allocate_fee_along_route` will only allocate the remaining budget to the remaining edges.

electrum/lnworker.py
ecdsa: removed

electrum/trampoline.py
ecdsa: done

electrum/lnworker.py
ecdsa: fixed

electrum/lnworker.py
ecdsa: indeed, this additional filtering makes sense. done.

electrum/lnworker.py
ecdsa: I see. Indeed this is an issue. Unfortunately it cannot be fixed by updating the dict, because the set of next trampolines might differ for each primary trampoline.

electrum/lnworker.py
ecdsa: Lol, I added `can_pay(amt_to_forward)` to the channel filter, and now Claude complains that this might disclose channel liquidity to third parties..

electrum/lnworker.py
ecdsa: ok, I believe it is fixed now: `PaySession.next_trampolines` is now a dict of dicts, where the first index is the `node_id` that sent the error message

electrum/trampoline.py
ecdsa: Indeed, we should keep it. This logic should also apply to `invoice_trampolines`.
I modified `failed_routes`, so that the routes are truncated at the erring `node_id`.
choose_next_trampoline now discards that node.

electrum/trampoline.py
ecdsa: Not sure what you mean. The budget is passed to `_allocate_fee_along_route`, so I do not think the sum of allocated fees could exceed the budget. Or am I missing something?

Anyway, that would be a pre-existing issue.

electrum/lnworker.py
ecdsa: ah, yes, but I wasn't sure whether we should duplicate code between `create_route_for_single_htlc` and `create_direct_route`. fixed now.

electrum/trampoline.py
ecdsa: sure, this never hurts

electrum/trampoline.py
ecdsa: oh, I see what you mean: the first edge does not have a placeholder fee, and `_allocate_fee_along_route` does not deduct that from its budget.

f321x: yes, the newly added `next_trampoline` here:
```python
        if next_trampolines and next_trampoline not in invoice_trampolines:
            next_trampoline = _choose_next_trampoline(next_trampoline, next_trampolines_ids, failed_routes)
            _extend_trampoline_route(route, end_node=next_trampoline, fee_info=next_trampolines[next_trampoline])
```
does already allocate fees for itself from `fee_info=next_trampolines[next_trampoline]`.

Then `_allocate_fee_along_route` following below will split the whole `PaymentFeeBudget` between the edges with `PLACEHOLDER_FEE` on `MAX_LEVEL` trampoline fees.

So after that the actual allocated fees are budget + next_trampolione fee which exceeds the budget in `is_route_within_budget`.

So `_allocate_fee_along_route` should probably consider the fees that have already been allocated before and then only allocate the remaining budget to the edges with placeholder to ensure the whole budget is not exceeded(?)

electrum/trampoline.py
```diff
@@ -294,6 +296,11 @@ def create_trampoline_route(
         if next_trampolines and next_trampoline not in invoice_trampolines:
             next_trampoline = _choose_next_trampoline(next_trampoline, next_trampolines_ids, failed_routes)
             _extend_trampoline_route(route, end_node=next_trampoline, fee_info=next_trampolines[next_trampoline])
+            # compute first_edge_fee (this is overshooting)
```
f321x: Isn't the same required for the legacy trampoline branch above, e.g. if we have 
A ---(free)--> B (my_trampoline) --(next trampoline edge fee)--> C (next Trampoline) --(placeholder fee)--> D (invoice_pubkey)  ?

electrum/trampoline.py
ecdsa: fixed in #10606

electrum/trampoline.py
ecdsa: outdated

PR #10547, opened 2026-03-25
set restrictive unix umask application-wide by default

SomberNight:
- I noticed we were creating the RPC server unix domain socket with 0o775 unix permissions.
  Instead of hunting down each individual line we create files/dirs, we should
  just set a restrictive umask by default. We can use chmod to relax this
  for individual files. -- but we should try to be secure by default
- note: bitcoin core [does the same](https://github.com/bitcoin/bitcoin/blame/2fe76ed8324af44c985b96455a05c3e8bec0a03e/src/common/system.cpp#L92)
    

Note: the first commit sets restrictive unix perms specifically for the RPC server unix socket - doing that is kind of redundant with setting the umask. We also manually set unix permissions in specific cases, e.g. for the [config file](https://github.com/spesmilo/electrum/blob/bd4439945a265b56205aa75121a48e4ab065a8b5/electrum/simple_config.py#L457) or for [wallet files](https://github.com/spesmilo/electrum/blob/bd4439945a265b56205aa75121a48e4ab065a8b5/electrum/storage.py#L108). I think we can keep these - for sensitive files it is worth the redundancy IMO.

### DIFF
--- a/run_electrum
+++ b/run_electrum
@@ -107,6 +107,15 @@ if is_android:
     ctypes.pythonapi = ctypes.PyDLL("libpython%d.%d.so" % sys.version_info[:2])  # replaces ctypes.PyDLL(None)
 
 
+# Set default application-wide file umask to more restrictive than typical.
+# We want to create all files and directories (esp. inside the datadir) with locked-down permissions.
+# note: this helps even on Windows! (see https://docs.python.org/3/library/os.html#os.mkdir)
+#       > `os.mkdir(path, mode=0o777, *, dir_fd=None)`
+#       >   On Windows, a mode of 0o700 is specifically handled to apply access control to the new
+#       >   directory such that only the current user and administrators have access.
+os.umask(0o0077)
+
+
 sys._ELECTRUM_RUNNING_VIA_RUNELECTRUM = True  # used by logging.py
 
 from electrum.logging import get_logger, configure_logging  # import logging submodule first


electrum/__init__.py
```diff
@@ -53,3 +53,12 @@ class GuiImportError(ImportError):
...
 
+
+# Set default application-wide file umask to more restrictive than typical.
+# We want to create all files and directories (esp. inside the datadir) with locked-down permissions.
+# note: this helps even on Windows! (see https://docs.python.org/3/library/os.html#os.mkdir)
+#       > `os.mkdir(path, mode=0o777, *, dir_fd=None)`
+#       >   On Windows, a mode of 0o700 is specifically handled to apply access control to the new
+#       >   directory such that only the current user and administrators have access.
+os.umask(0o0077)
+
```
ecdsa: Any reason why this is in `electrum/__init__.py` rather than in `run_electrum` ?
This will side effect any third-party application that imports the electrum module.

SomberNight: Indeed, good point. I had already considered this and was  on the verge re what to decide.
On one hand it's tempting to always want the more secure option, OTOH it's not nice to side-effect the calling code when they import a library.

Conceptually this is similar to ["harden_memory_linux"]( https://github.com/spesmilo/electrum/blob/d33212656f7a3f31f90d568adbc59d8467454304/run_electrum#L451-L453), where we decided not to do it for the library-scenario. So let's just do the same here.

Moved in https://github.com/spesmilo/electrum/pull/10547/changes/ab1c39110554e1b8921ac98e4c48e26d55b07a6e

SomberNight:
(note: I did a few force-pushes to see if the LLM security review from https://github.com/spesmilo/electrum/pull/10553 is working)

PR #10548, opened 2026-03-25
in GUI mode, only start a limited minimal RPC server

SomberNight:
see commit messages

Based on a report by `Zuzana Kotásková <36777@mail.vsfs.cz>`

### DIFF
--- a/electrum/gui/qt/__init__.py
+++ b/electrum/gui/qt/__init__.py
@@ -439,6 +439,14 @@ class ElectrumGui(BaseElectrumGui, Logger):
         window.activateWindow()
         if uri:
             window.show_send_tab()
+            # Handle URI defensively - local attacker with access to RPC server and config file could get here:
+            #   - tell user something happened
+            window.notify(_("Updated 'Pay To' field to handle external URI"))
+            #   - clear all fields in Send tab:
+            #     - perhaps user was just filling out the fields, trying to make another payment.
+            #       e.g. if the given URI does not have an amount, we should clear the amount field
+            window.send_tab.do_clear()
+            #   - update "Pay To" field (and maybe others)
             window.send_tab.set_payment_identifier(uri)
         return window
 


run_electrum
```diff
@@ -509,7 +510,9 @@ def handle_cmd(*, cmdname: str, config: 'SimpleConfig', config_options: dict):
         configure_logging(config)
         fd = daemon.get_file_descriptor(config)
         if fd is not None:
-            d = daemon.Daemon(config, fd, start_network=False)
+            # When running in GUI mode, only start a limited minimal RPC server, to limit attack surface.
+            only_minimal_jsonrpc = not constants.net.TESTNET
+            d = daemon.Daemon(config, fd, start_network=False, only_minimal_jsonrpc=only_minimal_jsonrpc)
```
SomberNight, line 515: Note: for basically the whole of 2018, we had the rpc server locked down like this, after https://github.com/spesmilo/electrum/issues/3374
(between https://github.com/spesmilo/electrum/commit/e65353c0626ac8c0e34e42369515c41298128888 and https://github.com/spesmilo/electrum/commit/3ec0ceba3e7c65404d2579c64fac9449cbc450f1, which was only released in December)
In fact that was even more strict, as it gutted the "gui" command as well.

run_electrum
```diff
@@ -509,7 +510,9 @@ def handle_cmd(*, cmdname: str, config: 'SimpleConfig', config_options: dict):
         configure_logging(config)
         fd = daemon.get_file_descriptor(config)
         if fd is not None:
-            d = daemon.Daemon(config, fd, start_network=False)
+            # When running in GUI mode, only start a limited minimal RPC server, to limit attack surface.
+            only_minimal_jsonrpc = not constants.net.TESTNET
```
SomberNight, line 514: An alternative we could branch on that is hard to fake is `is_git_clone`.
Using a config var would not help against an attacker that has filesystem write access to the config file.

f321x: In theory we could also have some CLI argument `--with-rpc` that starts the application with full RPC server independent of the network? Not sure if this is really that useful though.

f321x:
Tested both commits, seems like a useful security improvement.

run_electrum
SomberNight: Yes, that is a decent idea. We would make the option CLI-flag only, and ensure it cannot be set in the config file. Anyway, I think this is ok as-is.

PR #10553, opened 2026-03-26
ci: add claude code pr security review

f321x:
Adds a CI step to the Cirrus CI which will run claude code on the diff of a Pull Request and fail if it finds critical security vulnerabilities or serious code issues. Optinally it can be given a GitHub api key to create a comment in the pull request. We can play a bit with this, if its annoying we can as well just remove it again.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -243,6 +243,29 @@ task:
   main_script:
     - contrib/ban_unicode.py
 
+task:
+  name: "security review: Claude Code"
+  # NOTE: claude has access to all API keys available in the Cirrus CI environment.
+  # If we would add some critical api keys in here we should consider this.
+  matrix:
+    - trigger_type: automatic
+      only_if: $CIRRUS_PR != '' && ($CIRRUS_USER_PERMISSION == 'write' || $CIRRUS_USER_PERMISSION == 'admin')
+    - trigger_type: manual
+      only_if: $CIRRUS_PR != ''
+  container:
+    image: node:20
+    cpu: 1
+    memory: 2G
+  # CLAUDE_CODE_OAUTH_TOKEN is set as an encrypted "override" in https://cirrus-ci.com/settings/...
+  # It must be stored encrypted (ENCRYPTED[...]) so Cirrus CI refuses to decrypt it for
+  # fork PRs from users without write permission.
+  # Generate with: claude setup-token
+  # Optional: set GITHUB_TOKEN to enable PR comments on failure
+  install_script:
+    - npm install -g @anthropic-ai/claude-code
+  review_script:
+    - python3 contrib/ci/claude_security_review.py
+
 # Cron jobs configured in https://cirrus-ci.com/settings/...
 # - job "nightly" on branch "master" at "0 30 2 * * ?"  (every day at 02:30Z)
 task:
--- a/contrib/ci/claude_security_review.py
+++ b/contrib/ci/claude_security_review.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+"""
+Cirrus CI task: Claude Code security review for Electrum pull requests.
+
+Runs Claude Code against the PR diff to detect critical security
+vulnerabilities. Optionally posts findings as a GitHub PR comment.
+
+Exit codes:
+    0 -- PASS (no critical/high issues)
+    1 -- FAIL (critical/high issues found)
+    2 -- review could not run (infra error, logged as warning)
+
+Environment variables:
+    Required:
+        CLAUDE_CODE_OAUTH_TOKEN  -- OAuth token from `claude setup-token` (MAX subscription)
+    Optional:
+        GITHUB_TOKEN             -- GitHub token for posting PR comments
+    Set by Cirrus CI:
+        CIRRUS_PR                -- PR number (empty if not a PR build)
+        CIRRUS_BASE_BRANCH       -- target branch of the PR
+        CIRRUS_REPO_FULL_NAME    -- e.g. "spesmilo/electrum"
+        CIRRUS_TASK_ID           -- current Cirrus task ID
+"""
+
+import json
+import os
+import re
+import subprocess
+import sys
+import urllib.error
+import urllib.request
+
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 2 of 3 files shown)


.cirrus.yml
```diff
@@ -243,6 +243,29 @@ task:
...
 
+task:
+  name: "security review: Claude Code"
+  # NOTE: claude has access to all API keys available in the Cirrus CI environment.
+  # If we would add some critical api keys in here we should consider this.
+  matrix:
+    - trigger_type: automatic
+      only_if: $CIRRUS_PR != '' && ($CIRRUS_USER_PERMISSION == 'write' || $CIRRUS_USER_PERMISSION == 'admin')
+    - trigger_type: manual
+      only_if: $CIRRUS_PR != '' && ($CIRRUS_USER_PERMISSION == 'write' || $CIRRUS_USER_PERMISSION == 'admin')
```
SomberNight: I guess the `only_if` for the manual trigger should be either the inverse of the automatic trigger, or maybe manual could be unconditional?
Not sure we can even do the manual trigger for external contributors if you set `only_if` to write access.

.cirrus.yml
```diff
@@ -243,6 +243,29 @@ task:
...
+  container:
+    image: node:20
+    cpu: 1
+    memory: 2G
+  # CLAUDE_CODE_OAUTH_TOKEN is set as an encrypted "override" in https://cirrus-ci.com/settings/...
+  # It must be stored encrypted (ENCRYPTED[...]) so Cirrus CI refuses to decrypt it for
+  # fork PRs from users without write permission.
```
SomberNight, line 261: note: every CI task gets access to all "encrypted" variables.
atm the most sensitive secret is `crowdin_api_key` that has write access to the crowdin translations. we don't e.g. publish releases and push to PyPI from the CI.
but we have to keep in mind that we trust claude and its full supply chain with all the "encrypted" environment variables (not just `CLAUDE_CODE_OAUTH_TOKEN`)

.cirrus.yml
SomberNight: To clarify, as-is, I think if an external contributor opens a PR, the manual "trigger" button would not appear and the task would be skipped.

Look at e.g. https://github.com/spesmilo/electrum/pull/10548/checks?check_run_id=68592680450
That CI run has a task for `build: Android (QML arm64-v8a)` with a manual "trigger" button.
Note that it does not have e.g. the `check submodules` task at all as [its `only_if`](https://github.com/spesmilo/electrum/blob/35b44a1e6482ae29562a1091d3410e854705c5d8/.cirrus.yml#L416) was not met.

We want the task to appear, but not run, unless the manual "trigger" button is pressed.

f321x: I removed the `CIRRUS_USER_PERMISSION` checks from the manual trigger in https://github.com/spesmilo/electrum/pull/10553/commits/e094623245199cade4aaba3ddc0dc366766f42ed.

.cirrus.yml
f321x: I documented this in a comment here: [e094623#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bbR248-R249](https://github.com/spesmilo/electrum/pull/10553/changes/e094623245199cade4aaba3ddc0dc366766f42ed#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bbR248-R249)

contrib/ci/claude_security_review.py
```diff
@@ -0,0 +1,262 @@
...
+        instructions = f.read()
+
+    if len(diff) > MAX_DIFF_CHARS:
+        diff = diff[:MAX_DIFF_CHARS] + f"\n\n... [diff truncated at {MAX_DIFF_CHARS} characters]"
+
+    return (
+        f"{instructions}\n\n"
+        f"---\n\n"
+        f"## Changed files\n\n```\n{changed_files}\n```\n\n"
+        f"## Diff\n\n```diff\n{diff}\n```"
```
SomberNight: Claude correctly self-flagged this check here: if the git diff is too long, the script should probably never "exit 0" (pass) in the end. Note that wallet.py itself is around 200k characters, so it is very much conceivable to have a diff for a large PR that is more than that. It's ok not to send crazy long prompts to the LLM, but then we should also prevent the script silently giving a "success" result.

f321x: My assumption was basically that 0/success should be fine as we wouldn't have this huge diffs often, and failing doesn't provide any additional value except for confusion, as we cannot change the fact that the llm has a limited input size (assuming we wouldn't want to split PRs just so this CI can run on it). 

Though maybe it is good if it fails in that case so at least we are aware that this check didn't happen.
However it might not even make sense to run the check at all if the diff has to be truncated as i imagine the review quality will suffer if some part of the diff is just missing.

I changed it to just fail with status code 2 if the limit is exceeded and raised the limit to 800k chars, 200k was a bit conservative, the opus model which this script currently uses can handle this well apparently.

SomberNight: > Though maybe it is good if it fails in that case so at least we are aware that this check didn't happen.

Yes, that was my thinking as well.

contrib/ci/security_review_prompt.md
```diff
@@ -0,0 +1,92 @@
...
+developer time and erodes trust in this review.
+
+## Scope
+
+Focus your findings on the diff provided below -- only flag issues introduced or worsened by
+changes in this PR. You have access to the full Electrum codebase; use it freely to read
+surrounding code, trace call chains, and understand what the diff actually does. But do not
+audit code outside the diff -- the codebase is context, not the review target.
+Focus on changes that introduce, worsen, or fail to mitigate security vulnerabilities.
+Only flag issues introduced or worsened by added (`+`) lines in the diff. Do not flag
```
SomberNight: > - whole PR is deleting security checks
> - no new code added, only deletes
> - every line starts with (`-`) sign

Claude:
Image

f321x: Yeah ok the 'only + lines' might be a bit too explicit, i removed that https://github.com/spesmilo/electrum/compare/b1fa14f45a67e149544afffb4b79237ce024fc41..88f9c49a603d9a167ad9341264a2ecfdaf91db2c.

PR #10554, opened 2026-03-26
plugin.py: fix some type hints

SomberNight:


electrum/plugin.py
```diff
@@ -809,8 +810,7 @@ def read_file(self, name: str, filename: str) -> bytes:
             with open(path, 'rb') as myfile:
                 return myfile.read()
         else:
-            # no icon
-            return None
```
SomberNight, line 813: I don't understand what this branch was trying to do. Could this even be reached? Note that it is branching on `name`, which is the name of the plugin (not the file).
it was added in https://github.com/spesmilo/electrum/commit/79941529d2587d4055e327a090e15d0025b8af51

Also, pretty much everywhere we call `read_QIcon_from_bytes` on the return value, which would raise on `None`.

PR #10558, opened 2026-03-27
trampoline: prevent adding ourself on the route, fixes CI

f321x:
Followup #10541.
Fixes tests.regtest.TestLightningSwapserver.test_swapserver_forceclose.

In the regtest bob would now signal trampoline support due to #10541 and include Alice into the invoice trampoline as he is connected to Alice. Alice would then try to add herself onto the trampoline route (`if add_trampoline`), causing the payment to fail.

### DIFF
--- a/electrum/trampoline.py
+++ b/electrum/trampoline.py
@@ -224,6 +224,8 @@ def create_trampoline_route(
 ) -> LNPaymentTRoute:
     # we decide whether to convert to a legacy payment
     is_legacy, invoice_trampolines = is_legacy_relay(invoice_features, r_tags)
+    # we can be in the invoice_trampolines e.g. if we have a direct channel with the recipient
+    invoice_trampolines.discard(my_pubkey)
     _logger.debug(f"Creating trampoline route for invoice_pubkey={invoice_pubkey.hex()}, {is_legacy=}")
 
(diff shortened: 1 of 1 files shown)


electrum/trampoline.py
```diff
@@ -224,6 +224,8 @@ def create_trampoline_route(
 ) -> LNPaymentTRoute:
     # we decide whether to convert to a legacy payment
     is_legacy, invoice_trampolines = is_legacy_relay(invoice_features, r_tags)
+    # we can be in the invoice_trampolines e.g. if we have a direct channel with the recipient
```
SomberNight, line 227: I am a bit confused. Why is Alice even using trampoline to pay? Don't we force it to use gossip here: (?)
https://github.com/spesmilo/electrum/blob/d33212656f7a3f31f90d568adbc59d8467454304/tests/regtest/regtest.sh#L172

f321x: This config is overridden for `TestLightningSwapserver`:
https://github.com/spesmilo/electrum/blob/efcf1f056fc220b73ad3f054220d71a670abc8f9/tests/regtest.py#L89-L95

SomberNight: Hmm, I see.

PR #10565, opened 2026-04-01
prep 4.7.2: backports to 4.7.x

SomberNight:
These are the changes I would cherry-pick for 4.7.2

### DIFF
--- a/RELEASE-NOTES
+++ b/RELEASE-NOTES
@@ -1,3 +1,35 @@
+# Release 4.7.2 (April 1, 2026)
+ * General:
+   - changed: set restrictive unix umask (0077) application-wide by default (#10547)
+   - fix: failing assert for wallets with old (2023) still unpaid LN payment requests (#10502)
+ * Qt GUI (desktop):
+   - changed: move LN fee slider to payment dialog (#10516)
+   - fix: auto-close context menu in Addresses and Coins tab when underlying state changes (#10467)
+   - fix: Coins tab: only enable 'fully spend' menu if there are unfrozen coins in selection (#10503)
+ * QML GUI & Android:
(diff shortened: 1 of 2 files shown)


f321x:
Looks complete to me. Was thinking about #10524 as it makes the checksum **correct** but its kind of a breaking change anyways.

SomberNight:
> Was thinking about https://github.com/spesmilo/electrum/pull/10524

We never consume those checksums, only produce them. I don't know what the tools consuming them expect, but it is the original author who wanted to make the change. Also it seems to conform to the BIP text (then again, *both* the old code and the new code conforms to the BIP text -- it is simply that the old code did not produce the same checksum in some cases as the javascript *code example* in the BIP). I think it's fine.

SomberNight:
Added release notes.

RELEASE-NOTES
```diff
@@ -1,3 +1,35 @@
...
+ * Lightning:
+   - changed: lnwatcher now subscribes to fewer non-channel-related addresses (#10533)
+ * CLI/RPC:
+   - changed: daemon: forbid "setconfig" command to change rpcserver settings in-flight (#10534)
+   - changed: in GUI mode, only start a limited minimal RPC server (#10548)
+ * Plugins:
+   - nwc (Nostr Wallet Connect):
+     - sync with spec, fix some bugs (#10505, #10511, #10555)
+   - timelock_recovery:
+     - fix: checksum calculation to follow corresponding BIP-0128 (#10524)
```
f321x, line 30: This is not included, right?

SomberNight: it is there, in 29ddec075f5d03ea472ae10200b538a93231b7d9

f321x: oh, i was grepping for #10524 and missed that

PR #10573, opened 2026-04-08
qml: allow renaming wallets

### LINKED ISSUE #4377: Add option to rename wallet
It would be useful to be able to rename wallets in the Android app, because I tend to be lost by the weird name I can come up with for my wallets.

I know I could delete them and add the wallet back, but that's not so convenient.

f321x:
Allows to rename a wallet file from the QML Wallet Details view. This seems like a feature we should support as the use-case of a wallet can change or maybe the user didn't think about a proper name when setting up the wallet. Especially with lightning channels it is not possible to restore from seed to change the name.
Fixes #4377

Alternatively it could also be made a button in the bottom row, but we already have 4 potential buttons there, i think this looks cleaner.

[Screencast_20260408_103904.webm](https://github.com/user-attachments/assets/751660ab-dd36-4827-8d9c-83a2a0ba9efc)

### DIFF
--- a/electrum/gui/qml/qedaemon.py
+++ b/electrum/gui/qml/qedaemon.py
@@ -150,6 +150,7 @@ class QEDaemon(AuthMixin, QObject):
     walletRequiresPassword = pyqtSignal([str, str], arguments=['name', 'path'])
     walletOpenError = pyqtSignal([str], arguments=["error"])
     walletDeleteError = pyqtSignal([str, str], arguments=['code', 'message'])
+    walletRenameError = pyqtSignal([str], arguments=['message'])
 
     def __init__(self, daemon: 'Daemon', plugins: 'Plugins', parent=None):
         super().__init__(parent)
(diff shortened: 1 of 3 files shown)


electrum/gui/qml/qedaemon.py
```diff
@@ -310,6 +311,34 @@ def delete_wallet(self, wallet):
...
+    @pyqtSlot(str, result=bool)
+    def isValidWalletName(self, wallet_name: str) -> bool:
+        if not wallet_name:
+            return False
+        if self.availableWallets.wallet_name_exists(wallet_name):
+            return False
+        # ensure wallet_name is not interpreted as path
+        if os.path.basename(wallet_name) != wallet_name:  # '/foo/bar/' returns 'bar'
+            return False
+        return True
```
SomberNight, line 323: could this be deduped with https://github.com/spesmilo/electrum/blob/bca41d941dcc4e919042a3e1be42edb8b5984f82/electrum/gui/qml/qewizard.py#L132 ?

f321x: See https://github.com/spesmilo/electrum/pull/10604

PR #10575, opened 2026-04-10
pi: handle lud-17 lnurl URIs

f321x:
[LNURL LUD-17](https://github.com/lnurl/luds/blob/luds/17.md) is an alternative scheme for lnurl urls which formats the urls in cleartext with a `lnurlX://` scheme prefix instead of bech32 encoding them. 
Adding this is simple as we already support lnurl and makes Electrum compatible e.g. with [Boltcards](https://www.boltcard.org/).

### DIFF
--- a/electrum/gui/qml/qeapp.py
+++ b/electrum/gui/qml/qeapp.py
@@ -23,6 +23,7 @@ from electrum.network import Network
 from electrum.plugin import run_hook
 from electrum.gui.common_qt.util import get_font_id
 from electrum.util import profiler
+from electrum.lnurl import SUPPORTED_LNURL_SCHEMES
 
 from .qeconfig import QEConfig
 from .qedaemon import QEDaemon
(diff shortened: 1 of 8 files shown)


electrum/gui/qml/qeapp.py
```diff
@@ -235,7 +236,9 @@ def on_new_intent(self, intent):
         data = str(intent.getDataString())
         self.logger.debug(f'received intent: {repr(data)}')
         scheme = str(intent.getScheme()).lower()
-        if scheme == BITCOIN_BIP21_URI_SCHEME or scheme == LIGHTNING_URI_SCHEME:
+        if scheme == BITCOIN_BIP21_URI_SCHEME \
+                or scheme == LIGHTNING_URI_SCHEME \
+                or scheme in SUPPORTED_LNURL_SCHEMES:
```
SomberNight, line 241: do we want to handle these URI schemes also in the desktop Qt GUI?

(related https://github.com/spesmilo/electrum/commit/016c8b5f4cd693c2b954d4632a2b0aff63c557d6)

---

edit: "maybe later" is a sensible answer, just asking

f321x: Sure, why not, i didn't consider that we register URI handlers on Desktop as well, your comment from 016c8b5 was helpful there.
Rebased them into https://github.com/spesmilo/electrum/pull/10575/changes/ef702d74cdff5a17e018351c71b13737fc0d938f

accumulator:
Tested, working on android and linux desktop

SomberNight:
Looks reasonable (haven't tested)

PR #10578, opened 2026-04-10
backend support for 'close reason'

### LINKED ISSUE #10389: How to tell if channel was closed by local or remote in JSON-RPC API?
### Description

When requesting data about lightning channels via the JSON-RPC API, is there a concise way to tell *who* requested a channel close, whether cooperatively or via a force close?

I see in the code there is a channel state, but only FORCE_CLOSING and REQUESTED_FCLOSE seem to clearly indicate the "who".

Thank you :)

wdiffman:
Closes #10389 following @SomberNight 's suggestion. Used ` reason`  instead of `diagnosis`, though. 
In the test, bob's close reason isn't guaranteed to be set by the time alice's close_channel returns, so I used a short sleep; please advise if there is a better way.


electrum/lnchannel.py
```diff
@@ -319,6 +341,7 @@ def get_ctx_sweep_info(self, ctx: Transaction) -> Tuple[bool, Dict[str, MaybeSwe
                 self.logger.info(f'we (local) force closed')
             elif who_closed == REMOTE:
                 self.logger.info(f'they (remote) force closed.')
+                self.save_close_reason(ChanCloseReason.REMOTE_FORCE)
```
SomberNight, line 347: You should also set the close reason in the LOCAL case here.
First of all, it is nicer to make the code (in the local scope) symmetric/homogeneous.
But also, ultimately the final close reason should reflect what gets mined and weird chains of events could happen: e.g. local force-close gets broadcast to mempool first, then remote force-close replaces it in mempool, but then local force-close gets mined. If you call `save_close_reason` here, it will self-heal.

We could keep the other call in `lnworker._force_close_channel` in addition to this one, as that one runs much earlier.

electrum/gui/qt/channels_list.py
```diff
@@ -40,6 +40,7 @@ class Columns(MyTreeView.BaseColumnsEnum):
         LOCAL_BALANCE = enum.auto()
         REMOTE_BALANCE = enum.auto()
         CHANNEL_STATUS = enum.auto()
+        CLOSE_REASON = enum.auto()
```
SomberNight: I think it does not make sense to add a column for this. It takes a lot of horizontal space and adds noise.
Isn't it enough to show this in the channel details dialog?

electrum/commands.py
```diff
@@ -1877,6 +1877,7 @@ def _filter(chan):
                 'remote_reserve': chan.config[LOCAL].reserve_sat,
                 'local_unsettled_sent': chan.balance_tied_up_in_htlcs_by_direction(LOCAL, direction=SENT) // 1000,
                 'remote_unsettled_sent': chan.balance_tied_up_in_htlcs_by_direction(REMOTE, direction=SENT) // 1000,
+                'close_reason': (r := chan.get_close_reason()) and r.name or None, # Harder to read, but keeps it within a single line.
```
SomberNight: note: this would make the names of the enum members part of the external API

also see other comment re wallet db upgrades

electrum/lnchannel.py
```diff
@@ -192,6 +199,21 @@ class AbstractChannel(Logger, ABC):
     _state: ChannelState
     _who_closed: Optional[int] = None  # HTLCOwner (1 or -1).  0 means "unknown"
 
+    def save_close_reason(self, reason: ChanCloseReason) -> None:
+        self.storage['close_reason'] = reason.name
```
SomberNight, line 205: This would make the names of the enum members part of the wallet db. That is, if someone later renames e.g. `ChanCloseReason.LOCAL_FORCE` to `ChanCloseReason.LOCAL_FORCE_CLOSE`, they need to do a wallet db upgrade.

IMO it is preferable to use the names and not the int values in the wallet db, so this is fine, but it is completely unintuitive and error-prone for the programmer, so at least a comment must be added.

Just duplicate this comment:
https://github.com/spesmilo/electrum/blob/1235b4a6b9530fc1536a776c169613ad75872c6a/electrum/lnchannel.py#L77-L78

tests/test_lnpeer.py
```diff
@@ -1625,6 +1626,29 @@ async def set_settle():
...
+        w1.network.config.TEST_SHUTDOWN_FEE = 100
+        w2.network.config.TEST_SHUTDOWN_FEE = 100
+        w1.network.config.TEST_SHUTDOWN_LEGACY = True
+        w2.network.config.TEST_SHUTDOWN_LEGACY = True
+        async def action():
+            await util.wait_for2(p1.initialized, 1)
+            await util.wait_for2(p2.initialized, 1)
+            await p1.close_channel(alice_channel.channel_id)
+            # alice's side is completed, but bob need to wait to see the reason.
+            await asyncio.sleep(1)  # FIXME: use a better wait
```
SomberNight: On my machine, the test passes even with this sleep completely removed. Is that not the case for you?

Anyway, the simplest less-wasteful thing that could be done is:
```py
            async with util.async_timeout(1):
                while not bob_channel.is_closed_or_closing():
                    await asyncio.sleep(0.01)
```

Something more complicated:
```diff
diff --git a/tests/test_lnpeer.py b/tests/test_lnpeer.py
index a4950ba9d5..b65f78a041 100644
--- a/tests/test_lnpeer.py
+++ b/tests/test_lnpeer.py
@@ -1636,12 +1636,21 @@ class TestPeerDirect(TestPeer):
         w2.network.config.TEST_SHUTDOWN_FEE = 100
         w1.network.config.TEST_SHUTDOWN_LEGACY = True
         w2.network.config.TEST_SHUTDOWN_LEGACY = True
+
+        any_chan_changed = asyncio.Event()
+        async def on_chan_changed(*args):
+            any_chan_changed.set()
+            any_chan_changed.clear()
+        util.register_callback(on_chan_changed, ["channel"])
+
         async def action():
             await util.wait_for2(p1.initialized, 1)
             await util.wait_for2(p2.initialized, 1)
             await p1.close_channel(alice_channel.channel_id)
             # alice's side is completed, but bob need to wait to see the reason.
-            await asyncio.sleep(1)  # FIXME: use a better wait
+            async with util.async_timeout(1):
+                while not bob_channel.is_closed_or_closing():
+                    await any_chan_changed.wait()
             self.assertEqual(alice_channel.get_close_reason(), ChanCloseReason.LOCAL_COOP)
             self.assertEqual(bob_channel.get_close_reason(), ChanCloseReason.REMOTE_COOP)
             gath.cancel()

```

electrum/commands.py
SomberNight: re the comment `# Harder to read, but keeps it within a single line.` -- ok, but the comment itself is not needed IMO

SomberNight (commented):
Hey. Thanks for the PR. Added some comments.

electrum/gui/qt/channels_list.py
wdiffman: Agree. I only added because I thought someone might want to search for channels closed with a given reason. But if that's the case, we can fix it later. I will remove.

tests/test_lnpeer.py
wdiffman: Indeed it does, but that's will not be necessarily true for all runs.
I will use the simple wait, but thanks for sharing the more complete version.

wdiffman:
@SomberNight Just letting you know that this PR is ready for another pass.

PR #10582, opened 2026-04-16
Using GET_MASTER_FINGERPRINT for Legacy Client to get the root public key fingerprint

iartemov-ledger:
**This PR provides an example of how new GET_MASTER_FINGERPRINT instruction of Ledger app-bitcoin can be adopted in a software wallet.**

### Scope
Since the hardening introduced in https://github.com/LedgerHQ/app-bitcoin/pull/297 (and in https://github.com/LedgerHQ/app-bitcoin/pull/299 for `bitcoin_testnet_legacy`), it is no longer possible to get the public key at the master level. Therefore, the new `GET_MASTER_FINGERPRINT` instruction was introduced to replace it.
 
The changes in this PR apply only to the legacy `app-bitcoin` starting from version `2.4.11` (and the `Ledger_Client_Legacy` part). Specifically for Electrum, this only affects the `bitcoin_testnet_legacy` app variant (`Bitcoin Test Legacy`), as there is no hardening applied to the `bitcoin_legacy` variant (`Bitcoin Legacy`). 
For `app-bitcoin-new` and the corresponding `Ledger_Client_New`, everything is already handled well through the `get_master_fingerprint()` call.

The new behavior is to first try the new instruction. If an error is returned (meaning the instruction does not exist in the app), it falls back to `getWalletPublicKey()`. This ensures backward compatibility with app versions older than v2.4.11.

### Tests
This has been successfully tested as follows:
1. Wallet creation with the previous unhardened version of `app-bitcoin` (using the `m/84h/1h/0h` path corresponding to `testnet`), and connecting to it with the hardened version of the app.
2. Wallet creation and connection using the already hardened version.

In the corresponding traces, it was observed that `INS_GET_WALLET_PUBLIC_KEY` commands with `bip32Path.length = 0` were effectively replaced by `INS_GET_MASTER_FINGERPRINT` calls.

### DIFF
--- a/electrum/plugins/ledger/ledger.py
+++ b/electrum/plugins/ledger/ledger.py
@@ -347,8 +347,9 @@ class Ledger_Client(HardwareClientBase, ABC):
     def __init__(self, *, plugin: HW_PluginBase):
         HardwareClientBase.__init__(self, plugin=plugin)
 
+    @abstractmethod
     def get_master_fingerprint(self) -> bytes:
-        return self.request_root_fingerprint_from_device()
+        pass
 
     @abstractmethod
     def show_address(self, address_path: str, txin_type: str):
@@ -390,6 +391,25 @@ class Ledger_Client_Legacy(Ledger_Client):
         self._product_key = product_key
         self._soft_device_id = None
 
+    def _get_master_fingerprint(self, bip32_path: str = "") -> bytes:
+        """Return the 4-byte fingerprint for a BIP32 node.
+
+        If bip32_path is empty (root / master key), tries the dedicated
+        GET_MASTER_FINGERPRINT APDU first (INS 0xD0), which does NOT
+        require DERIVE_MASTER permission.  If the device does not support
+        it (old firmware), falls back to getWalletPublicKey + HASH160.
+
+        For non-empty paths the fingerprint is always computed via
+        getWalletPublicKey + HASH160.
+        """
+        if bip32_path == "":
+            try:
+                return self.dongleObject.getMasterFingerprint()
+            except Exception:
+                _logger.info("getMasterFingerprint APDU not supported, falling back to getWalletPublicKey")
+        nodeData = self.dongleObject.getWalletPublicKey(bip32_path)
+        return hash_160(compress_public_key(nodeData['publicKey']))[0:4]
+
     def is_pairable(self):
         return True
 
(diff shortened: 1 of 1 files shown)


SomberNight:
> it is no longer possible to get the public key at the root level

Note that is not what we have been doing.
Instead the client has been asking for the *xpub* at `m/0h`, and reading out the *parent fingerprint* from that xpub to learn the master fingerprint.
That is, there are two ways to get the master fingerprint:
- either get the root *pubkey* (33 bytes) and calculate the hash, or
- get the xpub (78 bytes) at *any* `m/x` (hardened or unhardened x) and read out the *parent fingerprint* from that

https://github.com/spesmilo/electrum/blob/3fbf597485f651cf920b5566c6ed0967ad110c49/electrum/hw_wallet/plugin.py#L257-L262

Was the new Ledger Bitcoin app changed to start refusing returning the xpub for `m/0h`?
If so, are you going to open another pull request here to adapt the code? :)

Also, in the PR title, what do you mean by `for Legacy Client`? If you mean `Ledger_Client_Legacy`, it looks to me `Ledger_Client_New` is also affected (?)

iartemov-ledger:
> > it is no longer possible to get the public key at the root level
> 
> Note that is not what we have been doing. Instead the client has been asking for the _xpub_ at `m/0h`, and reading out the _parent fingerprint_ from that xpub to learn the master fingerprint. That is, there are two ways to get the master fingerprint:
> 
> * either get the root _pubkey_ (33 bytes) and calculate the hash, or
> * get the xpub (78 bytes) at _any_ `m/x` (hardened or unhardened x) and read out the _parent fingerprint_ from that
> 
> https://github.com/spesmilo/electrum/blob/3fbf597485f651cf920b5566c6ed0967ad110c49/electrum/hw_wallet/plugin.py#L257-L262
> 
> Was the new Ledger Bitcoin app changed to start refusing returning the xpub for `m/0h`? If so, are you going to open another pull request here to adapt the code? :)
> 
> Also, in the PR title, what do you mean by `for Legacy Client`? If you mean `Ledger_Client_Legacy`, it looks to me `Ledger_Client_New` is also affected (?)

Thank you for your feedback!
And sorry, I was a bit too quick with my PR. I still need some time to process everything (which is why I closed it).

The change only affects the legacy `app-bitcoin` that we've been hardening, and I did indeed mean `Ledger_Client_Legacy`. For `app-bitcoin-new` and the corresponding `Ledger_Client_New`, everything is already handled well through the `get_master_fingerprint()` call.

IIUC, the `request_root_fingerprint_from_device()` function you mentioned above, as well as its caller `get_soft_device_id()` from `electrum/hw_wallet/plugin.py`, are not called in the Ledger case. This is because they are overridden in `electrum/plugins/ledger/ledger.py` for both the New and Legacy clients.

PR #10585, opened 2026-04-17
qml: ElTextArea default topPadding, hide placeholderText on input 

f321x:
Sets a larger (than `padding`) default `topPadding` on the `ElTextArea` as i noticed that all ElTextAreas look better with some more padding at the top
Followup to https://github.com/spesmilo/electrum/pull/10579

Also hides the `placeholderText` of `ElTextArea` as soon as the user enters something, otherwise it would conflict with the user input and look broken.

E.g. lnurlpay dialog:
before:
Screenshot_20260417_120555

after:
Screenshot_20260417_120446

### DIFF
--- a/electrum/gui/qml/components/NostrConfigDialog.qml
+++ b/electrum/gui/qml/components/NostrConfigDialog.qml
@@ -74,7 +74,6 @@ ElDialog {
                     Layout.fillWidth: true
                     Layout.fillHeight: true
                     font.family: FixedFont
-                    topPadding: constants.paddingLarge
                     wrapMode: TextEdit.WrapAnywhere
                     onTextChanged: valid = verify(text)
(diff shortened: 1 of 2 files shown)


accumulator:
The current styling is basically Qt6.4 specific. I assume you run it locally with a higher Qt version? That would explain the wrong padding.
In that case, don't bother with the styling until the https://github.com/spesmilo/electrum/pull/10485 branch is merged

f321x:
@accumulator Ah yes, on current the Qt 6.4 Android build this padding issue actually doesn't exist. I noticed it on Qt 6.5 (Desktop QML) and on your Qt 6.10 branch, so we should probably have a look again after #10485 is merged, or you include it in #10485.

f321x:
> In that case, don't bother with the styling until the https://github.com/spesmilo/electrum/pull/10485 branch is merged

Tested this again (Android and Desktop QML), the PR is working as intended.

SomberNight:
Looks reasonable, and I also tested at runtime.

PR #10586, opened 2026-04-17
millisat precision

### LINKED ISSUE #6253: user interface only has satoshi precision (msat truncated to sat, rounding)
Currently the user interface only has satoshi precision.
Amounts that are often in msat are truncated to sat.
For example, (mostly re Qt, but applies to kivy too in general)
- if the user tries to pay an invoice that has msat precision, we will only show the truncated sat value
- in the history, balances/deltas are in sat
- in the channels tab, balances are in sat

Not sure what to do in general, but perhaps at least for invoices we should show msat resolution amounts.

### LINKED ISSUE #10412: lnurl with msat precision cannot be paid
### Description

We currently cannot pay an lnurl-pay endpoint that requires msat precision.
For example, if they set `minSendable==maxSendable==100_000_500` (msat), we can't pay that invoice.

https://github.com/spesmilo/electrum/blob/ca597942fd80bab74cd83d7f338a28ac01b05e9b/electrum/lnurl.py#L153-L156
https://github.com/spesmilo/electrum/blob/ca597942fd80bab74cd83d7f338a28ac01b05e9b/electrum/payment_identifier.py#L420-L421
https://github.com/spesmilo/electrum/blob/5db4a16ace8cd00b63f80bb678bc402db2f08312/electrum/gui/qml/qeinvoice.py#L653-L654

could test with https://github.com/f321x/electrum-lnurl-server

accumulator:
support millisat precision throughout the codebase.

- kwarged backend wallet funcs so it is clear whether sats or millisats as `int`s are passed
- lnurl now native millisats
- `Decimal` is used on the GUI side where it's easier to assume sats are the base unit
    - avoids having to check `sat` and `msat` fields/parameters
    - can check easily if amount has millisat precision where not allowed
    - extra millisat precision is carried transparently for eventual passing to backend
- QML `QEAmount` is now synced between sats and millisats, simplifying assumptions and reducing if-else checks

should fix #6253, #10412


electrum/gui/qml/components/LnurlWithdrawRequestDialog.qml
```diff
@@ -20,20 +20,20 @@ ElDialog {
     needsSystemBarPadding: false
 
     property int walletCanReceive: 0
-    property int providerMinWithdrawable: parseInt(requestDetails.lnurlData['min_withdrawable_sat'])
-    property int providerMaxWithdrawable: parseInt(requestDetails.lnurlData['max_withdrawable_sat'])
+    property int providerMinWithdrawable: parseInt(requestDetails.lnurlData['min_withdrawable_msat'])
+    property int providerMaxWithdrawable: parseInt(requestDetails.lnurlData['max_withdrawable_msat'])
     property int effectiveMinWithdrawable: Math.max(providerMinWithdrawable, 1)
     property int effectiveMaxWithdrawable: Math.min(providerMaxWithdrawable, walletCanReceive)
```
SomberNight: The Claude CI task AFAICT correctly flagged (a preexisting) overflow issue here: `property int` is a [signed 32-bit int](https://doc.qt.io/qt-6/qml-int.html), which is too small for this usecase.
The max value is around 21 BTC with sat units, and around 21 mBTC with msat units.

accumulator: Yes this is a long-standing issue, hence the use of `QEAmount`. A short summary of the issues:
- integers in javascript context have a max range of +/- 2^54-1, which allows 21M BTC expressed in sats, but not in msats.
- QML `int` properties have a max range of +/- 2^31-1, which only allows 21 BTC expressed in sats, 21 mBTC expressed in msats (like you already indicated above)
- `int` parameters declared in the `pyqtProperty` and `pyqtSlot` decorators have a max range of 2^31-1, although this is somewhat alleviated in the QML->Python direction by using `q(u)int64`. Returning a `q(u)int64` does not work around the `int` limitation if I remember correctly (I think this was observed using Qt5, and when initially porting to Qt6)

In most of the QML code, `QEAmount` is already used for storing and passing around BTC values. The only exception is where amounts are compared (e.g. invoice amount < available balance etc) , so the `<`, `>`, `<=`, `>=` and `!=` operators, and where these operators are implied, like `Math.min()` and `Math.max()`

I'm adding patches to add these operators as `pyqtSlot`s so the operators are executed in python context. Possible caveat is that `pyqtSlot` calls might not be re-evaluated when used in a QML property binding.

electrum/wallet.py
```diff
@@ -3019,12 +3019,27 @@ def get_bolt11_invoice(self, req: Request) -> str:
...
+            amount_msat: Optional[int] = None,
+            message: Optional[str] = None,
+            exp_delay: Optional[int] = None,
+            address: Optional[str] = None
+    ):
         """ will create a lightning request if address is None """
         # for receiving
-        amount_sat = amount_sat or 0
-        assert isinstance(amount_sat, int), f"{amount_sat!r}"
-        amount_msat = None if not amount_sat else amount_sat * 1000  # amount_sat in [None, 0] implies undefined.
```
f321x, line 3056: Creating a 0 amount invoice in QML fails:
```
 78.79 | W | gui.qml.qetypes | amount_msat is undefined, returning 0
 78.79 | E | gui.qml.qeapp.Exception_Hook | exception caught by crash reporter
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qml/qewallet.py", line 706, in createRequest
    key = self.wallet.create_request(amount_msat=amount.msatsInt, message=message, exp_delay=expiration)
  File "/var/home/user/code/code_vm/electrum-2/electrum/wallet.py", line 3050, in create_request
    payment_hash = self.lnworker.create_payment_info(
        amount_msat=amount_msat,
        exp_delay=exp_delay,
        write_to_disk=False,
    )
  File "/var/home/user/code/code_vm/electrum-2/electrum/lnworker.py", line 2714, in create_payment_info
    raise ValueError("amount_msat must not be 0. Use None instead.")
ValueError: amount_msat must not be 0. Use None instead.
```

Creating an onchain address from Qt without amount fails as well:
```
  4.33 | E | gui.qt.receive_tab.ReceiveTab | Error adding payment request
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/receive_tab.py", line 328, in create_invoice
    key = self.wallet.create_request(amount_sat=amount_sat, message=message, exp_delay=expiry, address=address)
  File "/var/home/user/code/code_vm/electrum-2/electrum/wallet.py", line 3057, in create_request
    outputs = [PartialTxOutput.from_address_and_value(address, amount_sat)] if address else []
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/transaction.py", line 143, in from_address_and_value
    return cls(scriptpubkey=bitcoin.address_to_script(address),
               value=value)
  File "/var/home/user/code/code_vm/electrum-2/electrum/transaction.py", line 2094, in __init__
    TxOutput.__init__(self, *args, **kwargs)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/transaction.py", line 138, in __init__
    raise ValueError(f"bad txout value: {value!r}")
ValueError: bad txout value: None
```

electrum/gui/qt/amountedit.py
```diff
@@ -158,20 +171,35 @@ def setAmount(self, amount_sat):
         self.setFrozen(self.isFrozen()) # re-apply styling, as it is nuked by setText (?)
         self.repaint()  # macOS hack for #6269
 
+    def setMillisatPrecision(self, millisat_precision: bool):
+        if self.millisat_precision != millisat_precision:
+            self.millisat_precision = millisat_precision
+            self.setAmount(self._get_amount_from_text(self.text()))
 
-class FeerateEdit(BTCAmountEdit):
```
f321x, line 162: Calling `FeerateEdit.setAmount(None)` now raises. This can be triggered when opening a `ConfirmTxDialog` in offline mode:
```
  4.81 | D | gui.qt.send_tab.SendTab | update_fields
  4.81 | D | gui.qt.send_tab.SendTab | locking fields, r=False, a=False, m=False, d=False
  6.75 | E | gui.qt.exception_window.Exception_Hook | exception caught by crash reporter
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/send_tab.py", line 582, in do_pay_or_get_invoice
    self.do_pay_invoice(self.pending_invoice)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/send_tab.py", line 609, in do_pay_invoice
    self.pay_onchain_dialog(invoice.outputs, invoice=invoice)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/send_tab.py", line 334, in pay_onchain_dialog
    tx, is_preview, paid_with_swap = self.window.confirm_tx_dialog(
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        make_tx,
        ^^^^^^^^
    ...<2 lines>...
        batching_candidates=candidates,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/main_window.py", line 1512, in confirm_tx_dialog
    d = ConfirmTxDialog(
        window=self,
    ...<4 lines>...
        batching_candidates=batching_candidates,
    )
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/confirm_tx_dialog.py", line 1129, in __init__
    TxEditor.__init__(
    ~~~~~~~~~~~~~~~~~^
        self,
        ^^^^^
    ...<6 lines>...
        batching_candidates=batching_candidates,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/confirm_tx_dialog.py", line 122, in __init__
    self.create_fee_controls()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/confirm_tx_dialog.py", line 266, in create_fee_controls
    self.feerate_e.setAmount(self.fee_policy.fee_per_byte(self.network))
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/amountedit.py", line 124, in setAmount
    text = self._get_text_from_amount(amount)
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/amountedit.py", line 193, in _get_text_from_amount
    return super()._get_text_from_amount(quantize_feerate(amount))
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qt/amountedit.py", line 116, in _get_text_from_amount
    x = to_decimal(amount)
  File "/var/home/user/code/code_vm/electrum-2/electrum/util.py", line 244, in to_decimal
    return Decimal(str(x))
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]

```

electrum/gui/qt/main_window.py
f321x, line 971: Should `format_amount` pass `precision=3`?

Image

vs with `precision=3`:

Image

electrum/payment_identifier.py
```diff
@@ -388,15 +389,18 @@ async def _do_finalize(
             if not self.lnurl_data:
                 raise Exception("Unexpected missing LNURL data")
 
-            if not (self.lnurl_data.min_sendable_sat <= amount_sat <= self.lnurl_data.max_sendable_sat):
+            amount_msat = amount_sat * 1000
```
f321x: If `amount_sat` is coming from `QEInvoiceParser.lnurlGetInvoice()` as `Decimal(234333) / 1000 -> Decimal('234.333')` and here we
```
>>> Decimal('234.333') * 1000
Decimal('234333.000')
```
it will put `Decimal('234333.000')` into the lnurl request params and aiohttp might serialize it as '234333.000' str instead of msat int, causing an invalid lnurl reqest to be sent?

electrum/gui/qml/qefx.py
```diff
@@ -151,21 +151,18 @@ def fiatValueHistoric(self, satoshis, timestamp, plain=True):
...
             return ''
         v = fd / Decimal(rate) * COIN
-        if v.is_nan():
-            return ''
-        if plain:
-            return str(v.to_integral_value())
-        else:
-            return self.config.format_amount(v)
+        if not v.is_nan():
+            self._amount.satsInt = v.to_integral_value()
```
f321x: `to_integral_value()` returns a `Decimal` object, assigning this to `satsInt` makes `_amount._amount_msat` `Decimal` type instead of the intended `int`.

electrum/gui/qml/qetypes.py
```diff
@@ -81,42 +100,108 @@ def isMax(self, ismax):
...
+    @pyqtSlot('QVariant', result=bool)
+    def lte(self, other: 'QEAmount|None') -> bool:
+        return self.lt(other) or self == other
+
+    @pyqtSlot('QVariant', result=bool)
+    def gt(self, other: 'QEAmount|None') -> bool:
+        if other is None:
+            other = QEAmount()
+        assert isinstance(other, QEAmount)
+        if self.isMax or other.isMax:
```
f321x: The `isMax` checks here seem off.
MAX.gt(1) -> False

also
max(MAX, 1) -> 1

f321x:
I looked through the PR, but the issues above were all found by Claude. For those i verified (point 1 - 6) i wrote the comments above. I didn't verify the remaining points but considering the first were all real it makes sense to look at the remaining ones as well (8. -12.). 

<details>
<summary>This is the output:</summary>
## 🔴 High — confirmed crashes / broken core flows

### 1. `create_request` dropped its amountless-coercion — amountless requests break three ways

Root cause at `electrum/wallet.py:3023`: the old `amount_sat = amount_sat or 0` +
`amount_msat = None if not amount_sat else …` is gone, so an empty amount no longer maps
to "amountless".

- **(a) crash — QML amountless *Lightning* request.** `electrum/gui/qml/qewallet.py:706`
  passes `amount_msat=amount.msatsInt` (= `0` for an empty `QEAmount`). `create_request`
  keeps `amount_msat=0`, so `lnworker.create_payment_info` raises
  `ValueError("amount_msat must not be 0…")` at `electrum/lnworker.py:2714`. `createRequest`
  only catches `InvoiceError` (`qewallet.py:709`) → unhandled exception. This is the standard
  "donation / request-without-amount" flow and it's reachable — with inbound liquidity,
  `lightningCanReceive.gt(empty)` is `true`, so the Lightning button is enabled. Master
  produced an amountless invoice here.
- **(b) broken — Qt on-chain amountless request.** `receive_tab.py` empty field →
  `get_amount()` returns `None` → `create_request(amount_sat=None, address=addr)` →
  `wallet.py:3057` `PartialTxOutput.from_address_and_value(addr, None)` →
  `ValueError: bad txout value: None` (reproduced). Caught by the generic `except Exception`
  and shown as "Error adding payment request"; the request is never created (worked on
  master). Same crash, **uncaught**, via the `add_request` RPC with no amount
  (`satoshis(None)=None`, `commands.py:1383`).
- **(c) wrong — QML on-chain amountless request.** `qewallet.py:708` passes `amount_sat=0`
  → `amount_msat=0` → `Request.get_amount_sat()` returns `0` (not `None`) → a fixed
  **0-amount** request instead of "any amount".

**Fix:** re-introduce the coercion in `create_request` (treat `0`/`None` as amountless:
`amount_msat=None`, on-chain output value `0`).

### 2. `FeerateEdit.setAmount(None)` crashes

`electrum/gui/qt/amountedit.py:180` — `FeerateEdit` now subclasses `AmountEdit` instead of
`BTCAmountEdit`, losing `BTCAmountEdit.setAmount`'s `if amount is None: setText(" ")` guard
(`amountedit.py:165`). `AmountEdit.setAmount(None)` → `to_decimal(None)` → `Decimal('None')`
→ `InvalidOperation`. `confirm_tx_dialog.py:355` and `:436` call
`self.feerate_e.setAmount(None)` whenever there's no fee estimate / not enough funds — a
common path. (`fee_e` is fine; it's still a `BTCAmountEdit`.)

---

## 🟠 Medium — confirmed wrong output / wire format

### 3. Qt desktop lists & notification mis-render sub-sat amounts

`invoice_list.py:119`, `request_list.py:149`, `main_window.py:1375` (the "Payment received"
notification) and `:1690` (invoice dialog) now read `get_amount_sat_msat_precision()` (which
returns a `Decimal` for sub-sat amounts) but call `format_amount(...)` **without
`precision=3`**. Verified: `format_amount(Decimal('1.5'))` → rounds to 2 sat (`'0.00002'`),
and `format_amount(Decimal('0.4'))` → the malformed string `'0.'`. So the msat-precision
getter was wired up but the precision is dropped exactly where it should be shown (and
< 0.5 sat renders garbage). Pass `precision=3` (as the QML side does).

### 4. LNURL-pay sends a non-integer `amount` on the wire

`qeinvoice.py:609` computes `amount = Decimal(msatsInt)/1000`, then
`payment_identifier.py:392` does `amount_msat = amount_sat * 1000` → `Decimal('2345.000')`,
sent as `params={'amount': …}` (`:403`). aiohttp serializes that to `?amount=2345.000`
(trailing zeros). Since `BtcField` here has `msatPrecision: true`, any sub-sat amount (or an
LNURL whose `min_sendable_msat` isn't a whole sat) triggers it, and strict servers doing
`int(amount)` reject it. Master sent integer msat. **Fix:** `params={'amount': int(amount_msat)}`.

### 5. `qefx.satoshiValue` stores a `Decimal` in `_amount_msat`

`qefx.py:164` `self._amount.satsInt = v.to_integral_value()` (a `Decimal`); the `satsInt`
setter multiplies by 1000 and stores the `Decimal`, so `_amount_msat` is no longer an `int`
— violating the invariant the constructor asserts (`assert isinstance(amount_msat, int)`) and
the `'qint64'` property type. Wrap in `int(...)`.

---

## 🟡 Low / latent / robustness

6. **`QEAmount.max`/`min`/`lt`/`gt` `is_max` semantics are wrong** (`qetypes.py:177`):
   `gt()`/`lt()` return `False` whenever *either* operand is `is_max`, so `max(MAX, normal)`
   returns `normal` (loses MAX) and the result is order-dependent. Latent — no current caller
   passes an `is_max` amount — but the helper is incorrect for future use.
7. **`_msat_to_sat` truncates toward zero, not floor** (`qetypes.py:54`,
   `int(Decimal(msat)/1000)`): inconsistent with `// 1000` used everywhere else in the
   codebase; off-by-1-sat for negative sub-sat LN amounts (reachable via
   `qetransactionlistmodel.py` where `amount_msat < 0`).
8. **`__eq__` against an `int` silently changed from sat to msat** and the `str` branch was
   removed (`qetypes.py:192`). No broken caller found (comparisons are all against `0`), but
   `amount == <nonzero int>` now means msat — a footgun worth a comment or an explicit
   `msat`-named helper.
9. **`assert` used to validate untrusted input**: `lnurl.py:188` validates a remote server's
   min/max withdrawable with `assert`, and the `QEAmount` constructor asserts its args — both
   stripped under `python -O`, letting malformed LNURL ranges / bad amounts through.
   (Pre-existing pattern, but now on the canonical msat path.)
10. **`formatMilliSatsForEditing` lacks the `None` guard** its siblings `formatSats` /
    `formatMilliSats` have (`qeconfig.py:364`).
11. **`positive` uses `>= 0` (was `> 0`)** (`qetypes.py:110`): a zero-value entry is labeled
    received/credit in `LightningPaymentDetails.qml:55` / `TxDetails.qml:110` (mostly hidden
    by visibility guards).
12. **`min`/`max` return shared references** → `effectiveMaxWithdrawable` can alias the live
    wallet `QEAmount` in `LnurlWithdrawRequestDialog.qml`; self-corrected by `updateLimits()`
    but fragile. Minor: `ChannelBar.qml` switched to `.satsInt`, truncating msat in bar-width
    math (cosmetic).

---
</details>

accumulator:
> I looked through the PR, but the issues above were all found by Claude. For those i verified (point 1 - 6) i wrote the comments above. I didn't verify the remaining points but considering the first were all real it makes sense to look at the remaining ones as well (8. -12.).

Thanks for reviewing. I missed these Claude findings due to the green checkmark, and cirrus-ci is by now unavailable..

f321x:
> I missed these Claude findings due to the green checkmark

The results i commented were from a local run, not sure what the CI said, but yeah it often finds some things even though the checkmark is green, so makes sense to look at the output anyway.
Btw, let me know if you want an API key as well.

electrum/gui/qt/amountedit.py
```diff
@@ -159,19 +165,17 @@ def setAmount(self, amount_sat):
         self.repaint()  # macOS hack for #6269
 
 
-class FeerateEdit(BTCAmountEdit):
-
-    def __init__(self, decimal_point, is_int=False, parent=None, *, max_amount=_NOT_GIVEN):
-        super().__init__(decimal_point, is_int, parent, max_amount=max_amount)
+class FeerateEdit(AmountEdit):
+    def __init__(self, parent=None, *, max_amount=None):
+        super().__init__(lambda: self._base_unit(), False, parent, max_amount=max_amount)
```
SomberNight: The old base class, BTCAmountEdit, was setting a reasonable max_amount, 21 million BTC per vbyte. No one should want to pay a higher feerate than that. :)

Without having some upper bound, decimals will "overflow" and weird exceptions follow.

Image

```
 15.19 | E | gui.qt.exception_window.Exception_Hook | exception caught by crash reporter
  File "/home/user/wspace/electrum/electrum/gui/qt/confirm_tx_dialog.py", line 203, in timer_actions
    self.update()
    ~~~~~~~~~~~^^
  File "/home/user/wspace/electrum/electrum/gui/qt/confirm_tx_dialog.py", line 207, in update
    self.update_tx()
    ~~~~~~~~~~~~~~^^
  File "/home/user/wspace/electrum/electrum/gui/qt/confirm_tx_dialog.py", line 1155, in update_tx
    self.fee_policy = fee_policy = self.get_fee_policy()
                                   ~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/wspace/electrum/electrum/gui/qt/confirm_tx_dialog.py", line 394, in get_fee_policy
    feerate = self.feerate_e.get_amount()
  File "/home/user/wspace/electrum/electrum/gui/qt/amountedit.py", line 107, in get_amount
    amt = self._get_amount_from_text(str(self.text()))
  File "/home/user/wspace/electrum/electrum/gui/qt/amountedit.py", line 190, in _get_amount_from_text
    return quantize_feerate(super()._get_amount_from_text(text))
  File "/home/user/wspace/electrum/electrum/util.py", line 886, in quantize_feerate
    return Decimal(fee).quantize(_feerate_quanta, rounding=decimal.ROUND_HALF_DOWN)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]
```

SomberNight (commented):
I just looked at the first commit.
Please don't think just because the genai finds no concerns, there are no bugs.

accumulator:
> Btw, let me know if you want an API key as well.

Hmm, don't mind if I do :)

electrum/gui/qt/main_window.py
accumulator: Well, 5891e039b1db74c0eed14fbf9b9f4e61d9f3546c introduced a config option whether to show msat precision or not. `format_amount` just takes this into account.

QML bypasses this by using `Config.formatMilliSats` where msat precision is used, which explicitly passes the requested msat precision.

I'm fine with reverting 5891e039b1db74c0eed14fbf9b9f4e61d9f3546c though, I think for bitcoin amounts you almost always want to see full precision..

electrum/gui/qt/amountedit.py
accumulator: now clamped to 21M BTC, the reasonable value, like before :)

accumulator:
rebased on master, commits reorganised for better reviewability

electrum/gui/qt/main_window.py
f321x: Ah i didn't consider we have an option for this. Then i think it is fine as is, I assume msat could be considered useless by some users so it seems good to have the option to not show them.

f321x:
This would also benefit from using the new `amount_msat` kwarg:
https://github.com/spesmilo/electrum/blob/32f2c1875ee4cfd57930ea83c5da856c21191f44/electrum/plugins/nwc/nwcserver.py#L463-L466

electrum/gui/qml/qefx.py
```diff
@@ -151,21 +151,18 @@ def fiatValueHistoric(self, satoshis, timestamp, plain=True):
         else:
             return self.fx.historical_value_str(satoshis, dt)
 
-    @pyqtSlot(str, result=str)
-    @pyqtSlot(str, bool, result=str)
-    def satoshiValue(self, fiat, plain=True):
+    @pyqtSlot(str, result='QVariant')
+    def satoshiValue(self, fiat):
+        self._amount = QEAmount()
```
f321x, line 156: Why is `_amount` declared as instance attribute here, isn't it just used in the scope of this method?

electrum/gui/qml/qefx.py
```diff
@@ -151,21 +151,18 @@ def fiatValueHistoric(self, satoshis, timestamp, plain=True):
...
         rate = self.fx.exchange_rate()
         try:
             fd = Decimal(fiat)
         except Exception:
             return ''
```
f321x: Returning a `str` here violates the specified `QVariant` result type, e.g. when entering `.` as fiat amount:
```
 11.04 | W | gui.qml.qeapp | "Could not convert argument 0 at"
 11.04 | W | gui.qml.qeapp | 	 "expression for onTextChanged@file:///var/home/user/code/code_vm/electrum/electrum/gui/qml/components/controls/FiatField.qml:22"
 11.04 | W | gui.qml.qeapp | file:///var/home/user/code/code_vm/electrum/electrum/gui/qml/components/controls/FiatField.qml:22: TypeError: Passing incompatible arguments to C++ functions from JavaScript is not allowed.
```

electrum/gui/qml/components/controls/FormattedAmount.qml
```diff
@@ -26,7 +26,7 @@ GridLayout {
     }
     Label {
         visible: valid
-        text: amount.msatsInt != 0 ? Config.formatMilliSats(amount) : Config.formatSats(amount)
+        text: Config.formatMilliSats(amount)
```
f321x, line 30: This passes `None` to `Config.formatMilliSats()` and raises an Exception, e.g. when creating a lightning invoice:
```
  3.70 | I | lnworker.LNWallet.[default_wallet] | creating bolt11 invoice with routing_hints: [('r', [('02c3be1bcedffe1675617fc7ae9e4c43ce0f54cc3e58c48ef51398dc87a78e0ce0', '10052759x15511793x59990', 1000, 1, 144)]), ('r', [('02c3be1bcedffe1675617fc7ae9e4c43ce0f54cc3e58c48ef51398dc87a78e0ce0', '8427041x8597475x4058', 1000, 1, 144)]), ('r', [('02465ed5be53d04fde66c9418ff14a5f2267723810176c9212b722e542dc1afb1b', '16000000x0x18483', 1000, 1, 80)])], sat: 1898
  3.71 | E | gui.qml.qeapp.Exception_Hook | exception caught by crash reporter
  File "/var/home/user/code/code_vm/electrum/electrum/gui/qml/qeconfig.py", line 400, in formatMilliSats
    raise Exception(f"Unknown amount type: {str(type(amount))}")
Exception: Unknown amount type: <class 'NoneType'>
  3.73 | D | gui.qml.qerequestdetails | key=f3cd542ad64f230e65dfe4e7312d3a00394a3ce651ed4a5ab1c8072535c10061

```

electrum/gui/qt/amountedit.py
```diff
@@ -106,22 +107,32 @@ def get_amount(self) -> Union[None, Decimal, int]:
...
+        if amt is None:
+            return amt
+        # return as int if no millisats (Decimal otherwise)
+        return int(amt) if (int(amt) * 1000 == int(amt * 1000)) else amt
 
     def _get_text_from_amount(self, amount) -> str:
-        return "%d" % amount
+        x = to_decimal(amount)
+        scale_factor = pow(10, self.decimal_point())
+        nfmt = "{:." + str(self.decimal_point() + self.extra_precision()) + "f}"
```
f321x: When calling `.setAmount()` on a `FeerateEdit` with a `0` extra_precision currency like `KRW` this swallows trailing 0, though nothing seems to be affected by this:
```
>>> from electrum.gui.qt import amountedit
>>> fiat = amountedit.FiatAmountEdit(fx=daemon.fx)  # fx is KRW
>>> fiat.setAmount(100000)
>>> fiat.get_amount()
1
```

electrum/gui/qml/qetypes.py
```diff
@@ -20,32 +21,50 @@ class QEAmount(QObject):
 
     _logger = get_logger(__name__)
 
-    def __init__(self, *, amount_sat: int = 0, amount_msat: int = 0, is_max: bool = False, from_invoice=None, parent=None):
+    valueChanged = pyqtSignal()
+
+    def __init__(self, *, amount_sat: int = None, amount_msat: int = None, is_max: bool = False, from_invoice=None, parent=None):
```
f321x: The default argument change from `0` to `None` causes the log in `satsInt()` and `msatsInt()` to spam a bit when using default `QEAmount()`. Maybe the log could be removed as it is to be expected that the amount is `None` sometimes?
 
E.g. when opening a swap view:
```
 24.86 | D | gui.qml.qeswaphelper | run_swap_manager
 24.87 | I | submarine_swaps.NostrTransport | starting nostr transport with pubkey: 7992c2f29bf56da19b032e2aca6f785e8413058dd57f1c1482aa2486c4531299
 24.87 | I | submarine_swaps.NostrTransport | nostr relays: ['wss://nostr.einundzwanzig.space', 'wss://relay.primal.net', 'wss://relay.damus.io', 'wss://eu.purplerelay.com', 'wss://ftp.halifax.rwth-aachen.de/nostr', 'wss://relay.getalby.com/v1', 'wss://nostr.mom', 'wss://brb.io', 'wss://nos.lol']
 24.87 | W | gui.qml.qetypes | amount_msat is undefined, returning 0
 24.87 | W | gui.qml.qetypes | amount_msat is undefined, returning 0
 24.87 | W | gui.qml.qetypes | amount_msat is undefined, returning 0
 24.87 | W | gui.qml.qetypes | amount_msat is undefined, returning 0
```

electrum/gui/qml/components/ReceiveDetailsDialog.qml
```diff
@@ -103,22 +104,23 @@ ElDialog {
...
                 icon.source: '../../icons/bitcoin.png'
+                enabled: !amountBtc.textAsSats.hasMsatPrecision
                 onClicked: { dialog.isLightning = false; doAccept() }
             }
             FlatButton {
                 Layout.fillWidth: true
                 Layout.preferredWidth: 1
-                enabled: Daemon.currentWallet.isLightning && (Daemon.currentWallet.lightningCanReceive.satsInt
-                            > amountBtc.textAsSats.satsInt || Daemon.currentWallet.canGetZeroconfChannel)
+                enabled: Daemon.currentWallet.isLightning && (Daemon.currentWallet.lightningCanReceive.msatsInt
```
f321x: The `Lightning` button doesn't react to user input anymore, when entering an amount exceeding `lightningCanReceive` the button stays enabled.

electrum/gui/qt/send_tab.py
f321x, line 467: This behavior seems to have regressed, when entering `!` first and then pasting an address the amount turns empty instead of resolving max.

electrum/gui/qml/qefx.py
accumulator: It is returned to QML. The store as instance attribute is to avoid gc as soon as the method returns.

f321x: ahh good to know 👍

electrum/gui/qml/qefx.py
accumulator: > Returning a `str` here violates the specified `QVariant` result type, e.g. when entering `.` as fiat amount:

Actually, it is allowed to return a `str` here. It will be auto wrapped as `QVariant(QString)`. The reason for the error is the `str` getting passed to `Config.amountToBaseunitStr()`.

Nevertheless, I changed the method to always return a `QEAmount` (containing zero on exception), and `FiatField` now sets an empty text on `BtcField` if zero amount. (5a8f908f748beb7c484c5e0e63e2d2e8adb1b08c)

electrum/gui/qml/components/controls/FormattedAmount.qml
accumulator: The `x ? y : z` conditional likely hid this specific issue, due to `amount` being `null` for a moment.

da4c1970620bb63b2eb2170073ebc0ad56dbac0f

electrum/gui/qml/components/ReceiveDetailsDialog.qml
accumulator: 3def298be710297dc8837a9e02e7d6f978c415f5

electrum/gui/qml/qetypes.py
accumulator: yeah, fair point

afc1f00e9ab9b851e301fc563c0c5ccf6ba8587e

accumulator:
> This would also benefit from using the new `amount_msat` kwarg:

missed that one

e0629cd0b634e61dae3c0faa84f163321e1f03eb

electrum/gui/qt/send_tab.py
accumulator: 61eb3519a1daa3894f671942d5602b7dd60b7641

electrum/gui/qt/amountedit.py
accumulator: good catch. fab754df5bbcb0c71673f0798e42032f4884c3f7

electrum/gui/qt/amountedit.py
```diff
@@ -106,22 +107,32 @@ def get_amount(self) -> Union[None, Decimal, int]:
         amt = self._get_amount_from_text(str(self.text()))
         if self.max_amount and amt and amt >= self.max_amount:
             return self.max_amount
-        return amt
+        if amt is None:
+            return amt
+        # return as int if no millisats (Decimal otherwise)
+        return int(amt) if (int(amt) * 1000 == int(amt * 1000)) else amt
```
f321x: This only preserves fractional parts to 10^-3, when e.g. using a `FiatAmountEdit` with a currency requiring precision >3 this loses information. 

```suggestion
        return int(amt) if amt == int(amt) else amt
```

E.g. XAU currency (0.0002 == 1000sat), gets rounded to 0 sat:

Image

accumulator: ah yes, this was a BTC-ism left in the generic super

electrum/gui/qml/components/LnurlWithdrawRequestDialog.qml
```diff
@@ -13,41 +13,62 @@ ElDialog {
...
-    property bool liquidityWarning: providerMaxWithdrawable > walletCanReceive
-
-    property bool amountValid: !dialog.insufficientLiquidity &&
-        amountBtc.textAsSats.satsInt >= dialog.effectiveMinWithdrawable &&
-        amountBtc.textAsSats.satsInt <= dialog.effectiveMaxWithdrawable
+    property var walletCanReceive: Amount {}
+    property var providerMinWithdrawable: requestDetails.lnurlData['min_withdrawable_msat']
+    property var providerMaxWithdrawable: requestDetails.lnurlData['max_withdrawable_msat']
+    property var effectiveMinWithdrawable: onemsat.max(providerMinWithdrawable, onemsat)
+    property var effectiveMaxWithdrawable: onemsat.min(providerMaxWithdrawable, requestDetails.wallet.lightningCanReceive)
```
f321x: I think when `requestDetails.wallet.lightningCanReceive` is passed to `QEAmount.min()` here and is smaller than `providerMaxWithdrawable` the ownership of the min withdrawable `QEAmount` changes from `QEWallet._lightningcanreceive` to a standalone javascript object when being returned by the `min()` slot and is destroyed by the QML GC causing future access of `wallet.lightningCanReceive` to crash: 

```
 37.01 | D | gui.qml.qechannellistmodel | init_model
 37.01 | E | gui.qml.qeapp.Exception_Hook | exception caught by crash reporter
TypeError: unable to convert a Python 'QEAmount' object to a C++ 'QObject*' instance
 37.03 | E | gui.qml.qeapp.Exception_Hook | exception caught by crash reporter
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/qml/qeconfig.py", line 400, in formatMilliSats
    raise Exception(f"Unknown amount type: {str(type(amount))}")
Exception: Unknown amount type: <class 'NoneType'>
```

I can reproduce this by:
1. Opening a lnurl withdrawal dialog with `maxWithdrawable` > `lightningCanReceive`.
2. Closing it again
3. `lightningCanReceive` gets GC'd
4. Opening channel view, tries to access `lightningCanReceive`, crashes

This seems like a subtle footgun

electrum/gui/qt/send_tab.py
```diff
@@ -896,12 +902,12 @@ def request_lnurl_withdraw_dialog(self, lnurl_data: LNURL3Data):
...
+        min_amount = max(Decimal(lnurl_data.min_withdrawable_msat) / 1000, 1)
         max_amount = min(
-            lnurl_data.max_withdrawable_sat,
-            int(self.wallet.lnworker.num_sats_can_receive())
+            Decimal(lnurl_data.max_withdrawable_msat) / 1000,
+            self.wallet.lnworker.num_sats_can_receive()
         )
-        min_text = self.format_amount_and_units(lnurl_data.min_withdrawable_sat)
+        min_text = self.format_amount_and_units(Decimal(lnurl_data.min_withdrawable_msat) / 1000)
         if min_amount > int(self.wallet.lnworker.num_sats_can_receive()):
```
f321x:
```suggestion
        if min_amount > self.wallet.lnworker.num_sats_can_receive():
```

This would prevent a liquidity warning for:
```
>>> Decimal("10.5") > int(Decimal("10.9"))
True
```

electrum/gui/qml/components/LnurlWithdrawRequestDialog.qml
accumulator: sigh. python + QML is a gift that keeps on giving..

good find. 

I made a patch that returns a copy parented to its creating instance 

7f6643945afe9525997268ddbc044aaf26512fa7

(parenting `lightningCanReceive` to its `QEWallet` would also have worked, as that would keep its ownership in the `QObject` tree instead of JS, but this is IMO more robust)

electrum/gui/qt/send_tab.py
accumulator: c977ca4cdbd00352c1f6981f1bde3d74f982068c

electrum/gui/qml/qetypes.py
```diff
@@ -16,107 +17,221 @@ class QEAmount(QObject):
...
+    def lte(self, other: 'QEAmount|None') -> bool:
+        return self.lt(other) or self == other
+
+    @pyqtSlot('QVariant', result=bool)
+    def gt(self, other: 'QEAmount|None') -> bool:
+        if other is None:
+            other = QEAmount()
+        assert isinstance(other, QEAmount)
+        assert not (self.isMax or other.isMax), "'gt/gte' operator undefined for MAX amounts"
+        if self.isEmpty and not other.isEmpty:
```
f321x: Are the `isEmpty` checks in `lt()` and `gt()` necessary? 
The comparison below the check uses `self.msatsInt` which returns `0` if `self._amount_msat = None`, so this would be handled?
At the same time it produces incorrect comparisons with negative values:
```
`QEAmount(amount_msat=0).lt(QEAmount(amount_msat=-100)) -> True
`QEAmount(amount_msat=0).gt(QEAmount(amount_msat=-100)) -> False
```

electrum/gui/qml/qetypes.py
```diff
@@ -196,14 +196,27 @@ def min(self, one: 'QEAmount|None', two: 'QEAmount|None'):
...
         # - define MAX as always being larger than any value
         if one.isMax:
-            return two
+            return self.copy(two)
         if two.isMax:
-            return one
-        return one if one.lt(two) else two
+            return self.copy(one)
+
+        return self.copy(one if one.gt(two) else two)
```
f321x:
```suggestion
        return self.copy(one if one.lt(two) else two)
```

electrum/payment_identifier.py
```diff
@@ -507,7 +511,10 @@ def parse_amount(self, x: str) -> Union[str, int]:
             return x
         p = pow(10, self.config.BTC_AMOUNTS_DECIMAL_POINT)
         try:
-            return int(p * Decimal(x))
+            x = Decimal(x) * p
+            if int(x) != x:
+                raise InvalidOperation("no millisat precision allowed")
+            return int(x)
         except InvalidOperation:
             raise Exception("Invalid amount")
```
f321x, line 519: Maybe you could raise `Exception` directly here so the user gets displayed a more detailed error reason?

```suggestion
            if int(x) != x:
                raise Exception(_("no millisat precision allowed"))
            return int(x)
        except InvalidOperation:
            raise Exception(_("Invalid amount"))
```

accumulator: This is more for developers, users should never see this as the frontend should already filter for precision

f321x: Ah I didn't spot that pre-filtering.

electrum/gui/qt/amountedit.py
```diff
@@ -158,20 +178,39 @@ def setAmount(self, amount_sat):
...
 
     def _base_unit(self):
         return UI_UNIT_NAME_FEERATE_SAT_PER_VBYTE
 
     def _get_amount_from_text(self, text):
-        sat_per_byte_amount = super()._get_amount_from_text(text)
-        return quantize_feerate(sat_per_byte_amount)
+        amount = super()._get_amount_from_text(text)
+        if amount and amount > TOTAL_COIN_SUPPLY_LIMIT_IN_BTC * COIN:
+            return TOTAL_COIN_SUPPLY_LIMIT_IN_BTC * COIN
```
f321x, line 208: Instead of this clamp, would it be better to simply pass a `max_amount=TOTAL_COIN_SUPPLY_LIMIT_IN_BTC * COIN` to the `FeerateEdit.__init()`?
Currently the edit field text doesn't get capped if the user enters a large amount (unlike on master).

Arguably it might be even nicer to use a `max_amount` value/constant that defines a sane limit for transaction fees, maybe: https://github.com/spesmilo/electrum/blob/8c0adcda009474cbd2482cd94cbf1e26b636bee5/electrum/fee_policy.py#L23-L24

Image

SomberNight: > Arguably it might be even nicer to use a max_amount value/constant that defines a sane limit for transaction fees

The 1500 sat/byte max cap is for dynamic fee rates coming from servers. I think users should be allowed to pay higher fees if they want. (though the GUI shows warnings already above 600 sat/byte IIRC)

For example, around block 840k, so many people were trying to get txs included in the halving block that next-block feerates were around 2k sat/byte for a few hours. I think it's good that the dynamic limits are reasonably capped and it's fine that in extreme cases the limit can even be hit, but a manual limit should never not be enough.

That is, I think we should not cap manually input fees at these levels. If we really want a cap, it should be like 50k sat/byte, maybe even higher.

Also, what if I want to burn 1 testnet BTC as mining fees? :P

accumulator:
@f321x all findings should be covered now

accumulator:
If there are no further objections, I propose to merge this.

electrum/gui/qml/qeinvoice.py
f321x, line 295: `set_effective_invoice` doesn't clear `self._amountOverride`, this can cause paying a wrong, stale amount.
When e.g. saving an amountless ln invoice and an onchain invoice for 1k sats, then opening the saved ln invoice, entering 21k sat, going back and then opening the onchain invoice it will try to pay 21k sat instead of 1k sat.

electrum/gui/qml/components/LnurlPayRequestDialog.qml
```diff
@@ -74,12 +82,14 @@ ElDialog {
...
-                    enabled: invoiceParser.lnurlData['min_sendable_sat'] != invoiceParser.lnurlData['max_sendable_sat']
+                    text: Config.formatMilliSatsForEditing(invoiceParser.lnurlData['min_sendable_msat'])
+                    enabled: !invoiceParser.lnurlData['min_sendable_msat'].eq(invoiceParser.lnurlData['max_sendable_msat'])
                     color: Material.foreground // override gray-out on disabled
                     fiatfield: amountFiat
-                    onTextAsSatsChanged: {
+                    msatPrecision: true
+                    onValueChanged: {
                         invoiceParser.amountOverride = textAsSats
+                        dialog.amountValid = isValidAmount()
```
f321x, line 99: When scanning a fixed amount LNURL-pay before the channels are reestablished (e.g. when launching the app through an intent) the new `lightningCanSend` check in `isValidAmount()` will be false and then never update (as the value never changes), preventing it from being paid. There probably should be some connection updating it? 
Or we rely on the subsequent invoice view as it gives the user feedback on why they can't send instead of just disabling the button?

accumulator: Good catch, I actually had this left on my TODO list to bring in sync with LNURLw but it got buried :)

electrum/gui/qml/qeinvoice.py
accumulator: This actually accidentally worked on master, because before this `QEAmount` refactoring it glitched in a null value :)

Resetting `amountOverride` is a fix, but maybe we should tighten it further and ignore `amountOverride` entirely if the invoice contains an amount..

electrum/gui/qt/amountedit.py
f321x: >  Currently the edit field text doesn't get capped if the user enters a large amount (unlike on master).

Maybe a NIT but this still isn't clamped (in the UI), it seems nice to clamp it somehow like we already do on master.
Couldn't we move the `_NOT_GIVEN` sentinel check into the base class and also use it as default value for the `FeerateEdit` like this:
```diff
diff --git a/electrum/gui/qt/amountedit.py b/electrum/gui/qt/amountedit.py
index e5a4eb6660..a1e150d353 100644
--- a/electrum/gui/qt/amountedit.py
+++ b/electrum/gui/qt/amountedit.py
@@ -46,6 +46,8 @@ class AmountEdit(SizedFreezableLineEdit):
     shortcut = pyqtSignal()
 
     def __init__(self, parent=None, *, base_unit=None, is_int=False, max_amount=None, extra_precision=None):
+        if max_amount is _NOT_GIVEN:
+            max_amount = TOTAL_COIN_SUPPLY_LIMIT_IN_BTC * COIN
         # This seems sufficient for hundred-BTC amounts with 8 decimals
         width = 16 * char_width_in_lineedit()
         super().__init__(width=width, parent=parent)
@@ -134,8 +136,6 @@ class AmountEdit(SizedFreezableLineEdit):
 class BTCAmountEdit(AmountEdit):
 
     def __init__(self, decimal_point, parent=None, *, is_int=False, max_amount=_NOT_GIVEN, millisat_precision=False):
-        if max_amount is _NOT_GIVEN:
-            max_amount = TOTAL_COIN_SUPPLY_LIMIT_IN_BTC * COIN
         self.millisat_precision = millisat_precision
         AmountEdit.__init__(self, parent, base_unit=self._base_unit, is_int=is_int, max_amount=max_amount, extra_precision=self.extra_precision)
         self.decimal_point = decimal_point
@@ -186,7 +186,7 @@ class BTCAmountEdit(AmountEdit):
 
 
 class FeerateEdit(AmountEdit):
-    def __init__(self, parent=None, *, max_amount=None):
+    def __init__(self, parent=None, *, max_amount=_NOT_GIVEN):
         super().__init__(parent, base_unit=lambda: self._base_unit(), max_amount=max_amount)
         self.extra_precision = lambda: FEERATE_PRECISION
         self.decimal_point = lambda: 0
```

accumulator: > Couldn't we move the `_NOT_GIVEN` sentinel check into the base class and also use it as default value for the `FeerateEdit` like this:

I don't think the BTC supply limit is relevant to the generic `AmountEdit`, although some sort of cap seems reasonable. 

For feerate I added a cap at 100k.

see 9fb77e2e2d7b4fa5fd9baad3d4a8c79b15f188c2

electrum/gui/qt/amountedit.py
```diff
@@ -10,11 +10,16 @@
...
 
 from electrum.util import (format_satoshis_plain, decimal_point_to_base_unit_name,
-                           FEERATE_PRECISION, quantize_feerate, DECIMAL_POINT, UI_UNIT_NAME_FEERATE_SAT_PER_VBYTE)
+                           FEERATE_PRECISION, quantize_feerate, DECIMAL_POINT, UI_UNIT_NAME_FEERATE_SAT_PER_VBYTE,
+                           to_decimal)
 from electrum.bitcoin import COIN, TOTAL_COIN_SUPPLY_LIMIT_IN_BTC
 
 _NOT_GIVEN = object()  # sentinel value
 
+MAX_FEERATE = 100_000
```
SomberNight, line 19: note: the qml GUI lacks a similar limit. In "manual" fee edit mode, I can enter lots of "9"s and get a `decimal.InvalidOperation` exception (though it's caught).

I also wonder whether it would be better to put this limit in `fee_policy.py` near the other related constants. (esp if it's shared between the GUIs)

If you leave it here, please document the unit in a comment.
e.g. `MAX_FEERATE = 100_000  # sat/vbyte`

SomberNight:
> If there are no further objections, I propose to merge this.

All right. (I have not reviewed it.)
f321x has been looking at this, so he is your best bet for an ACK.

electrum/gui/qml/components/ReceiveDetailsDialog.qml
f321x, line 75: When entering a large fiat value here it will raise an overflow exception:
```
 24.66 | E | gui.qml.qeapp.Exception_Hook | exception caught by crash reporter
TypeError: unable to convert a Python 'int' object to a C++ 'qint64' instance
```

ecdsa:
> All right. (I have not reviewed it.)

4.8.2 should be a security release. I think it would be safer to delay new features, unless you want to cherry-pick and release from a branch.

f321x:
On QML there is no option to disable msat precision (like on Qt). I don't think we should show the balance and history items in msat precision, this doesn't provide value in real world usage and is confusing. 
It seems better to just show it in detailed views, e.g. when viewing the details of a specific transaction?
image

accumulator:
> On QML there is no option to disable msat precision (like on Qt). I don't think we should show the balance and history items in msat precision, this doesn't provide value and is confusing. It seems better to just show it in detailed views, e.g. when viewing the details of a specific transaction? 

I agree, and I also think the config option should go away in the desktop client.

Instead we should IMO only show sub-sat precision in 
- lightning invoice details/request details
- transaction details

Everywhere else, including history list and invoice/request lists it should be sat precision. Maybe one special-case here, e.g. <1 sat amount invoices/requests in list.

PR #10588, opened 2026-04-18
qml: allow entering key-origin info when importing a master key -> basic hardware wallet support

aydencook03:
When creating a watch-only wallet on mobile by pasting an xpub, Electrum calls root_fp_and_der_prefix_from_xkey() to recover the derivation path and master fingerprint. This works fine for depth-0 and depth-1 keys, but hardware wallets typically export at deeper paths (a Coldcard zpub, for example, is at m/84'/0'/0', depth 3). For keys at depth > 1, neither the master fingerprint nor the full derivation path can be read from the xpub, so they end up stored as null.

This matters because without key-origin metadata in the keystore, the PSBTs Electrum generates won't include it either, and hardware wallets that verify key-origin will reject signing.

This PR adds a key-origin section to the WCHaveMasterKey wizard screen in the QML GUI. When the user pastes a key that Electrum can't infer origin info for (depth > 1), two optional fields appear (derivation path and master fingerprint) so the user can supply them manually. The section is hidden entirely for depth <= 1 since Electrum already handles those automatically.

On the backend, a new verifyKeyOriginInfo slot does basic client-side validation (path structure and depth consistency with the xpub), and create_storage() in wizard.py passes the values through to keystore.add_key_origin() if provided. A masterKeyDepth slot is also exposed so the QML layer can check depth without duplicating the bip32 logic.

This is the minimum needed to support a basic hardware wallet signing loop on mobile: import an xpub, provide the derivation path and fingerprint shown on the device, and the resulting PSBTs will carry the key-origin data the device expects.


aydencook03:
new_UI

aydencook03:
The two fields are completely optional, and the default behavior is preserved if they're left blank. Also, they are only visible if the key that has been entered has depth > 1 and electrum can't infer the origin data.

aydencook03:
This is a simple solution until we're able to import skeleton wallet files into the mobile app.

aydencook03:
Are there any changes that I need to make?

PR #10110, opened 2025-08-07
initial support bolt12 offers

accumulator:
This PR contains initial support for BOLT12 
- parses bolt12 `offer` payment identifiers, 
- requests bolt12 `invoice` by sending `invoice_request` via onion message
- handles `invoice` and `invoice_error` responses
- initial support for paying bolt12 `invoice`
    - currently 'abusing' `bolt11` data structures like `LnAddr`, to minimize changes in payment sending code path
    - bolt12 `invoice` currently stored in `Invoice.lightning_invoice` with a prefix, because adding a new field to an `@attr.s` object is a PITA when switching branches, final storage location TBD.
- CLI commands `add_offer`, `get_offer`, `delete_offer` and `list_offers` for managing offers.
- handles incoming `invoice_request` and replies with `invoice` if request can be matched with an offer.
- accepts incoming HTLCs with blinding
- initial Qt GUI should be working for sending to offer
- initial QML GUI should be working for sending to offer

notes w.r.t current state:
- only first `path` in `invoice_paths` is tried, iteration of multiple `path` is TODO
- needs gossip enabled, no support for trampoline routing yet.
- expired offers are not automatically removed from the wallet db.
- adds extra blinding field to HTLCs in `adds` log, making the wallet file non-backward compatible

PR #10597, opened 2026-04-22
lnmsg/bolt12: merkle root signing, onion_wire types

f321x:
Implements the [signing scheme](https://github.com/lightning/bolts/blob/master/12-offer-encoding.md#signature-calculation) required for [bolt 12](https://github.com/lightning/bolts/blob/master/12-offer-encoding.md) (`invoice_request` and `invoice` signatures). 
Adds bolt 12 types to onion_wire.csv.

Test coverage will indirectly increase further when introducing the bolt 12 test vectors.


m0wer:
I've been reviewing this and the other BOLT12 PR (#10110). This one looks great as it is! Matches the spec and covers all edge cases I could think about (FWIW).

f321x:
Thanks for the review @m0wer. Please note #10110 is outdated, I currently work on BOLT12 here: https://github.com/spesmilo/electrum/pull/10671
We now have a `bolt12` milestone on GitHub, collecting all the bolt12 related PRs, review of (non-draft) PRs is appreciated.

electrum/lnmsg.py
```diff
@@ -321,6 +348,46 @@ def _parse_msgtype_intvalue_for_onion_wire(value: str) -> int:
...
+
+    return tlv_merkle_nodes[0]
+
+
+def _is_bolt12_signature_tlv_type(tlv_type: int) -> bool:
+    """
+    bolt12: each form is signed using one or more *signature TLV elements*: TLV
+    types 240 through 1000 (inclusive)
+    """
+    return tlv_type in range(240, 1001)
```
SomberNight: I mean... maybe the pyc compiler optimises this... but come on. xD
Naively this check takes like 800 ops.
why not just write `240 <= tlv_type <= 1000`?  perhaps combined with an isinstance(int) if you want

electrum/lnmsg.py
```diff
@@ -580,6 +580,11 @@ def read_field(
                     count=subtype_field_count)
             parsedlist.append(parsed)
 
+            # fd might contain more bytes, but we got passed a count. break when we have 'count' items.
+            # (e.g. nested complex types)
+            if isinstance(count, int) and len(parsedlist) == count:
+                break
+
         return parsedlist if count == '...' or count > 1 else parsedlist[0]
```
SomberNight, line 591: what if `count=10` but `len(parsedlist) == 3`, and we broke out of the `while` loop as there are no more bytes left in `fd`?
Should we not raise?

electrum/lnmsg.py
```diff
@@ -321,6 +348,46 @@ def _parse_msgtype_intvalue_for_onion_wire(value: str) -> int:
     return msg_type_int
 
 
+def _tlv_merkle_root(tlvs: List[Sequence[bytes]]) -> bytes:
```
SomberNight: type hint is incorrect.
maybe `tlvs: List[tuple[int, bytes]]`?

also maybe rename the param to `leaf_tlvs`?

electrum/lnmsg.py
```diff
@@ -321,6 +348,46 @@ def _parse_msgtype_intvalue_for_onion_wire(value: str) -> int:
...
+    first_tlv = None
+    tlv_merkle_nodes = []
+
+    for tlvt, tlv in tlvs:
+        if first_tlv is None:
+            first_tlv = tlv
+        tlv_val = tlv
```
SomberNight: `tlv` here is a full type-len-value serialized raw bytes, isn't it?  it's not just the "val" part of the tlv.
why this assignment anyway?

electrum/lnmsg.py
```diff
@@ -497,12 +575,35 @@ def read_field(
...
+        sign_over_tlvs = []
+        sig_tlv_type, sig_tlv_record_name = None, None
+        sig_records = self.in_tlv_stream_signature_tlv_records.get(tlv_stream_name, {})
+        if signing_key is not None:
+            sig_tlv_types = list(sig_records.keys())  # e.g. [240] ('signature')
+            if len(sig_tlv_types) != 1:
+                raise NotImplementedError
+            sig_tlv_type = sig_tlv_types[0]
+            sig_tlv_record_name = self.in_tlv_stream_get_record_name_from_type[tlv_stream_name][sig_tlv_type]
+            assert sig_tlv_record_name not in kwargs, f"pass either existing {sig_tlv_record_name} or signing_key"
```
SomberNight: assert message is confusing: `signing_key` was positively passed.

electrum/lnmsg.py
```diff
@@ -525,14 +626,29 @@ def write_tlv_stream(self, *, fd: io.BytesIO, tlv_stream_name: str, **kwargs) ->
...
+            # signature TLVs are excluded from the bolt 12 merkle root
+            if signing_key is not None and not is_signature_record:
+                with io.BytesIO() as tlvfd:
+                    _write_tlv_record(fd=tlvfd, tlv_type=tlv_record_type, tlv_val=tlv_val)
+                    sign_over_tlvs.append((tlv_record_type, tlvfd.getvalue()))
+
+    def read_tlv_stream(self, *,
+                        fd: io.BytesIO,
+                        tlv_stream_name: str,
+                        signing_key_path: Optional[Sequence[str]] = None) -> Dict[str, Dict[str, Any]]:
```
SomberNight: not a good way to indent lines: all lines move if we rename the method.
imho best way is:
```
    def read_tlv_stream(
        self, *,
        fd: io.BytesIO,
        tlv_stream_name: str,
        signing_key_path: Optional[Sequence[str]] = None,
    ) -> Dict[str, Dict[str, Any]]:
```

electrum/lnmsg.py
```diff
@@ -321,6 +355,46 @@ def _parse_msgtype_intvalue_for_onion_wire(value: str) -> int:
     return msg_type_int
 
 
+def _tlv_merkle_root(tlvs: List[Sequence[bytes]]) -> bytes:
+    first_tlv = None
+    tlv_merkle_nodes = []
```
SomberNight:
```suggestion
    tlv_merkle_nodes = []  # all nodes at same depth (in tree) at any given time
```

SomberNight (commented):
reviewed as of 622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2

Feels weird that the bolt12-specific signing code is being put in the generic TLV parser (e.g. write_tlv_stream)... but ok.

electrum/lnmsg.py
SomberNight: So apparently this test is fast. I did not know this.
See https://github.com/python/cpython/commit/3e124ae739db8dc6ad43687d6b14466a572a6f8e

Nevertheless, I think the check should be rewritten. If for nothing else, then for clarity.

Also, test:
```
>>> 777777777777777777777778 in range(0, 10**24, 2)
True
>>> 777777777777777777777778.0 in range(0, 10**24, 2)
^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C
```

f321x: Yeah right, changed it in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980

electrum/lnmsg.py
f321x: Added a check in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980

electrum/lnmsg.py
f321x: Fixed in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980

electrum/lnmsg.py
f321x: Yes that seems useless, removed in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980

electrum/lnmsg.py
f321x: Changed in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980

electrum/lnmsg.py
f321x: Added in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980

electrum/lnmsg.py
f321x: Not sure why you consider it confusing? What it means to say is that there should be either a `signing_key` passed, or a `sig_tlv_record_name` in the kwarks, not both at the same time.
I changed the string in https://github.com/spesmilo/electrum/compare/622eb7d9a9897c5ded90f38d5ad5e6035f54c6a2..d40810357cf355cc652ee921a7ff5604f6e2c980, hope it is more clear now.

SomberNight: I had interpreted the "either x or y" as an inclusive or, but I see you meant exclusive or. New phrasing is clearer to me, thanks. :)

SomberNight:
Thanks everyone (to both authors, and for reviewing)

PR #10600, opened 2026-04-22
lnpeer: channel_reestablish: split "they_are_ahead" into ctn vs revnum

SomberNight:
smarter handling of channel_reestablish for "they_are_ahead" case


SomberNight:
note: f321x added a [comment here](https://github.com/SomberNight/electrum/commit/21946e1e870206c0a33cc77ebcae92f6609a6447#commitcomment-183145502) before the PR was opened, which the interface kind of shows here as there being a comment somewhere but the comment itself is not visible AFAICT

PR #10603, opened 2026-04-23
Example: using GET_MASTER_FINGERPRINT for Legacy Client to get the master public key fingerprint

iartemov-ledger:
**This PR provides an example of how new GET_MASTER_FINGERPRINT instruction of Ledger app-bitcoin can be adopted in a software wallet. 
It overrides https://github.com/spesmilo/electrum/pull/10582 one.** 

### Scope
Since the hardening introduced in https://github.com/LedgerHQ/app-bitcoin/pull/297 (and in https://github.com/LedgerHQ/app-bitcoin/pull/299 for `bitcoin_testnet_legacy`), it is no longer possible to get the public key at the master level. Therefore, the new `GET_MASTER_FINGERPRINT` instruction was introduced to replace it.
 
The changes in this PR apply only to the legacy `app-bitcoin` starting from version `2.4.11` (and the `Ledger_Client_Legacy` part). Specifically for Electrum, this only affects the `bitcoin_testnet_legacy` app variant (`Bitcoin Test Legacy`), as there is no hardening applied to the `bitcoin_legacy` variant (`Bitcoin Legacy`). 
For `app-bitcoin-new` and the corresponding `Ledger_Client_New`, everything is already handled well through the `get_master_fingerprint()` call.

The new behavior is to first try the new instruction. If an error is returned (meaning the instruction does not exist in the app), it falls back to `getWalletPublicKey()`. This ensures backward compatibility with app versions older than v2.4.11.

### Tests
This has been successfully tested as follows:
1. Wallet creation with the previous unhardened version of `app-bitcoin` (using the `m/84h/1h/0h` path corresponding to `testnet`), and connecting to it with the hardened version of the app.
2. Wallet creation and connection using the already hardened version.

In the corresponding traces, it was observed that `INS_GET_WALLET_PUBLIC_KEY` commands with `bip32Path.length = 0` were effectively replaced by `INS_GET_MASTER_FINGERPRINT` calls.

### DIFF
--- a/electrum/plugins/ledger/ledger.py
+++ b/electrum/plugins/ledger/ledger.py
@@ -1251,7 +1251,7 @@ class Ledger_KeyStore(Hardware_KeyStore):
 
 class LedgerPlugin(HW_PluginBase):
     keystore_class = Ledger_KeyStore
-    minimum_library = (0, 2, 0)
+    minimum_library = (0, 4, 1)
     maximum_library = (1, 0)
     DEVICE_IDS = [(0x2581, 0x1807),  # HW.1 legacy btchip            # not supported anymore (but we log an exception)
                   (0x2581, 0x2b7c),  # HW.1 transitional production  # not supported anymore
--- a/electrum/plugins/ledger/ledger.py
+++ b/electrum/plugins/ledger/ledger.py
@@ -347,8 +347,9 @@ class Ledger_Client(HardwareClientBase, ABC):
     def __init__(self, *, plugin: HW_PluginBase):
         HardwareClientBase.__init__(self, plugin=plugin)
 
+    @abstractmethod
     def get_master_fingerprint(self) -> bytes:
-        return self.request_root_fingerprint_from_device()
+        pass
 
     @abstractmethod
     def show_address(self, address_path: str, txin_type: str):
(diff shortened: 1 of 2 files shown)


electrum/plugins/ledger/ledger.py
```diff
@@ -390,6 +391,31 @@ def __init__(self, hidDevice: 'HID', *, product_key: Tuple[int, int],
...
+
+        Tries the dedicated GET_MASTER_FINGERPRINT APDU first (INS 0xD0),
+        which does NOT require DERIVE_MASTER permission.  If the device
+        does not support it (old firmware), falls back to
+        getWalletPublicKey("") + HASH160.
+        """
+        try:
+            return self.dongleObject.getMasterFingerprint()
+        except BTChipException as e:
+            if e.sw in (0x6d00, 0x6a80):  # INS not supported / bad data
```
SomberNight, line 405: if you assume a newer version of `ledger_bitcoin`, you need to declare it and bump the minimum:
at runtime:
https://github.com/spesmilo/electrum/blob/7c433c5645efe81011f32ffc182946b9395af2a2/electrum/plugins/ledger/ledger.py#L1219
at build time (for binaries):
https://github.com/spesmilo/electrum/blob/7c433c5645efe81011f32ffc182946b9395af2a2/contrib/requirements/requirements-hw.txt#L16-L18
(the [pinned hashes](https://github.com/spesmilo/electrum/blob/7c433c5645efe81011f32ffc182946b9395af2a2/contrib/deterministic-build/requirements-hw.txt#L352-L357), I will bump correspondingly)
```
  6.35 | E | plugin.DeviceMgr | failed to create client for ledger at b'1-5:1.0': AttributeError("'btchip' object has no attribute 'getMasterFingerprint'")
  File "/home/user/wspace/electrum/electrum/plugin.py", line 1267, in list_pairable_device_infos
    soft_device_id = client.get_soft_device_id()
  File "/home/user/wspace/electrum/electrum/plugin.py", line 1000, in wrapper
    return run_in_hwd_thread(partial(func, *args, **kwargs))
  File "/home/user/wspace/electrum/electrum/plugin.py", line 993, in run_in_hwd_thread
    return fut.result()
           ~~~~~~~~~~^^
  File "/usr/lib/python3.13/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/user/wspace/electrum/electrum/plugins/ledger/ledger.py", line 453, in get_soft_device_id
    self._soft_device_id = self._get_master_fingerprint().hex()
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/wspace/electrum/electrum/plugins/ledger/ledger.py", line 403, in _get_master_fingerprint
    return self.dongleObject.getMasterFingerprint()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'btchip' object has no attribute 'getMasterFingerprint'
```

iartemov-ledger: @SomberNight , thank you for your feedback.
Sorry about that! The new ledger_bitcoin@v0.4.1 is indeed required, and I've just bumped the build-time and runtime versions in the lines you indicated.
Also tested with a `venv`  from scratch locally.

SomberNight:
merged in https://github.com/spesmilo/electrum/commit/5af40f435a15eef13669b6590b1ff21acbab73bc

Thanks for the patch.

PR #10605, opened 2026-04-24
qt: send start_new_window exc to reporter

f321x:
Send more types of exceptions happening during `daemon.load_wallet` in `ElectrumGui.start_new_window()` to the crash reporter to catch e.g. assertion failures instead of showing them to the user as a warning dialog.

Similar to https://github.com/spesmilo/electrum/pull/10121

### DIFF
--- a/electrum/gui/qt/__init__.py
+++ b/electrum/gui/qt/__init__.py
@@ -368,6 +368,17 @@ class ElectrumGui(BaseElectrumGui, Logger):
             self.logger.warning(f"terms of use not accepted, rejecting to start new window")
             return None
 
+        def __handle_wallet_loading_exc(exc: Exception, pos):
+            if isinstance(exc, UserFacingException) \
+                    or isinstance(exc, WalletFileException) and not exc.should_report_crash:
+                self.logger.exception(f"{pos=}")
+                custom_message_box(icon=QMessageBox.Icon.Warning,
+                                   parent=None,
+                                   title=_('Error'),
+                                   text=_('Cannot load wallet') + f' ({pos}):\n' + str(exc))
+            else:
+                send_exception_to_crash_reporter(exc)
+
         wallet = None
         # Try to open with daemon first. If this succeeds, there won't be a wizard at all
         # (the wallet main window will appear directly).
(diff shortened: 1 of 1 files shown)


electrum/gui/qt/__init__.py
```diff
@@ -386,12 +386,14 @@ def start_new_window(
...
                 self.logger.exception('')
-                err_text = str(e) if isinstance(e, WalletFileException) else repr(e)
-                custom_message_box(icon=QMessageBox.Icon.Warning,
-                                   parent=None,
-                                   title=_('Error'),
-                                   text=_('Cannot load wallet') + ' (1):\n' + err_text)
-                if isinstance(e, WalletFileException) and e.should_report_crash:
+                if isinstance(e, UserFacingException) \
+                        or isinstance(e, WalletFileException) and not e.should_report_crash:
+                    err_text = str(e) if isinstance(e, WalletFileException) else repr(e)
```
SomberNight: What about https://github.com/spesmilo/electrum/commit/2c541d2663a2e5779ecd8aaf9f8886b98e241804 then?

Might as well keep `Cannot load wallet (1)` and `Cannot load wallet (2)` in sync.
brilliant naming (`(1)` and `(2)`) on my part btw, don't you think? :D

electrum/gui/qt/__init__.py
```diff
@@ -386,12 +386,14 @@ def start_new_window(
...
+                    custom_message_box(icon=QMessageBox.Icon.Warning,
+                                       parent=None,
+                                       title=_('Error'),
+                                       text=_('Cannot load wallet') + ' (1):\n' + err_text)
```
SomberNight: oh hmm the spaces are inconsistent:
```
' (1):\n'
'(2) :\n'
```
(maybe it should be deduped somehow)

electrum/gui/qt/__init__.py
f321x: Ah yes good point. I deduped the exception now in a nested helper function, much simpler to keep in sync. https://github.com/spesmilo/electrum/compare/23f934978c079668e925b65a0d9cd7fbdbd5985c..c8c44e354c81277e9ece2dc209bc79a18c3d8c5f

electrum/gui/qt/__init__.py
f321x: Fixed by https://github.com/spesmilo/electrum/compare/23f934978c079668e925b65a0d9cd7fbdbd5985c..c8c44e354c81277e9ece2dc209bc79a18c3d8c5f

PR #10606, opened 2026-04-24
trampoline: handle edges with known fees during route edge fee allocation

f321x:
Handle `TrampolineEdge` with known fees when allocating fees to
`TrampolineEdge` with `PLACEHOLDER_FEE` during trampoline route
construction.

This allows to create a mixed route from edges where we know the exact
feerates (e.g. provided through lazy trampoline) and evenly spread the
remaining budget across the edges where the fees are unknown (`PLACEHOLDER_FEE`).

### DIFF
--- a/tests/test_lnrouter.py
+++ b/tests/test_lnrouter.py
@@ -1,18 +1,16 @@
-from math import inf
+import random
 import unittest
-import tempfile
-import shutil
-import asyncio
+from math import inf
 from typing import Optional
 from os import urandom
-from types import MappingProxyType
 
 from electrum import util
 from electrum.channel_db import NodeInfo
 from electrum.onion_message import is_onion_message_node
-from electrum.trampoline import create_trampoline_onion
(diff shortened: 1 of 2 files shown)


tests/test_lnrouter.py
```diff
@@ -551,3 +548,214 @@ async def test_find_path_for_onion_message(self):
...
+        ))
+        for edge in reversed(route[1:]):
+            self.assertEqual(amount, next(amounts_from_destination))
+            amount += edge.fee_for_edge(amount)
+        self.assertEqual([0, 2496, 1000, 2496, 1000], [e.fee_base_msat for e in route])
+
+    def test_fuzz(self):
+        invoice_amount = 1_000_000
+
+        for round_ in range(10**4):
```
f321x: I reduced the range of the fuzz test to `10**4` as runtime was quite long (couple seconds on my laptop) with `10**5`.

Edit: skipping test now

SomberNight (approved):
It's still a bit difficult to understand this logic but I think it cannot really be simplified more. It's as good as we can get it.

PR #10612, opened 2026-04-27
lnutil: bolt 12 LnFeatures changes

f321x:
Changes to `LnFeatures` required for bolt 12. 
* route blinding flags
* `to_tlv_bytes` helper method
* context dependent features (e.g. `BASIC_MPP_*` without `payment_secret` dep)


electrum/lnutil.py
```diff
@@ -1490,6 +1490,13 @@ class LnFeatures(IntFlag):
     _ln_feature_contexts[OPTION_TRAMPOLINE_ROUTING_REQ_ELECTRUM] = (LNFC.INIT | LNFC.NODE_ANN | LNFC.INVOICE)
     _ln_feature_contexts[OPTION_TRAMPOLINE_ROUTING_OPT_ELECTRUM] = (LNFC.INIT | LNFC.NODE_ANN | LNFC.INVOICE)
 
+    OPTION_ROUTE_BLINDING_REQ = 1 << 24
+    OPTION_ROUTE_BLINDING_OPT = 1 << 25
+
+    _ln_feature_direct_dependencies[OPTION_ROUTE_BLINDING_OPT] = {VAR_ONION_OPT}
+    _ln_feature_contexts[OPTION_ROUTE_BLINDING_REQ] = (LNFC.INIT | LNFC.NODE_ANN)
+    _ln_feature_contexts[OPTION_ROUTE_BLINDING_OPT] = (LNFC.INIT | LNFC.NODE_ANN)
```
SomberNight: what about `LNFC.INVOICE`?
looks like that's also specified in the spec:
https://github.com/lightning/bolts/blob/a3772650d8ebc06acf457fcadf97968ebfc4dfff/09-features.md?plain=1#L46

electrum/lnworker.py
```diff
@@ -195,6 +195,7 @@ class ErrorAddingPeer(Exception): pass
 )
```
SomberNight, line 195: can we add a line such as `BASE_FEATURES |= LN_FEATURES_ASSUMED` ?

electrum/lnutil.py
```diff
@@ -1783,9 +1801,12 @@ def ln_compare_features(our_features: 'LnFeatures', their_features: int) -> 'LnF
     """
     our_flags = set(list_enabled_bits(our_features))
     their_flags = set(list_enabled_bits(their_features))
+    assumed_flags = set(list_enabled_bits(LN_FEATURES_ASSUMED))
+    our_flags_effective = our_flags | assumed_flags
+    their_flags_effective = their_flags | assumed_flags
     # check that they have our required features, and disable the optional features they don't have
     for flag in our_flags:
-        if flag not in their_flags and get_ln_flag_pair_of_bit(flag) not in their_flags:
+        if flag not in their_flags_effective and get_ln_flag_pair_of_bit(flag) not in their_flags_effective:
```
SomberNight: This would mean that even if the remote node did not signal e.g. option_dataloss_protect in their INIT message, we just ignore that, assume they support it, and proceed as usual. I find this behaviour surprising and have some doubts this is what the spec really means by "ASSUMED".

electrum/lnutil.py
f321x: Added this in https://github.com/spesmilo/electrum/pull/10612/commits/f78e22e12ca231ac24e86fb4ada8e335de79bbd4, though i don't see the point why this should be signaled in a bolt 11 invoice.

electrum/lnworker.py
f321x: I removed `LN_FEATURES_ASSUMED` from this PR completely. 
I only relied on them to validate `BASIC_MPP_OPT` in bolt 12 invoices without the `payment_secret` dependency, *assuming* it is available.
Now instead i added context dependent transitive features and it doesn't seem worth declaring `LN_FEATURES_ASSUMED`.

electrum/lnutil.py
f321x: Ok according to [this PR comment](https://github.com/lightning/bolts/pull/1307#issuecomment-3655796025), iiuc, the intention of marking features *ASSUMED* really seems to be to just assume they are there and, at one point, ignore them completely (as bolt 9 says).

However in practice this doesn't seem to be done by implementations, and e.g. sending INIT without an assumed feature flag would not be accepted anywhere. ASSUMED seems to mean more like "always required to be set" in practice.

Maybe spec compliant thing would be to ignore them when validating features but keep signaling them forever for compatibility. But then, there are still nodes that don't support these features, so it is probably not safe to just ignore them.

For now i removed this assumed thing from the PR. Instead to get the bolt 12 feature validation working i rely on context dependent transitive dependencies, which seems conceptually cleaner than relying on assumed features anyway.

SomberNight: Does the spec plan to eventually reuse these feature bits? :D
This whole approach to "ASSUMED" just seems weird, and more trouble than is worth.
The reasonable thing would be to just burn the feature bits and require them to always be signalled.

f321x: > Does the spec plan to eventually reuse these feature bits? :D

From my understanding of this comment it seems like a goal of the change is to eventually be able to fully remove these assumed feature bits from the spec. Once there is no reference to them anymore i *assume* they could also be reused:

> This way, we'll eventually end up in a state where we can really assume those features and maybe even remove references to the feature bits they used (but maybe not 🤷). Meanwhile, we don't want to completely remove from the spec references to the fact that these features were at some point no assumed, and may or may not be supported.

SomberNight (approved):
Looks good.

PR #10615, opened 2026-04-28
qml: improve history list scrolling performance

f321x:
Some performance improvements attempting to make scrolling through a large history smoother on QML.

### DIFF
--- a/electrum/gui/qml/components/controls/HistoryItemDelegate.qml
+++ b/electrum/gui/qml/components/controls/HistoryItemDelegate.qml
@@ -75,6 +75,8 @@ Item {
                     Layout.preferredHeight: constants.iconSizeLarge
                     Layout.alignment: Qt.AlignVCenter
                     Layout.rowSpan: 2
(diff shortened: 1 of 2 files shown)


electrum/gui/qml/components/controls/HistoryItemDelegate.qml
```diff
@@ -21,6 +32,8 @@ Item {
         ItemDelegate {
             Layout.fillWidth: true
             Layout.preferredHeight: txinfo.height
+            // suppress Material press overlay, it flickers on rows during touch scroll
+            background: null
```
SomberNight, line 36: Note: ISTM while this happens on qt 6.10.2, it does not happen on qt 6.11.
But changing the qt version might introduce other regressions.
Also btw, qt 6.11 requires jdk 21, up from jdk 17, which conflicts with the markusfisch qr scanner (ref https://github.com/spesmilo/electrum/pull/10452)

SomberNight:
Looks ok. Tested. Thanks.

PR #10620, opened 2026-04-30
LNWallet: make anchor channels required

f321x:
If config.ENABLE_ANCHOR_CHANNELS is set, signal
OPTION_ANCHORS_ZERO_FEE_HTLC_REQ instead of OPT.

This will prevent opening new outgoing and incoming SRK channels. 
It will also prevent connecting to peers that don't signal anchor support at all (INIT features), so existing channels to them would need to get force closed or the peer needs to be bullied into updating their node software.
Existing SRK channels with peers that signal anchor support (e.g. SRK channel opened to Electrum Trampoline) remain usable.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1015,7 +1015,7 @@ class LNWallet(Logger):
         if features is None:
             features = LNWALLET_FEATURES
             if self.config.ENABLE_ANCHOR_CHANNELS:
-                features |= LnFeatures.OPTION_ANCHORS_ZERO_FEE_HTLC_OPT
+                features |= LnFeatures.OPTION_ANCHORS_ZERO_FEE_HTLC_REQ
             if self.config.OPEN_ZEROCONF_CHANNELS:
                 features |= LnFeatures.OPTION_ZEROCONF_OPT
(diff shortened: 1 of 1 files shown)


electrum/lnworker.py
```diff
@@ -1015,7 +1015,7 @@ def __init__(self, wallet: 'Abstract_Wallet', xprv, *, features: LnFeatures = No
         if features is None:
             features = LNWALLET_FEATURES
             if self.config.ENABLE_ANCHOR_CHANNELS:
-                features |= LnFeatures.OPTION_ANCHORS_ZERO_FEE_HTLC_OPT
+                features |= LnFeatures.OPTION_ANCHORS_ZERO_FEE_HTLC_REQ
```
SomberNight, line 1018: > This will prevent opening new [...] incoming SRK channels.

My reading of BOLT-02 suggests that just because option_anchors is negotiated on the peer session, it is still allowed to open an SRK channel by setting `channel_type` in `open_channel` accordingly.

Reviewing this and looking at the related code led me down a small rabbit hole, so I will open an alternative PR instead.

SomberNight: see https://github.com/spesmilo/electrum/pull/10622

PR #10621, opened 2026-04-30
lnonion/onion_wire: encrypted_data -> encrypted_recipient_data

f321x:
Previously there were different names for the encrypted_recipient_data field in blinded paths, depending on it being for a onion message or a blinded payment. 

This has been unified in the bolts, now it is always named encrypted_recipient_data.
This change allows us to simplify code and is required for compatibility.

See:
https://github.com/lightning/bolts/commit/3c0fd9ad901eeef9ef6890c1bb85592cbf59e3d3

Also renames `current_blinding_point` to `current_path_key` in onion_wire.csv. 
See https://github.com/lightning/bolts/pull/1181#issue-2402112734

### DIFF
--- a/electrum/lnonion.py
+++ b/electrum/lnonion.py
@@ -288,29 +288,16 @@ def decrypt_onionmsg_data_tlv(*, shared_secret: bytes, encrypted_recipient_data:
 
 
 def encrypt_hops_recipient_data(
-        tlv_stream_name: str,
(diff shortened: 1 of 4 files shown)


accumulator:
Yes that was very annoying, I considered deviating from the BOLT specs in our codebase, as it's serialized to numbered fields in packets anyway, but indeed it's confusing enough without multiple names

PR #10622, opened 2026-04-30
lnpeer: clean-up channel_type and enforce all newly opened channels use anchors

SomberNight:
alternative to https://github.com/spesmilo/electrum/pull/10620
and lots of related clean-up

As in https://github.com/spesmilo/electrum/pull/10620#issue-4357139535 :

> This will prevent opening new outgoing and incoming SRK channels.
> It will also prevent connecting to peers that don't signal anchor support at all (INIT features), so existing channels to them would need to get force closed or the peer needs to be bullied into updating their node software.
> Existing SRK channels with peers that signal anchor support (e.g. SRK channel opened to Electrum Trampoline) remain usable.

- also rm the `config.ENABLE_ANCHOR_CHANNELS` config var, which some users might have set (maybe even forgot they did)
  - some power-users might have set that to False, esp as that was a loophole to keep using a hardware-signer-backed wallet or a watch-only wallet with lightning 
  - instead add a new `config.TEST_LN_OPEN_SRK_CHANNELS`, explicitly for testing
    - if some user fiddles with config vars that have TEST or EXPERIMENTAL in their name, they are on their own

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -1243,12 +1243,15 @@ class Peer(Logger, EventListener):
         if channel_type is None:
             raise Exception("channel_type MUST be present in open_channel, but missing")
         # MUST fail the channel if channel_type is not suitable.
-        # TODO fail if channel_type does not have anchors
-        else:
-            channel_type = ChannelType.from_bytes(channel_type['type'], byteorder='big').discard_unknown_and_check()
-            if not channel_type.complies_with_features(self.features):
-                raise Exception("sender has sent a channel type we don't support")
-        assert isinstance(channel_type, ChannelType)
+        channel_type = ChannelType.from_bytes(channel_type['type'], byteorder='big').discard_unknown_and_check()
+        if not channel_type.complies_with_features(self.features):
+            raise Exception("sender has sent a channel type we don't support")
+        assert channel_type & ChannelType.OPTION_STATIC_REMOTEKEY, "new legacy channel?!"
+        if not self.config.TEST_LN_OPEN_SRK_CHANNELS:
+            if not channel_type & ChannelType.OPTION_ANCHORS:
+                # note: BOLT-02 does NOT forbid opening new SRK chan
+                #       just because ANCHORS has been negotiated as peer feature
+                raise Exception("refusing to open new static_remotekey channel")
 
         is_zeroconf = bool(channel_type & ChannelType.OPTION_ZEROCONF)
         if is_zeroconf and not self.config.ZEROCONF_TRUSTED_NODE.startswith(self.pubkey.hex()):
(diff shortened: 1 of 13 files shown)


SomberNight:
I cherry-picked the trivial commits directly to master to reduce the noise here:
https://github.com/spesmilo/electrum/commit/4357ef2f64f4972b234af0fd514f4c2f124940dd
https://github.com/spesmilo/electrum/commit/be2096f842c1c52e861aef95fc54fefcc6468038

electrum/lnpeer.py
```diff
@@ -994,16 +991,18 @@ async def channel_establishment_flow(
 
         channel_flags = CF_ANNOUNCE_CHANNEL if public else 0
         feerate: Optional[int] = self.lnworker.current_target_feerate_per_kw(
-            has_anchors=self.use_anchors()
+            has_anchors=not self.lnworker.config.TEST_LN_OPEN_SRK_CHANNELS,
```
ecdsa: didn't you just introduce a property `self.config`?  :-)

electrum/lnutil.py
```diff
@@ -1619,39 +1619,35 @@ class ChannelType(IntFlag):
...
+    def complies_with_features(self, peer_features: LnFeatures) -> bool:
+        """Returns whether channel_type complies with peer_features.
+        peer_features is negotiated features for the peer session.
+
+        note: enforces channel_type is-SUBSET-of peer_features
+              but does NOT enforce cflag-related-peer_features is-SUBSET-of channel_type.
+              For example, even if opt_anchors is a negotiated peer_feature, (as per my reading of BOLT-02),
+              it is still allowed to open an SRK channel (by setting channel_type accordingly).
+        """
+        self.check_combinations()  # test if raises
```
ecdsa, line 1643: `check_combinations` is called in `make_local_config_for_new_channel`, which is called in both directions (when we open or when remote opens). It feels redundant to call it a third time here (this is the non-initiator flow). If the goal is to call it earlier, maybe we should remove it from `make_local_config_for_new_channel`, and call it in `channel_establishment_flow` as well?

electrum/lnpeer.py
SomberNight: ah hmmm, yes, must have gotten lost when I reordered some commits before.
fixed in https://github.com/spesmilo/electrum/compare/f9297199bcf8e2e646e0cd02ee7207f0d70b6cff..621eac23f232852c7c4fa872271e47f97898b890

electrum/lnutil.py
SomberNight: I made this now symmetric in (force-push):
https://github.com/spesmilo/electrum/compare/f9297199bcf8e2e646e0cd02ee7207f0d70b6cff..621eac23f232852c7c4fa872271e47f97898b890

> If the goal is to call it earlier

yes, the goal is to call it early in the case of incoming-chan-open

but I would rather always call it in `make_local_config_for_new_channel`, just as a sanity check.

So now I added an extra assert for the outgoing-chan-open case, to make it all symmetric.

PR #10625, opened 2026-05-04
qt: wizard: add wizard data to wizard exception

f321x:
Enrich `InvalidPassword` thrown by `daemon.load_wallet` during `ElectrumGui._start_wizard_to_select_or_create_wallet` with the sanitized wizard data to help debugging #10623.
The wizard should have already refused the invalid password, though it is not clear to me how the invalid password escaped the wizard, maybe this helps.


f321x:
Let's wait for more reports on this issue.

PR #10630, opened 2026-05-06
lnpeer: send channel update also for private channels, if we are forw…

ecdsa:
…arding

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -598,7 +598,6 @@ class Peer(Logger, EventListener):
                 for chan in public_channels:
                     if chan.is_open() and chan.peer_state == PeerState.GOOD:
                         self.maybe_send_channel_announcement(chan)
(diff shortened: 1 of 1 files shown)


f321x:
If the peer doesn't get the channel_update it will never get it in the future. 
Maybe we should send it on reestablish as well?
Also the peer might prune old updates (though they shouldn't for their own channels), or our parameters have changed.

SomberNight:
> If the peer doesn't get the channel_update it will never get it in the future.

Ahh, we still have the [fallback code](https://github.com/spesmilo/electrum/blob/abc5e25a2b59573ce5004adb6bc1f97994ac80f6/electrum/lnworker.py#L3280-L3298) because lnd used to have the [same bug](https://github.com/lightningnetwork/lnd/issues/1347). Let's try not to repeat that.

>  Maybe we should send it on reestablish as well?

Yeah, I agree. Let's just send it near the end of `reestablish_channel()`.

PR #10631, opened 2026-05-06
qml: fix lightning payment status string race

f321x:
When paying a lightning invoice there was a race between the
`on_event_payment_succeeded` and `on_event_invoice_status` callbacks.
Depending on which was called last the string at the top
of the invoice view would either say "Paid!" (correct) or
"This invoice was already paid" (technically correct but confusing).

By keeping state if the invoice payment attempt was initiated in this
session the subsequent calls to `update_userinfo` can correctly
determine which string to show.

Can be reproduced with this diff:
```
diff --git a/electrum/gui/qml/qeinvoice.py b/electrum/gui/qml/qeinvoice.py
index 696506ffd..ff02c4cc6 100644
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -1,3 +1,4 @@
+import asyncio
 import copy
 import threading
 from enum import IntEnum
@@ -98,8 +99,9 @@ class QEInvoice(QObject, QtEventListener):
             self.userinfo = _('Payment failed: ') + reason
 
     @event_listener
-    def on_event_invoice_status(self, wallet, key, status):
+    async def on_event_invoice_status(self, wallet, key, status):
         if self._wallet and wallet == self._wallet.wallet and key == self.key:
+           await asyncio.sleep(1)
             self.update_userinfo()
             self.determine_can_pay()
             self.statusChanged.emit()
```

### DIFF
--- a/electrum/gui/qml/qeinvoice.py
+++ b/electrum/gui/qml/qeinvoice.py
@@ -65,6 +65,7 @@ class QEInvoice(QObject, QtEventListener):
         self._invoiceType = QEInvoice.Type.Invalid
         self._effectiveInvoice = None  # type: Optional[Invoice]
         self._userinfo = ''
+        self._paid_in_this_session = False
         self._lnprops = {}
         self._amount = QEAmount()
         self._amountOverride = QEAmount()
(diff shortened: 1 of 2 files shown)


ecdsa:
good catch!

PR #10635, opened 2026-05-08
lnpeer: send channel_update on channel_reestablish

f321x:
If we have forwarding enabled, send a channel_update message to the peer after a channel_reestablish.
The peer needs to know our channel constraints to receive a payment (e.g. for r_tags or blinded paths) and sending it just once on mark_open might not be enough as the peer could miss it, lose it or it might get out of date if we'd change our forwarding fees. See https://github.com/spesmilo/electrum/pull/10630#issuecomment-4390045965.


SomberNight (approved):
Looks good.

PR #10636, opened 2026-05-08
qml: allow hiding balances and amounts

f321x:
This allows hiding the wallet balance and transaction amounts by long-pressing the balance summary of the main view in QML (or toggling the config in the Preferences). 

When enabled the amounts will be replaced by a fixed `****` string and the units and fiat are hidden.  

[Screencast_20260512_165512.webm](https://github.com/user-attachments/assets/72e99bee-556c-4723-b308-350750b5382e)

### DIFF
--- a/electrum/gui/qml/components/controls/AddressDelegate.qml
+++ b/electrum/gui/qml/components/controls/AddressDelegate.qml
@@ -67,7 +67,7 @@ ItemDelegate {
                 }
                 Label {
                     font.family: FixedFont
-                    text: Config.formatSats(model.balance, false)
+                    text: Config.formatSats(model.balance, false, Config.hideAmounts)
                     visible: model.balance.satsInt != 0
                 }
                 Label {
--- a/electrum/gui/qml/components/controls/BalanceSummary.qml
+++ b/electrum/gui/qml/components/controls/BalanceSummary.qml
@@ -15,9 +15,10 @@ Item {
     property string formattedLightningBalance
 
     function setBalances() {
-        root.formattedConfirmedBalance = Config.formatSats(Daemon.currentWallet.confirmedBalance)
-        root.formattedTotalBalance = Config.formatSats(Daemon.currentWallet.totalBalance)
-        root.formattedLightningBalance = Config.formatSats(Daemon.currentWallet.lightningBalance)
+        let hide = Config.hideAmounts
+        root.formattedConfirmedBalance = Config.formatSats(Daemon.currentWallet.confirmedBalance, false, hide)
+        root.formattedTotalBalance = Config.formatSats(Daemon.currentWallet.totalBalance, false, hide)
+        root.formattedLightningBalance = Config.formatSats(Daemon.currentWallet.lightningBalance, false, hide)
         if (Daemon.fx.enabled) {
             root.formattedTotalBalanceFiat = Daemon.fx.fiatValue(Daemon.currentWallet.totalBalance, false)
         }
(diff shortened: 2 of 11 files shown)


electrum/util.py
```diff
@@ -879,6 +879,11 @@ def format_fee_satoshis(fee, *, num_zeros=0, precision=None):
     return format_satoshis(fee, num_zeros=num_zeros, decimal_point=0, precision=precision)
 
 
+def redact_amount_string(s: str) -> str:
+    """Replace ASCII digits in a formatted-amount string with '*'."""
+    return re.sub(r'[0-9]', '*', s)
+
```
SomberNight: > The length of the amount still exposes some information (*** sat vs ** *** *** sat), but i think that's a good tradeoff between usability and privacy.

I disagree. What is the goal of the feature? If someone has a thousand coins, it does not gain them much that a shoulder-surfer does not see whether it is 1000 or 9999, does it?

Just make it always be (say) four stars, both for shorter and longer "true" values.

That way you can still keep displaying the units (e.g. BTC vs sat) - I imagine hiding the unit would add more complexity.

So e.g.:
```
36.204 84 mBTC --> **** mBTC
3 620 484 sat --> **** sat 
```

but if it is simple, we could also hide the units, so e.g.:
```
36.204 84 mBTC --> ****
3 620 484 sat --> ****
```

btw, I just tested and this last example seems to align with what phoenix is doing

f321x: My thinking was that showing the exact amount of digits provides the benefit of being able to at least roughly estimate the balance as a user, improving the usability of the wallet with the mode enabled.

So e.g. when standing in line to pay for a drink in public you're still able to infer if it you'll be able to pay for the drink instead of having a.) deactivate the obfuscation in public or b.) attempt the payment and then awkwardly fail because the balance was too low. In the end the feature is only beneficial if it doesn't break the main usecase of a wallet, making transactions.

On the other hand I assumed the human mind is optimized to parse digits/numbers very quickly, but has a hard time counting `*` quickly, so a quick shoulder surfer likely isn't able to get much information as they are not used to the UI and counting `*` characters from distance.

But OTOH this mostly affects the "sat" unit, if set to "BTC" it will anyways most of the time be `*.*** *** ** BTC`, so we might as well just show `****` uniformly.

> What is the goal of the feature?

My motivation was actually not preventing shoulder surfers (though this is probably what the majority of users would benefit from through this feature), but instead being able to demonstrate the app and its features to others without having to use a dummy wallet or reveal my exact balance. For this i agree showing the exact digits is not sufficient but i found it better than not having the obfuscation at all and considered the tradeoffs above for the shoulder surfer usecase.

I'll change it to just use `****`.

f321x: Changed it to show `****` without unit in relevant places (main view, channels and utxo list), while still showing amounts in the other views (e.g. when confirming a payment or doing a swap).

electrum/gui/qml/qeconfig.py
```diff
@@ -362,9 +373,13 @@ def formatSatsForEditing(self, satoshis):
...
     @pyqtSlot('qint64', result=str)
     @pyqtSlot('qint64', bool, result=str)
+    @pyqtSlot('qint64', bool, bool, result=str)
     @pyqtSlot(QEAmount, result=str)
     @pyqtSlot(QEAmount, bool, result=str)
-    def formatSats(self, satoshis, with_unit=False):
+    @pyqtSlot(QEAmount, bool, bool, result=str)
+    def formatSats(self, satoshis, with_unit=False, hide_amounts: bool = False):
+        if hide_amounts:
+            return '****'
```
accumulator, line 382: Instead of passing `Config.hideAmounts` if you want the configured setting, you could define `hide_amounts` as a tri-state `bool|None` and fall back to `Config.hideAmounts` if it is `None`. This allows to only specify hide_amounts if it deviates from the configured setting.

electrum/gui/qml/components/controls/BalanceSummary.qml
```diff
@@ -15,9 +15,10 @@ Item {
     property string formattedLightningBalance
 
     function setBalances() {
-        root.formattedConfirmedBalance = Config.formatSats(Daemon.currentWallet.confirmedBalance)
-        root.formattedTotalBalance = Config.formatSats(Daemon.currentWallet.totalBalance)
-        root.formattedLightningBalance = Config.formatSats(Daemon.currentWallet.lightningBalance)
+        let hide = Config.hideAmounts
+        root.formattedConfirmedBalance = Config.formatSats(Daemon.currentWallet.confirmedBalance, false, hide)
+        root.formattedTotalBalance = Config.formatSats(Daemon.currentWallet.totalBalance, false, hide)
+        root.formattedLightningBalance = Config.formatSats(Daemon.currentWallet.lightningBalance, false, hide)
```
accumulator, line 21: e.g. here

electrum/gui/qml/components/controls/AddressDelegate.qml
```diff
@@ -67,7 +67,7 @@ ItemDelegate {
                 }
                 Label {
                     font.family: FixedFont
-                    text: Config.formatSats(model.balance, false)
+                    text: Config.formatSats(model.balance, false, Config.hideAmounts)
```
accumulator, line 70: and here

accumulator (changes requested):

electrum/gui/qml/qeconfig.py
f321x: This would work but I think the default case we want, is showing the amounts. If the default case (not passing `Config.hideAmounts`) would be to read the config, we would actually have to pass `False` as argument in a lot of places simply to not hide the amounts. This would change more lines than the current approach I think?

Also not passing `Config.hideAmounts` from QML would remove the QML binding to the config, so if the config changes and the label doesn't get reloaded otherwise it wouldn't update when changing the config.

accumulator: > This would work but I think the default case we want, is showing the amounts. If the default case (not passing `Config.hideAmounts`) would be to read the config, we would actually have to pass `False` as argument in a lot of places simply to not hide the amounts. This would change more lines than the current approach I think?

You might be right, it depends what % of amount occurrences we want to hide by config. In the most extreme case, we might only want to force-show amounts for requests and invoices (those not in listviews).

> Also not passing `Config.hideAmounts` from QML would remove the QML binding to the config, so if the config changes and the label doesn't get reloaded otherwise it wouldn't update when changing the config.

I'm quite sure binding doesn't work for `Config.formatSats(..., Config.hideAmounts)` . Using a property as a parameter in a slot-call is not a binding and doesn't update. This is probably not apparent in the Delegates as they are rerendered often (?) by the `ListView`s, and in places where updates are triggered using a `Connections` block + update function (e.g. `BalanceSummary`).

f321x: > You might be right, it depends what % of amount occurrences we want to hide by config

Right now out of 33 `Config.formatSats(` occurrences i pass the hide argument to 8 calls (~1/4).
I think it is simpler to hide the amounts explicitly, most views aren't really useful without showing amounts and aren't critical either as they are rarely exposed in public, e.g. when paying some coffee.

> Using a property as a parameter in a slot-call is not a binding and doesn't update

ah true, so changing this wouldn't remove any additional update guarantee

PR #10639, opened 2026-05-11
qt history: skip unchanged refreshes

KingParmenides:
Refs #6625.

This reduces unnecessary Qt history-list churn for large wallets by returning early when `wallet.get_full_history()` produces the same transaction data already displayed. In that case the model still reapplies the active filter and transaction counter, but avoids removing every row, recreating every `HistoryNode`, rebuilding status cache, and triggering a full insert/sort cycle.

The patch also fixes the inclusive row range passed to `beginRemoveRows()` and avoids calling `beginInsertRows()` for an empty model.

This does not claim to fully solve #6625 by itself; it targets one remaining expensive no-op refresh path in the large-history case.

Verification:
- `git diff --check`
- `python3 -m compileall -q electrum/gui/qt/history_list.py`

I could not run Qt GUI tests locally because the system Python environment in this workspace does not have `PyQt6` installed.

### DIFF
--- a/electrum/gui/qt/history_list.py
+++ b/electrum/gui/qt/history_list.py
@@ -302,9 +302,14 @@ class HistoryModel(CustomModel, Logger):
             onchain_domain=self.get_domain(),
             include_lightning=self.should_include_lightning_payments(),
         )
+        if transactions == self.transactions:
+            self.view.filter()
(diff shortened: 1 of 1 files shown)


f321x:
before being submitted.

PR #10640, opened 2026-05-11
qt: skip unchanged address and coin refreshes

Dewaldnel11:
## Summary
- Avoid rebuilding the full Qt Addresses model when the displayed address rows have not changed.
- Avoid rebuilding the full Qt Coins/UTXO model when the displayed UTXO rows have not changed.
- Keep the existing full refresh behavior whenever balances, labels, frozen state, filters, UTXO set, or coin-control state changes.

This targets one remaining large-wallet freeze source discussed in #6625: repeated full redraws of large Addresses and Coins tabs during wallet updates. It is scoped as a partial performance fix and is separate from the history-list-only optimization in #10639.

## Verification
- `python3 -m compileall -q electrum/gui/qt/address_list.py electrum/gui/qt/utxo_list.py`
- `git diff --check`

I could not run a full Qt GUI benchmark in this environment because the local Python environment does not have the Electrum Qt runtime dependencies installed.

### DIFF
--- a/electrum/gui/qt/address_list.py
+++ b/electrum/gui/qt/address_list.py
@@ -101,6 +101,7 @@ class AddressList(MyTreeView):
         )
         self.wallet = self.main_window.wallet
         self._address_list_status = 0  # type: int
+        self._last_address_rows = None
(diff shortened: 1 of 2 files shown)


f321x:
before being submitted.

PR #10649, opened 2026-05-14
ci: code review: improvements (use system prompt, give more permissions)

f321x:
Skimming through the [headless docs](https://code.claude.com/docs/en/headless) I think these changes could improve the results of the CI:
* pass `--dangerously-skip-permissions` so it can actually execute and modify code
* append the prompt text to the system prompt, this prioritizes it higher and keeps it more available in the context
* log effort level
* increase timeout: 20 -> 60 min (20 are too short for complex/large contexts)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -264,7 +264,9 @@ task:
   install_script:
     - npm install -g @anthropic-ai/claude-code
   review_script:
-    - python3 contrib/ci/claude_security_review.py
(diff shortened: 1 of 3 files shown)


SomberNight (approved):
Ok.  but note the only real guardrail is that people with write permissions on the repo are trusted, and they are also trusted not to blindly manually trigger this task on problematic external contributions

PR #10650, opened 2026-05-15
submarine_swaps: Reintroduce support for the old submarine swap protocol

### LINKED ISSUE #10611: [Electrum Swap,WEX] Third party client forward swaps require different protocol on swap server side
### Description

Follow up based on X request https://x.com/ElectrumWallet/status/2048374535578062988 to move the discussion to an issue.

We are trying to expand Electrum Swap ecosystem by creating interface for people who do not use Electrum Wallet. We have released the first public version where we implemented reverse swaps and we would like to support forward swap as well in order to provide complete swap gateway to Electrum Swaps.

The research of forward swaps lead us to discovery of a change in Electrum's implementation (https://github.com/spesmilo/electrum/blob/9b26c1812dd06ef716a7a9f3369c407b193011e7/electrum/submarine_swaps.py#L871) from an "old flow" (based on standard Bolt11) to a "new flow" (based on HODL invoices).

As we do not have any control of the client's LN wallet and because HODL invoices are not commonly supported by wallets, the new flow does not seem to be viable for us to use. The old flow only requires the user's wallet to be able to generate a standard Bolt11 invoice, so it'd be perfect for us. However, the support for the old flow on the server side has been removed (https://github.com/spesmilo/electrum/blob/9b26c1812dd06ef716a7a9f3369c407b193011e7/electrum/submarine_swaps.py#L1466) from Electrum's code base.

We would like to ask for ideas on how to reach support for forward swaps. 

The only idea we currently have is to reintroduce the server side support for the old flow:
- new extra flag goes into swap announcement, so that we can distinguish providers that we can use for forward swaps from others; this should be backwards compatible with old swap servers and clients, who would still use the old swap announcement and ignore the extra flag
- swap server code would implement both versions of the swap, the new flow that is used by current Electrum swap clients as is, and the old flow, for the purpose of enabling third party access to the ecosystem
- optionally, we could add a setting that could enable/disable the old flow, depending on whether it would be opt-in or opt-out feature

Would Electrum be happy to accept such contribution? Or is there a better solution?

Further, it has been suggested that

> Note: If the server uses our protocol (not Boltz), then a reverse swap failing will not result in the user losing fees, because the invoices are bundled.

I am personally unaware of how we could use that. Bundled invoices is something that does not seem to have wide support either. I'd be happy to discuss how we can improve our reverse swap support. Currently, the swap server does not complete the fee invoice before the main swap invoice is also sent, so an honest swap server would not take the fee if the main invoice is not paid. However, a malicious swap server can steal the funds, but the original post I found about this (https://diyhpl.us/~bryan/irc/bitcoin/bitcoin-dev/linuxfoundation-pipermail/lightning-dev/2023-June/003977.txt) mentions that the theft by malicious provider is not prevented either in case of bundled. Of course, the advantages of having just one invoice are still great, so even if the theft cannot be prevented, we would like to improve the UX. So any help understanding what can be done here would also be appreciated.

MartyIX:
Closes #10611

This PR

- Reintroduces the old v1 flow for submarine swaps.
    - A preliminary check ensures the LN payment route has a high chance of success before proceeding.
- Updates the swap server to signal support for `forwardv1` (the old protocol).

## Motivation

We've been trying to expand Electrum Swap ecosystem by creating interface &ndash; https://whales.exchange &ndash; for people who do not use Electrum Wallet. We have released the first public version of https://whales.exchange where we implemented support for reverse swaps. We would like to support forward swap as well in order to provide complete swap gateway to Electrum Swaps.

The current implementation of submarine swaps in Electrum require control over user's LN wallet which https://whales.exchange does not control (see the [design](https://github.com/AITIS-s-r-o/whales-exchange-web-app#design)). However, Electrum used to implement a different protocol, so called _old flow_ or simply _v1 flow_, for submarine swaps which did not have the requirement. The old implementation was removed [because](https://x.com/ElectrumWallet/status/2048328014459490491):

> The old flow was removed because servers sometimes fail to find a route for the LN payment, in which case users have to get an on-chain refund (funds locked + mining fees). The new flow cannot be supported with bolt11 invoices; it requires a wallet that supports hold invoices.

This PR attempts to bring back the old flow, so Electrum would support the old and the new submarine protocols. Moreover, to address the above issue, a new preliminary "is-there-a-LN-payment-route" check was added to somewhat mitigate the issue.

A short description of both protocols follows:

Old submarine swap flow (v1) is

- User generates an LN invoice with RHASH, and knows preimage. 
- User creates on-chain output locked to RHASH.
- **Server checks if there is a LN payment route, otherwise the swap is rejected** [new]
- Server pays LN invoice. By completing payment, user reveals preimage.
- Server spends the on-chain output using preimage.

[New submarine swap flow](https://github.com/spesmilo/electrum/pull/8566) is:

- User requests swap  (RPC 'createnormalswap')
- Server creates preimage, sends RHASH to user
- User creates hold invoice, sends it to server  (RPC 'addswapinvoice')
- Server sends HTLC, user holds it
- User creates on-chain output locked to RHASH
- Server spends the on-chain output using preimage (revealing the preimage)
- User fulfills HTLC using preimage

## Discussions

* #10611
* https://x.com/ElectrumWallet/status/2048374535578062988

(edit: Rebased to b9be9749f61224d51c6c96ab6b693385b9904172. Originally, it was based on 9b26c18.)

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1868,6 +1868,7 @@ class LNWallet(Logger):
             full_path: LNPaymentPath = None,
             channels: Optional[Sequence[Channel]] = None,  # my own direct channels
             budget: Optional[PaymentFeeBudget] = None,  # to limit max fee
+            probe_only: bool = False,  # checks if a route for the payment exists, without actually sending the payment
     ) -> Tuple[bool, List[HtlcLog]]:
         """Attempt to pay a Lightning invoice (find routes, do MPP, send HTLCs).
 
@@ -1912,7 +1913,7 @@ class LNWallet(Logger):
             attempts = 30
         success = False
         try:
-            await self.pay_to_node(
+            probe_only_result = await self.pay_to_node(
                 node_pubkey=invoice_pubkey,
                 payment_hash=payment_hash,
                 payment_secret=payment_secret,
@@ -1924,7 +1925,12 @@ class LNWallet(Logger):
                 full_path=full_path,
                 channels=channels,
                 budget=budget,
+                probe_only=probe_only,
             )
+
+            if probe_only:
+                return probe_only_result, []
+
             success = True
         except PaymentFailure as e:
             self.logger.info(f'payment failure: {e!r}')
@@ -1959,10 +1965,13 @@ class LNWallet(Logger):
             budget: PaymentFeeBudget,
             channels: Optional[Sequence[Channel]] = None,
             fw_payment_key: str = None,  # for forwarding
-    ) -> None:
+            probe_only: bool = False,  # checks if a route for the payment exists, without actually sending the payment
+    ) -> bool:
         """
         Can raise PaymentFailure, ChannelDBNotLoaded,
         or OnionRoutingFailure (if forwarding trampoline).
+
+        returns bool: In case of probe_only, true/false denoting if a route to the destination exists.
         """
 
         assert budget
@@ -2021,8 +2030,15 @@ class LNWallet(Logger):
                             channels=channels,
                             budget=budget._replace(fee_msat=remaining_fee_budget_msat),
                         )
+
+                        can_route = False
+
                         # 2. send htlcs
                         async for sent_htlc_info, cltv_delta, trampoline_onion in routes:
+                            if probe_only:
+                                can_route = True
+                                continue
+
                             await self.pay_to_route(
                                 paysession=paysession,
                                 sent_htlc_info=sent_htlc_info,
@@ -2030,6 +2046,9 @@ class LNWallet(Logger):
                                 trampoline_onion=trampoline_onion,
                                 fw_payment_key=fw_payment_key,
                             )
+
+                        if probe_only:
+                            return can_route
                     # invoice_status is triggered in self.set_invoice_status when it actually changes.
                     # It is also triggered here to update progress for a lightning payment in the GUI
                     # (e.g. attempt counter)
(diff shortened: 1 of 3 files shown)


electrum/lnworker.py
```diff
@@ -2018,15 +2027,25 @@ async def pay_to_node(
...
                             await self.pay_to_route(
                                 paysession=paysession,
                                 sent_htlc_info=sent_htlc_info,
                                 min_final_cltv_delta=cltv_delta,
                                 trampoline_onion=trampoline_onion,
                                 fw_payment_key=fw_payment_key,
                             )
+
+                        if probe_only:
+                            return can_route
```
SomberNight, line 2051: So you are not actually probing? HTLCs are not sent, this is just testing if the offline path finder can create some routes at all.

E.g. if Alice is trying to pay Bob, this is just testing that Alice has enough outbound capacity on the direct channel, and that some path can be found in the graph from Alice to Bob based on the gossip and the invoice r hints only. If Alice is using trampoline, it is checking even less, as Alice does not even have the graph.

Consider that Bob has a channel with the ACINQ node, and Alice has a channel with the  "Electrum Trampoline" node. Bob likely only generates a bolt11 invoice if the Bob<->ACINQ channel has liquidity. Similarly, Alice knows if the Alice<->"Electrum Trampoline" channel has liquidity. Whether there is liquidity on any path from "Electrum Trampoline" to ACINQ is not checked here.

MartyIX: > So you are not actually probing? HTLCs are not sent, this is just testing if the offline path finder can create some routes at all.

Yes, the check is very basic. It was somewhat intentional to get some feedback.

> E.g. if Alice is trying to pay Bob, this is just testing that Alice has enough outbound capacity on the direct channel, and that some path can be found in the graph from Alice to Bob based on the gossip and the invoice r hints only. If Alice is using trampoline, it is checking even less, as Alice does not even have the graph.
> 
> Consider that Bob has a channel with the ACINQ node, and Alice has a channel with the "Electrum Trampoline" node. Bob likely only generates a bolt11 invoice if the Bob<->ACINQ channel has liquidity. Similarly, Alice knows if the Alice<->"Electrum Trampoline" channel has liquidity. Whether there is liquidity on any path from "Electrum Trampoline" to ACINQ is not checked here.

Thank you for the elaboration. Please help me understand if you find this to be a hard requirement for this PR, or not. I'm not sure how to add this behavior at the moment (perhaps @dejfcz does). If you do know how to add it, any pointer would be very helpful and much appreciated. It would save us time. :)

dejfcz: > some path can be found in the graph from Alice to Bob based on the gossip 

Yes. The suggestion is to start with something trivial, which can later be improved by someone with deeper knowledge of Electrum code base. As it is now, it does prevent accepting swap in some scenarios, but of course full probing would be much better. So it is more about setting up the interface (probe_only parameter) for now.

PR #10651, opened 2026-05-15
ci: replace Cirrus with Github Actions

f321x:
* `secrets.COVERALLS_REPO_TOKEN` needs to be set for `tests.yml`
* `secrets.CROWDIN_API_KEY` needs to be set for `locale.yml`
* `secrets.CLAUDE_CODE_OAUTH_TOKEN` needs to be set for `security-review.yml`

I tested `tests.yml` and `regtest.yml` on my fork ([here](https://github.com/f321x/electrum-fork/actions)).

I disabled the non-mandatory flake8 check as it would always fail as our codebase is not adapted to the lints, there doesn't seem to be a "neutral/grey" status in github actions like we had with Cirrus, and both the red cross (which marks the whole PR red) and the green checkmark would be somewhat misleading. AFAIK no one ever really looked at these non-mandatory lints anyway?

I also dropped the unused intel based MacOS build which was commented out in the cirrus definition.

We should also monitor if the scheduled binary builds succeed during the night.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,207 @@
+name: tests
+
+on:
+  push:
+    branches: [master]
+    tags: ['*']
+  pull_request:
+  workflow_dispatch:
```
SomberNight, line 8: Say `Eve` opens a PR against `spesmilo`.

What are the semantics of `secrets`-access for the `workflow_dispatch` trigger?
On the `pull_request` trigger, Eve's `secrets` will be used (as discussed last time).
On the `push` trigger, Eve contextually does not exist, so `spesmilo`'s secrets will be used.
On the `pull_request_target` trigger, spesmilo's secrets will be used.

What about `workflow_dispatch`? Is it Eve's secrets?

.github/workflows/regtest.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Start electrumx
+        run: nohup tests/regtest/run_electrumx.sh > /tmp/electrumx.log 2>&1 &
+      - name: Run regtest tests
+        run: |
+          sleep 10
+          python3 -m unittest tests/regtest.py --failfast || TEST_EXIT_CODE=$?
+          tar -czf test_wallets.tar.gz /tmp/alice /tmp/bob /tmp/carol || true
+          exit ${TEST_EXIT_CODE:-0}
+      # if any test fails, the test will get aborted (--failfast) and the wallet directories will be
+      # available for download in the Cirrus UI
```
SomberNight: outdated comment "Cirrus UI"

.github/workflows/tests.yml
```diff
@@ -0,0 +1,207 @@
...
+          flake8 . --count \
+            --select="E9,E101,E129,E273,E274,E703,E71,E722,F5,F6,F7,F8,W191,W29,B,B909" \
+            --ignore="B007,B009,B010,B036,B042,F541,F841" \
+            --show-source --statistics \
+            --exclude "*_pb2.py,electrum/_vendor/"
+
+# Disabled because it lets the PRs CI run appear failed and the codebase isn't
+# adapted to these lints (so it will always fail).
+#  flake8-nonmandatory:
+#    name: "linter: Flake8 Non-Mandatory"
```
SomberNight: > I disabled the non-mandatory flake8 check as it would always fail as our codebase is not adapted to the lints, there doesn't seem to be a "neutral/grey" status in github actions like we had with Cirrus, and both the red cross (which marks the whole PR red) and the green checkmark would be somewhat misleading. AFAIK no one ever really looked at these non-mandatory lints anyway?

Ok, sounds good. I guess we could just delete the code, instead of commenting it out.

The original motivation of running the non-mandatory flake8 checks was that perhaps someone would take a look and slowly "fix" the issues it finds. But that's just not how things work out in practice :)

SomberNight (commented):
(reviewed 55fcd13ae5560f4c4b4d6d9c147415a31cbd2dfd)

I guess the conservative approach would be to first add GHA CI, merge it, run both GHA and Cirrus for a few days, and then rm Cirrus. But the Cirrus shutdown is imminent now.

.github/workflows/tests.yml
f321x: I cannot find any documentation about this specifically but i assume that `workflow_dispatch` runs in `spesmilo` context with `spesmilo` secrets. I assume this because you [need write access to trigger it](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow#configuring-a-workflow-to-run-manually) and it is declared independently of the `pull_request` target, similar to the `push:` directive.

.github/workflows/tests.yml
f321x: Removed in https://github.com/spesmilo/electrum/compare/55fcd13ae5560f4c4b4d6d9c147415a31cbd2dfd..a41c76f3f676e718a34c65448bf2a52cc7c23873

.github/workflows/regtest.yml
f321x: Updated in https://github.com/spesmilo/electrum/compare/55fcd13ae5560f4c4b4d6d9c147415a31cbd2dfd..a41c76f3f676e718a34c65448bf2a52cc7c23873

f321x:
> I guess the conservative approach would be to first add GHA CI, merge it, run both GHA and Cirrus for a few days, and then rm Cirrus.

I see the appeal but I also changed the security review script to use the github action environment variables (instead of the `CIRRUS_CI*` ones, so running both in parallel is requires more changes than just dropping the last commit, i don't think that's worth the effort at this point.

SomberNight:
Thanks for the work.
Let's see what broke :D

SomberNight:
oh right, we need to set the secrets.

SomberNight:
> * `secrets.COVERALLS_REPO_TOKEN` needs to be set for `tests.yml`
> * `secrets.CROWDIN_API_KEY` needs to be set for `locale.yml`
> * `secrets.CLAUDE_CODE_OAUTH_TOKEN` needs to be set for `security-review.yml`

I could not find a way to scope a secret for a specific .yml file.
Instead, all secrets are available to all .yml files, as "Repository secrets".

f321x:
> I could not find a way to scope a secret for a specific .yml file.

It might be possible to create an environment for certain secrets (e.g. `locale-env`), and then give a specific job access to this environment (by declaring `environment: locale-env` in the job).

PR #10653, opened 2026-05-18
lnwallet: fix rebalancing/pathfinding regression, add unittests 

f321x:
Don't exclude r_tags for frozen channels from the route creation in
create_route_for_single_htlc if the start_node of the routing hint is not our node id.

When doing a rebalance from us (chan_1) -> bob -> us (chan_2, frozen for sending),
we would exclude the invoice r_tag for chan_2 because the chan
is our chan and frozen for sending, resulting in us being unable to find a route back to us trough bob.

This is a regression from https://github.com/spesmilo/electrum/pull/9692 (https://github.com/spesmilo/electrum/commit/964ffbd2c441c155d568a5702e332b881cda9389).

Also prevents rebalancing if the required channel direction is frozen as this can lead to
opaque failures somewhere downstream (e.g. during pathfinding).

### DIFF
--- a/tests/test_lnwallet.py
+++ b/tests/test_lnwallet.py
@@ -3,13 +3,12 @@ import os
 import asyncio
 from unittest import mock
 from decimal import Decimal
+from typing import Optional
 
 from electrum.address_synchronizer import TX_HEIGHT_LOCAL
+from electrum import bitcoin
 import electrum.trampoline
-from . import ElectrumTestCase
-from .test_lnchannel import create_test_channels
-
-from electrum.lnutil import RECEIVED, MIN_FINAL_CLTV_DELTA_ACCEPTED, serialize_htlc_key, LnFeatures
+from electrum.lnutil import RECEIVED, MIN_FINAL_CLTV_DELTA_ACCEPTED, serialize_htlc_key, LnFeatures, HTLCOwner
 from electrum.logging import console_stderr_handler
 from electrum.lntransport import LNPeerAddr
 from electrum.invoices import LN_EXPIRY_NEVER, PR_UNPAID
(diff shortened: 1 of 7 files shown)


tests/test_lnwallet.py
```diff
@@ -19,6 +21,46 @@
...
+        local_msat=local_msat,
+        remote_msat=remote_msat,
+    )
+    peer_ab.lnworker._add_channel(chan_ab)
+    peer_ba.lnworker._add_channel(chan_ba)
+    chan_ab._state = ChannelState.FUNDED
+    chan_ba._state = ChannelState.FUNDED
+    peer_ab.mark_open(chan_ab)
+    peer_ba.mark_open(chan_ba)
+    return chan_ab, chan_ba
```
SomberNight: I understand why you are adding this, but I don't like it.

It is mostly just duplicating `test_lnpeer.TestPeer.prepare_chans_and_peers_in_graph`. We should de-dupe it somehow.

We could split that off into a new module and reuse it. Except we also need to extend `graph_definition` to allow multiple direct channels between a given pair of nodes - but that could come in handy for other things too anyway.

I hacked together something along those lines:
https://github.com/spesmilo/electrum/compare/master...SomberNight:electrum:202605_tests_lnhelpers_2 (b8491bdaf1d6b2270e5e56b8e65b6afd4552ab92)

but my approach is also a bit ugly :(

idk, what do you think?

f321x: Yes I was already considering including the test in `test_lnpeer.py` but it felt misplaced to test the `LNWallet` logic there. 

I looked at your approach, i think it makes sense to have a shared helper module as there are quite some similarities between the setups required in `test_lnpeer` and `test_lnwallet`. I cherry-picked it into this branch so we can move forward with this.

Should we move the graph definitions and the `create_test_channels()` function into `lnhelpers.py` as well? Or did you consider them too specific to move them into `lnhelpers`?

SomberNight: > Should we move the graph definitions

atm they are mostly used from lnpeer, we could wait to see if it spreads more

> Should we move [...] `create_test_channels()`

you're right, that seems to be used from all over

SomberNight: I'll move more stuff

SomberNight: (ref https://github.com/spesmilo/electrum/pull/10660)

PR #10654, opened 2026-05-18
channel: give peer some time to fail htlc offchain on startup

f321x:
Resolves this path of unnecessary force closes:
1. Alice sends out HTLC through Bob
2. Alice closes wallet with pending HTLC
3. HTLC gets failed back to Bob
4. HTLC times out, but Bob doesn't force close
5. Alice starts her wallet and immediately force closes

With this patch, Alice will instead give Bob 30 seconds at step 5 to fail the HTLC offchain before force closing the channel. This can prevent some force closes.

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1972,12 +1972,15 @@ class Channel(AbstractChannel):
         # If there is an offered HTLC which has already expired (+ some grace period after), we
         # will unilaterally close the channel and time out the HTLC
         offered_htlc_deadline_delta = lnutil.NBLOCK_DEADLINE_DELTA_AFTER_EXPIRY_FOR_OFFERED_HTLCS
+        time_since_startup = now() - self.lnworker.instantiation_timestamp
         for sub, dir, ctn in ((LOCAL, SENT, self.get_latest_ctn(LOCAL)),
                               (REMOTE, RECEIVED, self.get_oldest_unrevoked_ctn(REMOTE)),
                               (REMOTE, RECEIVED, self.get_latest_ctn(REMOTE)),):
             for htlc_id, htlc in self.hm.htlcs_by_direction(subject=sub, direction=dir, ctn=ctn).items():
                 if htlc.cltv_abs + offered_htlc_deadline_delta > local_height:
                     continue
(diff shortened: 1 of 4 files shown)


electrum/lnchannel.py
```diff
@@ -1972,12 +1972,15 @@ def should_be_closed_due_to_expiring_htlcs(self, local_height: int) -> bool:
...
         # will unilaterally close the channel and time out the HTLC
         offered_htlc_deadline_delta = lnutil.NBLOCK_DEADLINE_DELTA_AFTER_EXPIRY_FOR_OFFERED_HTLCS
+        time_since_startup = now() - self.lnworker.instantiation_timestamp
         for sub, dir, ctn in ((LOCAL, SENT, self.get_latest_ctn(LOCAL)),
                               (REMOTE, RECEIVED, self.get_oldest_unrevoked_ctn(REMOTE)),
                               (REMOTE, RECEIVED, self.get_latest_ctn(REMOTE)),):
             for htlc_id, htlc in self.hm.htlcs_by_direction(subject=sub, direction=dir, ctn=ctn).items():
                 if htlc.cltv_abs + offered_htlc_deadline_delta > local_height:
                     continue
+                if time_since_startup < lnutil.TIME_FOR_OFFERED_HTLCS_TO_GET_FAILED_OFFCHAIN_ON_RESTART:
```
f321x, line 1982: Note, there is no guarantee that LNWatcher will call this again anytime soon (similar case for received htlcs loop above).
I propose we add a simple asyncio task in LNWatcher that triggers the callbacks from time to time if they haven't been triggered for X minutes to have some guarantee of them getting called regularly.

SomberNight (approved):
Makes sense. Looks good. Thanks.

electrum/lnchannel.py
SomberNight: > Note, there is no guarantee that LNWatcher will call this again anytime soon (similar case for received htlcs loop above).

True.

> we add a simple asyncio task in LNWatcher that triggers the callbacks from time to time

Sure. The question is where to put the task, and e.g. how to notice if it dies. I guess LNWatcher should have its own taskgroup and we just reuse the existing patterns.

f321x: Doing this now in https://github.com/spesmilo/electrum/pull/10656

PR #10656, opened 2026-05-19
lnwatcher: introduce loop to trigger callbacks

f321x:
Introduce a taskgroup and polling loop to LNWatcher to guarantee
the callbacks get called at least once every
LNWatcher.MAX_CALLBACK_TRIGGER_DELAY_SEC (10 min).
This should prevent callbacks that operate on time instead of
blockheight from becoming (very) stale if there are no blockchain
events triggering the callbacks for a longer time.
Not entirely set about the 10 min delay, might as well be 5/2/1 min?

Related to https://github.com/spesmilo/electrum/pull/10654

### DIFF
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -2,11 +2,14 @@
 # Distributed under the MIT software license, see the accompanying
 # file LICENCE or http://www.opensource.org/licenses/mit-license.php
 
+import asyncio
 from typing import TYPE_CHECKING, Optional, Dict, Callable, Awaitable
 
 from . import util
-from .util import TxMinedInfo, BelowDustLimit, NoDynamicFeeEstimates
-from .util import EventListener, event_listener, log_exceptions, ignore_exceptions
+from .util import (
+    TxMinedInfo, BelowDustLimit, NoDynamicFeeEstimates, OldTaskGroup, EventListener, event_listener, log_exceptions,
+    ignore_exceptions, now
+)
 from .transaction import Transaction, TxOutpoint
 from .logging import Logger
 from .address_synchronizer import TX_HEIGHT_LOCAL
@@ -20,6 +23,8 @@ if TYPE_CHECKING:
 
 
 class LNWatcher(Logger, EventListener):
+    MAX_CALLBACK_TRIGGER_DELAY_SEC = 600
+    CALLBACK_LOOP_POLL_INTERVAL_SEC = 5
 
     def __init__(self, lnworker: 'LNWallet'):
         self.lnworker = lnworker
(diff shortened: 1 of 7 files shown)


electrum/onion_message.py
```diff
@@ -45,9 +45,9 @@
 from electrum.util import OldTaskGroup, log_exceptions, random_shuffled_copy
 
 
-def now():
+def now() -> float:
     return time.time()
-
+assert type(now()) == float, "OnionMessageManager requires float timestamps"
```
SomberNight, line 50: Right... that's part of why so far there has not been a global "now" function. Different use cases might want different behaviour.
For example, in several cases it would probably be more correct to use `time.monotonic()`.

but ok, sure, just noting all that. The diff is fine.

electrum/lnwatcher.py
```diff
@@ -30,13 +35,40 @@ def __init__(self, lnworker: 'LNWallet'):
...
+        """
+        Triggers the callbacks if no event has triggered them within the
+        last MAX_CALLBACK_TRIGGER_DELAY_SEC
+        (e.g. during a prolonged time without new blocks)
+        """
+        while True:
+            time_since_last_cb_trigger = now() - self._last_callback_trigger_ts
+            if time_since_last_cb_trigger > self.MAX_CALLBACK_TRIGGER_DELAY_SEC:
+                await self.trigger_callbacks()
+            await asyncio.sleep(self.CALLBACK_LOOP_POLL_INTERVAL_SEC)
```
SomberNight, line 73: >  Not entirely set about the 10 min delay, might as well be 5/2/1 min?

Esp considering https://github.com/spesmilo/electrum/pull/10654, where it could happen that there are expired htlcs, but `should_be_closed_due_to_expiring_htlcs` decides to wait 30 secs, but then there is no new block for a long time and `trigger_callbacks` would only get called after 10 mins, note that the user might close their wallet without waiting 10 mins. In fact it is perfectly believable that some users ~never leave the wallet open for 10 mins; and blocks would be processed within the first few seconds of wallet-open.

What do you think about something like this?
```diff
$ git diff
diff --git a/electrum/lnwatcher.py b/electrum/lnwatcher.py
index cb3cb9dbf..55bad7778 100644
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -63,9 +63,13 @@ class LNWatcher(Logger, EventListener):
         last MAX_CALLBACK_TRIGGER_DELAY_SEC
         (e.g. during a prolonged time without new blocks)
         """
+        t0 = now()
         while True:
             time_since_last_cb_trigger = now() - self._last_callback_trigger_ts
-            if time_since_last_cb_trigger > self.MAX_CALLBACK_TRIGGER_DELAY_SEC:
+            max_delay = self.MAX_CALLBACK_TRIGGER_DELAY_SEC
+            if now() - t0 < max_delay:  # if wallet just recently opened, be much more eager
+                max_delay /= 10
+            if time_since_last_cb_trigger > max_delay:
                 await self.trigger_callbacks()
             await asyncio.sleep(self.CALLBACK_LOOP_POLL_INTERVAL_SEC)
 
```

SomberNight (commented):
ACK 3254bcdf7527b3b6523d9035efee2d04f49548a5. Just some comments.

electrum/lnwatcher.py
f321x: Yes this seems like a good enhancement, added in https://github.com/spesmilo/electrum/compare/3254bcdf7527b3b6523d9035efee2d04f49548a5..2460b030ff109cf40bb309273335f415ad577722

electrum/onion_message.py
f321x: yeah, thats a little footgun, but also, using `now()` seems cleaner than using `int(time.time())` everywhere and it felt a bit wrong to define yet another `now()`.

Issue #4183, opened 2018-03-27
Very slow processing many transactions


alexandrz:
Opening a big wallet containing many transactions, as well as doing any action like switching tabs, opening transaction preview, closing it, changing transaction amount etc freezes Electrum for several minutes. It now takes about 10 minutes to open a wallet, create a transaction and broadcast it. It was fast before.

Electrum 3.0.5, MacOS

bauerj:
Can you be a bit more specific about the number of transactions?

AbdussamadA:
Upgrade to 3.1 There have been performance improvements since 3.0.5

alexandrz:
How can I get the number of transaction using console? I only know how to get the number of addresses which is about 6000.

alexandrz:
@AbdussamadA I've made a mistake, I am already using 3.1.1.

alexandrz:
@AbdussamadA Thank you, it's faster now. It still takes some time to open the wallet though, and it slower than it used to be. I downloaded 3.1.1 two weeks ago, and forget to install.

AbdussamadA:
You can create a new wallet via file > new/restore and move your coins there. Then once that transaction has confirmed perhaps you can share your old wallet with the developers so that they can look at optimizing Electrum for large wallets? I know @SomberNight was looking for a large wallet. I think even a watch only wallet will suffice.

alexandrz:
@AbdussamadA I can share a master public key to some developers. @SomberNight are you interested?

SomberNight:
Yes I am. An xpub would be welcome. See email in my github profile.

PR #10658, opened 2026-05-20
wallet: cache paid-invoice keys to fix GUI freeze after broadcast

sashazykov:
Fix for a long-standing GUI freeze when broadcasting a transaction from a wallet with many already-paid outgoing invoices that share output scriptpubkeys.

This issue has bothered me personally for nearly ten years — I originally reported it back in 2018 in #4183. The wallet I use most has accumulated 500+ invoices over the years (typical for repeat payments to the same set of addresses), and every time I broadcast a new transaction the GUI would lock up for several seconds — up to a minute on bigger wallets. The workaround I'd been using is `wallet.clear_invoices(); wallet.save_db()` in the console, but that throws away history I'd rather keep.

## Symptom

After clicking *Send*, the entire Qt GUI becomes unresponsive for several seconds (tens of seconds on larger wallets). Scales with the number of stored invoices, not with the size of the new transaction.

## Root cause

`_invoices_from_scriptpubkey_map` maps `scriptpubkey -> set[invoice_key]`. When many invoices share an output scriptpubkey (very common in real wallets — repeat payments to the same destination), this set is large. On broadcast:

1. `set_broadcasting(tx, PR_BROADCASTING)` iterates every touched invoice and calls `get_invoice_status` on each.
2. `_update_onchain_invoice_paid_detection` runs the same loop again after the tx is added to the address synchronizer.
3. `set_broadcasting(tx, PR_BROADCAST)` runs it a third time on success.

Each `get_invoice_status` on an onchain invoice calls `_is_onchain_invoice_paid`, which scans every output scripthash against `_prevouts_by_scripthash` and does a `get_tx_height` per prevout. With shared inputs/outputs and many past payments, both inner dimensions blow up together. All of this runs on the GUI thread.

## Fix

Two commits:

**`wallet: cache paid-invoice keys to avoid prevout rescan`**
Adds `_paid_invoice_keys: Set[str]` to `Abstract_Wallet`, an in-memory cache of outgoing invoice ids known to be `PR_PAID`, maintained incrementally:
- populated at wallet load by `_prepare_onchain_invoice_paid_detection`
- updated in `_update_onchain_invoice_paid_detection` (discard then recompute, so reorgs still demote PAID->UNPAID)
- updated in `save_invoice`, `delete_invoice`, `clear_invoices`, `clear_history`
- new `on_event_invoice_status` listener keeps it in sync with LN-driven transitions from `LNWallet.set_invoice_status`

Both `save_invoice` and `_update_onchain_invoice_paid_detection` reuse the result of the existing `(is_paid, conf)` computation instead of triggering a second prevout scan via `get_invoice_status`.

`get_invoice_status` short-circuits to `PR_PAID` on cache hit. `set_broadcasting` skips already-paid invoices entirely — `broadcasting_status` has no effect on a paid invoice and the callback churn is what made the GUI freeze.

**`qt/send_tab: drop redundant invoice_list.update() after broadcast`**
The explicit full rebuild after a successful broadcast is unnecessary: the `set_broadcasting(PR_BROADCAST)` call immediately after fires the `invoice_status` callback for any touched invoices, and `main_window.on_event_invoice_status` updates the affected rows incrementally. Tx ingestion also sets `need_update`, which causes `update_tabs()` to refresh.

## Testing

**Unit tests:** added `TestOutgoingInvoicesPaidCache` (9 tests) covering: empty/populated cache state, invoice-paid event flow, delete/clear cleanup, the short-circuit in `get_invoice_status`, `set_broadcasting` skipping cached-paid invoices, the shared-scriptpubkey regression scenario, cache population at wallet load, **cache demotion on reorg** (verified tx unverified), and **cache reset on `clear_history`**.

```
$ python -m unittest tests.test_invoices tests.test_wallet tests.test_wallet_vertical
Ran 128 tests in 1.65s
OK
```

Also ran `tests.test_lnpeer` and `tests.test_lnwallet` — 143/143 pass.

**Real-world testing:** I ran a patched build on macOS against my actual long-lived wallet — the one with the 500+ invoices that motivated the fix. The freeze is gone; sending a transaction is now instant.

## Notes for reviewers

- The fix is fully invariant-preserving: every status transition path (including reorg-driven PAID->UNPAID via `on_event_adb_removed_verified_tx`, and history wipe via `clear_history`) still flows through code that discards the cache entry before any subsequent read. Both transitions have explicit regression tests.
- Thread-safety: `_paid_invoice_keys` is only ever touched with single set operations (`add` / `discard` / `in` / `clear`), no iteration, which is safe under the GIL.
- The `on_event_invoice_status` listener firing on the wallet's own callbacks is idempotent and intentional (defense in depth for any future callsite that fires `invoice_status` without going through the existing paths).
- Wallet load itself is incidentally a bit faster: `_update_onchain_invoice_paid_detection` now derives status from the already-computed `(is_paid, conf_needed)` tuple instead of calling `get_invoice_status`, which previously re-ran `is_onchain_invoice_paid` — halving the per-invoice prevout work in `_prepare_onchain_invoice_paid_detection`.

### DIFF
--- a/electrum/gui/qt/send_tab.py
+++ b/electrum/gui/qt/send_tab.py
@@ -809,7 +809,6 @@ class SendTab(QWidget, MessageBoxMixin, Logger):
                 success, msg = result
                 if success:
                     parent.show_message(_('Payment sent.') + '\n' + msg)
-                    self.invoice_list.update()
                     self.wallet.set_broadcasting(tx, broadcasting_status=PR_BROADCAST)
                 else:
                     msg = msg or ''
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -422,6 +422,8 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         self.lock = self.adb.lock
         self._last_full_history = None
         self._tx_parents_cache = {}
+        self._paid_invoice_keys_cache = set()  # type: Set[str]
+        self._coin_price_cache = {}
         self._default_labels = {}
         self._accounting_addresses = set()  # addresses counted as ours after successful sweep
 
@@ -445,15 +447,12 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         self.txbatcher = TxBatcher(self)
         self._init_lnworker()
         self._init_requests_rhash_index()
-        # cache of outgoing-invoice ids known to be PR_PAID, to skip the prevout scan
-        self._paid_invoice_keys = set()  # type: Set[str]
         self._prepare_onchain_invoice_paid_detection()
         self._calc_unused_change_addresses()
         # save wallet type the first time
         if self.db.get('wallet_type') is None:
             self.db.put('wallet_type', self.wallet_type)
         self.contacts = Contacts(self.db)
-        self._coin_price_cache = {}
 
         # true when synchronized. this is stricter than adb.is_up_to_date():
         # to-be-generated (HD) addresses are also considered here (gap-limit-roll-forward)
@@ -632,7 +631,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         self.clear_tx_parents_cache()
         if self.lnworker:
             self.lnworker.maybe_add_backup_from_tx(tx)
-        self._update_invoices_and_reqs_touched_by_tx(tx_hash)
+        self._update_invoices_and_reqs_touched_by_tx(tx)
         util.trigger_callback('new_transaction', self, tx)
 
     @event_listener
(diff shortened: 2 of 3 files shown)


sashazykov:
Rebased onto current master.

I've been running a patched build with these two commits for the past two weeks. The broadcast freeze is gone and I haven't observed any regressions in normal day-to-day use.

f321x:
Looks reasonable to me, also tested for some time on Android. 
I added a followup commit https://github.com/spesmilo/electrum/pull/10658/commits/4f89720087e8c8c9965e615183b3a2ced5eb8cbc with some minor renaming and refactoring.
Thanks for the PR.

SomberNight (commented):
Even after Felix's changes, I believe I still simplified the diff quite a bit (in 63ee2f3a60cad514074c7b83261871d43393cdfb).
Your new unit tests still pass. Please re-test if your performance regression is still fixed.

Congrats on baiting two people to make sense of the slop :/

(but I understand you just want your old bug fixed)

sashazykov:
Thanks a lot for the cleanup — routing cache population through `on_event_invoice_status` is much nicer than my scattered version, and good catch on the removed-tx path.

One thing I noticed: dropping the inline status derivation means `_update_onchain_invoice_paid_detection` now scans prevouts twice per invoice (once for `relevant_txs`, then again via `get_invoice_status` after the cache discard). It's off the GUI-freeze path so probably fine, but adds some cost at load on wallets with many invoices — flagging in case it matters.

I'll test against my wallet and report back. Thanks again!

sashazykov:
Tested it against my wallet. There is no broadcast freeze.

PR #10663, opened 2026-05-21
unit tests: do not mock save_db

ecdsa:
Create in-memory wallets instead.

The logic of calling save_db() at specific places is specific to the current JsonDB backend, and it does not generalize well.


SomberNight (approved):
Sounds good.

PR #10665, opened 2026-05-22
lnpeer: add some rate-limiting against ping flood

SomberNight:
A remote peer could send us lots of small ping messages, requesting large pong responses. This is cheap for them but potentially expensive for us. Does not seem too serious, but I think we could add some rate-limiting.

note: There are many ways for a remote peer to inflate our incoming traffic usage, but the cost of that is usually shared between them and us (they need to send the data, we receive it).

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -386,7 +386,8 @@ class Peer(Logger, EventListener):
             await asyncio.sleep(min_delay - elapsed_since_last)
         self._last_ping_recv_time = time.monotonic()
         l = payload['num_pong_bytes']
-        self.send_message('pong', byteslen=l)
+        raw_msg = encode_msg('pong', byteslen=l)
+        await self.transport.send_bytes_and_drain(raw_msg)
 
     def on_pong(self, payload):
         self.pong_event.set()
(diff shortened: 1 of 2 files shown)


electrum/lnpeer.py
```diff
@@ -371,7 +372,18 @@ def send_error(self, channel_id: bytes, message: str = None, *, force_close_chan
...
+            self.logger.debug("remote sending PINGs too often, sleeping a bit")
+            # note: This rate-limiting helps limit our outbound traffic usage.
+            #       (max inc msg size 65 KB, max out msg size 65 KB, decoupled)
+            #       Does not really help for inbound traffic-usage:
+            #       there are many other ways for the peer to flood us, e.g. unknown 'odd' message types.
+            # note: this blocks processing *any* further incoming message from this peer
+            await asyncio.sleep(min_delay - elapsed_since_last)
+        self._last_ping_recv_time = time.monotonic()
         l = payload['num_pong_bytes']
         self.send_message('pong', byteslen=l)
```
f321x: Maybe it would be good to drain the transport like this as well?
```suggestion
        msg = encode_msg('pong', byteslen=l)
        await self.transport.send_bytes_and_drain(msg)
```
Otherwise, maybe a peer could just fill up our sockets outgoing buffer by ~234mb/h if they stop reading from their socket and keep sending pings every second.

SomberNight: uuuuuu nice observation.
I don't know if such an attack would actually work though -- I would expect some kind of TCP timeout to get triggered from the OS. But maybe I am wrong?
In any case, let's do this, defense in depth.

SomberNight: > if they stop reading

I guess if they keep reading reeeeeeaaaalllyy slow, but send stuff at a steady pace, that might work, maybe.

PR #10668, opened 2026-05-26
lnonion/trampoline: stop double counting HMAC size

f321x:
`OnionHopsDataSingle.to_bytes()` already accounts for the onions hmac internally, 
it seems incorrect that we were additionally adding `PER_HOP_HMAC_SIZE` to the payload 
size of the payload returned by `.to_bytes()`.

https://github.com/spesmilo/electrum/blob/8e96fef1e1d6859204d918685e41c0d002b718a6/electrum/lnonion.py#L76-L92


SomberNight (approved):
Right.. Yes, that seems to have been a logic bug.

PR #10670, opened 2026-05-27
ci: builds: android: fix cache key

f321x:
Fix the cache key of the scheduled Android CI build which failed at the `Cache buildozer (p4a)` step with the following error:
```
Error: The template is not valid. .github/workflows/builds.yml (Line: 113, Col: 16): hashFiles('contrib/android/**') failed. Fail to hash files under directory '/home/runner/work/electrum/electrum'
```

Probably it failed because the `contrib/android/.cache` directory got included in the cache key, which worked on Cirrus. This specifies the cache key hash inputs explicitly.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -101,7 +101,7 @@ jobs:
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
         with:
           path: packages
-          key: android-packages-${{ env.APK_ARCH }}-${{ hashFiles('contrib/deterministic-build/requirements.txt', 'contrib/make_packages.sh', 'contrib/android/**') }}
+          key: android-packages-${{ env.APK_ARCH }}-${{ hashFiles('contrib/deterministic-build/**', 'contrib/make_packages.sh', 'contrib/android/**', '!contrib/android/.cache') }}
 
       - name: Cache buildozer (p4a)
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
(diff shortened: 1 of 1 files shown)


SomberNight:
> Probably it failed because the `contrib/android/.cache` directory got included in the cache key

I don't understand.
We use a similar pattern for the other builds, e.g.:
https://github.com/spesmilo/electrum/blob/b2d6f58b05ae4f8a6b9046f040dd6ae98d6f1b1b/.github/workflows/builds.yml#L145
Why did that one succeed?


If it's really about the `.cache` folder for some reason, we could just exclude that using the [`!` pattern from the GHA examples](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#examples-with-multiple-patterns).

f321x:
>  Why did that one succeed?

I think the Android build failed due to symlinks created by the barcode scanner build inside the `.cache` directory.
Yes, the exclude pattern is cleaner, using this now in https://github.com/spesmilo/electrum/compare/a0e6880d61f4b5f8d3f2d482d803e6c179036d7f..39e9e23eb97f8da8c6cdf2ef386c4df1f28c0634.

I verified the build works on my fork, see: https://github.com/f321x/electrum-fork/actions/runs/26628230417/job/78470164529

SomberNight:
> I think the Android build failed due to symlinks created by the barcode scanner build inside the .cache directory.

That sounds plausible.

OTOH the appimage build cache also contains symlinks.

see
```
$ find contrib/build-linux/appimage/.cache -type l -ls
$ find contrib/android/.cache -type l -ls
```
anyway, in this case, we don't have to root-cause it.

PR #10674, opened 2026-05-28
mpp_split: stop returning splits without amounts

f321x:
`mpp_split.suggest_splits()` would return payment split configurations with no split amounts like:
`{(channel_id, node_id): []}`.

This is unintuitive and introduced two bugs:

1. `LNWallet.create_routes_for_payment()` evaluates `is_multichan_mpp = len(sc.config.items()) > 1`. So even if the actual payment amount gets split onto a single channel this would falsely evaluate `True` if there is an empty, additional split.

2. `is_direct_path = all(node_id == paysession.invoice_pubkey for (chan_id, node_id) in sc.config.keys())` similarly might incorrectly evaluates `False` if the split contains an empty split for another `node_id`.

### DIFF
--- a/electrum/mpp_split.py
+++ b/electrum/mpp_split.py
@@ -143,11 +143,10 @@ def suggest_splits(
                 random.shuffle(channel_keys)
                 # we check each channel and try to put the funds inside, break if we succeed
                 for c in channel_keys:
-                    if c not in config:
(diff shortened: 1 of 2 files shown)


SomberNight (approved):
Nice catch. Looks good.

PR #10677, opened 2026-05-30
trezor: support Safe 7

romanz:
Trezor Safe 7 is using [Trezor-Host Protocol, for encrypted and authenticated communication](https://docs.trezor.io/trezor-firmware/common/thp/specification.html).

Since it requires a pairing process before device features are available, an additional `WCTrezorPair` wizard step is introduced, allowing the user to enter 6-digit pairing code (as shown on the device).

## TODOs:
- [x] [retest](https://github.com/spesmilo/electrum/pull/10465#issue-3902421542) on other devices

### DIFF
--- a/electrum/plugins/trezor/clientbase.py
+++ b/electrum/plugins/trezor/clientbase.py
@@ -11,9 +11,11 @@ from electrum.logging import Logger
 from electrum.plugin import runs_in_hwd_thread
 from electrum.hw_wallet.plugin import OutdatedHwFirmwareException, HardwareClientBase
 
-from trezorlib.client import TrezorClient, PassphraseSetting, get_default_client
+from trezorlib.client import TrezorClient, PassphraseSetting, AppManifest, get_client
 from trezorlib.exceptions import TrezorFailure, Cancelled, OutdatedFirmwareError
-from trezorlib.messages import WordRequestType, FailureType, ButtonRequestType, Capability
+from trezorlib.messages import WordRequestType, FailureType, ButtonRequestType, Capability, Features
+from trezorlib import models
+from trezorlib.thp.pairing import CodeEntry, ControllerLifecycle
 import trezorlib.btc
 import trezorlib.device
 
@@ -42,23 +44,36 @@ MESSAGES = {
     'default': _("Check your {} device to continue"),
 }
 
+# trezorlib THP does not support cross-thread cancellation.
+# https://github.com/trezor/trezor-firmware/issues/7112
+CANCEL_SUPPORTED = frozenset({models.T1B1, models.T2T1, models.T2B1, models.T3T1, models.T3B1})
 
 class TrezorClientBase(HardwareClientBase, Logger):
     def __init__(self, transport, handler, plugin):
         HardwareClientBase.__init__(self, plugin=plugin)
+        Logger.__init__(self)
+
         if plugin.is_outdated_fw_ignored():
             TrezorClient.is_outdated = lambda *args, **kwargs: False
 
-        self.client = get_default_client(
-            app_name="Electrum",
-            path_or_transport=transport,
-            button_callback=self.button_request,
-            pin_callback=self.get_pin,
-        )
         self._session = None
         self.device = plugin.device
         self.handler = handler
-        Logger.__init__(self)
+
+        self.transport = transport
+        self.app = AppManifest(
+            app_name="Electrum",
+            credentials=(),
+            button_callback=self.button_request,
+            pin_callback=self.get_pin,
+        )
+        self._client = None
+        # Makes sure the client is connected (on THP, a channel has been established)
+        model = self.client.model
+        if model.is_unknown:
+            self.logger.warning("Unknown Trezor model: %s", model)
+
+        # Pairing cannot be done during device enumeration, since UI handler is unset).
 
         self.msg = None
         self.creating_wallet = False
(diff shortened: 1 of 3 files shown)


electrum/plugins/trezor/qt.py
```diff
@@ -922,3 +923,50 @@ def initialize_device_task(settings, method, device_id, handler):
...
     def apply(self):
         pass
+
+
+class WCTrezorPair(WalletWizardComponent, Logger):
+    def __init__(self, parent, wizard):
+        WalletWizardComponent.__init__(self, parent, wizard, title=_('Trezor Pairing'))
+        Logger.__init__(self)
+        self.plugins = wizard.plugins
+        self._busy = True
```
SomberNight, line 933: > Since it requires a pairing process before device features are available, an additional WCTrezorPair wizard step is introduced

Oh. So it works differently from the previous devices. In that case I would like to ask for 2 or 3 samples we can use for manual testing in the future when refactoring code. Maybe during btcprague? Then shipping could be avoided.

electrum/plugins/trezor/qt.py
```diff
@@ -922,3 +923,50 @@ def initialize_device_task(settings, method, device_id, handler):
...
+            try:
+                with client.run_flow(f"Confirm pairing with {client.device_model_name()}"):
+                    client.pair_if_needed()
+
+                if self.wizard_data['wallet_exists']:
+                    next_view = 'trezor_unlock'
+                else:
+                    next_view = ['trezor_not_initialized', 'trezor_start'][client.features.initialized]
+
+                self.wizard_data['_after_pairing'] = next_view
```
accumulator: don't determine the next view here. Instead, record state (paired, not paired), and determine next view in the plugin's wizard extension declaration.

romanz: > Instead, record state (paired, not paired), and determine next view in the plugin's wizard extension declaration.

https://github.com/spesmilo/electrum/pull/10677/commits/cff4a656ba05a7cd6f5bdd3c79a25be4e90d9a10 will keep initialization state in `wizard_data['trezor_initialized']` and choose the next view using `_after_pairing()` callback at `TrezorPlugin`.

Please let me know if you have meant something else.

romanz:
@SomberNight would it be OK to allow cancellation only for non-THP devices (to avoid bumping `trezorlib` due to https://github.com/trezor/trezor-firmware/issues/7112)?

Asking since the user can cancel the current flow also from the device itself.

electrum/plugins/trezor/clientbase.py
```diff
@@ -325,7 +375,8 @@ def wipe_device(self, *args, **kwargs):
 
     def button_request(self, br):
         message = self.msg or MESSAGES.get(br.code) or MESSAGES['default']
-        self.handler.show_message(message.format(self.device), self.client.cancel)
+        on_cancel = self.transport.interrupt and self.transport.interrupt.request
+        self.handler.show_message(msg=message.format(self.device), on_cancel=on_cancel)
```
SomberNight: @romanz 
> would it be OK to allow cancellation only for non-THP devices (to avoid bumping trezorlib due to https://github.com/trezor/trezor-firmware/issues/7112)?

You can disable/hide the cancel functionality if you think that's the most practical approach.
I guess that's better than not having support for a device at all because of something like this :P

Just add a code comment to document the reasoning.

> Asking since the user can cancel the current flow also from the device itself.

Was there no way to cancel on-device for the older devices?

romanz: > Was there no way to cancel on-device for the older devices?

Yes - older devices also allow the flows to be cancelled via button click / screen touch.

In the worst case, it is also possible to disconnect the device from the host - cancelling the flow by shutting down the device :)

romanz: Changed in https://github.com/spesmilo/electrum/compare/7469710c0ba511b7abdb4e1670f645081934a888..35de77662adbea42612fb7f5fe246a719a74da3f.

romanz:
[Squashed fixups](https://github.com/spesmilo/electrum/compare/35de77662adbea42612fb7f5fe246a719a74da3f..1547c5b4c1985fcc6235d903e01fda1a45cd8782) into a single commit.

electrum/plugins/trezor/qt.py
```diff
@@ -922,3 +923,45 @@ def initialize_device_task(settings, method, device_id, handler):
...
+
+        def pair_task(client):
+            try:
+                with client.run_flow(f"Confirm pairing with {client.device_model_name()}"):
+                    client.pair_if_needed()
+
+                self.wizard_data['trezor_initialized'] = client.features.initialized
+                self.wizard.requestNext.emit()  # triggers Next GUI thread from event loop
+            except Exception as e:
+                self.error = repr(e)  # TODO: handle user interaction exceptions (e.g. invalid pin) more gracefully
```
f321x: Here, and in `WCTrezorInit` above you could display `str(e)` instead of `repr(e)` to display the exception in a more readable way:

Image

Image

romanz: Thanks! Fixed in 154d79ebbf6840843bd74e621c3b81b09ac313b8.

f321x:
Tested with Safe 7 and Safe 3, lgtm.

SomberNight:
Thanks

PR #10680, opened 2026-06-02
config: add option to disable memory hardening

ecdsa:
This is needed for running with a memory profiler

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -2462,6 +2462,10 @@ def add_global_options(parser, suppress=False):
     group.add_argument(
         "--forgetconfig", action="store_true", dest=SimpleConfig.CONFIG_FORGET_CHANGES.key(), default=None,
         help=argparse.SUPPRESS if suppress else "Forget config on exit")
+    group.add_argument(
+        # Note: default value is False and not None, so that behaviour cannot be modified by editing the config file
+        "--nohardening", action="store_true", dest=SimpleConfig.DISABLE_MEMORY_HARDENING_LINUX.key(), default=False,
+        help=argparse.SUPPRESS if suppress else "Disable memory hardening (linux)")
 
 
 def get_simple_parser():
--- a/run_electrum
+++ b/run_electrum
@@ -408,6 +408,7 @@ def main():
             'cmd': 'gui',
             SimpleConfig.GUI_NAME.key(): 'qml',
             SimpleConfig.WALLET_SHOULD_USE_SINGLE_PASSWORD.key(): True,
+            SimpleConfig.DISABLE_MEMORY_HARDENING_LINUX.key(): True,
         }
         SimpleConfig.set_chain_config_opt_based_on_android_packagename(config_options)
     else:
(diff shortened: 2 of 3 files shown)


run_electrum
```diff
@@ -458,7 +459,7 @@ def main():
         print_stderr('unknown command:', uri)
         sys.exit(1)
 
-    if sys.platform == "linux" and not is_android:
+    if sys.platform == "linux" and not config.DISABLE_MEMORY_HARDENING_LINUX:
```
SomberNight, line 462: what is our threat model for the memory hardening?
looks like the same hostile process we are trying to defend against could just turn off the hardening by writing into the config file

SomberNight: threat model is explicitly to also protect against other processes running as same uid:
https://github.com/spesmilo/electrum/blob/f29ef6f68cbd96900f3432413e7d55e62082839f/electrum/harden_memory_linux.py#L8-L11

electrum/commands.py
```diff
@@ -2462,6 +2462,10 @@ def add_global_options(parser, suppress=False):
     group.add_argument(
         "--forgetconfig", action="store_true", dest=SimpleConfig.CONFIG_FORGET_CHANGES.key(), default=None,
         help=argparse.SUPPRESS if suppress else "Forget config on exit")
+    group.add_argument(
+        # Note: default value is False and not None, so that behaviour cannot be modified by editing the config file
+        "--disable_hardening", action="store_true", dest=SimpleConfig.DISABLE_MEMORY_HARDENING_LINUX.key(), default=False,
```
SomberNight: for the CLI flags specificly, if we want to separate words, the usual convention is to use the `-` char: for example this should be `--disable-hardening` then

electrum/commands.py
```diff
@@ -2462,6 +2462,10 @@ def add_global_options(parser, suppress=False):
     group.add_argument(
         "--forgetconfig", action="store_true", dest=SimpleConfig.CONFIG_FORGET_CHANGES.key(), default=None,
         help=argparse.SUPPRESS if suppress else "Forget config on exit")
+    group.add_argument(
+        # Note: default value is False and not None, so that behaviour cannot be modified by editing the config file
+        "--nohardening", action="store_true", dest=SimpleConfig.DISABLE_MEMORY_HARDENING_LINUX.key(), default=False,
+        help=argparse.SUPPRESS if suppress else "Disable memory hardening (linux)")
```
SomberNight, line 2468:
```suggestion
        help=argparse.SUPPRESS if suppress else SimpleConfig.DISABLE_MEMORY_HARDENING_LINUX.get_short_desc())
```

run_electrum
SomberNight: ok, setting a hard-default in commands.py so that the value is always set on the CLI (which has priority over the config) works.

SomberNight:
ACK 22e0f03c5cf5c090c9fb236079fa47d0d4248937
but feel free to do the comment dedupe in addition. or not.

electrum/commands.py
ecdsa: actually, it would be more consistent to keep the help in commands.py and to remove it from simple_config.py. That is what we already do with other config vars that are passed on the cmdline, such as `rpcpassword`

PR #10686, opened 2026-06-04
Modify the 'stored_at' syntax, so that it includes the full path

ecdsa:
This may prevent collisions. More importantly, it also makes it possible to cache the subdict of 'registered_names' that is relevant to a given StoredDict, so that we do not need to walk the path on every conversion. (performance optimization, not part of this commit)

This is a backport from the levelDB branch

### DIFF
--- a/electrum/stored_dict.py
+++ b/electrum/stored_dict.py
@@ -43,23 +43,25 @@ def locked(func):
     return wrapper
 
 
-registered_names = defaultdict(dict)
-registered_keys = defaultdict(dict)
+registered_names = {}
+registered_keys = {}
 
-def _parse_path(path):
-    path2 = path.split('/')
-    name, suffix = path2[0], path2[1:]
-    n = len(suffix)
-    assert suffix == n * ['*']
-    return name, n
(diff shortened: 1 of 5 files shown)


electrum/stored_dict.py
```diff
@@ -54,8 +54,14 @@ def _parse_path(path):
     return name, n
 
 def register_name(path, _type, func):
-    name, n = _parse_path(path)
-    registered_names[name][n] = _type, func
+    path = path.split('/')
+    path, key = path[0:-1], path[-1]
```
SomberNight: so does `register_name` always take an absolute path now, but `register_key` still takes a relative path (always)? I find this a bit confusing.

Can we assert at the beginning of `register_name()` that the `path` str starts with a `/` character, to signal it is an abs path? And then also update all the callers to prepend that to the path to make this clear.

ecdsa: Good points. I modified `register_key` in a similar way.
Also added a heading ` / ` to all paths.

SomberNight: The new interface and the concept looks good.
(but I haven't reviewed the correctness of the implementation)

PR #10687, opened 2026-06-04
wallet: do not call synchronize() from event handler

ecdsa:
If the wallet is online, synchronize is already called from do_synchronize_loop, which seems sufficient. Most importantly, this task is stopped with the taskgroup, before the wallet database is closed. This order is not guaranteed for event-driven actions: we may try to write to the database after the wallet is stopped.

backport from levelDB branch


ecdsa:
closing: not needed

PR #10689, opened 2026-06-05
plugin: nwc: regularly rebroadcast info event

f321x:
I noticed with my long running NWC daemon that after a
couple days some clients relying on the info event being
available stopped working. Alby e.g. shows
"Error: no info event (kind 13194) returned from relay".
After restarting the daemon it works again for some days.
It seems that relays drop the info events after some days,
even though they are not ephemeral and don't have an expiry tag.
So rebroadcasting them once a day should make the NWC server more
reliable.


ecdsa:
lgtm

PR #10692, opened 2026-06-06
create_channel_storage: return dict instead of StoredDict

ecdsa:
we should avoid dangling StoredDict; this does not make sense if the DB is not in memory.

(a unit test with dangling StoredDict will be added in the new_storage_api branch)


SomberNight (approved):
Looks good -- though the flow is a bit weird (`lnworker.add_new_channel`) :/
Still I could not see a way to simplify it.

Added a boring follow-up commit.

PR #10693, opened 2026-06-07
New storage api

ecdsa:
I am marking this as ready because it is advanced enough that it could use some review.

My hope was to make the first commit pass all the tests, but that seems to be impossible. Commit 4 (types conversions) restores functionality. Note that commit 2 (perf optimization) and 3 (rename) are	trivial, and they could	potentially be reordered after commit 4.

There are 3 remaining design issues:

   1. should `StoredDict`.get() return `dict`/`list` instead of `StoredList`/`StoredDict`	?
       - pro: `get`() would be consistent with `pop`()
       - con: `get`() would be inconsistent with `__getitem__` , which must return `StoredList`/`StoredDict`
       - pro: in order to access/create `StoredDict`/`StoredList`, we would move `get_dict`, `get_list` from `wallet_db` to `stored_dict`. This would make the returned types more explicit.
       - pro: this would simplify DB upgrades, and code in general
      	
   2. should we remove the `init_db` parameter in the constructor, and use an explicit `DictStorage.open`() call instead?
       - probably cleaner, there is already a `close`() method
       - `open` would take the password as parameter, and behave as `decrypt`(). Howver, it would need to be called even for non-encrypted wallets.

   3.  currently if an exception is raised during a batch write, memory and disk become inconsistent (see FIXME)
       - ideally, we should get rid of `save_db`, and always write to disk if we are not in a batch.
       - this would be easy if we had partial writes always enabled (not implemented yet with encrypted wallets)

EDIT: point 3 is now fixed in #10339

### DIFF
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -30,9 +30,9 @@ from typing import TYPE_CHECKING, Optional, Sequence, List, Union, Dict, Any
 import jsonpatch
 import jsonpointer
 
-from .util import WalletFileException, profiler, sticky_property, MyEncoder
+from .util import WalletFileException, profiler, sticky_property
 from .logging import Logger
-from .stored_dict import _FLEX_KEY, BaseDB, _convert_dict_key, _convert_dict_value
+from .stored_dict import _FLEX_KEY, BaseDB
 from .storage import FileStorage
 
 
@@ -89,12 +89,10 @@ class JsonDB(BaseDB):
             *,
             allow_partial_writes = True,
             init_db = True,
-            encoder = MyEncoder,
     ):
         BaseDB.__init__(self, path)
         self._is_closed = True
         self.lock = threading.RLock()
-        self.encoder = encoder
         self.pending_changes = []  # type: List[str]
         self._modified = False
         if self.path:
@@ -105,18 +103,17 @@ class JsonDB(BaseDB):
                 self.init_db()
         else:
             self.storage = None
-            self.set_data('{}')
+            self.json_data = {}
             self._is_closed = False
 
     def set_data(self, json_str):
-        data = self.load_data(json_str)
-        self.json_data = self._convert_dict([], data)
+        self.json_data = self.load_data(json_str)
 
     def init_db(self):
         if self.storage.is_encrypted():
             assert self.storage.is_past_initial_decryption()
         json_str = self.storage.read()
-        self.set_data(json_str)
+        self.json_data = self.load_data(json_str)
         # write file in case there was a db upgrade
         self.write_and_force_consolidation()
         self._is_closed = False
@@ -286,7 +283,7 @@ class JsonDB(BaseDB):
 
     @locked
     def add_patch(self, patch):
-        self.pending_changes.append(json.dumps(patch, cls=self.encoder))
+        self.pending_changes.append(json.dumps(patch))
         self.set_modified(True)
 
     def db_add(self, path, key: _FLEX_KEY, value) -> None:
@@ -310,28 +307,7 @@ class JsonDB(BaseDB):
             self.json_data,
             indent=4 if human_readable else None,
             sort_keys=bool(human_readable),
-            cls=self.encoder,
         )
-
-    def _convert_dict_key(self, path: List[str], key: str) -> _FLEX_KEY:
-        return _convert_dict_key(path, key)
-
-    def _convert_dict_value(self, path: List[str], v) -> Any:
-        v = _convert_dict_value(path, v)
-        if isinstance(v, dict):
-            v = self._convert_dict(path, v)
-        return v
-
-    def _convert_dict(self, path: List[str], data: dict):
-        # recursively convert json dict to StoredDict
-        assert all(isinstance(x, str) for x in path), repr(path)
-        d = {}
-        for k, v in list(data.items()):
-            child_path = path + [k]
-            k = self._convert_dict_key(path, k)
-            v = self._convert_dict_value(child_path, v)
-            d[k] = v
-        return d
 
     @locked
     def write(self):
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -30,14 +30,11 @@ from typing import TYPE_CHECKING, Optional, Sequence, List, Union, Dict, Any
 import jsonpatch
 import jsonpointer
 
-from . import util
-from .util import WalletFileException, profiler, sticky_property
+from .util import WalletFileException, profiler, sticky_property, MyEncoder
 from .logging import Logger
-from .stored_dict import StoredDict, _FLEX_KEY, registered_names, registered_keys, _convert_dict_key, _convert_dict_value
+from .stored_dict import _FLEX_KEY, BaseDB, _convert_dict_key, _convert_dict_value
+from .storage import FileStorage
 
-
-if TYPE_CHECKING:
-    from .storage import WalletStorage
 
 
 # We monkeypatch exceptions in the jsonpatch package to ensure they do not contain secrets from the DB.
@@ -84,36 +81,154 @@ def locked(func):
 
 
 
-
-class JsonDB(Logger):
+class JsonDB(BaseDB):
 
     def __init__(
-        self,
-        s: str,
-        *,
-        storage: Optional['WalletStorage'] = None,
-        encoder=None,
-        upgrader=None,
+            self,
+            path: Optional[str],
+            *,
+            allow_partial_writes = True,
+            init_db = True,
+            encoder = MyEncoder,
     ):
-        Logger.__init__(self)
+        BaseDB.__init__(self, path)
+        self._is_closed = True
         self.lock = threading.RLock()
-        self.storage = storage
         self.encoder = encoder
         self.pending_changes = []  # type: List[str]
         self._modified = False
-        # load data
-        data = self.load_data(s)
-        if upgrader:
-            data, was_upgraded = upgrader(data)
-            self._modified |= was_upgraded
-        # convert json to python objects
-        data = self._convert_dict([], data)
-        # convert dict to StoredDict
-        self.data = StoredDict(data, self)
-        self.data.set_parent(key='', parent=None)
+        if self.path:
+            self.storage = FileStorage(path, allow_partial_writes=allow_partial_writes)
+            if init_db and not self.is_encrypted():
+                # open DB if file is not encrypted
+                # otherwise, this will be called in self.decrypt
+                self.init_db()
+        else:
+            self.storage = None
+            self.set_data('{}')
+            self._is_closed = False
+
+    def set_data(self, json_str):
+        data = self.load_data(json_str)
+        self.json_data = self._convert_dict([], data)
+
+    def init_db(self):
+        if self.storage.is_encrypted():
+            assert self.storage.is_past_initial_decryption()
+        json_str = self.storage.read()
+        self.set_data(json_str)
         # write file in case there was a db upgrade
-        if self.storage and self.storage.file_exists():
-            self.write_and_force_consolidation()
+        self.write_and_force_consolidation()
+        self._is_closed = False
+
+    def decrypt(self, password: str):
+        self.storage.decrypt(password)
+        json_str = self.storage.read()
+        self.set_data(json_str)
+        self._is_closed = False
+
+    def check_password(self, password):
+        self.storage.check_password(password)
+
+    def supports_file_encryption(self):
+        return bool(self.storage)
+
+    def get_encryption_version(self):
+        return self.storage.get_encryption_version()
+
+    def is_encrypted(self):
+        return self.storage and self.storage.is_encrypted()
+
+    def is_encrypted_with_user_pw(self) -> bool:
+        return self.storage and self.storage.is_encrypted_with_user_pw()
+
+    def is_encrypted_with_hw_device(self) -> bool:
+        return self.storage and self.storage.is_encrypted_with_hw_device()
+
+    def set_password(self, password: str, enc_version=None):
+        self.storage.set_password(password, enc_version=enc_version)
+
+    def file_exists(self):
+        return self.storage and self.storage.file_exists()
+
+    def _subdict(self, path):
+        d = self.json_data
+        for k in path[1:]:
+            d = d[k]
+        return d
+
+    def iter_keys(self, path):
+        d = self._subdict(path)
+        return d.__iter__()
+
+    def dict_len(self, path):
+        d = self._subdict(path)
+        return len(d)
+
+    def contains(self, path, key):
+        d = self._subdict(path)
+        return key in d
+
+    def replace(self, path, key, value):
+        # called by setattr
+        self.put(path, key, value)
+
+    @modifier
+    def put(self, path, key, value):
+        d = self._subdict(path)
+        is_new = key not in d
+        if not is_new and d[key] == value:
+            return
+        d[key] = value
+        self.db_add(path, key, value) if is_new else self.db_replace(path, key, value)
+
+    @modifier
+    def clear(self, path):
+        d = self._subdict(path)
+        d.clear()
+        path, key = path[:-1], path[-1]
+        self.db_replace(path, key, {})
+
+    def get(self, hint, key):
+        return hint[key]
+
+    def get_hint(self, path):
+        return self._subdict(path)
+
+    @modifier
+    def remove(self, path, key):
+        d = self._subdict(path)
+        d.pop(key)
+        self.db_remove(path, key)
+
+    @modifier
+    def list_append(self, path, item):
+        _list = self._subdict(path)
+        n = len(_list)
+        _list.append(item)
+        self.db_add(path, str(n), item)
+
+    def list_index(self, path, item):
+        _list = self._subdict(path)
+        return _list.index(item)
+
+    def list_len(self, path):
+        _list = self._subdict(path)
+        return len(_list)
+
+    @modifier
+    def list_clear(self, path):
+        _list = self._subdict(path)
+        _list.clear()
+        self.db_remove(path[:-1], path[-1])
+        self.db_add(path[:-1], path[-1], [])
+
+    @modifier
+    def list_remove(self, path, item):
+        _list = self._subdict(path)
+        n = _list.index(item)
+        _list.remove(item)
+        self.db_remove(path, str(n)) # fixme: keys
 
     def load_data(self, s: str) -> Dict[str, Any]:
         if s == '':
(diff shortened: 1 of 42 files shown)


electrum/stored_dict.py
```diff
@@ -43,11 +43,21 @@ class StorageEncryptionVersion(IntEnum):
...
 
 
-def locked(func):
-    def wrapper(self, *args, **kwargs):
-        with self.lock:
-            return func(self, *args, **kwargs)
-    return wrapper
+def normalize_key(x: _FLEX_KEY) -> str:
+    if isinstance(x, int):
+        return int(x)
```
f321x, line 48: Is this intended to cast x as `str` instead? Otherwise the annotated return type is incorrect.

```suggestion
        return str(x)
```

electrum/stored_dict.py
```diff
@@ -131,118 +153,243 @@ def set_parent(self, *, key: _FLEX_KEY, parent: Optional['BaseStoredObject']) ->
...
+    def __len__(self) -> int:
+        return self._db.dict_len(self.hint, self._path)
+
+    # ---- Dict-like extras ----
+
+    def __contains__(self, key: object) -> bool:
+        key = key_to_str(key)
+        return self._db.dict_contains(self.hint, self._path, key)
+
+    def keys(self) -> Iterable[str]:
```
f321x, line 342: The iterator returned by the `keys()/values()/items()` methods is exhausted after one iteration, unlike the [dictionary view objects](https://docs.python.org/3/glossary.html#term-dictionary-view) returned by a regular dict. 

This seems like a footgun as it behaves unexpected in situations like this:
```diff
diff --git a/tests/test_stored_dict.py b/tests/test_stored_dict.py
index 26e96fc63..cf8f92701 100644
--- a/tests/test_stored_dict.py
+++ b/tests/test_stored_dict.py
@@ -98,3 +98,14 @@ class TestStorage(ElectrumTestCase):
         storage = DictStorage(self.path)
         self.assertEqual(storage.dump(), {'a':{}})
 
+    async def test_show_iterator_type(self):
+        storage = DictStorage(self.path)
+        storage['a'] = {'k1': 1, 'k2': 2, 'k3': 3}
+        a = storage.get('a')
+        # len(a.keys())  # TypeError: object of type 'generator' has no len()
+        keys = a.keys()
+        b = list(keys)
+        c = list(keys)
+        self.assertEqual(len(b), len(c))
+        storage['b'] = {}
+        self.assertFalse(storage.get('b').keys())  # iterator is always True
```  

In fact, `_convert_version_14()` already surfaces this issue:
```python
                pubkeys = self.get('keystore').get('keypairs').keys()
                assert len(addresses) == len(list(pubkeys))  # drains pubkeys
                d = {}
                for pubkey in pubkeys:  # pubkeys is empty here
                    addr = bitcoin.pubkey_to_address('p2pkh', pubkey)
```

A simple and nice fix seems to be making `StoredDict` inherit from [`collections.abc.MutableMapping`](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes):

<details>
<summary> See patch (click to expand) </summary>

```diff

diff --git a/electrum/stored_dict.py b/electrum/stored_dict.py
index 1917ca478..41df28a2f 100644
--- a/electrum/stored_dict.py
+++ b/electrum/stored_dict.py
@@ -27,7 +27,8 @@ import threading
 import os
 from enum import IntEnum
 from collections import defaultdict
-from typing import Any, Optional, Tuple, Union, Iterator, Iterable, List, Sequence
+from collections.abc import MutableMapping
+from typing import Any, Optional, Union, Iterator, List, Sequence
 from .logging import Logger
 
 
@@ -272,7 +273,7 @@ class StoredObject(BaseStoredObject):
         return d
 
 
-class StoredDict(BaseStoredObject):
+class StoredDict(BaseStoredObject, MutableMapping):
     """
     dict-like object that queries the DB
     type conversions are performed here
@@ -327,8 +328,9 @@ class StoredDict(BaseStoredObject):
         key = key_to_str(key)
         self._db.remove(self.hint, self._path, key)
 
-    def __iter__(self) -> Iterator[str]:
-        return self._db.iter_keys(self.hint, self._path)
+    def __iter__(self) -> Iterator[_FLEX_KEY]:
+        for k in self._db.iter_keys(self.hint, self._path):
+            yield self._convert_key(k)
 
     def __len__(self) -> int:
         return self._db.dict_len(self.hint, self._path)
@@ -339,18 +341,6 @@ class StoredDict(BaseStoredObject):
         key = key_to_str(key)
         return self._db.dict_contains(self.hint, self._path, key)
 
-    def keys(self) -> Iterable[str]:
-        for k in self._db.iter_keys(self.hint, self._path):
-            yield self._convert_key(k)
-
-    def values(self) -> Iterator[Any]:
-        for k in self._db.iter_keys(self.hint, self._path):
-            yield self[k]
-
-    def items(self) -> Iterator[Tuple[str, Any]]:
-        for k in self._db.iter_keys(self.hint, self._path):
-            yield (self._convert_key(k), self[k])
-
     def get(self, key: _FLEX_KEY, default: Any = None, add_if_missing=False) -> Any:
         # If add_if_missing is True, create DB entry if it does not exist.
         # This will return StoredDict/StoredList if default is dict/list
@@ -378,15 +368,6 @@ class StoredDict(BaseStoredObject):
         del self[key]
         return v
 
-    def update(self, other=(), /, **kwargs) -> None:
-        if isinstance(other, dict):
-            pairs = list(other.items())
-        else:
-            pairs = list(other)
-        pairs.extend(kwargs.items())
-        for k, v in pairs:
-            self[k] = v
-
     def as_dict(self) -> dict:
         """used by keystore"""
         return self.dump()
```

</details>

And `StoredList` could similarly inherit from `collections.abc.MutableSequence`.

electrum/json_db.py
```diff
@@ -84,36 +82,141 @@ def wrapper(self, *args, **kwargs):
...
+
+    def init_db(self):
+        if self.storage.is_encrypted():
+            assert self.storage.is_past_initial_decryption()
+        json_str = self.storage.read()
+        self.json_data = self.load_data(json_str)
         # write file in case there was a db upgrade
-        if self.storage and self.storage.file_exists():
-            self.write_and_force_consolidation()
+        self.write_and_force_consolidation()
```
f321x, line 120: On master `self.write_and_force_consolidation()` was only called after the wallet upgrade, so if the upgrade failed the patches weren't applied. Now it is called before applying the upgrade.
This causes `test_mainnet_testnet_mixup` to apply the patches in `/tests/test_storage_upgrade/client_4_5_2_9dk_with_ln` on each test run which modifies the test wallet and shows up in the git diff.
Can the db write be delayed until the wallet upgrade succeeds?

tests/test_stored_dict.py
```diff
@@ -0,0 +1,100 @@
...
+        storage = DictStorage(self.path)
+        storage['a'] = {'b': {'c': 0}}
+        storage.write()
+        a = storage.get('a')
+        b = a['b']
+        self.assertEqual(type(b), StoredDict)
+        b2 = a.pop('b')
+        self.assertEqual(type(b2), dict)
+        # replace item. this must not been written to db
+        with self.assertRaises(KeyError):
```
f321x, line 94: This only raises if `self.hint` is not yet populated. When e.g. accessing (`__getitem__`) `StoredDict` first and the hint property sets `self.hint`, subsequent `__setitem__` don't raise. 
This fails e.g.:

```diff
diff --git a/tests/test_stored_dict.py b/tests/test_stored_dict.py
index 26e96fc63..0e25480ad 100644
--- a/tests/test_stored_dict.py
+++ b/tests/test_stored_dict.py
@@ -88,6 +88,7 @@ class TestStorage(ElectrumTestCase):
         a = storage.get('a')
         b = a['b']
         self.assertEqual(type(b), StoredDict)
+        print(b['c'])  # reads from b, populates cache
         b2 = a.pop('b')
         self.assertEqual(type(b2), dict)
         # replace item. this must not been written to db
```

electrum/wallet_db.py
```diff
@@ -1280,6 +1292,7 @@ def _move_unprocessed_onion(short_channel_id: str, htlc_id: Optional[int]) -> Op
                     # as the htlcs won't get failed due to the new SETTLING state
                     # unless a forwarding error is set.
                     recv_mpp_status[0] = 4  # RecvMPPResolution.SETTLING
+                    mpp_sets[payment_key] = recv_mpp_status
```
f321x, line 1295: Should this be moved one intendation layer out? 
Otherwise the new type mpp htlcs will only be saved if a forwarding_key is set.

electrum/wallet_db.py
f321x, line 591: This crashes in `to_default()` as `r['buckets']` is a `StoredList`.
```suggestion
            r = c['revocation_store'].dump()
```

Could `to_default()` handle `StoredDict/StoredList` by calling `dump()` on them so this would be prevented?

electrum/lnhtlc.py
f321x, line 238: `_maybe_active_htlc_ids` contains `Set[int]` but `StoredDict.__iter__()` returns `Iterator[str]` because it doesn't call `_convert_key()` on the yielded keys. 
So this might mixes up `int` and `str` htlc ids in the same set.
This is also fixed by the patch in comment https://github.com/spesmilo/electrum/pull/10693#discussion_r3614996503

f321x, line 28: Is this comment still up-to-date, shouldn't they should be converted to `int` on exposure as the keys are registered in `wallet_db.py`?

electrum/wallet.py
```diff
@@ -4524,8 +4538,6 @@ def restore_wallet_from_text(
         if gap_limit_for_change is not None:
             db.put('gap_limit_for_change', gap_limit_for_change)
         wallet = wallet_factory(db, config=config)
-    if db.storage:
-        assert not db.storage.file_exists(), "file was created too soon! plaintext keys might have been written to disk"
```
f321x, line 4528: The wallet file will already get created when instantiating `db = WalletDB(storage)` above due to the upgrader setting the initial seed version etc and `write_batch()` then writing the file. 
If wallet recovery then fails later on here the user ends up with an unusable wallet file.
Would it be safer to just not create a new file/call `self.write()` in `JsonDB.write_batch()` if the path doesn't exist yet and rely on the explicit call to `wallet.save_db()` below to create the file?

electrum/stored_dict.py
```diff
@@ -131,118 +153,243 @@ def set_parent(self, *, key: _FLEX_KEY, parent: Optional['BaseStoredObject']) ->
...
+
+    def should_convert(self):
+        return self._db._should_convert
+
+    def write_batch(self):
+        return self._db.write_batch()
+
+    def dump(self) -> dict:
+        data = {}
+        for k, v in self.items():
```
f321x, line 304: Should `StoredDict|StoredList.dump()` take `self.lock` to prevent `self` from changing during the iteration? Otherwise things like `Abstract_Wallet.save_backup()` might end up with a weird state if something from the asyncio thread modifies the dict/list while the GUI dumps it.
```suggestion
with self.lock:
        for k, v in self.items():
```

electrum/stored_dict.py
```diff
@@ -131,118 +153,243 @@ def set_parent(self, *, key: _FLEX_KEY, parent: Optional['BaseStoredObject']) ->
...
+        # This will return dict/list
+        try:
+            v = self[key]
+        except KeyError:
+            if default is _RaiseKeyError:
+                raise
+            return default
+        if isinstance(v, (StoredList, StoredDict)):
+            v = v.dump()
+        del self[key]
```
f321x, line 378: As this is not done under a lock i guess it could raise even if a `default` is given if multiple threads access/modify the dict simultaneously.

electrum/wallet.py
f321x, line 3040: Now that objects are re-instantiated on every access this probably won't work anymore as `._broadcasting_status` is not persisted and then lost on the next db access when a fresh `Invoice` is instantiated?

tests/test_jsondb.py
```diff
@@ -10,7 +11,7 @@
 
 from . import ElectrumTestCase
 
-from electrum.json_db import JsonDB
```
f321x, line 13: Removing the json_db import prevents the monkeypatch from redacting the secrets, causing the test to fail if `test_jsondb.py` is tested on its own (otherwise jsondb gets imported in `test_commands.py` and the monkeypatch applies).

```
====================================================================== short test summary info ======================================================================
SUBFAILED[replace_dict_inner_key_missing] tests/test_jsondb.py::TestJsonpatch::test_patch_does_not_leak_privatekeys - AssertionError: 'dictlevel' unexpectedly found in "can't replace a non-existent object 'dictlevelX2'"
...
```

electrum/stored_dict.py
```diff
@@ -78,49 +88,61 @@ def decorator(func):
...
+        return dict([(key_to_str(k), to_default(v)) for k, v in obj.items()])
+    if isinstance(obj, list):
+        return [to_default(x) for x in obj]
+    if isinstance(obj, tuple):
+        return tuple([to_default(x) for x in list(obj)])
+    raise Exception('unsupported type', type(obj))
+
+
+
+class BaseDB(Logger):
```
f321x, line 118: Could `BaseDB` be made a proper [abstract base class](https://docs.python.org/3/library/abc.html#module-abc) defining all the methods that it is supposed to implement (get/put/replace/remove/clear/iter_keys/dict_len/dict_contains/get_hint/list_append/etc...)?

This would add some boilerplate code but would also be nice to understand what `BaseDB` is actually supposed to support when implementing support for a new DB and would look better with a type checking IDE.

electrum/json_db.py
```diff
@@ -84,36 +82,141 @@ def wrapper(self, *args, **kwargs):
...
+        return key in d
+
+    def replace(self, d, path, key, value):
+        # called by setattr
+        self.put(d, path, key, value)
+
+    @modifier
+    def put(self, d, path, key, value):
+        is_new = key not in d
+        if not is_new and d[key] == value:
```
f321x, line 175: When storing a tuple in a `StoredDict` the `to_default()` call doesn't convert it to a json array but keeps it as tuple. 
When the value is still a `list` after loading the data from file this comparison 
will be False as `[1, 2] != (1, 2)`  and a useless replace patch is appended.
Could `to_default()` simply return a list for passed tuples?

f321x:
> should StoredDict.get() return dict/list instead of StoredList/StoredDict ?

I think this sounds good as `StoredDict.get()` already is a bit ambiguous with the `add_if_missing` parameter. However it might be worth to change the name to something else, e.g. `get_copy()`, to prevent confusion and make it explicit that the returned object is not persisted.

> should we remove the init_db parameter in the constructor, and use an explicit DictStorage.open() call instead?

It seems cleaner, I think it would make it easier to reason about what happens when and might allow for better error handling. E.g. when reading the file fails, the caller could still check `db.file_exists()` as there is a instantiated `JsonDB` object, if the constructor raises there is no object at all. Might also be better for testability.
Also right now there are 3 places where `_is_closed` gets set `False`, with an `open()` method that could be consolidated in a single place and the API would be simpler as `open()`/`decrypt()` could be a single thing.

electrum/stored_dict.py
```diff
@@ -251,28 +398,130 @@ def setdefault(self, key: _FLEX_KEY, default = None, /):
...
+        if type(s) is int:
+            s = n + s if s < 0 else s
+            return self._get_list_item(s)
+        elif type(s) is slice:
+            start = 0 if s.start is None else s.start if s.start >= 0 else n + s.start
+            stop = n if s.stop is None else s.stop if s.stop >= 0 else n + s.stop
+            step = 1 if s.step is None else s.step
+            return [self._get_list_item(i) for i in range(start, stop, step)]
+        else:
+            raise Exception()
```
f321x, line 432: This could be simplified by using the [builtin `.indices()` method](https://docs.python.org/3/reference/datamodel.html#slice.indices) of the `slices` type instead of manually calculating `start`, `stop` and `step`. Also the builtin list `__getitem__` raises `IndexError` for out of bound access.
```suggestion
def __getitem__(self, s: int | slice) -> Any:
    n = self._db.list_len(self.hint, self._path)
    if isinstance(s, int):
        if s < 0:
            s += n
        if not 0 <= s < n:
            raise IndexError(f'list index out of range: {s}')
        return self._get_list_item(s)
    elif isinstance(s, slice):
        return [self._get_list_item(i) for i in range(*s.indices(n))]
    raise TypeError(f'indices must be integers or slices, not {type(s).__name__}')
```

electrum/wallet_db.py
ecdsa: indeed, thanks

electrum/json_db.py
ecdsa: done.

tests/test_stored_dict.py
ecdsa: fixed

electrum/stored_dict.py
ecdsa: looks good. done.

electrum/lnhtlc.py
ecdsa: patch applied

electrum/json_db.py
ecdsa: Actually, this is wanted. The conversion of tuple to list is a limitation of the json backend. I do not want that limitation to be carried over to other backends, e.g. levelDB. So, to_default should return a tuple, because it should be storage agnostic.

ecdsa: I think this can be fixed in json_db.py, without changing the behavior of `to_default`

electrum/wallet_db.py
ecdsa: no, better call `dump()` here. to_default is expected to return `StoredDict/List` objects. I will add that to the docstring.

tests/test_jsondb.py
ecdsa: fixed. new tests were added in the meantime, that require importing JsonDB

electrum/json_db.py
ecdsa: fixed: added an extra converter, `to_json_data` to `json_db.py`
Note that list-to-tuple back-conversions are in `wallet_db.py`, search `register_name`

electrum/stored_dict.py
ecdsa: fixed

electrum/wallet_db.py
ecdsa: fixed

electrum/stored_dict.py
ecdsa: Right... actually my initial plan was to have all keys to be `str`, but then I backed off because it is not so simple. 
I think I will just fix the type hint, and postpone that decision.

electrum/stored_dict.py
ecdsa: this patch looks good, thanks! 
The same issue was flagged by Claude in security review

electrum/stored_dict.py
ecdsa: indeed. I added the lock there too.

electrum/lnhtlc.py
ecdsa: This is related to your `normalize_key` question. 
I believe the comment means that int can be used as keys, which is still up to date.

electrum/wallet.py
ecdsa: ok, I changed that: `WalletDB.__init__` no longer writes the file if the wallet is new. 

This allows to reintroduce the assert.

electrum/wallet.py
ecdsa: This is not so easy. Note that the `swap_manager` has exactly the same issue, and solves it by having a cache of `SwapData`. We could apply the same solution to invoices, but I do not like it... 

Note that I added a cache to `StoredDict` (4th commit). This 'solves' the issue, but this really is not **how** it should be solved; Indeed, unlike the `swaps_manager` cache, this cache is there for performance, and we should not rely on its existence. It should be possible to flush this cache without having weird side effects. 

So, this issue is temporarily 'fixed', but probably deserves a big FIXME.
I will defer this for now, because it is related to whether we use `dataclasses` instead of `attr` for invoices

electrum/stored_dict.py
ecdsa: sounds good. abstract methods were added.

PR #10694, opened 2026-06-08
refactor change-to-swap and submarine-payment parts from GUI to backend 

accumulator:
Move submarine swap support code from qt gui to backend and common_qt, where applicable.
    
This covers the functionality for swap use during payments (so, change-to-ln and
submarine-payments) present in `gui/qt/confirm_tx_dialog`, not the 'standalone' swap.

This is in preparation for adding the same functionality to qml.

- move backend code present in GUI to backend
- move common Qt GUI swap support code from `TxEditor` to `SubmarineSwapMixin`


ecdsa:
looks good.

PR #10695, opened 2026-06-08
add Iroh P2P transport for Electrum

rueckwaerts:
Adds Iroh QUIC-based P2P transport as an alternative to TCP/SSL/Tor.

## What this does
- Adds `electrum/iroh_transport.py` — asyncio Transport over Iroh QUIC streams
- Adds `i` as a valid protocol in `KNOWN_ELEC_PROTOCOL_TRANSPORTS`
- `ServerAddr` bypasses NetAddress validation for protocol `i`
- Iroh connection path in `open_session()` before `_RSClient`
- Persistent secret key in `~/.electrum/iroh_secret_key.bin`
- `remote_address()` stub for compatibility with `interface.py`

## How to connect
Server address: `<NODE_ID>:1:i`

## Dependencies
- `iroh` Python package (0.35.0): `pip install iroh`

## Tested with
- electrs with Iroh listener (rueckwaerts/electrs, branch iroh-test)
- StartOS 0.4.0-beta.9 (x86_64)
- End-to-end: wallet sync, transaction history, real-time block updates
- Handshake duration: ~165ms over relay, ~4ms direct LAN

## Notes
- Uses iroh 0.35.0 (latest version with Python bindings)
- ALPN: `electrs/electrum/0`
- Relay: `https://use1-1.relay.iroh.network./`
- No changes to existing TCP/SSL/Tor transports

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -467,14 +467,21 @@ class ServerAddr:
             protocol = 's'
         if not host:
             raise ValueError('host must not be empty')
+        if protocol not in KNOWN_ELEC_PROTOCOL_TRANSPORTS:
+            raise ValueError(f"invalid network protocol: {protocol}")
+        if protocol == 'i':
+            # Iroh: host is a Node ID (hex string), not a network address
+            self.host = host
+            self.port = int(port)
+            self.protocol = protocol
+            self._net_addr_str = f"{host}:{port}"
+            return
         if host[0] == '[' and host[-1] == ']':  # IPv6
             host = host[1:-1]
         try:
             net_addr = NetAddress(host, port)  # this validates host and port
         except Exception as e:
             raise ValueError(f"cannot construct ServerAddr: invalid host or port (host={host}, port={port})") from e
-        if protocol not in KNOWN_ELEC_PROTOCOL_TRANSPORTS:
-            raise ValueError(f"invalid network protocol: {protocol}")
         self.host = str(net_addr.host)  # canonical form (if e.g. IPv6 address)
         self.port = int(net_addr.port)
         self.protocol = protocol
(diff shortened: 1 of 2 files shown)


SomberNight:
why? what's the pitch?

-----

Also,

> Dependencies
>
>    `iroh` Python package (0.35.0): `pip install iroh`

No, that's [just the bindings](https://pypi.org/project/iroh). And they are [unmaintained already](https://github.com/n0-computer/iroh-ffi/blob/c96e4288ebeadb9c1a2122f431c8614de5b4780d/README.md). The [real logic](https://github.com/n0-computer/iroh) is in rust.
You really have to justify adding new dependencies, triply so if they are not pure python.

rueckwaerts:
Thanks for the quick review. The unmaintained bindings are a valid concern and understood.
To address the pitch more directly: this removes all connection complexity for self-hosters. Today, connecting Electrum to your own server requires knowing your IP, setting up Tor, or managing SSL certificates. With Iroh you share one string — the Node ID. It works behind NAT, no IP needed, no certificates.
On the dependency concern: Fedimint recently shipped Iroh in production (v0.7) for exactly the same reason — NAT traversal without infrastructure. They use the Rust library directly. A future iteration of this transport could use a subprocess or socket bridge to the Rust binary instead of Python bindings, eliminating the unmaintained dependency entirely.
I'll revisit with that approach or when Iroh 1.x Python bindings land.

rueckwaerts:
Updated based on feedback:

1. **No more pip dependency** — removed the `iroh` Python package entirely. 
   The transport now spawns `iroh-electrum-bridge` as a subprocess (a small 
   Rust binary). No pip install required.

2. **The pitch:**
   Users running their own Bitcoin node at home (Start9, Umbrel, RaspiBlitz) 
   currently need port forwarding, Tor, or a VPN to connect from outside their 
   home network. This is a real barrier for digital nomads and travelers.
   
   With Iroh P2P, the connection works from anywhere — no router config, 
   no Tor, no VPN. The user only enters a 64-char Node ID once.

Bridge binary source: https://github.com/rueckwaerts/iroh-electrum-bridge

PR #10709, opened 2026-06-25
qml: improve InvoiceDialog when launching from URI

f321x:
When launching the Electrum Android app from another app through an URI there were two UX issues with the `InvoiceDialog`:
1. The fiat exchange rate would not be shown as it is unavailable directly after start and didn't automatically update once it became available.
2. The dialog says `Insufficient balance` at the top while channels are still reestablishing, this is a bit confusing, it now tells the user that we are still connecting to the lightning peers instead.


accumulator:
LGTM, thanks

PR #10711, opened 2026-06-26
qt: prevent setting locktime on channel funding tx

f321x:
Prevents the user from manually modifying the locktime of a lightning channel funding transaction.
This seems like a potential footgun without much practical use.


electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -590,7 +591,7 @@ def set_fee_edit_visible(self):
             w.show()
 
     def set_locktime_visible(self):
-        b = self.config.GUI_QT_TX_EDITOR_SHOW_LOCKTIME
+        b = self.config.GUI_QT_TX_EDITOR_SHOW_LOCKTIME and self.context != TxEditorContext.CHANNEL_FUNDING
```
SomberNight, line 594: Oooh nice, I did not remember this context concept!

Issue #10688, opened 2026-06-05
QML "hidden wallet" feature


SomberNight:
It would be neat if there was a feature in the qml gui where a wallet could be marked as hidden: it would not show up in the list of wallets but it could still be opened somehow. For example, maybe a user could have a watch-only wallet for their cold storage, on their phone, as a hidden wallet.

---

There are several UI/UX questions that arise:
1. how to mark/unmark a wallet as hidden
2. how to open a hidden wallet
    - we could repurpose the wizard: user clicks "Create Wallet" and on the "New Wallet" screen, they enter a "Wallet name"
      - we already check if the name is available and only enable the "Next" button if it is
      - if the user sets the name to the name of an existing hidden wallet, we enable the "Next" button, and clicking it opens it
3. what to do if the last wallet used when the app is cold-started is a hidden wallet? (`config.CURRENT_WALLET`)
    - I guess we should not auto-open it
    - maybe prevent the situation from happening in the first place by not setting `config.CURRENT_WALLET` to a hidden wallet?

---

In the past I had considered question (1) to be the most problematic (especially the "unmark" from hidden part).
But now I just realised we have it already implemented!!! It is already there!
@f321x implemented renaming wallet files in https://github.com/spesmilo/electrum/pull/10573, and this interacts with the little-known feature where if your wallet file name starts with a `.`, it won't get listed, lol.
So if you prepend the dot, you mark it as hidden;  if you remove the dot, you unmark it.

So now only question (2) needs to be handled and implemented (and (3), but that's probably easy) :D

Alternatively, we should probably disallow prepending dots to wallet file names as users will not be able to open those wallets, but that would just be lame.

ZenulAbidin:
If a person knows the Electrum wallet directory (`~/.electrum/wallets`), then all the wallet files are visible there, and hiding won't really offer much.

I guess it can be hidden from the Recent Wallets list, but I'd think it would be better to just have a Clear Recent button in that case.

ecdsa:
@ZenulAbidin this issue is about QML, the Android GUI

SomberNight:
done in https://github.com/spesmilo/electrum/pull/10741

PR #10715, opened 2026-06-29
qml: disallow wallet names starting with '.'

f321x:
Prevent qml users from creating or renaming wallet files starting with a `.` as those files are not listed in the wallet list, making it impossible to access the wallet.
Later we could allow this again and implement opening such a "hidden" wallet file e.g. through the wizard.

Related https://github.com/spesmilo/electrum/issues/10688


SomberNight (approved):
Looks good.

later, we should add the "hidden wallet" feature :)

PR #10720, opened 2026-06-30
exchange rate: remove CoinDesk provider

f321x:
The API is non-functional and their new API requires an authorization token, so we cannot keep using it.


SomberNight:
https://data.coindesk.com/blogs/changes-to-coindesk-data-indices-api-free-tier-access

> As of May 21, 2026, we will be retiring our free API tier and accounts without a subscription will no longer have API access.

----

there are not many providers for historical data  :(

PR #10723, opened 2026-07-01
QML send change to lightning

accumulator:
Add change-to-lightning option in `ConfirmTxDialog`. 

This builds on the prep work done in #10694

Options available to the user in `ConfirmTxDialog` is now a `flag` on the finalizer, instead of a `showOptions` boolean property on the QML side. Sending change to lightning is disabled for sweeps and channel-opens.

### DIFF
--- a/electrum/gui/qml/components/ConfirmTxDialog.qml
+++ b/electrum/gui/qml/components/ConfirmTxDialog.qml
@@ -205,6 +205,27 @@ ElDialog {
                             helptext: Config.longDescFor('WALLET_COIN_CHOOSER_OUTPUT_ROUNDING')
                         }
 
+                        ElCheckBox {
+                            Layout.fillWidth: true
+                            visible: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > 0
+                            text: Config.shortDescFor('WALLET_SEND_CHANGE_TO_LIGHTNING')
+                            onCheckedChanged: {
+                                if (activeFocus) {
+                                    Config.sendChangeToLightning = checked
+                                    finalizer.doUpdate()
+                                }
+                            }
+                            Component.onCompleted: {
+                                checked = Config.sendChangeToLightning
+                            }
+                        }
+
+                        HelpButton {
+                            visible: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > 0
+                            heading: Config.shortDescFor('WALLET_SEND_CHANGE_TO_LIGHTNING')
+                            helptext: Config.longDescFor('WALLET_SEND_CHANGE_TO_LIGHTNING')
+                        }
+
                     }
                 }
 
--- a/electrum/gui/qml/components/ConfirmTxDialog.qml
+++ b/electrum/gui/qml/components/ConfirmTxDialog.qml
@@ -18,6 +18,8 @@ ElDialog {
     property alias amountLabelText: amountLabel.text
     property alias sendButtonText: sendButton.text
 
+    signal confirmed
+
     title: qsTr('Transaction Fee')
     iconSource: Qt.resolvedUrl('../../icons/question.png')
 
@@ -284,7 +286,7 @@ ElDialog {
                         : qsTr('Pay...')
                 icon.source: '../../icons/confirmed.png'
                 enabled: finalizer.valid
-                onClicked: doAccept()
+                onClicked: confirmed()
             }
         }
     }
--- a/electrum/gui/qml/components/ConfirmTxDialog.qml
+++ b/electrum/gui/qml/components/ConfirmTxDialog.qml
@@ -165,6 +165,7 @@ ElDialog {
                         id: optionslayout
                         width: parent.width
                         columns: 2
+                        rowSpacing: 0
 
                         ElCheckBox {
                             Layout.fillWidth: true
@@ -206,6 +207,7 @@ ElDialog {
                         }
 
                         ElCheckBox {
+                            id: cb_send_change_to_lightning
                             Layout.fillWidth: true
                             visible: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > 0
                             text: Config.shortDescFor('WALLET_SEND_CHANGE_TO_LIGHTNING')
@@ -224,6 +226,16 @@ ElDialog {
                             visible: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > 0
                             heading: Config.shortDescFor('WALLET_SEND_CHANGE_TO_LIGHTNING')
                             helptext: Config.longDescFor('WALLET_SEND_CHANGE_TO_LIGHTNING')
+                        }
+
+                        Label {
+                            visible: cb_send_change_to_lightning.visible && cb_send_change_to_lightning.checked
+                            color: constants.mutedForeground
+                            font.pixelSize: constants.fontSizeSmall
+                            text: finalizer.swapStatusMsg
+                            Layout.topMargin: -constants.paddingSmall
+                            Layout.leftMargin: cb_send_change_to_lightning.contentItem.leftPadding
+                                + cb_send_change_to_lightning.padding
                         }
 
                     }
--- a/electrum/gui/qml/components/ConfirmTxDialog.qml
+++ b/electrum/gui/qml/components/ConfirmTxDialog.qml
@@ -324,7 +324,7 @@ ElDialog {
                         ? qsTr('Finalize...')
                         : qsTr('Pay...')
                 icon.source: '../../icons/confirmed.png'
-                enabled: finalizer.valid
+                enabled: finalizer.valid && !finalizer.busy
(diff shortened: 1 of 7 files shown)


electrum/gui/qml/components/ConfirmTxDialog.qml
```diff
@@ -202,14 +206,16 @@ ElDialog {
...
                             heading: Config.shortDescFor('WALLET_COIN_CHOOSER_OUTPUT_ROUNDING')
                             helptext: Config.longDescFor('WALLET_COIN_CHOOSER_OUTPUT_ROUNDING')
                         }
 
                         ElCheckBox {
                             id: cb_send_change_to_lightning
                             Layout.fillWidth: true
-                            visible: Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > 0
+                            visible: finalizer.txOptions & TxFinalizer.TxOptions.SEND_CHANGE_TO_LIGHTNING
+                                && Daemon.currentWallet.isLightning && Daemon.currentWallet.lightningCanReceive.satsInt > 0
```
f321x, line 218: Maybe using `> submarine_swaps.MIN_SWAP_AMOUNT_SAT` would be more reliable UX than `> 0` as it is still not possible to receive change on lightning if the inbound liquidity is lower than the minimum swap amount.

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -561,11 +613,97 @@ def signAndSend(self):
...
+        assert tx.get_dummy_output(DummyAddress.SWAP)
+
+        async def handle_swap_task():
+            try:
+                swap_dummy_output = tx.get_dummy_output(DummyAddress.SWAP)
+                swap_manager = self.swap_manager
+                transport = self.swap_transport
+
+                try:
+                    if not swap_manager.is_initialized.is_set():
```
f321x, line 640: If the user hasn't configured a swap provider already previously by doing a manual swap this will always time out and show "Will not send change to lightning" without a reason.

Would it be possible to let the `NostrSwapServerDialog` pop-up similar to the manual swap view if no swapserver was configured or the configured swapserver isn't active anymore?

Otherwise the user might never figure out how to actually find and select a swapserver.

electrum/gui/qml/components/WalletMainView.qml
```diff
@@ -703,12 +710,43 @@ Item {
...
+                        title: qsTr('Error'),
+                        text: [qsTr('Could not swap change'), message].join('\n\n'),
+                        iconSource: '../../../icons/warning.png'
+                    })
+                    dialog.open()
+                }
+                onSwapStart: {
+                    _swapwaitdialog = app.messageDialog.createObject(mainView, {
+                        title: qsTr('Please wait...'),
+                        text: [qsTr('waiting for lightning invoice'), ''].join('\n\n'),
```
f321x, line 726:
```suggestion
                        text: [qsTr('Requesting Submarine Swap for change...'), ''].join('\n\n'),
```

Might be easier to understand? It might not be clear to the user why they have to wait for a lightning invoice if they aren't aware of the swap technicalities.

Image

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -561,11 +657,104 @@ def signAndSend(self):
...
+                except Exception as e:
+                    self._logger.info(f'swap error while requesting swap from server: {str(e)}')
+                    self.swapError.emit(str(e))
+                    return
+
+                try:
+                    if not self._wallet.wallet.sign_transaction(tx, self._wallet.password):
+                        raise Exception('tx not signed')
+                    self.swapStart.emit()
+                    funding_txid = await swap_manager.wait_for_htlcs_and_broadcast(
```
f321x, line 718: It's a bit weird that after the swap is funded it goes back to the `InvoiceDialog`, allowing the user to modify the `AmountEdit` and showing a incorrect `This invoice was already paid`, instead of showing the regular `Payment Sent` message popup. Maybe this is because the broadcast doesn't emit `broadcastSucceeded`?

Image

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -561,11 +657,104 @@ def signAndSend(self):
...
+        async def handle_swap_task():
+            try:
+                swap_dummy_output = tx.get_dummy_output(DummyAddress.SWAP)
+                swap_manager = self.swap_manager
+                transport = self.swap_transport
+
+                try:
+                    if not swap_manager.is_initialized.is_set():
+                        await asyncio.wait_for(swap_manager.is_initialized.wait(), timeout=5)
+                except Exception as e:
```
f321x, line 692: Should the `Exception` branches in the swap task call `self.update()` after the tx has been mutated with the dummy address, so that if the user clicks "Pay" again after a failure he doesn't accidentally fund the failed swaps dummy address?

electrum/gui/qml/components/WalletMainView.qml
```diff
@@ -703,12 +710,43 @@ Item {
...
+                        iconSource: Qt.resolvedUrl('../../icons/info.png'),
+                        buttonText: qsTr('Cancel'),
+                        buttonIcon: Qt.resolvedUrl('../../icons/closebutton.png')
+                    })
+                    _swapwaitdialog.accepted.connect(function() {
+                        cancelSwap()
+                    })
```
f321x, line 733: Should clicking outside the dialog/back maybe also cancel the swap?
```qml
_swapwaitdialog.rejected.connect(function() {
    cancelSwap()
})
```
Alternatively we could also disallow closing the dialog by clicking outside so the user must either wait or click cancel.

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -402,7 +439,9 @@ def __init__(
         make_tx: Callable[[int, FeePolicy], PartialTransaction] = None,
         accept: Callable[[PartialTransaction], None] = None,
     ):
-        super().__init__(parent)
+        TxFeeSlider.__init__(self, parent)
+        SubmarineSwapMixin.__init__(self, None)
```
f321x, line 443: `self.register_callbacks()` is not called in the `QETxFinalizer.__init__()`, so the `SubmarineSwapMixin` callbacks will never emit `swapAvailabilityChanged`

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -490,10 +576,20 @@ def update(self):
             self._logger.debug('wallet not set, ignoring update()')
             return
 
+        send_change_to_lightning = self._send_change_to_lightning()
+        if send_change_to_lightning:
```
f321x, line 580: If the relay connection fails `swapAvailabilityChanged` calls `update()` which then tries to prepare the swap transport again, so this is a infinite loop. On desktop we try to connect only once, maybe we should temporarily disable change to lightning if it isn't possible to connect or discover swap providers?

electrum/gui/qml/qetxfinalizer.py
```diff
@@ -411,6 +450,24 @@ def __init__(
...
+        self._busy = False
+        self.swap_task: Future = None
+
+        self.swapAvailabilityChanged.connect(self.on_swap_availability_changed, Qt.ConnectionType.QueuedConnection)
+
+        self.destroyed.connect(lambda: self.on_destroy())
+
+    def on_destroy(self):
+        self._logger.debug('on_destroy')
+        self.swap_transport_cleanup()
```
f321x, line 463: This raises trying to emit `swapAvailabilityChanged` after the `QETxFinalizer` has been destroyed:
```
311.03 | I | submarine_swaps.NostrTransport | connected relays: ['wss://relay.getalby.com/v1', 'wss://relay.damus.io', 'wss://nostr.einundzwanzig.space', 'wss://nostr.mom', 'wss://nos.lol', 'wss://relay.primal.net', 'wss://ftp.halifax.rwth-aachen.de/nostr']
311.07 | D | gui.qml.qetxfinalizer | on_destroy
311.08 | E | concurrent.futures | exception calling callback for <Future at 0x7f98a020b1e0 state=cancelled>
  File "/usr/lib64/python3.14/concurrent/futures/_base.py", line 335, in _invoke_callbacks
    callback(self)
    ~~~~~~~~^^^^^^
  File "/var/home/user/code/code_vm/electrum-2/electrum/gui/common_qt/swaps.py", line 85, in transport_initialize_done
    self.swapAvailabilityChanged.emit()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: wrapped C/C++ object of type QETxFinalizer has been deleted
311.08 | I | submarine_swaps.NostrTransport | shutting down nostr transport
311.14 | I | submarine_swaps.NostrTransport | taskgroup stopped.
311.15 | I | submarine_swaps.NostrTransport | nostr transport shut down
```

electrum/gui/qml/components/WalletMainView.qml
f321x, line 161:
```suggestion
                        finalizerDialog.close()
```

Pressing "Sweep" in a watch-only wallet doesn't close the sweep view anymore. I guess `close()` needs to be replaced with `finalizerDialog.close()`.

PR #10728, opened 2026-07-01
update block header checkpoints

SomberNight:
```
>>> network.export_checkpoints(f"/home/user/wspace/electrum/electrum/chains/{electrum.constants.net.NET_NAME}/checkpoints.json")
```


electrum/chains/mainnet/checkpoints.json
```diff
@@ -1854,5 +1854,41 @@
     [
         "00000000000000000001209d8bd5b083d0eec0a0cfb72a9f922a5f46e3b4744f",
         214194756963942469886095641713233006794734161633476608
+    ],
+    [
+        "00000000000000000001d7ccda9078724e0514418daa90f0a8bca6eb407b84c4",
+        186702449664110855091766126325007006911986414567555072
+    ],
```
SomberNight, line 1861: I wonder if Claude comes up with some "attack" that requires me faking the checkpoints

SomberNight: Ah well it did not.
Could have mentioned that if the values are fake, the software (e.g. release binaries) becomes unusable. Maybe next model.

Issue #9519, opened 2025-02-03
Decrease `funding_sat`


matevz:
The current `funding_sat` minimum amount for opening a channel is `200,000` sats which roughly equals to $200 these days. This amount is relatively high for the newcomers or even those running LN node themselves and want to play around with it. I suggest to reduce this amount to something more reasonable like `10,000` sats (CLN default) or `20,000` sats (LND default) or even better - remove this check locally and let the target LN node decide.

PR #10730, opened 2026-07-01
lnutil: make minimum channel funding amount configurable

nothing-stops-this-train:
Rationale:

Electrum comes with a hardcoded constant (MIN_FUNDING_SAT) which specifies the smallest allowable lightning channel, set to 200,000 sats, roughly equivalent to $200 USD. This is a burdensome amount for new Bitcoin users and puts lightning out of reach for them. This is especially relevant to lightning since it is really designed for small payments in the first place.

Some minimum, however, is needed for funds safety since if a channel partner published an old channel state (in an attempt to steal), it must be economical to challenge that state on-chain.

Other popular lightning wallets like Zeus set this at 20,000 sats (the LND default). Given that LND is the most popular lightning node client, this is a good reference amount.

This change enables users to set their own risk tolerance and advises them about how to do so, while sticking with a conservative 50,000 sat default.

Solution:
The minimum channel funding amount was a hardcoded constant (MIN_FUNDING_SAT = 200_000). Lower it to a 50_000 sat default and expose it as a new config setting, LIGHTNING_MIN_FUNDING_SAT, mirroring the existing LIGHTNING_MAX_FUNDING_SAT.

All call sites that previously read the hardcoded constant now read the value from config, so it is loaded at startup and honored at runtime like any other config value. The Qt Preferences dialog (Tools -> Preferences -> Lightning) gains an editor for the setting, together with a warning describing the risks of small channels in a high onchain-fee environment.

MIN_FUNDING_SAT is kept in lnutil as the default value for the new config var.

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -14,7 +14,7 @@ from .util import base_units, base_unit_name_to_decimal_point, decimal_point_to_
 from .util import format_satoshis, format_fee_satoshis, os_chmod
 from .util import user_dir, make_dir
 from .util import is_valid_websocket_url
-from .lnutil import LN_MAX_FUNDING_SAT_LEGACY
+from .lnutil import LN_MAX_FUNDING_SAT_LEGACY, MIN_FUNDING_SAT
 from .i18n import _
 from .logging import get_logger, Logger
 
@@ -751,6 +751,16 @@ Note that static backups only allow you to request a force-close with the remote
 If this is enabled, other nodes cannot open a channel to you. Channel recovery data is encrypted, so that only your wallet can decrypt it. However, blockchain analysis will be able to tell that the transaction was probably created by Electrum."""),
     )
     LIGHTNING_TO_SELF_DELAY_CSV = ConfigVar('lightning_to_self_delay', default=7 * 144, type_=int)
+    LIGHTNING_MIN_FUNDING_SAT = ConfigVar(
+        'lightning_min_funding_sat', default=MIN_FUNDING_SAT, type_=int,
+        short_desc=lambda: _("Minimum channel funding amount (sat)"),
+        long_desc=lambda: _(
+            "⚠️ Warning: Small channel sizes (<50 000 sats) come with some risks. "
+            "If your channel partner tries to cheat you by publishing an old channel state and "
+            "does so in a high onchain fee environment, on-chain fees could make challenging that "
+            "uneconomical (you could lose money). Don't open small channels with channel partners "
+            "you don't trust."),
+    )
     LIGHTNING_MAX_FUNDING_SAT = ConfigVar('lightning_max_funding_sat', default=LN_MAX_FUNDING_SAT_LEGACY, type_=int)
     LIGHTNING_MAX_HTLC_VALUE_IN_FLIGHT_MSAT = ConfigVar('lightning_max_htlc_value_in_flight_msat', default=None, type_=int)
     INITIAL_TRAMPOLINE_FEE_LEVEL = ConfigVar('initial_trampoline_fee_level', default=1, type_=int)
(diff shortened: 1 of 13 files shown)


electrum/simple_config.py
```diff
@@ -751,6 +751,16 @@ def __setattr__(self, name, value):
...
     LIGHTNING_TO_SELF_DELAY_CSV = ConfigVar('lightning_to_self_delay', default=7 * 144, type_=int)
+    LIGHTNING_MIN_FUNDING_SAT = ConfigVar(
+        'lightning_min_funding_sat', default=MIN_FUNDING_SAT, type_=int,
+        short_desc=lambda: _("Minimum channel funding amount (sat)"),
+        long_desc=lambda: _(
+            "⚠️ Warning: Small channel sizes (<50 000 sats) come with some risks. "
+            "If your channel partner tries to cheat you by publishing an old channel state and "
+            "does so in a high onchain fee environment, on-chain fees could make challenging that "
+            "uneconomical (you could lose money). Don't open small channels with channel partners "
+            "you don't trust."),
```
SomberNight, line 762: The whole model breaks down if the capacity is comparable to (or less than) the onchain fees. Are you sure you understand? I would probably have to spend 10+ minutes to think all the implications through so I highly doubt the typical user has any idea.

For example, did you realise that adding an HTLC to the channel usually increases the size of the commitment transaction and the onchain fees needed? Meaning if the prevalent mempool feerates are high, [you might not be able to add an HTLC at all](https://github.com/lightning/bolts/blob/94eb038c42e664dd7862faeec6508ccd25f63ff8/02-peer-protocol.md?plain=1#L2802-L2821). Notice that I have not mentioned trust or cheating here.

Having a large minimum is annoying a single well-defined time: when the user wants to open a channel. Having a small minimum is a lingering risk. It is also a constant looming danger for us that suddenly when the mempool fees spike we have to explain to all those users why their channels are unusable. It is bad UX at unexpected times.

Also note that ACINQ (and eclair in general by default) sets `max_htlc_value_in_flight_msat` to 45% of the capacity, meaning pending htlcs in either direction (separately-per-direction) cannot exceed 45%. So if you open a channel with ACINQ for 200k sats, actually the max invoice you are able to pay is 90k sats. Unless you are a Phoenix client because then you get special treatment. This is even more annoying if you open a small channel.

SomberNight:
ref https://github.com/spesmilo/electrum/issues/9519

nothing-stops-this-train:
You are right to point out the max_htlc_value_in_flight issue, this is a concern in high-fee environments.  There is obviously no "correct" answer, it comes down to the user and their risk tolerance. LND is the dominant lightning node software, not just for routing nodes but also for user wallets, and because of that, lightning users are used to regularly being able to move a much greater % of their channel balance than Eclair's defaults allow. Is that a good argument for allowing unsafe channel behaviour? Not really. As a dev you have to choose the right balance. Just because another wallet packages a convenient option, you shouldn't do the same thing if it's unsafe.

Your point about "one pain point now, avoid multiple pain points later" is sound. Do we apply this same logic to small UTXOs? A user can receive small on-chain payments only to find out they are functionally unspendable in a high fee environment, and of course, they would blame Electrum devs somehow for this even though it's a basic fact about how the Bitcoin network works. Just like "small lightning channels become useless in high-fee environments" is. A high-fee environment, in the typical lifecyle of a wallet, might also be considered either a one time or very rare occurrence.

Keep in mind that Electrum is the only mature desktop user wallet on Linux that supports lightning (and one of the few for other platforms). LND and other clients, of course, could also be run, but depend on complex plumbing like docker compose files. If users cant afford the $200 entry price and aren't a whiz in the terminal, they are essentially locked out of being able to use lightning at all. Again, not a great *technical* argument, but worth considering.  I think a 200,000 sat limit is letting the perfect be the enemy of the good.

A $200 entry fee to use a system to cover what is essentially an edge case is steep. Even if the default minimum isn't lowered, I think letting the user have control of their own risk tolerance and adding a setting for this is valuable and it puts Electrum on par with other wallets. It also fits with the Bitcoin ethos about returning autonomy to the user.

PR #10732, opened 2026-07-02
Coincontrol to backend

accumulator:
Refactor coin control from GUI to backend, in preparation of adding coin control features to QML

Edit: Opus 5 found a number of issues (1619a75bc3c16a81678012b730cbfd4f2be85d23) Opus 4.8 didn't find. I've tried to resolve these but this ripples out to a bit more code than I'd like. Now analyzing the coincontrol state of `master`, and it finds basically the same issues pre-existing. 

Short summary:
- `main_window.py:1419` docstring declares coincontrol a tri-state, and wallet layer honours it (`get_spendable_coins`), but GUI's `utxo_list.get_spend_list` collapses empty set to `None`, so it might inadvertendly select coins outside coincontrol if the last coin is removed from coincontrol.
- interplay with frozen state is flaky; `add_to_coincontrol` frozen-filters only at insertion time. `update_coincontrol_bar` frozen-filters, but `get_spend_list` doesn't, so the status bar can understate the number and amount. Frozen coins can be spent (add to CC, freeze, send)
- coincontrol bypasses every non-frozen filter; `get_coins(**kwargs)` discards all kwargs when a spend list exists, so no `mature_only`, `nonlocal_only`, `confirmed_only` etc. Think of funding a channel from a local, never broadcasted tx, or `get_candidates_for_batching` wanting a conservative coin selection but getting the unfiltered coin set.

I'm tempted to limit this PR to just the refactor to the backend, without regressing, but also without fixing all of the above issues, and incrementally fix the issues in other PRs

### DIFF
--- a/electrum/gui/qt/confirm_tx_dialog.py
+++ b/electrum/gui/qt/confirm_tx_dialog.py
@@ -193,6 +193,7 @@ class TxEditor(WindowModalDialog, SubmarineSwapMixin, Logger):
             self.needs_update = False
 
     def update(self):
+        self.error = ''
         self.update_tx()
         self.set_locktime()
         self._update_widgets()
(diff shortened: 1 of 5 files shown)


accumulator:
superseded by #10918

PR #10727, opened 2026-07-01
build: update some dependencies, all over

SomberNight:
I tried to be somewhat conservative.

(But really should do this mid-cycle, not just before a release. Then it would be possible to do major version bumps as well.)

-----

done manual QA, like in the stone age:
- built appimage OK
    - at runtime, used wizard, created wallet for Trezor1, connected to server, used "show address" on Trezor1
- built windows OK
    - at runtime, used wizard, created wallet for Trezor1, connected to server, used "show address" on Trezor1 
- built macos OK
    - at runtime, used wizard, created standard wallet, opened wallet, connected to server
- built android OK
    - at runtime
        - used wizard, created 2fa wallet, connected to server
        - opened wallet with lightning channels

PR #10733, opened 2026-07-02
util.make_aiohttp_session: wrap aiohttp-socks 0.11+ excs to ClientError

SomberNight:
older versions of aiohttp-socks and python-socks used to raise
- ProxyConnectionError(OSError)
- ProxyTimeoutError(TimeoutError)
- ProxyError(Exception)

now they raise:
- ProxyConnectionError(Exception)
- ProxyTimeoutError(Exception)
- ProxyError(Exception)

In many call sites, we currently handle OSError and TimeoutError, usually by simply logging the error or showing it to the user. Another exceptions our call sites handle similarly is aiohttp.ClientError, which is the aiohttp base class for any client connection error.

A simple "fix" for us to restore the old behaviour is converting the new aiohttp_socks exception types to aiohttp.ClientError.

ref https://github.com/romis2012/aiohttp-socks/commit/db4235c2b9805f088e47074647c946873a231487
ref https://github.com/romis2012/python-socks/commit/50a3024a8ac10888085f963e44ab3a79d6574703

-----

related:
https://github.com/spesmilo/electrum/pull/10727#issuecomment-4863659725 (LLM output:)
> Worth passing upstream (correctness, not security): aiohttp-socks 0.11.0's ProxyConnectionError is no o Electrum call sites like electrum/exchange_rate.py:90 that catch (aiohttp.ClientError, TimeoutError,OSError) now fall through to their generic except Exception when a SOCKS/Tor proxy is unreachable — functional but noisier; catching the aiohttp_socks exceptions explicitly would restore clean handling.

### DIFF
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -55,6 +55,7 @@ import functools
 from functools import partial
 from abc import abstractmethod, ABC
 import enum
+import contextlib
 from contextlib import nullcontext, suppress
 import traceback
 import inspect
@@ -62,6 +63,7 @@ import weakref
 
 import aiohttp
 from aiohttp_socks import ProxyConnector, ProxyType
+from aiohttp_socks import ProxyConnectionError, ProxyTimeoutError, ProxyError
 import aiorpcx
 import certifi
 import dns.asyncresolver
@@ -1354,7 +1356,7 @@ def make_aiohttp_proxy_connector(proxy: 'ProxySettings', ssl_context: Optional[s
     )
 
 
-def make_aiohttp_session(proxy: Optional['ProxySettings'], headers=None, timeout=None):
+def _make_aiohttp_session(proxy: Optional['ProxySettings'], headers=None, timeout=None):
     if headers is None:
         headers = {'User-Agent': 'Electrum'}
     if timeout is None:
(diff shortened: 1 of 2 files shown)


f321x:
Tested with exchange rate fetching logic. Works and seems reasonable:
```
 28.11 | I | exchange_rate.BitFinex | failed fx quotes: ClientError('proxy error: ProxyConnectionError("[Errno 111] Couldn\'t connect to proxy 127.0.0.1:9053 [Connect call failed (\'127.0.0.1\', 9053)]")')
```

electrum/util.py
```diff
@@ -55,13 +55,20 @@
...
 import weakref
 
 import aiohttp
 from aiohttp_socks import ProxyConnector, ProxyType
+try:
+    from aiohttp_socks import ProxyConnectionError, ProxyTimeoutError, ProxyError
+except ImportError:  # above requires aiohttp_socks>=0.11
+    class _FakeProxyError(Exception):
+        pass
+    ProxyConnectionError = ProxyTimeoutError = ProxyError = _FakeProxyError
```
SomberNight: note: actually older aiohttp_socks just imports and re-exports the same names from python_socks:
https://github.com/romis2012/aiohttp-socks/blob/v0.9.2/aiohttp_socks/__init__.py#L4-L8

PR #10736, opened 2026-07-03
wallet: avoid redundant prevout rescans in invoice paid detection

sashazykov:
Follow-up to #10658 (paid-invoice keys cache). Four small commits addressing the remaining prevout-rescan hotspots in invoice paid detection, all measured on a real wallet with 686 outgoing invoices, accumulated from a year of daily payments to 40-90 recipients — many of whom appear on the list every day, so their invoices pile up on the same addresses: 572 invoices share the single most-reused destination scriptpubkey.

| scenario | master | this PR |
|---|---|---|
| wallet open (incl. detection pass) | ~37 s | ~0.7 s |
| one tx event, invoices confirmed-paid (normal use) | ~35 s | <1 ms |
| one tx event, invoices unconfirmed (initial sync) | ~35 s | ~0.6 s |
| full sync from master key with invoices present | 12+ hours | ~5 min |

The ~35s cases run synchronously on the asyncio event loop thread — longer than the 30s request timeout, so during initial sync every in-flight request (merkle proofs, txs) was guaranteed to time out, servers kept disconnecting, and the GUI froze. Both threads were observed inside `_is_onchain_invoice_paid` via py-spy: the event loop under `on_event_adb_added_tx`, and the GUI thread under `InvoiceList.refresh_item`.

**Commit 1** — since 63ee2f3a6, the detection pass discards the cache key, scans prevouts (for `relevant_txs`), then calls `get_invoice_status()`, which rescans. Re-add the key from the fresh scan result (confirmed-paid only; LN status precedence preserved; reorg demotion unaffected). Halves the wallet-open pass. Side effect: on master the cache was actually empty right after `__init__` (`register_callbacks()` runs after `_prepare_onchain_invoice_paid_detection()`), so this also makes load-time cache population effective.

**Commit 2** — a tx addition can only promote an invoice towards paid; demotion only comes from the removal paths (reorg/conflict), which still rescan unconditionally. So additions skip invoices already cached as `PR_PAID`, like `set_broadcasting` does since d2d4251c8. Fixes the ~35s per tx in a synced wallet. Does not help unconfirmed invoices (`conf==0` cannot be cached) — that is addressed by the next commit.

**Commit 3** — invoices sharing a scriptpubkey each redid the identical prevouts + `get_tx_height` lookups; one pass over 572 sharing invoices did 572+ lookups of the same data. Memoize the per-scriptpubkey lookup for the duration of one detection pass; per-invoice height filtering applies on top. No invalidation is needed: the cache exists only while a pass runs, adb state cannot change mid-pass, and all access is serialized under `wallet.lock`. Note the diff is dominated by a `try/finally` re-indent — `git show -w` shows the ~37-line logic change.

**Commit 4** — each `invoice_status` event made the Qt invoice list recompute the status (a prevout scan) on the GUI thread; while syncing these arrive per touched invoice per tx. While not up-to-date, just set `need_update` — `update_wallet()` already defers `update_tabs()` until `is_up_to_date()` and then rebuilds the list once.

**Testing:** three new unit tests spying on the scan/lookup calls (load pass scans each invoice once; tx additions and verifications don't rescan cached-paid invoices; a shared scriptpubkey is looked up once per pass, with per-invoice height filtering pinned by a height-excluded invoice). The existing reorg test still pins paid->unpaid demotion. `tests.test_invoices`, `tests.test_wallet`, `tests.test_wallet_vertical`: 131 tests pass. Real-world: watch-only wallet restored from the hw wallet's master public key with the invoices imported — full sync from scratch took 12+ hours on master (timeout spam, disconnections, frozen GUI) vs ~5 minutes on this branch.

Known remaining (pre-existing, not introduced here): `timer_actions` refreshes the visible invoice list every ~0.5s with a per-row status recomputation, so keeping the Send tab open during initial sync still costs GUI-thread scans; the `request_status` and QML paths have the same per-event pattern. Happy to follow up on those if there's interest.

Happy to privately share the GPG-encrypted watch-only reproduction wallet if useful for testing.

### DIFF
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -1387,6 +1387,12 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger, QtEventListener):
     def on_event_invoice_status(self, wallet, key, status):
         if wallet != self.wallet:
             return
+        if not wallet.is_up_to_date():
+            # while syncing, these events arrive in bulk (per touched invoice per tx), and
+            # refresh_item recomputes the status, scanning prevouts on the GUI thread.
+            # update_wallet() rebuilds the list in one batch once we are up_to_date again.
+            self.need_update.set()
+            return
         if status == PR_PAID:
             self.send_tab.invoice_list.delete_item(key)
         else:
--- a/electrum/gui/qt/main_window.py
+++ b/electrum/gui/qt/main_window.py
@@ -1388,6 +1388,12 @@ class ElectrumWindow(QMainWindow, MessageBoxMixin, Logger, QtEventListener):
     def on_event_invoice_status(self, wallet, key, status):
         if wallet != self.wallet:
             return
+        if not wallet.is_up_to_date():
+            # while syncing, these events arrive in bulk (per touched invoice per tx), and
+            # refresh_item recomputes the status, scanning prevouts on the GUI thread.
+            # update_wallet() rebuilds the list in one batch once we are up_to_date again.
+            self.need_update.set()
+            return
         if status == PR_PAID:
             self.send_tab.invoice_list.delete_item(key)
         else:
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -631,7 +631,7 @@ class Abstract_Wallet(ABC, Logger, EventListener):
         self.clear_tx_parents_cache()
         if self.lnworker:
             self.lnworker.maybe_add_backup_from_tx(tx)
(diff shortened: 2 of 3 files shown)


sashazykov:
A quick update from real-world usage: I've been running Electrum with this patch applied for almost two months now and haven't noticed any regressions.

At this point I honestly can't go back — the difference is big enough that I've changed how I use Electrum: I now close it every day, since starting it no longer takes minutes.

I also haven't seen the exception from #10737 anymore when closing the app (I have both patches applied on my local installation).

PR #10737, opened 2026-07-03
hw_wallet: fix crash on exit if device unpairing fails

sashazykov:
Quitting Electrum with a hardware wallet open crashes the process (SIGABRT) if the paired device was unplugged during the session.

`HW_PluginBase.close_wallet` unpaired the device *before* stopping the keystore `TaskThread`. Unpairing closes the hardware client, which does device I/O and raises if the device is gone:

```
E | plugin | Plugin error. plugin: trezor, hook: close_wallet
  File "electrum/plugin.py", line 833, in run_hook
    r = f(*args)
  File "electrum/hw_wallet/plugin.py", line 89, in close_wallet
    self.device_manager().unpair_pairing_code(keystore.pairing_code())
  File "electrum/plugin.py", line 1118, in unpair_pairing_code
    self._close_client(_id)
  File "electrum/plugin.py", line 1134, in _close_client
    client.close()
  ...
  File "electrum/plugins/trezor/clientbase.py", line 286, in close
    self.client.lock()
  ...
trezorlib.transport.bridge.BridgeException: trezord: acquire/62/null failed with code 400: device not found
```

`run_hook()` swallows the exception, so `keystore.thread.stop()` never runs. The still-running `TaskThread` is a QObject child of the wallet window, so at interpreter shutdown PyQt's atexit cleanup destroys the window and Qt aborts:

```
QThread: Destroyed while thread '' is still running
```

### Fix

- `hw_wallet/plugin.py`: stop the keystore thread before unpairing, so a device-I/O failure cannot leak a running thread. This ordering is also safer in itself: previously the client could be closed while a queued task was still using it.
- `plugin.py`: `DeviceMgr._close_client` treats `client.close()` as best-effort and logs failures — a missing device is a normal condition during cleanup, and this also covers the other unpair callers (disconnect events, session timeouts).

### Testing

- Regression tests added in `tests/plugins/test_hw_wallet.py`.
- Reproduced and verified on macOS with a Trezor One (open hw wallet → pair → unplug → quit): before, the traceback above followed by the Qt fatal and abort; after, a clean shutdown with the bridge failure logged:
  `plugin.DeviceMgr | failed to close hardware client cleanly: BridgeException('trezord: acquire/64/null failed with code 400: device not found')`

### DIFF
--- a/electrum/hw_wallet/plugin.py
+++ b/electrum/hw_wallet/plugin.py
@@ -86,9 +86,11 @@ class HW_PluginBase(BasePlugin, ABC):
     def close_wallet(self, wallet: 'Abstract_Wallet'):
         for keystore in wallet.get_keystores():
             if isinstance(keystore, self.keystore_class):
-                self.device_manager().unpair_pairing_code(keystore.pairing_code())
+                # stop the thread first: if unpairing raises, the thread must not be leaked,
+                # as a still-running QThread would make Qt abort() the process at shutdown
                 if keystore.thread:
                     keystore.thread.stop()
+                self.device_manager().unpair_pairing_code(keystore.pairing_code())
 
     def get_client(self, keystore: 'Hardware_KeyStore', force_pair: bool = True, *,
                    devices: Sequence['Device'] = None,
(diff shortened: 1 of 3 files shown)


f321x:
https://github.com/spesmilo/electrum/pull/10737/commits/b0ae14a5b41164f98a4e1685541f084e71fdecfe lgtm.
Thanks. I dropped the tests, they didn't seem that useful.

PR #10741, opened 2026-07-07
qml: add wallets search option

accumulator:
qml: add wallets search option, allow loading hidden wallet iff search string matches exact wallet name

ref #10715
ref #10688

### DIFF
--- a/electrum/gui/qml/qedaemon.py
+++ b/electrum/gui/qml/qedaemon.py
@@ -72,7 +72,7 @@ class QEWalletListModel(QAbstractListModel):
         wallet_folder = os.path.dirname(self.daemon.config.get_wallet_path())
         with os.scandir(wallet_folder) as it:
             for i in it:
-                if i.is_file() and not i.name.startswith('.'):
+                if i.is_file():
                     available.append(i.path)
         for path in sorted(available):
             wallet = self.daemon.get_wallet(path)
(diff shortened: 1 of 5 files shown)


electrum/gui/qml/qedaemon.py
f321x, line 223: `Daemon.load_wallet()` sets `config.RECENTLY_OPEN_WALLET_FILES` and `config.CURRENT_WALLET` causing the app to open a hidden wallet on the next restart if it was the last opened wallet before.
Should we allow this? In the end the user might be forced to open the app (e.g. with the fingerprint) and then their secret wallet would show up. Maybe we should just open the wallet list on startup or the last non-hidden wallet.

f321x:
Should we make the user aware of this feature somehow, either in the wallet details view (renaming) or in the wizard?
Otherwise I'd expect only few will find this feature.

Considering that the experience of creating a new wallet in the wizard is already stressful for users it might be better to not add even more mental load to the wizard and show it somewhere else?

Or maybe a "Hide" button (e.g. in the rename box or somewhere else in the wallet details view), that when clicked, automatically changes the name and opens a popup explaining the user how they can access their wallet through the search box?

electrum/gui/qml/qedaemon.py
SomberNight: done in https://github.com/spesmilo/electrum/commit/c43cf8e46f5fbbbf361d12bf696baff7ba5094a3

SomberNight:
> Should we make the user aware of this feature somehow, either in the wallet details view (renaming) or in the wizard?
Otherwise I'd expect only few will find this feature.

Sure the feature would not really be discoverable for now. We can improve that later. Anyway, I want it for myself and it's relatively simple code-wise :)

SomberNight:
@accumulator I pushed a follow-up here in the branch. Otherwise your commits look good. Thanks :)

Issue #10717, opened 2026-06-29
DNS-leak outside network-proxy when resolving openalias/lightning-address


SomberNight:
DNSSEC-resolution does not respect the user-configured network-proxy. (ref https://github.com/spesmilo/electrum/issues/4754)

When we try to pay to an "email-address" or "domain", we don't know if it's an openalias or an lnurl-type lightning address. So we resolve both.

When we resolve the openalias, we do DNS queries that do not use the proxy.
I wonder if the proxy is a Tor SOCKS proxy, could we perhaps use the [`RESOLVE` command](https://spec.torproject.org/socks-extensions.html) to fix the DNS-leak?
(perhaps that only works for simple DNS resolution, and not for DNSSEC though?)
ref https://github.com/spesmilo/electrum/commit/bb1aaf60eb31c973c12d210f1010d55604863ba4

If we added support for bip353, that would also be affected.
ref https://github.com/spesmilo/electrum/issues/10531

SomberNight:
tangentially related is
https://github.com/spesmilo/electrum/blob/bb1aaf60eb31c973c12d210f1010d55604863ba4/electrum/util.py#L1961-L1962

(called from https://github.com/spesmilo/electrum/blob/bb1aaf60eb31c973c12d210f1010d55604863ba4/electrum/lnworker.py#L525)

accumulator:
> I wonder if the proxy is a Tor SOCKS proxy, could we perhaps use the [`RESOLVE` command](https://spec.torproject.org/socks-extensions.html) to fix the DNS-leak?

Nope. `RESOLVE` only support `A`/`AAAA`/`PTR` records..

PR #10742, opened 2026-07-07
network: proxy DNSSEC requests if a proxy is configured and enabled

accumulator:
network: proxy DNSSEC requests if a proxy is configured and enabled
adding ConfigVar NETWORK_PROXY_DOH_ENDPOINT (only used for OpenAlias/DNSSEC lookup for now), defaulting to cloudflare's DoH endpoint.

ref #10717

### DIFF
--- a/electrum/dnssec.py
+++ b/electrum/dnssec.py
@@ -36,8 +36,11 @@ import dns
 import dns.name
 import dns.asyncquery
 import dns.dnssec
+import dns.exception
 import dns.message
 import dns.asyncresolver
+import dns.resolver
+import dns.rcode
 import dns.rdatatype
 import dns.rdtypes.ANY.NS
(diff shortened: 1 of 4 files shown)


SomberNight:
ah, sorry, did not see that you *just recently* marked this for 4.8.1.  I did not want to passive-aggressively overrule that xD  The change warrants a bit of thought and it's not that urgent, so I think it could wait. but ofc if someone reviews it, we could have it sooner.

Concept ACK, btw.

accumulator:
> ah, sorry, did not see that you _just recently_ marked this for 4.8.1. I did not want to passive-aggressively overrule that xD The change warrants a bit of thought and it's not that urgent, so I think it could wait. but ofc if someone reviews it, we could have it sooner.

No worries, I added it to 4.8.1 to have it at least considered, but no problem moving it to later.

PR #10744, opened 2026-07-08
android build: cache p4a recipe downloads even for FRESH_CLONE builds

SomberNight:
I propose we cache some downloads even for "fresh clone" builds (that have `ELECBUILD_COMMIT` env var set), with an opt-out reusing the existing `ELECBUILD_NOCACHE` env var. Currently `ELECBUILD_NOCACHE` only opts-out from the docker build cache, now I would reuse it for p4a recipe downloads.

[`qt-everywhere-src-6.10.2.tar.xz`](https://github.com/spesmilo/python-for-android/blob/8c0fcc9ef2e559918ca96ecde6e09fe521bb1427/pythonforandroid/recipes/qt6/__init__.py#L11) is 1.3 GiB, and we download it twice per arch. (once for the qt6 recipe, once for the hostqt6 recipe) It seems to me, download.qt.io is heavily rate-limiting downloads from some IPs.

When doing a release, we build 3 archs, so we download Qt six times. With this patch, we would only download it twice.
(note: I see no need to have separate caches per arch atm.)

Also, when doing multiple builds to debug reproducibility, with this patch, only the first attempt needs to download Qt (twice). Subsequent attempts hit the cache.

---

note: the git tag is part of the cache key out of cache-staleness paranoia. It is not the full commit though, as that would kill the reproducibility-debugging use case.

### DIFF
--- a/contrib/android/build.sh
+++ b/contrib/android/build.sh
@@ -53,8 +53,11 @@ docker build \
 if [ ! -z "$ELECBUILD_COMMIT" ] ; then
     info "ELECBUILD_COMMIT=$ELECBUILD_COMMIT. doing fresh clone and git checkout."
     FRESH_CLONE_BASE=${FRESH_CLONE_BASE:-"/var/tmp/electrum_build/android"}
-    FRESH_CLONE="$FRESH_CLONE_BASE/fresh_clone/electrum"
-    rm -rf "$FRESH_CLONE" 2>/dev/null || ( info "we need sudo to rm prev FRESH_CLONE." && sudo rm -rf "$FRESH_CLONE" )
+    FRESH_CLONE="$FRESH_CLONE_BASE/electrum"
+    rm -rf "$FRESH_CLONE" 2>/dev/null || (
+        info "we need sudo to rm prev FRESH_CLONE." &&
+        sudo chown "$(id -u)" "$FRESH_CLONE_BASE" &&
+        sudo rm -rf "$FRESH_CLONE" )
     umask 0022
     git clone "$PROJECT_ROOT" "$FRESH_CLONE"
     cd "$FRESH_CLONE"
--- a/contrib/android/build.sh
+++ b/contrib/android/build.sh
@@ -32,9 +32,10 @@ mkdir -p "${PROJECT_ROOT}/.buildozer_$1"
 ln -s ".buildozer_$1" ${PROJECT_ROOT}/.buildozer
 
 DOCKER_BUILD_FLAGS=""
+DOCKER_RUN_FLAGS=""
 if [ ! -z "$ELECBUILD_NOCACHE" ] ; then
     info "ELECBUILD_NOCACHE is set. forcing rebuild of docker image."
-    DOCKER_BUILD_FLAGS="--pull --no-cache"
+    DOCKER_BUILD_FLAGS="$DOCKER_BUILD_FLAGS --pull --no-cache"
 fi
 
 if [ -z "$ELECBUILD_COMMIT" ] ; then  # local dev build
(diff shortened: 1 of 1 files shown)


contrib/android/build.sh
```diff
@@ -51,21 +52,31 @@ docker build \
...
     PROJECT_ROOT_OR_FRESHCLONE_ROOT="$FRESH_CLONE"
+    if [ -z "$ELECBUILD_NOCACHE" ] ; then
+        info "ELECBUILD_NOCACHE is not set. mounting p4a download cache."
+        GIT_TAG="$(git describe --abbrev=0)"
+        mkdir -p "$FRESH_CLONE_BASE/downloads_cache/$GIT_TAG/p4a_packages"
+        mkdir -p "$FRESH_CLONE"/.buildozer/android/platform/build-{armeabi-v7a,arm64-v8a,x86,x86_64}/packages
+        DOCKER_RUN_FLAGS="$DOCKER_RUN_FLAGS -v $FRESH_CLONE_BASE/downloads_cache/$GIT_TAG/p4a_packages:/home/user/wspace/electrum/.buildozer/android/platform/build-armeabi-v7a/packages"
+        DOCKER_RUN_FLAGS="$DOCKER_RUN_FLAGS -v $FRESH_CLONE_BASE/downloads_cache/$GIT_TAG/p4a_packages:/home/user/wspace/electrum/.buildozer/android/platform/build-arm64-v8a/packages"
+        DOCKER_RUN_FLAGS="$DOCKER_RUN_FLAGS -v $FRESH_CLONE_BASE/downloads_cache/$GIT_TAG/p4a_packages:/home/user/wspace/electrum/.buildozer/android/platform/build-x86/packages"
+        DOCKER_RUN_FLAGS="$DOCKER_RUN_FLAGS -v $FRESH_CLONE_BASE/downloads_cache/$GIT_TAG/p4a_packages:/home/user/wspace/electrum/.buildozer/android/platform/build-x86_64/packages"
```
SomberNight, line 74: note: `.buildozer/android/platform/build-$ARCH/packages` is the folder p4a internally uses for caching downloads. This PR would just persist the cache through build attempts.

note2: p4a already puts a marker next to each download to detect staleness, with the version in the name. Though I remark it does not work for git clones (see `plyer`):


<details>
 <summary>dir structure of packages folder</summary>

```
$ tree -L2 -a /var/tmp/electrum_build/android/downloads_cache/4.8.0/p4a_packages/
/var/tmp/electrum_build/android/downloads_cache/4.8.0/p4a_packages/
├── cffi
│   ├── .mark-v2.0.0.tar.gz
│   └── v2.0.0.tar.gz
├── hostpython3
│   ├── .mark-Python-3.11.15.tgz
│   └── Python-3.11.15.tgz
├── hostqt6
│   ├── .mark-qt-everywhere-src-6.10.2.tar.xz
│   └── qt-everywhere-src-6.10.2.tar.xz
├── libffi
│   ├── .mark-v3.4.8.tar.gz
│   └── v3.4.8.tar.gz
├── libiconv
│   ├── .mark-libiconv-1.18.tar.gz
│   └── libiconv-1.18.tar.gz
├── libsecp256k1
│   ├── .mark-1a53f4961f337b4d166c25fce72ef0dc88806618.zip
│   └── 1a53f4961f337b4d166c25fce72ef0dc88806618.zip
├── libzbar
│   ├── .mark-bb05ec54eec57f8397cb13fb9161372a281a1219.zip
│   └── bb05ec54eec57f8397cb13fb9161372a281a1219.zip
├── openssl
│   ├── .mark-openssl-3.0.21.tar.gz
│   └── openssl-3.0.21.tar.gz
├── packaging
│   ├── .mark-packaging-26.0.tar.gz
│   └── packaging-26.0.tar.gz
├── plyer
│   ├── .mark-plyer
│   └── plyer
├── pycparser
│   ├── .mark-pycparser-2.22.tar.gz
│   └── pycparser-2.22.tar.gz
├── pycryptodomex
│   ├── .mark-v3.23.0x.tar.gz
│   └── v3.23.0x.tar.gz
├── pyjnius
│   ├── .mark-1.7.0.zip
│   └── 1.7.0.zip
├── pyparsing
│   ├── .mark-pyparsing-3.0.7.tar.gz
│   └── pyparsing-3.0.7.tar.gz
├── pyqt6
│   ├── .mark-pyqt6-6.10.2.tar.gz
│   └── pyqt6-6.10.2.tar.gz
├── pyqt6sip
│   ├── .mark-pyqt6_sip-13.10.3.tar.gz
│   └── pyqt6_sip-13.10.3.tar.gz
├── pyqt_builder
│   ├── .mark-pyqt_builder-1.19.1.tar.gz
│   └── pyqt_builder-1.19.1.tar.gz
├── python3
│   ├── .mark-Python-3.11.15.tgz
│   └── Python-3.11.15.tgz
├── qt6
│   ├── .mark-qt-everywhere-src-6.10.2.tar.xz
│   └── qt-everywhere-src-6.10.2.tar.xz
├── sip
│   ├── .mark-sip-6.15.1.tar.gz
│   └── sip-6.15.1.tar.gz
├── six
│   ├── .mark-six-1.17.0.tar.gz
│   └── six-1.17.0.tar.gz
└── sqlite3
    ├── .mark-sqlite-amalgamation-3500000.zip
    └── sqlite-amalgamation-3500000.zip

24 directories, 43 files
```
</details>

contrib/android/build.sh
```diff
@@ -77,8 +88,8 @@ mkdir --parents "$PROJECT_ROOT_OR_FRESHCLONE_ROOT"/.buildozer/.gradle
 # check uid and maybe chown. see #8261
 if [ ! -z "$ELECBUILD_COMMIT" ] ; then  # fresh clone (reproducible build)
     if [ $(id -u) != "1000" ] || [ $(id -g) != "1000" ] ; then
-        info "need to chown -R FRESH_CLONE dir. prompting for sudo."
-        sudo chown -R 1000:1000 "$FRESH_CLONE"
+        info "need to chown -R FRESH_CLONE_BASE dir. prompting for sudo."
+        sudo chown -R 1000:1000 "$FRESH_CLONE_BASE"
```
f321x, line 95: Calling `chown -R 1000:1000` on `$FRESH_CLONE_BASE` seems to break repeated builds for users without `1000` uid. 
In the first run with `ELECBUILD_COMMIT=X`, `$FRESH_CLONE_BASE` gets chown'd to uid 1000 with sudo fallback, then in the second build attempt the uid 1500 host doesn't have permissions to rm `$FRESH_CLONE`.

```
 => => unpacking to docker.io/library/electrum-android-builder-img:latest  0.0s
💬 INFO:  ELECBUILD_COMMIT=dc1072c93e96f37bdbb76c18742a6d0ed329a93a. doing fresh clone and git checkout.
💬 INFO:  we need sudo to rm prev FRESH_CLONE.
fatal: could not create work tree dir '/var/tmp/electrum_build/android/fresh_clone/electrum': Permission denied
alice@debian:~/electrum$ id -u
1500
```

SomberNight: but it has already worked like that before this PR, hasn't it?

see commit message of https://github.com/spesmilo/electrum/commit/c9b6a6c01e20f66c9d00d517d42947e991b50950

f321x: > but it has already worked like that before this PR, hasn't it?

this doesn't happen on master

> doesn't have permissions to rm $FRESH_CLONE

actually it fails at the step after `rm`, when trying to clone the repo again.

The issue is that now the whole `$FRESH_CLONE_BASE` is owned by uid 1000, and the git clone 

https://github.com/spesmilo/electrum/blob/58ae11d8800e571e8da20b0c727c8573530bc668/contrib/android/build.sh#L57

doesn't have a sudo fallback like the rm before it. On master this worked because the `$FRESH_CLONE` dir was the only dir owned by 1000 and the rm on a subsequent build could utilize the sudo fallback to delete it.
Now all the commands following the `rm` still act in the uid 1000 context (inside `$FRESH_CLONE_BASE`) and fail as a result if the user is another uid.

SomberNight: Hmm, I see, thanks for pointing this out!
Fixed in 7785865c372931e057a9aacee665f2513fb00f8c

Issue #7557, opened 2021-11-08
Native Apple Silicon Support


fishcharlie:
Electrum should release a Universal or separate Apple Silicon build of the application. This would allow users to use the application with increased performance without having to rely on Rosetta 2.

SomberNight:
1. This might be non-trivial to do with our current scripts for codesigning and being able to strip the codesign signatures to verify reproducible builds.
2. Not sure if pyinstaller supports this yet.
    - EDIT(2026-04): looks like it does since https://github.com/pyinstaller/pyinstaller/pull/5581 
3. Looks like one needs xcode 12.2 to build universal2 binaries, which requires macos 10.15 for the build system, which based on past experience would mean bumping the min supported macos version for users to the ~same. (Currently we support 10.13+)
    - EDIT(2026-04): Since https://github.com/spesmilo/electrum/commit/86f6e3dbcb23e5c232655adf5a5f34b68e8dede2, the release build system uses macOS 11.7.10 and xcode 13.2.  We could bump this further a bit if that helped.
4. Note that we would want to build on amd64.
    - that is, ether building a universal2 binary, or two separate x86_64 and arm64 binaries would be fine, however if at all possible the build host machine should stay x86_64

Contributions are welcome.

ballo:
Why was this closed? Apple hasn't sold Intel Macs in a long while

ballo:
pyinstaller can build native binaries BTW
https://github.com/pyinstaller/pyinstaller/issues/5315

Ishaanahuja7:
Any update on this? Qt 5.15.9 community version also supports building on apple silicon now natively. Official Qt download link [here](https://download.qt.io/official_releases/qt/5.15/5.15.9/single)

Ishaanahuja7:
@SomberNight

Ishaanahuja7:
Yes, arm64 can be built from amd64 mac. But min mac os then needs to be updated to 10.15+ and which requires xcode 12.2. Builds will be reproducible as in the amd64 mac case so that shouldn't be an issue.

Ishaanahuja7:
Any updates on this yet? @SomberNight

SomberNight:
We will be moving the build system for the mac binary to a new VM (macOS 10.14.6, xcode 11.3.1 -> macOS 11.7.10, xcode 13.2). See https://github.com/spesmilo/electrum/pull/8738
This itself is unrelated to arm64.

[...]

PR #10745, opened 2026-07-09
contrib/osx: add native arm64 (Apple Silicon) build support

ssotomayor:
## Native Apple Silicon (arm64) macOS build

Ref #7557. Adds an arm64 macOS build, **cross-compiled on the existing x86_64 host**. The x86_64 path is unchanged (byte-identical) when `ELECTRUM_MACOS_ARCH` is unset.

**Approach**
- New `ELECTRUM_MACOS_ARCH` env var (`x86_64` default, `arm64` opt-in). In arm64 mode, native deps + C extensions build **universal2** (so PyInstaller can import them on the x86_64 host); `pyinstaller.spec` then sets PyInstaller's `target_arch` (`--target-architecture`) to thin the bundle down to arm64.
- `PyQt6-Qt6` has no universal2 wheel, so `make_universal_qt.py` lipo-merges the two hash-pinned thin wheels in the venv.
- Ships two thin per-arch `.dmg`s (not universal2), keeping the x86_64 release and the `extract_sigs`/`apply_sigs` flow untouched.
- `apply_sigs.sh`: detect arch per file (`lipo -archs`) instead of hardcoding `x86_64`.

**Verified** (x86_64 host, Xcode CLT 13.2, Big Sur, python.org 3.12.10)
- **Reproducible**: two clean rebuilds → byte-identical `.app` (`95be818c…`, empty `diff`).
- **Codesign + detached-sig flow work for arm64**: self-signed `sign_osx.sh` → `compare_dmg IS_NOTARIZED=false` → `DMGs match. success`.

Note: **Homebrew no longer supports macOS 11** (Tier 3) and fails to build several required tools from source on Big Sur, so I installed `autoconf/automake/libtool/gettext/pkgconfig/coreutils` via **MacPorts**.

### DIFF
--- a/contrib/osx/README.md
+++ b/contrib/osx/README.md
@@ -13,8 +13,16 @@ This guide explains how to build Electrum binaries for macOS systems.
 
 This needs to be done on a system running macOS or OS X.
 
-The script is only tested on Intel-based (x86_64) Macs, and the binary built
-targets `x86_64` currently.
+The script is only tested on Intel-based (x86_64) Macs.
+By default, the binary built targets `x86_64`. To build the binary for
+Apple Silicon (`arm64`) Macs, set the `ELECTRUM_MACOS_ARCH` env var:
+
+    ELECTRUM_MACOS_ARCH=arm64 ./contrib/osx/make_osx.sh
+
+Note that the arm64 binary is *cross-compiled*: the build host stays x86_64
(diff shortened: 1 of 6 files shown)


harryqt:
@SomberNight Can we have this merged please?

ecdsa:
related: https://github.com/csd113/electrum-apple-silicon-patch

PR #10747, opened 2026-07-10
contrib: change release procedure, to avoid reusing tags

SomberNight:
Towards making git tags immutable...
Let's avoid reusing tags and ambiguous binary names.

-----

- primary goal: immutable git tags
    - don't delete tag (e.g. "4.8.0") and then create another with the same name
- secondary goal: immutable distributable names
    - don't make a signed "Electrum-4.8.0-arm64-v8a-release.apk" file publicly downloadable, and then later delete it, and release another identically named file

-----

Problem: it happened several times during releases that we had to "cancel" a release
at the last minute, after there already was a git tag, after we already built binaries,
and sometimes even after those binaries were publicly available on download.electrum.org,
but always before running `release_www.sh` (so before the main webserver linked to them
or before qt gui update notifications were triggered).
At almost any point during making a release we might notice that we cannot reproduce
some distributables, meaning we have to cancel the release. In our current workflow,
very early in the process we create a git tag, e.g. "4.8.0". If we have to cancel
the release and apply some fixes, we will need a new git tag, and in that case historically
we have deleted and reused git tags (so the second attempt would also be from a git tag named "4.8.0" but now pointing to a new commit). This is bad practice. We should aim
to have immutable git tags.

The idea here is:
- construct binary name from:
    version.py::ELECTRUM_VERSION + git commit hash
    (independent of the tag)
- when we build, only create an "-rc1" git tag, e.g. "4.8.0-rc1"
- after we are satisfied to finalise the release, near the very end,
  create another git tag, e.g. "4.8.0", pointing to the same commit

- so in the typical happy path, we will have two git tags:
    - "4.8.0-rc1" == commit1
    - "4.8.0" == commit1
- if something goes wrong and we have to retry, we will have:
    - "4.8.0-rc1" == commit1
    - "4.8.0-rc2" == commit2
    - "4.8.0" == commit2

### DIFF
--- a/contrib/release_www.sh
+++ b/contrib/release_www.sh
@@ -34,8 +34,8 @@ if [ -z "$ELECTRUM_SIGNING_WALLET" ] || [ -z "$ELECTRUM_SIGNING_ADDRESS" ]; then
     exit 1
 fi
 
-VERSION=$("$CONTRIB"/print_electrum_version.py)
-info "VERSION: $VERSION"
+VERSIONB=$("$CONTRIB"/print_electrum_version.py)
+info "VERSIONB: $VERSIONB"
 
 ANDROID_VERSIONCODE_NULLARCH=$("$CONTRIB"/android/get_apk_versioncode.py "null")
 # ^ note: should parse as an integer in the final json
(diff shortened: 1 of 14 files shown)


contrib/release_www.sh
SomberNight, line 61: perhaps we should also put VERSIONC in the version announcement file, e.g. in extradata

contrib/upload.sh
```diff
@@ -31,12 +31,12 @@ fi
 
 sftp -oBatchMode=no -b - "$SSHUSER@uploadserver" << !
    cd electrum-downloads-airlock
-   -mkdir "$VERSION"
-   -chmod 777 "$VERSION"
-   cd "$VERSION"
+   -mkdir "$VERSIONB"
+   -chmod 777 "$VERSIONB"
+   cd "$VERSIONB"
```
SomberNight, line 36: Perhaps we should also name the folder VERSIONC.
That would help f-droid, which is programmatically constructing download URLs, with a limited API.

ref https://gitlab.com/fdroid/fdroiddata/-/blob/85be50faee4b2f78a3641a0af41d62e4d4551766/metadata/org.electrum.electrum.yml#L1480

ref https://f-droid.org/docs/Build_Metadata_Reference/#build_binary

ecdsa: note: with the current PR, nothing seems to clean up the airlock. If we need to change the commit for a given version number, the old file with the old commit name will remain in the airlock forever.

SomberNight:
(edited the OP to clarify the goal is not only "immutable git tags" but also "immutable distributable names")

Issue #10758, opened 2026-07-17
Seed confirmation on mobile (QML): no per-word feedback when typing back seed phrase


c03rad0r:
When creating a new wallet, Electrum asks the user to type back their seed phrase to confirm it was written down correctly. On the mobile (QML/Android) GUI, autocomplete suggests valid BIP39 words, but if the user ignores suggestions and types a word not in the word list, there is no immediate feedback — they can continue adding more words without knowing an earlier word is invalid.

**Root cause:** In `SeedTextArea.qml`, `onTextChanged` only generates suggestions for the last word fragment (`text.split(' ').pop()`). It never validates previously typed words. The custom keyboard appends characters with no per-word check. Full seed validation runs on a debounce timer and only sets a `valid` boolean — individual invalid words are not flagged.

- `SeedTextArea.qml` `onTextChanged` — last-word-only suggestions, no prior word validation:
  https://github.com/spesmilo/electrum/blob/cda4143/electrum/gui/qml/components/controls/SeedTextArea.qml#L56-L66

- `SeedKeyboard` key handler — appends characters without word validation:
  https://github.com/spesmilo/electrum/blob/cda4143/electrum/gui/qml/components/controls/SeedTextArea.qml#L135-L142

- Qt desktop GUI, for comparison, disables suggestions if a prior word is invalid:
  https://github.com/spesmilo/electrum/blob/cda4143/electrum/gui/qt/seed_dialog.py#L338-L343

Is this the intended behavior? If not, would you be open to a PR? I'm thinking of something lightweight — e.g., visually flagging the invalid word when a space follows it, without blocking input.

PR #10759, opened 2026-07-17
fix(qml): add per-word validation feedback in SeedTextArea

c03rad0r:
## Summary

Addresses #10758.

On the mobile (QML/Android) GUI, seed confirmation previously only suggested completions for the current word fragment but provided no feedback when a completed word was not in the BIP39 word list.

This PR adds lightweight per-word validation:

- A new `isWordValid()` method on `QEBitcoin` checks a word against the BIP39 + old-mnemonic wordlist
- `SeedTextArea.qml` validates all completed words (all but the last) on every text change
- A red warning label below the suggestions row shows which word number(s) are invalid
- **Input is not blocked** — the user can continue typing and correct words later

### Files changed

- `electrum/gui/qml/qebitcoin.py` — added `isWordValid()` slot
- `electrum/gui/qml/components/controls/SeedTextArea.qml` — added `_invalidWordIndices` property, validation in `onTextChanged`, and a warning `Label`

### Testing

The change is UI-only and requires the QML GUI to test visually. The logic is straightforward: split text by spaces, check all but the last word against the wordlist.

### DIFF
--- a/electrum/gui/qml/components/controls/SeedTextArea.qml
+++ b/electrum/gui/qml/components/controls/SeedTextArea.qml
@@ -18,7 +18,7 @@ Pane {
 
     // Whether to validate individual words against the wordlist.
     // When true, invalid words are visually flagged (red underline area).
-    // Does not block input — user can still type and correct.
+    // Does not block input - user can still type and correct.
     property bool validateWords: false
 
     property var _suggestions: []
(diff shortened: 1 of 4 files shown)


c03rad0r:
Probably too many incoming PRs for the reviewers closing for now to reduce cognitive load.

PR #10762, opened 2026-07-20
android: show password validation feedback

### LINKED ISSUE #8497: feat: Android: display error in password change modal when password too short
version: Android, 4.4.5.0

if the password is less than 6 characters long, the re-entry field can't be focused on, but the UI won't tell you that this is because there is a minimum length that your password doesn't satisfy.

ZenulAbidin:
## What changed

- show an inline warning when a new password is shorter than six characters
- show an inline warning when the confirmation does not match
- reuse one minimum-length value for both feedback and button validation

## Why

The Android password dialog already prevented submission of passwords shorter than six characters, but the requirement was only encoded in the disabled OK button. Users were not told why the password could not be accepted.

This makes the existing validation visible without changing the password policy.

Fixes #8497.

## Validation

- existing QML test suite: 8 passed
- headless state checks against the real `PasswordDialog` component, including short, mismatched, valid, and non-confirmation modes
- repeated the component checks against an isolated, non-editable install built from the source distribution
- `git diff --check`

### DIFF
--- a/electrum/gui/qml/components/PasswordDialog.qml
+++ b/electrum/gui/qml/components/PasswordDialog.qml
@@ -16,6 +16,7 @@ ElDialog {
     property bool confirmPassword: false
     property string infotext
     property string errorMessage
+    readonly property int minimumPasswordLength: 6
 
     signal passwordEntered(string password)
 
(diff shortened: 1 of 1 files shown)


electrum/gui/qml/components/PasswordDialog.qml
```diff
@@ -78,6 +79,21 @@ ElDialog {
...
+                    if (pw_1.text.length > 0 && pw_1.text.length < minimumPasswordLength)
+                        return qsTr('Password must be at least %1 characters.').arg(minimumPasswordLength)
+                    if (pw_2.text.length > 0 && pw_1.text !== pw_2.text)
+                        return qsTr("Passwords don't match")
+                    return ''
+                }
+                iconStyle: InfoTextArea.IconStyle.Warn
+                backgroundColor: constants.darkerDialogBackground
+                compact: true
+            }
```
accumulator: Instead of adding an extra `InfoTextArea`, the message should preferably be combined with the `errorMessage` `Label` below, into a single element

accumulator (changes requested):

electrum/gui/qml/components/PasswordDialog.qml
ZenulAbidin: Implemented

PR #10764, opened 2026-07-21
json_db: escape '\' and '~' in json patch pointer

f321x:
> Because the characters `~` (%x7E) and `/` (%x2F) have special
   meanings in JSON Pointer, `~` needs to be encoded as `~0` and `/`
   needs to be encoded as `~1` when these characters appear in a
   reference token.

https://www.rfc-editor.org/info/rfc6901/

### DIFF
--- a/tests/test_jsondb.py
+++ b/tests/test_jsondb.py
@@ -152,3 +152,15 @@ class TestJsonDB(ElectrumTestCase):
                 jpatch = jsonpatch.JsonPatch(patches)
                 data = jpatch.apply(data)
                 self.assertEqual(data, {'d': 3})
+
+    async def test_jsondb_pointer_escaping(self):
+        # keys containing '/' or '~' must be escaped per RFC 6901 in emitted patches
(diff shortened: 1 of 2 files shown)


tests/test_jsondb.py
```diff
@@ -152,3 +152,15 @@ async def test_jsondb_replace_after_remove_nested(self):
                 jpatch = jsonpatch.JsonPatch(patches)
                 data = jpatch.apply(data)
                 self.assertEqual(data, {'d': 3})
+
+    async def test_jsondb_pointer_escaping(self):
+        # keys containing '/' or '~' must be escaped per RFC 6901 in emitted patches
+        data = {'labels': {'some/label~key': 'hello'}, 'x1/': {'type': 'bip32'}}
```
SomberNight, line 158: Ah. Wanted to ask how you found this bug but based on the "x1/" key I can guess now :)

SomberNight (approved):
Looks good to me.

PR #10765, opened 2026-07-21
json_db: set_modified after removing incomplete data

f321x:
Set the JsonDB modified after `maybe_load_incomplete_data` removed a corrupted patch. Otherwise we might not override the file on disk, append more patches and lose them all again on the next reload.

### DIFF
--- a/electrum/json_db.py
+++ b/electrum/json_db.py
@@ -126,6 +126,7 @@ class JsonDB(Logger):
                 data, patches = r, []
             elif r := self.maybe_load_incomplete_data(s):
                 data, patches = r, []
+                self.set_modified(True)
             else:
                 raise WalletFileException("Cannot read wallet file. (parsing failed)")
         if not isinstance(data, dict):
@@ -170,7 +171,7 @@ class JsonDB(Logger):
             if n == 0:
                 s = s[0:i]
                 assert s[-2:] == ',\n'
-                self.logger.info('found incomplete data {s[i:]}')
+                self.logger.info('found incomplete data')
                 return self.load_data(s[0:-2])
 
     def set_modified(self, b):
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -91,6 +91,26 @@ class TestWalletStorage(WalletTestCase):
         for key, value in some_dict.items():
             self.assertEqual(d[key], value)
 
+    async def test_appends_after_incomplete_data_recovery_survive_reload(self):
+        storage = WalletStorage(self.wallet_path, allow_partial_writes=True)
+        db = JsonDB('', storage=storage)
+        db.put("a", "b")
+        db.write()
+        # simulate a crash that truncated an appended patch
+        with open(self.wallet_path, "a") as f:
+            f.write(',\n{"op": "add", "path": "/x", "value": {"inco')
+        # reopen: recovery drops the incomplete patch
+        storage = WalletStorage(self.wallet_path, allow_partial_writes=True)
+        db = JsonDB(storage.read(), storage=storage)
+        self.assertEqual(db.get("a"), "b")
+        # append a new patch to the recovered wallet
+        db.put("c", "d")
+        db.write()
+        # reopen: the appended patch must not be lost
+        storage = WalletStorage(self.wallet_path, allow_partial_writes=True)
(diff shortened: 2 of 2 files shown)


electrum/json_db.py
```diff
@@ -170,7 +171,7 @@ def maybe_load_incomplete_data(self, s):
             if n == 0:
                 s = s[0:i]
                 assert s[-2:] == ',\n'
-                self.logger.info('found incomplete data {s[i:]}')
+                self.logger.info('found incomplete data')
                 return self.load_data(s[0:-2])
```
SomberNight, line 175: I guess this is a separate bug but now that I am looking at this:

this whole function incorrectly assumes that "{" and "}" characters cannot appear inside json keys or values, only as control characters.

```diff
diff --git a/tests/test_jsondb.py b/tests/test_jsondb.py
index bb07eaf1a2..db04508525 100644
--- a/tests/test_jsondb.py
+++ b/tests/test_jsondb.py
@@ -152,3 +152,10 @@ class TestJsonDB(ElectrumTestCase):
                 jpatch = jsonpatch.JsonPatch(patches)
                 data = jpatch.apply(data)
                 self.assertEqual(data, {'d': 3})
+
+    async def test_json_db_maybe_load_incomplete_data_with_control_characters(self):
+        # user-supplied text such as tx labels can contain arbitrary characters, such as "}"
+        raw_json_db = '{"a": {"b": "c1"}, "d": "e"},\n{"op": "replace", "path": "/a/b", "value": "user_supp}}}lied_text"}'
+        raw_json_db = raw_json_db[:-4]  # truncate some characters, to trigger maybe_load_incomplete_data()
+        db = JsonDB(raw_json_db)
+        self.assertEqual({"a": {"b": "c1"}, "d": "e"}, dict(db.data))

```

electrum/json_db.py
```diff
@@ -170,7 +171,7 @@ def maybe_load_incomplete_data(self, s):
             if n == 0:
                 s = s[0:i]
                 assert s[-2:] == ',\n'
-                self.logger.info('found incomplete data {s[i:]}')
+                self.logger.info('found incomplete data')
```
SomberNight, line 174: note1: this literal was not marked as f-string
note2: we prob don't want logging secrets

so the change is good.

tests/test_wallet.py
```diff
@@ -91,6 +91,26 @@ def test_write_dictionary_to_file(self):
         for key, value in some_dict.items():
             self.assertEqual(d[key], value)
 
+    async def test_appends_after_incomplete_data_recovery_survive_reload(self):
+        storage = WalletStorage(self.wallet_path, allow_partial_writes=True)
```
SomberNight, line 95: I think what this is testing is so specific to JsonDB that perhaps it would be better to put it into test_jsondb.py. Though I have similar thoughts about the two tests just above. Hmm well, not sure.

electrum/json_db.py
```diff
@@ -126,6 +126,7 @@ def load_data(self, s: str) -> Dict[str, Any]:
                 data, patches = r, []
             elif r := self.maybe_load_incomplete_data(s):
                 data, patches = r, []
+                self.set_modified(True)
```
SomberNight, line 129: note: this works as it triggers a patch-consolidation here:
https://github.com/spesmilo/electrum/blob/eb30c9844c7146250231b9258c024c4553739625/electrum/json_db.py#L114-L116

SomberNight (approved):
Thanks.
how did you find this bug?

electrum/json_db.py
SomberNight: moved to https://github.com/spesmilo/electrum/issues/10766

electrum/json_db.py
f321x: Also `s[i:]` was always empty as it is sliced above.

tests/test_wallet.py
f321x: Yeah i wasn't sure about that either, in the end i put it here because of the similarity to the other two tests.

f321x:
>  how did you find this bug?

I was poking at the new_storage_api branch with Claude Code (Fable) and it pointed out these issues (including #10764 and #10763). It regularly discovers subtle bugs when passing by something.

PR #10768, opened 2026-07-23
qt: rename color themes (the user-visible names of the themes)

### LINKED ISSUE #10767: Light theme does not work when Windows 11 is in Dark mode
### Description

Environment:

Electrum version: 4.8.0
Operating System: Windows 11 Pro 25H2 (latest updates installed)
Windows Appearance: Dark mode
No third-party theme or UI customization software installed (e.g. ExplorerPatcher, StartAllBack, WindowBlinds, etc.)

Description:
When Windows 11 is configured to use Dark appearance, the Light Theme in Electrum does not apply a light theme, selecting View → Theme → Light.

The theme setting is saved correctly (it remains set to Light after restarting Electrum), but the entire application continues to use a dark color palette. The appearance changes only slightly (different shades of dark gray), but it never becomes a true light theme.

Steps to reproduce

Set Windows 11 Appearance to Dark.
Start Electrum 4.8.0.
Select View → Theme → Light.
Restart Electrum.

Expected behavior:
Electrum should display the complete light theme regardless of the Windows appearance setting.

Actual behavior:
The entire application remains dark.

Additional observations:
If Windows Appearance is changed to Light and Electrum is restarted, the Light theme works correctly.
Switching Windows back to Dark causes Electrum's Light theme to become dark again.
Electrum correctly remembers the selected theme.
The issue affects the entire application, not just specific dialogs or widgets.

This suggests that Electrum's Light theme is not fully overriding the Windows/Qt dark palette when Windows is using Dark mode.

Maybe there should be 3 settings: Dark, Light, System. 
Where Light is always light.

SomberNight:
"Light" is actually the Platform-dependent system theme, where we let Qt define the palette and it leave it alone. Qt in turn mostly defers to the OS. On modern windows/macos/gnome/kde/etc, if the OS is set to "dark mode", Electrum will end up using a system-specific dark mode. Hence the name "Light" is confusing.

"Dark" is the qdarkstyle theme, which is largely platform-independent.

I think it's fine if these names are not completely "user-friendly".

closes https://github.com/spesmilo/electrum/issues/10767

### DIFF
--- a/electrum/gui/qt/settings_dialog.py
+++ b/electrum/gui/qt/settings_dialog.py
@@ -205,8 +205,8 @@ class SettingsDialog(QDialog, QtEventListener):
         qr_combo.currentIndexChanged.connect(on_video_device)
 
         colortheme_combo = QComboBox()
-        colortheme_combo.addItem(_('Light'), 'default')
-        colortheme_combo.addItem(_('Dark'), 'dark')
+        colortheme_combo.addItem(_("System"), "default")
+        colortheme_combo.addItem("qdarkstyle", "dark")
         index = colortheme_combo.findData(self.config.GUI_QT_COLOR_THEME)
         colortheme_combo.setCurrentIndex(index)
         colortheme_label = QLabel(self.config.cv.GUI_QT_COLOR_THEME.get_short_desc() + ':')


f321x:
yeah this is less confusing than the current state

PR #10773, opened 2026-07-27
lnwallet: fix fee budget check regression

f321x:
The fee budget during local pathfinding route creation incorrectly
compared the full payment amount against the full budget on every
route instead of the partial amount against its fraction of the budget.
This is a regression introduced in https://github.com/spesmilo/electrum/commit/df5c8c4c98adee8cd477d54afc4a98beb4100639 and makes it possible for a payment to 
exceed the budget or a route to be incorrectly considered too expensive.


SomberNight (approved):
oof...
thanks!

PR #10774, opened 2026-07-27
lnrouter/lnworker: move liquidity hint update to htlc removal callbacks

f321x:
Move the liquidity hint updates from the htlc log processing to the htlc removal callbacks so they
get called for all htlcs we sent out.
On master we would decrement the inflight htlc counter on the liquidity hint only for the first
successful htlc's route, after which `PaymentSuccess` would end the payment flow.
The remaining (successful) routes used for the other htlcs of a payment wouldn't get their counter decremented and 
be penalized for this in subsequent payment attempts until the in-memory liquidity hints get reset through a restart.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2093,14 +2093,6 @@ class LNWallet(Logger):
         or OnionRoutingFailure (if forwarding trampoline).
         """
         if htlc_log.success:
-            if self.network.path_finder:
-                # TODO: report every route to liquidity hints for mpp
-                # in the case of success, we report channels of the
-                # route as being able to send the same amount in the future,
-                # as we assume to not know the capacity
-                self.network.path_finder.update_liquidity_hints(htlc_log.route, htlc_log.amount_msat)
-                # remove inflight htlcs from liquidity hints
-                self.network.path_finder.update_inflight_htlcs(htlc_log.route, add_htlcs=False)
             raise PaymentSuccess()
         # htlc failed
         # if we get a tmp channel failure, it might work to split the amount and try more routes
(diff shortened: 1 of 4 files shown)


electrum/lnworker.py
```diff
@@ -3214,6 +3206,8 @@ def htlc_failed(
         shi = self.sent_htlcs_info.get((payment_hash, chan.short_channel_id, htlc_id))
         if shi and htlc_id in chan.onion_keys:
             onion_key = chan.pop_onion_key(htlc_id)
+            if self.network.path_finder:
+                self.network.path_finder.update_inflight_htlcs(shi.route, add_htlcs=False)
```
f321x, line 3210: `update_liquidity_hints` for failed htlcs is still called through `_process_htlc_log()` as it also might raise `PaymentFailure`, it seems cleaner to just keep it there

electrum/lnworker.py
```diff
@@ -2093,14 +2093,6 @@ async def _process_htlc_log(
         or OnionRoutingFailure (if forwarding trampoline).
         """
         if htlc_log.success:
-            if self.network.path_finder:
-                # TODO: report every route to liquidity hints for mpp
-                # in the case of success, we report channels of the
-                # route as being able to send the same amount in the future,
-                # as we assume to not know the capacity
```
SomberNight, line 2100: I would say this comment contains important information needed to understand what assumptions the liquidity hints make and should be kept (somewhere). It is no longer a TODO but it is still valuable. Could be moved to e.g. the docstring of update_liquidity_hints.

SomberNight: I will add it back in a followup.

electrum/lnworker.py
SomberNight: Ok, agreed. It's also not that important that we could miss some failed htlcs - and not that likely that we miss them in the first place.

SomberNight (approved):
Looks good. Thanks.

electrum/lnworker.py
SomberNight: done in https://github.com/spesmilo/electrum/pull/10778

PR #10776, opened 2026-07-28
lnwallet: use split config channel for single channel mpp

f321x:
Use the channel returned by the split config when sending a
single channel mpp, instead of allowing the pathfinding to use any channel.
Otherwise it can happen that we have a good split config fitting a
specific channel, but the pathfinding then tries to put both parts
on another channel that cannot handle the sum of the parts, raising
at `_assert_can_add_htlc` later on and failing the payment.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2421,7 +2421,6 @@ class LNWallet(Logger):
             receiver_pubkey=paysession.invoice_pubkey,
         )
         for sc in split_configurations:
-            is_multichan_mpp = len(sc.config.items()) > 1
             is_mpp = sc.config.number_parts() > 1
             if is_mpp and not paysession.invoice_features.supports(LnFeatures.BASIC_MPP_OPT):
                 continue
@@ -2520,7 +2519,7 @@ class LNWallet(Logger):
                                     invoice_pubkey=paysession.invoice_pubkey,
                                     r_tags=paysession.r_tags,
                                     invoice_features=paysession.invoice_features,
-                                    my_sending_channels=[channel] if is_multichan_mpp else my_active_channels,
+                                    my_sending_channels=[channel] if is_mpp else my_active_channels,
                                     full_path=full_path,
                                 ))
                             if not is_route_within_budget(
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -597,6 +597,9 @@ class LNPathFinder(Logger):
                         continue
                     if not ignore_amount_constraints \
                             and not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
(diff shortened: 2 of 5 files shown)


electrum/lnworker.py
```diff
@@ -2421,7 +2421,6 @@ async def create_routes_for_payment(
             receiver_pubkey=paysession.invoice_pubkey,
         )
         for sc in split_configurations:
```
SomberNight, line 2423: So there is also a `Channel.can_pay` check done by lnrouter but that is insufficient to avoid the bug you are describing as it does not yet see the other HTLCs as they are not added yet. Here:
https://github.com/spesmilo/electrum/blob/f925081ea3768054904a1412bc48e4e9f6b7605f/electrum/lnrouter.py#L612

Perhaps we should add a comment there stating something along these lines. e.g.:
```diff
diff --git a/electrum/lnrouter.py b/electrum/lnrouter.py
index 023395a685..d50e16917f 100644
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -610,6 +610,9 @@ class LNPathFinder(Logger):
                         continue
                     if not ignore_amount_constraints \
                             and not my_sending_channels[edge_channel_id].can_pay(amount_msat, check_frozen=True):
+                        # note: in case of MPP, we might be constructing multiple routes for different HTLCs around the same time,
+                        #       and only a tiny bit later will the HTLCs get added to the channels. If another HTLC has not been
+                        #       added yet but we already selected to use the same channel for it, this "can_pay()" can return a false positive.
                         continue
                 edge_cost, fee_for_edge_msat = self._edge_cost(
                     short_channel_id=edge_channel_id,
```

f321x: good idea, added the comment in https://github.com/spesmilo/electrum/pull/10776/commits/82b6602a5cb81a3be75535a225671e321a038374

electrum/lnworker.py
```diff
@@ -2520,7 +2519,7 @@ async def create_routes_for_payment(
                                     invoice_pubkey=paysession.invoice_pubkey,
                                     r_tags=paysession.r_tags,
                                     invoice_features=paysession.invoice_features,
-                                    my_sending_channels=[channel] if is_multichan_mpp else my_active_channels,
+                                    my_sending_channels=[channel] if is_mpp else my_active_channels,
                                     full_path=full_path,
```
SomberNight, line 2523: note: in case of `my_sending_channels=my_active_channels`, lnrouter gets an extra degree of freedom as it can choose which channel to use. It could make things a bit simpler if we removed this degree of freedom and just always set `my_sending_channels=[channel]`, i.e. stick to the channel chosen by the splitter. OTOH, unlike the splitter, lnrouter sees the fee cost/penalty of the full paths, so by giving it more freedom it can choose an overall better path.

SomberNight:
thx

PR #10779, opened 2026-07-29
lnrouter: use separate timestamp per liquidity hint value

f321x:
Use a separate timestamp for each value in a `LiquidityHint` instead of a single shared one for all values.

Fixes:
1. We send 1M sat, failure, record cannot_send = 1M
2. Two hours pass, hint expired (shared ts reset)
3. We send 10k, update_can_send resets the shared ts
   -> cannot_send is active again, even though it is outdated

-> We won't try sending >1M sat through the channel until the
process is restarted and the in-memory data is gone.

### DIFF
--- a/electrum/lnrouter.py
+++ b/electrum/lnrouter.py
@@ -25,7 +25,7 @@
 
 import queue
 from collections import defaultdict
-from typing import Sequence, Tuple, Optional, Dict, TYPE_CHECKING, Set, Callable
+from typing import Sequence, Tuple, Optional, Dict, TYPE_CHECKING, Set, Callable, NamedTuple
 import time
 import threading
 from threading import RLock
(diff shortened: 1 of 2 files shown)


electrum/lnrouter.py
```diff
@@ -192,7 +192,8 @@ def __init__(self):
 
     @property
     def can_send_forward(self) -> Optional[int]:
-        return self._can_send_forward.get_valid_amount() if self._can_send_forward else None
+        la = self._can_send_forward
+        return la.get_valid_amount() if la else None
```
SomberNight, line 196: > `return self._can_send_forward.get_valid_amount() if self._can_send_forward else None`

note: the LLM review pointed out that the setters are all locked but the getters are not, and so the pattern used in the first commit is not thread-safe. That observation seems correct, hence my follow-up.

PR #10780, opened 2026-07-30
lnwallet: don't blacklist channels on missing or unchanged channel_update in htlc failure

f321x:
Don't blacklist channels from pathfinding if we receive a htlc failure with a channel_update we already know about or no channel_update at all.
The spec was changed so including a `channel_update` in htlc failure messages is not required anymore (https://github.com/lightning/bolts/pull/1173).
We can just record the liquidity hint and don't try to send the same amount through this channel again.

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2194,13 +2194,18 @@ class LNWallet(Logger):
             raise PaymentFailure(f'payment destination reported error: {failure_msg.code_name()}') from None
 
         # TODO: handle unknown next peer?
-        # handle failure codes that include a channel update
+        # handle failure codes that may include a channel update
         if code in failure_codes:
             offset = failure_codes[code]
             channel_update_len = int.from_bytes(data[offset:offset+2], byteorder="big")
             channel_update_as_received = data[offset+2: offset+2+channel_update_len]
-            payload = self._decode_channel_update_msg(channel_update_as_received)
-            if payload is None:
+            if channel_update_len == 0:
+                # the channel_update became optional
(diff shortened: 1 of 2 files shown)


electrum/lnworker.py
```diff
@@ -2211,17 +2216,15 @@ def handle_error_code_from_failed_htlc(
...
-                if code == OnionFailureCode.TEMPORARY_CHANNEL_FAILURE:
-                    self.network.path_finder.update_liquidity_hints(
-                        route,
-                        amount_msat,
-                        failing_channel=ShortChannelID(failing_channel))
-                # if we can't decide on some action, we are stuck
-                if not (blacklist or handled):
-                    raise PaymentFailure(failure_msg.code_name())
-        # for errors that do not include a channel update
+                assert blacklist or handled, "some action has to be taken on a failure with channel update"
```
SomberNight, line 2219: you changed this condition from `raise PaymentFailure` to an `assert`.
Then I suggest also applying this diff, as an assumption is made that `_handle_chanupd_from_failed_htlc` handles all possible UpdateStatuses:
```diff
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index 785af42375..8ce68f4bb1 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -2274,6 +2274,8 @@ class LNWallet(Logger):
                 handled = True
             else:
                 blacklist = True
+        else:
+            raise Exception(f"unexpected chan upd UpdateStatus: {r}")
         return blacklist, handled
 
     @classmethod

```

f321x: Added in https://github.com/spesmilo/electrum/compare/1b48ac0bec2f847b7222513379350b4332a93c83..a865e336e0db30833cf9be03170b0eb35abdd958

SomberNight:
Looks good. Thanks.

PR #10782, opened 2026-07-30
lnrouter: liquidity hints: add extra penalty if amt near cannot_send

SomberNight:
We guide LNPathFinder away from channels where there is a known cannot_send amount just barely above the HTLC value.

Somewhat follows 6a97e74ce82b9e2055b1ccf31f43288c9becfa02, as before that commit, we were often (unintentionally) blacklisting chans with liquidity failures but now we allow retrying them. Without current change, the liqudity-based penalty would disallow sending an HTLC over the the known "cannot_send" but it would basically not distinguish between 0.999*cannot_send and 0.

Instead, just as a completely naive bandaid, I propose penalising HTLC values greater than 80% of the known cannot_send.


f321x:
lgtm, simple improvement over the current state

> just as a completely naive bandaid

Would it be useful to add a fixme comment to make this more obvious to future readers?

PR #10627, opened 2026-05-04
Protocol 1.7

ecdsa:

PR #10783, opened 2026-07-31
android service onchain monitoring watched addresses/outpoints

accumulator:
Background service for checking onchain addresses/outpoints. Runs once per hour through android's `PeriodicWorkRequest` (periodic work services need the least amount of permissions, execution can be delayed so it doesn't run *exactly* every interval, but can be minutes delayed)

Wallets register their objects of interest in a `json` file, which is read by the service. A notification is shown when conditions are met for address/outpoint. Tapping the notification will open the corresponding wallet.

TODOs
- [ ] background service runs also when app is open, should not emit notifications for wallets that are already open in the app
- [ ] notifications can pass multiple wallets to open, app only opens first

### DIFF
--- a/electrum/gui/qml/service/chainwatch_service.py
+++ b/electrum/gui/qml/service/chainwatch_service.py
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+# p4a service entry point for "Chainwatch". Launched by org.electrum.worker.ElectrumWorker
+# once per period.
+#
+# Declared in contrib/android/buildozer_qml.spec as:
+#   services = Chainwatch:electrum/gui/qml/service/chainwatch_service.py:foreground
+#
+# Runs in a separate process with its own interpreter.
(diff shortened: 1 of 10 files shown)


electrum/gui/qml/service/chainwatch_service.py
```diff
@@ -0,0 +1,429 @@
...
+        return txids, confirmations
+
+    def _query_outpoint(outpoint):
+        """Spend detection: find the tx that spends this specific outpoint.
+
+        The funding tx gives us the watched output's scriptpubkey -> scripthash;
+        its history contains the funder plus anything spending from that script.
+        A history tx spends our outpoint iff one of its inputs' prevout equals
+        txid:idx. An outpoint is spent at most once, so we stop at the first hit
+        and report that tx and its confirmations."""
```
ecdsa, line 292: This will probably not get merged before #10627, so it would be reasonable to use outpoint subscriptions.

Issue #4327, opened 2018-05-01
sign message: inconsistent handling of whitespaces


SomberNight:
Message signing/verification gets stripped of leading/trailing whitespaces in Qt, but not when e.g. using the `signmessage` command.

https://github.com/spesmilo/electrum/blob/bc76d63f61961d1a875ba32b07da7fbe8fc7a4d1/gui/qt/main_window.py#L2148
https://github.com/spesmilo/electrum/blob/bc76d63f61961d1a875ba32b07da7fbe8fc7a4d1/gui/qt/main_window.py#L2176

SomberNight:
In the command line, having whitespaces is very explicit, the user is fully aware;
unlike in Qt, where it's not so clear, so it's better to strip there.

fametrano:
For the record: Bitcoin Core's Qt GUI does *not* strip whitespace. Both signing and verifying take the message verbatim — `toPlainText().toStdString()` with no `trimmed()`:

https://github.com/bitcoin/bitcoin/blob/556988790a7f961693a8fd93f73725baea66476a/src/qt/signverifymessagedialog.cpp#L141
https://github.com/bitcoin/bitcoin/blob/556988790a7f961693a8fd93f73725baea66476a/src/qt/signverifymessagedialog.cpp#L202

which is consistent with its RPCs (`signmessage`, `signmessagewithprivkey`, `verifymessage`) and with `MessageHash` hashing the exact bytes.

So the two Qt GUIs disagree: a signature produced here over " hello " actually commits to "hello" and verifies in Bitcoin-Qt only against the trimmed text, while a Bitcoin-Qt signature over " hello " never verifies here, even when the exact message is pasted.

PR #10787, opened 2026-08-01
qml: strip whitespace from message before signing, as in qt gui

fametrano:
The qt gui strips leading/trailing whitespace from the message (and address) both when signing (`do_sign`) and when verifying (`do_verify`) in `main_window.py`. This behaviour was discussed in #4327 and kept as intended.

The qml gui already strips in `verifyMessage` (`qedaemon.py`), but not in `signMessage` (`qewallet.py`): a message with leading/trailing whitespace (e.g. a trailing newline) signed in the qml gui produces a signature that fails verification even in the same Sign/Verify dialog, and is inconsistent with the qt gui.

This PR strips address and message in `signMessage` too, replicating the qt gui behaviour.

**Note:** this PR takes the current gui behaviour (#4327) as given and just fixes the qml inconsistency. The complementary alternative #10788 removes the message trimming everywhere in the gui instead, aligning with Bitcoin Core (which does not trim, not even in its gui) and with the electrum CLI/RPC — in my opinion the better resolution of the two. The two PRs are mutually exclusive: whichever is preferred, the qml sign/verify asymmetry should not stay as it is.

Related: #4327
Related: https://github.com/btclib-org/btclib/issues/200

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/electrum/gui/qml/qewallet.py
+++ b/electrum/gui/qml/qewallet.py
@@ -846,6 +846,9 @@ class QEWallet(AuthMixin, QObject, QtEventListener):
     @pyqtSlot(str, str)
     @auth_protect(message=_("Sign message?"))
     def signMessage(self, address, message):
+        # strip, as in qt gui and in qml verifyMessage (see #4327)
+        address = address.strip()
+        message = message.strip()
         sig = self.wallet.sign_message(address, message, self.password)
         result = base64.b64encode(sig).decode('ascii')
         self.messageSigned.emit(result)


SomberNight:
Hey. Thanks for reporting this.
Clearly it is a bug that even the qml sign/verify dialog is not self-consistent.

The bug suggests that this code should be refactored and the stripping should not be done by the callers. If we want stripping, it should be done in a more consistent way. I am refactoring the relevant code but will merge your fix now first.

PR #10791, opened 2026-08-02
wallet: sign_message: strip whitespaces in GUis, do not strip in CLI

SomberNight:
(stacked on top of https://github.com/spesmilo/electrum/pull/10790, which is mostly clean-up, merge that first)

-----

The design decision to make is if we stand-by the stance from https://github.com/spesmilo/electrum/issues/4327:
> In the command line, having whitespaces is very explicit, the user is fully aware;
unlike in Qt, where it's not so clear, so it's better to strip there.

-----

- stripping whitespaces leads to better UX for GUIs IMO, but it's counter-productive for CLI
- our previous behaviour was unintentionally inconsistent:
  - qt stripped the message in both sign and verify
  - qml stripped the message in verify but not in sign (until https://github.com/spesmilo/electrum/pull/10787)
  - cli stripped the message in both sign and verify
  - qml also stripped the signature in verify, which was not done anywhere else
- now this code is de-duped, and the default becomes stripping both message and signature; with the CLI explicitly opting out of that

closes https://github.com/spesmilo/electrum/pull/10787
closes https://github.com/spesmilo/electrum/pull/10788

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -918,7 +918,12 @@ class Commands(Logger):
             raise UserFacingException(f"address must be a str instead of {type(address)}")
         if not isinstance(message, str):
             raise UserFacingException(f"message must be a str instead of {type(message)}")
-        sig = wallet.sign_message(address=address, message=message, password=password)
+        sig = wallet.sign_message(
+            address=address,
+            message=message,
+            password=password,
+            strip_inputs=False,  # respect whitespaces for CLI
+        )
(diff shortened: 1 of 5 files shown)


SomberNight:
> The design decision to make is if we stand-by the stance from https://github.com/spesmilo/electrum/issues/4327

I think we stand-by that.
Hence this PR does not change the intent of the code and is just a clean-up.

PR #10794, opened 2026-08-03
add crandom.py, for our RNG, where we mix stuff with os.urandom

SomberNight:
(still WIP)

Currently we rely on `os.urandom()` ~everywhere for cryptographically secure randomness. Existing code already checks at runtime that the output of `os.urandom()` looks at least somewhat sane (see if it compresses with zlib) and hard-fails if it does not.

However, `os.urandom` could still be subtly "broken" (undetected by us) and produce bad quality output. That's the motivation of the new code here. We expect `os.urandom` to work well, BUT if it undetectably does not, mixing in other sources of entropy mitigates the situation somewhat.

This PR introduces a new module `crandom.py` that manages the RNG, and that our other code should call. Extreme care should be taken here not to make things worse than the status quo by "rolling our own" thing.

The logic is split across two modules: `crandom.py` and `crandom_env.py`.
- The core sensitive logic (RNG mixing, extracting random bytes) is in `crandom.py`, which is absolutely security critical and is kept concise to ease review.
- `crandom_env.py` contains secondary sources of entropy and potentially platform-specific code. It is just a companion to `crandom.py` and is only intended to be accessed from there. Even if all the entropy sources listed in `crandom_env.py` are broken, assuming `os.urandom()` produces high quality random, `crandom.py` should never produce low-quality random output. Hence `crandom_env.py` is much less critical to review in depth.

This is inspired by https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.h#L25, which is significantly more advanced. But I wanted to (1) lessen our dependence on `os.urandom()`, (2) while keeping it simple.

### DIFF
--- a/electrum/crandom.py
+++ b/electrum/crandom.py
@@ -3,19 +3,151 @@
 # file LICENCE or http://www.opensource.org/licenses/mit-license.php
 #
 # Cryptographically secure RNG.
+#
+# This mostly uses os.urandom, and extreme care should be taken here not to make things worse
+# compared to just directly using that.
+# We check os.urandom is not trivially broken (passes the zlib test), in which case we panic and runtime exit.
+# However, os.urandom could still be subtly "broken" (undetected by us) and produce bad quality output.
+# That's the motivation of all this code. We expect os.urandom to work well, BUT if it undetectably does not,
+# hopefully mixing in other sources of entropy mitigates the situation somewhat.
+#
+# inspired by https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.h#L25
+#
+# The logic is split across two modules: crandom.py and crandom_env.py.
+# - The core sensitive logic (RNG mixing, extracting random bytes) is in this module (crandom.py),
+#   which is absolutely security critical and is kept concise to ease review.
+# - crandom_env.py contains secondary sources of entropy and potentially platform-specific code.
+#   Even if all the entropy sources listed in crandom_env.py are broken, assuming os.urandom()
+#   produces high quality random, this module should never produce low-quality random output.
 
+import hashlib
 import os
-import secrets
+import threading
+from typing import Callable
+import zlib
+
+from . import crandom_env
 
 
-def get_rand_bytes(nbytes: int) -> bytes:
-    """Returns uniformly distributed bytes, of length nbytes."""
-    assert nbytes >= 0, nbytes
-    return os.urandom(nbytes)
+# Check that os.urandom works
+length = len(zlib.compress(os.urandom(1000)))
+if length <= 900:
+    raise ImportError("Broken PRNG. Refusing to continue. Exiting...")
 
 
-def get_rand_below(upper_bound: int) -> int:
-    """Return a uniformly distributed int in the range [0, upper_bound)."""
-    assert upper_bound > 0, upper_bound
-    return secrets.randbelow(upper_bound)
+def sha512(x: bytes) -> bytes:
+    assert isinstance(x, bytes)
+    return hashlib.sha512(x).digest()
 
+
+CRANDOM_FEEDER_API = Callable[[bytes | str | int], None]
+
+
+class RNGState:
+
+    def __init__(self):
+        self.lock = threading.Lock()
+        self._state = os.urandom(32)  # secret!  access needs lock.
+        # gather ghetto-entropy:
+        self.rand_add_refresh()  # clock
+        crandom_env.rand_add_static_env(self.feed_entropy)
+        self.rand_add_refresh()  # clock again
+
+    def rand_add_refresh(self) -> None:
+        """Gather dynamic environment data that changes over time and mix it in.
+        This includes a high-precision clock.
+
+        Never raises.
+        """
+        crandom_env.rand_add_dynamic_env(self.feed_entropy)
+
+    def feed_entropy(self, data: bytes | str | int) -> None:
+        """Mix in some data into our internal RNG state, in hopes of increasing entropy.
+
+        We MUST be robust for given 'data' not to contain any randomness, it could even be static.
+        Assuming our internal hash function is cryptographically secure, our internal state
+        MUST not be left with less entropy than before the call.
+
+        Never raises (assuming input type-checks).
+        Matches CRANDOM_FEEDER_API.
+        """
+        if not data:
+            return
+        if isinstance(data, int):
+            data = hex(data)
+        if isinstance(data, str):
+            # we must not raise UnicodeError, hence "backslashreplace"
+            data = data.encode("utf-8", errors='backslashreplace')
+        with self.lock:
+            self._state = sha512(data + self._state)[0:32]
+            assert len(self._state) == 32
+
+    def _mix_extract(self) -> bytes:
+        """Return 32 bytes of secure randomness.
+
+        Mix in some new entropy from os.urandom first, and then extract 32 bytes.
+        When mixing in new entropy, H = SHA512(new_entropy || old_rng_state) is computed, and
+        the first 32 bytes of H are produced as output, while the last 32 bytes
+        become the new RNG state.
+
+        Never raises.
+        """
+        with self.lock:
+            fresh_entropy = os.urandom(32)
+            h = sha512(fresh_entropy + self._state)
+            out, self._state = h[0:32], h[32:64]
+            assert len(out) == 32
+            assert len(self._state) == 32
+            return out
+
+    def get_rand_bytes(self, nbytes: int) -> bytes:
+        """Returns uniformly distributed bytes, of length nbytes.
+
+        Never raises.
+        """
+        assert nbytes >= 0, nbytes
+        out = b""
+        while len(out) < nbytes:
+            out += self._mix_extract()
+        return out[:nbytes]
+
+    def _get_rand_bits(self, nbits: int) -> int:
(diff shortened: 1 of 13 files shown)


electrum/crandom.py
```diff
@@ -0,0 +1,153 @@
...
+            assert len(out) == 32
+            assert len(self._state) == 32
+            return out
+
+    def get_rand_bytes(self, nbytes: int) -> bytes:
+        """Returns uniformly distributed bytes, of length nbytes.
+
+        Never raises.
+        """
+        assert nbytes >= 0, nbytes
```
renepickhardt, line 126: also nbytes needs to by smaller equal than 32 I guess

electrum/crandom.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        if not data:
+            return
+        if isinstance(data, int):
+            data = hex(data)
+        if isinstance(data, str):
+            # we must not raise UnicodeError, hence "backslashreplace"
+            data = data.encode("utf-8", errors='backslashreplace')
+        with self.lock:
+            self._state = sha512(data + self._state)[0:32]
```
renepickhardt, line 93: this also applies to other places. I am not sure about taking the first half of SHA512. Thus wouldn't it make more sense to do `self._state = XOR(SHA256(data),self._state)`. 

The - admitedly paranoid - reasoning: The hash is an additional component capable, at least in principle, of damaging a good source if the implementation or construction is defective while xor should always work?

electrum/crandom.py
```diff
@@ -0,0 +1,153 @@
...
+        Mix in some new entropy from os.urandom first, and then extract 32 bytes.
+        When mixing in new entropy, H = SHA512(new_entropy || old_rng_state) is computed, and
+        the first 32 bytes of H are produced as output, while the last 32 bytes
+        become the new RNG state.
+
+        Never raises.
+        """
+        with self.lock:
+            fresh_entropy = os.urandom(32)
+            h = sha512(fresh_entropy + self._state)
```
renepickhardt, line 115: see comment from `feed_entropy` in line 82. I think xor preserves the randomness property better if at least one source was random

electrum/crandom.py
```diff
@@ -0,0 +1,153 @@
...
+import os
+import threading
+from typing import Callable
+import zlib
+
+from . import crandom_env
+
+
+# Check that os.urandom works
+length = len(zlib.compress(os.urandom(1000)))
```
renepickhardt, line 35: I am not sure about this sanity check. I think there is a non zero chance that true random output is compressable. That probability should be small however and I see what you are trying to achieve here but I am not sure if other means are more useful

electrum/crandom.py
```diff
@@ -0,0 +1,153 @@
...
+
+_rng = RNGState()
+
+
+########################################
+# External API (thread-safe):
+
+get_rand_bytes = _rng.get_rand_bytes
+get_rand_below = _rng.get_rand_below
+feed_entropy = _rng.feed_entropy
```
renepickhardt, line 192: It absolutely makes sense to expose this API. however I am not sure if users who need this API will understand that they need it and build their application accordingly. See other comment in `__init__`

electrum/crandom.py
```diff
@@ -0,0 +1,153 @@
...
+
+class RNGState:
+
+    def __init__(self):
+        self.lock = threading.Lock()
+        self._state = os.urandom(32)  # secret!  access needs lock.
+        # gather ghetto-entropy:
+        self.rand_add_refresh()  # clock
+        crandom_env.rand_add_static_env(self.feed_entropy)
+        self.rand_add_refresh()  # clock again
```
renepickhardt, line 59: as far as I understand you pull OS specific values when creating the state `RNGState` object (for example at startup of electrum) if users do this within a local machine and clone the local machine I believe the additional source of randomness is just copied and never updated. I understand that `feed_entropy` could help here but I wonder if the feeding of additional runtime components should not also occasionally happen during random number generation.

electrum/crandom_env.py
```diff
@@ -0,0 +1,91 @@
...
+def rand_add_static_env(feed: 'CRANDOM_FEEDER_API') -> None:
+    """Gather non-cryptographic environment data that does not change over time
+    and feed it into feed().
+    """
+    # os
+    feed(str(os.environ))
+    feed(getattr(os, "ctermid", lambda: "")())
+    feed(os.getcwd())
+    feed(str(os.get_exec_path()))
+    feed(str(os.getgroups()))
```
renepickhardt: Seems to be supported only under linux? c.f. https://docs.python.org/3/library/os.html#os.getgroups

renepickhardt (commented):
Disclaimer: I am no expert with randomness and security of it. So please take this feedback not as advise or proper audit. It is just some thoughts that I had while looking over this PR. 

In particular not commenting on things does not mean I approve them as secure or good. It rather means I have no opinion / idea. 

As a general question / comment I wonder if pulling all the data in `crandom_env` might not be predictable or at least bruteforcable in some scenarios.

electrum/crandom_env.py
SomberNight: Indeed, thanks. Fixed in ac41f354534434326f49ecb0fa1175ff6ffacdeb, though I am still uncertain exactly how to structure the code to nicely catch and silence these platform-specific exceptions.

electrum/crandom.py
SomberNight: No, any reasonably large value should work. There is a test case `test_get_rand_bytes_api` that covers this. `_mix_extract` keeps pulling 32 bytes potentially many times, until we have enough.

electrum/crandom.py
SomberNight: In general XOR can be dangerous here as it can cancel out. Though your combined sha+xor usage looks okay: but even there if you feed the same "data" twice, that part cancels out.

With another hash function call, that could be fixed:
`self._state = XOR(SHA256(data),SHA256(self._state))`
but depending on what you have in mind, that might already "damage" the state

However, TBH if sha256/sha512 is broken, we probably will not even be able to correctly derive private keys.
(for example, bip32 uses sha512)

electrum/crandom.py
SomberNight: This construction is based on Bitcoin Core's: https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.h#L63
though in their implementation, they also have a counter, I wonder if that helps though:
https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.cpp#L424

electrum/crandom.py
SomberNight: This check is pre-existing on master and has been there for 10 years, since https://github.com/spesmilo/electrum/commit/e8d30129eae10b5efeaa9e95a900576e41dd2bf1

(Though it was only recently moved to early-init, in https://github.com/spesmilo/electrum/commit/dceece1cda9bb21ec8544b2ab88876277ed6f45a. Prior to that, it is unclear how early it got executed, but I think it usually got imported via e.g. `gui/qt/__init__` -> `gui/qt/main_window` -> `paymentrequest` -> `rsakey`)

The check is bare-bones. I think it failing is extremely unlikely if the RNG is at all decent. If it fails, it's perfectly reasonable to bail out. We could try to add more, better, checks for the RNG quality, yes - though not sure how exactly.

electrum/crandom.py
SomberNight: Electrum, the wallet, itself is the primary consumer of this module (even though `electrum` can be used as a python library of course).

Regardless, even if code external to the module never calls `feed_entropy`, `get_rand_bytes` will not produce unsafe output. `get_rand_bytes`->`_mix_extract` always mixes in `os.urandom()` into the result, so the output of `get_rand_bytes` should never be worse quality than directly calling `os.urandom()`.

electrum/crandom.py
SomberNight: > I wonder if the feeding of additional runtime components should not also occasionally happen during random number generation

Added a `strengthen()` method in https://github.com/spesmilo/electrum/pull/10794/changes/3b61150148b90e781e224108b0b9b52061d94c7f, which `mix_extract()` calls if it hasn't been called in a while. `strengthen()` spends some CPU-time hashing the internal state and also calls `rand_add_refresh()` to extract some entropy from the dynamic environment.

electrum/crandom.py
SomberNight: > I think there is a non zero chance that true random output is compressable

I think we can completely disregard false-positive panics.

-----

It is questionable though whether the check is useful at all, as, at least on Linux, both /dev/random and /dev/urandom (which are the sources os.urandom is using) output the result of a cryptographic hash function (see e.g. https://lore.kernel.org/lkml/20220106235920.995517-1-Jason@zx2c4.com/ where sha1 was replaced with blake2s), completely masking the quality of the underlying source entropy. I think the hash output will ~never compress well, even if the original entropy is non-existent. I cannot think of any test we could practically apply to such hash output to guess its "true entropy".

electrum/crandom.py
SomberNight: note: blockstream jade uses the same construction, seemingly also copied from Bitcoin Core:
https://github.com/Blockstream/Jade/blob/15ce915a20898dda4ca0e3d7ba55ca556f5271f2/main/random.c#L134

PR #10799, opened 2026-08-05
qml/android: protect WIF keys from screenshots in more places

SomberNight:
also fix bug in AddressDetails where we were never unsetting the secure flag

### DIFF
--- a/electrum/gui/qml/components/ImportAddressesKeysDialog.qml
+++ b/electrum/gui/qml/components/ImportAddressesKeysDialog.qml
@@ -123,4 +123,11 @@ ElDialog {
         id: bitcoin
     }
 
+    Binding {
+        target: AppController
+        property: 'secureWindow'
+        when: root.visible  // enables stacking multiple secureWindow dialogs
+        value: true
+    }
+
 }
--- a/electrum/gui/qml/components/WalletDetails.qml
+++ b/electrum/gui/qml/components/WalletDetails.qml
@@ -677,6 +677,7 @@ Pane {
     Binding {
         target: AppController
         property: 'secureWindow'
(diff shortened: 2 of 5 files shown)


electrum/gui/qml/components/ImportAddressesKeysDialog.qml
```diff
@@ -123,4 +123,10 @@ ElDialog {
         id: bitcoin
     }
 
+    Binding {
+        target: AppController
+        property: 'secureWindow'
+        value: root.visible
+    }
```
SomberNight, line 131: the clanker correctly flagged that this construction does not compose: if we stack two "secure" dialogs and then pop one of them, the remaining "secure" dialog will have `secureWindow == false`

SomberNight: fixed in https://github.com/spesmilo/electrum/pull/10799/changes/65aae21ae15a332f1c8704235c0fc4eb2b09910d

electrum/gui/qml/components/WalletDetails.qml
```diff
@@ -677,7 +677,8 @@ Pane {
     Binding {
         target: AppController
         property: 'secureWindow'
-        value: seedText.visible
+        when: rootItem.visible  // enables stacking multiple secureWindow dialogs
+        value: true
```
accumulator: Now this page will always be 'secured', even when no secrets are shown?

SomberNight: thanks, that was not intentional.
force-pushed https://github.com/spesmilo/electrum/compare/65aae21ae15a332f1c8704235c0fc4eb2b09910d..739cba5d7e582571cc7a7201294b5f66d7a9b52f

Issue #10804, opened 2026-08-06
Warn users about ColdCard Mk3 seed-entropy flaw (Coinkite advisory, Jul 2026)


nrobi144:
## Summary

On 30–31 July 2026, Coinkite published a security advisory disclosing that a firmware build error caused ColdCard seed generation to silently fall back to a software RNG (MicroPython) instead of the STM32 hardware RNG. On the **Mk3 this reduced generated-seed entropy to roughly ~40 bits** (vs. the intended 128) — brute-forceable on commodity hardware. Approximately **594 BTC (~$38M) was drained from ~500 wallets on 31 July 2026** (Coinkite's investigation into that sweep is ongoing).

Affected, per Coinkite:
- **Mk2 / Mk3:** firmware **4.0.1 – 4.1.9** (fixed in **v4.2.0**+; device-generated seeds)
- **Mk4 / Mk5 / Q:** reduced to ~72 bits before fixed firmware (standard **v5.6.0** / Edge **v6.6.0X**; Q **v1.5.0Q** / Edge **v6.6.0QX**)
- Seeds created with **≥50 user dice rolls**, or protected by a strong BIP-39 passphrase, are not affected in the same way.
- Fix: update firmware, generate a **new** seed, and migrate funds.

Advisory: https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/

## Why this concerns Electrum

Electrum documents and supports ColdCard as a signing device:
- Docs: https://electrum.readthedocs.io/en/latest/hardware-linux.html (dedicated "Coldcard" subsection; ColdCard listed among supported devices) — source in `docs/`
- Plugin: `electrum/plugins/coldcard/` (`coldcard.py`, `README.md`, etc.)

Electrum does **not** generate the ColdCard seed (the device does), so this is **not a flaw in Electrum's code**. But users who set up ColdCard following Electrum's docs may hold funds on affected seeds and would reasonably look to Electrum for guidance.

## Suggested changes

1. **Add a visible warning** on the ColdCard setup docs (and, where practical, in the ColdCard plugin/connect UI) linking Coinkite's advisory and the fixed-firmware / reseed guidance.
2. **Deprioritize ColdCard** from prominent "supported devices" placement until a user is known to be on fixed firmware with a freshly generated seed, so new users aren't steered onto an affected setup path.

Happy to open a PR for the docs/UI change if that's welcome.

## Disclosure note

[...]

PR #10805, opened 2026-08-06
docs: add Coldcard Mk3 seed-entropy security notice to plugin README

nrobi144:
## Summary

Adds a short **Security notice** to the Coldcard plugin README (`electrum/plugins/coldcard/README.md`).

In July 2026 Coinkite [disclosed](https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/) a firmware build error that reduced generated-seed entropy on **Coldcard Mk3** (firmware 4.0.1–5.0.3) and pre-fix **Mk4/Q**. Electrum doesn't generate the Coldcard seed, so this isn't an Electrum flaw — but users who set up a Coldcard via Electrum may hold funds on an affected seed. Docs-only change.

Refs #10804.

### DIFF
--- a/electrum/plugins/coldcard/README.md
+++ b/electrum/plugins/coldcard/README.md
@@ -1,5 +1,9 @@
 # Coldcard Hardware Wallet Plugin
 
+## ⚠ Security notice
+
+Coinkite disclosed (July 2026) a firmware build error that reduced generated-seed entropy on **Coldcard Mk3** (firmware 4.0.1–5.0.3) and pre-fix **Mk4/Q**. Seeds generated on-device without extra dice rolls may be brute-forceable. Update to fixed firmware, generate a new seed and move your funds. See the [Coinkite advisory](https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/).
+
 ## Just the glue please
 
 This code connects the public USB API and Electrum. Leverages all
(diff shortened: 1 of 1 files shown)


ecdsa:
thanks. please remove unicode from your notice; we ban unicode from our codebase for security reasons

electrum/plugins/coldcard/README.md
```diff
@@ -1,5 +1,9 @@
 # Coldcard Hardware Wallet Plugin
 
+## ⚠ Security notice
+
+Coinkite disclosed (July 2026) a firmware build error that reduced generated-seed entropy on **Coldcard Mk3** (firmware 4.0.1–5.0.3) and pre-fix **Mk4/Q**. Seeds generated on-device without extra dice rolls may be brute-forceable. Update to fixed firmware, generate a new seed and move your funds. See the [Coinkite advisory](https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/).
```
SomberNight: > please remove unicode from your notice

(using unicode dashes vs ascii hyphen-minus)
Can't even be bothered to write a single line of text by hand anymore, huh?

nrobi144: @SomberNight @ecdsa  updated. Scanning and updating references to ColdCards in multiple repos is way more effective with LLMs. Why would I do that manually if it's slower and more error prone?

PR #10809, opened 2026-08-06
android: download pinned platform-tools

accumulator:
platform-tools is actually not used, but pulled in by gradle/AGP. Google's repo XML only ever advertises the single newest revision. So when AGP's auto-installer resolves it, it can only install "latest".

Adding `android.builder.sdkDownload=false` to `gradle.properties` inhibits platform-tools download, and without it we still get a successful build.

~~pre-installing (older) platform-tools does avoid gradle/AGP downloading the newest version.~~


accumulator:
> pre-installing (older) platform-tools does avoid gradle/AGP downloading the newest version.

Seems that is not true, `android.builder.sdkDownload=false` apparently needed

ecdsa:
> Seems that is not true, android.builder.sdkDownload=false apparently needed

can you clarify what is not true? If this property is set, does it make the pre-installing redundant? or do we need both?

accumulator:
> > Seems that is not true, android.builder.sdkDownload=false apparently needed
> 
> can you clarify what is not true? If this property is set, does it make the pre-installing redundant? or do we need both?

it is not true that pre-installing a not-latest version of platform-tools inhibits downloading the latest version of platform-tools.

So if we want to inhibit installing a latest version of platform-tools, we must add `android.builder.sdkDownload=false` to `build.properties`. This allows us to pin platform-tools to a particular version.

ecdsa:
so, again, does setting that property make the pre-installing in this PR redundant?

I am asking because, if this is redundant, then you might as well close this pull request instead of marking it as draft.

accumulator:
We can go three ways here:
1. do nothing; platform-tools gets pulled in, but is not used in the build
2. only add `android.builder.sdkDownload=false` to our p4a branch; platform-tools is not pulled in, builds ok regardless
3. add `android.builder.sdkDownload=false` to our p4a branch AND merge this PR; platform-tools is pinned, builds ok as it is not used anyway

So yeah, fine by me to close this and go with (1), unless paranoid then go with either (2) or (3)

ecdsa:
Ah, thank you for answering my question.

And what is the point of downloading (3) if it is not needed?

accumulator:
> Ah, thank you for answering my question.
> 
> And what is the point of downloading (3) if it is not needed?

exactly :) 

(I noticed the build doesn't actually need it *after* making this PR, so..)

Option (2) might be the better option, as the build will fail iff platform-tools will be a requirement in the future and we don't want to have an unpinned platform-tools *ever*

ecdsa:
> exactly :)

well, you could have said that earlier, instead of me spending 30 minutes figuring it out

> (I noticed the build doesn't actually need it after making this PR, so..)

in that case the correct thing to do is to close this pull request, instead of leaving it around as draft

SomberNight:
> only add android.builder.sdkDownload=false to our p4a branch; platform-tools is not pulled in, builds ok regardless

> Option (2) might be the better option, as the build will fail iff platform-tools will be a requirement in the future and we don't want to have an unpinned platform-tools ever

Sounds good. We should do option 2.

SomberNight:
Thank you for looking into this.

accumulator:
#10832

PR #10813, opened 2026-08-07
android: limit biometric auth to biometric only

accumulator:
android: limit biometric auth to biometric only, disallow standard device unlock (pin, pattern, device pw)

### DIFF
--- a/electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
+++ b/electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
@@ -55,7 +55,7 @@ public class BiometricActivity extends Activity {
         Executor executor = getMainExecutor();
         BiometricPrompt biometricPrompt = new BiometricPrompt.Builder(this)
                 .setTitle("Electrum Wallet")
-                .setAllowedAuthenticators(BiometricManager.Authenticators.BIOMETRIC_STRONG | BiometricManager.Authenticators.DEVICE_CREDENTIAL)
+                .setAllowedAuthenticators(BiometricManager.Authenticators.BIOMETRIC_STRONG)
                 .setSubtitle(authMessage)
                 .build();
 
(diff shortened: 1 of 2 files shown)


f321x:
Why do you want to remove this?
This was intentionally enabled so users without fingerprint sensor (or users unwilling to enable it) don't have to enter the long Electrum wallet password to authorize payments and instead can enter the shorter system pin/pattern. Also allowing us to remove the pin input from QML.
See the discussion here: https://github.com/spesmilo/electrum/pull/10340#discussion_r2690146079

electrum/gui/qml/java_classes/org/electrum/biometry/BiometricActivity.java
```diff
@@ -55,7 +55,7 @@ private void handleIntent() {
         Executor executor = getMainExecutor();
         BiometricPrompt biometricPrompt = new BiometricPrompt.Builder(this)
                 .setTitle("Electrum Wallet")
-                .setAllowedAuthenticators(BiometricManager.Authenticators.BIOMETRIC_STRONG | BiometricManager.Authenticators.DEVICE_CREDENTIAL)
+                .setAllowedAuthenticators(BiometricManager.Authenticators.BIOMETRIC_STRONG)
```
SomberNight, line 58: Allowing `DEVICE_CREDENTIAL` was a deliberate design decision. The thinking was:
- some phones might not have a fingerprint reader (or other biometric sensor)
- even phones that do have it, the user might intentionally not set up the biometric sensor (not trust the device with its fingerprint/retina/etc)
- even so, we should probably allow such users to configure a very strong wallet password, but not have them re-type it every time they open their wallet file. Instead they can rely on e.g. an OS lockscreen PIN code which is hardware-SE-rate-limited
- also if the user enables auth on making payments, but they don't have/use the fingerprint sensor, and they have a strong wallet password, it is extremely annoying to have to type that at a POS terminal when buying beer/pizza
  - and to remove code complexity, we deleted the fallback software config PIN UI and concept
  - with current master, they can just use their OS lockscreen PIN for that too

accumulator:
ok ok :)

maybe this should be configurable. biometric-only is an improvement for devices that can, and for users that want.

PR #10818, opened 2026-08-07
qml: trustedcoin: add type hints, reduce excessive logging

accumulator:


electrum/plugins/trustedcoin/qml.py
```diff
@@ -14,16 +14,20 @@
...
 class Plugin(TrustedCoinPlugin):
     def __init__(self, *args):
         super().__init__(*args)
-        self._app = None
-        self.so = None
-        self.on_success = None
-        self.on_failure = None
-        self.tx = None
+        self._app = None  # type: ElectrumQmlApplication
+        self.wallet = None  # type: Wallet_2fa
```
SomberNight, line 26: It is pre-existing before this PR, but note that it is bad design to have the Plugin object store a reference to a Wallet object as the Plugin objects are singletons (for each type of plugin class, we only instantiate a single object) but there can be multiple wallets open. Other plugins don't do this.

SomberNight: (My point is that the plugin should not assume it is only being used by a single wallet object. Multiple wallet objects share a plugin.)

accumulator: agreed, I'll have a look if this is an easy fix, otherwise I will make an issue.

accumulator: see #10830

PR #10819, opened 2026-08-07
if we are accepting channels, check for channel_id collisions

ecdsa:


electrum/lnworker.py
```diff
@@ -1669,6 +1669,7 @@ def add_new_channel(self, temp_chan: Channel) -> Channel:
         assert type(temp_chan.storage) is dict
         channel_id = temp_chan.channel_id.hex()
         channels_db = self.db.get_dict('channels')
+        assert channel_id not in channels_db
```
SomberNight, line 1672: maybe also add
```py
assert bfh(channel_id) not in self._channel_backups
```

ecdsa: this was also pointed out by Claude. I added the check in `lnpeer.py` rather than here

SomberNight (approved):
ACK 0593c2bf22555afedba6b10712d89b639e941893
Looks okay to me.

PR #10820, opened 2026-08-07
lnutil: don't trim htlcs equal to dust level

f321x:
We should include htlc outputs in the ctx if their values is equal to the dust limit

https://github.com/lightning/bolts/blob/35e79db504560b9d3494a0ed07bf1e8379c3663a/03-transactions.md?plain=1#L340-L342

### DIFF
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -1146,18 +1146,18 @@ def make_commitment_outputs(
     has_anchors: bool,
     local_anchor_script: Optional[str],
     remote_anchor_script: Optional[str]
-) -> Tuple[List[PartialTxOutput], List[PartialTxOutput]]:
+) -> List[PartialTxOutput]:
 
-    # determine HTLC outputs and trim below dust to know if anchors need to be included
+    # convert htlcs to tx outputs
     htlc_outputs = []
     for script, htlc in htlcs:
         addr = bitcoin.redeem_script_to_address('p2wsh', script)
-        if htlc.amount_msat // 1000 > dust_limit_sat:
-            htlc_outputs.append(
-                PartialTxOutput(
-                    scriptpubkey=address_to_script(addr),
-                    value=htlc.amount_msat // 1000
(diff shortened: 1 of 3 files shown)


electrum/lnutil.py
```diff
@@ -1147,7 +1147,7 @@ def make_commitment_outputs(
     htlc_outputs = []
     for script, htlc in htlcs:
         addr = bitcoin.redeem_script_to_address('p2wsh', script)
-        if htlc.amount_msat // 1000 > dust_limit_sat:
+        if htlc.amount_msat // 1000 >= dust_limit_sat:
```
ecdsa: besides the strict inequality, do we also have a bug for non-anchor channels, where we are not deducting the htlc fee?

f321x: No, it seems correct. Actually this check here is redundant and I now replaced it by an assert.
`Channel.make_commitment()` already trims the htlcs through `Channel.included_htlcs()`, also considering the feerate:
https://github.com/spesmilo/electrum/blob/537653ad6f59b1bf727f56c34c13f59b34852e07/electrum/lnchannel.py#L1820

https://github.com/spesmilo/electrum/blob/537653ad6f59b1bf727f56c34c13f59b34852e07/electrum/lnchannel.py#L1633-L1634

https://github.com/spesmilo/electrum/blob/537653ad6f59b1bf727f56c34c13f59b34852e07/electrum/lnutil.py#L1196-L1212

So this check was only misleading and caused the conditions to drift apart.

ecdsa: great. I think we should also remove the filtering at the bottom of `make_commitment_outputs`

ecdsa: looks good now

f321x: > remove the filtering

Done in https://github.com/spesmilo/electrum/pull/10820/commits/3bcb39e3686dc9351b3686667850f89b619dbda9

PR #10821, opened 2026-08-07
interface: hardening against resource exhaustion (take2)

SomberNight:
see individual commits.

This is a simpler alternative to https://github.com/spesmilo/electrum/pull/10810.

### DIFF
--- a/electrum/interface.py
+++ b/electrum/interface.py
@@ -126,9 +126,13 @@ def assert_hex_str(val: Any) -> None:
         raise RequestCorrupted(f'{val!r} should be a hex str')
 
 
-def assert_dict_contains_field(d: Any, *, field_name: str) -> Any:
+def assert_dict(d: Any) -> None:
     if not isinstance(d, dict):
         raise RequestCorrupted(f'{d!r} should be a dict')
+
+
+def assert_dict_contains_field(d: Any, *, field_name: str) -> Any:
+    assert_dict(d)
     if field_name not in d:
         raise RequestCorrupted(f'required field {field_name!r} missing from dict')
     return d[field_name]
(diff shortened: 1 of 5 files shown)


electrum/interface.py
```diff
@@ -1095,10 +1099,17 @@ async def run_fetch_blocks(self):
         await self.session.subscribe('blockchain.headers.subscribe', [], header_queue)
```
ecdsa, line 1099: Why only subscriptions, and not responses to requests? I am asking this because I had a look at `server.peers.subscribe`, having forgotten that it is not an actual subscription.

Nevertheless, it seems that `parse_servers` in `network.py` could accept a dict instead of a list, and iterate over the dict keys without raising an exception (so, the extra data stuffing the response would be the dict values).

ecdsa: Ah, actually this only affects `network.get_peers`, not `network.get_server_peers`, which is what we use. 
The latter calls `random.shuffle`, which raises if passed a dict.

fun fact: `random.shuffle` does not raise if passed a dict of length one

electrum/interface.py
```diff
@@ -271,6 +281,40 @@ async def close(self, *, force_after: int = None):
...
+        # - Secondary interfaces should generate minimal traffic:
+        #   as they are only used for polled fee estimates (minimal data) and header subs (minimal data).
+        #   Note a header notification can force us into fork resolution and downloading lots
+        #   of headers. The initial headers download might also happen on any interface.
+        #   Headers download uses non-trivial amounts of data, e.g. 100k x 160 hex headers take 16 MB.
+        assert hasattr(NotificationSession, "cost_hard_limit")  # in base class
+        self.bw_cost_per_byte = 1 / 1_000
+        self.cost_hard_limit = 30_000  # 30 MB of bandwidth, in+out
+        self.cost_soft_limit = self.cost_hard_limit - 1  # this effectively disables the soft limit
+        self.cost_decay_per_sec = self.cost_hard_limit / 600  # refund over 10 minutes
```
ecdsa, line 303: `self.cost_decay_per_sec` is not reset in `remove_resource_limits`. 
Maybe its previous value should be saved and then restored?

ecdsa: well, I guess setting soft and hard limits to zero in remove_resource_limits makes other values irrelevant

electrum/interface.py
SomberNight: added some checks for `server.peers.subscribe` in https://github.com/spesmilo/electrum/pull/10834

PR #10823, opened 2026-08-08
channel_db: skip non-ASCII DNS hostnames in node announcements

ekzyis:
BOLT-07 mentions [here](https://github.com/lightning/bolts/blob/35e79db504560b9d3494a0ed07bf1e8379c3663a/07-routing-gossip.md?plain=1#L324-L327) that senders of `node_announcement` need to encode DNS hostname bytes with ASCII:

```
   * `5`: DNS hostname; data = `[1:hostname_len][hostname_len:hostname][2:port]` (length up to 258)
       * `hostname` bytes MUST be ASCII characters.
       * Non-ASCII characters MUST be encoded using Punycode:
         https://en.wikipedia.org/wiki/Punycode
```

(Punycode maps non-ASCII characters to ASCII characters.)

However, receivers need to check that the bytes are indeed decodable as ASCII. Currently, the code to parse address descriptors would throw `UnicodeDecodeError` on the first DNS hostname with non-ASCII characters and drop the remaining descriptors.[^1] With this PR, such invalid address descriptors are skipped. 

[^1]: I haven't tested this against running code, only using the existing tests, so I don't know if Electrum would crash during runtime or not.

FYI, LND [additionally checks](https://github.com/lightningnetwork/lnd/blob/d932f48678228e337a7cc75ee4e864245e6e2b99/lnwire/dns_addr.go#L56-L92) that the characters are within `[a-zA-Z0-9_.]` but this is not part of the specification.

### DIFF
--- a/electrum/channel_db.py
+++ b/electrum/channel_db.py
@@ -281,8 +281,12 @@ class NodeInfo(NamedTuple):
                 addresses.append((host, port))
             elif atype == 5:  # dns hostname
                 len_hostname = int.from_bytes(read(1), 'big')
-                host = read(len_hostname).decode('ascii')
+                host_bytes = read(len_hostname)
                 port = int.from_bytes(read(2), 'big')
+                try:
+                    host = host_bytes.decode('ascii')
+                except UnicodeDecodeError:
+                    continue
                 if not NodeInfo.invalid_announcement_hostname(host) and port > 0:
                     addresses.append((host, port))
             else:
--- a/tests/test_lnmsg.py
+++ b/tests/test_lnmsg.py
@@ -380,6 +380,7 @@ class TestLNMsg(ElectrumTestCase):
         invalid_inputs_parsing = (
             b'', # empty input
             b'\x06\x00', # address type 6 (\x06) is not specified
+            b'\x05\x02\xff\xff\x26\x07', # non-ASCII dns hostname
         )
         invalid_inputs_serialization = (
             ("::1", 9735),  # local ipv6


electrum/channel_db.py
```diff
@@ -281,8 +281,12 @@ def read(n):
                 addresses.append((host, port))
             elif atype == 5:  # dns hostname
                 len_hostname = int.from_bytes(read(1), 'big')
-                host = read(len_hostname).decode('ascii')
+                host_bytes = read(len_hostname)
                 port = int.from_bytes(read(2), 'big')
+                try:
+                    host = host_bytes.decode('ascii')
+                except UnicodeDecodeError:
+                    continue
```
SomberNight, line 289: > receivers need to check that the bytes are indeed decodable as ASCII

are we allowed to forward the node announcement message if it fails this check?

ekzyis: I have only found [this](https://github.com/lightning/bolts/blob/35e79db504560b9d3494a0ed07bf1e8379c3663a/07-routing-gossip.md?plain=1#L393-L395 ) in the spec regarding not forwarding node announcements:

```
  - if more than one `type 5` address is announced:
    - SHOULD ignore the additional data.
    - MUST not forward the `node_announcement`.
```

Fwiw, LND does not forward `node_announcement`s if they contain an invalid DNS hostname, see [here](https://github.com/lightningnetwork/lnd/blob/d932f48678228e337a7cc75ee4e864245e6e2b99/discovery/chan_series.go#L314-L319).

Before, we also did not forward it. With this PR, we do, like the other test cases that fail to parse. Maybe it's better to actually throw like LND?

accumulator:
To be totally conformant, we should 'punydecode' the hostname, but since at least LND does not seem to do that, it's likely node operators avoid unicode domain names anyway.

ekzyis:
> To be totally conformant, we should 'punydecode' the hostname

I only see the spec mentioning that the sender must encode DNS hostname bytes as ASCII, and Punycode for non-ASCII characters. I don't see where it mentions the receiver needs to decode Punycode back into Unicode. For connection purposes, the Punycode ASCII characters should be enough. Did you mean for display purposes?

> since at least LND does not seem to do that, it's likely node operators avoid Unicode domain names anyway.

Yes, I agree, I think Unicode domains are rare.

accumulator:
> I don't see where it mentions the receiver needs to decode Punycode back into Unicode

Indeed, it is actually what DNS resolvers do when resolving a unicode domain name; convert to punycode :)
So no need to decode before presenting to the resolver.

PR #10827, opened 2026-08-10
swaps: introduce unittest suite for submarine swaps

f321x:
Introduce new unittest suite for submarine swaps. Also do some minor refactoring for increased reliability.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -76,9 +76,11 @@ SWAP_TX_SIZE = 150  # default tx size, used for mining fee estimation
 
 MIN_SWAP_AMOUNT_SAT = 20_000
 MIN_LOCKTIME_DELTA = 60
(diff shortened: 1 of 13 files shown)


electrum/submarine_swaps.py
```diff
@@ -723,6 +725,16 @@ def _sanity_check_swap_costs(
             if costs_ratio > 0.15:
                 raise exc
 
+    def _sanity_check_prepayment(self, *, prepayment_sat: int, lightning_amount_sat: int):
+        max_prepayment_sat = min(
```
ecdsa, line 733: why min and not max here?

f321x: Both conditions should be enforced, so the lower one (`min`) covers both

ecdsa: ok :-)

ecdsa:
looks good to me

PR #10828, opened 2026-08-10
LNURL validation improvements

accumulator:


electrum/lnurl.py
```diff
@@ -169,12 +176,23 @@ def _parse_lnurl6_response(lnurl_response: dict) -> LNURL6Data:
...
     )
     return data
 
 
+def validate_lnurl6_invoice(lnurl_data: LNURL6Data, amount_sat: Decimal, invoice: 'Invoice'):
+    # invoice amount should match exactly what we requested
+    if invoice.get_amount_sat() != amount_sat:
+        raise Exception("lnurl returned invoice with wrong amount")
+    # the invoice must commit to the metadata we showed the user, so that the
+    # payee's signature covers the description we are about to pay for.
```
f321x: This doesn't seem to be part of the spec requirements anymore:
https://github.com/lnurl/luds/pull/234

electrum/lnurl.py
```diff
@@ -273,10 +273,18 @@ async def callback_lnurl(url: str, params: dict) -> dict:
     return response
 
 
+LN_ADDRESS_RE = re.compile(
+    r"[a-zA-Z0-9\-_][a-zA-Z0-9\-_.]*"          # username
+    r"@"
+    r"[a-zA-Z0-9\-]+(\.[a-zA-Z0-9\-]+)*"       # domain labels
+    r"\.[a-zA-Z]{2,}"                          # tld (alphabetic, so no bare IPv4)
+)
```
f321x, line 263: This would break supported addresses with `+` like: alice+donations@example.com
https://github.com/lnurl/luds/blob/913c49e5b65081473eed7889eee5f806dd30f049/16.md?plain=1#L10

electrum/lnurl.py
```diff
@@ -273,10 +273,18 @@ async def callback_lnurl(url: str, params: dict) -> dict:
...
+
+
 def lightning_address_to_url(address: str) -> Optional[str]:
     """Converts an email-type lightning address to a decoded lnurl.
     see https://github.com/fiatjaf/lnurl-rfc/blob/luds/16.md
     """
-    if re.match(r"^[^@]+@[^.@]+(\.[^.@]+)+$", address):
+    if LN_ADDRESS_RE.fullmatch(address):
         username, domain = address.split("@")
         return f"https://{domain}/.well-known/lnurlp/{username}"
```
f321x: Preexisting but i guess for `.onion` domains we should prepend `http://` instead of `https://`:
https://github.com/lnurl/luds/blob/913c49e5b65081473eed7889eee5f806dd30f049/16.md?plain=1#L12

electrum/lnurl.py
accumulator: thx

electrum/lnurl.py
accumulator: I don't think you can host email on an onion domain. You can have emaillike notation on an onion domain of course, but is that a usecase we want to support?

accumulator: well, a bit strange, but easy to implement, so why not..

SomberNight (approved):

SomberNight:
Looks good to me. Thanks.

PR #10830, opened 2026-08-10
qml: wallet, tx and callbacks were incorrectly scoped to class level for trustedcoin plugin.

accumulator:
Pass these instances in the callback instead of retrieving it from the plugin.


SomberNight:
I added a trivial follow-up to the branch (38a2e1ab49ccda46e0eb5ce85d76acbdd69ffb31).

SomberNight (approved):

SomberNight:
Thanks.

PR #10834, opened 2026-08-10
interface: add method for "server.peers.subscribe"

SomberNight:
Instead of having callers manually construct and call the electrum protocol RPC, add convenience method in Interface class. Minimal validation of response shape is also done here, with some further deeper checks left in network.parse_servers().

ref https://github.com/spesmilo/electrum-protocol/blob/b3db795b3efd2a0e45a5b3f67fb360b65d4cd4d9/docs/protocol-methods.rst?plain=1#L1503

### DIFF
--- a/electrum/network.py
+++ b/electrum/network.py
@@ -86,12 +86,12 @@ class ConnectionState(IntEnum):
     CONNECTED     = 2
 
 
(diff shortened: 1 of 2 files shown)


electrum/network.py
f321x, line 513: nit: `session` is now unused

PR #10836, opened 2026-08-10
reverse swaps: lower acceptable cltv_delta value

ecdsa:


SomberNight (approved):
f58d37e1c2cb2dacccf6d828856738166b5bc467 looks good to me, if CI passes
EDIT: ah but it does not! concept/approach ACK anyway.

PR #10837, opened 2026-08-10
config: log warnings if using experimental features on mainnet

SomberNight:
Users who discover experimental settings in the first place should already be kind of aware of the risks anyway. However we now log a warning to raise more awareness.

Alternatively we could raise an exception to trigger a runtime exit. This is a bit less extreme and should already be seen even without verbose logging enabled.

In the case of LN forwarding, people should really not use it on mainnet yet.


krynerovamonika827-commits (commented):
[e](url)

PR #10839, opened 2026-08-10
android: refactor deprecated android:allowBackup

accumulator:
android: update p4a to ref b1b528cd8b681c0b5185ab82a319527037cc2421:
    
- qt6: refactor deprecated `android:allowBackup` and `android:fullBackupContent` to Android12+ `android:dataExtractionRules` + xml definition.
- qt6: retain `android:allowBackup` for android 8-11
    
Disallow device-transfer (nearby share, USB)
    
(old `allowBackup="false"` behaviour was only considering cloud backup)

### DIFF
--- a/contrib/android/Dockerfile
+++ b/contrib/android/Dockerfile
@@ -235,7 +235,7 @@ RUN cd /opt \
     && /opt/venv/bin/python3 -m pip install --no-build-isolation --no-dependencies -e .
 
 # install python-for-android
-ENV P4A_CHECKOUT_COMMIT="315bb8adf8022972f0ee87a45896d05c38134e1c"
+ENV P4A_CHECKOUT_COMMIT="b1b528cd8b681c0b5185ab82a319527037cc2421"
 # ^ from branch electrum_202602 (note: careful with force-pushing! see #8162)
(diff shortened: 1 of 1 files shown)


accumulator:
crap, removal of android:extractNativeLibs="true" snuck in..
Should not be a problem though

SomberNight:
> Disallow device-transfer (nearby share, USB)

I don't know if we want to do that. e.g. when migrating to a new phone, it might be nice to move the Lightning state and keep using the channels. I wonder if the user can still launch the app on the old phone though :/

SomberNight:
but I think it's a good idea to opt-out of the cloud backups, which IIUC the current (deprecated?) setting does.
That is, I think device-to-device transfers we could allow, but cloud backups, maybe not.

If the currently used `android:allowBackup` pref is being deprecated, maybe we still need half of that new `backup_rules.xml` file you wrote in https://github.com/spesmilo/python-for-android/commit/29b481864aebf95ab0b203a47f6efdcb8ac598ba ?

accumulator:
For now the current status-quo is fine IMO, we should revisit after we have strengthened wallet KDF

PR #10840, opened 2026-08-10
cli/swaps: fast fail swap cli on changed fees

### LINKED ISSUE #10826: [Electrum Swap] Claim fee calculation problem
### Description

We have received report on whales.exchange that swaps do not work.

Upon investigation of logs, it seems the following is the culprit. Can someone please check if this is a bug or intended behavior?

In https://github.com/spesmilo/electrum/blob/master/electrum/commands.py#L2205-L2214 we have calculation of `claim_fee`. This is in `reverse_swap` method which is called on the client. Then `wallet.lnworker.swap_manager.reverse_swap` is called with calculated values.

This then sends message to the swap provider and receives information about the swap, including how much the client will receive on-chain.

Then there is check https://github.com/spesmilo/electrum/blob/master/electrum/submarine_swaps.py#L1212-L1214

that compares the received on-chain amount (what client receives) with what client expected. If the received value is smaller, call fails.

On the server side, the server calculates using its fees, obviously. On the client side, however, it looks like the `claim_fee` is also calculated locally. 

So imagine that the client has fee 200 sats, but provider has fee 180 sats. 
Then `onchain_amount_sat = satoshis(onchain_amount) + claim_fee` will be `onchain_amount + 200` on the client but the provider will send `onchain_amount + 180` and the call fails.

So it seems that if client view of fees are greater than provider fees then the reverse swap call fails on the check. Is that intended? Shouldn't the calculation be using the provider's advertised `mining_fee` amount?

f321x:
CLI swaps happen in two steps. 
First the command is ran as "dryrun" to calculate the expected amounts, then the swap is initiated with these amounts. 

If the mining fee estimates increase during the time between those two calls, the swap provider might return an unexpected amount for us in their response, causing the response checks in `SwapManager` to fail a swap after a the network round trip with the swap provider. 

Instead we should fail early before the swap is registered with the provider, to avoid stale swaps in the servers db and give the CLI user a clear exception.
Note: the issue can still happen if the fees change in the brief moment during the request, so scripts consuming the CLI should handle a failure either way.

Also fixes a bug in that the `onchain_amount` returned by the swap branch of the `reverse_swap` command was inconsistent with the value returned by the `dryrun` branch (it included the `claim_fee`).

Closes https://github.com/spesmilo/electrum/issues/10826

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -56,7 +56,6 @@ from .util import (
 from . import bitcoin
 from .bitcoin import is_address,  hash_160, COIN
 from .bip32 import BIP32Node
-from .i18n import _
 from .transaction import (
     Transaction, multisig_script, PartialTransaction, PartialTxOutput, tx_from_any, PartialTxInput, TxOutpoint,
     convert_raw_tx_to_hex
@@ -85,6 +84,10 @@ if TYPE_CHECKING:
     from .network import Network
     from .daemon import Daemon
     from electrum.lnworker import PaymentInfo
+
+
+def _(_):  # break translation
+    raise Exception("The CLI is intentionally always non-localized")
 
 
 known_commands = {}  # type: Dict[str, Command]
@@ -2451,7 +2454,7 @@ def subparser_call(self, parser, namespace, values, option_string=None):
         parser = self._name_parser_map[parser_name]
     except KeyError:
         tup = parser_name, ', '.join(self._name_parser_map)
-        msg = _('unknown parser {!r} (choices: {})').format(*tup)
+        msg = 'unknown parser {!r} (choices: {})'.format(*tup)
         raise ArgumentError(self, msg)
     # parse all the remaining options into the namespace
     # store any unrecognized options on the object, so that the top
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -2132,7 +2132,8 @@ class Commands(Logger):
     @command('wnpl')
     async def normal_swap(self, onchain_amount, lightning_amount, password=None, wallet: Abstract_Wallet = None):
         """
-        Normal submarine swap: send on-chain BTC, receive on Lightning
+        Normal submarine swap: send on-chain BTC, receive on Lightning.
+        Note: fees can change between the dryrun and the following swap request, causing the request to error and require a new dryrun.
 
         arg:decimal_or_dryrun:lightning_amount:Amount to be received, in BTC. Set it to 'dryrun' to receive a value
         arg:decimal_or_dryrun:onchain_amount:Amount to be sent, in BTC. Set it to 'dryrun' to receive a value
@@ -2156,6 +2157,13 @@ class Commands(Logger):
             else:
                 lightning_amount_sat = satoshis(lightning_amount)
                 onchain_amount_sat = satoshis(onchain_amount)
+                required_onchain_amount_sat = sm.get_send_amount(lightning_amount_sat, is_reverse=False)
+                # same 1 sat rounding tolerance as in `request_normal_swap()`
+                if not required_onchain_amount_sat \
+                        or not (onchain_amount_sat - 1 <= required_onchain_amount_sat <= onchain_amount_sat):
+                    raise UserFacingException(
+                        "Swap fees have changed since the dryrun was calculated. Do a new dryrun first."
+                        + f" ({required_onchain_amount_sat} != {onchain_amount_sat} sat)")
                 txid = await wallet.lnworker.swap_manager.normal_swap(
                     transport=transport,
                     lightning_amount_sat=lightning_amount_sat,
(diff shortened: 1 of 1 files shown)


electrum/commands.py
```diff
@@ -2156,6 +2157,13 @@ async def normal_swap(self, onchain_amount, lightning_amount, password=None, wal
             else:
                 lightning_amount_sat = satoshis(lightning_amount)
                 onchain_amount_sat = satoshis(onchain_amount)
+                required_onchain_amount_sat = sm.get_send_amount(lightning_amount_sat, is_reverse=False)
+                # same 1 sat rounding tolerance as in `request_normal_swap()`
+                if not required_onchain_amount_sat \
+                        or not onchain_amount_sat - 1 <= required_onchain_amount_sat <= onchain_amount_sat:
+                    raise UserFacingException(
+                        _("Swap fees have changed since the dryrun was calculated. Do a new dryrun first.")
+                        + f" ({required_onchain_amount_sat} != {onchain_amount_sat} sat)")
```
SomberNight, line 2169: https://github.com/spesmilo/electrum/blob/a357e86f68fc8f8665b39d9ff9d2ea84b523d613/run_electrum#L433

(though we maybe should re-state that at the top of commands.py)

electrum/commands.py
```diff
@@ -2156,6 +2157,13 @@ async def normal_swap(self, onchain_amount, lightning_amount, password=None, wal
             else:
                 lightning_amount_sat = satoshis(lightning_amount)
                 onchain_amount_sat = satoshis(onchain_amount)
+                required_onchain_amount_sat = sm.get_send_amount(lightning_amount_sat, is_reverse=False)
+                # same 1 sat rounding tolerance as in `request_normal_swap()`
+                if not required_onchain_amount_sat \
+                        or not onchain_amount_sat - 1 <= required_onchain_amount_sat <= onchain_amount_sat:
```
SomberNight: Can we use parens here? There is so much operator precedence stuff going on in this `if`, it's hard to imagine the AST in my head.

```suggestion
                        or not (onchain_amount_sat - 1 <= required_onchain_amount_sat <= onchain_amount_sat):
```


it's a small nit, but I also mean in general in the codebase :)
Too many layers of parens makes it hard to read but too few makes it easy to make a mistake in operator precedence. Delicate tradeoff :P

electrum/commands.py
```diff
@@ -2190,32 +2199,38 @@ async def reverse_swap(
...
+                requested_recv_amount_sat = satoshis(onchain_amount)
                 claim_fee = sm.get_fee_for_txbatcher()
-                onchain_amount_sat = satoshis(onchain_amount) + claim_fee
-                assert prepayment != "dryrun", "Provide the 'prepayment' obtained from the dryrun."
+                funding_utxo_value_sat = requested_recv_amount_sat + claim_fee
+                onchain_recv_amount_sat = sm.get_recv_amount(lightning_amount_sat, is_reverse=True)
+                if not onchain_recv_amount_sat or onchain_recv_amount_sat < requested_recv_amount_sat:
+                    raise UserFacingException(
+                        _("Swap fees have changed since the dryrun was calculated. Do a new dryrun first.")
+                        + f" ({onchain_recv_amount_sat} < {requested_recv_amount_sat} sat)")
```
SomberNight, line 2225: Do I understand correctly that this is the "same" check as in `reverse_swap()`?
https://github.com/spesmilo/electrum/blob/a357e86f68fc8f8665b39d9ff9d2ea84b523d613/electrum/submarine_swaps.py#L1296-L1299

Except there in sswaps.py, both amounts being compared are offset by `+claim_fee` ?
Uff took me like 10 minutes to untangle that >.<

SomberNight (commented):
52a961f2f2c11dfa27d02bcdf5cc2276dabafdcf is mostly ok. some nits.

electrum/commands.py
f321x: Removed.
I added a commit https://github.com/spesmilo/electrum/pull/10840/commits/43680185789f3212d9910e3b2e7575d8ae058a72 which breaks the `_()` function call in `commands.py` to enforce this programmatically.

electrum/commands.py
f321x: Changed in https://github.com/spesmilo/electrum/pull/10840/commits/89b10fe256ff81d6bcb6a8dd584cbe150cd8e47c. I see your point.

electrum/commands.py
f321x: Yes it's a similar check, it just re-calculates the amount with the latest known fees client side to reduce the time window for disagreements and potentially prevent a useless request to the swap provider as the other check only happens after we got the servers response.

electrum/commands.py
```diff
@@ -87,6 +86,10 @@
     from electrum.lnworker import PaymentInfo
 
 
+def _(_):  # break translation
+    raise ImportError("The CLI is intentionally always non-localized")
```
SomberNight: Why is it an `ImportError` instead of some other exception type (such as generic `Exception`)?

electrum/commands.py
SomberNight: Did you forget to push that?
No relevant diff in https://github.com/spesmilo/electrum/compare/52a961f2f2c11dfa27d02bcdf5cc2276dabafdcf..43680185789f3212d9910e3b2e7575d8ae058a72

f321x: Indeed this got lost somehow. Pushed now in https://github.com/spesmilo/electrum/compare/43680185789f3212d9910e3b2e7575d8ae058a72..e4b6d3a44b33a1bf060b95f474cfc0f01b973c7f

electrum/commands.py
f321x: I thought it is somewhat fitting as it is related to the `_` import, but yeah might as well just raise a generic Exception. Changed in https://github.com/spesmilo/electrum/compare/43680185789f3212d9910e3b2e7575d8ae058a72..e4b6d3a44b33a1bf060b95f474cfc0f01b973c7f

PR #10843, opened 2026-08-12
 test: test_we_disconnect_on_incoming_notification_spam fails on Macos

Alwoch:
This test for me fails on macOS 26 (arm64, python 31.4). `await asyncio.sleep(0)` yields a single event loop iteration which isnt enough here for the client to disconnect and the server to observe it. With `sleep(0.5)`, the test passes in full including `recv_count/cost` assertions and the log check so i figured the resource limiting itself works but the check could be too tight? I didnt dig into why one iteration suffices on the CI runners but not here.

Screenshot 2026-08-12 at 09 40 58

I did change it to wait on the event with a timeout to match what the next two lines already do the for the Client side and it still passes. All CI runners are ubuntu so i dont think this would've showed up there

Screenshot 2026-08-12 at 09 49 42

### DIFF
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -187,8 +187,8 @@ class TestInterface(ElectrumTestCase):
                     break
                 await srv_sess.send_notification('blockchain.headers.subscribe', headersub_res)
             # both parties should close their end of the session (triggered by the client DC-ing)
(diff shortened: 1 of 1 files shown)


f321x:
Thank you! Also noticed this flakiness on my linux machine.

PR #10846, opened 2026-08-12
Android gradle9 pin barcode scanner and its deps

accumulator:
This PR upgrades gradle used in P4A to 9.4.1 (AGP 9.1.0) and patches the barcode scanner components
to pin to the p4a provided toolchain, upgrades xzing-cpp to 3.1.0.0, CameraView to 1.10.2


contrib/android/Dockerfile
```diff
@@ -235,8 +235,8 @@ RUN cd /opt \
     && /opt/venv/bin/python3 -m pip install --no-build-isolation --no-dependencies -e .
 
 # install python-for-android
-ENV P4A_CHECKOUT_COMMIT="315bb8adf8022972f0ee87a45896d05c38134e1c"
-# ^ from branch electrum_202602 (note: careful with force-pushing! see #8162)
+ENV P4A_CHECKOUT_COMMIT="167c3e42db2bb3cfa28f96126cdb82d3de5189de"
+# ^ from branch electrum_202602_gradle9 (note: careful with force-pushing! see #8162)
```
SomberNight: btw, the branch protection rules on spesmilo/p4a are set up for `electrum_*`. 
so `electrum_202602_gradle9` cannot be force-pushed.

You can use other branch names for non-prod development work and then you can force-push them. (or you can also use branches on your own fork)


just wanted to point this out :)

accumulator: yeah I noticed :/

contrib/android/make_barcode_scanner.sh
```diff
@@ -7,19 +7,46 @@
...
 BARCODE_SCANNER_VIEW_COMMIT_HASH="0bdb69269c252bb6daef2f871b76403c8b051945"  # 1.6.5
 BARCODE_SCANNER_VIEW_REPO="https://github.com/markusfisch/BarcodeScannerView.git"
 
-CAMERA_VIEW_COMMIT_HASH="745597d05bc6abfdb3637a09a8ecaf30fdce7b6e"  # 1.10.0
+CAMERA_VIEW_COMMIT_HASH="c806afadaf2ea81d454c9b81ef5938be1c9855cd"  # 1.10.2
 CAMERA_VIEW_REPO="https://github.com/markusfisch/CameraView.git"
+CAMERA_VIEW_VERSION="1.10.2"
 
-ZXING_CPP_COMMIT_HASH="79f5adc6250e90de0bd635eb9181c5f8a18affda"  # v2.3.0.4 using kotlin-stdlib 1.8.22
+ZXING_CPP_COMMIT_HASH="e88bb1d9e43502ad7073d29ee72cea4b758c1125"  # v3.1.0.0 using kotlin-stdlib 2.2.10
```
SomberNight, line 29: if `markusfisch/zxing-cpp` now uses gradle 9, can we then use use Java 21 (or 25) and undo the hacks in https://github.com/spesmilo/electrum/commit/4da7b7f55d85f11d84cfb5995bd6ad1d9d9cb311 (from https://github.com/spesmilo/electrum/pull/10452)?

contrib/android/buildozer_qml.spec
```diff
@@ -169,7 +169,7 @@ android.add_src = electrum/gui/qml/java_classes/
 # kotlin-stdlib is required for zxing-cpp (BarcodeScannerView)
 android.gradle_dependencies =
     androidx.core:core:1.16.0,
-    org.jetbrains.kotlin:kotlin-stdlib:1.8.22
+    org.jetbrains.kotlin:kotlin-stdlib:2.2.10
```
SomberNight, line 172: How come all the verification-metadata xml files are for the QR scanner?  I would have expected some java deps to be pulled in through python-for-android, for the main Electrum code.  For example, what about this `androidx.core`?  How come that's not pinned in an xml file? Does gradle not download that?

contrib/android/verification-metadata/barcodescannerview.xml
SomberNight, line 1: This is madness. Are all these transitive deps really pulled in *only* for the QR scanner? 500+ deps. Completely unrealistic to vet/review. Thanks a lot for looking into pinning this -- without that I would not have realised we are apparently pulling in the world...

If we tried to switch to your PoC QtMultimedia scanner, could we drop all these?
(Maybe since switching to Qt 6.10, it would no longer be crashing?)

contrib/android/make_barcode_scanner.sh
accumulator: probably, but will need to test.

contrib/android/buildozer_qml.spec
accumulator: > How come all the verification-metadata xml files are for the QR scanner? I would have expected some java deps to be pulled in through python-for-android, for the main Electrum code. For example, what about this `androidx.core`? How come that's not pinned in an xml file? Does gradle not download that?

The QR scanner was the scope for this PR, pinning the dependencies above probably needs additional changes in our P4A fork.

contrib/android/verification-metadata/barcodescannerview.xml
accumulator: > This is madness. Are all these transitive deps really pulled in _only_ for the QR scanner? 500+ deps. Completely unrealistic to vet/review. Thanks a lot for looking into pinning this -- without that I would not have realised we are apparently pulling in the world...

Agreed, it feels a bit.. fragile. I was in doubt whether to include the last commit in this PR for the same reasons, but on the other hand, what's the use of pinning only the top of the dep hierarchy..

> If we tried to switch to your PoC QtMultimedia scanner, could we drop all these? (Maybe since switching to Qt 6.10, it would no longer be crashing?)

Yes. However, the last time I dug into this, I saw there were some patches upstream around Qt6.6-6.8 claiming to fix some stability problems with it. No idea if it's stable now, at least it seemed to received some eyeballs.

Note that there were issues on Android8 w.r.t QtMultimedia (don't remember exactly what, likely a NPE/SEGV at init) and was disabled in https://github.com/spesmilo/python-for-android/commit/375a05de21b538d704174b1efeb3fc85d151f94e 

BTW, It wasn't really a PoC scanner, it was the implementation we used while on Qt5.

contrib/android/buildozer_qml.spec
ecdsa: How many more dependencies would that represent? (the ones that are not pinned yet)

PR #10850, opened 2026-08-14
swap_server: fix some issues running swap server

accumulator:
- `SwapServerPlugin.server` was never set, so the guard on 'use the first wallet loaded' was ineffective
- exclude non-lightning wallets for consideration
- disabling the plugin left the server running
- use `async with` .. in `run_nostr_server()`. non-async would block on `__exit__`, `fut.result(timeout=5)` would time-out since it runs on the `asyncio` thread.

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -291,6 +291,9 @@ class SwapManager(Logger):
                     self.lnworker.bundle_payments([payment_hash, swap.prepay_hash])
                 self.lnworker.register_hold_invoice(payment_hash, self.hold_invoice_callback)
         self.is_server = False # overridden by swapserver plugin if enabled
+        self.http_server = None  # set by the swapserver plugin if enabled
+        self._server_tasks = []  # type: List[asyncio.Task]
+        self._stopping_server = False
         self.is_initialized = asyncio.Event()
         self.pairs_updated = asyncio.Event()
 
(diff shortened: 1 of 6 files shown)


electrum/submarine_swaps.py
```diff
@@ -317,7 +323,7 @@ async def run_nostr_server(self):
             self.logger.info("This wallet is password-protected. Please unlock it to start the swapserver plugin")
             await asyncio.sleep(10)
 
-        with NostrTransport(self.config, self, self.lnworker.nostr_keypair) as transport:
+        async with NostrTransport(self.config, self, self.lnworker.nostr_keypair) as transport:
```
ecdsa, line 326: Note that we still use non-async with in the GUI.
The commit message says "non-async would block on __exit__", any idea why?

accumulator: Because `run_nostr_server` is itself a coroutine on the asyncio loop thread. The `__exit__` deadlocks on `fut.result`. The GUI call-sites run on the UI thread, so it won't deadlock there.

ecdsa: I see.

ecdsa:
Note: a big part of this PR fixes issues with `HttpSwapServer`. This class only exists so that we can use it in regtests; it is not supposed to be used on mainnet, so I doubt these issues were organically met. Were they raised by Claude?

Ideally, we should write a very simple local nostr relay, that would only relay our own stuff. This local relay could be used in regtests, so that these tests would use nostr instead of http, and we could remove `HttpSwapServer`.  A local nostr relay would also be useful for the NWC plugin.

Adding AI-generated unit tests for `HttpSwapServer`, which only serves to support other tests, feels, well... recursive :-( 

EDIT: actually, HttpSwapServer might be used by whales.exchange, I have not checked that.

accumulator:
> This class only exists so that we can use it in regtests

Where is this documented?

ecdsa:
> Where is this documented?

it is not, my bad

Issue #10785, opened 2026-07-31
[BUG] Funds from force-closed Lightning channel not swept after Electrum Mobile crash


MarcosBarceloPerez:
Hello,

I would like to report what appears to be a bug related to Lightning channel recovery and force-closed channels in Electrum Mobile.

Here is what happened:

1. I recently created a wallet in Sparrow Wallet and imported it into Electrum Mobile using the BIP84 extended private master key (xprv).

2. I deposited approximately €100 worth of Bitcoin into the wallet and decided to open a Lightning channel with the Electrum Trampoline service to test Lightning functionality.

3. I saved the Base64 channel backup provided by Electrum.

4. Unfortunately, I later lost the application data from Electrum Mobile. I restored my wallet by importing the extended master private key again and also restored the Lightning channel using the Base64 channel backup.

5. Since Electrum no longer had information about the latest state of the Lightning channel, I decided to request a remote force-close of the channel.

6. The force-close was successfully executed, and the channel funds were sent to the following Bitcoin address:

`bc1q89a5ynlqlyg07ps32d8sgrygtlcpm9ggsh5f2ex35qlmzrqwt4dqj6xl75`

The corresponding transaction is:

`b9017a5a826af5a2b9c09d545516ce23a78760a84d229cf0e324f116cc68b7c3`

7. As far as I understand, after the force-close and once the CSV timelock expired, Electrum should have detected the relevant output and created a sweep transaction to move the funds back into the wallet.

However, Electrum Mobile suddenly crashed. After restarting the application, the Lightning channel and its history had disappeared. The only Lightning transaction that had previously appeared in the history was the channel closure. No sweep transaction was ever created.

I then tried importing the Base64 channel backup again. The channel briefly appears in Electrum, but then disappears again. I assume this may be because the channel has already been closed. I have observed the same behavior when trying to restore the channel using Electrum Desktop.

At this point, I am unable to recover the funds from the force-closed channel. It appears that Electrum did not perform the expected sweep transaction, and I can no longer access the channel information through the user interface.

This seems to indicate a potential bug in the handling of force-closed Lightning channels after restoring a wallet/channel backup, particularly when the application loses the latest channel state or crashes during the recovery process.

Could you please investigate whether this is a known bug or an issue with Electrum's Lightning channel recovery logic?

[...]

Issue #7513, opened 2021-09-25
"Save backup" for an LN wallet results in a lightning-disabled wallet copy with imported channels backups hidden in Qt GUI


SomberNight:
Steps:
- have wallet with some existing lightning channels in Qt GUI open
- go to menubar>`File`>`Save backup`, complete dialog and create a backup wallet
- open backup wallet file
- notice that Channels tab is empty
    - ^ bug. there should be "imported channel backups" there
    - note that the channel backups exist, they are just not shown (see e.g. `wallet.db.get("imported_channel_backups")`)
- notice that lightning is disabled
- go to menubar>`Wallet`>`Information`, enable Lightning
- the Channels tab is now populated with the imported channel backups

ecdsa:
This can be fixed by removing the line `new_db.put('lightning_xprv', None) in `save_backup
However, weird things arise from the fact that we create imported backups from channels that have an onchain backup.
For example, I can delete the imported backup, and no "onchain backup" is shown.

ecdsa:
the above issue was fixed in af1fb1ae9f5776ca8f197e2d14bd2bdca7f0afe5
in addition, backups of redeemed channels are auto-removed: 90610de58ec99a83a419d14ac27f445312a9672b

PR #10851, opened 2026-08-14
wallet: don't remove ln xprv from wallet backup

f321x:
Fixes the same issue described in https://github.com/spesmilo/electrum/issues/7513 for non-deterministic lightning wallets.

### DIFF
--- a/electrum/wallet.py
+++ b/electrum/wallet.py
@@ -513,7 +513,6 @@ class Abstract_Wallet(ABC, Logger, EventListener):
             for chan_id, chan in self.lnworker.channels.items():
                 channel_backups[chan_id.hex()] = self.lnworker.create_channel_backup(chan_id)
             new_db.put('channels', None)
-            new_db.put('lightning_privkey2', None)
         new_db.set_modified(True)
         new_db.write()
         return new_path


electrum/wallet.py
```diff
@@ -513,7 +513,6 @@ def save_backup(self, backup_dir):
             for chan_id, chan in self.lnworker.channels.items():
                 channel_backups[chan_id.hex()] = self.lnworker.create_channel_backup(chan_id)
             new_db.put('channels', None)
-            new_db.put('lightning_privkey2', None)
```
SomberNight, line 516: While this change is similar to https://github.com/spesmilo/electrum/issues/7513 from a GUI point of view, actually it is substantially different from a security of funds point of view.

This change basically fixes https://github.com/spesmilo/electrum/issues/10785 for the "file-backup" scenario (as opposed to individual exported chan backups). As `lightning_privkey2` is randomly generated, deleting it means we would generate new ones when re-enabling lightning in the .backup file, and then v2 channel backups would not be sufficient to spend the funds from anchors channels. With this change, `lightning_privkey2` is not deleted, and hence v2 channel backups *are* sufficient to spend anchor channels.

So the change is good and it is more than a simple GUI refresh bugfix.

SomberNight (approved):

PR #10852, opened 2026-08-14
lightning: fix anchor channel backup

### LINKED ISSUE #10785: [BUG] Funds from force-closed Lightning channel not swept after Electrum Mobile crash

f321x:
Add missing information to lightning channel backup so users of non-deterministic
lightning wallets trying to recover anchor channels are able to sweep the to_remote
output of a remote ctx. 

Also shows a warning to affected users, urging them to export a new backup.

Fixes https://github.com/spesmilo/electrum/issues/10785

### DIFF
--- a/electrum/gui/qml/components/main.qml
+++ b/electrum/gui/qml/components/main.qml
@@ -698,6 +698,30 @@ ApplicationWindow
         }
     }
 
+    function showStartupWarnings() {
+        if (!Daemon.currentWallet)
+            return
+        let warnings = Daemon.currentWallet.startupWarnings
+        // show the warnings one after another, as the dialogs are not modal
+        function showWarning(i) {
+            if (i >= warnings.length)
+                return
+            let dialog = app.messageDialog.createObject(app, {
+                title: warnings[i].title,
+                iconSource: Qt.resolvedUrl('../../icons/warning.png'),
+                text: warnings[i].message
+            })
+            dialog.accepted.connect(function() {
+                Daemon.currentWallet.acknowledgeWarning(warnings[i].key)
+            })
+            dialog.closed.connect(function() {
+                showWarning(i + 1)
+            })
+            dialog.open()
+        }
+        showWarning(0)
+    }
+
     Connections {
         target: Daemon
         function onWalletRequiresPassword(name, path) {
@@ -736,6 +760,7 @@ ApplicationWindow
         }
         function onWalletLoaded() {
             app._loadingWalletContext = null  // either biometric auth or manual auth was successful
+            showStartupWarnings()
         }
     }
 
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -613,7 +613,7 @@ class ChannelBackup(AbstractChannel):
             to_self_delay=cb.local_delay,
             channel_type=cb.channel_type,
             # there are three cases of backups:
-            # 1. legacy: payment_basepoint will be derived
+            # 1. legacy: deprecated, it cannot be recovered
             # 2. static_remotekey: to_remote sweep not necessary due to wallet address
             # 3. anchor outputs: sweep to_remote with local_payment_basepoint if it is a private key, otherwise
             #           by deriving the key from the funding pubkeys
@@ -680,8 +680,8 @@ class ChannelBackup(AbstractChannel):
         return sweep_their_ctx_to_remote_backup(chan=self, ctx=ctx, funding_tx=funding_tx)
 
     def create_sweeptxs_for_our_ctx(self, ctx):
-        if self.is_imported:
-            return sweep_our_ctx(chan=self, ctx=ctx)
+        if self.is_imported and self.config[LOCAL].payment_basepoint.pubkey is not None:
+            return sweep_our_ctx(chan=self, ctx=ctx)  # v0 backups miss payment_basepoint (see #8536/1a46460)
         else:
             return {}
 
@@ -753,12 +753,12 @@ class ChannelBackup(AbstractChannel):
 
     def get_wallet_addresses_channel_might_want_reserved(self) -> Sequence[str]:
         if self.is_imported:
-            # For v1 imported cbs, we have the local_payment_pubkey, which is
+            # For v1+ imported cbs, we have the local_payment_basepoint, which is
             # directly used as p2wpkh() of static_remotekey channels.
-            # (for v0 imported cbs, the correct local_payment_pubkey is missing, and so
-            #  we might calculate a different address here, which might not be wallet.is_mine,
-            #  but that should be harmless)
+            # (for v0 imported cbs, it is missing, so we have no address to reserve)
             our_payment_pubkey = self.config[LOCAL].payment_basepoint.pubkey
+            if our_payment_pubkey is None:
+                return []
             to_remote_address = make_commitment_output_to_remote_address(our_payment_pubkey, has_anchors=self.has_anchors())
             return [to_remote_address]
         else:  # on-chain backup
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -672,6 +672,20 @@ class ChannelBackup(AbstractChannel):
             announcement_bitcoin_sig=b'',
         )
 
+    def can_sweep_their_ctx_to_remote(self) -> bool:
+        cb = self.cb
+        if not isinstance(cb, ImportedChannelBackupStorage):
+            return True  # on-chain backups only exist for deterministic wallets
+        v = cb.backup_version
+        if v >= 3:
+            return True  # v3+ backups contain the payment_basepoint secret needed for the to_remote sweep
+        elif v == 2:
+            # pre-v3: only sweepable if we still have the LN keys that created this backup
+            return (not self.has_anchors()) or cb.privkey == self.lnworker.node_keypair.privkey
+        # srk backups can sweep to_remote (but to_local was broken in v0), legacy channels are not considered
+        assert not self.has_anchors()
+        return True
+
     def can_be_deleted(self):
         return self.is_imported or self.is_redeemed()
 
@@ -763,7 +777,7 @@ class ChannelBackup(AbstractChannel):
 
     def get_close_options(self) -> Sequence[ChanCloseOption]:
         ret = []
-        if self.get_state() == ChannelState.FUNDED:
+        if self.get_state() == ChannelState.FUNDED and self.can_sweep_their_ctx_to_remote():
             ret.append(ChanCloseOption.REQUEST_REMOTE_FCLOSE)
         return ret
 
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -620,7 +620,6 @@ class ChannelBackup(AbstractChannel):
             payment_basepoint=local_payment_basepoint,
             multisig_key=multisig_funding_keypair,
             # dummy values
-            static_remotekey=None,
             static_payment_key=None,
             dust_limit_sat=None,
             max_htlc_value_in_flight_msat=None,
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -618,10 +618,6 @@ class ChannelBackup(AbstractChannel):
 
     def init_config(self, cb: ImportedChannelBackupStorage):
         local_payment_basepoint = cb.local_payment_basepoint
-        if local_payment_basepoint is None:
-            self.logger.warning(
-                f"local_payment_basepoint missing from (old-type) channel backup. "
-                f"You should export and re-import a newer backup.")
         multisig_funding_keypair = None
         if multisig_funding_secret := cb.multisig_funding_privkey:
             multisig_funding_keypair = Keypair(
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -742,6 +742,8 @@ class ChannelBackup(AbstractChannel):
         return self.lnworker.wallet.get_new_sweep_address_for_channel()
 
     def has_anchors(self) -> Optional[bool]:
+        if isinstance(self.cb, ImportedChannelBackupStorage):
+            return bool(self.cb.channel_type & ChannelType.OPTION_ANCHORS)
         return None
 
     def is_zeroconf(self) -> bool:
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -571,6 +571,26 @@ class AbstractChannel(Logger, ABC):
 
 class ChannelBackup(AbstractChannel):
     """
+    * v0: added in first LN release, 4.0
+      - can be either for a pre-SRK (legacy) channel or an SRK channel
+    * v1: added in 4.4.6 (#8536), to fix sweeping local fclose
+      - implies SRK channel
+    * v2: added together with anchor chans, in 4.6
+      - can be either for an SRK or an anchors chan
+    * v3: added in 4.8.2 (#10852), to fix anchor chan to_remote sweep
+      - can be either for an SRK or an anchor chan
+
+    Channel types:
+    * legacy (pre-SRK):
+      - pre-SRK channels could only be opened strictly before first LN release
+      - pre-SRK support was removed in 4.3.1
+      - payment_basepoint was derived from backup (removed in #10852)
+    * static_remotekey:
+      - to_remote sweep not necessary due to wallet address
+    * anchors:
+      - sweep to_remote with local_payment_basepoint if it is a private key,
+        otherwise by deriving the key from the funding pubkeys (requires deterministic lightning)
+
     current capabilities:
       - detect force close
       - request force close
@@ -612,11 +632,6 @@ class ChannelBackup(AbstractChannel):
             channel_seed=cb.channel_seed,
             to_self_delay=cb.local_delay,
             channel_type=cb.channel_type,
-            # there are three cases of backups:
-            # 1. legacy: deprecated, it cannot be recovered
-            # 2. static_remotekey: to_remote sweep not necessary due to wallet address
-            # 3. anchor outputs: sweep to_remote with local_payment_basepoint if it is a private key, otherwise
-            #           by deriving the key from the funding pubkeys
             payment_basepoint=local_payment_basepoint,
             multisig_key=multisig_funding_keypair,
             # dummy values
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -575,8 +575,7 @@ class ChannelBackup(AbstractChannel):
       - detect force close
       - request force close
       - sweep my ctx to_local
-    future:
-      - will need to sweep their ctx to_remote
+      - sweep their ctx to_remote (anchor channels, with srk it is a wallet address)
     """
 
     def __init__(self, cb: ChannelBackupStorage, *, lnworker: 'LNWallet'):
@@ -598,10 +597,10 @@ class ChannelBackup(AbstractChannel):
         self.unconfirmed_closing_txid = None # not a state, only for GUI
 
     def init_config(self, cb: ImportedChannelBackupStorage):
-        local_payment_pubkey = cb.local_payment_pubkey
-        if local_payment_pubkey is None:
+        local_payment_basepoint = cb.local_payment_basepoint
+        if local_payment_basepoint is None:
             self.logger.warning(
-                f"local_payment_pubkey missing from (old-type) channel backup. "
+                f"local_payment_basepoint missing from (old-type) channel backup. "
                 f"You should export and re-import a newer backup.")
         multisig_funding_keypair = None
         if multisig_funding_secret := cb.multisig_funding_privkey:
@@ -612,13 +611,16 @@ class ChannelBackup(AbstractChannel):
         self.config[LOCAL] = LocalConfig.from_seed(
             channel_seed=cb.channel_seed,
             to_self_delay=cb.local_delay,
+            channel_type=cb.channel_type,
             # there are three cases of backups:
             # 1. legacy: payment_basepoint will be derived
             # 2. static_remotekey: to_remote sweep not necessary due to wallet address
-            # 3. anchor outputs: sweep to_remote by deriving the key from the funding pubkeys
-            static_remotekey=local_payment_pubkey,
+            # 3. anchor outputs: sweep to_remote with local_payment_basepoint if it is a private key, otherwise
+            #           by deriving the key from the funding pubkeys
+            payment_basepoint=local_payment_basepoint,
             multisig_key=multisig_funding_keypair,
             # dummy values
+            static_remotekey=None,
             static_payment_key=None,
             dust_limit_sat=None,
             max_htlc_value_in_flight_msat=None,
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -5,7 +5,7 @@ from enum import IntFlag, IntEnum
 import enum
 from typing import (
     NamedTuple, List, Tuple, Mapping, Optional, TYPE_CHECKING, Union, Dict, Set, Sequence, FrozenSet,
-    TypedDict,
+    TypedDict, Literal
 )
 import sys
 import time
@@ -78,6 +78,15 @@ def hex_to_bytes(arg: Optional[Union[bytes, str]]) -> Optional[bytes]:
 
 def bytes_to_hex(arg: Optional[bytes]) -> Optional[str]:
     return repr(arg.hex()) if arg is not None else None
+
+
+def int_min_byte_len(n: int) -> int:
+    """Returns the smallest number of bytes that can represent n (zero -> 0 bytes)."""
+    return (n.bit_length() + 7) // 8
+
+
+def int_to_bytes_minimal(n: int, byteorder: Literal['big', 'little'] = 'big') -> bytes:
+    return int.to_bytes(n, length=int_min_byte_len(n), byteorder=byteorder)
 
 
 def json_to_keypair(arg: Union['OnlyPubkeyKeypair', dict]) -> Union['OnlyPubkeyKeypair', 'Keypair']:
@@ -1628,10 +1637,6 @@ class LnFeatures(IntFlag):
                 features |= (1 << flag)
         return features
 
-    def min_len(self) -> int:
-        b = int.bit_length(self)
-        return b // 8 + int(bool(b % 8))
-
     def supports(self, feature: 'LnFeatures') -> bool:
         """Returns whether given feature is enabled.
 
@@ -1718,12 +1723,6 @@ class ChannelType(IntFlag):
             if not peer_features.supports(feature):
                 return False
         return True
-
-    def to_bytes_minimal(self):
-        # MUST use the smallest bitmap possible to represent the channel type.
-        bit_length = self.value.bit_length()
-        byte_length = bit_length // 8 + int(bool(bit_length % 8))
-        return self.to_bytes(byte_length, byteorder='big')
 
     @property
     def name_minimal(self):
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -256,6 +256,9 @@ class LocalConfig(ChannelConfig):
         kwargs['revocation_basepoint'] = keypair_generator(LnKeyFamily.REVOCATION_BASE)
         static_remotekey = kwargs.pop('static_remotekey')
         static_payment_key = kwargs.pop('static_payment_key')
+        channel_type = kwargs.pop('channel_type')
+        payment_basepoint = kwargs.pop('payment_basepoint', None)
+        assert bool(static_remotekey) + bool(static_payment_key) + bool(payment_basepoint) <= 1
         if static_payment_key:
             # We derive the payment_basepoint from a static secret (derived from
             # the wallet seed) and a public nonce that is revealed
@@ -267,10 +270,20 @@ class LocalConfig(ChannelConfig):
             )
         elif static_remotekey:  # we automatically sweep to a wallet address
             kwargs['payment_basepoint'] = OnlyPubkeyKeypair(static_remotekey)
+        elif payment_basepoint:  # channel backup
+            if len(payment_basepoint) == 32:  # privkey
+                assert channel_type & ChannelType.OPTION_ANCHORS
+                privkey = ecc.ECPrivkey(payment_basepoint)
+                kwargs['payment_basepoint'] = Keypair(privkey=privkey.get_secret_bytes(), pubkey=privkey.get_public_key_bytes())
+            else:
+                assert len(payment_basepoint) == 33  # pubkey
+                kwargs['payment_basepoint'] = OnlyPubkeyKeypair(payment_basepoint)
         else:
-            # we expect all our channels to use option_static_remotekey, so ending up here likely indicates an issue...
+            # v0 channel backup for srk channel, note the derived key is bogus and cannot be used
+            # see: https://github.com/spesmilo/electrum/pull/8536
             kwargs['payment_basepoint'] = keypair_generator(LnKeyFamily.PAYMENT_BASE)
 
+        assert ecc.ECPubkey.is_pubkey_bytes(kwargs['payment_basepoint'].pubkey)
         return LocalConfig(**kwargs)
 
     def validate_params(self, *, funding_sat: int, config: 'SimpleConfig', peer_features: 'LnFeatures') -> None:
@@ -308,8 +321,8 @@ class ChannelConstraints(StoredObject):
     funding_txn_minimum_depth = attr.ib(type=int)
 
 
-CHANNEL_BACKUP_VERSION_LATEST = 2
-KNOWN_CHANNEL_BACKUP_VERSIONS = (0, 1, 2, )
+CHANNEL_BACKUP_VERSION_LATEST = 3
+KNOWN_CHANNEL_BACKUP_VERSIONS = (0, 1, 2, 3, )
 assert CHANNEL_BACKUP_VERSION_LATEST in KNOWN_CHANNEL_BACKUP_VERSIONS
 
 
@@ -342,26 +355,37 @@ class OnchainChannelBackupStorage(ChannelBackupStorage):
         return OnchainChannelBackupStorage(**kwargs)
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True, kw_only=True)
 class ImportedChannelBackupStorage(ChannelBackupStorage):
+    backup_version: int = CHANNEL_BACKUP_VERSION_LATEST
     node_id: bytes  # remote node pubkey
     privkey: bytes  # local node privkey
     host: str
     port: int
     channel_seed: bytes
+    channel_type: int
     local_delay: int
     remote_delay: int
     remote_payment_pubkey: bytes
     remote_revocation_pubkey: bytes
-    local_payment_pubkey: Optional[bytes]
+    # can either be a pubkey or a privkey (for anchor channels)
+    local_payment_basepoint: Optional[bytes]
     multisig_funding_privkey: Optional[bytes]
 
+    def __post_init__(self):
+        # strip the variation flags, they are irrelevant for the backup
+        channel_type = int(self.channel_type) & ~(ChannelType.OPTION_SCID_ALIAS | ChannelType.OPTION_ZEROCONF)
+        object.__setattr__(self, 'channel_type', channel_type)
+
     def to_bytes(self) -> bytes:
+        if self.backup_version != CHANNEL_BACKUP_VERSION_LATEST:
+            raise Exception("cannot re-serialize old-version channel backup")
         vds = BCDataStream()
         vds.write_uint16(CHANNEL_BACKUP_VERSION_LATEST)
         vds.write_boolean(self.is_initiator)
         vds.write_bytes(self.privkey, 32)
         vds.write_bytes(self.channel_seed, 32)
+        vds.write_string(int_to_bytes_minimal(self.channel_type))
         vds.write_bytes(self.node_id, 33)
         vds.write_bytes(bfh(self.funding_txid), 32)
         vds.write_uint16(self.funding_index)
@@ -372,7 +396,13 @@ class ImportedChannelBackupStorage(ChannelBackupStorage):
         vds.write_uint16(self.remote_delay)
         vds.write_string(self.host)
         vds.write_uint16(self.port)
-        vds.write_bytes(self.local_payment_pubkey, 33)
+        if len(self.local_payment_basepoint) == 32:  # private key
+            assert self.channel_type == ChannelType.OPTION_STATIC_REMOTEKEY | ChannelType.OPTION_ANCHORS
+            vds.write_bytes(b"\x00" + self.local_payment_basepoint, 33)
+        else:
+            assert len(self.local_payment_basepoint) == 33  # pubkey
+            assert self.channel_type == ChannelType.OPTION_STATIC_REMOTEKEY
+            vds.write_bytes(self.local_payment_basepoint, 33)
         vds.write_bytes(self.multisig_funding_privkey, 32)
         return bytes(vds.input)
 
@@ -386,6 +416,10 @@ class ImportedChannelBackupStorage(ChannelBackupStorage):
         is_initiator = vds.read_boolean()
         privkey = vds.read_bytes(32)
         channel_seed = vds.read_bytes(32)
+        channel_type = None
+        if version >= 3:
+            channel_type_length = vds.read_compact_size()
+            channel_type = ChannelType.from_bytes(vds.read_bytes(channel_type_length), byteorder='big')
         node_id = vds.read_bytes(33)
         funding_txid = vds.read_bytes(32).hex()
         funding_index = vds.read_uint16()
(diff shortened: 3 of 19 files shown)


electrum/lnutil.py
```diff
@@ -247,6 +247,8 @@ def keypair_generator(family: 'LnKeyFamily') -> 'Keypair':
         kwargs['revocation_basepoint'] = keypair_generator(LnKeyFamily.REVOCATION_BASE)
         static_remotekey = kwargs.pop('static_remotekey')
         static_payment_key = kwargs.pop('static_payment_key')
+        channel_type = kwargs.pop('channel_type', None)
+        payment_basepoint = kwargs.pop('payment_basepoint', None)
```
SomberNight: can we add:
```py
assert (static_remotekey is not None) + (static_payment_key is not None) + (payment_basepoint is not None) == 1
```
Also, in the last `else` branch, "ending up here likely indicates an issue", just raise an exception.

tests/test_lnwallet.py
```diff
@@ -584,3 +592,230 @@ async def test_missing_channel_update_from_failed_htlc(self):
...
+        imports the channel backup, and sweeps her to_local output once the CSV delay expired."""
+        csv_delay = 5  # demanded of alice by bob
+        self.config.LIGHTNING_TO_SELF_DELAY_CSV = csv_delay
+        alice = self.create_non_deterministic_xprv_wallet("alice")
+        await self.pay_to_address(alice.get_receiving_address(), 1 * COIN)
+        await self.mine_blocks(1)
+
+        chan = await self.open_channel(alice, self.bob)
+        self.assertTrue(chan.has_anchors())
+        self.assertEqual(csv_delay, chan.config[REMOTE].to_self_delay)
```
SomberNight: This should not work.
`self.config` is Alice's config, and setting `config.LIGHTNING_TO_SELF_DELAY_CSV = 5` means Alice will impose a 5 lock CSV on Bob when Bob force-closes.
But in this case Alice will force-close. So the timelock is chosen by Bob. So how come the rest of the test works if Bob is supposed to be choosing the default of 1008 blocks but after Alice force-closes and we wait only 5 blocks and then Alice can sweep?
Well it's because Alice and Bob share the ToyNetwork, which has a config object, so they sort-of share a config object. (well Alice has a single config object, and Bob has his own but also has a reference to Alice's through the network object... confusing)

This assert should not fail ideally:
```diff
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index 58cc77c330..3f4e6b8876 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -424,6 +424,7 @@ class LNPeerManager(Logger, EventListener, NetworkRetryManager[LNPeerAddr]):
         assert network
         assert self.network is None, "already started"
         self.network = network
+        assert network.config is self.config
         self._add_peers_from_config()
         asyncio.run_coroutine_threadsafe(self.main_loop(), get_asyncio_loop())
         if listen:
```

electrum/lnutil.py
```diff
@@ -258,6 +260,12 @@ def keypair_generator(family: 'LnKeyFamily') -> 'Keypair':
             )
         elif static_remotekey:  # we automatically sweep to a wallet address
             kwargs['payment_basepoint'] = OnlyPubkeyKeypair(static_remotekey)
+        elif payment_basepoint:
+            if channel_type is not None and channel_type & ChannelType.OPTION_ANCHORS:
+                privkey = ecc.ECPrivkey(payment_basepoint)
+                kwargs['payment_basepoint'] = Keypair(privkey=privkey.get_secret_bytes(), pubkey=privkey.get_public_key_bytes())
+            else:
+                kwargs['payment_basepoint'] = OnlyPubkeyKeypair(payment_basepoint)
```
SomberNight: maybe assert here that `len(payment_basepoint) == 33` or even `ECPubkey.is_pubkey_bytes(payment_basepoint)`

electrum/lnutil.py
f321x: Added an assert in https://github.com/spesmilo/electrum/compare/bf0f96dcc0452397cecaf2d52355d799c57ae8c8..a8a5b2d2ad1c2defdc98adf374a3983ae57c8dfc
But we can't assert == 1 or raise at the else clause as v0 channel backups contain none of the fields.

The `LocalConfig.from_seed()` method is a bit confusing in general, IIUC this else clause was useful for pre-srk channels (for which channel backups never got exposed) and then wasn't removed later on.
For v0 backups / srk channels the derivation is basically nonsense?
So I added a commit which makes this explicit instead of misleading the reader: https://github.com/spesmilo/electrum/pull/10852/commits/8d9544cf1e0d6ed62661b5ef80dbd0d89faac3ac

tests/test_lnwallet.py
f321x: Right, good catch, this is confusing. I changed the `ToyServerTestCase` in https://github.com/spesmilo/electrum/pull/10852/commits/1362998dce9ce74a8117652e3a309eb2f80fbd88 so it now handles separate instances (similar to a separate Electrum process) with their own configs (see `ToyInstance`). This separates the owners of the config transparently.

f321x:
@SomberNight as discussed I also added more unittests in https://github.com/spesmilo/electrum/pull/10852/commits/c649d51f223aafc2b5dea7bedf09281ee8674d6a, some might overlap a bit, e.g. the local force close tests are taking a pretty similar path to claim the `to_local` output. Maybe still useful considering the previous subtle bugs we had with this logic.

electrum/lnutil.py
SomberNight: > IIUC this else clause was useful for pre-srk channels (for which channel backups never got exposed) and then wasn't removed later on.

Right, the else branch was only useful for pre-srk channels. I think we had channel backups for those too, on master. However already the first release that supported lightning required `option_static_remotekey`.
For pre-srk, the branch was needed, see e.g. https://github.com/spesmilo/electrum/commit/2255b0715704bf9e32a4e5b9d1fda511be7b5938 (lnsweep line 327), which relied on `option_data_loss_protect` (i.e. the peer sending us `their_pcp` in `channel_reestablish`)

electrum/lnutil.py
```diff
@@ -258,8 +261,15 @@ def keypair_generator(family: 'LnKeyFamily') -> 'Keypair':
             )
         elif static_remotekey:  # we automatically sweep to a wallet address
             kwargs['payment_basepoint'] = OnlyPubkeyKeypair(static_remotekey)
+        elif payment_basepoint:  # channel backup
+            if channel_type is not None and channel_type & ChannelType.OPTION_ANCHORS:
+                privkey = ecc.ECPrivkey(payment_basepoint)
+                kwargs['payment_basepoint'] = Keypair(privkey=privkey.get_secret_bytes(), pubkey=privkey.get_public_key_bytes())
+            else:
+                kwargs['payment_basepoint'] = OnlyPubkeyKeypair(payment_basepoint)
```
SomberNight, line 277: maybe assert here that `len(payment_basepoint) == 33` or even `ECPubkey.is_pubkey_bytes(payment_basepoint)`

electrum/wallet_db.py
```diff
@@ -1436,6 +1437,17 @@ def _convert_version_71(self):
...
+    def _convert_version_72(self):
+        """Rename 'local_payment_pubkey' to 'local_payment_basepoint' in imported channel
+        backups, and add 'channel_type' (None, as pre-v3 backups do not contain it)."""
+        if not self._is_upgrade_method_needed(71, 71):
+            return
+        cbs = self.data.get('imported_channel_backups', {})
+        for cb in cbs.values():
+            cb['local_payment_basepoint'] = cb.pop('local_payment_pubkey', None)
+            cb['channel_type'] = None
+        self.data['seed_version'] = 72
```
SomberNight, line 1481: note: This db upgrade is only needed because there are two ways to (de)serialize `lnutil.ImportedChannelBackupStorage`: we can serialize it to bytes used in the exports or serialize it to the storage as json. I guess it would be simpler if we removed the json-serialization, I don't see the point of it. Instead of subclassing `StoredObject`, `ImportedChannelBackupStorage` could be a purely in-memory object, maybe even frozen and be a NamedTuple, and we could serialize it using its existing to_bytes() method into the db.

We kind of need to support importing old type chan backups, so on wallet db open we would just run the same deser logic on what is in the wallet db as the logic we would run on importing another backup string.

Well ok, I guess this is mostly orthogonal to current PR. And maybe there are issues I don't foresee.

electrum/lnutil.py
```diff
@@ -380,18 +401,27 @@ def from_bytes(s: bytes) -> 'ImportedChannelBackupStorage':
...
+
+        local_payment_basepoint = None
+        if version >= 3:
+            if channel_type == ChannelType.OPTION_STATIC_REMOTEKEY | ChannelType.OPTION_ANCHORS:
+                local_payment_basepoint = vds.read_bytes(32)
+            else:
+                assert channel_type == ChannelType.OPTION_STATIC_REMOTEKEY, channel_type
+                local_payment_basepoint = vds.read_bytes(33)
+        elif version >= 1:
+            local_payment_basepoint = vds.read_bytes(33)
```
SomberNight: What do you think about simplifying this branching a bit by always writing 33 bytes?
Similar to [how xprvs are serialized](https://github.com/bitcoin/bips/blob/7fe0b034ec967b52a5a28276419117326df93263/bip-0032.mediawiki?plain=1#L136), private keys could be prefixed by a zero byte:
```diff
diff --git a/electrum/lnutil.py b/electrum/lnutil.py
index a79b4f048c..25b997c423 100644
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -372,11 +372,13 @@ class ImportedChannelBackupStorage(ChannelBackupStorage):
         vds.write_uint16(self.remote_delay)
         vds.write_string(self.host)
         vds.write_uint16(self.port)
-        if self.channel_type == ChannelType.OPTION_STATIC_REMOTEKEY | ChannelType.OPTION_ANCHORS:
-            vds.write_bytes(self.local_payment_basepoint, 32)  # private key
+        if len(self.local_payment_basepoint) == 32:  # private key
+            assert self.channel_type == ChannelType.OPTION_STATIC_REMOTEKEY | ChannelType.OPTION_ANCHORS
+            vds.write_bytes(b"\x00" + self.local_payment_basepoint, 33)
         else:
+            assert len(self.local_payment_basepoint) == 33  # pubkey
             assert self.channel_type == ChannelType.OPTION_STATIC_REMOTEKEY
-            vds.write_bytes(self.local_payment_basepoint, 33)  # pubkey
+            vds.write_bytes(self.local_payment_basepoint, 33)
         vds.write_bytes(self.multisig_funding_privkey, 32)
         return bytes(vds.input)
 
@@ -401,17 +403,11 @@ class ImportedChannelBackupStorage(ChannelBackupStorage):
         remote_delay = vds.read_uint16()
         host = vds.read_string()
         port = vds.read_uint16()
-
-        local_payment_basepoint = None
-        if version >= 3:
-            if channel_type == ChannelType.OPTION_STATIC_REMOTEKEY | ChannelType.OPTION_ANCHORS:
-                local_payment_basepoint = vds.read_bytes(32)
-            else:
-                assert channel_type == ChannelType.OPTION_STATIC_REMOTEKEY, channel_type
-                local_payment_basepoint = vds.read_bytes(33)
-        elif version >= 1:
+        local_payment_basepoint = None  # type: Optional[bytes]
+        if version >= 1:
             local_payment_basepoint = vds.read_bytes(33)
-
+            if local_payment_basepoint[0] == 0:  # private key
+                local_payment_basepoint = local_payment_basepoint[1:]
         if version >= 2:
             multisig_funding_privkey = vds.read_bytes(32)
         else:

```

SomberNight (commented):
some comments (but still looking)

Also, I merged the first two commits into master in 8c260c9cae4406ae9a987dbf73e90c1c944a163d

electrum/lnutil.py
f321x: nice trick, doing this now in https://github.com/spesmilo/electrum/compare/d18913ad56c0e076a46f920a1eb2db6164aef242..b8812c06595d1eb6678091d79dda945834fb7c2e

electrum/lnutil.py
f321x: Added an `ECPubkey.is_pubkey_bytes(payment_basepoint)` assert in https://github.com/spesmilo/electrum/compare/d18913ad56c0e076a46f920a1eb2db6164aef242..b8812c06595d1eb6678091d79dda945834fb7c2e

electrum/wallet_db.py
f321x: Yeah I also think it would be easier have just a single format, seems like less risk for drift and easier to reason about.

I gave this a try in https://github.com/spesmilo/electrum/pull/10852/commits/ec4a0365787da342ee1be0736dc259c69adb6f46. I pushed the same branch without this change to `fix_10785_backup` to have a clear diff of the change.
I used a dataclass so the inheritance from `ChannelBackupStorage` stays as is.

electrum/lnutil.py
SomberNight, line 1: note: you have not signed these commits. Consider enabling ["vigilant mode"](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits) in the [github acc settings](https://github.com/settings/keys  ).

electrum/wallet_db.py
SomberNight: I think this is better. I merged that commit in https://github.com/spesmilo/electrum/pull/10920.

electrum/lnutil.py
SomberNight: If there will be a setting to force maintainers to sign commits, we should enable that.
Looks like it does not exist atm: https://github.com/orgs/community/discussions/69706

There is a setting that requires *every* commit to be signed but that's unusable in practice given we accept contributions from anyone.

f321x: Must have rebased without `-S`, pushed with sigs now and enabled vigilant mode.

electrum/lnchannel.py
```diff
@@ -656,6 +658,20 @@ def init_config(self, cb: ImportedChannelBackupStorage):
...
         )
 
+    def can_sweep_their_ctx_to_remote(self) -> bool:
+        cb = self.cb
+        if not isinstance(cb, ImportedChannelBackupStorage):
+            return True  # on-chain backups only exist for deterministic wallets
+        if self.has_anchors():
+            return True  # has_anchors can only be determined with >=v3, which also has the secret for the to_remote sweep
+        if cb.local_payment_basepoint is None:
+            return True  # v0 backup, only had srk channels
```
SomberNight: v1 backups were only introduced in 4.4.6 (https://github.com/spesmilo/electrum/pull/8536)

- pre-SRK channels could only be opened strictly before first LN release
- some time later, still before first LN release, we implemented and started requiring SRK channels for newly opened chans
- v0 backups were added in first LN release, 4.0
- support for pre-SRK channel was removed in 4.3.1. I think wallets that might still contain them can no longer be opened.
- v1 backups were added in 4.4.6
- v2 backups were added together with anchor chans, in 4.6

so

- a v0 backup can be either for a pre-SRK channel or an SRK channel
    - as someone who opened a pre-SRK chan before the first LN release and then kept using the wallet into the 4.0, 4.1, 4.2 releases could have exported a chan backup of their old (at-the-time-still-open) chan
    -  however we don't support pre-SRK anymore
- a v1 backup implies SRK chan
- a v2 backup can be either for an SRK or an anchors chan

electrum/lnutil.py
```diff
@@ -340,19 +351,30 @@ class ImportedChannelBackupStorage(ChannelBackupStorage):
...
+
     def to_bytes(self) -> bytes:
+        if self.channel_type is None:
+            raise Exception("cannot re-serialize pre-v3 channel backup")
         vds = BCDataStream()
         vds.write_uint16(CHANNEL_BACKUP_VERSION_LATEST)
         vds.write_boolean(self.is_initiator)
         vds.write_bytes(self.privkey, 32)
         vds.write_bytes(self.channel_seed, 32)
+        vds.write_uint64(self.channel_type)
```
SomberNight: uint64 is too small. it already overflows for `option_simple_taproot` (bits 80/81)

in the bolts, it is variable length:
https://github.com/lightning/bolts/blob/152897261850d93c4f4597f39cf22d7d22d6ede6/02-peer-protocol.md#L713

I guess we should use something like `to_bytes_minimal` instead:
https://github.com/spesmilo/electrum/blob/ace379b0da9aa8141a5533d714298103d790cbcc/electrum/lnutil.py#L1722
maybe factor that out so that it can be used with a generic int?

and then you can use `vds.write_string` for it

electrum/lnchannel.py
```diff
@@ -656,6 +658,20 @@ def init_config(self, cb: ImportedChannelBackupStorage):
...
+        to_remote_addr = make_commitment_output_to_remote_address(cb.local_payment_basepoint, has_anchors=False)
+        if self.lnworker.wallet.is_mine(to_remote_addr):
+            return True  # legacy static_remotekey: to_remote is a wallet address
+        # pre-v3: only sweepable if we still have the LN keys that created this backup
+        return cb.privkey == self.lnworker.node_keypair.privkey
```
SomberNight: If we always knew the channel_type for an ImportedChannelBackupStorage, I think this whole block could be simplified (some of it can be kept as asserts maybe).
Relatedly, we could store "backup_version" as well.

see https://github.com/spesmilo/electrum/commit/0ce17ddc800bdc20ab8e275b1571041db8fd6f16 (from https://github.com/SomberNight/electrum/compare/202609_ln_chan_backup_infer_channeltype)

electrum/lnchannel.py
```diff
@@ -612,7 +612,7 @@ def init_config(self, cb: ImportedChannelBackupStorage):
             channel_seed=cb.channel_seed,
             to_self_delay=cb.local_delay,
             # there are three cases of backups:
-            # 1. legacy: payment_basepoint will be derived
+            # 1. v0: no payment_basepoint, it cannot be recovered
             # 2. static_remotekey: to_remote sweep not necessary due to wallet address
             # 3. anchor outputs: sweep to_remote with local_payment_basepoint if it is a private key, otherwise
             #           by deriving the key from the funding pubkeys
```
SomberNight: This comment is now confusing. It is mixing chan backup versions and chan types. Originally it was strictly about chan types (it predates chan backup version 1: https://github.com/spesmilo/electrum/pull/7509/changes/6426e3a7661d9c0ae42460ce7ae7656094b8ee57).

electrum/lnutil.py
```diff
@@ -258,10 +261,18 @@ def keypair_generator(family: 'LnKeyFamily') -> 'Keypair':
...
+                privkey = ecc.ECPrivkey(payment_basepoint)
+                kwargs['payment_basepoint'] = Keypair(privkey=privkey.get_secret_bytes(), pubkey=privkey.get_public_key_bytes())
+            else:
+                kwargs['payment_basepoint'] = OnlyPubkeyKeypair(payment_basepoint)
         else:
-            # we expect all our channels to use option_static_remotekey, so ending up here likely indicates an issue...
-            kwargs['payment_basepoint'] = keypair_generator(LnKeyFamily.PAYMENT_BASE)
+            # v0 channel backup for srk channel: the real basepoint is a wallet pubkey that is
+            # not part of the backup and cannot be derived, see: https://github.com/spesmilo/electrum/pull/8536
+            kwargs['payment_basepoint'] = OnlyPubkeyKeypair(None)
```
SomberNight: small simplification (maybe??).
It is subjective so feel free to skip.


```diff
diff --git a/electrum/lnchannel.py b/electrum/lnchannel.py
index 57326f41ba..0b3b397711 100644
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -620,7 +620,6 @@ class ChannelBackup(AbstractChannel):
             payment_basepoint=local_payment_basepoint,
             multisig_key=multisig_funding_keypair,
             # dummy values
-            static_remotekey=None,
             static_payment_key=None,
             dust_limit_sat=None,
             max_htlc_value_in_flight_msat=None,
diff --git a/electrum/lnutil.py b/electrum/lnutil.py
index 09ca42135d..cb9b00a047 100644
--- a/electrum/lnutil.py
+++ b/electrum/lnutil.py
@@ -245,11 +245,10 @@ class LocalConfig(ChannelConfig):
         kwargs['htlc_basepoint'] = keypair_generator(LnKeyFamily.HTLC_BASE)
         kwargs['delayed_basepoint'] = keypair_generator(LnKeyFamily.DELAY_BASE)
         kwargs['revocation_basepoint'] = keypair_generator(LnKeyFamily.REVOCATION_BASE)
-        static_remotekey = kwargs.pop('static_remotekey')
         static_payment_key = kwargs.pop('static_payment_key')
         channel_type = kwargs.pop('channel_type')
-        payment_basepoint = kwargs.pop('payment_basepoint', None)
-        assert bool(static_remotekey) + bool(static_payment_key) + bool(payment_basepoint) <= 1
+        payment_basepoint = kwargs.pop('payment_basepoint', None)  # type: bytes | None
+        assert bool(static_payment_key) + bool(payment_basepoint) <= 1
         if static_payment_key:
             assert channel_type & ChannelType.OPTION_ANCHORS
             # We derive the payment_basepoint from a static secret (derived from
@@ -260,10 +259,7 @@ class LocalConfig(ChannelConfig):
                 static_payment_secret=static_payment_key.privkey,
                 funding_pubkey=kwargs['multisig_key'].pubkey
             )
-        elif static_remotekey:  # we automatically sweep to a wallet address
-            assert channel_type == ChannelType.OPTION_STATIC_REMOTEKEY
-            kwargs['payment_basepoint'] = OnlyPubkeyKeypair(static_remotekey)
-        elif payment_basepoint:  # channel backup
+        elif payment_basepoint:  # channel backup (or new SRK chan in unit tests)
             if len(payment_basepoint) == 32:  # privkey
                 assert channel_type & ChannelType.OPTION_ANCHORS
                 privkey = ecc.ECPrivkey(payment_basepoint)
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index c5c2d9edd4..99c0995dde 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1747,7 +1747,7 @@ class LNWallet(Logger):
         channel_type.check_combinations()  # test if raises
         if channel_type & ChannelType.OPTION_ANCHORS:  # anchors
             static_payment_key = self.static_payment_key
-            static_remotekey = None
+            payment_basepoint = None
         else:  # static_remotekey
             assert channel_type & channel_type.OPTION_STATIC_REMOTEKEY
             assert self.config.TEST_LN_OPEN_SRK_CHANNELS
@@ -1755,7 +1755,7 @@ class LNWallet(Logger):
             assert wallet.txin_type == 'p2wpkh'
             addr = wallet.get_new_sweep_address_for_channel()
             static_payment_key = None
-            static_remotekey = bytes.fromhex(wallet.get_public_key(addr))
+            payment_basepoint = bytes.fromhex(wallet.get_public_key(addr))
 
         if multisig_funding_keypair:
             for chan in self.channels.values():  # check against all chans of lnworker, for sanity
@@ -1775,7 +1775,7 @@ class LNWallet(Logger):
         local_config = LocalConfig.from_seed(
             channel_seed=channel_seed,
             channel_type=channel_type,
-            static_remotekey=static_remotekey,
+            payment_basepoint=payment_basepoint,
             static_payment_key=static_payment_key,
             multisig_key=multisig_funding_keypair,
             upfront_shutdown_script=upfront_shutdown_script,

```

tests/test_lnwallet.py
```diff
@@ -618,3 +630,304 @@ async def test_missing_channel_update_from_failed_htlc(self):
...
+        wallet.init_lightning(password=None)
+        self.assertFalse(wallet.lnworker.has_deterministic_node_id())
+        return wallet
+
+    def create_config(self, name: Optional[str] = None) -> SimpleConfig:
+        config = super().create_config(name)
+        config.LIGHTNING_USE_RECOVERABLE_CHANNELS = False  # onchain backup tests opt in explicitly
+        return config
+
+    async def open_channel(self, from_wallet, to_wallet) -> Channel:
```
SomberNight: missing type-hints
```py
async def open_channel(self, from_wallet: Abstract_Wallet, to_wallet: Abstract_Wallet) -> Channel:
```

tests/test_lnwallet.py
```diff
@@ -618,3 +630,304 @@ async def test_missing_channel_update_from_failed_htlc(self):
...
+            self.assertTrue(alice.is_mine(ctx.outputs()[to_remote_idx].address))
+            return ctx
+        await self.wait_for_spender_of(TxOutpoint.from_str(f"{ctx.txid()}:{to_remote_idx}"))
+        return ctx
+
+    async def assert_balance_recovered(
+        self,
+        alice: Abstract_Wallet,
+        cb: ChannelBackup, *,
+        balance_before: int,
```
SomberNight: maybe add comment:
```
        balance_before: int,  # onchain balance before closing chan
```

tests/test_lnwallet.py
```diff
@@ -618,3 +630,304 @@ async def test_missing_channel_update_from_failed_htlc(self):
...
+        chan_id = chan.channel_id
+        alice.storage = WalletStorage(os.path.join(alice.config.get_datadir_wallet_path(), 'alice_wallet'))
+        backup_path = alice.save_backup(self.electrum_path)
+        with open(backup_path) as f:
+            backup_db = WalletDB(f.read(), storage=None, upgrade=False)
+        cb_blob = backup_db.get_dict('imported_channel_backups')[chan_id.hex()]
+        self.assertEqual(
+            alice.lnworker.create_channel_backup(chan_id),
+            ImportedChannelBackupStorage.from_bytes(bfh(cb_blob)),
+        )
```
SomberNight, line 782:
```diff
diff --git a/tests/test_lnwallet.py b/tests/test_lnwallet.py
index b2cf3b6e9a..5092c87cea 100644
--- a/tests/test_lnwallet.py
+++ b/tests/test_lnwallet.py
@@ -780,6 +780,9 @@ class TestChannelBackup(ToyServerTestCase):
             alice.lnworker.create_channel_backup(chan_id),
             ImportedChannelBackupStorage.from_bytes(bfh(cb_blob)),
         )
+        # check full channels are removed from file backups:
+        self.assertEqual(1, len(alice.db.get_dict('channels')))
+        self.assertEqual(0, len(backup_db.get_dict('channels')))
 
     @staticmethod
     def serialize_v1_channel_backup(cb: ImportedChannelBackupStorage) -> bytes:

```

tests/test_lnwallet.py
```diff
@@ -618,3 +630,304 @@ async def test_missing_channel_update_from_failed_htlc(self):
...
+        vds.write_uint16(cb.local_delay)
+        vds.write_uint16(cb.remote_delay)
+        vds.write_string(cb.host)
+        vds.write_uint16(cb.port)
+        # v1 stores the payment pubkey
+        vds.write_bytes(ecc.ECPrivkey(cb.local_payment_basepoint).get_public_key_bytes(), 33)
+        return bytes(vds.input)
+
+    async def test_request_fclose_from_pre_v2_anchor_chan_backup_deterministic_wallet(self):
+        """Test recovery with an older channel backup on a deterministic lnwallet with anchor channel"""
```
SomberNight: Anchor chan support was added together with v2 chan backups. There cannot exist v1 backups for anchor chans. So this test does not make sense.

What you should test instead is serializing the chan backup as v2 (not v1). That would be a small change to this test and that would simulate something that is realistic: someone using Electrum 4.6-4.8 opening an anchor chan and exporting a backup of it.

tests/test_lnwallet.py
```diff
@@ -618,3 +630,304 @@ async def test_missing_channel_update_from_failed_htlc(self):
...
+    async def test_local_fclose_from_srk_chan_backup_non_deterministic_lightning(self):
+        await self._test_local_fclose_then_sweep_to_local_from_chan_backup(
+            create_alice_cb=self.create_non_deterministic_xprv_wallet,
+            anchors=False,
+        )
+
+    async def _test_request_fclose_from_onchain_chan_backup(self, *, anchors: bool) -> None:
+        """Alice has a deterministic LNWallet, so her funding txs contain an OP_RETURN onchain backup.
+        She loses her wallet db, restores from her seed, discovers the backup in the funding tx,
+        and recovers her balance by asking bob to force-close."""
```
SomberNight, line 907:
```diff
diff --git a/tests/test_lnwallet.py b/tests/test_lnwallet.py
index b2cf3b6e9a..411278b3ed 100644
--- a/tests/test_lnwallet.py
+++ b/tests/test_lnwallet.py
@@ -895,7 +895,7 @@ class TestChannelBackup(ToyServerTestCase):
         )
 
     async def _test_request_fclose_from_onchain_chan_backup(self, *, anchors: bool) -> None:
-        """Alice has a deterministic LNWallet, so her funding txs contain an OP_RETURN onchain backup.
+        """Alice has a deterministic LNWallet, and her funding txs contain an OP_RETURN onchain backup.
         She loses her wallet db, restores from her seed, discovers the backup in the funding tx,
         and recovers her balance by asking bob to force-close."""
         self.alice_instance.config.LIGHTNING_USE_RECOVERABLE_CHANNELS = True

```
the first clause does not imply the second

tests/regtest/regtest.sh
```diff
@@ -251,39 +251,6 @@ if [[ $1 == "breach" ]]; then
 fi
 
 
-if [[ $1 == "backup" ]]; then
```
SomberNight, line 254: just curious, why remove this one ("backup") but keep "backup_local_forceclose"?

SomberNight (commented):
Some more comments.

The new tests in `test_lnchannel.TestChannelBackup` are amazing! Well done.

self-notes:
- ok  27249b338df2c67fde0232ef78d80192ddd9b933
    - except uint64 for channel_type is too small
- ok 07dc1c0273b9ed3c2f103ba0ec6a881c6b74c572
    - except comments are confusing
- ok 63828ffacf3afd55eb4e743f6f03fd75a6473dd6
- ok 5a7b5a7a0d8d03f5e256b071bbd695ec54842062

electrum/lnutil.py
SomberNight: tangent: we should a have sparse representation for feature bits, where a feature vector that has e.g. bits `1,3,7,40000` enabled does not take 5 kbytes to encode. Then we could also remove the upper limit of 10k we impose on feature bits (https://github.com/spesmilo/electrum/commit/c9536180c50c230af43aa75bd22c7bdd97cc70b0).

Even here, imagine if a channel type uses bit 40000. That would mean we put 5 kbytes of data into the channel backup to encode that. Would not even fit in a QR code. :P

tests/regtest/regtest.sh
f321x: I just saw `backup` and thought that's is. Missed that. Removed in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

tests/test_lnwallet.py
f321x: Right, users can still disable onchain backups. Changed in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

tests/test_lnwallet.py
f321x: Right, doesn't really make sense. It's basically just testing the anchor recovery without the secret (and deterministic wallet). Changed that to a v2 backup in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9.

tests/test_lnwallet.py
f321x: Added in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

tests/test_lnwallet.py
f321x: Added in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

tests/test_lnwallet.py
f321x: Added in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

electrum/lnutil.py
f321x: Yeah it does seem simpler, less variables to think about. Added this patch in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

electrum/lnchannel.py
f321x: Right, i moved these comments to a docstring in https://github.com/spesmilo/electrum/pull/10852/commits/b82507ecd582ada3da86bb9092e199521ff27566

electrum/lnchannel.py
f321x: Added this into a docstring in https://github.com/spesmilo/electrum/pull/10852/commits/b82507ecd582ada3da86bb9092e199521ff27566 and adapted the comment.

electrum/lnchannel.py
f321x: Yes, i have added your commit to the branch, `can_sweep_their_ctx_to_remote` is now based mostly on the version, that seems simpler.
https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9

electrum/lnutil.py
f321x: > uint64 is too small. it already overflows for option_simple_taproot (bits 80/81)

my thinking was we can just change it once a new channel type is being implemented, but yeah it seems nicer to have a future-proof solution. Now using a variable encoding as you proposed in https://github.com/spesmilo/electrum/compare/b2d4d964d079e980dfe250fb1cfe40ab7f9c74fb..e7a8343f1602cf74a895b548a7a4c845334244a9
(https://github.com/spesmilo/electrum/pull/10852/commits/cf25870a224fd4fdc9ac9efdb3f3781f32ac1b37 factors out the functions)

> tangent: we should a have sparse representation for feature bits

I did some initial exploration with claude, but this is probably out of scope to seriously work on right now:
https://github.com/f321x/electrum-fork/tree/sparse_feature_representation

electrum/gui/qml/components/main.qml
```diff
@@ -698,6 +698,28 @@ ApplicationWindow
...
+            let dialog = app.messageDialog.createObject(app, {
+                title: warnings[i].title,
+                iconSource: Qt.resolvedUrl('../../icons/warning.png'),
+                text: warnings[i].message
+            })
+            dialog.closed.connect(function() {
+                showWarning(i + 1)
+            })
+            dialog.open()
+            Daemon.currentWallet.acknowledgeWarning(warnings[i].key)
```
SomberNight: we should only call acknowledgeWarning if the user clicks "ok" I think

SomberNight (commented):
Great! just one more small thing
(at e7a8343f1602cf74a895b548a7a4c845334244a9)

tests/test_lnwallet.py
```diff
@@ -618,3 +630,311 @@ async def test_missing_channel_update_from_failed_htlc(self):
         pubkey_b = graph.workers['bob'].node_keypair.pubkey
         pubkey_d = graph.workers['dave'].node_keypair.pubkey
         self.assertEqual(amount_msat, hint_bd.cannot_send(pubkey_b < pubkey_d))
+
+
+class TestChannelBackup(ToyServerTestCase):
+    """Alice and Bob are two LNWallets sharing a ToyServer and talking to each other over a real BOLT-08 transport."""
```
SomberNight, line 636: maybe another test case: Alice already has a v2 imported chan backup imported in her wallet file for chan_id1, and she tries to import a v3 backup for the same chan id. Test that the import is successful and the new backup overwrites the existing one.

(question: should downgrades be allowed too? e.g. importing a v2 backup overwrites an existing v3 backup? I think it's not too important to have special handling for this - but if it's little code, we could disallow downgrades. idk)

electrum/gui/qml/components/main.qml
f321x: Changed in https://github.com/spesmilo/electrum/compare/e7a8343f1602cf74a895b548a7a4c845334244a9..2313e45ef19ed5d92ad7e47772fe1ac5d0dff1e8

Startup warnings for existing, broken channel backups will now also show up on every startup until the upgrades them.

tests/test_lnwallet.py
f321x: > maybe another test case

added in https://github.com/spesmilo/electrum/pull/10852/commits/2313e45ef19ed5d92ad7e47772fe1ac5d0dff1e8

> question: should downgrades be allowed too

well it is easy to prevent, so i added a check that guards against this footgun in https://github.com/spesmilo/electrum/pull/10852/commits/0c6736092353c594738aa8930bb94a93a60963d8.
It is also covered by the new unittest.

electrum/lnworker.py
```diff
@@ -1126,6 +1126,35 @@ def has_anchor_channels(self) -> bool:
...
+            warnings.append(WalletWarning(
+                key='ln_outdated_channel_backups',
+                title=_('Outdated channel backups'),
+                show_once=True,
+                message=''.join([
+                    _("The Lightning channels of this wallet cannot be recovered from seed."), ' ',
+                    _("Channel backups that were exported with an older version of Electrum "
+                      "cannot be used to request a force close of these channels."), '\n\n',
+                    _("Please export new channel backups and store them in a safe place."),
+                ])))
```
SomberNight, line 1154: If all these strings are translated, it is hard to refer to these warnings or deduce which one is being shown from a localised screenshot. Should we put some unique ids somewhere? e.g.:
```diff
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index d6ec7b7303..5d02aa610b 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1134,7 +1134,7 @@ class LNWallet(Logger):
             # (backup v3) cannot sweep the to_remote output of an anchor channel
             warnings.append(WalletWarning(
                 key='ln_outdated_channel_backups',
-                title=_('Outdated channel backups'),
+                title="(gh-10852-1) " + _('Outdated channel backups'),
                 show_once=True,
                 message=''.join([
                     _("The Lightning channels of this wallet cannot be recovered from seed."), ' ',
@@ -1145,7 +1145,7 @@ class LNWallet(Logger):
         if any(not cb.can_sweep_their_ctx_to_remote() for cb in self.channel_backups.values()):
             warnings.append(WalletWarning(
                 key='ln_unusable_channel_backups',
-                title=_('Unusable channel backups'),
+                title="(gh-10852-2) " + _('Unusable channel backups'),
                 show_once=False,  # show on every startup
                 message=''.join([
                     _("This wallet contains channel backups that cannot be used to request a force close, "

```
EDIT: (or it could be in the message body)

f321x: Interesting idea. I added it right of the title (as most read from left to right), https://github.com/spesmilo/electrum/compare/2313e45ef19ed5d92ad7e47772fe1ac5d0dff1e8..b0fde53043a4ba93abe3aeffb9aca2ac346ded83

electrum/lnworker.py
```diff
@@ -1126,6 +1126,35 @@ def has_anchor_channels(self) -> bool:
...
+                key='ln_unusable_channel_backups',
+                title=_('Unusable channel backups') + ' [gh-10852-2]',
+                show_once=False,  # show on every startup
+                message=''.join([
+                    _("This wallet contains channel backups that cannot be used to request a force close, "
+                      "because they were exported with an older version of Electrum."), ' ',
+                    _("Please import new backups, exported by the wallet these channels belong to."), '\n\n',
+                    _("If you have lost access to that wallet, please open an issue on GitHub."),
+                ])))
+        return warnings
```
SomberNight, line 1166: wdyt about https://github.com/SomberNight/electrum/commit/0587b48f1f7c28a2c66acc4a8dadeb5f651b3ba2,
to remove the old log warning?

Turns out I still have testnet wallets with v0 backups imported :DDD

SomberNight: maybe also
```diff
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index 103cbb21c5..f584a85202 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1158,7 +1158,7 @@ class LNWallet(Logger):
                 title=_('Unusable channel backups') + ' [gh-10852-2]',
                 show_once=False,  # show on every startup
                 message=''.join([
-                    _("This wallet contains channel backups that cannot be used to request a force close, "
+                    _("This wallet contains old (v2) channel backups that cannot be used to request a force close, "
                       "because they were exported with an older version of Electrum."), ' ',
                     _("Please import new backups, exported by the wallet these channels belong to."), '\n\n',
                     _("If you have lost access to that wallet, please open an issue on GitHub."),

```

f321x: Yeah makes sense, the log won't be seen by gui users anyway. Integrated your commit into the branch: https://github.com/spesmilo/electrum/compare/b0fde53043a4ba93abe3aeffb9aca2ac346ded83..409d9ba16f67bf46d5ce61de39a5f21338ffb892

SomberNight:
ACK 409d9ba16f67bf46d5ce61de39a5f21338ffb892

PR #10853, opened 2026-08-14
swaps: handle negative PoW nonce gracefully

f321x:
Handle negative nonces in swap announcement events gracefully.


SomberNight (approved):
Nice catch. Looks good to me.

PR #10855, opened 2026-08-14
ci: lint QML files with qmllint

### LINKED ISSUE #10802: could we use qmllint (or similar) as linter for the qml gui?
Would be great if we had a linter that caught trivial-typos/syntax-errors/etc, in the qml code.
Something that we could run on the CI.

-----

What about [`qmllint`](https://doc.qt.io/qt-6/qtqml-tooling-qmllint.html)?

Clearly it is more complicated than just invoking that from the CLI:
```
$ /usr/lib/qt6/bin/qmllint electrum/gui/qml/components/About.qml 
Warning: electrum/gui/qml/components/About.qml:40:23: Unqualified access [unqualified]
                text: BUILD.electrum_version
                      ^^^^^
Warning: electrum/gui/qml/components/About.qml:77:55: Unqualified access [unqualified]
                onLinkActivated: Qt.openUrlExternally(link)
                                                      ^^^^
Warning: electrum/gui/qml/components/About.qml:88:25: Unqualified access [unqualified]
                height: constants.paddingXLarge
                        ^^^^^^^^^
```

ref https://doc.qt.io/qtforpython-6/tools/pyside-qmllint.html  (re pyside, not pyqt)

Alwoch:
Closes #10802

This PR creates a `qml-lint` CI job that runs qmllint over all tracked `*.qml` files (qml gui and plugins). The project currently uses `PyQt6` which doesn't ship `qmllint`, so in this PR `qmllint` is taken from the PySide6 wheels pinned in the workflow. `PySide6-Addons` is also needed for the QtMultimedia types used by QRScan.qml. It is a lint-only dependency so nothing in electrum imports it.

As anticipated in the issue, the default behaviour is not usable as a gate because a plain `qmllint` over the tree currently emits approximately 3400 warnings, nearly all from three structural sources that static analysis cannot see: context properties injected at runtime (qeapp.py), dynamically scoped root-window properties (`constants`/`app` in main.qml), and the `org.electrum` types being registered from Python. To address this:

- `contrib/qml_lint/run_qml_lint.py` generates a minimal qmldir+qmltypes stub for `org.electrum` from the `qmlRegisterType` calls in qeapp.py (regenerated on every run, so it cannot drift) and passes it to qmllint via `-I`. With the module resolvable, a misspelled type name fails the lint instead of producing unresolved-import noise.
- `.qmllint.ini` at the repo root disables six warning categories; each carries a comment with the measured warning count and what it would take to enable it. Everything else gates: syntax errors, unknown types, bad imports, and the 56 remaining default-enabled categories the tree already passes (duplicate bindings, writes to read-only properties, use-before-declaration, deprecations, ...).

```
$ python3 -m venv /tmp/qml-lint-venv && /tmp/qml-lint-venv/bin/pip install "PySide6-Essentials==6.11.1" "PySide6-Addons==6.11.1"
$ /tmp/qml-lint-venv/bin/python contrib/qml_lint/run_qml_lint.py
[... 96 info-level unused-import notices, which do not gate ...]
qmllint: checked 131 files (29 org.electrum stub types), exit code 0

# after changing one "Label" to "Labell" in About.qml:
$ /tmp/qml-lint-venv/bin/python contrib/qml_lint/run_qml_lint.py
Warning: electrum/gui/qml/components/About.qml:35:13: Labell was not found. Did you add all imports and dependencies? [import]
            Labell {
            ^^^^^^
Info: Did you mean "Label"?
qmllint: checked 131 files (29 org.electrum stub types), exit code 255
```

**Known gaps:**

Property-name typos and misspelled identifiers inside bindings are not caught yet because they live in the disabled MissingProperty/UnqualifiedAccess categories. The disabled categories are intended as a ratchet, not a permanent state; this PR is deliberately infra-only, and I plan to follow up in roughly this order:

1. Fix the 32 sites of Qt-documented undefined behavior that `Quick.LayoutsPositioning` flags (raw width/height on items managed by a Layout, e.g. About.qml:87 -- the file the issue quotes), then enable the category.
2. Modernize the 43 legacy `PropertyChanges` sites, then enable `Quick.PropertyChangesParsed`.
3. Extend the stub generator to introspect the Python bridge classes (pyqtProperty/signals/slots) so the org.electrum types carry real property lists, then enable `MissingProperty` (catching property-name typos) and `UnresolvedAlias` (1 site).
4. Longer term, and only if the maintainers approve: refactor Constants into a QML singleton and migrate the context properties toward singletons (the direction Qt recommends), which is what it would take to ever enable `UnqualifiedAccess` (~3000 warnings today: ~800 `constants`, ~790 context properties, ~260 `app`, rest delegate scope). Unlike the items above, this changes runtime code, so it needs discussion first.

The 96 unused-import notices the job prints are also a candidate for a small cleanup PR; they are info-level and do not gate. The one remaining disable, `EqualityTypeCoercion`, is intentional and permanent: its single hit relies on url/string coercion, and switching it to `!==` would change behavior.

**Defaults I picked that are easy to change:** 

The linter is pinned to 6.11.1 (newest) while the android gui ships Qt 6.10.2. I can test and repin to 6.10.x if matching the runtime Qt is preferred. `.qmllint.ini` sits at the repo root so it also covers the plugin qml files; it could instead live under electrum/gui/qml/, at the cost of not covering plugins. And plugin qml files are in scope, since they currently pass clean.

Let me know if this is along the lines of what you had in mind @SomberNight .

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -72,6 +72,11 @@ jobs:
         # (contrib/android/p4a_recipes/), so lint semantics can be newer than runtime
         run: pip install "PySide6-Essentials==6.11.1" "PySide6-Addons==6.11.1"
       - name: Run qmllint
+        # advisory while this job is new: its runtime deps are not exercised
+        # anywhere else in CI, so a problem with the job itself should not
+        # block unrelated PRs. The known warning backlogs are info-level in
+        # .qmllint.ini, so this can be dropped without them becoming fatal.
+        continue-on-error: true
         run: ./contrib/qml_lint/run_qml_lint.py
 
   # unittests using the 'latest' runtime python-dependencies
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,6 +53,27 @@ jobs:
       - name: Run ban_unicode
         run: ./contrib/ban_unicode.py
 
+  qml-lint:
+    name: "linter: qmllint"
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+        with:
+          python-version: "3.10"
+          cache: 'pip'
+          # key the pip cache on this file so the PySide6 wheels (~255MB) are
+          # actually cached (and invalidated when the pin below changes)
+          cache-dependency-path: .github/workflows/tests.yml
+      - name: Install qmllint (from PySide6; lint-only dependency, never shipped)
+        # newest qmllint at time of writing; NB the Android gui ships Qt 6.10.2
+        # (contrib/android/p4a_recipes/), so lint semantics can be newer than runtime
+        run: pip install "PySide6-Essentials==6.11.1" "PySide6-Addons==6.11.1"
+      - name: Run qmllint
+        run: ./contrib/qml_lint/run_qml_lint.py
+
   # unittests using the 'latest' runtime python-dependencies
   unittests:
     name: "unittests: py${{ matrix.python }}${{ matrix.debug && ', debug-mode' || '' }}"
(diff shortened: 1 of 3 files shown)


accumulator:
@Alwoch thank you for this, it looks quite comprehensive.

Maybe we could for now mark the job/step as `continue-on-error` so it doesn't fail the tests, and incrementally enable the currently disabled categories in `.qmllint.ini`, so we have a list of improvements to work towards.

PR #10856, opened 2026-08-14
swaps: prioritize confirmed funding utxos over unconfirmed ones

f321x:
When picking the funding utxo to look at in `SwapManager._claim_swap()`, prefer picking one that is already confirmed.


SomberNight (approved):
ACK febb381baa5eab952fe8c99611a55443ec6a914d,  if tests pass

krynerovamonika827-commits (commented):
6a94e460b50bc5afc334ca0d6feead47d3b50539f

PR #10859, opened 2026-08-17
util, swaps: fix nostr announcement PoW mining

### LINKED ISSUE #10858: swapserver nostr announcement PoW miner: worker returns the builtin hash, gen_nostr_ann_pow can discard the winning nonce, result stored unvalidated
**Summary**

The nostr announcement PoW miner of the swapserver plugin has three defects, introduced in `947094c1b` (PR #9551, "Change nostr swap event type, add PoW, rebroadcast events", merged 2025-02-12) and present — byte-identical — in every release since (4.6.0 → 4.8.1) and in current master.

Reproduction is deterministic via the test suite in the fix branch linked at the bottom: 5 of 7 tests fail on unpatched master, all 7 pass with the fix. The tests run no PoW searches (precomputed nonces, faked multiprocessing), so they are CI-safe.

**The three defects**

1. `nostr_pow_worker` returns the Python builtin instead of the digest — `electrum/util.py` (~line 2417 at master):

```python
            if int.from_bytes(digest, 'big') < (1 << (hash_len_bits - target_bits)):
                shutdown.set()
                return hash, nonce      # <-- `hash` is the builtin, `digest` was meant
```

The winning worker returns `(builtin_hash_function, nonce)`. Latent today (the digest is unused by the caller) but a landmine for any future consumer, and clear evidence the line was never exercised by a test.

2. `gen_nostr_ann_pow` can discard the freshly-mined winning nonce — `electrum/util.py` (~line 2453):

```python
        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
        hash_res, nonce_res = done.pop().result()
```

`done` is a **set**; `pop()` returns an arbitrary element. Non-winning workers return `(None, None)` at their next 1M-hash block boundary after `shutdown` is set — and that result can be delivered to the event loop **before** the winner's own result: the two travel over independent channels (the multiprocessing-manager socket for the event vs the executor's result pipe + `call_soon_threadsafe` for the return value), so there is no ordering guarantee. When an observer's future lands in `done` first (or in the same wakeup), `pop()` can yield `(None, None)` and the winner's nonce is silently thrown away. Reproduced deterministically in `test_observer_result_completing_first_is_ignored` with faithful fake workers.

3. `set_nostr_proof_of_work` stores the result without validating it — `electrum/submarine_swaps.py` (~line 448):

```python
        nonce, pow_amount = await gen_nostr_ann_pow(...)
        self.logger.debug(f"Found {pow_amount} bits of work for Nostr announcement.")
        self.config.SWAPSERVER_ANN_POW_NONCE = nonce   # no check pow_amount >= target
```

Note `ConfigVar._set_config_value` skips the type check when the value is `None`, so the race outcome of defect 2 is persisted as-is.

**Impact chain (defects 2+3 combined)**

~~Race hits → `gen_nostr_ann_pow` returns `(None, 0)` → `None` stored in `plugins.swapserver.ann_pow_nonce` → the next `publish_offer` executes `hex(sm.config.SWAPSERVER_ANN_POW_NONCE)` → `TypeError: 'NoneType' object cannot be interpreted as an integer` **outside** the try/except around `aionostr._add_event` → `run_nostr_server` dies → the swapserver stops announcing and rebroadcasting over nostr entirely until restart — and on restart re-mines from scratch (the reuse check sees `0` bits), discarding the previous run's work.~~

**Correction (2026-08):** the struck-through chain above is wrong. Assigning `None` to a ConfigVar deletes the key (`_set_key_in_user_config`), the next read falls back to the default `0`, and `hex(0)` works fine — nothing crashes. The real consequence is quieter: the announcement is published with a 0-bit nonce and every client silently discards it (`too low pow`) — the swapserver looks healthy in its own logs but is undiscoverable over nostr. Thanks to @f321x for the REPL check that caught this.

Separately, any below-target nonce that reaches the store (e.g. a manually pinned bad value, or a future miner regression) is published without warning while every electrum client silently discards the announcement (`too low pow`, submarine_swaps.py ~2085) — the server logs `published offer <id>` on schedule and looks healthy, but is undiscoverable.

We first noticed this while running a signet swapserver that published an announcement with a far-below-target nonce; we could not deterministically attribute that specific nonce to these defects (our checkout carries local patches), which is why this report stands on the deterministic reproduction below instead of the live anecdote.

**Reproduce / A/B**

```
git clone https://github.com/Amperstrand/electrum -b nostr-pow-bug-repro
cd electrum
python -m pytest tests/test_nostr_pow.py -v   # 5 failed, 2 passed (+2 skipped)
git checkout nostr-pow-fix
python -m pytest tests/test_nostr_pow.py -v   # 7 passed, 2 skipped
```

The two skipped tests (`TestNostrPowSearch`) perform actual PoW searches and are deliberately gated off CI — run locally with `ELECTRUM_TESTS_NOSTR_POW=1 python -m pytest tests/test_nostr_pow.py -v` (9/9 pass with the fix, ~2 s at a 16-bit target).

Race-frequency note (honest): in 150 consecutive real-pool mining runs on unpatched master (target 20 bits) the race did not fire — the winner's result normally leads the observers' by a full 1M-hash block (~1 s), so the defect is real but rare per mining run. The deterministic unit test pins the worst-case interleaving regardless.

**Fix**

Branch `Amperstrand/electrum:nostr-pow-fix` (2 commits on master `a94e460b5`, +195/−3):

1. `nostr_pow_worker`: `return digest, nonce`
2. `gen_nostr_ann_pow`: wait for all workers (`ALL_COMPLETED`) and collect the result that carries a nonce; raise if none does. The winner's result is the only one guaranteed to carry a nonce, so waiting ≤1 block for the observers to return is the simplest correct collection strategy.
3. `set_nostr_proof_of_work`: raise when `pow_amount < SWAPSERVER_POW_TARGET` instead of storing.

Adjacent suites `tests/test_util.py` + `tests/test_submarine_swaps.py` remain green (64/64). Happy to open a PR from that branch if the approach looks acceptable.

Amperstrand:
Fixes #10858

## Summary

Three defects in the swapserver nostr announcement PoW path, all introduced in `947094c1b` (PR #9551):

1. `nostr_pow_worker` returned the builtin `hash` instead of the sha256 `digest` of the winning preimage.
2. `gen_nostr_ann_pow` collected the result via `done.pop()` on the unordered set returned by `asyncio.wait(FIRST_COMPLETED)`. Workers that observe the shutdown event return early at their next 1M-hash block boundary, and their results can be delivered to the event loop before the winner's own result (independent channels: manager socket vs executor result pipe), so the freshly-mined winning nonce could be silently discarded.
3. `set_nostr_proof_of_work` stored the returned nonce without checking `pow_amount >= SWAPSERVER_POW_TARGET`.

## Changes (electrum/util.py + electrum/submarine_swaps.py)

- `nostr_pow_worker`: return only the winning `nonce` — the digest was never used by the caller (it is recomputed via `get_nostr_ann_pow_amount`), so the return value is removed entirely
- `gen_nostr_ann_pow`: wait for all workers (`ALL_COMPLETED`), collect the first result carrying a nonce, raise if none does
- `set_nostr_proof_of_work`: assert the achieved work meets `SWAPSERVER_POW_TARGET` before storing the nonce

## Impact

If the race in (2) hits, the winning nonce is discarded and the stored nonce reads as the default `0`: the announcement is then published with insufficient PoW and every client silently discards it (`too low pow`) — the swapserver looks healthy in its own logs but is undiscoverable over nostr until it re-mines.

~~An earlier version of this description claimed the race outcome would be persisted as `None`, and that the next `publish_offer` would then raise `TypeError: hex(None)` outside its try/except, killing `run_nostr_server`.~~

**Correction:** that chain was wrong — assigning `None` to a ConfigVar deletes the key, the read falls back to the default `0`, and `hex(0)` works fine, so nothing crashes. Thanks to @f321x for the REPL check that caught this.

## Testing

- Adjacent suites: `python -m pytest tests/test_util.py tests/test_submarine_swaps.py` — all green
- The deterministic repro suite from #10858 stays on the `nostr-pow-bug-repro` branch (per review, the in-tree module was overkill for this fix size)

### DIFF
--- a/electrum/submarine_swaps.py
+++ b/electrum/submarine_swaps.py
@@ -450,6 +450,7 @@ class SwapManager(Logger):
             self.config.SWAPSERVER_POW_TARGET,
         )
         self.logger.debug(f"Found {pow_amount} bits of work for Nostr announcement.")
+        assert pow_amount >= self.config.SWAPSERVER_POW_TARGET, pow_amount
         self.config.SWAPSERVER_ANN_POW_NONCE = nonce
 
     async def pay_invoice(self, key):
--- a/electrum/util.py
+++ b/electrum/util.py
@@ -2420,10 +2420,10 @@ def nostr_pow_worker(nonce, nostr_pubk, target_bits, hash_function, hash_len_bit
             digest = hash_function(hash_preimage + nonce.to_bytes(32, 'big')).digest()
             if int.from_bytes(digest, 'big') < (1 << (hash_len_bits - target_bits)):
                 shutdown.set()
-                return hash, nonce
+                return nonce
             nonce += 1
         if shutdown.is_set():
-            return None, None
+            return None
 
 
 async def gen_nostr_ann_pow(nostr_pubk: bytes, target_bits: int) -> Tuple[int, int]:
@@ -2456,10 +2456,15 @@ async def gen_nostr_ann_pow(nostr_pubk: bytes, target_bits: int) -> Tuple[int, i
             if start_nonce > max_nonce:  # make sure we don't go over the max_nonce
                 start_nonce = random.randint(0, int(max_nonce * 0.75))
 
-        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
-        hash_res, nonce_res = done.pop().result()
+        # workers that observe the shutdown event return None; their results can be
+        # delivered before the winner's own result, so wait for all workers and
+        # collect the result that carries a nonce.
+        done, _pending = await asyncio.wait(tasks, return_when=asyncio.ALL_COMPLETED)
+        nonce_res = next((n for n in (fut.result() for fut in done) if n is not None), None)
         executor.shutdown(wait=False, cancel_futures=True)
 
+    if nonce_res is None:
+        raise Exception("nostr announcement PoW mining failed: no worker returned a nonce")
     return nonce_res, get_nostr_ann_pow_amount(nostr_pubk, nonce_res)
 
 


Amperstrand:
Full A/B logs for the record: https://gist.github.com/Amperstrand/4b477e0fbc14b22f8601c601f964b210

- unpatched master `a94e460b5` + tests commit: **5 failed, 2 passed, 2 skipped** (deterministic across 3 runs)
- this branch: **7 passed, 2 skipped** (across 5 runs); full default suite `pytest tests --ignore=tests/regtest --ignore=tests/qml` locally: **1024 passed, 8 skipped, 275 subtests, 0 failures** (~4 min)
- the 2 skipped tests (`TestNostrPowSearch`) run real process-pool searches and are intentionally gated off CI; locally with `ELECTRUM_TESTS_NOSTR_POW=1` they pass 9/9 (~2 s at a 16-bit target). Also ran the real pool directly at target-20 bits: 30/30 valid nonces with the patch.

If a different collection strategy is preferred (e.g. return as soon as any nonce-bearing result completes, instead of waiting for all workers — slightly more code, but zero added startup latency), happy to rework.

tests/test_nostr_pow.py
```diff
@@ -0,0 +1,197 @@
+"""Tests for the nostr announcement PoW of the swapserver plugin.
```
f321x: Adding this whole new test module for the simple pow functionality feels a bit overkill, I'd rather drop it

electrum/submarine_swaps.py
```diff
@@ -450,6 +450,10 @@ async def set_nostr_proof_of_work(self) -> None:
             self.config.SWAPSERVER_POW_TARGET,
         )
         self.logger.debug(f"Found {pow_amount} bits of work for Nostr announcement.")
+        if pow_amount < self.config.SWAPSERVER_POW_TARGET:
```
f321x:
```suggestion
        assert pow_amount >= self.config.SWAPSERVER_POW_TARGET, pow_amount
```

Considering this shouldn't happen I think an assert would make this more clear here instead of making the reader reason about how this could happen

electrum/util.py
```diff
@@ -2414,7 +2414,7 @@ def nostr_pow_worker(nonce, nostr_pubk, target_bits, hash_function, hash_len_bit
             digest = hash_function(hash_preimage + nonce.to_bytes(32, 'big')).digest()
             if int.from_bytes(digest, 'big') < (1 << (hash_len_bits - target_bits)):
                 shutdown.set()
-                return hash, nonce
+                return digest, nonce
```
f321x: considering the `hash`/`digest` return value is never used, it is probably best to remove it completely, making the result access in `gen_nostr_ann_pow` more readable too

f321x:
> ConfigVar._set_config_value skips its type-check for None, so the race outcome was persisted — the next publish_offer then raised TypeError: hex(None)

This is hallucinated, see:

```
>>> config.SWAPSERVER_ANN_POW_NONCE = None
>>> config.SWAPSERVER_ANN_POW_NONCE
0
>>> hex(config.SWAPSERVER_ANN_POW_NONCE)
'0x0'
```

However still, good catch, it could cause a swap provider to announce itself without PoW, making it invisible to clients

f321x:
Thanks!

PR #10860, opened 2026-08-17
tests: make new swap tests less flaky

f321x:
Prevent assertion failures through early return in `_claim_swap` by ensuring the wallet is synced when we call `_claim_swap`.

### DIFF
--- a/tests/test_submarine_swaps.py
+++ b/tests/test_submarine_swaps.py
@@ -109,10 +109,15 @@ class TestSwapClaim(ElectrumTestCase):
 
     async def sync(self) -> None:
         """wait until the wallet has caught up with the server"""
-        await self.wait_until(
-            lambda: self.adb.get_local_height() == self.server.cur_height
-            and self.adb.is_up_to_date()
-            and self.wallet.is_up_to_date())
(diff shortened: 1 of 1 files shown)


tests/test_submarine_swaps.py
```diff
@@ -109,10 +109,15 @@ async def wait_until(self, predicate, *, timeout: int = 10) -> None:
...
-            and self.wallet.is_up_to_date())
+        def is_synced() -> bool:
+            return self.adb.get_local_height() == self.server.cur_height \
+                    and self.adb.is_up_to_date() \
+                    and self.wallet.is_up_to_date()
+        while True:
+            await self.wait_until(is_synced)
+            # the adb_set_up_to_date callback might unsync the wallet again when syncing new addresses
+            if self.wallet.synchronize() == 0:
+                return
```
SomberNight, line 120: just noting that instead of time-based polling, this Event exists:
https://github.com/spesmilo/electrum/blob/61e5b26f6bef9b31b738ca95976fb26568ad0d7e/electrum/wallet.py#L462

SomberNight (approved):
looks ok

PR #10861, opened 2026-08-17
lnpeer: drop support for legacy closing fee negotiation

ecdsa:
Note that BOLT2 used to bound fee_satoshis by the base fee of the final ctx, but that requirement was removed in lightning/bolts#847.

Also fix the "strictly between" check, which was skipped when their previously proposed fee was 0.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -53,7 +53,9 @@ from .lntransport import LNTransport, LNTransportBase, LightningPeerConnectionCl
 from .lnmsg import encode_msg, decode_msg, UnknownOptionalMsgType, FailedToParseMsg
 from .interface import GracefulDisconnect
 from .invoices import PR_PAID
-from .fee_policy import FEE_LN_ETA_TARGET, FEERATE_PER_KW_MIN_RELAY_LIGHTNING, FEERATE_MAX_DYNAMIC
+from .fee_policy import (
+    FEE_LN_ETA_TARGET, FEERATE_PER_KW_MIN_RELAY_LIGHTNING, FEERATE_MAX_DYNAMIC, FEE_LN_MINIMUM_ETA_TARGET, FEERATE_DEFAULT_RELAY,
+)
 from .channel_db import FLAG_DIRECTION
 
 if TYPE_CHECKING:
@@ -64,6 +66,9 @@ if TYPE_CHECKING:
 
 
 LN_P2P_NETWORK_TIMEOUT = 20
+
+
+class CoopCloseFailure(Exception): pass
 
 
 class Peer(Logger, EventListener):
@@ -2622,8 +2627,9 @@ class Peer(Logger, EventListener):
         # Note: A malicious Electrum server can make us pay high closing fees, capped only by FEERATE_MAX_DYNAMIC
         # The same issue exists with commitment transactions; we only make sure it is not worse here.
         config = self.config
+        is_initiator = chan.constraints.is_initiator
         our_fee = None
-        if config.TEST_SHUTDOWN_FEE:
+        if config.TEST_SHUTDOWN_FEE is not None:
             our_fee = config.TEST_SHUTDOWN_FEE
         else:
             fee_rate_per_kb = self.network.fee_estimates.eta_target_to_fee(FEE_LN_ETA_TARGET)
@@ -2637,19 +2643,29 @@ class Peer(Logger, EventListener):
             our_fee = fee_rate_per_kb * closing_tx.estimated_size() // 1000
         # max value
         max_fee = min(our_fee * 2, FEERATE_MAX_DYNAMIC * closing_tx.estimated_size() // 1000)
-        our_fee = min(our_fee, max_fee)
         # make sure fee is payable by initiator
-        affordable = chan.balance(LOCAL if chan.constraints.is_initiator else REMOTE) // 1000
-        our_fee = min(our_fee, affordable)
+        affordable = chan.balance(LOCAL if is_initiator else REMOTE) // 1000
         max_fee = min(max_fee, affordable)
         # min value. We aim at a fee between next block inclusion and some lower value.
-        min_fee = our_fee // 2
+        fee_rate_per_kb = self.network.fee_estimates.eta_target_to_fee(FEE_LN_MINIMUM_ETA_TARGET) or FEERATE_DEFAULT_RELAY
+        superlow_min_fee = fee_rate_per_kb * closing_tx.estimated_size() // 1000
+        if is_initiator:
+            min_fee = our_fee // 2
+            min_fee = max(min_fee, superlow_min_fee)
+        else:
+            # The sending node, if it is not the funder:
+            # SHOULD set min_fee_satoshis to a fairly low value
+            min_fee = superlow_min_fee
+        # ensure order
+        min_fee = min(min_fee, max_fee)  # note: if min_fee was > max_fee, the tx might not relay... unclear what to do.
+        our_fee = max(min_fee, our_fee)
+        our_fee = min(our_fee, max_fee)
+        assert min_fee <= our_fee <= max_fee
         if config.TEST_SHUTDOWN_FEE_RANGE:
             our_fee_range = config.TEST_SHUTDOWN_FEE_RANGE
         else:
             our_fee_range = {'min_fee_satoshis': min_fee, 'max_fee_satoshis': max_fee}
         self.logger.info(f"Our fee range: {our_fee_range} and fee: {our_fee}")
-        assert our_fee_range['min_fee_satoshis'] <= our_fee <= our_fee_range['max_fee_satoshis']
         return our_fee, our_fee_range
 
     @log_exceptions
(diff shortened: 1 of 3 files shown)


electrum/lnpeer.py
```diff
@@ -2667,7 +2672,8 @@ async def _shutdown(self, chan: Channel, payload, *, is_local: bool):
         our_sig = None  # type: Optional[bytes]
         closing_tx = None  # type: Optional[PartialTransaction]
         is_initiator = chan.constraints.is_initiator
-        our_fee, our_fee_range = self.get_shutdown_fee_range(chan, dummy_tx, is_local)
+        our_fee, min_fee, max_fee = self.get_shutdown_fee_range(chan, dummy_tx, is_local)
+        our_fee_range = {'min_fee_satoshis': min_fee, 'max_fee_satoshis': max_fee} if min_fee is not None else None
 
         def send_closing_signed(our_fee, our_fee_range, drop_remote):
             nonlocal our_sig, closing_tx
```
SomberNight, line 2693:
```suggestion
            nonlocal our_sig, closing_tx
            if is_initiator:
                assert our_fee <= max_fee, f"{our_fee=}, {max_fee=}"
```

electrum/lnpeer.py
```diff
@@ -2617,36 +2617,41 @@ async def send_shutdown(self, chan: Channel):
...
+                our_fee = chan.get_latest_fee(LOCAL if is_local else REMOTE)
+        # max value
+        max_fee = min(our_fee * 2, FEERATE_MAX_DYNAMIC * closing_tx.estimated_size() // 1000)
+        our_fee = min(our_fee, max_fee)
+        # make sure fee is payable by initiator
+        affordable = chan.balance(LOCAL if chan.constraints.is_initiator else REMOTE) // 1000
+        our_fee = min(our_fee, affordable)
+        max_fee = min(max_fee, affordable)
+        # min value. None for legacy negotiation.
+        min_fee = our_fee // 2 if not config.TEST_SHUTDOWN_LEGACY else None
```
SomberNight: should we check that `min_fee` is above the relay fee?
e.g. we could set it to the 25 block ETA

electrum/lnpeer.py
```diff
@@ -2617,36 +2617,41 @@ async def send_shutdown(self, chan: Channel):
...
-        if config.TEST_SHUTDOWN_LEGACY:
-            our_fee_range = None
-        elif config.TEST_SHUTDOWN_FEE_RANGE:
-            our_fee_range = config.TEST_SHUTDOWN_FEE_RANGE
-        else:
-            # we aim at a fee between next block inclusion and some lower value
-            our_fee_range = {'min_fee_satoshis': our_fee // 2, 'max_fee_satoshis': our_fee * 2}
-        self.logger.info(f"Our fee range: {our_fee_range} and fee: {our_fee}")
-        return our_fee, our_fee_range
+                our_fee = chan.get_latest_fee(LOCAL if is_local else REMOTE)
```
SomberNight: That's a weird fallback given that the weight of the closing tx is likely different (smaller?) than the weight of the ctx.

Perhaps keep the feerate instead?  Not sure though.
```diff
diff --git a/electrum/lnpeer.py b/electrum/lnpeer.py
index 68a7a9d389..1c1186237a 100644
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2633,11 +2633,11 @@ class Peer(Logger, EventListener):
             if fee_rate_per_kb is None:  # fallback
                 from .fee_policy import FeePolicy
                 fee_rate_per_kb = FeePolicy(config.FEE_POLICY).fee_per_kb(self.network)
-            if fee_rate_per_kb is not None:
-                our_fee = fee_rate_per_kb * closing_tx.estimated_size() // 1000
-            if our_fee is None:  # fallback
+            if fee_rate_per_kb is None:  # fallback
                 self.logger.warning(f"got no fee estimates for co-op close! falling back to chan.get_latest_fee")
-                our_fee = chan.get_latest_fee(LOCAL if is_local else REMOTE)
+                feerate_per_kw = chan.get_latest_feerate(LOCAL if is_local else REMOTE)
+                fee_rate_per_kb = 4 * feerate_per_kw
+            our_fee = fee_rate_per_kb * closing_tx.estimated_size() // 1000
         # max value
         max_fee = min(our_fee * 2, FEERATE_MAX_DYNAMIC * closing_tx.estimated_size() // 1000)
         our_fee = min(our_fee, max_fee)
```

electrum/lnpeer.py
SomberNight: If we are serious about `min_fee`, we could compare against that as well.

electrum/lnpeer.py
```diff
@@ -2701,6 +2707,10 @@ async def receive_closing_signed():
             their_fee = cs_payload['fee_satoshis']
             their_fee_range = cs_payload['closing_signed_tlvs'].get('fee_range')
```
SomberNight, line 2719: These `closing_signed_tlvs` and `fee_range` fields were added to the bolts [5 years ago](https://github.com/lightning/bolts/pull/847). I think it would be reasonable to assume they are always set by now. Otherwise force-closing is an okay fallback.
Hopefully removing the legacy case would simplify this code somewhat, as it is quite complicated as-is.

In fact, btw, there is an even-newer way to do mutual close since then, that we haven't yet implemented: [`option_simple_close`](https://github.com/lightning/bolts/pull/1205).

ecdsa: I agree, I would not mind removing the legacy pathway.

electrum/lnpeer.py
ecdsa: sure, why not

electrum/lnpeer.py
ecdsa: done

electrum/lnpeer.py

electrum/lnpeer.py
```diff
@@ -2617,35 +2622,50 @@ async def send_shutdown(self, chan: Channel):
...
+            # SHOULD set min_fee_satoshis to a fairly low value
+            min_fee = superlow_min_fee
+        if config.TEST_SHUTDOWN_FEE_RANGE:
             our_fee_range = config.TEST_SHUTDOWN_FEE_RANGE
         else:
-            # we aim at a fee between next block inclusion and some lower value
-            our_fee_range = {'min_fee_satoshis': our_fee // 2, 'max_fee_satoshis': our_fee * 2}
+            our_fee_range = {'min_fee_satoshis': min_fee, 'max_fee_satoshis': max_fee}
         self.logger.info(f"Our fee range: {our_fee_range} and fee: {our_fee}")
+        if not (our_fee_range['min_fee_satoshis'] <= our_fee <= our_fee_range['max_fee_satoshis']):
```
ecdsa: I think this should be an assert, as in, a property of this method.
how about the following:

```diff --git a/electrum/lnpeer.py b/electrum/lnpeer.py
index ac2331e3c..f23cc2d5b 100644
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2643,10 +2643,8 @@ class Peer(Logger, EventListener):
             our_fee = fee_rate_per_kb * closing_tx.estimated_size() // 1000
         # max value
         max_fee = min(our_fee * 2, FEERATE_MAX_DYNAMIC * closing_tx.estimated_size() // 1000)
-        our_fee = min(our_fee, max_fee)
         # make sure fee is payable by initiator
         affordable = chan.balance(LOCAL if is_initiator else REMOTE) // 1000
-        our_fee = min(our_fee, affordable)
         max_fee = min(max_fee, affordable)
         # min value. We aim at a fee between next block inclusion and some lower value.
         fee_rate_per_kb = self.network.fee_estimates.eta_target_to_fee(FEE_LN_MINIMUM_ETA_TARGET) or FEERATE_DEFAULT_RELAY
@@ -2658,14 +2656,16 @@ class Peer(Logger, EventListener):
             # The sending node, if it is not the funder:
             # SHOULD set min_fee_satoshis to a fairly low value
             min_fee = superlow_min_fee
+        # ensure order
+        max_fee = max(min_fee, max_fee)
+        our_fee = max(min_fee, our_fee)
+        our_fee = min(our_fee, max_fee)
+        assert min_fee <= our_fee <= max_fee
         if config.TEST_SHUTDOWN_FEE_RANGE:
             our_fee_range = config.TEST_SHUTDOWN_FEE_RANGE
         else:
             our_fee_range = {'min_fee_satoshis': min_fee, 'max_fee_satoshis': max_fee}
         self.logger.info(f"Our fee range: {our_fee_range} and fee: {our_fee}")
-        if not (our_fee_range['min_fee_satoshis'] <= our_fee <= our_fee_range['max_fee_satoshis']):
-            self.schedule_force_closing(chan.channel_id)
-            raise CoopCloseFailure(f"min<max inequality failure for fee range: {our_fee_range=}, {our_fee=}")
         return our_fee, our_fee_range
 

```

SomberNight: applied similar patch in https://github.com/spesmilo/electrum/compare/01fb37b51010d363802eed6143f4a6c2e303c0a5..75bd374a04f4da32d94eb11c0bf1e5b8c5af7919

SomberNight: note: related comments also at https://github.com/spesmilo/electrum/commit/01fb37b51010d363802eed6143f4a6c2e303c0a5#r196793647 :

>> this really should be an assert; the code in this method should guarantee this property.
>
> it seems hard to guarantee as - in theory - the mempoolminfee could increase to so high (or the relayfee reported by the server be so high) that the local chan balance is too low to pay for the fees

ecdsa:
LGTM

PR #10871, opened 2026-08-19
hww: add a deprecation message popup at get_xpub time for bitbox01 and safe-t

accumulator:
popup at `get_xpub()` seems most convenient, since this reports to user at wallet creation time, and at existing wallet pairing time.

### DIFF
--- a/electrum/plugins/digitalbitbox/digitalbitbox.py
+++ b/electrum/plugins/digitalbitbox/digitalbitbox.py
@@ -125,17 +125,27 @@ class DigitalBitbox_Client(HardwareClientBase):
             raise UserFacingException(_('This device does not reveal xpubs corresponding to non-hardened paths'))
 
         reply = self._get_xpub(bip32_path)
-        if reply:
-            xpub = reply['xpub']
-            # Change type of xpub to the requested type. The firmware
-            # only ever returns the mainnet standard type, but it is agnostic
-            # to the type when signing.
-            if xtype != 'standard' or constants.net.TESTNET:
-                node = BIP32Node.from_xkey(xpub, net=constants.BitcoinMainnet)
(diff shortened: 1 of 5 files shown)


SomberNight:
Is the reason this is draft that you cannot test it as you lack the hardware?

If so, say so, and I will try to test it. I have both but have not plugged them in for a long while.

electrum/plugins/digitalbitbox/digitalbitbox.py
```diff
@@ -125,18 +125,28 @@ def get_xpub(self, bip32_path, xtype):
...
+            xpub = node._replace(xtype=xtype).to_xpub()
+
+        deprecation_warning = (
+            "DigitalBitbox (BitBox01) is being deprecated.\n\nIt is no longer supported by the manufacturer.\n"
+            "Future versions of Electrum will no longer be compatible with it.\n\n"
+            "You should move your coins and migrate to a modern hardware device.")
+        _logger.warning(deprecation_warning.replace("\n", " "))
+
+        if self.handler:
+            self.handler.show_warning(deprecation_warning, blocking=True)
```
SomberNight, line 148: maybe only show this once per process lifecycle?
You could declare a global `DEPRECATION_WARNING_SHOWN = False` at the top, and set it to true here.

accumulator:
> Is the reason this is draft that you cannot test it as you lack the hardware?
> 
> If so, say so, and I will try to test it. I have both but have not plugged them in for a long while.

I have a bitbox01, but no safe-t. 
I'll put it on Ready after implementing the global `DEPRECATION_WARNING_SHOWN` you described above

SomberNight (approved):
ACK 4dfe946911e2e80cec7e26d873e283e6908dc6dd

I tested both bitbox01 and safe-t.

Issue #10868, opened 2026-08-18
Request the passphrase through the interface when the device prompts for it.


Perlover:
### Description

If Trezor, specifically the Trezor Safe 5 model, has the option enabled in its settings to always request the passphrase on the device, version 4.8.1 of Electrum behaves differently from older releases. For instance, I previously used version 4.7.*, but now the latest version enforces the passphrase request even after entering it on the device.

Here's how it works. You connect the Trezor device, which is already set to always request the passphrase on the device. After entering the PIN code, the Electrum wallet signals the Trezor, prompting it to request the passphrase on the Trezor screen. After entering this passphrase, the latest version 4.8.1 now presents a dialog asking you to enter the passphrase again, either on the device or directly in Electrum.

If you choose to enter it again on the device, the Trezor device prompts for the passphrase a second time. Thus, when connecting the hardware wallet to Electrum, you have to enter the passphrase twice.

Perlover:
Good afternoon,

today I updated Electrum to 4.8.2 from the AppImage on your website (Linux). During testing, it turned out that the passphrase is still requested twice, but now it happens differently. Previously, you had to enter the passphrase on the device, then a modal window would appear on your computer, offering the choice to enter it on the computer or the device, and you would choose to enter it on the device and enter it again. 

Now, if the settings indicate that the passphrase is always entered on the Trezor device (in Trezor's settings), you have to enter the passphrase once, then a few seconds later enter the passphrase a second time. Meanwhile, the modal window (from Electrum) that says to look at the device still appears as before. So, the issue remains, it just has taken on a different form. :(

Perlover:
I am currently studying the causes of this problem. I think I'll be sending a patch here soon. Claude Caude seems to have discovered a problem. He is currently assembling the Trezor emulator to address this issue using the emulator.

Perlover:
I dug into this and found the root cause. I reproduced it on the official Trezor emulator and verified a patch. Details below.

> **Transparency note:** the analysis, the emulator testing and the patch were prepared with [Claude Code](https://claude.com/claude-code) (Claude Opus 5) at my request. I'm only mentioning it so you know where the text and the code come from, not as a warning about their quality. Everything below was actually run on the official Trezor core emulators, and the logs are shown, so please judge the patch on its merits.

## TL;DR

[...]

PR #10872, opened 2026-08-19
plugin: trezor: fix usage with passphrase

f321x:
Fixes 3 issues when using passphrases with Trezor hww:
1. V1 devices with `passphrase_always_on_device` would
    require the passphrase to be entered twice, see https://github.com/spesmilo/electrum/issues/10868
2. THP devices (Safe 7) with `passphrase_always_on_device` couldn't be unlocked at all
3. The settings dialog would raise after disabling the passphrase (https://github.com/spesmilo/electrum/commit/5796dec1cbd3f61ef76c944192f65d3b3eee75ce)


SomberNight (approved):
Looks reasonable but I only skimmed the diff.
I tested on a model T a bit.

SomberNight:
thx

PR #10874, opened 2026-08-19
plugins: load zip plugins from the bytes whose signature was verified

ecdsa:
see commits. `MemoryZipImporter` and its unit tests were generated by Claude.


electrum/zip_importer.py
```diff
@@ -0,0 +1,193 @@
...
+        return self._root_name + '.' + '.'.join(parts), is_pkg
+
+    def _synthetic_path(self, member: str) -> str:
+        return self._archive_path + '/' + member
+
+    @property
+    def sha256(self) -> bytes:
+        """sha256 of the archive bytes held in memory. This is what a caller
+        must verify against, rather than re-hashing the file on disk."""
+        return self._sha256
```
f321x: This doesn't seem to be used, so it could be dropped with its unittest?

electrum/plugin.py
```diff
@@ -646,7 +666,22 @@ def derive_privkey(pw: str, salt: bytes) -> ECPrivkey:
...
+        # in memory under this name is stale, and was signed for the old one
+        self._drop_zip_importer(name)
+
+    def remove_external_plugin_metadata(self, name: str) -> None:
+        """Unregisters an external plugin that did not end up being installed."""
+        self.external_plugin_metadata.pop(name, None)
+        self._drop_zip_importer(name)
+
     def uninstall(self, name: str):
+        self._drop_zip_importer(name)
```
f321x: `uninstall()` should also disable the plugin, otherwise it will stay in `self.plugins` and its `on_close` callback won't get executed before it gets removed. (the GUI calls `uninstall` directly without first disabling it)
```suggestion
    def uninstall(self, name: str):
        self.disable(name)
        self._drop_zip_importer(name)
```

tests/test_plugin.py
```diff
@@ -0,0 +1,200 @@
...
+        with self.assertRaises(IncorrectPluginHash):
+            plugin = plugins.load_plugin_by_name(PLUGIN_NAME)
+
+    def test_read_file_raises_after_file_replaced(self):
+        make_plugin_zip(self.zip_path, secret='benign')
+        plugins = self._start_plugins()
+        self._authorize(plugins)
+        self._replace_zip_on_disk('evil')
+        with self.assertRaises(IncorrectPluginHash):
+            self.assertEqual(b'benign', plugins.read_file(PLUGIN_NAME, 'icon.txt'))
```
f321x: The `assertEqual` is not checking anything assuming the `read_file` raises an `IncorrectPluginHash`.

tests/test_plugin.py
```diff
@@ -0,0 +1,200 @@
...
+        manifest = plugins.read_manifest(self.zip_path)
+        self.assertEqual(sha256(blob).hex(), manifest['zip_hash_sha256'])
+
+    def test_uninstall_drops_the_cached_archive(self):
+        make_plugin_zip(self.zip_path, secret='benign')
+        plugins = self._start_plugins()
+        self._authorize(plugins)
+        self.assertTrue(plugins.is_authorized(PLUGIN_NAME))
+        plugins.uninstall(PLUGIN_NAME)
+        self.assertFalse(any(isinstance(f, MemoryZipImporter) for f in sys.meta_path))
```
f321x: The plugin is never enabled in this test:
```suggestion
        self._authorize(plugins)
        plugins.enable(PLUGIN_NAME)
        self.assertTrue(plugins.is_authorized(PLUGIN_NAME))
        self.assertTrue(any(isinstance(f, MemoryZipImporter) for f in sys.meta_path))
        plugins.uninstall(PLUGIN_NAME)
        self.assertIsNone(plugins.get(PLUGIN_NAME))
        self.assertFalse(any(isinstance(f, MemoryZipImporter) for f in sys.meta_path))
```

Issue #8624, opened 2023-09-25
add batching transactions (multi-output txs) on android qml


pony-montana:
Hi, following electrum docs, I'm not able to do a batching transaction on android. Is it missing? Would be great to have this feature :)

PR #10875, opened 2026-08-20
qml: fix paying multi-output invoices

sashazykov:
Paying a multi-output ("pay to many") onchain invoice from the QML gui builds a transaction that sends the **entire invoice total to the first output's address**: the first recipient is overpaid, the others receive nothing — and the invoice then remains marked Unpaid, since onchain paid-detection requires every output to be satisfied.

The QML gui assumed onchain invoices have a single output, in two places:

1. `WalletMainView.payOnchain()` passes only `(invoice.address, invoice.amount)` to the finalizer, and `QETxFinalizer.make_tx()` rebuilds `[PartialTxOutput.from_address_and_value(self.address, amount)]` — where `address` is `invoice.get_address()` (the first output) and `amount` is the invoice total.
2. `InvoiceDialog` only displays the first output's address, so there is no way to notice before confirming.

To reproduce: on desktop Qt, Send tab → "Pay to many" with several `address, amount` lines → Save invoice; open the same wallet with `electrum -g qml` (or on Android) and pay the saved invoice from the invoice list. The resulting tx has a single output paying the full total to the first address.

The fix makes the finalizer spend to the invoice outputs as-is when the invoice has more than one output — matching the qt gui, which already pays `invoice.outputs`, and following the direction of the existing TODO in `gui/qt/send_tab.py` ("instead of passing outputs, use an invoice instead"). Single-output invoices keep the existing `(address, amount)` path, since the paid amount can legitimately differ from the invoice output value (zero-amount invoices, lightning invoices paid via their onchain fallback). It keeps `make_unsigned_transaction`'s default `merge_duplicate_outputs=False`, as the QML gui has no UI for `WALLET_MERGE_DUPLICATE_OUTPUTS`; happy to pass the config through like qt's send tab if preferred.

`InvoiceDialog` now lists all outputs with their amounts (a lightweight list rather than `controls/TxOutput`, which needs tx-level model roles and can't render `'!'` values), and the "Max" option is hidden for multi-output invoices, as amountOverride/updateMaxAmount only support a single address. The amount override does not apply to multi-output invoices at all: the pay path spends to the outputs as-is, so a leftover override is cleared on load, the amount editor is not offered, and an amountless multi-output invoice — all zero-value outputs, which the desktop multiline parser permits — cannot be paid. An entered amount can thus never silently diverge from what the tx pays.

`ConfirmTxDialog` needs no changes: it already renders the built tx's outputs, so the confirmation screen shows all recipients before signing.

<details>
<summary>Before / after screenshots (same invoices, desktop QML, testnet)</summary>

3-output invoice, before / after:

invoice-dialog-multi-output-BEFORE invoice-dialog-multi-output

Invoice with a `'!'` line, before / after:

invoice-dialog-max-BEFORE invoice-dialog-max
</details>

Includes tests driving the same python objects the QML flow uses (`QEInvoice`/`QETxFinalizer`), covering: multi-output payment, duplicate-address outputs, a `'!'` max line, weighted-max display values, the zero-amount override path, a lightning invoice paid via its onchain fallback address, and that the finalizer rebuilds its tx when the invoice property is set after the wallet binding has already triggered the first update(). They are async tests on `ElectrumTestCase` (same as `test_qml_qetransactionlistmodel.py`) rather than the existing `QETestCase` in `tests/qml/qt_util.py`, as they need wallet fixtures and asyncio support that harness doesn't provide; happy to move them onto the refactored harness once #10699 lands.

Tested on desktop QML (`-g qml`) on testnet, and on an arm64 Android 16 device with a debug testnet apk carrying this fix:

<details>
<summary>Device screenshots (Android 16)</summary>

device-invoice-dialog-multi-output device-invoice-dialog-max
</details>

Note: this overlaps textually with #10723 (`QETxFinalizer.__init__`/`make_tx` and `WalletMainView.payOnchain`) — happy to rebase on top of it or vice versa, whichever lands first.

The QML gui still cannot *create* multi-output invoices; this only fixes displaying and paying ones that already exist in the wallet. Groundwork for #8624, but stands alone as a bug fix.

### DIFF
--- a/electrum/gui/qml/components/InvoiceDialog.qml
+++ b/electrum/gui/qml/components/InvoiceDialog.qml
@@ -22,7 +22,7 @@ ElDialog {
 
     padding: 0
 
-    property bool _canMax: invoice.invoiceType == Invoice.OnchainInvoice
+    property bool _canMax: invoice.invoiceType == Invoice.OnchainInvoice && invoice.outputs.length == 1
 
     property var _invoice_amount: invoice.amount  // type: Amount
 
@@ -75,7 +75,7 @@ ElDialog {
                 Label {
                     Layout.columnSpan: 2
                     Layout.topMargin: constants.paddingSmall
-                    visible: invoice.invoiceType == Invoice.OnchainInvoice
+                    visible: invoice.invoiceType == Invoice.OnchainInvoice && invoice.outputs.length == 1
                     text: qsTr('Address')
                     color: Material.accentColor
                 }
@@ -83,7 +83,7 @@ ElDialog {
                 DialogHighlightPane {
                     Layout.columnSpan: 2
                     Layout.fillWidth: true
-                    visible: invoice.invoiceType == Invoice.OnchainInvoice
+                    visible: invoice.invoiceType == Invoice.OnchainInvoice && invoice.outputs.length == 1
                     leftPadding: constants.paddingMedium
 
                     RowLayout {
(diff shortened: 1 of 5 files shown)


ecdsa:
@accumulator, could you have a look?

PR #10876, opened 2026-08-20
blockchain: move blockchain logic from module to class

f321x:
Move the blockchain logic in `blockchain.py` from the module level
into a `BlockchainManager` class whose instance is owned by the `Network`.

This separates the blockchain handling from the process and allows
to have different blockchain setups in the same process, which can
e.g. be used in unittests, as those might simulate different
"users" in a single process.


SomberNight (approved):
Thanks. Looks good!
I added some opinionated trivial follow-ups. :)

PR #10814, opened 2026-08-07
reverse swaps: delete invoices when failing

ecdsa:

PR #10877, opened 2026-08-20
swap server: simplify LN invoice payment

f321x:
Replaces #10814

Simplifies the way a swap server pays lightning invoices (when doing a reverse swap/serving a client forward swap). 
On master we have the `pay_pending_invoices` task, however the retries are never exercised as the retry delay is higher than the invoice expiry. 
This made the whole logic unnecessarily complicated and leaked `Invoice` objects in the database when swaps failed.
Now it just one-shot attempts payments and never even persists an `Invoice`.

This uncovered that the `status == PR_INFLIGHT` check in `LNWallet.pay_invoice()` was a no-op which could lead to starting multiple `PaySession` for the same payment hash if there were no htlcs on the channel yet (which would fail the `payment_hash in self.get_payments(status='inflight')` check, and the payment secret of the invoices differed (causing a different payment key).

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1063,7 +1063,7 @@ class LNWallet(Logger):
         self._channel_sending_capacity_lock = asyncio.Lock()
 
         # detect inflight payments
-        self.inflight_payments = set()        # (not persisted) keys of invoices that are in PR_INFLIGHT state
+        self.inflight_payments = set()  # type: set[str]  # (not persisted) keys of invoices that are in PR_INFLIGHT state
         for payment_hash in self.get_payments(status='inflight').keys():
             self.set_invoice_status(payment_hash.hex(), PR_INFLIGHT)
 
@@ -1907,7 +1907,7 @@ class LNWallet(Logger):
         invoice_features = lnaddr.get_features()
         r_tags = lnaddr.get_routing_info('r')
         amount_to_pay = lnaddr.get_amount_msat()
-        status = self.get_payment_status(payment_hash, direction=SENT)
+        status = self.get_invoice_status(invoice)
         if status == PR_PAID:
             raise PaymentFailure(_("This invoice has been paid already"))
         if status == PR_INFLIGHT:
@@ -1931,7 +1931,7 @@ class LNWallet(Logger):
         if attempts is None and self.uses_trampoline():
             # we don't expect lots of failed htlcs with trampoline, so we can fail sooner
             attempts = 30
-        success = False
+        success, reason = False, _("unknown")
         try:
             await self.pay_to_node(
                 node_pubkey=invoice_pubkey,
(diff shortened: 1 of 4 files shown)


electrum/lnworker.py
```diff
@@ -1947,20 +1947,17 @@ async def pay_invoice(
...
-            util.trigger_callback('payment_succeeded', self.wallet, key)
-        else:
-            self.set_invoice_status(key, PR_UNPAID)
-            util.trigger_callback('payment_failed', self.wallet, key, reason)
+            if success:
+                self.set_invoice_status(key, PR_PAID)
+                util.trigger_callback('payment_succeeded', self.wallet, key)
+            else:
+                self.set_invoice_status(key, PR_UNPAID)  # allows retries (unless there are still unresolved htlcs)
+                util.trigger_callback('payment_failed', self.wallet, key, reason)
```
f321x, line 1960: Note, this triggers the callback quite late on shutdown if there are pending htlcs. We could catch `CancelledError` and not trigger the callback then, but not sure if this is worth the added complexity as this doesn't really cause problems in practice.
```
 60.20 | I | network | taskgroup stopped (0x7f3469054740).
 60.20 | I | daemon.Daemon | stopping plugins
 60.20 | I | plugin.Plugins | stopped
 60.20 | I | daemon.Daemon | removing lockfile
 60.20 | I | daemon.Daemon | stopped
 60.21 | I | lnworker.PaySession.[952d2906-4edf] | pay_to_node ending session for RHASH=952d29065a0a13abde81bc7d8470df07611e1a79deff67799c1616085e7508ee
 60.21 | I | lnworker.LNWallet.[default_wallet] | pay_invoice ending session for RHASH=952d29065a0a13abde81bc7d8470df07611e1a79deff67799c1616085e7508ee. success=False
 60.21 | D | gui.qml.qewallet.[default_wallet] | invoice status update for key 952d29065a0a13abde81bc7d8470df07611e1a79deff67799c1616085e7508ee to 0
 60.21 | I | lnworker.LNWallet.[default_wallet] | set_invoice_status 952d29065a0a13abde81bc7d8470df07611e1a79deff67799c1616085e7508ee: 0
(env) user@fedora-2:/var/home/user/code/electrum$ 
```

ecdsa: why did you right-indent this block?

ecdsa:
> This uncovered that the status == PR_INFLIGHT check in LNWallet.pay_invoice() was a no-op

good catch

electrum/submarine_swaps.py
```diff
@@ -346,17 +346,14 @@ async def run_nostr_server(self):
...
+        # nostr and http are not mutually exclusive
         async with self.taskgroup as group:
-            for task in tasks:
-                await group.spawn(task)
+            if self.is_server:
+                if self.config.SWAPSERVER_PORT:
+                    await group.spawn(self.http_server.run())
+                if self.config.NOSTR_RELAYS:
+                    await group.spawn(self.run_nostr_server())
+            await group.spawn(asyncio.Event().wait())  # run until cancel
```
ecdsa, line 356: this looks like pure refactoring...

electrum/lnworker.py
f321x: Otherwise the `Invoice` status won't get set back to `PR_UNPAID` if `pay_to_node` raises some generic `Exception` which now changes behavior due to the fixed check and fails some existing unittests. Without setting it back to `PR_UNPAID` the payment cannot be retried.
Without indent:
```
SUBFAILED[bad path: last node id differs from invoice pubkey] tests/test_lnpeer.py::TestPeerForwardinNoAnchors::test_payment_multihop_with_preselected_path - electrum.lnutil.PaymentFailure: A payment was already initiated for this invoice
SUBFAILED[good path] tests/test_lnpeer.py::TestPeerForwardinNoAnchors::test_payment_multihop_with_preselected_path - electrum.lnutil.PaymentFailure: A payment was already initiated for this invoice
SUBFAILED[bad path: last node id differs from invoice pubkey] tests/test_lnpeer.py::TestPeerForwarding::test_payment_multihop_with_preselected_path - electrum.lnutil.PaymentFailure: A payment was already initiated for this invoice
SUBFAILED[good path] tests/test_lnpeer.py::TestPeerForwarding::test_payment_multihop_with_preselected_path - electrum.lnutil.PaymentFailure: A payment was already initiated for this invoice
```

electrum/submarine_swaps.py
f321x: To some degree it is, because the only unconditional task `tasks = [self.pay_pending_invoices()]` went away i thought it is simpler to express it like this instead of keeping the list logic. 
I can also restore it or move it into a separate commit if you prefer this.

ecdsa: no need this time. 
next time, yes, moving refactoring changes in a separate commit would simplify review.

PR #10880, opened 2026-08-21
bip32: reject extended private key with non-zero key prefix

ekzyis:
BIP-32 mentions [here](https://github.com/bitcoin/bips/blob/7fe0b034ec967b52a5a28276419117326df93263/bip-0032.mediawiki?plain=1#L136) that private key data must start with 0x00:

> ===Serialization format===
>
>Extended public and private keys are serialized as follows:
> * ...
> * 33 bytes: the public key or private key data (ser<sub>P</sub>(K) for public keys, **0x00 || ser<sub>256</sub>(k) for private keys**)

Electrum does currently not check that prefix. [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/bf8402c8803f085a50df96cb7956033cd252e9ab/src/key.cpp#L420) and [embit](https://github.com/diybitcoinhardware/embit/blob/eb6104fd85d3becabba628756cd5e1b75619f3a1/src/embit/bip32.py#L118) do.

_This was found with differential fuzzing and embit in my [fuzz branch](https://github.com/ekzyis/electrum/tree/fuzz)._

### DIFF
--- a/electrum/bip32.py
+++ b/electrum/bip32.py
@@ -155,6 +155,9 @@ class BIP32Node(NamedTuple):
         if not allow_custom_headers and xtype != "standard":
             raise ValueError(f"only standard xpub/xprv allowed. found custom xtype={xtype}")
         if is_private:
+            if xkey[13 + 32] != 0:
+                raise BitcoinException('Invalid extended private key: '
+                                       'key data must be prefixed with 0x00')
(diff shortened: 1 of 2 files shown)


SomberNight (approved):
Good find. Thanks.

ACK ea285d5e57790fa9ee8a4a9192884d9f7ec90d09

PR #10881, opened 2026-08-21
bip32: reject master key with non-zero child/fingerprint

ekzyis:
BIP-32 mentions [here](https://github.com/bitcoin/bips/blob/7fe0b034ec967b52a5a28276419117326df93263/bip-0032.mediawiki?plain=1#L133-L134) that the fingerprint of the parent's key and child number must be four zero bytes for master keys: 

> ===Serialization format===
>
> Extended public and private keys are serialized as follows:
> * ...
> * 4 bytes: the fingerprint of the parent's key (0x00000000 if master key)
> * 4 bytes: child number. This is ser<sub>32</sub>(i) for i in x<sub>i</sub> = x<sub>par</sub>/i, with x<sub>i</sub> the key being serialized. (0x00000000 if master key)

Together with #10880, electrum now covers all invalid BIP-32 test vectors.

_This was found with differential fuzzing and embit in my [fuzz branch](https://github.com/ekzyis/electrum/tree/fuzz)._

### DIFF
--- a/electrum/bip32.py
+++ b/electrum/bip32.py
@@ -154,6 +154,9 @@ class BIP32Node(NamedTuple):
         xtype = headers_inv[header]
         if not allow_custom_headers and xtype != "standard":
             raise ValueError(f"only standard xpub/xprv allowed. found custom xtype={xtype}")
+        if depth == 0 and (child_number != bytes(4) or fingerprint != bytes(4)):
+            raise BitcoinException('Invalid extended key: a depth-0 (master) key must have '
+                                   'zero child number and zero parent fingerprint')
         if is_private:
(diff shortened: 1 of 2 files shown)


SomberNight:
I have force-pushed a rebase: https://github.com/spesmilo/electrum/compare/64ec2bc5ad65ab62ed8ea6d0de6e565017d67c56..afb5c7ade274f5b0d4fa5b4b22fa3e1f864684bd

SomberNight (approved):
Thanks. Nice catch.

ACK afb5c7ade274f5b0d4fa5b4b22fa3e1f864684bd

PR #10882, opened 2026-08-21
logging: get_git_version: add timeout

SomberNight:
note: this is currently only used by the crash reporter, on-demand when constructing a report

### DIFF
--- a/electrum/logging.py
+++ b/electrum/logging.py
@@ -5,6 +5,7 @@
 import logging
 import logging.handlers
 import datetime
+from functools import lru_cache
 import sys
 import pathlib
 import os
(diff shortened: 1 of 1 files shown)


electrum/logging.py
```diff
@@ -356,8 +356,12 @@ def describe_os_version() -> str:
 def get_git_version() -> Optional[str]:
     dir = os.path.dirname(os.path.realpath(__file__))
     try:
+        # note: this searches $PATH and perhaps $PWD for an executable named "git".  attack surface?
         version = subprocess.check_output(
-            ['git', 'describe', '--always', '--dirty'], cwd=dir)
+            ['git', 'describe', '--always', '--dirty'],
+            cwd=dir,
+            timeout=0.1,  # seconds
```
f321x: Some git commands can be really slow on certain filesystems setups, this is for example the output on a virtio9p mount in a vm:
```
user@debian-vm:~/code/electrum$ time git describe --always --dirty
4.8.1-62-gab813173cd

real    0m2.614s
user    0m0.109s
sys     0m0.196s
user@debian-vm:~/code/electrum$ 
```

Not sure if we want to increase the timeout for edge cases like this? But 0.1s might be a bit tight on slow machines. Also `get_git_version` is potentially called twice per report, a `functools.lru_cache` could store the result for the process lifetime.

SomberNight: > real    0m2.614s

o.O

ok, well, changed to 5 sec timeout in df770bb05ae133d1d2b9513d923c51c62fa9e8f5. I think it is worth having some kind of max timeout on it. Ultimately we don't want to block forever.

btw I started looking at this as I was wondering whether this is fast enough to call during startup, as I was considering using it as an entropy source. Evidently it is much too slow for that.

PR #10884, opened 2026-08-21
config: make sure running the unittests can never mutate prod config

SomberNight:
related: https://github.com/spesmilo/electrum/pull/10883

Running the unittests must never mutate the mainnet/production config file (or other files in the datadir) of a dev machine. I am not sure how exactly to prevent that, but this is one approach.

The idea is that a test case constructing e.g. a config file should use a tempfolder for it (and then clean-up after itself), however in case that is accidentally not done, we should have a belt-and-suspenders last-resort protection. In this PR I propose using e.g. `/home/user/.electrum/unittests/config` instead of `/home/user/.electrum/config`: we always know we are in a unittest context, and if so, we add an extra level of nesting.

### DIFF
--- a/electrum/simple_config.py
+++ b/electrum/simple_config.py
@@ -24,6 +24,7 @@ _logger = get_logger(__name__)
 
 FINAL_CONFIG_VERSION = 3
 
+_RUNNING_UNITTESTS = False
 
 _config_var_from_key = {}  # type: Dict[str, 'ConfigVar']
 
@@ -236,6 +237,9 @@ class SimpleConfig(Logger):
         # Otherwise use the user's default data directory.
         path = self.get('electrum_path') or self.user_dir()
         make_dir(path, allow_symlink=False)
+        if _RUNNING_UNITTESTS:
+            path = os.path.join(path, "unittests")
+            make_dir(path, allow_symlink=False)
         return path
 
     @classmethod
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -30,6 +30,7 @@ FAST_TESTS = False
 electrum.logging._configure_stderr_logging(verbosity="*")
 
 electrum.util.AS_LIB_USER_I_WANT_TO_MANAGE_MY_OWN_ASYNCIO_LOOP = True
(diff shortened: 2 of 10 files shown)


f321x:
Might be nice if regtest tests also have some fallback, the `unixsocket` test for example writes into the `~/.electrum/regtest` dir.

This might also work and gets passed on to the regtests without requiring modifications to `SimpleConfig` or writing in the user dir at all. Otoh setting an env variable might be more fragile?
```diff
diff --git a/tests/__init__.py b/tests/__init__.py
index 1dc6c680ae..dc64d4228d 100644
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,4 +1,4 @@
-import asyncio
+import atexit
 import os
 import unittest
 import threading
@@ -30,7 +30,8 @@ FAST_TESTS = False
 electrum.logging._configure_stderr_logging(verbosity="*")
 
 electrum.util.AS_LIB_USER_I_WANT_TO_MANAGE_MY_OWN_ASYNCIO_LOOP = True
-electrum.simple_config._RUNNING_UNITTESTS = True
+os.environ["ELECTRUMDIR"] = _unittests_datadir = tempfile.mkdtemp(prefix="electrum-unittests-datadir-")
+atexit.register(lambda: shutil.rmtree(_unittests_datadir, ignore_errors=True))
 
 
 class ElectrumTestCase(unittest.IsolatedAsyncioTestCase, Logger):
```

electrum/simple_config.py
```diff
@@ -236,6 +237,9 @@ def electrum_path_root(self):
         # Otherwise use the user's default data directory.
         path = self.get('electrum_path') or self.user_dir()
         make_dir(path, allow_symlink=False)
+        if _RUNNING_UNITTESTS:
+            path = os.path.join(path, "unittests")
+            make_dir(path, allow_symlink=False)
```
ecdsa, line 242: I do not like the multiplication of paths.
#10833 introduced `tempfile.mkdtemp(prefix="electrum-unittest-base-")`
Why not assert here that `electrum_path` has been set and is starting with that prefix?

PR #10885, opened 2026-08-22
crypto.py: add some sanity checks

SomberNight:
see commits

### DIFF
--- a/electrum/crypto.py
+++ b/electrum/crypto.py
@@ -504,3 +504,64 @@ def get_ecdh(priv: bytes, pub: bytes) -> bytes:
 
 def privkey_to_pubkey(priv: bytes) -> bytes:
     return ecc.ECPrivkey(priv[:32]).get_public_key_bytes()
+
+
+########################################
+# Run-time sanity checks.
+# - If one of the important cryptographic primitives is broken, we better panic.
+# - For several primitives we support multiple backends. By now at runtime,
+#   we have already selected the backend: check if it really works and fail early if not.
+# - This whole section takes 5-10 msec.
+#
+# Hash functions. SHA2. single backend: hashlib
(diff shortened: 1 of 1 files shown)


electrum/crypto.py
```diff
@@ -350,6 +354,12 @@ def ripemd(x: bytes) -> bytes:
         return md.digest()
 
 
+# Run-time sanity checks: if one of the important hash functions is broken, we better panic.
+assert sha256(b"satoshi_nakamoto").hex() == "5f94a8490efe9e06c590dd34e37b5ab8f482f1af7578c6a41542761938a42426"
+assert hashlib.sha512(b"satoshi_nakamoto").digest().hex() == "cae681a7f07bd26128f8536c59a38f10c9e648898426cd1dd94144c7d3ea0512186400edc5d38ac677d43e97ebb877bf76d69c44de1f6e074435adc79caf8f14"
+assert hash_160(b"satoshi_nakamoto").hex() == "b3abbb1cb430ef62152b0b1c34330a4a926cb615"
+
```
f321x: Considering we support 3 different backends for AES, would it make sense to add a sanity check for AES (and ChaCha) too?

SomberNight: Right... I was already considering that and was on the fence.
but yes, I think it would be better if we checked those too, see 1e3792f08a3612490b31562324533099517eebdd

One counterpoint is that checking aes+chacha+chachapoly takes around 100x as much time (~7 ms) as sha256+sha512+ripemd (~0.07 ms) (assuming ripemd is from the openssl backend). Though if ripemd is using the pure-python fallback from ripemd.py, that takes 1 ms.

Still, I think the checks are worth the time.

f321x: > Though if ripemd is using the pure-python fallback from ripemd.py, that takes 1 ms

Fyi, ripemd.py already does some sanity checks internally too
https://github.com/spesmilo/electrum/blob/29991f4ab8f055a4d8cd734fa47f38bc79bdeca2/electrum/ripemd.py#L388-L393

PR #10888, opened 2026-08-23
lnchannel: do not remove funded channel once funding tx is confirmed

ecdsa:


electrum/lnchannel.py
```diff
@@ -846,7 +843,7 @@ def has_onchain_backup(self):
         return self.storage.get('has_onchain_backup', False)
 
     def can_be_deleted(self) -> bool:
-        if self.has_funding_timed_out():
+        if (not self.is_funded() or self.is_zeroconf()) and self.has_funding_timed_out():
```
SomberNight: why the check for `is_zeroconf`?

Consider a channel where
- `is_zeroconf` (meaning the channel type has the flag), and
- we are not the funder
- funding tx is deeply mined but the electrum server suddenly claims it is unconfirmed

I think this `can_be_deleted()` would return True, which we probably do not want.

f321x: note, the zeroconf flag is removed once `is_funding_tx_mined` became True, so a channel with deeply mined funding tx shouldn't carry the flag anymore.

https://github.com/spesmilo/electrum/blob/a042ee9fe6c20faa500c8ad28c970b424700897a/electrum/lnchannel.py#L424-L426

amylesueurtheallahmadre:
> …no longer funded.

I fund myself

electrum/lnchannel.py
ecdsa: I removed that change; it was a belt-and-suspenders thing, but it was also introducing another bug: a channel that was never funded but is in the force-closing state could not be deleted

electrum/lnchannel.py
```diff
@@ -262,24 +263,25 @@ def need_to_subscribe(self) -> bool:
...
 
-    def save_funding_height(self, *, txid: str, height: int, timestamp: Optional[int]) -> None:
-        self.storage['funding_height'] = txid, height, timestamp
+    def _maybe_save_tx_height(self, *, name, txid: str, info: TxMinedInfo) -> None:
+        height = info.height()
+        h = self.storage.get(name)
+        prev_height = h[1] if h else TX_HEIGHT_LOCAL
+        if prev_height > 0 and height <= 0:
+            # do not revert from confirmed to unconfirmed
+            return
```
SomberNight, line 272: what is the desired behaviour if the funding tx gets 1 conf (but fewer than `funding_txn_minimum_depth`) and then gets reorged out and never mined again?

AFAICT atm the channel will stay in the OPENING state forever, will never time out, and can never be deleted.

SomberNight: well, if we are the funder, `update_unfunded_state` actually has reasonable behaviour:
https://github.com/spesmilo/electrum/blob/da98f9a68039c0f1a5e12489210444fc51132f57/electrum/lnchannel.py#L354

but if the remote is the funder, we are stuck with the channel

tests/test_lnchannel.py
```diff
@@ -582,6 +584,84 @@ def test_unfunded_channel_can_be_removed(self):
...
+        )
+
+        # the saved funding height must not be overwritten, and the channel must not be removed
+        self.assertEqual((funding_txid, funding_confirmed_height, 1600000000), chan.get_funding_height())
+        self.assertFalse(chan.has_funding_timed_out())
+        self.assertFalse(chan.can_be_deleted())
+        mock_lnworker.remove_channel.assert_not_called()
+
+        # the server now omits the funding tx entirely, so that we forget the saved height,
+        # and then claims the funding tx is in the mempool again
```
SomberNight: where is the `and then claims the funding tx is in the mempool again` part?

SomberNight: comment was just stale. removed in 1678ce7c848e5b841ed4638293070cdcabdfe291

electrum/lnchannel.py
SomberNight: bcf5be933137e736fc35d5b5d576ecbecc0bc0c6 fixes the non-funder case

ecdsa: looks good

PR #10889, opened 2026-08-24
github: add anti-slop PR template

f321x:
Adds a PR template which asks the creator of a PR for a handwritten description and understanding of the patch. 
This hopefully reduces the amount of fully generated PRs whose underlying issues could be solved more efficient if the author just opened an Issue and let us fix the issue instead of going back-and-forth with the author to clean it up.


SomberNight (approved):
fully agree with the idea, and I think the text is good

ecdsa:
ack

PR #10890, opened 2026-08-24
lnwatcher/qt: fix regressed pending force close warning

f321x:
The warning dialog supposed to be shown when we wait for a ctx to get confirmed so we don't miss claiming the htlc was not shown anymore due to a regression from f337b4782d. 
We were looking at our cltv (which is 0) instead of the remote one, and always considered the htlc too old for the warning.
Also now shows the exact remaining blocks in the warning dialog instead of the constant, resolving a `fixme` comment.


ecdsa:
looks good.

ecdsa (approved):

PR #10892, opened 2026-08-25
lnpeer: rm early return in send_revoke_and_ack

ecdsa:
This method is a response to us receiving a new commitment, which is gated on `can_update_ctx(proposer=REMOTE)`. We need to send that response. `can_update_ctx(proposer=LOCAL)` is stricter, because _can_send_ctx_updates is set during shutdown.

ref: CLOSE-6

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -1958,7 +1958,9 @@ class Peer(Logger, EventListener):
 
     def maybe_send_commitment(self, chan: Channel) -> bool:
         assert util.get_running_loop() == util.get_asyncio_loop(), f"this must be run on the asyncio thread!"
-        if not chan.can_update_ctx(proposer=LOCAL):
+        # proposer=REMOTE, because commitment_signed only commits changes that were already proposed,
+        # so it must not be gated on chan._can_send_ctx_updates (see send_shutdown).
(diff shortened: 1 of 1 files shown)


ecdsa:
Note: I think it would be better to rename `can_update_ctx(proposer=xxx)`, because it does not describe what it is doing.

electrum/lnpeer.py
```diff
@@ -1958,7 +1958,9 @@ def on_update_fail_htlc(self, chan: Channel, payload):
 
     def maybe_send_commitment(self, chan: Channel) -> bool:
         assert util.get_running_loop() == util.get_asyncio_loop(), f"this must be run on the asyncio thread!"
-        if not chan.can_update_ctx(proposer=LOCAL):
+        # proposer=REMOTE, because commitment_signed only commits changes that were already proposed,
+        # so it must not be gated on chan._can_send_ctx_updates (see send_shutdown).
+        if not chan.can_update_ctx(proposer=REMOTE):
```
SomberNight, line 1963: I don't like this approach. Makes it hard to understand what `proposer` means.
I would prefer splitting `can_update_ctx` somehow. See https://github.com/spesmilo/electrum/pull/10897

ecdsa: I agree. see my comment in #10897

PR #10895, opened 2026-08-25
Qml wizard warn wrong electrum seed

accumulator:

### DIFF
--- a/electrum/gui/qml/components/wizard/WCHaveSeed.qml
+++ b/electrum/gui/qml/components/wizard/WCHaveSeed.qml
@@ -20,6 +20,7 @@ WizardComponent {
 
     property string _seedType
     property string _validationMessage
+    property string _warning
     property bool _canPassphrase
     property bool _seedValid
 
@@ -72,11 +73,18 @@ WizardComponent {
         var verifyResult = wiz.verifySeed(seedtext.text, seed_variant_cb.currentValue, wizard_data['wallet_type'])
 
         _validationMessage = verifyResult.message
+        _warning = verifyResult.warning
         _seedType = verifyResult.type
         _canPassphrase = verifyResult.can_passphrase
 
         if (!cosigner || !verifyResult.valid) {
             _seedValid = verifyResult.valid
+            if (_warning) {
+                infotext.text = _warning
+                infotext.visible = true
+            } else {
+                setSeedTypeHelpText()
+            }
         } else {
             // bip39 validate after derivation path is known
             if (seed_variant_cb.currentValue == 'electrum') {
@@ -190,6 +198,7 @@ WizardComponent {
                 Layout.fillWidth: true
                 Layout.columnSpan: 2
                 Layout.topMargin: constants.paddingLarge
+                iconStyle: root._warning === infotext.text ? InfoTextArea.IconStyle.Warn : InfoTextArea.IconStyle.Info
                 compact: true
                 backgroundColor: constants.darkerDialogBackground
             }
--- a/electrum/wizard.py
+++ b/electrum/wizard.py
@@ -312,10 +312,23 @@ class KeystoreWizard(AbstractWizard):
         run_hook('init_wallet_wizard', self)  # TODO: currently only used for hww, hook name might be confusing
         return plugin.wizard_entry_for_device(_info, new_wallet=new_wallet)
 
-    def validate_seed(self, seed: str, seed_variant: str, wallet_type: str) -> Tuple[bool, str, str, bool]:
+    def _warning_wrong_electrum_seed_type(self, seed_type: str) -> str:
+        if not seed_type:
+            return ''
+        wiztype_fullname = _('Wallet with two-factor authentication') if is_any_2fa_seed_type(seed_type) \
+            else _("Standard wallet")
+        msg = ' '.join([
+            _("Looks like you have entered a valid seed of type '{}' but this dialog does not support such seeds.").format(
+                seed_type),
+            _("If unsure, try restoring as '{}'.").format(wiztype_fullname),
+        ])
+        return msg
+
+    def validate_seed(self, seed: str, seed_variant: str, wallet_type: str) -> Tuple[bool, str, str, str, bool]:
         seed_type = ''
         seed_valid = False
         validation_message = ''
+        warning = ''
         can_passphrase = True
 
         if seed_variant == 'electrum':
@@ -345,14 +358,17 @@ class KeystoreWizard(AbstractWizard):
         # check if seed matches wallet type
         if wallet_type == '2fa' and not is_any_2fa_seed_type(seed_type):
             seed_valid = False
+            warning = self._warning_wrong_electrum_seed_type(seed_type)
         elif wallet_type == 'standard' and seed_type not in ['old', 'standard', 'segwit', 'bip39', 'slip39']:
             seed_valid = False
+            if seed_variant == 'electrum':
+                warning = self._warning_wrong_electrum_seed_type(seed_type)
         elif wallet_type == 'multisig' and seed_type not in ['standard', 'segwit', 'bip39', 'slip39']:
             seed_valid = False
 
         self._logger.debug(f'seed verified: {seed_valid}, type={seed_type!r}, validation_message={validation_message}')
 
-        return seed_valid, seed_type, validation_message, can_passphrase
+        return seed_valid, seed_type, validation_message, warning, can_passphrase
 
     def keystore_from_data(self, wallet_type: str, data: dict):
         if data['keystore_type'] in ['createseed', 'haveseed'] and 'seed' in data:
--- a/electrum/gui/qml/qewizard.py
+++ b/electrum/gui/qml/qewizard.py
@@ -116,11 +116,12 @@ class QENewWalletWizard(NewWalletWizard, QEAbstractWizard):
 
     @pyqtSlot(str, str, str, result='QVariantMap')
     def verifySeed(self, seed, seed_variant, wallet_type='standard'):
-        seed_valid, seed_type, validation_message, can_passphrase = self.validate_seed(seed, seed_variant, wallet_type)
+        seed_valid, seed_type, validation_message, warning, can_passphrase = self.validate_seed(seed, seed_variant, wallet_type)
         return {
             'valid': seed_valid,
             'type': seed_type,
             'message': validation_message,
+            'warning': warning,
             'can_passphrase': can_passphrase
         }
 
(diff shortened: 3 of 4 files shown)


electrum/gui/qml/components/wizard/WCHaveSeed.qml
```diff
@@ -72,11 +73,18 @@ WizardComponent {
...
 
         _validationMessage = verifyResult.message
+        _warning = verifyResult.warning
         _seedType = verifyResult.type
         _canPassphrase = verifyResult.can_passphrase
 
         if (!cosigner || !verifyResult.valid) {
             _seedValid = verifyResult.valid
+            if (_warning) {
+                infotext.text = _warning
```
f321x, line 83: Might look nicer if the text is shown with the yellow border, so the user also doesn't need to read the `Warning:` in the string?

Image

instead of:

Image

This seemed to do the job:
```diff
diff --git a/electrum/gui/qml/components/wizard/WCHaveSeed.qml b/electrum/gui/qml/components/wizard/WCHaveSeed.qml
index 4c8b8047a1..63199db633 100644
--- a/electrum/gui/qml/components/wizard/WCHaveSeed.qml
+++ b/electrum/gui/qml/components/wizard/WCHaveSeed.qml
@@ -198,6 +198,7 @@ WizardComponent {
                 Layout.fillWidth: true
                 Layout.columnSpan: 2
                 Layout.topMargin: constants.paddingLarge
+                iconStyle: root._warning === infotext.text ? InfoTextArea.IconStyle.Warn : InfoTextArea.IconStyle.Info
                 compact: true
                 backgroundColor: constants.darkerDialogBackground
             }
diff --git a/electrum/wizard.py b/electrum/wizard.py
index d8773f82a1..2f566f46e9 100644
--- a/electrum/wizard.py
+++ b/electrum/wizard.py
@@ -318,7 +318,6 @@ class KeystoreWizard(AbstractWizard):
         wiztype_fullname = _('Wallet with two-factor authentication') if is_any_2fa_seed_type(seed_type) \
             else _("Standard wallet")
         msg = ' '.join([
-            '<b>' + _('Warning') + ':</b>  ',
             _("Looks like you have entered a valid seed of type '{}' but this dialog does not support such seeds.").format(
                 seed_type),
             _("If unsure, try restoring as '{}'.").format(wiztype_fullname),

```

electrum/wizard.py
```diff
@@ -312,10 +312,24 @@ def on_hardware_device(self, wizard_data: dict, new_wallet=True) -> str:
         run_hook('init_wallet_wizard', self)  # TODO: currently only used for hww, hook name might be confusing
         return plugin.wizard_entry_for_device(_info, new_wallet=new_wallet)
 
-    def validate_seed(self, seed: str, seed_variant: str, wallet_type: str) -> Tuple[bool, str, str, bool]:
+    def _warning_wrong_electrum_seed_type(self, seed_type: str) -> str:
```
f321x, line 315: This could be used to deduplicate the identical warning in `seed_dialog.py`: https://github.com/spesmilo/electrum/blob/a8ff16af7efd9661f9443f2ceb38758355107d1c/electrum/gui/qt/seed_dialog.py#L325-L331
<details>
<summary> This works (click to expand) </summary>

```diff
diff --git a/electrum/gui/qt/seed_dialog.py b/electrum/gui/qt/seed_dialog.py
index 0b0da25ae3..e5ffe9c6b9 100644
--- a/electrum/gui/qt/seed_dialog.py
+++ b/electrum/gui/qt/seed_dialog.py
@@ -36,6 +36,7 @@ from electrum.mnemonic import Mnemonic, calc_seed_type, is_any_2fa_seed_type
 from electrum import old_mnemonic
 from electrum import slip39
 from electrum.util import ChoiceItem
+from electrum.wizard import KeystoreWizard
 
 from .util import (
     Buttons, OkButton, WWLabel, ButtonsTextEdit, icon_path, EnterButton,
@@ -322,13 +323,7 @@ class SeedWidget(QWidget):
             t = calc_seed_type(s)
             label = _('Seed Type') + ': ' + t if t else ''
             if t and not valid:  # electrum seed, but does not conform to dialog rules
-                wiztype_fullname = _('Wallet with two-factor authentication') if is_any_2fa_seed_type(t) else _("Standard wallet")
-                msg = ' '.join([
-                    '<b>' + _('Warning') + ':</b>  ',
-                    _("Looks like you have entered a valid seed of type '{}' but this dialog does not support such seeds.").format(t),
-                    _("If unsure, try restoring as '{}'.").format(wiztype_fullname),
-                ])
-                self.seed_warning.setText(msg)
+                self.seed_warning.setText(KeystoreWizard.warning_wrong_electrum_seed_type(t))
             else:
                 self.seed_warning.setText("")
 
diff --git a/electrum/wizard.py b/electrum/wizard.py
index d8773f82a1..2b15a06e47 100644
--- a/electrum/wizard.py
+++ b/electrum/wizard.py
@@ -312,7 +312,8 @@ class KeystoreWizard(AbstractWizard):
         run_hook('init_wallet_wizard', self)  # TODO: currently only used for hww, hook name might be confusing
         return plugin.wizard_entry_for_device(_info, new_wallet=new_wallet)
 
-    def _warning_wrong_electrum_seed_type(self, seed_type: str) -> str:
+    @staticmethod
+    def warning_wrong_electrum_seed_type(seed_type: str) -> str:
         if not seed_type:
             return ''
         wiztype_fullname = _('Wallet with two-factor authentication') if is_any_2fa_seed_type(seed_type) \
@@ -359,11 +360,11 @@ class KeystoreWizard(AbstractWizard):
         # check if seed matches wallet type
         if wallet_type == '2fa' and not is_any_2fa_seed_type(seed_type):
             seed_valid = False
-            warning = self._warning_wrong_electrum_seed_type(seed_type)
+            warning = self.warning_wrong_electrum_seed_type(seed_type)
         elif wallet_type == 'standard' and seed_type not in ['old', 'standard', 'segwit', 'bip39', 'slip39']:
             seed_valid = False
             if seed_variant == 'electrum':
-                warning = self._warning_wrong_electrum_seed_type(seed_type)
+                warning = self.warning_wrong_electrum_seed_type(seed_type)
         elif wallet_type == 'multisig' and seed_type not in ['standard', 'segwit', 'bip39', 'slip39']:
             seed_valid = False
 
```

</details>

electrum/wizard.py
```diff
@@ -531,6 +548,8 @@ def on_keystore_type(self, wizard_data: dict) -> str:
         }.get(t)
 
     def on_have_or_confirm_seed(self, wizard_data: dict) -> str:
+        if is_any_2fa_seed_type(wizard_data['seed_type']):
+            return 'trustedcoin_have_ext' if self.wants_ext(wizard_data) else 'trustedcoin_keep_disable'
```
f321x: This is not reachable as the Next button is disabled for `2fa` seeds and the trustedcoin plugin provides its own `next`, not calling `is_any_2fa_seed_type` during the 2fa wizard.

If it would be reachable somehow, it would raise already at the `accept` callback:
```
  File "trace_probe.py", line 15, in test_probe
    w.resolve_next('have_seed', d)
  File "electrum/wizard.py", line 80, in resolve_next
    view_accept(wizard_data)
  File "electrum/wizard.py", line 454, in <lambda>
    'accept': lambda d: None if self.wants_ext(d) else self.maybe_master_pubkey(d),
  File "electrum/wizard.py", line 566, in maybe_master_pubkey
    wizard_data['multisig_master_pubkey'] = self.keystore_from_data(wizard_data['wallet_type'], wizard_data).get_master_public_key()
  File "electrum/wizard.py", line 379, in keystore_from_data
    return keystore.from_seed(data['seed'], passphrase=seed_extension, for_multisig=for_multisig)
  File "electrum/keystore.py", line 1228, in from_seed
    raise BitcoinException('Unexpected seed type {}'.format(repr(t)))
electrum.util.BitcoinException: Unexpected seed type '2fa_segwit'
```

with a wizard state like this:
```python
{'wallet_name': 'test', 'wallet_type': 'standard', 'keystore_type': 'haveseed',
     'seed': 'fortune theme add purity broom bubble tortoise smart acid tiny safe sample',
     'seed_variant': 'electrum', 'seed_type': '2fa_segwit', 'seed_extend': False}
```

Maybe a simple assert would be clearer?

electrum/wizard.py
```diff
@@ -345,14 +359,17 @@ def validate_seed(self, seed: str, seed_variant: str, wallet_type: str) -> Tuple
         # check if seed matches wallet type
         if wallet_type == '2fa' and not is_any_2fa_seed_type(seed_type):
             seed_valid = False
+            warning = self._warning_wrong_electrum_seed_type(seed_type)
         elif wallet_type == 'standard' and seed_type not in ['old', 'standard', 'segwit', 'bip39', 'slip39']:
             seed_valid = False
+            if seed_variant == 'electrum':
+                warning = self._warning_wrong_electrum_seed_type(seed_type)
         elif wallet_type == 'multisig' and seed_type not in ['standard', 'segwit', 'bip39', 'slip39']:
             seed_valid = False
```
f321x, line 368: On desktop we also show a warning if the user enters the wrong seed type into the multisig wizard:
```suggestion
            warning = self._warning_wrong_electrum_seed_type(seed_type)

```

electrum/wizard.py
accumulator: No this was not supposed to be in the commit, just local testing :)

electrum/wizard.py
accumulator: Yes I saw that, but I held off on refactoring this like you suggested above because it felt like a circular dependency using the class and its static member in a control that is typically imported in a wizard.

SomberNight: somewhat related, I have long thought we have too many top-level modules (.py files directly in `electrum/`). One idea for a new folder is `mnemonic/`, which would include:
- mnemonic.py
- old_mnemonic.py
- slip39.py
- the "bip39" parts extracted from keystore.py, in a new bip39.py

(EDIT: there is also bip39_recovery.py, bip39_wallet_formats.json, perhaps even the whole wordlist/ folder)

If we had that folder, you could somewhat more cleanly also add another module there for shared UI code, and put this function there.

SomberNight: You can also add a FIXME pointing out where the duplication is, and we can fix it later, if you want. We should definitely try to avoid unmarked duplication though.
I can do the above mentioned refactor but don't want to hold up this PR on that.

ecdsa:
it would be better to redirect the flow if the wrong seed type is entered. 

EDIT: see #10982

electrum/wizard.py
ecdsa: > Yes I saw that, but I held off on refactoring this like you suggested above because it felt like a circular dependency using the class and its static member in a control that is typically imported in a wizard.

why would that be a circular dependency? it looks fine to me

Issue #10801, opened 2026-08-05
CI: GHA "security-review" workflow cannot be run on external contributors' PRs


SomberNight:
We currently have all CI tasks auto-run on PRs from maintainers.
For other PRs, the CI only runs after manual approval from a maintainer.

But even if I manually approve a non-maintainer PR, the ["security-review"](https://github.com/spesmilo/electrum/blob/b1aa52d701eb826d3c68ae9c03737cc05163c0e2/.github/workflows/security-review.yml#L25
) task will get skipped. The logic there is quite convoluted, and note that task uses the `pull_request_target` GHA trigger instead of the safer `pull_request` trigger, because it needs access to a cloud-LLM API token from spesmilo.

Maintainers should be able to manually trigger the "security-review" CI task for any PR.

-----

question: what if I click "approve workflows" for a PR, the tasks run, and then later the PR author pushes more commits?  I would expect and want to have to click "approve workflows" again. Is that so?

SomberNight:
Felix pointed out temporary workaround: the "security-review" task can be manually started from the "Actions" tab UI: https://github.com/spesmilo/electrum/actions/workflows/security-review.yml

SomberNight:
> question: what if I click "approve workflows" for a PR, the tasks run, and then later the PR author pushes more commits? I would expect and want to have to click "approve workflows" again. Is that so?

At least experimentally it seems we have to click it after every push. At least that is sane then :)

SomberNight:
Hmm, AFAIU the ["Approve workflows to run"-gating button](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks) only applies to the `pull_request` trigger tasks.
As the "security-review" task uses `pull_request_target`, it auto-starts without any gating, so the only protection for the API key is the custom check inside the yml:
https://github.com/spesmilo/electrum/blob/b1aa52d701eb826d3c68ae9c03737cc05163c0e2/.github/workflows/security-review.yml#L25-L30

SomberNight:
Even if it did work for `pull_request_target`, I wonder if the `Approve workflows to run` button is vulnerable to a TOCTOU:
is there a window between the maintainer clicking "approve" and the the workflow pulling the pull request HEAD when the branch could be force-pushed? Or does the button bind the commit hash being shown on the maintainer's browser to the workflows somehow?

Here it looks like it is pulling the PR HEAD some ~seconds after the workflow started:
[...]

PR #10896, opened 2026-08-25
ci: llm review: expand comment and add fixme re race

SomberNight:
This was reported by `jackieisbad`.

ref https://github.com/spesmilo/electrum/issues/10801

Given we hardly use the "workflow_dispatch" trigger, I think it's ok to just publicly document this for now.


SomberNight:
Thomas remarked that using the web interface to click the "merge" button has the same race.

PR #10897, opened 2026-08-25
lnpeer: split can_update_ctx checks

SomberNight:

commit `lnpeer: when sending commitsig/revack, ignore _can_send_ctx_updates`  is an alternative to https://github.com/spesmilo/electrum/pull/10892

### DIFF
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1092,19 +1092,13 @@ class Channel(AbstractChannel):
             return False
         return True
 
-    def can_update_ctx(self, *, proposer: HTLCOwner) -> bool:
-        """Whether proposer is allowed to send update_* messages.
-        (update_fee, update_*_htlc)
-        """
-        if not self.can_progress_ctx():
-            return False
-        if proposer == LOCAL:
-            if not self._can_send_ctx_updates:
-                return False
-        return True
+    def can_send_ctx_updates(self) -> bool:
+        """Whether *we* can send update_* messages."""
+        return self.can_progress_ctx() and self._can_send_ctx_updates
 
     def can_send_update_add_htlc(self) -> bool:
-        return self.can_update_ctx(proposer=LOCAL) and self.is_open()
+        """Whether *we* can send update_add_htlc."""
+        return self.can_send_ctx_updates() and self.is_open()
 
     def is_frozen_for_sending(self) -> bool:
         if self.lnworker.uses_trampoline() and not self.lnworker.is_trampoline_peer(self.node_id):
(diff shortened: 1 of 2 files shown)


electrum/lnchannel.py
```diff
@@ -1084,20 +1084,27 @@ def get_state_for_GUI(self):
...
+    def can_update_ctx(self, *, proposer: HTLCOwner) -> bool:
+        """Whether proposer is allowed to send update_* messages.
+        (update_fee, update_*_htlc)
+        """
+        if not self.can_sign_or_rev_ctns():
+            return False
         if proposer == LOCAL:
             if not self._can_send_ctx_updates:
                 return False
         return True
```
ecdsa: I do not like this. I think this is making things unnecessarily complicated.

I agree that `can_update_ctx` is confusing. I wanted to minimize the diff in my PR. However, if we decide to clarify things, then we should remove that method.

```diff
-    def can_sign_or_rev_ctns(self) -> bool:
+    def can_progress_ctx(self) -> bool:
         """Whether chan is in a state that generally allows signing new commitments or revoking them."""
         if self.get_state() not in (ChannelState.OPEN, ChannelState.SHUTDOWN):
             return False
@@ -1092,20 +1092,13 @@ class Channel(AbstractChannel):
             return False
         return True
 
-    def can_update_ctx(self, *, proposer: HTLCOwner) -> bool:
-        """Whether proposer is allowed to send update_* messages.
-        (update_fee, update_*_htlc)
-        """
-        if not self.can_sign_or_rev_ctns():
-            return False
-        if proposer == LOCAL:
-            if not self._can_send_ctx_updates:
-                return False
-        return True
+    def can_send_ctx_updates(self) -> bool:
+        """Whether *we* can send updates."""
+        return self.can_progress_ctx() and self._can_send_ctx_updates
 
-    def can_send_update_add_htlc(self) -> bool:
+    def can_add_htlcs(self) -> bool:
         """Whether *we* can send update_add_htlc."""
-        return self.can_update_ctx(proposer=LOCAL) and self.is_open()
+        return self.can_send_ctx_updates() and self.is_open()


```

and then replace all calls of `can_update_ctx` with the corresponding method, depending on `proposer`.

Also please rename `can_send_update_add_htlcs` -> `can_add_htlcs`, and `can_sign_or_rev_ctns` -> `can_progress_ctx`.  In general, 4 words should to be sufficient for most method names.

SomberNight: > Also please rename `can_send_update_add_htlcs` -> `can_add_htlcs` [...]`.  In general, 4 words should to be sufficient for most method names.

We already have a method named `_assert_can_add_htlc`. The distinction between the two would be even less clear if this one was named `can_add_htlcs` IMO. I would keep it as `can_send_update_add_htlc`.

Re other points, ok.

PR #10899, opened 2026-08-26
lnpeer: harden _check_unfulfilled_htlc

ecdsa:
see commits

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2227,6 +2227,10 @@ class Peer(Logger, EventListener):
             if not fw_enabled:
                 _log_fail_reason("forwarding is disabled")
                 raise OnionRoutingFailure(code=OnionFailureCode.PERMANENT_CHANNEL_FAILURE, data=b'')
+            # we must not forward an htlc whose payment_hash matches a payment request we created
+            if self.lnworker.maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(payment_hash):
+                _log_fail_reason(f"RHASH corresponds to payreq we created")
+                raise OnionRoutingFailure(code=OnionFailureCode.TEMPORARY_NODE_FAILURE, data=b'')
             if outer_onion_payment_secret:
                 # this is a trampoline forwarding htlc, multiple incoming trampoline htlcs can be collected
                 payment_key = (payment_hash + outer_onion_payment_secret).hex()
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -4073,9 +4073,6 @@ class LNWallet(Logger):
         if htlc.amount_msat - next_amount_msat_htlc < forwarding_fees:
             data = next_amount_msat_htlc.to_bytes(8, byteorder="big") + outgoing_chan_upd_message
             raise OnionRoutingFailure(code=OnionFailureCode.FEE_INSUFFICIENT, data=data)
-        if self._maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(htlc.payment_hash):
-            log_fail_reason(f"RHASH corresponds to payreq we created")
-            raise OnionRoutingFailure(code=OnionFailureCode.TEMPORARY_NODE_FAILURE, data=b'')
         self.logger.info(
             f"maybe_forward_htlc. will forward HTLC: inc_chan={incoming_chan.short_channel_id}. inc_htlc={str(htlc)}. "
             f"next_chan={next_chan.get_id_for_log()}.")
@@ -4137,12 +4134,6 @@ class LNWallet(Logger):
         except Exception as e:
             self.logger.exception('')
             raise OnionRoutingFailure(code=OnionFailureCode.INVALID_ONION_PAYLOAD, data=b'\x00\x00\x00')
-
-        if self._maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(payment_hash):
-            self.logger.debug(
-                f"maybe_forward_trampoline. will FAIL HTLC(s). "
-                f"RHASH corresponds to payreq we created. {payment_hash.hex()=}")
-            raise OnionRoutingFailure(code=OnionFailureCode.TEMPORARY_NODE_FAILURE, data=b'')
 
         # these are the fee/cltv paid by the sender
         # pay_to_node will raise if they are not sufficient
@@ -4240,7 +4231,7 @@ class LNWallet(Logger):
                 data = b''
             raise OnionRoutingFailure(code=OnionFailureCode.UNKNOWN_NEXT_PEER, data=data)
 
-    def _maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(self, payment_hash: bytes) -> bool:
+    def maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(self, payment_hash: bytes) -> bool:
         """Returns True if the HTLC should be failed.
         We must not forward HTLCs with a matching payment_hash to a payment request we created.
         Example attack:
--- a/tests/test_lnpeer.py
+++ b/tests/test_lnpeer.py
@@ -1301,6 +1301,79 @@ class TestPeerDirect(TestPeer):
                 await f()
 
 
+    async def test_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(self):
+        """Alice holds an invoice created by Bob, hence she knows RHASH and Bob's payment_secret.
+        She sends Bob a dust htlc whose outer onion is addressed to Bob (using the invoice's
+        payment_secret, and claiming a tiny total_msat), but whose inner trampoline onion asks Bob
+        to *forward* the payment.
+        Bob must fail such htlcs, and he must not release the preimage.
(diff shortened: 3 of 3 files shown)


electrum/lnpeer.py
```diff
@@ -2250,17 +2250,21 @@ def _check_unfulfilled_htlc(
...
+            #       for the final node only, but we apply them when we are asked to relay as well.
+            #       Without this check, a single dust htlc claiming a tiny total_msat would push a
+            #       whole htlc set into COMPLETE/SETTLING.
+            try:
+                # note: the inner payload is attacker-chosen, these might be missing (None)
+                assert (inner_amt_to_forward := trampoline_onion.amt_to_forward) is not None
+                assert (inner_cltv_abs := trampoline_onion.outgoing_cltv_value) is not None
+                assert processed_onion.total_msat >= inner_amt_to_forward
+                assert processed_onion.outgoing_cltv_value >= inner_cltv_abs
+            except AssertionError:
```
SomberNight, line 2286: I am not convinced these checks are equivalent.
Now you are never looking at the value of `processed_onion.amt_to_forward`.

~~Perhaps we want this:
`assert processed_onion.total_msat >= processed_onion.amt_to_forward >= inner_amt_to_forward`~~

EDIT: in case of MPP, `processed_onion.amt_to_forward >= inner_amt_to_forward` does not hold, so my suggestion is wrong too. Still, my point is that some checks are missing here.

ecdsa: Indeed, the missing check is  `processed_onion.total_msat >= processed_onion.amt_to_forward`.
If these values are equal, then we are in the else branch of the `if is_mpp` construct, 
Thus, the former assert becomes equivalent to:
`assert processed_onion.total_msat >= inner_amt_to_forward`

SomberNight: Ok, this now looks equivalent.
I think there might be value in implementing the checks exactly as written in the BOLTS, OTOH this is now somewhat easier to read. So, ok.

electrum/lnworker.py
```diff
@@ -2958,7 +2958,7 @@ def update_or_create_mpp_with_received_htlc(
     ):
         # Payment key creation:
         #   * for regular forwarded htlcs -> "scid.hex() + ':%d' % htlc_id" [htlc key]
-        #   * for trampoline forwarding -> "payment hash + payment secret from outer onion"
+        #   * for trampoline forwarding -> "fwd:" + (payment hash + payment secret from outer onion).hex()
         #   * for final non-trampoline htlcs (we are receiver) -> "payment hash + payment secret from onion"
         #   * for final trampoline htlcs (we are receiver) -> 2. step grouping:
         #           1. grouping of htlcs by "payments hash + outer onion payment secret", a 'multi-trampoline mpp part'.
```
SomberNight, line 2967: Ideally we should introduce full domain-separation, so each category would have its own namespace.
I understand it is simpler not to touch the non-forwarding categories as that would definitely require a wallet DB upgrade -- whereas for forwarding, we can skip DB upgrades atm.

At the very least, please choose a prefix for "trampoline forwarding" that makes it clear it is for *trampoline* forwarding. E.g. `fwd_t`.
We could also add a prefix to regular forwarding now, e.g. `fwd_r`.

electrum/lnworker.py
```diff
@@ -4063,7 +4063,7 @@ def log_fail_reason(reason: str):
         if htlc.amount_msat - next_amount_msat_htlc < forwarding_fees:
             data = next_amount_msat_htlc.to_bytes(8, byteorder="big") + outgoing_chan_upd_message
             raise OnionRoutingFailure(code=OnionFailureCode.FEE_INSUFFICIENT, data=data)
-        if self._maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(htlc.payment_hash):
+        if self.maybe_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(htlc.payment_hash):
```
SomberNight: Does it make sense to have this check in three different places? Can't we just check it a single time in `_check_unfulfilled_htlc`?

tests/test_lnpeer.py
```diff
@@ -1173,6 +1231,197 @@ async def on_htlc_failed(*args):
...
+            nhtlc_success += 1
+            htlc_resolved.set()
+            htlc_resolved.clear()
+        async def on_htlc_failed(*args):
+            nonlocal nhtlc_failed
+            nhtlc_failed += 1
+            htlc_resolved.set()
+            htlc_resolved.clear()
+        util.register_callback(on_htlc_fulfilled, ["htlc_fulfilled"])
+        util.register_callback(on_htlc_failed, ["htlc_failed"])
```
SomberNight: Please review https://github.com/spesmilo/electrum/pull/10904, merge that, and then rebase this and rm the code duplication from here.
Same for other test case you added.

electrum/lnworker.py
ecdsa: indeed. done. this was moved into a separate commit, with the new unit test

tests/test_lnpeer.py
ecdsa: done

electrum/lnworker.py
ecdsa: note: I now realize that this namespace separation does not prevent the attack we discussed. it is certainly cleaner, but I have yet to find a scenario where it would actually prevent something.

tests/test_lnpeer.py
```diff
@@ -1301,6 +1301,79 @@ async def wait_for_htlcs_failed(*, num_htlcs: int):
                 await f()
 
 
+    async def test_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created(self):
+        """Alice holds an invoice created by Bob, hence she knows RHASH and Bob's payment_secret.
+        She sends Bob a dust htlc whose outer onion is addressed to Bob (using the invoice's
+        payment_secret, and claiming a tiny total_msat), but whose inner trampoline onion asks Bob
+        to *forward* the payment.
+        Bob must fail such htlcs, and he must not release the preimage.
+        """
```
SomberNight, line 1310: I am confused about this test.
The commit message says:

> Also rewrite test_refuse_to_forward_htlc_that_corresponds_to_payreq_we_created:
> The new test uses send_trampoline_htlc_to_forward and
> only two peers, so it belongs in the TestPeerDirect class.
> The new test is logically equivalent and 4~5 times faster.

Without the new check in lnpeer.py, the new test fails but the old test passes.
With that, how could the two tests be logically equivalent?

SomberNight (commented):
I merged the first commit into master in https://github.com/spesmilo/electrum/commit/ddd22c8f071ce8835275f767c06f81b0d126d327.

tests/test_lnpeer.py
SomberNight: First I thought that the difference between the two tests is that the old one has the attacker send an htlc with matching `paymenthash` but different `payment_secret`, whereas in the new test both `paymenthash` and `payment_secret` match.

but now I think the whole `payment_secret` stuff and the combined `payment_key` colliding is a complete red herring.

The meaningful difference between the old test and the new test is the timing!
Try with this diff for the old test, now it will fail as the preimage gets released:

```diff
diff --git a/tests/test_lnpeer.py b/tests/test_lnpeer.py
index db86707637..e78afc4429 100644
--- a/tests/test_lnpeer.py
+++ b/tests/test_lnpeer.py
@@ -2123,6 +2123,14 @@ class TestPeerForwarding(TestPeer):
         # now graph is linear: A <-> B <-> D
         graph = self.prepare_chans_and_peers_in_graph(graph_def)
         peers = graph.peers.values()
+
+        w2 = graph.workers['bob']
+        orig_maybe_forward_htlc_set = w2.maybe_forward_htlc_set
+        async def maybe_forward_htlc_set(*args, **kwargs):
+            await asyncio.sleep(0.15)
+            return await orig_maybe_forward_htlc_set(*args, **kwargs)
+        w2.maybe_forward_htlc_set = maybe_forward_htlc_set
+
         async def pay():
             lnaddr1, pay_req1 = self.prepare_invoice(
                 graph.workers['bob'],

```

electrum/lnworker.py
SomberNight: right, the bug has nothing to do with payment_keys colliding. It is a timing race. see https://github.com/spesmilo/electrum/pull/10899#discussion_r3896914929

ecdsa: I removed the namespace separation commit from this PR. 
Not that I am against it, but it seems to only add confusion at this point.

tests/test_lnpeer.py
ecdsa: > Without the new check in lnpeer.py, the new test fails but the old test passes.
> With that, how could the two tests be logically equivalent?

By logically equivalent I mean that they are testing against the same attack. 
The old test only passed because we did not give  lnworker enough time to release the preimage.

ecdsa: Note: Claude is also complaining that the tests are not logically equivalent because the old test had two subtests: trampoline and non-trampoline. Please disregard that. The new check is **upstream** of the trampoline branching, so it does not make sense to test both cases.

PR #10900, opened 2026-08-27
tests: Unskip, repair and make discoverable tests under TestChanReserve

Alwoch:
`TestChanReserve` currently has 3 tests, one skipped (`part_1`) and the remaining two tests  (`part_2` and `part_3`) were not discoverable by pytest since they were written because both were written as normal functions without being prefixed with `test_` This PR makes the tests discoverable, fixes the breaking chages in part 1,2,3 and makes additional changes to make the tests more meaningful.

```shell
❯ pytest tests/test_lnchannel.py -v 
=============================================================== 31 passed, 2 skipped in 3.86s ================================================================

```

### DIFF
--- a/tests/test_lnchannel.py
+++ b/tests/test_lnchannel.py
@@ -865,7 +865,6 @@ class TestChanReserve(ElectrumTestCase):
         self.alice_channel = alice_channel
         self.bob_channel = bob_channel
 
-    @unittest.skip("broken probably because we haven't implemented detecting when we come out of a situation where we violate reserve")
     def test_part1(self):
         # Add an HTLC that will increase Bob's balance. This should succeed,
         # since Alice stays above her channel reserve, and Bob increases his
--- a/tests/test_lnchannel.py
+++ b/tests/test_lnchannel.py
@@ -905,7 +905,7 @@ class TestChanReserve(ElectrumTestCase):
         with self.assertRaises(lnutil.RemoteMisbehaving):
             self.alice_channel.receive_htlc(htlc)
 
-    def part2(self):
+    def test_part2(self):
         paymentPreimage = b"\x01" * 32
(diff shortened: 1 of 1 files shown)


tests/test_lnchannel.py
```diff
@@ -948,31 +947,37 @@ def part3(self):
...
+        # The HTLC is still in-flight, so Bob's balance is unchanged and still
+        # below his channel reserve: he cannot send anything.
+        self.assertEqual(0, self.bob_channel.available_to_spend(LOCAL))
         self.bob_channel.settle_htlc(paymentPreimage, bob_idx)
         self.alice_channel.receive_htlc_settle(paymentPreimage, alice_idx)
         force_state_transition(self.alice_channel, self.bob_channel)
-        self.check_bals(one_bitcoin_in_msat * 3
-                        - self.alice_channel.get_next_fee(LOCAL),
-                        one_bitcoin_in_msat * 7)
+        self.check_bals(one_bitcoin_in_msat * 3, one_bitcoin_in_msat * 7)
```
SomberNight, line 957: Note: this is pre-existing, but it really should be `force_state_transition(self.bob_channel, self.alice_channel)`, as Bob initiates the update (with "settle_htlc").

As-is, your `check_bals()` test only succeeds because `balance_minus_outgoing_htlcs` by default sets `ctn = self.get_next_ctn()`. If it set `ctn = self.get_latest_ctn()`, the checks would fail.

SomberNight (approved):
Looks good, or at least better than the old dysfunctional state! :)
Thanks.

tests/test_lnchannel.py
SomberNight: related https://github.com/spesmilo/electrum/pull/10913

PR #10902, opened 2026-08-27
lnworker: add sanity check to trampoline forwarding

SomberNight:
These new asserts should not be triggered unless there are bugs in the code.
Which there are.
(still the asserts are not intended as primary checks, just defense-in-depth)

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -3954,10 +3954,13 @@ class LNWallet(Logger):
                 min_inc_cltv_abs = min(
                     mpp_htlc.htlc.cltv_abs
                     for mpp_htlc in processed_htlc_set.keys())  # take "min" to assume worst-case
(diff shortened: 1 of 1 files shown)


electrum/lnworker.py
```diff
@@ -4134,6 +4137,7 @@ async def _maybe_forward_trampoline(
                 f"RHASH corresponds to payreq we created. {payment_hash.hex()=}")
             raise OnionRoutingFailure(code=OnionFailureCode.TEMPORARY_NODE_FAILURE, data=b'')
 
+        assert total_msat >= amt_to_forward  # sanity check: money_in >= money_out
```
ecdsa, line 4140: maybe add `closest_inc_cltv_abs >= out_cltv_abs` ?

ecdsa:
superseded by #10929

PR #10903, opened 2026-08-27
type hints: (trivial) add missing "Optional" qualifiers

SomberNight:
In many cases with our type hinting, we do not explicitly specify that variables are allowed to be None when they have a default value of None.
This PR now adds the missing explicit "Optional" markers.

-----

Consider:
```
class Foo(NamedTuple):
    field: str = None
```
This type-hinting is technically incorrect, we should type hint it as either one of these options:
- `field: Optional[str] = None`
- `field: str | None = None`

Thomas said he prefers the second, so `str | None`.
However I am not going to touch lines with existing `Optional`s, I am only fixing lines that are type-hinted incorrectly.

PyCharm has had soft-warnings about these lines for quite a while,
however when doing type-inference, it used to internally "fix" the type by allowing None.
Recently however it seems it is not doing the internal inference-fixing anymore, but assumes that `path` has type `str` (as we specified) and then make "incorrect" inferences:

```py
class Foo(NamedTuple):
    field: str = None

def func(foo: Foo):
    assert foo.field is None
    print("function body")  # <<< PyCharm now says this code is unreachable
```
I have whole function bodies grayed out and marked as dead code due to this.

Example: https://github.com/spesmilo/electrum/blob/c82a0d12ef1e03a96d6180839f215c05efe9efe3/electrum/lnpeer.py#L2320, where htlc_set.parent_set_key is type-hinted as str, so the assert "must" always fail

### DIFF
--- a/electrum/plugins/psbt_nostr/psbt_nostr.py
+++ b/electrum/plugins/psbt_nostr/psbt_nostr.py
@@ -254,7 +254,7 @@ class CosignerWallet(Logger):  # children have to inherit EventListener and regi
         self.known_events[event_id] = now()
         run_sync_function_on_asyncio_thread(self.pending.set, block=False)
 
-    def prepare_messages(self, tx: Union[Transaction, PartialTransaction], label: str = None) -> List[Tuple[str, dict]]:
+    def prepare_messages(self, tx: Union[Transaction, PartialTransaction], label: str | None = None) -> List[Tuple[str, dict]]:
         messages = []
         for xpub, pubkey in self.cosigner_list:
             if not self.cosigner_can_sign(tx, xpub):
@@ -265,7 +265,7 @@ class CosignerWallet(Logger):  # children have to inherit EventListener and regi
             messages.append((pubkey, payload))
         return messages
 
-    def send_psbt(self, tx: Union[Transaction, PartialTransaction], label: str):
+    def send_psbt(self, tx: Union[Transaction, PartialTransaction], label: str | None):
         self.do_send(self.prepare_messages(tx, label), tx.txid())
 
     def do_send(self, messages: List[Tuple[str, dict]], txid: Optional[str] = None):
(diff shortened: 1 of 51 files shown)


electrum/transaction.py
f321x:
```suggestion
        sighash_cache: SighashCache | None = None,
```

Not sure why the suggestion is not rendered correctly, maybe I am using it wrong or Github has a bug.

f321x:
```suggestion
        sighash_cache: SighashCache | None = None,
```

electrum/plugins/psbt_nostr/psbt_nostr.py
f321x:
```suggestion
    def send_psbt(self, tx: Union[Transaction, PartialTransaction], label: str | None):
```

f321x:
lgtm, thanks this is helpful when working with pycharm

SomberNight:
force-pushed https://github.com/spesmilo/electrum/compare/cdfa3d3e596ab32525032871901a62c714c9e18e..9522628b105b929ec9e679ab9eec5a38391ca895 to fix conflicts, had to rebase. I also added your suggestions.

You can check the clean diff ignoring the rebase:
`$ git range-diff c82a0d12ef1e03a96d6180839f215c05efe9efe3..cdfa3d3e596ab32525032871901a62c714c9e18e da98f9a68039c0f1a5e12489210444fc51132f57..9522628b105b929ec9e679ab9eec5a38391ca895`

electrum/transaction.py
SomberNight: done in 5d319a489bb13da2320978f7b09e7c88a53ff8ef

electrum/plugins/psbt_nostr/psbt_nostr.py
SomberNight: done in 5d319a489bb13da2320978f7b09e7c88a53ff8ef

electrum/transaction.py
SomberNight: done in 5d319a489bb13da2320978f7b09e7c88a53ff8ef

PR #10905, opened 2026-08-28
lnpeer: don't log a warning and force-close suggestion for anchor channels we did not initiate

accumulator:
The feerate sanity check for channels initiated by the remote peer should not be done when channel is an anchor channel.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -2487,6 +2487,7 @@ class Peer(Logger, EventListener):
         # want to get kicked out of the mempool immediately if it grows
         fee_buffer = current_feerate_per_kw * 0.5 if chan.has_anchors() else 0
         update_feerate_per_kw = int(current_feerate_per_kw + fee_buffer)
+
         def does_chan_fee_need_update(chan_feerate: Union[float, int]) -> Optional[bool]:
             if chan.has_anchors():
                 # TODO: once package relay and electrum servers with submitpackage are more common,
@@ -2507,8 +2508,9 @@ class Peer(Logger, EventListener):
                 low_fee = min(low_fee, current_feerate_per_kw - FEERATE_PER_KW_MIN_RELAY_LIGHTNING)
                 assert low_fee < high_fee, (low_fee, high_fee)
                 return not (low_fee < chan_feerate < high_fee)
+
         if not chan.constraints.is_initiator:
-            if constants.net is not constants.BitcoinRegtest:
+            if constants.net is not constants.BitcoinRegtest and not chan.has_anchors():
                 chan_feerate = chan.get_latest_feerate(LOCAL)
                 ratio = chan_feerate / update_feerate_per_kw
                 if ratio < 0.5:


electrum/lnpeer.py
SomberNight, line 2523: This check warns if the chan feerate is too low compared to our current desired estimate (`update_feerate_per_kw`).
`update_feerate_per_kw` already takes it into account whether the channel uses anchors, in which case it uses the 1008-block ETA target + 50%.

The reason this gets triggered for you is likely because we lower-bound the 1008-block ETA at 5 sat/vbyte (+50% = 7.5 s/b). That was an arbitrary decision. (ref https://github.com/spesmilo/electrum/pull/9826)

However, note the spirit of the check makes sense also for anchor channels: you can only CPFP the anchor channel ctx if you can get it into the mempool in the first place. So the chan feerate need to be higher than the mempoolminfee. We broadcast the parent and the child transactions separately, so they each need high enough fees to get into the mempool individually.

(except, perhaps we could now use [blockchain.transaction.broadcast_package](https://github.com/spesmilo/electrum-protocol/blob/b3db795b3efd2a0e45a5b3f67fb360b65d4cd4d9/docs/protocol-methods.rst#L788) from electrum protocol 1.6 to do package-relay and broadcast the ctx and the high-fee child as a single package (?). Do you think that would work @f321x, without newer channel types, etc?  -- but we should not spend time working on this *now*! :D)

-----

Anyway, the immediate annoyance here is that the `if ratio < 0.5` check can easily fail because of our 7.5 s/b floor.

ecdsa: why would package-relay require a new channel type?

SomberNight: > why would package-relay require a new channel type?


Some of the cool package relay stuff [only works for nVersion=3 transactions](https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki). The even cooler [generic package relay](https://github.com/bitcoin/bips/blob/master/bip-0331.mediawiki) stuff has not been implemented at all. Some of it does work though for non-v3 txs, it's just hard to remember what exactly. :P

from https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-28.0.md :

> Transactions having a feerate that is too low will be opportunistically paired with their child transactions and submitted as a package, thus enabling the node to download 1-parent-1-child packages using the existing transaction relay protocol. Combined with other mempool policies, this change allows limited "package relay" when a parent transaction is below the mempool minimum feerate. Topologically Restricted Until Confirmation (TRUC) parents are additionally allowed to be below the minimum relay feerate (i.e., pay 0 fees). Use the submitpackage RPC to submit packages directly to the node. Warning: this P2P feature is limited (unlike the submitpackage interface, a child with multiple unconfirmed parents is not supported) and not yet reliable under adversarial conditions.

So looks like nVersion=2 anchor output ctxs combined with submitpackage (in electrum protocol 1.6) could have feerates below mempoolminfee (but must be above the relayfee).

f321x: > perhaps we could now use [blockchain.transaction.broadcast_package](https://github.com/spesmilo/electrum-protocol/blob/b3db795b3efd2a0e45a5b3f67fb360b65d4cd4d9/docs/protocol-methods.rst#L788) from electrum protocol 1.6 to do package-relay and broadcast the ctx and the high-fee child as a single package (?). Do you think that would work @f321x, without newer channel types, etc?

As you say a v2 ctx still has to pay some fee so it can be broadcast in a package, so I think broadcasting packages could mostly be viewed as additional safety mechanism?
But we can't actively lower the feerate to rely on package broadcast either as other implementations probably enforce the ctx getting into the mempool by itself.
Working on the new zero fee ctx seems more beneficial, there we don't have to do the update_fee dance and checks at all which seems like a real simplification.

This is what other implementations seem to enforce (compiled by claude):
| Implementation | Floor for accepting a peer's `update_fee` (anchors) | What it proposes as opener (anchors) | Accept-side code | Propose-side code |
|---|---|---|---|---|
| **lnd** | 253 sat/kw, static | `min(3-block est., 2500 sat/kw)`, floored at its relay fee | `lnwallet/channel.go:3921` (`validateCommitmentSanity`), const `lnwallet/chainfee/rates.go:14` | `lnwallet/channel.go:9964` (`IdealCommitFeeRate`), cap `sample-lnd.conf:492` |
| **CLN** | its own bitcoind `mempoolminfee` (dynamic) | `max(100-block est., 1250 sat/kw)` + opener offset | `lightningd/channel_control.c:93-97`, floor from `plugins/bcli.c:520` | `lightningd/chaintopology.c:644` (`unilateral_feerate`) |
| **eclair** | 253 sat/kw, static | `min(fast, 2500 sat/kw)`, floored at `1.25 × mempoolminfee` | `channel/Commitments.scala:1089` (`receiveFee`), const `blockchain/fee/FeeProvider.scala:107` | `blockchain/fee/OnChainFeeConf.scala` (`getCommitmentFeerate`) |
| **LDK** | `MinAllowedAnchorChannelRemoteFee` (user-supplied, ≥253) | `AnchorChannelFee` (user-supplied, ≥253) | `ln/channel.rs:7896` (`check_remote_fee`), floor `chain/chaininterface.rs:373` | `ln/chan_utils.rs:323` (`selected_commitment_sat_per_1000_weight`) |

PR #10906, opened 2026-08-28
lnchannel: extract_preimage: fix len(witness) check 

f321x:
`Channel.extract_preimage_from_htlc_txin()` would incorrectly assume a htlc revocation spends revocationpubkey to be the preimage of the htlc and then fail at the comparison against the payment hash. 
This happened because the preimage was selected solely based on the length of the spending txin witness which consists of 3 elements both for the preimage spend path and the revocation spend path.

This PR takes over @SomberNight's b0650f1bf7fbf6ba92237644610d6d5e7dc0fe3b commit as fix and adds some additional commits.

### DIFF
--- a/electrum/lnwatcher.py
+++ b/electrum/lnwatcher.py
@@ -14,7 +14,7 @@ from .transaction import Transaction, TxOutpoint
 from .logging import Logger
 from .address_synchronizer import TX_HEIGHT_LOCAL
 from .lnutil import REDEEM_AFTER_DOUBLE_SPENT_DELAY
-from .lnsweep import KeepWatchingTXO, SweepInfo
+from .lnsweep import KeepWatchingTXO, SweepInfo, MaybeSweepInfo
 
 if TYPE_CHECKING:
     from .network import Network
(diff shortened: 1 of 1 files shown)


f321x:
The security review CI finding is addressed by https://github.com/spesmilo/electrum/pull/10910

SomberNight:
Thanks! Looks excellent. I merged all but the last commit, in https://github.com/spesmilo/electrum/commit/d69838959f4f60d528620051cd1a9afdb0a02c38.
I ran out of time before really thinking about the last commit.

electrum/lnwatcher.py
```diff
@@ -201,58 +201,74 @@ async def sweep_commitment_transaction(self, funding_outpoint: str, closing_tx:
...
-                spender_txid=spender_txid,
-                is_local_ctx=is_local_ctx,
-                sweep_info=sweep_info,
-            )
+        for prevout, sweep_info in sweep_info_dict.items():
+            try:
+                keep_watching |= self._sweep_ctx_output(prevout, sweep_info, chan, closing_tx, is_local_ctx)
+            except Exception as e:
+                # in case a single sweep crashes we keep sweeping the other outputs
+                send_exception_to_crash_reporter(e)  # logs too
```
ecdsa: I would remove that; see my comment in the other PR.

f321x: Removed in https://github.com/spesmilo/electrum/pull/10906/commits/df8fd78e3e9ac04a189aeee866a747b922890ce5

ecdsa (approved):

PR #10909, opened 2026-08-28
lnpeer: fix wedged peer connections

accumulator:
On android, when the app is suspended in the background long enough that LN peer sockets die, electrum failed to recover these wedged connections. On desktop, the same issue occurs when suspending the system.

- add timeout on ping messages. avoid wait forever on a dead socket
- ~add timeout on channel_reestablish response message, avoid staying in state REESTABLISHING forever~
- add connection monitor task on Peer taskgroup, similar to Interface.

### DIFF
--- a/electrum/lnpeer.py
+++ b/electrum/lnpeer.py
@@ -225,7 +225,10 @@ class Peer(Logger, EventListener):
         if time.time() - self.last_message_time > 30:
             self.send_message('ping', num_pong_bytes=4, byteslen=4)
             self.pong_event.clear()
-            await self.pong_event.wait()
+            try:
+                await util.wait_for2(self.pong_event.wait(), LN_P2P_NETWORK_TIMEOUT)
+            except asyncio.TimeoutError as e:
+                raise GracefulDisconnect("pong timed out") from e
 
     async def _process_message(self, message: bytes) -> None:
         try:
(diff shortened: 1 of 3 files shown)


electrum/lnpeer.py
```diff
@@ -1725,7 +1736,15 @@ async def reestablish_channel(self, chan: Channel):
...
         fut = self.channel_reestablish_msg[chan_id]
-        await fut
+        try:
+            # future is shielded here, as it is shared with on_channel_reestablish(), which
+            # would raise InvalidStateError when trying to resolve a cancelled future.
+            await util.wait_for2(asyncio.shield(fut), LN_P2P_NETWORK_TIMEOUT)
+        except asyncio.TimeoutError as e:
+            # we must disconnect, so that peer_closed() resets the state to DISCONNECTED.
+            raise GracefulDisconnect(
+                f"channel_reestablish ({chan.get_id_for_log()}): timed out waiting for remote") from e
```
SomberNight: I wonder if we want to enforce reestablishing a channel.
Imagine Alice uses the same electrum seed on a phone and on a laptop. In the typical case of "deterministic lightning", we derive the lightning node id from the seed. Alice's phone could try to open channel1 with remote node1, and Alice's laptop could try to open channel2 with same remote node1.

Now imagine node1 is running this code. node1 thinks it has two channels with Alice, kind of rightly so. Whenever either of Alice's devices connect, node1 will try to reestablish both channels, and then disconnect after LN_P2P_NETWORK_TIMEOUT.

note that Alice's devices simply ignore received `channel_reestablish` messages for unknown channels:
https://github.com/spesmilo/electrum/blob/c82a0d12ef1e03a96d6180839f215c05efe9efe3/electrum/lnpeer.py#L258-L262

Anyway, I agree the whole situation I am describing is a mess... and untested. Could you split out this change from the rest? The rest look good.

accumulator: removed the `channel_reestablish` part

SomberNight:
Thanks!

PR #10910, opened 2026-08-28
crash reporter: keep queue bounded, log enqueued exceptions

f321x:
* Keep the `EarlyExceptionsQueue` bounded so it doesn't leak memory on Electrum processes without registered exc hook (e.g. when running as daemon).
* Log exceptions enqueued on the `EarlyExceptionsQueue` so they don't go silent on processes without exc hook if there is no additional logging. Otherwise we'd need to use a double "log first, then send to crash reporter" pattern which logs the exception twice if a GUI exc hook was registered (as the crash reporter itself logs again).
~* Forward `check_onchain_situation` exceptions to the crash reporter so we can learn from the userbase executing more code paths and don't have to rely on developers looking at the logs to discover bugs~

### DIFF
--- a/electrum/base_crash_reporter.py
+++ b/electrum/base_crash_reporter.py
@@ -32,7 +32,7 @@ from .version import ELECTRUM_VERSION
 from . import constants
 from .i18n import _
 from .util import make_aiohttp_session, error_text_str_to_safe_str
-from .logging import describe_os_version, Logger, get_git_version
+from .logging import describe_os_version, Logger, get_logger, get_git_version
 from .crypto import sha256
 
 if TYPE_CHECKING:
@@ -221,7 +221,8 @@ class EarlyExceptionsQueue:
     """
 
     _is_exc_hook_ready = False
-    _exc_queue = queue.Queue()
+    _exc_queue = queue.Queue(maxsize=100)
+    _logger = get_logger(__name__)
 
     @classmethod
     def set_hook_as_ready(cls):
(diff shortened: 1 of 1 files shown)


electrum/lnwatcher.py
```diff
@@ -134,8 +134,7 @@ def add_channel(self, chan: 'AbstractChannel') -> None:
         callback = lambda: self.check_onchain_situation(address, outpoint)
         self.add_callback(address, callback, subscribe=chan.need_to_subscribe())
 
-    @ignore_exceptions
-    @log_exceptions
+    @report_exceptions
     async def check_onchain_situation(self, address: str, funding_outpoint: str) -> None:
```
f321x: I think it should be ok to forward the exceptions to the crash reporter unredacted (in opposition to the CI review's complaint)..?
Alternatively we could also redact the exception similar to the way it's done in the htlc switch:
https://github.com/spesmilo/electrum/blob/66f73cb5ba9f8fbe17258ce96c35f1efdd6500cc/electrum/lnpeer.py#L2866-L2871

electrum/util.py
```diff
@@ -1254,6 +1254,21 @@ async def wrapper(*args, **kwargs):
     return wrapper
 
 
+def report_exceptions(func):
+    """Decorator to send exceptions to the crash reporter and swallow them.
```
ecdsa: I don't really like the idea of introducing a new decorator for handling exceptions. It makes things difficult to reason about, and it violates separation between levels, because it is a GUI-only behavior.

IMO `ignore_exceptions` and `log_exceptions` already cover all the logic we should care about at the code level. Whether the exception should propagate to the crash reporter should not be decided by every method.

How about sending to the crash reporter everything that goes into the log? it might sound a bit extreme, but in the long run I think it would make things simpler.

f321x: I see the point, reporting exceptions at the logger level sounds nice as it would work in a more generic way. I will have a look at this. 
I removed the decorator commit from this PR so only the uncontroversial `EarlyExceptionsQueue` change is left.

PR #10912, opened 2026-08-28
fee_policy: don't quantize explicitly set feerates

sashazykov:
`estimate_fee()` rounds the feerate to `FEERATE_PRECISION` (1 decimal place, half-down) before computing the fee, so `payto --feerate 0.45` pays 0.4 sat/vB and `--feerate 3.75` (the example from #4259) pays 3.7, with no indication that the requested rate was not used.

The rounding was added in f12798e91 for consistency with the GUI display, but both GUIs already constrain feerate input to one decimal place before the policy is constructed, so for `FeeMethod.FEERATE` it only affected CLI- and
config-originated rates. ETA and mempool based estimates keep being quantized. Since 58af1c493 lowered the supported relay floor to 0.1 sat/vB, one-decimal granularity is coarsest exactly where these rates now live.

Note: `bump_fee`/`dscancel` have their own `quantize_feerate` call (which also feeds the new-vs-old rate sanity check); deliberately left untouched here.

### DIFF
--- a/electrum/fee_policy.py
+++ b/electrum/fee_policy.py
@@ -265,7 +265,9 @@ class FeePolicy(Logger):
             else:
                 raise NoDynamicFeeEstimates()
 
-        return self.estimate_fee_for_feerate(fee_per_kb=fee_per_kb, size=size)
+        # a feerate explicitly set by the user is honored at its full precision
+        quantize = self.method != FeeMethod.FEERATE
+        return self.estimate_fee_for_feerate(fee_per_kb=fee_per_kb, size=size, quantize=quantize)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


sashazykov:
Tested on mainnet: `paytomany --feerate 0.315 --unsigned` on a hardware wallet built a 532 vB transaction paying 168 sat = ceil(0.315 × 532), signed and broadcast fine. Without this patch the same command pays 160 sat (0.315 rounded to 0.3).

f321x:
lgtm [44c9c47](https://github.com/spesmilo/electrum/pull/10912/commits/44c9c47ccbd5b00493ccf9ccd77b31f805cecfde), thanks for the contribution.

PR #10913, opened 2026-08-29
tests: lnchan: clarify "force_state_transition" function

SomberNight:


tests/test_lnchannel.py
f321x: This seems like a (preexisting) copy paste mistake

```suggestion
        self.assertEqual(1000000000, bob_channel.available_to_spend(REMOTE))
```

SomberNight: indeed. fixed in https://github.com/spesmilo/electrum/pull/10913/commits/17e194f009e0beb4154d487a6ed06471695e5778

PR #10914, opened 2026-08-29
lnhtlc: add sanity check we don't revoke our last valid commitment

SomberNight:
adds asserts, and some test cases


ecdsa:
Nice. CTN drift has never been tested :-)

PR #10932, opened 2026-09-04
qt: qr code reader: speed up camera

### LINKED ISSUE #10380: Apple M4 iMac - Internal Camera Glitches with Electrum 4.6.2 using OS Tahoe 26.2
### Description

**Apple M4 iMac - Internal Camera Glitches with Electrum 4.6.2 using OS Tahoe 26.2** 

I just bought a new Apple iMac (M4 chips) in December  2025. It uses OS Tahoe 26.0.1. I installed Electrum 4.6.2 and all was working well until I tried to sweep a paper wallet using the QR code scan with the Mac internal camera. The camera freezes up and then Electrum stops responding. I checked the camera  using FaceTime and it functions perfectly. I was so disappointed. Please consider an update for Electrum's camera scan QR code functionality when using Tahoe 26. THANK YOU so much for making Electrum. 

**EDITED:** Friday December 19 at 11:01AM EST:
I think perhaps Apple may have changed something with the front facing MAC camera code in OS Tahoe 26.0.1 - example: image mirroring. That's just a guess and doesn't really explain why the camera window in Electrum 4.6.2 hangs, glitches and stalls.

**EDITED:** Saturday December 20 at 9:22AM EST:
I thought about a work around solution - that being to install a QR code reader App on my Mac and use it to read and  then paste addresses into the Electrum window. I intend to report the issue to Apple as well since not hearing much on here. Updated: Google A.I. returns the following result: "Your Electrum app glitches with the camera on macOS Tahoe (version 26) because of a known Electron framework bug where its use of a private Apple API causes GPU overload, leading to stuttering/freezing in Electron apps like Electrum when using the camera. The fix involves waiting for Electrum to update to a patched Electron version or using the temporary Photo Booth workaround (opening it in the background) until the app developer releases an update for Tahoe."
https://9to5mac.com/2025/11/21/mac-tahoe-electron-performance-bug/ 

**EDITED:** Sunday December 21 at 9:03AM EST:
Leaving this comment here for the search engines: I updated to Mac OS Tahoe 26.2 in which Apple attempted to address the latency in third party Apps interface with Electron coding. However, even after the update install the Electrum wallet App still is unable to  scan and read QR codes without becoming unresponsive and still has the camera lag latency glitch issue original described above. As an aside - the Sparrow Wallet does not have the issue that Electrum has.

Could someone comment if this Electrum wallet issue is being worked on and if so whether an Electrum App update may happen - so people searching this same issue will know?

### LINKED ISSUE #10928: [Regression][Windows][Camera] QR camera preview severely stutters since Electrum 4.6.x; 4.5.8 is smooth on the same hardware
### Description

<html><head></head><body><h2>Description</h2><p>There appears to be a significant regression in Electrum's QR-code camera preview on Windows starting somewhere between Electrum 4.5.8 and 4.6.x.</p><p>The camera preview is very smooth in Electrum 4.5.8 on the same hardware, but becomes severely laggy/stuttery in Electrum 4.6.2 and remains the same in Electrum 4.8.1.</p><p>I have reproduced this on two different Windows laptops.</p><p>Importantly, the lag occurs even when there is <strong>no QR code visible in the camera image</strong>, so the problem does not appear to depend on QR decoding itself.</p><h2>Test results</h2><h3>Laptop 1</h3><p><strong>Lenovo ThinkBook 14 G4</strong></p><p>Same laptop, same built-in camera, same Windows installation and camera driver:</p>
Application / Version | Camera preview
-- | --
Windows Camera app | Smooth
Electrum 4.5.8 | Smooth
Electrum 4.6.2 | Severely laggy
Electrum 4.8.1 | Severely laggy

<p>So the issue does not appear to be specific to the Lenovo camera or driver.</p><h2>Important observations</h2><ol><li><p>The camera preview is already severely laggy when pointed at a blank scene with <strong>no QR code present</strong>.</p></li><li><p>Windows' built-in Camera application is smooth on both machines.</p></li><li><p>CPU usage increases only slightly when the Electrum camera window is opened. The CPU is not close to being saturated.</p></li><li><p>The problem reproduces on two different laptop models from different manufacturers.</p></li><li><p>Electrum 4.5.8 works smoothly on the exact same Lenovo laptop where 4.6.2 and 4.8.1 are severely laggy.</p></li></ol><p>This makes a Windows camera-driver-only problem seem unlikely.</p><h2>Steps to reproduce</h2><ol><li><p>Start Electrum on Windows.</p></li><li><p>Open any function that starts the QR-code camera scanner.</p></li><li><p>Point the camera at a normal scene or even a blank wall. No QR code is required.</p></li><li><p>Observe the live camera preview.</p></li></ol><h3>Expected behavior</h3><p>Camera preview should be reasonably smooth, similar to Electrum 4.5.8 and the Windows Camera application.</p><h3>Actual behavior</h3><p>The camera preview runs at a very low apparent frame rate and stutters heavily.</p><p>QR scanning may still eventually work, but positioning a QR code in front of the camera becomes unnecessarily difficult because of the very low preview frame rate.</p><h2>Regression / Qt6 migration</h2><p>Electrum 4.6.0 migrated the Qt desktop GUI from Qt5 to Qt6:</p><p><a href="https://github.com/spesmilo/electrum/issues/8007">https://github.com/spesmilo/electrum/issues/8007</a></p><p>PR:</p><p><a href="https://github.com/spesmilo/electrum/pull/9189">https://github.com/spesmilo/electrum/pull/9189</a></p><p>Since Electrum 4.5.8 is smooth and 4.6.2 is not, it may be worth investigating whether the regression is related to the Qt5 → Qt6 / QtMultimedia camera migration.</p><p>The current Qt GUI also uses <code inline="">PyQt6.QtMultimedia</code> for camera handling.</p><p>Possible areas worth comparing between 4.5.8 and 4.6.x include:</p><ul><li><p>camera format selection</p></li><li><p>capture resolution selection</p></li><li><p>pixel format conversion</p></li><li><p>frame-rate negotiation</p></li><li><p><code inline="">QCamera</code> / <code inline="">QCameraFormat</code></p></li><li><p><code inline="">QVideoSink</code></p></li><li><p><code inline="">QVideoFrame.toImage()</code></p></li><li><p>whether Qt6 selects a much higher resolution or less efficient pixel format than the old Qt5 implementation</p></li><li><p>whether every incoming camera frame is being converted synchronously before display</p></li></ul><p>It might also be useful to explicitly select a lower-resolution camera format, such as 640×480 or 1280×720, rather than relying on the QtMultimedia default camera format.</p><p>I am not claiming that Qt6 itself is necessarily the root cause, but the version comparison makes the Qt6 migration a useful place to investigate.</p><h2>Related issue</h2><p>This may also be related to:</p><p>#10380 — Apple M4 iMac internal camera glitches with Electrum 4.6.2</p><p><a href="https://github.com/spesmilo/electrum/issues/10380">https://github.com/spesmilo/electrum/issues/10380</a></p><p>That report is on macOS rather than Windows, but there are some notable similarities:</p><ul><li><p>system/native camera application works normally</p></li><li><p>Electrum camera preview stalls or lags</p></li><li><p>the issue is present in Electrum 4.6.2</p></li><li><p>another Bitcoin wallet reportedly does not have the same camera issue</p></li></ul><p>The Windows regression described here may therefore help determine whether #10380 is actually a broader QtMultimedia/Electrum camera regression rather than being specific to macOS.</p><h2>Summary</h2><p>The most useful comparison is:</p><p><strong>Same Lenovo ThinkBook 14 G4 + same built-in camera + same Windows environment</strong></p><ul><li><p>Electrum 4.5.8 → smooth</p></li><li><p>Electrum 4.6.2 → severely laggy</p></li><li><p>Electrum 4.8.1 → severely laggy</p></li><li><p>Windows Camera → smooth</p></li></ul><p>And the problem is also reproducible with Electrum 4.8.1 on a Dell Inspiron 5593.</p><p>I would be happy to perform additional tests if the developers would like specific camera-format, resolution, logging, or debug information.</p></body></html>

f321x:
I tried the Qt QR camera on a slow windows VM. Electrum freezes and the scanner is unusable.
With these changes it is snappy again.

* Only keep the most recent camera frame in memory
* specify smaller camera format instead of giving choice to qt (qt picks high resolution) 
* some cleanup from the qt5->qt6 transition

Fixes https://github.com/spesmilo/electrum/issues/10380
Fixes https://github.com/spesmilo/electrum/issues/10928

### DIFF
--- a/electrum/gui/qt/qrreader/qtmultimedia/video_surface.py
+++ b/electrum/gui/qt/qrreader/qtmultimedia/video_surface.py
@@ -23,13 +23,12 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from typing import List
+from typing import Optional
 
-from PyQt6.QtMultimedia import (QVideoFrame, QVideoFrameFormat, QVideoSink)
+from PyQt6.QtMultimedia import QVideoFrame, QVideoSink
 from PyQt6.QtGui import QImage
-from PyQt6.QtCore import QObject, pyqtSignal
+from PyQt6.QtCore import QObject, QTimer, pyqtSignal
 
-from electrum.i18n import _
 from electrum.logging import get_logger
 
 
@@ -38,44 +37,41 @@ _logger = get_logger(__name__)
 
 class QrReaderVideoSurface(QVideoSink):
     """
-    Receives QVideoFrames from QCamera, converts them into a QImage, flips the X and Y axis if
-    necessary and sends them to listeners via the frame_available event.
+    Receives QVideoFrames from QCamera, converts the newest one into a QImage and
+    sends it to listeners via the frame_available signal.
     """
+
+    frame_available = pyqtSignal(QImage)
 
     def __init__(self, parent: QObject = None):
         super().__init__(parent)
+        self._pending_frame: Optional[QVideoFrame] = None
+        self._process_timer = QTimer(self)
+        self._process_timer.setSingleShot(True)
+        self._process_timer.setInterval(0)
+        self._process_timer.timeout.connect(self._process_pending_frame)
         self.videoFrameChanged.connect(self._on_new_frame)
 
     def _on_new_frame(self, frame: QVideoFrame) -> None:
         if not frame.isValid():
             return
+        # only keep the newest frame
+        self._pending_frame = QVideoFrame(frame)  # keep our own reference (the received frame is owned by Qt)
+        if not self._process_timer.isActive():
+            self._process_timer.start()
 
-        image_format = QVideoFrameFormat.imageFormatFromPixelFormat(frame.pixelFormat())
-        if image_format == QVideoFrameFormat.PixelFormat.Format_Invalid:
-            _logger.info(_('QR code scanner for video frame with invalid pixel format'))
+    def _process_pending_frame(self) -> None:
+        frame, self._pending_frame = self._pending_frame, None
+        if frame is None:
             return
-
         if not frame.map(QVideoFrame.MapMode.ReadOnly):
-            _logger.info(_('QR code scanner failed to map video frame'))
+            _logger.warning(f"failed to map video frame. pixel format: {frame.pixelFormat()}", only_once=True)
             return
-
         try:
             img = frame.toImage()
-
-            # Check whether we need to flip the image on any axis
-            surface_format = frame.surfaceFormat()
-            flip_x = surface_format.isMirrored()
-            flip_y = surface_format.scanLineDirection() == QVideoFrameFormat.Direction.BottomToTop
-
-            # Mirror the image if needed
-            if flip_x or flip_y:
-                img = img.mirrored(flip_x, flip_y)
-
-            # Create a copy of the image so the original frame data can be freed
-            img = img.copy()
         finally:
             frame.unmap()
-
+        if img.isNull():
+            _logger.warning(f"failed to convert video frame to image. pixel format: {frame.pixelFormat()}", only_once=True)
+            return
         self.frame_available.emit(img)
-
-    frame_available = pyqtSignal(QImage)
--- a/electrum/gui/qt/qrreader/qtmultimedia/camera_dialog.py
+++ b/electrum/gui/qt/qrreader/qtmultimedia/camera_dialog.py
@@ -192,12 +192,13 @@ class QrReaderCameraDialog(Logger, MessageBoxMixin, QDialog):
             self.logger.info("Warning: start_scan already called for this instance.")
 
         self.camera = QCamera(device_info)
-        self.camera.start()
         self.camera.errorOccurred.connect(self._on_camera_error)  # log the errors we get, if any, for debugging
 
         self.media_capture_session = QMediaCaptureSession()
         self.media_capture_session.setCamera(self.camera)
         self.media_capture_session.setVideoSink(self.video_surface)
+
+        self.camera.start()
 
         self.open()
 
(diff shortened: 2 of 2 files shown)


ecdsa:
Looks good. Tested on MacOS.

electrum/gui/qt/qrreader/qtmultimedia/video_surface.py
```diff
@@ -38,44 +37,41 @@
...
-            _logger.info(_('QR code scanner failed to map video frame'))
+            _logger.warning(f"failed to map video frame. pixel format: {frame.pixelFormat()}", only_once=True)
             return
-
         try:
             img = frame.toImage()
-
-            # Check whether we need to flip the image on any axis
-            surface_format = frame.surfaceFormat()
-            flip_x = surface_format.isMirrored()
```
ecdsa, line 67: what was this logic doing and why can it be removed?

f321x: In Qt5 the `QImage` was constructed from the raw pixels (`int(frame.bits()`) and did not have information about it's orientation:
https://github.com/spesmilo/electrum/blob/88a1c1a618062db0ad49ba3b98d6e47e9e80d5d9/electrum/gui/qt/qrreader/qtmultimedia/video_surface.py#L63
It had to be flipped manually so the preview doesn't show it e.g. upside down (depending on the camera), zbar itself should be able to handle frames of either orientation.

Now, since the Qt6 introduction, the `QImage` is being constructed directly from the `QVideoFrame` object using its `.toImage()` method.
The `.toImage()` method does the same flipping internally:
https://github.com/qt/qtmultimedia/blob/v6.9.2/src/multimedia/video/qvideoframe.cpp#L862
https://github.com/qt/qtmultimedia/blob/v6.9.2/src/multimedia/qmultimediautils.cpp#L106

PR #10933, opened 2026-09-05
prep crandom.py: trivial refactor to centralise our RNG code

SomberNight:
prep commits for https://github.com/spesmilo/electrum/pull/10794 to make that less invasive

### DIFF
--- a/electrum/commands.py
+++ b/electrum/commands.py
@@ -79,6 +79,7 @@ from . import GuiImportError
 from . import crypto
 from . import constants
 from . import descriptor
+from . import crandom
 
 if TYPE_CHECKING:
     from .network import Network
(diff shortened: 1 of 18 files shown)


ekzyis (commented):
I noticed not all occurrences of `randrange(x)` were replaced with `crandom.get_rand_below(x - 1) + 1` like in tests/test_bitcoin.py:176. Is that intentional?

SomberNight:
> not all occurrences of `randrange(x)` were replaced with `crandom.get_rand_below(x - 1) + 1` like in tests/test_bitcoin.py:176. Is that intentional?

It is. See the respective commit message:

> util: rm randrange function. just use crandom.get_rand_below
>
> The only difference was that randrange never generated zero,
> while get_rand_below does. Most consumers do not care.
> This behaviour originally came from python-ecdsa, where the
> use case was to generate secret scalars for ecc private keys,
> and zero would generate the point at infinity.

There were only three call sites of `randrange()`:
- in `tests/test_bitcoin.py`, it is used to generate a secret scalar for an ecc private key, so `0` would not be valid
- in `daemon.py`, it is used to generate a password, `0` is fine
- in `mnemonic.py` (`make_seed`), the "try again if seed would not contain enough words" while loop nukes all small values anyway
    - but even if that was not there, `mnemonic_encode()` can handle 0 (though the nonce starts at 1 anyway)

Though in practice, all these call sites want random numbers so large that generating the previously not allowed 0 is quite unlikely.

PR #10935, opened 2026-09-05
lnworker: dont fail invoice with unresolved htlcs

ecdsa:

### DIFF
--- a/electrum/lnhtlc.py
+++ b/electrum/lnhtlc.py
@@ -128,6 +128,17 @@ class HTLCManager:
         self.log[LOCAL]['fails'][htlc_id] = {LOCAL: next_ctn, REMOTE: None}
 
     @with_lock
+    def mark_htlc_resolved_onchain(self, *, htlc_proposer: HTLCOwner, htlc_id: int, success: bool) -> None:
+        # simulate the removal of the htlc
+        # so that lnworker.has_unresolved_sent_htlcs() does not return True
+        if self.was_htlc_failed(htlc_id=htlc_id, htlc_proposer=htlc_proposer):
+            return
+        if self.was_htlc_preimage_released(htlc_id=htlc_id, htlc_proposer=htlc_proposer):
+            return
+        MAX_CTN = 2 ** 48 + 1  # means that it was settled offchain
+        self.log[htlc_proposer]['settles' if success else 'fails'][htlc_id] = {htlc_proposer: MAX_CTN, htlc_proposer.inverted(): None}
+
+    @with_lock
     def send_update_fee(self, feerate: int) -> None:
         fee_update = FeeUpdate(rate=feerate,
                                ctn_local=None, ctn_remote=self.ctn_latest(REMOTE) + 1)
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -3234,6 +3234,28 @@ class LNWallet(Logger):
             upstream_peer.downstream_htlc_resolved_event.set()
             upstream_peer.downstream_htlc_resolved_event.clear()
 
+    def _set_sent_payment_succeeded(self, payment_hash: bytes) -> None:
+        key = payment_hash.hex()
+        info = self.get_payment_info(payment_hash, direction=SENT)
+        if info is not None and info.status != PR_PAID:
+            self.set_invoice_status(key, PR_PAID)
+            util.trigger_callback('payment_succeeded', self.wallet, key)
+
+    def _set_sent_payment_failed(self, payment_hash: bytes) -> None:
+        key = payment_hash.hex()
+        if self.has_unresolved_sent_htlcs(payment_hash):
+            return
+        if self.get_preimage(payment_hash) and self.wallet.get_request(key) is None:
+            # if we know the preimage don't consider the payment failed (unless we pay ourselves).
+            # maybe another htlc of the same mpp got fulfilled, or we saw a htlc-success tx in the mempool
+            # before claiming a revoked htlc with a justice tx which ultimately would make LNWatcher try to fail the htlc here
+            return
+        info = self.get_payment_info(payment_hash, direction=SENT)
+        if info is not None and (info.status == PR_UNPAID and key in self.inflight_payments):
+            # invoice status is PR_INFLIGHT, set it to PR_UNPAID
+            self.set_invoice_status(key, PR_UNPAID)
+            util.trigger_callback('payment_failed', self.wallet, key, '')
+
     def htlc_fulfilled(self, chan: Channel, payment_hash: bytes, htlc_id: int):
         """Called when an HTLC *WE proposed* becomes irrevocably fulfilled."""
         # note: this may be called several times for the same htlc
@@ -3265,15 +3287,13 @@ class LNWallet(Logger):
                 paysession_active = False
             else:
                 paysession_active = True
+            if not fw_key and not paysession.is_active:
+                self._set_sent_payment_succeeded(payment_hash)
         else:
             if fw_key:
                 paysession_active = False
             else:
-                key = payment_hash.hex()
-                info = self.get_payment_info(payment_hash, direction=SENT)
-                if info is not None and info.status != PR_PAID:
-                    self.set_invoice_status(key, PR_PAID)
-                    util.trigger_callback('payment_succeeded', self.wallet, key)
+                self._set_sent_payment_succeeded(payment_hash)
 
         if fw_key:
             fw_htlcs = self.active_forwardings[fw_key]
@@ -3344,20 +3364,14 @@ class LNWallet(Logger):
                 paysession_active = False
             else:
                 paysession_active = True
+            if not fw_key and not paysession.is_active:
+                self._set_sent_payment_failed(payment_hash)
         else:
             if fw_key:
                 paysession_active = False
             else:
                 self.logger.info(f"received unknown htlc_failed, probably from previous session (phash={payment_hash.hex()})")
-                key = payment_hash.hex()
-                invoice = self.wallet.get_invoice(key)
-                if invoice and self.get_invoice_status(invoice) != PR_UNPAID \
-                        and (self.get_preimage(payment_hash) is None or self.wallet.get_request(key) is not None):
-                    # if we know the preimage don't consider the payment failed (unless we pay ourselves).
-                    # maybe another htlc of the same mpp got fulfilled, or we saw a htlc-success tx in the mempool
-                    # before claiming a revoked htlc with a justice tx which ultimately would make LNWatcher try to fail the htlc here
-                    self.set_invoice_status(key, PR_UNPAID)
-                    util.trigger_callback('payment_failed', self.wallet, key, '')
+                self._set_sent_payment_failed(payment_hash)
 
         if fw_key:
             fw_htlcs = self.active_forwardings[fw_key]
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1316,6 +1316,12 @@ class LNWallet(Logger):
                 out[payment_hash] += plist
         return out
 
+    def has_unresolved_sent_htlcs(self, payment_hash: bytes) -> bool:
+        """Returns whether there are htlcs we sent for this payment that have neither
+        been failed nor fulfilled yet, i.e. the receiver might still take the money.
+        """
+        return payment_hash in self.get_payments(status='inflight', direction=SENT)
+
     def get_payment_value(
             self, sent_info: Optional['PaymentInfo'],
             plist: List[HTLCWithStatus]
@@ -1960,7 +1966,7 @@ class LNWallet(Logger):
             raise PaymentFailure(_("This invoice has been paid already"))
         if status == PR_INFLIGHT:
             raise PaymentFailure(_("A payment was already initiated for this invoice"))
-        if payment_hash in self.get_payments(status='inflight'):
+        if self.has_unresolved_sent_htlcs(payment_hash):
             raise PaymentFailure(_("A previous attempt to pay this invoice did not clear"))
         info = PaymentInfo(
             payment_hash=payment_hash,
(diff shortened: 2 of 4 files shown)


ecdsa:
related: PAY-1, PAY-2

tests/test_lnpeer.py
```diff
@@ -915,6 +915,58 @@ async def f():
...
+            self.assertTrue(w1.has_unresolved_sent_htlcs(payment_hash))
+            # the money is still at risk, so the invoice must not look failed/unpaid
+            self.assertEqual(PR_INFLIGHT, w1.get_invoice_status(pay_req))
+            with self.assertRaises(PaymentFailure):
+                await w1.pay_invoice(pay_req)
+            # now let bob fulfill the htlc. nobody is waiting for the payment anymore,
+            # but the invoice must still end up as paid.
+            w2.enable_htlc_settle = True
+            while w1.get_invoice_status(pay_req) != PR_PAID:
+                await asyncio.sleep(0.01)
```
SomberNight, line 966: nit: add a timeout so that a failing test does not hang forever
```diff
diff --git a/tests/test_lnpeer.py b/tests/test_lnpeer.py
index 502b6d809c..13005ac6b0 100644
--- a/tests/test_lnpeer.py
+++ b/tests/test_lnpeer.py
@@ -951,8 +951,9 @@ class TestPeerDirect(TestPeer):
             # now let bob fulfill the htlc. nobody is waiting for the payment anymore,
             # but the invoice must still end up as paid.
             w2.enable_htlc_settle = True
-            while w1.get_invoice_status(pay_req) != PR_PAID:
-                await asyncio.sleep(0.01)
+            async with util.async_timeout(5):
+                while w1.get_invoice_status(pay_req) != PR_PAID:
+                    await asyncio.sleep(0.01)
             raise SuccessfulTest()
 
         async def f():

```

electrum/lnworker.py
```diff
@@ -3328,20 +3348,14 @@ def htlc_failed(
                 paysession_active = False
             else:
                 paysession_active = True
+            if not fw_key and not paysession.is_active:
+                self._set_sent_payment_failed(payment_hash)
```
SomberNight, line 3368: `paysession.is_active` is not the same as `paysession_active` (the latter is `== not paysession.can_be_deleted()`).

This test probably needs the stricter `not paysession_active` check.

SomberNight: Ok, turns out I was wrong, as `_set_sent_payment_failed` is more aptly named `_maybe_set_*` :)

I wrote a unit test before I realised I was wrong, to showcase the issue I believed was there:

```diff
diff --git a/tests/test_lnpeer.py b/tests/test_lnpeer.py
index 502b6d809c..d67d296f37 100644
--- a/tests/test_lnpeer.py
+++ b/tests/test_lnpeer.py
@@ -44,9 +44,9 @@ from electrum.lnworker import LNWallet, NoPathFound, SentHtlcInfo, PaySession, L
 from electrum.lnmsg import encode_msg, decode_msg
 from electrum import lnmsg
 from electrum.logging import console_stderr_handler, Logger
-from electrum.lnonion import OnionFailureCode, OnionRoutingFailure, OnionHopsDataSingle, OnionPacket
+from electrum.lnonion import OnionFailureCode, OnionRoutingFailure, OnionHopsDataSingle, OnionPacket, OnionParsingError
 from electrum.lnutil import LOCAL, REMOTE, UpdateAddHtlc, RecvMPPResolution, RevocationStore
-from electrum.invoices import PR_PAID, PR_UNPAID, PR_INFLIGHT, Invoice
+from electrum.invoices import PR_PAID, PR_UNPAID, PR_INFLIGHT, Invoice, PR_FAILED
 from electrum.interface import GracefulDisconnect
 from electrum.fee_policy import FeeTimeEstimates, FEE_ETA_TARGETS
 from electrum.mpp_split import split_amount_normal
@@ -915,11 +915,12 @@ class TestPeerDirect(TestPeer):
         for _test_trampoline in [False, True]:
             await run_test(_test_trampoline)
 
-    async def test_invoice_stays_inflight_while_htlcs_unresolved(self):
+    async def test_invoice_stays_inflight_while_htlcs_unresolved_then_paid(self):
         """Tests that we don't mark an invoice as failed while htlcs we sent for it are still
         unresolved. The receiver can still fulfill those htlcs, so telling the user the payment
         failed would trick them into paying twice, e.g. using a fresh invoice, which the
         "did not clear" check in pay_invoice cannot detect as a retry.
+        After we give up on a paysession, the remote fulfills the htlc, and we mark the invoice PAID.
         """
         graph = self.prepare_chans_and_peers_in_graph(self.GRAPH_DEFINITIONS['single_chan'])
         p1, p2 = graph.peers.values()
@@ -967,6 +968,62 @@ class TestPeerDirect(TestPeer):
         with self.assertRaises(SuccessfulTest):
             await f()
 
+    async def test_invoice_stays_inflight_while_htlcs_unresolved_then_failed(self):
+        """Tests that we don't mark an invoice as failed while htlcs we sent for it are still
+        unresolved. After we give up on paysession, we keep receiving htlc failures.
+        Only after the very last htlc failure is received shall we mark the invoice as FAILED.
+        """
+        graph = self.prepare_chans_and_peers_in_graph(self.GRAPH_DEFINITIONS['single_chan'])
+        p1, p2 = graph.peers.values()
+        w1, w2 = graph.workers.values()
+        w1.config.TEST_FORCE_MPP = True  # force alice to send mpp
+        w2.features |= LnFeatures.BASIC_MPP_OPT
+        w2.enable_htlc_settle = False  # bob holds the htlc: he neither settles nor fails it
+        lnaddr, pay_req = self.prepare_invoice(w2)
+        self.assertTrue(lnaddr.get_features().supports(LnFeatures.BASIC_MPP_OPT))
+        payment_hash = lnaddr.paymenthash
+        bob_channel = graph.channels[('bob', 'alice')][0]
+
+        def bobs_pending_htlcs():
+            return bob_channel.hm.get_htlcs_in_latest_ctx(HTLCOwner.LOCAL)
+
+        async def pay():
+            # Alice starts a payment attempt
+            task = asyncio.create_task(w1.pay_invoice(pay_req))
+            while len(bobs_pending_htlcs()) < 2:
+                await p1.received_commitsig_event.wait()
+            # but she quickly gives up. with htlcs left pending
+            task.cancel()
+            self.assertTrue(w1.has_unresolved_sent_htlcs(payment_hash))
+            # the money is still at risk, so the invoice must not look failed/unpaid
+            self.assertEqual(PR_INFLIGHT, w1.get_invoice_status(pay_req))  # money is still at risk, hence PR_INFLIGHT
+            # Bob fails one HTLC, but keeps the rest pending (using "fail_malformed" as it's easier to simulate here)
+            pending_htlcs_orig = bobs_pending_htlcs()
+            assert len(pending_htlcs_orig) >= 2
+            p2.fail_malformed_htlc(chan=bob_channel, htlc_id=0, reason=OnionParsingError(data=bytes(32)))
+            await p2.received_commitsig_event.wait()
+            pending_htlcs = bobs_pending_htlcs()
+            assert len(pending_htlcs) == len(pending_htlcs_orig) - 1
+            self.assertEqual(PR_INFLIGHT, w1.get_invoice_status(pay_req))
+            # Bob fails all HTLCs, now Alice can safely mark the invoice as FAILED
+            for htlc_dir, htlc in pending_htlcs:
+                p2.fail_malformed_htlc(chan=bob_channel, htlc_id=htlc.htlc_id, reason=OnionParsingError(data=bytes(32)))
+            await p2.received_commitsig_event.wait()
+            self.assertEqual(PR_FAILED, w1.get_invoice_status(pay_req))
+            raise SuccessfulTest()
+
+        async def f():
+            async with OldTaskGroup() as group:
+                await group.spawn(p1._message_loop())
+                await group.spawn(p1.htlc_switch())
+                await group.spawn(p2._message_loop())
+                await group.spawn(p2.htlc_switch())
+                await asyncio.sleep(0.01)
+                await group.spawn(pay())
+
+        with self.assertRaises(SuccessfulTest):
+            await f()
+
     async def test_payment_race(self):
         """Alice and Bob pay each other simultaneously.
         They both send 'update_add_htlc' and receive each other's update

```

electrum/lnhtlc.py
```diff
@@ -127,6 +127,16 @@ def recv_fail(self, htlc_id: int) -> None:
             raise Exception(f"(remote) cannot remove htlc that is not there...")
         self.log[LOCAL]['fails'][htlc_id] = {LOCAL: next_ctn, REMOTE: None}
 
+    @with_lock
+    def mark_resolved_onchain(self, htlc_id: int, success: bool) -> None:
+        # so that lnworker.has_unresolved_sent_htlcs() does not return True
```
SomberNight, line 133: ehhh :/  somewhat unsatisfying.

anyway, nits:
`def mark_htlc_resolved_onchain(self, *, htlc_proposer: HTLCOwner, htlc_id: int, success: bool) -> None:`

SomberNight (commented):
ACK first two commits (4025cea73078cac3a27dd558b16c8b3def80b4d2)

still thinking about third commit... (7752591b0a981319b22730fad6faf7fbbfd71d75)

tests/test_lnpeer.py
ecdsa: done

electrum/lnworker.py
ecdsa: this test is still useful, because the paid and resolved situations are not symmetric

electrum/lnhtlc.py
```diff
@@ -127,6 +127,16 @@ def recv_fail(self, htlc_id: int) -> None:
...
 
+    @with_lock
+    def mark_resolved_onchain(self, htlc_proposer: HTLCOwner, htlc_id: int, success: bool) -> None:
+        # so that lnworker.has_unresolved_sent_htlcs() does not return True
+        if self.was_htlc_failed(htlc_id=htlc_id, htlc_proposer=htlc_proposer):
+            return
+        if self.was_htlc_preimage_released(htlc_id=htlc_id, htlc_proposer=htlc_proposer):
+            return
+        MAX_CTN = 2 ** 48 + 1  # means that it was settled offchain
+        self.log[htlc_proposer]['settles' if success else 'fails'][htlc_id] = {LOCAL: MAX_CTN, REMOTE: None}
```
SomberNight: you added `htlc_proposer` like I asked but did so incorrectly.
(before:)
```
self.log[LOCAL]['settles' if success else 'fails'][htlc_id] = {LOCAL: next_ctn, REMOTE: None}
```
(after:)
```
self.log[htlc_proposer]['settles' if success else 'fails'][htlc_id] = {LOCAL: MAX_CTN, REMOTE: None}
```

actually the value needs to be `{htlc_proposer: MAX_CTN, htlc_proposer.inverted(): None}`

electrum/lnhtlc.py
SomberNight: in my "nit section" above I packaged multiple nits :D
- add `htlc_proposer` param
- make it kwarg-only
- put `htlc` in name of method

electrum/lnworker.py
```diff
@@ -1320,7 +1320,22 @@ def has_unresolved_sent_htlcs(self, payment_hash: bytes) -> bool:
...
+                if htlc.payment_hash != payment_hash:
+                    continue
+                if direction != RECEIVED:
+                    continue
+                if chan.hm.was_htlc_failed(htlc_id=htlc.htlc_id, htlc_proposer=LOCAL):
+                    continue
+                if chan.hm.was_htlc_preimage_released(htlc_id=htlc.htlc_id, htlc_proposer=LOCAL):
+                    continue
+                return True
+        return False
```
SomberNight, line 1338: > i.e. the receiver might still take the money

just because the htlc was failed/fulfilled, the receiver can still take the money, until the removal is irrevocably committed, by force-closing the channel

That is to say, both the state before the highlighted diff and after the highlighted diff, is unsafe.

This also implies that `hm.mark_htlc_resolved_onchain` is also insufficient as it would have to "mark" an htlc as irrevocably removed, not just failed/fulfilled.

SomberNight: > ```
>                chan.hm.get_htlcs_in_oldest_unrevoked_ctx(REMOTE),
>                chan.hm.get_htlcs_in_latest_ctx(REMOTE),
>                chan.hm.get_htlcs_in_next_ctx(REMOTE)
> ```

oh boy this loop is confusing. I almost just commented again that I was wrong and the loop is safe, until re-realising it is not safe...

Imagine the remote sends `update_fail_htlc` and then pauses the transport/wire (so that's the last message).
`chan.hm.was_htlc_failed(htlc_id=htlc.htlc_id, htlc_proposer=LOCAL)` is going to be True. So the inner loop hits `continue` for that HTLC (potentially three times).
Of course we would want `has_unresolved_sent_htlcs()` to return True.

The remote at this point can still broadcast e.g. their oldest_unrevoked_ctx, which still contains the htlc, and can then claim the htlc onchain using the preimage.

Basically `hm.was_htlc_failed`/`hm.was_htlc_preimage_released` are the wrong things to check here.
(hmm well, I guess if we got the preimage but the htlc is not yet irrevocably rm-ed, that's ok?
but if the htlc was failed but not yet irrevocably rm-ed, that's not ok.)

see https://github.com/spesmilo/electrum/pull/10962

> ```
>                chan.hm.get_htlcs_in_oldest_unrevoked_ctx(REMOTE),
>                chan.hm.get_htlcs_in_latest_ctx(REMOTE),
>                chan.hm.get_htlcs_in_next_ctx(REMOTE)
> ```

Another minor remark is that conceivably there could be multi-threading issues if we are not running on the asyncio thread here, as we don't hold `hm.lock` and so the state machine could progress between these calls.

ecdsa: > Imagine the remote sends update_fail_htlc and then pauses the transport/wire (so that's the last message).
chan.hm.was_htlc_failed(htlc_id=htlc.htlc_id, htlc_proposer=LOCAL) is going to be True. So the inner loop hits continue for that HTLC (potentially three times).

indeed. what is inside the loop is not consistent with the set being iterated

> This also implies that hm.mark_htlc_resolved_onchain is also insufficient as it would have to "mark" an htlc as irrevocably removed, not just failed/fulfilled.

How about marking both sides:
```diff
-        self.log[htlc_proposer]['settles' if success else 'fails'][htlc_id] = {htlc_proposer: MAX_CTN, htlc_proposer.inverted(): None}
+        self.log[htlc_proposer]['settles' if success else 'fails'][htlc_id] = {LOCAL: MAX_CTN, REMOTE: MAX_CTN}

```

EDIT: I realize that `MAX_CTN` will not work with the 'irrevocably removed' logic, we probably need to use oldest unrevoked (local and remote), at the expense of traceability...

SomberNight:
superseded by https://github.com/spesmilo/electrum/pull/10962

PR #10939, opened 2026-09-07
trampoline forwarding: make sure cltv budget is not negative

ecdsa:

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -4252,11 +4252,10 @@ class LNWallet(Logger):
                     next_onion=next_onion)
                 return
 
-        if not direct_channels:
-            if budget.fee_msat < 1000:
-                raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT, data=b'')
-            if budget.cltv < 576:
-                raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON, data=b'')
+        if budget.fee_msat < (1000 if not direct_channels else 0):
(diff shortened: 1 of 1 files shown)


electrum/lnworker.py
```diff
@@ -4252,11 +4252,10 @@ async def _maybe_forward_trampoline(
...
 
-        if not direct_channels:
-            if budget.fee_msat < 1000:
-                raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT, data=b'')
-            if budget.cltv < 576:
-                raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON, data=b'')
+        if budget.fee_msat < (1000 if not direct_channels else 0):
+            raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_FEE_INSUFFICIENT, data=b'')
+        if budget.cltv < (576 if not direct_channels else 0):
+            raise OnionRoutingFailure(code=OnionFailureCode.TRAMPOLINE_EXPIRY_TOO_SOON, data=b'')
```
SomberNight, line 4258: note: even in the `direct_channels` case, 0 is too low. we should have some positive minimum, like 144, for the ctlv

note that the budget considers the first edge to be free (no cost):
https://github.com/spesmilo/electrum/blob/e7ee1e6cbf1049254b96ed964c6e9b007c982881/electrum/lnrouter.py#L146

SomberNight: split off to separate issue: https://github.com/spesmilo/electrum/issues/10942

PR #10947, opened 2026-09-08
descriptor: add unit tests, rejecting multiple pubkeys

fametrano:
Two error messages in `electrum/descriptor.py` are wrong:

- the `wpkh` branch raises `"More than one pubkey in pkh descriptor"`, naming `pkh` instead of `wpkh`;
- the bare-multisig error `"Cannot have {} pubkeys in bare multisig: only at most 3 pubkeys"` is never formatted, so it prints the literal `{}` rather than the key count.

Both are corrected, and a regression test covers each.

These mirror the same fix in Bitcoin Core's HWI, which shares this descriptor code: bitcoin-core/HWI#857.


ecdsa:
Looks good to me. 
The commit message should be about the added unit tests, this is more important than the fixed error messages.

PR #10953, opened 2026-09-09
qt: fix regression for #6791 introduced in 48916f56b

accumulator:

### DIFF
--- a/electrum/gui/qt/utxo_list.py
+++ b/electrum/gui/qt/utxo_list.py
@@ -306,6 +306,9 @@ class UTXOList(MyTreeView):
         if not coins:
             return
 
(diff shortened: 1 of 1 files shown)


SomberNight:
I would prefer if we wrote commit messages where the first line gives a human-parseable summary of the diff:

so instead of:
```
qt: fix regression for #6791 introduced in 48916f56b
```

the commit message could have been (for example):
```
qt: utxo_list: (fix) copy coins to avoid mutating

fix regression for #6791 introduced in 48916f56b
```

The first line of the commit message is important as many git commands only display that line, same for the github history overview. Even the PR title gets set to that.

PR #10955, opened 2026-09-09
qr scanning: rm zbar, rm barcodescannerview, use zxing-cpp

f321x:
The [java activity based QR code scanner](https://github.com/markusfisch/BarcodeScannerView) we use on Android devices pulls in many dependencies: https://github.com/spesmilo/electrum/pull/10846. 
It also prevents us from bumping the java version used during the APK build:
https://github.com/spesmilo/electrum/blob/c7b6b8d2354ad790b6259b95d143f71f8971291e/contrib/android/Dockerfile#L82-L84
Additionally, having a java activity is a higher maintenance burden than a QML based one which fits better in the codebase and gives more levers like string translation.

This PR replaces the Java based scanner with @accumulator's QtMultimedia based one. 
This scanner was only used on desktop as the QtMultimedia camera interface was unstable, however now with Qt6.10 it seems to work well. 

The PR also replaces the zbar scanner with zxing-cpp (on Android and Desktop). 
This allows to have a single QR scanner dependency instead of two and unify some code paths. 
Zxing performance is also *much* better than zbar and it supports binary QR codes required for some animated QRs. Zxing is accessed through a simple ctypes wrapper to prevent pulling in dependencies required by its official python bindings.

This PR requires the [`electrum_202602_qtmultimedia`](https://github.com/spesmilo/python-for-android/tree/electrum_202602_qtmultimedia) python-for-android branch which enables QtMultimedia and builds FFmpeg.

#### Tested on:

* Pixel 9 (GrapheneOS)
* Samsung Galaxy S8+ (Android 9/API28)
* OnePlus Smartphone with popup camera (see [comment](https://github.com/spesmilo/electrum/pull/10955#issuecomment-5712118195))
* webcam (source): HP Fedora Linux laptop
* webcam (AppImage): same laptop
* webcam (AppImage): debian 13 vm
* screen reader + file reader (AppImage): debian 13 vm
* webcam: windows 11 vm

Doesn't seem to work properly in an Android Emulator with API28 (camera stays black), this seems to be an issue with the emulator.

#### Followups:
* bump Java in Android build to v21 (see https://github.com/spesmilo/electrum/commit/4da7b7f55d85f11d84cfb5995bd6ad1d9d9cb311)
* remove `androidx.core:core:1.16.0` gradle dependency in [`buildozer_qml.spec`](https://github.com/spesmilo/electrum/blob/c7b6b8d2354ad790b6259b95d143f71f8971291e/contrib/android/buildozer_qml.spec#L171) to actually get rid of the huge dependency bloat (see [gist](https://gist.github.com/f321x/073f8318cf3351d23ad630b1d38712fd))
* use zxing for QR encoding?

##### AppImage

```bash
ELECBUILD_COMMIT=3e4be70bbc1fe3993e6e74ec331485cb0e52c89d ELECBUILD_NOCACHE=1 ./contrib/build-linux/appimage/build.sh

reproduced on independent machines:
901adee5d8412451e7ad4c2085dc6103393778a307256317c8f29fe6e5f4b84a  electrum-4.8.2-31-g3e4be70bb-x86_64.AppImage
```

##### Windows

```bash
ELECBUILD_COMMIT=3e4be70bbc1fe3993e6e74ec331485cb0e52c89d ELECBUILD_NOCACHE=1 ./contrib/build-wine/build.sh

reproduced on independent machines:
8b46b3a5f09f7d694bea142255853da313ca921920fc8a0bc3086d7b3d8a8d86  electrum-4.8.2-31-g3e4be70bb-portable.exe
dc7370ab86c7fb0a33e619f6ea7b82178ebd7f0968f25096af84aacd8846dbc9  electrum-4.8.2-31-g3e4be70bb-setup.exe
59f127bebdfa891c6b7ae9d3f46d84a51da7672575b90efc6c35c9d32bbaca9f  electrum-4.8.2-31-g3e4be70bb.exe

```

##### OSX

```bash
ELECBUILD_COMMIT=3e4be70bbc1fe3993e6e74ec331485cb0e52c89d ELECBUILD_NOCACHE=1 ./contrib/osx/make_osx.sh

(note this build will be very slow as homebrew needs to compile new packages)
> only built once (only have a single osx vm)
💬 INFO:  Finished building unsigned dist/Electrum.app. This hash should be reproducible:
7918df6d7ffff1d1696067f2e7bf7af0e6ccc54a8bc4ffc76c674e585ff5a710  -
💬 INFO:  Creating unsigned .DMG
.........................................................................................................................................................................................................
created: /Users/vagrant/electrum/dist/electrum-4.8.2-31-g3e4be70bb-unsigned.dmg
```

##### Android

```bash
ELECBUILD_COMMIT=3e4be70bbc1fe3993e6e74ec331485cb0e52c89d ELECBUILD_NOCACHE=1 ./contrib/android/build.sh qml arm64-v8a release-unsigned

reproduced on independent machines:
5550f70c3b9971ceafefd7c2e3b79c1248e9579809284e66f79b5c8c373cfbbf  Electrum-4.8.2-arm64-v8a-release-unsigned.apk
```


accumulator:
Please check on Android 8/API26, because I ran into troubles when including QtMultimedia on that API level.

ecdsa:
Tested on my phone. The qr scanner works. 

When the app is started and the camera is used for the first time, it first uses the selfie camera before switching to the main camera. (it's a OnePlus, so the selfie camera mechanically pops out)

ecdsa:
> When the app is started and the camera is used for the first time, it first uses the selfie camera before switching to the main camera. (it's a OnePlus, so the selfie camera mechanically pops out)

this is fixed now (fc690b2)

olgahounslow-dev:
- Felix updated Electrum's QR scanning by replacing zbar with zxing-cpp.
   - ThomasV reported that ecdsa confirmed the initial camera bug is fixed.

f321x:
> this is fixed now (https://github.com/spesmilo/electrum/commit/fc690b2b1b757380cd7bd39012011b5b213714e4)

The fix was using FFmpeg as multimedia backend for QtMultimedia instead of the native implementation.
FFmpeg accesses the cameras through the modern Android Camera2 API which provides camera metadata without opening the camera. The native Qt backend required opening the camera to gather this information.
Qt itself recommends using the FFmpeg backend and only uses the native implementation as fallback according to https://doc.qt.io/qt-6.11/qtmultimedia-index.html#target-platform-and-backend-notes.

The downside is we now have to compile FFmpeg and maintain an additional python-for-android recipe.

f321x:
Force pushed from [3e4be70](https://github.com/spesmilo/electrum/commit/3e4be70bbc1fe3993e6e74ec331485cb0e52c89d) to [8002d19](https://github.com/spesmilo/electrum/commit/8002d19c425bde954b6ef7ce34704e718728d80f) for MacOS reproducibility (see https://github.com/spesmilo/electrum/compare/3e4be70bbc1fe3993e6e74ec331485cb0e52c89d..8002d19c425bde954b6ef7ce34704e718728d80f)

```
ELECBUILD_COMMIT=6309c480f407d1f7e9d15762c3c328015981253f ELECBUILD_NOCACHE=1 ./contrib/osx/make_osx.sh

💬 INFO:  Finished building unsigned dist/Electrum.app. This hash should be reproducible:
4513039550ce451dbc46f3df2513b75b8623639698179bd543283e1eca4b9861  -
💬 INFO:  Creating unsigned .DMG
created: /Users/vagrant/electrum/dist/electrum-4.8.2-57-g6309c480f-unsigned.dmg
```

contrib/osx/make_osx.sh
```diff
@@ -155,13 +157,17 @@ else
...
 
-if [ ! -f "$DLL_TARGET_DIR/libzbar.0.dylib" ]; then
-    info "Building ZBar dylib..."
-    "$CONTRIB"/make_zbar.sh || fail "Could not build ZBar dylib"
+if [ ! -f "$DLL_TARGET_DIR/libZXing.dylib" ]; then
+    info "Building zxing-cpp dylib..."
+    # Build zxing with the installed gcc 14 for C++20 support and statically link the runtimes into the library.
+    CC="$(brew --prefix gcc@14)/bin/gcc-14" \
+    CXX="$(brew --prefix gcc@14)/bin/g++-14" \
+    LDFLAGS="-static-libgcc -static-libstdc++ -Wl,-S" \
```
f321x, line 165: Note: libgcc gets statically bundled in the barcode scanner build, and the homebrew package [updates regularly](https://github.com/Homebrew/homebrew-core/commits/main/Formula/g/gcc%4014.rb), this could make the builds nondeterministic if builders end up with different gcc versions.
We could pin all homebrew packages like in this commit: https://github.com/spesmilo/electrum/commit/d509827129f230f6e96967b185b88586bd731e76
But for now i left this out of the branch to not grow it even more, we could try to see how it goes and pick this up if issues arise.

Maybe gcc also becomes obsolete once we bump the MacOS version (e.g. as part of the https://github.com/spesmilo/electrum/pull/10745 migration)

ecdsa:
> 4513039550ce451dbc46f3df2513b75b8623639698179bd543283e1eca4b9861

same for me on OSX.
also, the DMG works for me on OSX

PR #10957, opened 2026-09-09
contrib: build appimage: fetch ca-certificates from pinned sources

f321x:
Fetch ca-certificates from the pinned sources similar to the other packages.
This needs to be done because the default sources of debian 11 stopped working.
Because the pinned sources are https they need to be patched to http before using them to get ca-certificate, as https requires ca-certificates. This is safe as packages are signed by debian (the previous way of installing ca-certificates also used http).


SomberNight (approved):
thx. Looks ok.
will remove this soonish after bumping to newer debian

Issue #5420, opened 2019-06-11
using Trezor on Windows 10, getting LIBUSB_ERROR_ACCESS


c1xx:
Hello,

I'm trying to connect my TREZOR-T device to Electrum 3.3.6 but without success. I get this error:
![electrum-3 3 6_2019-06-11_20-46-43](https://user-images.githubusercontent.com/24668586/59299932-2e354b00-8c8e-11e9-8a52-6ee1ee938590.png)

What can be the problem? I tried some older Electrum versions, but same issues.
On TREZOR web wallet I can connection without any problems.

Thank you,
Lukas

c1xx:
**Issue resolved:** 

I replaced libusb.dll from this GitHub respository: https://github.com/libusb/libusb/releases/tag/v1.0.22

That's all.

Regards,
Lukas

SomberNight:
Where did you replace it? Where was this DLL on your local filesystem? We should already be packaging that version IIRC

c1xx:
> Where did you replace it? Where was this DLL on your local filesystem? We should already be packaging that version IIRC

I took the Electrum Setup package and replaced the "libusb.dll" in installation directory, so in "C:\Program Files (x86)\Electrum\"

SomberNight:
I've just double-checked; we are already bundling that latest version of libusb in the Windows binaries.

The bundled dll is the file at `libusb-1.0.22.7z\MS32\dll\libusb-1.0.dll`
with sha256 hash `1269e6338da5ee738b4b4a24c091773a56559f0b515d002624e675e854c3ea3a`

Could you please try to reproduce the issue with the standalone or the portable binary?
Also, are you sure there was no other application accessing the Trezor at the same time (e.g. the webwallet)?

c1xx:
Yes, there was not other application accesing the Trezor at the same time.
I checked my libsub-1.0.dll and sha256 hash is `67F2551E0132C21214BE01A079E9E41003B06D420DD98794E4981230D41B74BB`

I tried to use my TREZOR with standalone binary, but got same issue...

SomberNight:
> I checked my libusb-1.0.dll and sha256 hash is 67f2551e0132c21214be01a079e9e41003b06d420dd98794e4981230d41b74bb

What is this hash for? Is this the one you downloaded and it started working with? Or is this the hash of the file you got installed alongside Electrum?

c1xx:
[...]

Issue #1887, opened 2016-08-07
getrequest improvment


klerms:
Hi,

firstly, thank you for your excellent work on this project.

I use old merchant.py script, which isn't supported anymore (I have to fix it almost every time a new relase of Electrum is published). The reason for that is, that I want to accept payments to my website with 3 confirmations only.

My modified merchant.py sends the following information to my website when an address status changes: (address, amount, number_of_confirmations). This is very important, because I can inform my clients about the payment status (BROADCASTED, WAITING_FOR_LAST_CONFIRMATION, CONFIRMED)

I suppose that there is no efficient way how can this be achieved via Electrun API. Could you please change the API to contain this functionality?

For example adding lower number of confirmations of all incoming transactions (verified by SPV) and sum of incoming transactions to the getrequest result?

... or adding list of incoming txids to getrequest result (SPV verified)
... or implementing a no-network-request version of getaddresshistory (SPV verified)

Regards

ecdsa:
I agree we will have to improve that.

ecdsa:
would it work if I add a configuration variable for the number of confirmations that is required?

ecdsa:
closing, since duplicate of #1757

PR #10958, opened 2026-09-09
build: update some dependencies

SomberNight:
update some bundled deps:
- cpython
- openssl
- secp
- aiohttp

and some misc others

----

tested:
- [x] build-time: windows build succeeds, still repro
- [x] build-time: appimage build succeeds, still repro
- [x] build-time: macos build succeeds, still repro
- [x] build-time: android build succeeds, still repro
- [x] runtime: windows binary still works
- [x] runtime: appimage binary still works
- [x] runtime: macos binary still works
- [x] runtime: android binary still works


SomberNight:
macOS build is failing on d71b7e2a13016e2013b7052ed23e33e5fc9b717c
```
  File "/Users/vagrant/electrum/contrib/osx/build-venv/lib/python3.13/site-packages/pip/_internal/build_env/installer.py", line 13, in <module>
    from pip._internal.cli.spinners import open_rich_spinner, open_spinner
ImportError: cannot import name 'open_rich_spinner' from 'pip._internal.cli.spinners' (/Users/vagrant/electrum/contrib/osx/build-venv/lib/python3.13/site-packages/pip/_internal/cli/spinners.py)
🗯 ERROR: Could not install build dependencies (mac)
```

Likely due to our pinned pip version being older than what is [bundled](https://github.com/python/cpython/blob/v3.13.15/Lib/ensurepip/__init__.py#L13) by default in cpython 3.13.
Postponing the cpython version bump because of this.

SomberNight:
ah, I forgot to test the android repro build with this. :/
Added checkboxes in the OP for that now.

contrib/deterministic-build/requirements.txt
```diff
@@ -12,14 +12,14 @@ async-timeout==5.0.1 \
...
-    --hash=sha256:024c88eeec92ca068db80f02b8b07c9cef7b9fe261d1d535abfd5abd6f6af432
+certifi==2026.7.22 \
+    --hash=sha256:741e2c3b351ddf169a738da9f2c048608ff7f2c5cc02f1ebc6b118bb090d5d55
 dnspython==2.4.2 \
     --hash=sha256:8dcfae8c7460a2f84b4072e26f1c9f4101ca20c071649cb7c34e8b6a93d58984
 electrum-aionostr==0.1.0 \
     --hash=sha256:3774f8e8312388272e10851a869c9f4d3d4a54d8d564851c36e2dc40297bec84
-electrum-ecc==0.0.7 \
-    --hash=sha256:ed4134e1dbff0fd83022764c6acc97a02cde3512927d7c41f4d48b9a06e91fb2
+electrum-ecc==0.0.8 \
```
f321x, line 21: Should we also bump `electrum_ecc` in `requirements/requirements.txt` to >= 0.0.8 due to the new ABI version libsecp is built with:
https://github.com/spesmilo/electrum/blob/06602fcfa9f1ce1c6b13c30b7308615edf312dbe/contrib/requirements/requirements.txt#L8

https://github.com/spesmilo/electrum-ecc/blob/1ff19059cf8e60df994d518b7fdc91abad310d39/src/electrum_ecc/ecc_fast.py#L47

SomberNight: why? I think new electrum will be compatible with:
- new electrum_ecc + new libsec
- new electrum_ecc + old libsecp
- old electrum_ecc + old libsecp

The only thing we are not compatible with is:
- old electrum_ecc + new libsecp
    - or is this your point?

f321x: > old electrum_ecc + new libsecp
    * or is this your point?

Yeah that's what I meant. Someone building libsecp with the new script but still being on an older `electrum_ecc`. Maybe ok to ignore though.?

f321x:
lgtm at https://github.com/spesmilo/electrum/pull/10958/commits/810274aa69b2f51ba315eafc2dfb110c8a12f3a4
Our apk hashes matched on IRC, also tested a built apk on a Pixel 9.

<details>
<summary> Also ran a deep review of the dependencies with claude: </summary>

# Dependency review — branch `202609_deps`

**Range reviewed:** `b45d9f516..810274aa6` (5 commits, 12 files, +57 / −56)
**Question:** do the new dependencies contain backdoors, and do any API changes regress Electrum?
**Verdict:** **No backdoor found.** Every changed pin was re-downloaded from upstream and re-hashed; all match byte-for-byte. No regression in the test suite. Recommend merging, with the follow-ups in §4.

---

## 1. Provenance — every changed pin re-downloaded and re-hashed

| Pin | Result |
|---|---|
| CPython 3.11.15 → **3.11.16**, 3.12.13 → **3.12.14** | sha512 / sha256 match upstream; detached PGP signatures good (release managers pablogsal / thomas); Sigstore bundles verify. The **old** 3.11.15 / 3.12.13 pins also still match upstream → the pinning process is honest across the bump. |
| OpenSSL 3.0.21 → **3.0.22** | sha512 matches; sha256/sha1 match OpenSSL's published checksum assets; good PGP signature from the documented release key. Download path is `openssl.org` only — no third-party mirror. |
| libsecp256k1 `1a53f496` → **`6e2c8bc4`** | `6e2c8bc4…` **is** `v0.8.0^{commit}`, reachable from upstream master. Annotated, GPG-signed tag. Archive zip sha512 matches today, **and** the old v0.7.1 archive hash still matches → this is a genuine pin, not archive-regeneration drift. Zip contents ≡ `git archive` of the commit. |
| 9 PyPI pins in `requirements.txt` | All are the real **sdist** digests (re-downloaded and hashed locally). None yanked. File remains strictly source-only (27 packages / 27 hashes). |
| `requirements-hw` / `-binaries` / `-binaries-mac` / `-build-base` / `-build-android` | Pinned hash **sets** are exactly the real sdist+wheel pairs for each version — no extra hash, none missing. |
| All 6 changed requirements files | Regenerated from scratch with the project's own `hashin` against live PyPI → **byte-identical to the branch**. 534 referenced artifacts re-downloaded and re-hashed: 534 OK, 0 mismatches. No hand-edited hash block. |
| The 5 commits | All carry **good GPG signatures** from SomberNight's signing subkey `6D7A2116DA909E00AC21108BB33B5F232C6271E9` — the `[S]` subkey in the repo's own `pubkeys/sombernight.asc`. |
| The diff itself | All 113 changed lines classify as version string / hash string / pinned-package line / the one `export PROPCACHE_NO_EXTENSIONS=1`. **Zero** in an "other" bucket. No URL, download source, `verify_hash` call, `assert` or build flag changed or removed; no hash-algorithm downgrade; no non-ASCII / bidi / homoglyph characters; no file mode changes. |

The only non-numeric change, `aiohttp-socks` → `aiohttp_socks`, is genuine tool output: PyPI's `info.name` for that project is now underscored and `hashin` writes that field. pip canonicalises both spellings identically (PEP 503) and nothing in the tree greps the old literal.

## 2. Backdoor hunt on the code that actually changed

**electrum-ecc 0.0.7 → 0.0.8** — the package that handles private keys and signing nonces, so this got the deepest look:

- The PyPI sdist was reproduced **byte-for-byte from the GPG-signed upstream git tag `0.0.8`** (all 195 overlapping files identical; the only 6 extra files are setuptools-generated metadata). The signing key is the SomberNight key already checked into `pubkeys/`.
- The entire Python delta is four things: version string, an asserts-enabled `ImportError` guard, `KNOWN_COMPATIBLE_ABI_VERSIONS [6,5,2,1,0] → [7,6,5,2,1,0]`, and one clarifying comment. **`keys.py` (all signing / pubkey logic), `util.py`, `setup.py`, `pyproject.toml` and `_custom_build/backend.py` are byte-identical to 0.0.7** — so no new install-time code either.
- Behavioural check: 320/320 output fields (RFC6979 ECDSA, low-R grinding, recoverable sigs, Schnorr, pubkeys, point ops) are **bit-identical** between 0.0.7 and 0.0.8; 25 RFC6979 nonces reconstructed from scratch in pure Python match libsecp exactly; all official BIP-340 vectors pass.
- The vendored libsecp256k1 tree inside the sdist is byte-identical to upstream `v0.8.0` — i.e. PyPI users get exactly the version Electrum pins.

**libsecp256k1 v0.7.1 → v0.8.0** — 81 files, +15,549 / −843, but **13,631 of those insertions are the new silentpayments module**, leaving +608 / −460 across the 31 files of the actual crypto core.

- The EC arithmetic diff (`group*`, `ecmult_const*`, `modinv*`, `int128*`, `scalar_impl.h`) is **literally empty**.
- Both tags were built and differentially tested through ctypes over **every symbol electrum_ecc binds** (40 keys × 2 context states; ECDSA via three nonce paths, recoverable ECDSA, ECDH, BIP-340, x-only pubkeys, seckey/pubkey tweak add+mul, tagged SHA256): **output byte-identical**. An independent pure-Python RFC6979 + ECDSA reimplementation reproduces libsecp's exact pubkeys and signatures → nonce derivation is unmodified RFC6979.
- All 167 commits in the range arrived through numbered, maintainer-merged PRs; zero direct pushes.
- Note: `v0.8.0` is signed by a **different maintainer key** (theStack) than v0.7.1. This is a documented, legitimate rotation — the key was added to upstream `SECURITY.md` by existing maintainer Tim Ruffing in PR #1887, a commit that is an ancestor of v0.8.0.

**certifi 2026.6.17 → 2026.7.22** — exact root delta is **4 added / 1 removed** (net 118 → 121): added SECOM TLS RSA/ECC Root CA 2024 and Telia EC/RSA TLS Root CA v3; removed Atos TrustedRoot 2011. The new bundle is an **exact set match** with the roots CCADB lists as Mozilla-included with the Websites trust bit — zero extras, zero omissions. No rogue CA.

**python-socks 2.8.1 → 2.8.2** (Tor / SOCKS path) — the whole delta is 4 files: an IPv6-bracketing fix in the HTTP CONNECT path, a `singledispatchmethod` refactor of the SOCKS5 state machine, the version string, and 3 new tests. **No resolver, connector, rdns, auth or error-handling code changed.** ~120 SOCKS5 wire-byte cases plus 10 error cases are byte- and message-identical between the two versions. Confirmed against real loopback SOCKS5 servers: rdns intact (hostname sent as `ATYP=DOMAIN`, never resolved locally), no direct-connection fallback when the proxy is down or refuses, TLS-over-SOCKS cert/hostname verification unchanged.

**aiohttp, yarl, multidict, propcache, aiohappyeyeballs, typing-extensions, poetry-core, setuptools-scm, build, expandvars** — every sdist matched its upstream git tag (only setuptools-generated metadata differs), and PyPI PEP 740 Sigstore attestations bind them to the projects' own GitHub Actions workflows. aiohttp's pre-generated C (`_http_parser.c`, `_http_writer.c`, `mask.c`, …) was regenerated from the git sources with the upstream-pinned Cython and matched **bit-for-bit**; the vendored llhttp C is byte-identical to nodejs/llhttp's official `release/v9.4.2` artifacts. No install-time code execution, no obfuscated blobs, no network calls, no unexplained hunks in any of them.

## 3. What the bump actually fixes

- **aiohttp 3.14.1 → 3.14.3 is a security update, not housekeeping.** 3.14.1 is affected by **CVE-2026-69244 / GHSA-cq5v-8q36-5273** (HIGH) — an out-of-bounds heap read / use-after-release in the C HTTP response parser error path. 3.14.2 is *not* sufficient; only 3.14.3 fixes it. It also clears CVE-2026-59881 (WebSocket client decompresses RSV1 frames without negotiation — live on the Nostr relay transport) and CVE-2026-69243.
  *Reachability caveat, stated honestly:* the bug was reproduced by driving `HttpResponseParser` directly (216 bytes of adjacent heap leaked into the exception text; SIGSEGV in another harness), but **could not be triggered through `ClientSession`** — on a parse error aiohttp's `ResponseHandler` closes the transport and `connection_lost()` sets `_parser = None`, so the second `feed_data()` the bug needs never happens (verified across 4 malicious-server scenarios + 176 fuzzed malformed-chunked cases). So: real bug, correct fix, no demonstrated exposure for Electrum. Worth noting only the AppImage and macOS builds compile that C parser at all (see §4.5).
- **multidict 6.8.0 fixes five segfaults that reproduce in the pinned 6.7.1** — including a printf argument-order bug where a `const char*` is printed as an integer and a `Py_ssize_t` index is dereferenced as a string pointer, plus a use-after-free on `d.extend(d)`.
- **yarl 1.24.3 closes real host-confusion holes.** Hosts containing Unicode default-ignorable / format code points are now rejected instead of being silently folded — previously an LNURL for `ex<U+00AD>ample.com` displayed as such while the request went to `example.com`.
- **OpenSSL 3.0.22** fixes 5 CVEs; **CPython 3.11.16 / 3.12.14** carry 51 Security-labelled NEWS entries between them.

One change that *could* have been catastrophic was checked explicitly: CPython 3.11.16 rewrites unicodedata's canonical-ordering step, which feeds Electrum's NFKD seed normalisation. Both 3.11.15 and 3.11.16 were built and NFD/NFKD/NFC/NFKC hashed over all 1,112,064 codepoints plus 200k random combining runs plus Electrum's own wordlists — **identical digests**.

## 4. Follow-ups

### 4.1 Version-agnostic build cache can silently ship the old libsecp256k1

All three binary pipelines gate the libsecp build on the mere *existence* of any versioned artifact in `$DLL_TARGET_DIR` (`$CACHEDIR/dlls`, host-persistent and bind-mounted into the containers):

- `contrib/build-linux/appimage/make_appimage.sh:84` — `ls "$DLL_TARGET_DIR"/libsecp256k1.so.*`
- `contrib/build-wine/make_win.sh:46` — `ls "$DLL_TARGET_DIR"/libsecp256k1-*.dll`
- `contrib/osx/make_osx.sh:150` — `ls "$DLL_TARGET_DIR"/libsecp256k1.*.dylib`

`LIBSECP_VERSION` appears nowhere outside `make_libsecp256k1.sh`, so a builder with a warm cache never re-runs the build and ships **v0.7.1** while the branch claims v0.8.0. It fails *open* — electrum-ecc 0.0.8 still lists ABI 6, so the binary works and nothing errors. Worth keying the cache directory or the guard on `LIBSECP_VERSION`.

### 4.2 p4a recipe pins did not move with the requirements files

These were in sync before this branch and now diverge:

| Recipe pin | Requirements pin |
|---|---|
| `contrib/android/p4a_recipes/hostpython3/__init__.py:29` → `build==1.4.0` | `requirements-build-android.txt` → `build==1.5.0` |
| `contrib/android/p4a_recipes/sip/__init__.py:15` → `setuptools-scm==8.3.1` | `requirements-build-base.txt` → `setuptools-scm==9.2.2` |

The recipe copy of `build` is the one that actually shapes the artifact: p4a's `PyProjectRecipe.build_arch()` runs `python -m build --no-isolation --wheel` under hostpython using exactly that pin. Nothing breaks today (8.3.1 and 1.4.0 still install, hashes verified), but it means one release build now uses two different majors of the same tool.

### 4.3 `propcache 0.5.2` declares `setuptools >= 82.0.1`; every build env pins `80.9.0`

Latent only — builds pass because Electrum always uses `--no-build-isolation`, which skips the check (`pip --check-build-dependencies` does flag it). Upstream reverted the identical bump in yarl as spurious, so the cleanest action may be to do nothing and note it.

### 4.4 `NO_EXTENSIONS` asymmetry across build scripts

| Script | AIOHTTP | YARL | MULTIDICT | FROZENLIST | PROPCACHE |
|---|---|---|---|---|---|
| `contrib/make_packages.sh` (android, sdist) | ✅ | ✅ | ✅ | ✅ | ✅ |
| `contrib/build-wine/build-electrum-git.sh` | ✅ | ✅ | ✅ | ✅ | ✅ |
| `contrib/build-linux/appimage/make_appimage.sh` | ❌ | ✅ | ❌ | ✅ | ✅ *(added here)* |
| `contrib/osx/make_osx.sh` | ❌ | ✅ | ❌ | ❌ | ✅ |

Both of the latter install with `--no-binary :all:` and have a C toolchain, so the AppImage and the .dmg compile aiohttp's C parser and multidict's C extension — the latter being precisely the file this branch rewrites — while Windows and Android ship the pure-Python paths.

### 4.5 Strategic / forward-looking

- **OpenSSL 3.0 reached EOL on 2026-09-07.** The bump to 3.0.22 is correct and is the newest 3.0.x, but it is very likely the last release that branch will ever get. The Android APK is the only artifact that ships its own OpenSSL, so from here it has no supported upgrade path on 3.0 — moving the p4a recipe to the 3.5 LTS line (supported to 2030-04-08) is the follow-up.
- **`setuptools_scm[toml]>=3.4.1` is unbounded** (`contrib/requirements/requirements-build-base.txt:15`). setuptools-scm 10.x added a hard runtime dependency on `vcs-versioning`, a separate distribution that is not pinned or hash-pinned anywhere. A dry-run resolve on the documented interpreter today yields setuptools-scm 10.2.3 + vcs-versioning 2.3.4, not the 9.2.2 pinned here. (Context: every 9.0.x and 9.1.x release is yanked on PyPI, which is why the jump is two majors.)
- The AppImage now ships **pure-Python propcache** — a deliberate reproducibility-for-speed trade. Root cause of the reproducibility regression the commit message flags as unexplained: propcache 0.5.0 added a `-ffile-prefix-map` flag that rewrites the build dir to pip's random `/tmp/pip-req-build-<rand>` path, and unlike 0.3.1's DWARF-only paths that string lands in `.rodata` where the AppImage's `strip -R .note.gnu.build-id -R .comment` cannot remove it. Reproduced end-to-end; `PROPCACHE_NO_EXTENSIONS=1` is load-bearing, correctly spelled, and correctly placed (same shell, before the install at `make_appimage.sh:148`).

### 4.6 Low-priority polish

- **yarl's new `ValueError`** on rejected hosts escapes the `except aiohttp.client_exceptions.ClientError` in `electrum/lnurl.py:129,240` and `electrum/submarine_swaps.py:1811`. No crash — the broad `except Exception` handlers at `electrum/payment_identifier.py:363,421` catch it — but the user gets a bare `ValueError` text instead of the intended `LNURLError` message. Adding `ValueError` to those `except` clauses would be tidier.
- **certifi's Atos TrustedRoot 2011 removal** is a Mozilla policy removal, not an expiry (the cert is valid to 2030). `electrum/interface.py:676-683` only treats `verify_code == 18` ("self signed") as the pin-it case; an unknown-issuer failure (code 20) is re-raised, so a server chaining to that root becomes unusable — and if an empty cert file already exists for it, every reconnect fails until the file is deleted by hand. Real-world impact is near zero since Atos is defunct as a public CA, but it is the only certifi change in this branch with a code path behind it.

## 5. Pre-existing issues surfaced (not introduced by this branch)

Listed only because the sweep touched neighbouring pins; none of these are defects in this diff.

- **`urllib3==1.26.20`** (`requirements-hw.txt`) is on a permanently dead branch — 5 advisories, all fixed only on 2.x, and 1.26.20 is the final 1.26 release. Reachability is limited: the headline header-leak CVE requires `ProxyManager…urlopen(assert_same_host=False)` while following cross-origin redirects, which `requests`' high-level API does not do, and urllib3 does not appear in Electrum's own code at all (it arrives via `trezor → requests`, desktop builds only, not Android). Hygiene rather than exposure — but the pin's stated reason (`# prefer older urllib3 to avoid needing hatchling`) is build convenience, which is worth re-weighing now that the branch is unmaintained.
- **`cryptography==46.0.7`** wheels statically link an OpenSSL with a HIGH advisory (GHSA-537c-gmf6-5ccf), and all three desktop builds install it with `--only-binary cryptography`.
- **`pip==25.1.1`** (and `pip==24.0` hardcoded in two p4a recipes) carries 6 advisories, including a fallback-tar-extraction symlink escape — relevant because the builds install nearly everything from sdists.
- **`dnspython==2.4.2`** (CVE-2023-29483, fixed 2.6.1) backs OpenAlias resolution; **`protobuf==3.20.3`**, **`wheel==0.45.1`** (CVE-2026-24049), **`setuptools==80.9.0`** also carry published advisories.
- **Windows and macOS remain on CPython 3.12.10** while the AppImage moves to 3.12.14. This is **forced, not an oversight**: python.org publishes no Windows MSI layout and no macOS `.pkg` for any 3.12 release after 3.12.10 (3.12.11–3.12.14 are source-only). The AppImage and Android builds compile from source, which is exactly why only they could move. The cost is a widening gap of 51 CPython security fixes on those two platforms.
- **`tomli`** is pinned at 2.0.2 (build-base) and 2.4.1 (build-android) and both land in the same Android Docker venv, silently violating the `tomli<=2.0.2` loose constraint. Inert on that venv's Python 3.12, and pre-existing.

## 6. No regression — test evidence

Electrum's full test suite gives **identical results in every configuration**: **1061 passed, 4 skipped, 285 subtests passed** under

1. the old pins,
2. the new pins,
3. an AppImage-style pure-Python propcache/yarl/frozenlist stack,
4. the new pins against an externally built libsecp256k1 v0.8.0, and
5. the new pins on a from-source CPython 3.12.14 + OpenSSL 3.0.22.

pip reported zero resolution conflicts in either direction, and `pip install --require-hashes --no-binary :all:` succeeds for both `requirements.txt` and `requirements-build-base.txt`. The AppImage path was reproduced end-to-end inside the pinned `debian:bullseye` builder: CPython 3.12.14 compiles with the existing flags (its `configure`/`configure.ac` are byte-identical to 3.12.13), the reproducible-buildinfo patch still applies, and every pinned package installs at its pinned version under hash checking. On Android, all four p4a recipe `assert`s still hold upstream, every p4a patch applies to 3.11.16 at identical offsets, secp256k1 v0.8.0 keeps autotools and accepts all ten configure flags, and OpenSSL 3.0.22's `Configurations/` tree is byte-identical to 3.0.21's.

Two API-surface notes, neither of which affects Electrum: aiohttp 3.14.3 removes `hdrs.METH_HEAD_ALL` / `METH_CONNECT_ALL` and makes two `WebSocketReader.__init__` defaults required (no consumer in the pinned set uses them); and electrum-ecc 0.0.8's new asserts-enabled guard refuses to import under `python -O` — a no-op for Electrum, which has had the identical guard at `electrum/__init__.py:39-47` for years.

</summary>

contrib/deterministic-build/requirements.txt
SomberNight: I think it's fine to ignore. That scenario is not the concern of the `electrum` package IMO, a constraint is violated deeper in the dependency-graph: electrum_ecc has both a lower and an upper version limit on libsecp (except electrum_ecc has a shortcoming that it can only check the limits at runtime).

ecdsa:
LGTM

f321x:
<details>
<summary>Claude investigation regarding urllib3/requests/trezor:</summary>

Why ghosting requests doesn't work cleanly

It fails at three layers, each deeper than the last:

Layer 1 — the plugin import. As expected: trezor.py:27 raises, the blanket except Exception swallows it, TREZORLIB=False, plugin silently gone. The user isn't even told the truth — they get "Library version for 'trezor' is incompatible. Installed: 0.20.1", which is a lie about the cause.

Layer 2 — the one nobody predicted, and it's decisive. I verified this myself in the 0.20.1 wheel:

def all_transports():                      # transport/__init__.py:163
    from .ble import BleTransport
    from .bridge import BridgeTransport    # ← unconditional, before the filter
    ...
    return [t for t in transports if t.ENABLED]

.bridge is imported before the ENABLED filter, so Electrum setting BridgeTransport.ENABLED = False at trezor.py:195 cannot prevent it. That means ghosting requests doesn't just kill the bridge — it kills enumerate_devices() (trezor.py:214) and get_transport() (trezor.py:227), i.e. HID, WebUSB and UDP too. Deleting Electrum's own bridge import is necessary but not sufficient; you'd have to vendor a copy of trezorlib's transport list (~45 lines) and keep it in sync forever.

Notably this is an upstream inconsistency: hid, webusb and ble all guard their optional dep with try: import X / ENABLED = X_IMPORTED. Bridge is the only one that doesn't. That's worth a PR to trezor-firmware regardless of what you decide here.

Layer 3 — requests has a second, independent source. The comment at requiremenvia trezor -> requests -> urllib3". That's incomplete — safet==0.1.5 declaresrequests (>=2.4.0) in its own metadata, and safetlib.client → coins → tx_api ils worse than Trezor: its get_library_version() only does import safetlib, which succeeds, so the failure escapes as a traceback from plugin init rather than a release was 2020-02-21 — abandoned for six years.)

And the payoff is smaller than it looks: ghosting drops only requests, urllib3, charset-normalizer from the binaries. certifi and idna stay, because requirements.txt pins them for the base install.

The genuinely good news: the bridge path is already dead

Independent of urllib3, three findings say the bridge preference should go:

- It can't reach a modern Trezor Suite. Suite's built-in bridge dropped port 2-05-26, stable since v26.6.1) while trezorlib still hardcodes TREZORD_HOST ="http://127.0.0.1:21325" — I confirmed that line in the pinned wheel. Electrum's bridge path now only finds a legacy standalone trezord, the product Trezor tells users to uninstall.   - The #5420 rationale expired in 2019. Windows 10 1903 blocked non-admin accesusage — a libusb bug, fixed in libusb v1.0.23 (2019-08-28), three weeks afterElectrum added the bridge preference. Electrum builds libusb v1.0.27.                                                                                                                   - HWI already did it — deleted bridge.py from its vendored trezorlib, no reque
                                                                                                                                                                                          My recommendation: take the other routeAllow urllib3 2.x and accept hatchling. The cost is exactly 6 new pure-python,quirements-build-base.txt — hatchling 1.32.0, hatch-vcs 0.5.0, pathspec 1.1.1,pluggy 1.6.0, tomlkit 0.15.1, trove-classifiers 2026.6.1.19 — 223 KiB total, same shape as existing entries. Nothing in the graph actually caps urllib3 below 2 (requests 2.34.2 declares urllib3<3,>=1.26); the <2 is purely Electrum's own line.

The clincher, which I verified myself: expandvars==1.1.2 — bumped by this veryquirements-build-base.txt — is itself a hatchling-built package. Electrum justnever runs hatchling because build-base installs from wheels. Meanwhile requirements-hw.txt is installed --no-binary :all: --no-build-isolation on all three desktop builds, which is why
urllib3 2.x needs the backend present.

So the trade is: 6 pure-python build-time packages vs. a vendored transport shng bridge + owning trezorlib-internals drift forever — and the hatchling routekeeps Bridge and Safe-T working untouched.

Caveats worth weighing: it adds six upstream maintainer sets to the build-time trust surface (a compromised hatchling runs during every sdist build); hatchling requires-python is >=3.10 while freeze_packages.sh:11 deliberately targets python3.10 — clearing by the smallest margin; and urllib3 2.x hard-requires OpenSSL ≥1.1.1 at import, with the AppImage's bullseye base at 1.1.1w — also the smallest possible margin.

If you do want requests gone

The best version isn't the ghost and isn't aiohttp — it's a ~40-line stdlib htsys.modules before trezorlib imports. It keeps upstream's bridge.pybyte-for-byte, fixes safe_t for free, and was verified end-to-end against a fake trezord including an 8-second blocking read. Its costs are a global sys.modules slot and dev/ship divergence (developers running pip install -e .[hw] never exercise the stub, sion in it).

aiohttp specifically is a bad idea, and this was demonstrated rather than assumed: it forces a sync→async hop into the hwd thread, and stopping Electrum's loop mid-device-read hung the single hwd thread permanently — the test process had to be killed.

</details>

Issue #10921, opened 2026-09-02
PyInstaller/loader/pyimod02_importers.py:450: UserWarning: pkg_resources is deprecated as an API


harryqt:
### Description

`electrum version`

> PyInstaller/loader/pyimod02_importers.py:450: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.

Macos 26, aarch64, M2 chip

Alwoch:
I've taken a look at this. The first thing I noticed is this warning is invisible when installing Electrum the normal way. I Launched
from Finder on macOS 26.5 (Apple M5) there's no warning anywhere ,not on screen and not  in the
system log. The warning goes to stderr, and Finder just throws stderr away. It only becomes visible if you start Electrum from a terminal, which is what the reporter did (`electrum version`). So this doesn't affect normal GUI users at all, only people using the CLI.

Running the official 4.8.1 dmg from a terminal reproduces it exactly:

```
$ /Volumes/Electrum/Electrum.app/Contents/MacOS/run_electrum --offline version
PyInstaller/loader/pyimod02_importers.py:450: UserWarning: pkg_resources is deprecated as an API. ...
4.8.1
```

It shows up on every command including`--help` too, so it happens at startup rather than in `version`. The real output still goes to stdout, so it doesn't break scripts that parse it.

I grepped the electrum repository and found that nothing under `electrum/` imports `pkg_resources`. It comes
from PyInstaller:

- we bundle `packaging` (via qdarkstyle -> qtpy, and via ledger-bitcoin)
- PyInstaller 6.13.0 ships `hook-packaging.py`, whose entire contents are `hiddenimports = ["pkg_resources"]`
- once `pkg_resources` is bundled, PyInstaller automatically runs its `pyi_rth_pkgres.py` startup hook, which imports it
- setuptools 80.9.0, which we pin, prints the warning on import

PyInstaller fixed this in 6.14.0. They deleted `hook-packaging.py` and made the startup hook suppress the warning (pyinstaller/pyinstaller#9151). We pin 6.13.0, set in 538fc37f0 on 2025-06-05, which was two days after 6.14.0 came out, so it looks like it was just missed. 


So the fix would be to bump the pin, in both `contrib/osx/make_osx.sh` and `contrib/build-wine/prepare-wine.sh`:

```diff
[...]

PR #10960, opened 2026-09-10
build: bump pyinstaller to 6.14.0

Alwoch:
This minimal bump is necessary to suppress the warning reported in #10921 (I did explain more of this in the issue thread).  Also bumps the declared minimum for `pyinstaller-hooks-contrib` from `2025.2` to `2025.4` to match what pyinstaller `6.14.0` requires. The pinned version is already `2025.4`, so nothing actually changes at build time.

Ps: I've only tested this on macOS 26.5, python 3.12.14.

### DIFF
--- a/contrib/build-wine/prepare-wine.sh
+++ b/contrib/build-wine/prepare-wine.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 PYINSTALLER_REPO="https://github.com/pyinstaller/pyinstaller.git"
-PYINSTALLER_COMMIT="306d4d92580fea7be7ff2c89ba112cdc6f73fac1"
-# ^ tag "v6.13.0"
+PYINSTALLER_COMMIT="776a57a6b07cb2c71f02018953610bcbfcdaf0aa"
+# ^ tag "v6.14.0"
 
 PYTHON_VERSION=3.12.10
 
--- a/contrib/osx/make_osx.sh
+++ b/contrib/osx/make_osx.sh
@@ -89,8 +89,8 @@ brew install autoconf automake libtool gettext coreutils pkgconfig
 
 info "Building PyInstaller."
 PYINSTALLER_REPO="https://github.com/pyinstaller/pyinstaller.git"
-PYINSTALLER_COMMIT="306d4d92580fea7be7ff2c89ba112cdc6f73fac1"
-# ^ tag "v6.13.0"
+PYINSTALLER_COMMIT="776a57a6b07cb2c71f02018953610bcbfcdaf0aa"
+# ^ tag "v6.14.0"
 (
     if [ -f "$CACHEDIR/pyinstaller/PyInstaller/bootloader/Darwin-64bit/runw" ]; then
         info "pyinstaller already built, skipping"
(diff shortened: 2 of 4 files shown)


SomberNight:
I plan to investigate "major" bumps in many dependencies for non-android next week. (not just pyinstaller, but all over, e.g. freeze_packages, cpython, etc)

After this kind of change, multiple things need to be tested, unfortunately manually atm:
- build still succeeds
- build is still reproducible
- built executable still "works" at runtime (e.g. it at least launches without crashing)

if you are interested in pyinstaller specifically, you could try to bump it to its latest release, or at least see how far it can be bumped. but basically it is not worth it to do a minimal bump just to fix a warning message.

Alwoch:
> if you are interested in pyinstaller specifically, you could try to bump it to its latest release, or at least see how far it can be bumped. but basically it is not worth it to do a minimal bump just to fix a warning message.

Oh I thought bumping to the latest release would be too big of a jump. i'll go ahead and test with the latest release then

Alwoch:
I looked into how far we can actually bump pyinstaller, and ended up finding something that changes
the picture.

The latest version of pyinstaller is 6.22.2. But I don't think we can go past **6.15.0** without bringing in hatchling as build dependency. 6.16.0 switched its build backend to hatchling, and since we install pyinstaller with `--no-build-isolation` (`make_osx.sh:122`, `prepare-wine.sh:98`) the backend has to already be installed. With hatchling absent, `pip install ./pyinstaller` just fails at 6.22.2 with `BackendUnavailable: Cannot import 'hatchling.build'`. It installs fine at 6.15.0. I also noticed the project pins upper bounds in a few places specifically to avoid hatchling (`requirements.txt:11`, `:17`, `requirements-hw.txt:30`, `requirements-build-android.txt:15`), so this looks like something thats being deliberately avoided.

There's a second limit before that one. Anything above 6.14.0 needs a newer `pyinstaller-hooks-contrib` than the 2025.4 thats currently pinned (6.15.0 wants >=2025.8), so it can't be done without a `freeze_packages` run.

While chasing that, I found the warning in #10921 will probably fix itself. It comes from `pkg_resources`, which ships inside setuptools. setuptools deleted `pkg_resources` in 82.0.0 and we currently pin 80.9.0, and `freeze_packages` installs with `--upgrade`, so a re-freeze lands on a setuptools that doesn't have it at all. I tested this:  I left pyinstaller at 6.13.0, untouched, changed only setuptools to 84.0.0, and the warning was gone. `pkg_resources` isn't collected into the bundle at all, and the run-time hook that imports it never gets activated. I also checked that `collect_submodules('pkg_resources')` in our spec files returns `[]` rather than raising when setuptools no longer ships it. It doesnt break the build and just becomes dead code that can be safely dropped.

So given you're doing the dependency upgrades next week, and that involves `freeze_packages`, I figured it doesn't make sense for me to generate one too  so i'll just hand over the findings. Happy for you to close this PR.

For reference, the two ceilings:

| pyinstaller | needs hooks-contrib | build backend |
|---|---|---|
| 6.13.0 (current) | >=2025.2 | setuptools |
| 6.14.0 | >=2025.4 | setuptools |
| 6.15.0 | >=2025.8 | setuptools  (last one before hatchling) |
| 6.16.0+ | >=2025.8 | hatchling |
| 6.21.0–6.22.2 | >=2026.6 | hatchling |

I built and ran 6.14.0, 6.15.0 and 6.22.2 on macOS. All three built with the spec unmodified and produced a binary that launches, reports its version and generates a seed.

PR #10962, opened 2026-09-10
lnworker: dont fail invoice with unresolved htlcs (take 2)

SomberNight:
alternative to https://github.com/spesmilo/electrum/pull/10935

### DIFF
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -3227,6 +3227,28 @@ class LNWallet(Logger):
             upstream_peer.downstream_htlc_resolved_event.set()
             upstream_peer.downstream_htlc_resolved_event.clear()
 
+    def _set_sent_payment_succeeded(self, payment_hash: bytes) -> None:
+        key = payment_hash.hex()
+        info = self.get_payment_info(payment_hash, direction=SENT)
+        if info is not None and info.status != PR_PAID:
+            self.set_invoice_status(key, PR_PAID)
+            util.trigger_callback('payment_succeeded', self.wallet, key)
+
+    def _set_sent_payment_failed(self, payment_hash: bytes) -> None:
+        key = payment_hash.hex()
+        if self.has_unresolved_sent_htlcs(payment_hash):
+            return
+        if self.get_preimage(payment_hash) and self.wallet.get_request(key) is None:
+            # if we know the preimage don't consider the payment failed (unless we pay ourselves).
+            # maybe another htlc of the same mpp got fulfilled, or we saw a htlc-success tx in the mempool
+            # before claiming a revoked htlc with a justice tx which ultimately would make LNWatcher try to fail the htlc here
+            return
+        info = self.get_payment_info(payment_hash, direction=SENT)
+        if info is not None and (info.status == PR_UNPAID and key in self.inflight_payments):
+            # invoice status is PR_INFLIGHT, set it to PR_UNPAID
+            self.set_invoice_status(key, PR_UNPAID)
+            util.trigger_callback('payment_failed', self.wallet, key, '')
+
     def htlc_fulfilled(self, chan: Channel, payment_hash: bytes, htlc_id: int):
         """Called when an HTLC *WE proposed* becomes irrevocably fulfilled."""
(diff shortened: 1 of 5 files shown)


electrum/lnworker.py
SomberNight, line 1072: maybe we should change the definition of the payment status

temp patch to illustrate thinking:
```diff
diff --git a/electrum/lnchannel.py b/electrum/lnchannel.py
index 26f495dde7..02de118826 100644
--- a/electrum/lnchannel.py
+++ b/electrum/lnchannel.py
@@ -1087,6 +1087,8 @@ class Channel(AbstractChannel):
             if direction is not None and htlc_direction != direction:
                 continue
             htlc_proposer = LOCAL if htlc_direction is SENT else REMOTE
+            # TODO maybe we should consider hm.is_htlc_irrevocably_removed_yet(),
+            #      before we set status to failed/settled.
             if self.hm.was_htlc_failed(htlc_id=htlc.htlc_id, htlc_proposer=htlc_proposer):
                 _status = 'failed'
             elif self.hm.was_htlc_preimage_released(htlc_id=htlc.htlc_id, htlc_proposer=htlc_proposer):
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index 7a0ce8e5e2..ba6957ddde 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1068,8 +1068,9 @@ class LNWallet(Logger):
 
         # detect inflight payments
         self.inflight_payments = set()  # type: set[str]  # (not persisted) keys of invoices that are in PR_INFLIGHT state
-        for payment_hash in self.get_payments(status='inflight', direction=SENT).keys():
-            self.set_invoice_status(payment_hash.hex(), PR_INFLIGHT)
+        for payment_hash in self.get_payments(direction=SENT).keys():
+            if self.has_unresolved_sent_htlcs(payment_hash):  # filter out redeemed channels
+                self.set_invoice_status(payment_hash.hex(), PR_INFLIGHT)
 
         # payment forwarding
         self.active_forwardings = self.db.get_dict('active_forwardings')    # type: Dict[str, List[str]]        # Dict: payment_key -> list of htlc_keys

```

SomberNight: > change the definition of the payment status

perhaps the safe definition would be inherently not symmetric :(

- ~if *we* release a preimage offchain, we should already consider it settled,~
- ~but if *they* release a preimage offchain, it's not settled yet until it is irrevocably committed.~

but this seems like a non-trivial change

EDIT: will need to think more about this.

ecdsa: I realize `get_payments(status=xxx)` is really confusing: it does not return the list of payments in status xxx.

ecdsa: the `get_payments` part of the patch should definitely be applied

electrum/lnworker.py
```diff
@@ -1316,6 +1316,19 @@ def get_payments(
                 out[payment_hash] += plist
         return out
 
+    def has_unresolved_sent_htlcs(self, payment_hash: bytes) -> bool:
+        """Returns whether there are htlcs we sent for this payment that have neither
+        been failed nor fulfilled yet, i.e. the receiver might still take the money.
+        """
+        for chan in self.channels.values():
+            if chan.is_redeemed():
+                continue  # skip channel
```
ecdsa, line 1330: Note for the future: `is_redeemed()` might never be reached, if there are htlc outputs that are not economical to sweep. this means an invoice might remain stuck in `PR_INFLIGHT` indefinitely. Not worth fixing for now, IMO.

electrum/lnworker.py
SomberNight: > the `get_payments` part of the patch should definitely be applied

added in 44ac7a27a58b75b73dcba5abe7391cb11e34efe5

ecdsa:
Thanks, it looks good to me.

I pushed an extra commit on top of the branch: d4fbccd7f4dd919316aa8c9471ac7c371ff7482d

SomberNight:
ack d4fbccd7f4dd919316aa8c9471ac7c371ff7482d

PR #10963, opened 2026-09-11
plugin: nwc: fix stuck shutdown, followup #10689

f321x:
The `NWCServer` was still catching `asyncio.CancelledError` however this is not neccessary anymore since 9b425692 (#10689). Catching the `CancelledError` made it block daemon shutdown:
```
 24.48 | I | daemon.Daemon | stopping plugins
 24.48 | I | plugin.Plugins | stopped
 24.48 | I | daemon.Daemon | removing lockfile
 24.48 | I | daemon.Daemon | stopped
 24.48 | D | plugins.nwc.nwcserver.NWCServer | Restarting nwc event handler
^CException ignored while joining a thread in _thread._shutdown():
  File "/usr/lib64/python3.14/threading.py", line 1583, in _shutdown
    _thread_shutdown()
KeyboardInterrupt:
```

Also check the wallet triggering the `close_wallet` callback so the server doesn't get stopped when an unrelated wallet gets closed.

### DIFF
--- a/electrum/plugins/nwc/nwcserver.py
+++ b/electrum/plugins/nwc/nwcserver.py
@@ -79,17 +79,24 @@ class NWCServerPlugin(BasePlugin):
         self.initialized = True
 
     @hook
-    def close_wallet(self, *args, **kwargs):
+    def close_wallet(self, wallet: 'Abstract_Wallet', *args, **kwargs):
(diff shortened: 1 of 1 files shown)


SomberNight (approved):
just skimmed this but looks ok.

PR #10966, opened 2026-09-11
contrib: appimage: respect ELECBUILD_NOCACHE during runtime build, bump deps

f321x:
Consider the ELECBUILD_NOCACHE environment variable also for the
build of the type2-runtime built in a separate docker container.

Bump the pinned alpine packages (and base image) of the docker container in which we
build the type2-runtime for appimages.


SomberNight (approved):
looks good. thx

PR #10967, opened 2026-09-13
tests: qml: restore the process locale after each QETestCase

fametrano:
I hit this running the test suite on a machine whose locale is Italian:
tests/qml fails one test, and the same file passes when I run it alone.

QETestCase.setUp() constructs a QCoreApplication, and Qt sets the process
locale from the environment while doing so. Nothing puts it back, so every
test that runs later in the same process formats dates and numbers in that
locale instead of the one the interpreter started in.

pytest collects tests/qml/test_qml_qeconfig.py, which is a QETestCase,
before tests/qml/test_qml_qetransactionlistmodel.py, which is not. By the
time test_format_date_by_section checks strftime('%a, %H:%M') it gets
'gio, 14:30' instead of 'Thu, 14:30'.

Under LANG=it_IT.UTF-8, pytest tests/qml gives 1 failed, 7 passed on
master and 8 passed with this patch. Under LC_ALL=C both give 8 passed,
which is why CI never sees it: the runner images default to C.UTF-8,
where '%a' renders the same either way.

The patch saves the locale in setUp(), before the Qt thread starts, and
restores it from a cleanup, so the restore happens after tearDown(), with
the application already exited and its thread joined. I tried restoring
right after the application is constructed instead: under LC_ALL=C, Qt
moves LC_CTYPE to UTF-8 for itself and reports that it did, and an early
restore undoes that while Qt is still running.


f321x:
I shortened the comment as it seemed a bit excessive. Thanks for the contribution.

PR #10970, opened 2026-09-14
bolt11: bugfixes and cleanup

f321x:
Fix some minor bugs in the bolt11.py module and remove t-tag support. See individual commits.


electrum/bolt11.py
```diff
@@ -65,14 +65,18 @@ def unshorten_amount(amount) -> Decimal:
         'u': 10**6,
         'm': 10**3,
     }
-    unit = str(amount)[-1]
+
     # BOLT #11:
     # A reader SHOULD fail if `amount` contains a non-digit, or is followed by
     # anything except a `multiplier` in the table above.
-    if not re.fullmatch("\\d+[pnum]?", str(amount)):
+    if not re.fullmatch(f"[1-9][0-9]*[{''.join(units)}]?", amount):
```
f321x, line 72: note: if a user has saved a malformed invoice this stricter parsing could raise on them. 
However the risk seems very low as 
a.) the invoice would already be non spec compliant, and 
b.) we were just doing a wallet upgrade (v73) master which drops raising invoices, which would land in the same release.

electrum/bolt11.py
SomberNight, line 340: this makes the `get_routing_info` method's API confusing: the `tag` input param should just be removed I guess?

SomberNight (commented):
ACK 8c5a946c903b3e55c70110a98d591cab16c13c7c.
Just a minor nit.

electrum/bolt11.py
SomberNight: Sounds fine.

electrum/bolt11.py
f321x: Yeah true, removed in https://github.com/spesmilo/electrum/pull/10970/commits/e78fba28f8a435742b32de3ada4eca66384059f7

SomberNight:
thx

Issue #6167, opened 2020-05-20
RBF of CPFP tx


ziggamon:
Current master

If a big inbound transaction gets stuck due to high mempool it's common to do CPFP. However, if the initial CPFP doesn't work electrum does not allow the user to bump further.
CPFP option gets disabled on the original transaction (not quite sure why), and RBF of the child transaction shows an interface that measures tx fee in sat/B for the child transaction, requiring quite complex math to recalculate to what that should be based on the original transaction, the size of the child transaction.

This could be solved either by:
- Don't disable CPFP on a transaction after it has been CPFP'd.
- Show a full RBF UI to RBF the bump transaction, calculating the fees based on the entirety of the transactions that are being pulled.

ecdsa:
A transaction cannot be CPFP'd if all its outputs that belong to you are already spent.
That is typically be the case if you already CPFP'd it.

ziggamon:
Why not?

And if it really can't, then why not show the proper RBF UI? I'd say it's a common use case to try to RBF gradually and then increase it further if it's needed.

ecdsa:
> Why not?
because that does not make sense. 
CPFP means that it shhould be possible to create a child transaction.

bumping the fee of the child tx is what you actually want to do.

ziggamon:
>because that does not make sense.

I think you're confusing. CPFP happens to an incoming transaction. The child transaction is still incoming to my wallet. I could CPFP it again. 

I agree with you that bumping the fee of it would also help, and I'm asking for better UI for it. It's not reasonable that the user should have to calculate the fee manually and retrofit it into a sats/byte rate of the child transaction. It's far too complex. Better just show the original CPFP dialog and let the user set the fee for the entire block of transactions that are being confirmed.

ecdsa:
> I think you're confusing. CPFP happens to an incoming transaction. The child transaction is still incoming to my wallet. I could CPFP it again

So you want to create a grand-child transaction?
That sounds pretty wasteful compared to RBF
However, I do not think that is disabled.

ziggamon:
It is wasteful, a decent RBF UI for CPFP would also work of course, as mentioned.

>However, I do not think that is disabled.

It is.

Please try this out yourself by CPFPing a transaction, and then try to bump it further and you'll see what I mean.

ecdsa:
oh indeed, I just checked the code, CPFP is disabled when RBF is possible.
which makes sense, we should not propose CPFP when RBF is possible.

[...]

Issue #2396, opened 2017-05-07
Pay attention when parent is unconfirmed AND low fee


gits7r:
Scenario: Using Dynamic fees on Electrum, with setting: within 5 blocks. We receive an incoming payment, transaction A. It is marked as Low fee and at date it says: unknown (low fee) - this is OK and expected.

We send a payment that is comprised of parents:
transaction A (the low fee one from above)
transaction B (another confirmed input from the wallet or another unconfirmed input with high enough fee).

This payment results in transaction C. Set to dynamic fees with confirmation within 5 blocks, Electrum will calculate a fee good enough for transaction C so that it gets mined within the expected time frame, but instead it will take about 5 days because of input transaction A with low fee. This is very problematic, the user expects a confirmation within 5 blocks and gets it in few days - received a report of someone who lost some money until the coins reached the exchange wallet to be sold (price dropped significantly). If the entire amount of the wallet (max) is sent in this transaction C, CPFP is also out of the question so there's nothing the user can do. Having the user know in advance about a certain transaction (either incoming or outgoing) that might not get confirmed soon means a LOT and makes the difference in many cases. 

Suggestions: 
a) When such cases appear, we mark the resulting transaction C as **unconfirmed parent + low fee for at least one parent** with **Date: unknown - at least one of the parents has a low fee.**

b) If we plan to automatically enable RBF for low fee transactions, it should also include it for transactions with good fees but which have at least 1 unconfirmed parent with low fee. (if there is a ticket open for auto-RBF on low fee please someone link it here, I couldn't find it).

c) Somehow include this logic in our sweep function - we should handle fees better in this function.

d) (Not sure about this?) calculate a bigger dynamic fee for transaction C in such cases, so that it will cover transaction A as well and get it mined faster. This might be unreliable and have little effect, because not all miners implement CPFP. 

This issue closes also #2393 (same issue, better explained here) as well as other issues closed as "not a bug". These changes will not affect anything in case some solution for capacity increase is adopted in the close future and fees drop but ensures we have it ready for when there's competition for block space.

gonderie:
What if I want to cancel Uncomfirmed? Or wait only?

i send FEE within 25 blocks , i send payment 2 times in week. At today i do not finist

gits7r:
[...]

Issue #3470, opened 2017-12-07
Child pays for parent does not include grandparents in the size calculation


Raulo:
If you have unconfirmed grandparents, only direct parent will be presented in the total size dialog after selecting "child pays for parent" dialogue. 

Obviously, the calculated fee (or manually selected fee computed with the suggested size) will be smaller than the value with grandparents that will be taken into account  by the miners.

SomberNight:
I have spent some time thinking about this recently but failed to find a "trustless" solution.

First note that there is little point in considering just the grandparent, what we would want is consider the whole unconfirmed tx "package": the minimum set of unconfirmed transactions that would need to get mined to also get "tx" mined.

What the client would need in the end is to get the size of this package and the overall fee these transactions are paying. E.g. see the `getmempoolentry` [bitcoind RPC](
https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getmempoolentry/) (`ancestorsize` and `fees`>`ancestor`).

-----

If we were okay with trusting the server with these stats, the server could just call `getmempoolentry` and forward the results to the client. We already trust the server with the fees of incoming transactions.

I think trusting the server here might be more dangerous than trusting it just for fees of individual incoming txs, as
- the client would only call this RPC if they want to CPFP (or maybe RBF fee bump)
- this package can be quite large, considering current Bitcoin p2p policy, I think up to 100 transactions, each potentially 400k weight
- the server provides the size of the package, even if the client uses a relatively low feerate the absolute fee could be huge

and so a malicious server could trick a user to significantly overpay on fees.

-----

The client could try to calculate the package size/fee itself. For this it could download all the relevant unconfirmed transactions from the server (traverse prevouts until it reaches confirmed txns), along with merkle proofs for the confirmed ancestors.

This would be several network requests for a flow (CPFP) that currently needs none (the initial dialog at least). The package can be large, as above, this could be a lot of data. We could bound the max number/size of txs to download and just display an error if it's over that (although that would be really annoying for the user).

Still, at first glance, traversing the ancestors back to confirmed txns and SPV-checking those could provide a trustless solution. However, even this scheme does not work due to the lack of tx non-inclusion proofs for the blockchain. That is, a malicious server could inflate the package by claiming that parent transactions which have been mined are "unconfirmed". This way, the client could be forced to overpay on fees greatly, again.

Transisto:
How does a server benefit from over-estimating the fees needed for a CPFP?

[...]

PR #10973, opened 2026-09-15
qt: TxEditor: show effective feerate when spending unconfirmed coins

sashazykov:
Added an effective feerate hint to the tx editor. It is shown below the existing _"This transaction will spend unconfirmed coins."_ message and it's display only.

This solution is deliberately simple for reviewability and does not cover the general case. It only considers the chain of ancestors already in the wallet history, at most one unconfirmed parent transaction at each level. I believe this covers most of the cases when the user is interested in effective feerate (e.g. using CPFP to bump their own stuck transaction).

I also have a solution that considers all unconfirmed ancestors that are in the wallet history, but it is much more complex and with the [cluster mempool](https://bitcoin.org/en/releases/31.0/#mempool) in Bitcoin Core 31 it is no longer exact either when ancestors are shared.

Tested on mainnet by creating a CPFP transaction while the previous one was still unconfirmed:

Effective feerate in the tx editor

Related: #2396, #6167


sashazykov:
Tested it with RBF:

Screenshot 2026-09-16 at 12 30 53

Also tested with a chain of two unconfirmed ancestors.

sashazykov:
Bumped the fee with an RBF transaction. mempool.space now shows 2 decimals for fee rates. Do you think Electrum could also switch to showing two decimals?

mempool.space RBF timeline

SomberNight:
some prior discussion in https://github.com/spesmilo/electrum/issues/3470

> This solution is deliberately simple for reviewability and does not cover the general case. It only considers the chain of ancestors already in the wallet history, at most one unconfirmed parent transaction at each level. I believe this covers most of the cases when the user is interested in effective feerate (e.g. using CPFP to bump their own stuck transaction).

Arguably, one just as interesting case for CPFP is using it for an "incoming" tx. In that scenario there might be a much more complex graph of mempool txs that would need to be considered :/

Another CPFP usecase is to get a lightning channel-related tx mined. In that case usually it is a simple 1-parent-1-child CPFP, where the parent should already be in the wallet history.

sashazykov:
Thanks for pointing to #3470.

I did not think of the incoming tx case when I wrote the description. My bias: I send transactions much more often than I receive them.

The simple incoming case works when the incoming coin is spent from the send tab or in the RBF dialog: the tx is in the wallet db, its fee is reported by the server (same as in the history), and the server tells us whether it has unconfirmed parents itself. If it does, we cannot see them and the hint says that not all parents are taken into account. The CPFP dialog does not show the hint yet.

The lightning case works too, as far as I can tell from the code. lnwatcher shares the wallet's adb, so the funding or commitment tx is in the wallet db and the chain walk finds it as the parent. Its fee is known: computed if we funded the channel, and for a commitment tx because the funding output is adb.is_mine; otherwise it is the server fee, as in the incoming case. The anchor sweep after a force close goes through the txbatcher without a fee dialog, so the hint does not apply there.

A possible follow-up would be to use `get_unconf_ancestors_info` in the "Child pays for parent" dialog too.

sashazykov:
Tested it with an incoming transaction and it works.

When there is more than one unconfirmed parent transaction, the effective feerate is not shown, only the note that it could be lower:

Screenshot 2026-09-22 at 13 07 40

Is that wording ok for this case?

PR #10975, opened 2026-09-15
qt: TxEditor: friendlier strings for send change to lightning

f321x:
The strings in `TxEditor` shown to the user if the send change to lightning option is enabled can be somewhat scary if the user isn't very familiar. 

This rewords them a bit, e.g. `not sending change to lightning` -> `Change will stay on-chain, ...`.

I have not added an indicator on which configuration caused this string to show up, like `Change will stay on-chain, ... [Option Send change to Lightning].`, as this seems quite long. 
However it could help the user know what actually makes this message show up, not sure.


electrum/gui/qt/confirm_tx_dialog.py
```diff
@@ -187,6 +188,9 @@ def on_tab_changed(self, index):
     def is_batching(self) -> bool:
         return self._base_tx is not None
 
+    def send_change_to_lightning_available(self) -> bool:
+        return self.config.WALLET_SEND_CHANGE_TO_LIGHTNING and self.context == TxEditorContext.PAYMENT
```
SomberNight, line 192: What about sending the change to ln when funding a new channel? :P
Partly kidding: I have never tried that, not sure if it works - but I guess it might

SomberNight (commented):
mostly ACK 0a186637834031ea4db5b4fa1e848cb6ecdc874b

electrum/gui/qt/confirm_tx_dialog.py
f321x: We don't currently do that, `send_change_to_lightning` is not passed to `make_unsigned_transaction` in `mktx_for_open_channel`. 
To do so we would also have to change how the funding tx is broadcast (after the swap htlcs have arrived) and consider the anchor channel reserve?
We could, but this is probably not worth it.

ecdsa:
LGTM

PR #10979, opened 2026-09-17
2fa mobile cosigner

ecdsa:
- this probably requires UR2.0 QR codes (#7905)
- we should consider adding the `xprv` fingerprint(s) of a wallet file to its encrypted header


electrum/plugins/trustedcoin/trustedcoin.py
```diff
@@ -64,187 +58,176 @@ def get_signing_xpub(xtype):
...
+    xpub2 = keystore.from_xprv(xprv2).get_master_public_key()
+    cosigner = {
+        'xpub1': xpub1,
+        'xpub2': xpub2,
+        'xpub3': xpub3,
+        'xprv2': pw_encode_with_version_and_mac(xprv2.encode('utf8'), password),
+    }
+    cosigners = [c for c in get_cosigners(config) if c['xpub2'] != xpub2]
+    cosigners.append(cosigner)
+    config.PLUGIN_TRUSTEDCOIN_COSIGNERS = cosigners
```
SomberNight: This is really blurring the boundaries of wallet vs config. It is abusing the config IMO.

Also, what if someone has multiple 2fa wallets?

ecdsa: > This is really blurring the boundaries of wallet vs config. It is abusing the config IMO.

could be moved to a separate file. the point is that is should not be a wallet.

> Also, what if someone has multiple 2fa wallets?

that's why `cosigners` is a list.

PR #10980, opened 2026-09-18
add_info_from_network: do not swallow exceptions

ecdsa:
An attacker can provide a witness_utxo whose value disagrees  with the real previous transaction. That would raise PSBTInputConsistencyFailure, which is swallowed because it is not a NetworkException.

This would lead us to sign a tx with the wrong input amount. (no fund loss: the tx is invalid under BIP143)


SomberNight (approved):
looks good

PR #10982, opened 2026-09-20
wizard: redirect flow if a 2fa seed is entered for a standard wallet.

ecdsa:
This fixes a regression introduced by the new wizard. Note that the reverse direction (standard seed entered for a 2FA wallet) is not redirected; I believe that is less of an issue.


SomberNight:
test_wizard.py would love to learn about this scenario :)

f321x:
[3dfd254](https://github.com/spesmilo/electrum/pull/10982/commits/3dfd254755e98396c1aebb5115d384af7b231b66) lgtm

> test_wizard.py would love to learn about this scenario :)

I added [70b4d27](https://github.com/spesmilo/electrum/pull/10982/commits/70b4d27cf55d284c41e1d22b5ff4ef684d36de91) which covers the new wizard path

PR #10984, opened 2026-09-21
lnwallet: fix discovery of onchain backups

f321x:
Stop gating `maybe_add_backup_from_tx` behind `tx_is_related`. 
If the wallet is being restored from seed and the funding tx is sent 
before we knew about the transactions funding its inputs `tx_is_related`
would skip adding the backup and the channel would never be discovered.

Also prevent onchain channel backups from overriding imported channel backups
and add new unittests covering this.


electrum/lnworker.py
```diff
@@ -1831,14 +1833,41 @@ def make_local_config_for_new_channel(
...
-        return chacha20_decrypt(key=self.backup_key, data=encrypted_data, nonce=nonce)
+        if version == 0:
+            return funding_scripthash[0:12]
+        # prevent replaying the channel backup in another tx by binding it to the funding tx inputs
+        prevouts = b''
+        for txin in sorted(funding_inputs, key=lambda txin: (txin.prevout.txid, txin.prevout.out_idx)):
+            prevouts += txin.prevout.serialize_to_network()
+        preimage = funding_scripthash + prevouts
+        nonce = bip340_tagged_hash(tag=b'electrum/onchain_channel_backup/nonce', msg=preimage)[0:12]
+        return nonce
```
SomberNight: if we change the nonce, we should certainly commit to the full output: not just the scriptpubkey but also the satoshis value

electrum/lnworker.py
```diff
@@ -4009,24 +4009,41 @@ def maybe_add_backup_from_tx(self, tx):
...
+        channel_id: bytes = cb_storage.channel_id()
+        for chan in itertools.chain[AbstractChannel](self.channels.values(), self.channel_backups.values()):
+            if chan.get_funding_address() != funding_address:
+                assert channel_id != chan.channel_id
+                continue
+            if not isinstance(chan, ChannelBackup) or chan.is_imported \
+                    or channel_id == chan.channel_id:  # same channel backup added again
+                return  # not overriding existing channel or imported channel backup
+            # there is a conflicting onchain backup tx. we prefer the older one. this only happens with v0 backups.
+            height_new_backup = max(self.wallet.adb.get_tx_height(funding_txid)._height, 0) or TX_HEIGHT_INF
```
SomberNight: don't use `TxMinedInfo._height`, use `.height()` instead.
`_height` is not SPV-verified, it fully trusts the server.

f321x: Changed the approach, the height is not used anymore.

electrum/lnworker.py
f321x: Changed the approach, the backup version stays unchanged.

electrum/lnworker.py
```diff
@@ -3994,6 +3994,9 @@ async def _request_fclose(addresses):
             raise Exception('failed to connect')
 
     def maybe_add_backup_from_tx(self, tx):
+        funding_txid = tx.txid()
+        if any(funding_txid == c.funding_outpoint.txid for c in self.get_channel_objects().values()):
+            return
```
SomberNight, line 3999: I think some code comments would be helpful for this check.
```
# Check we don't override imported backups or full channels.
# note: currently no support for batched channel opens
```

electrum/address_synchronizer.py
```diff
@@ -366,15 +366,22 @@ def add_value_from_prev_output():
                     # give v to txi that spends me
                     next_tx = self.db.get_spent_outpoint(tx_hash, n)
                     if next_tx is not None:
+                        spender_had_txi = bool(self.db.get_txi_addresses(next_tx))
                         self.db.add_txi_addr(next_tx, addr, ser, v)
                         self._add_tx_to_local_history(next_tx)
+                        if not spender_had_txi:  # the spender was added before we knew it spends our coins
+                            spender_tx = self.db.get_transaction(next_tx)
+                            assert spender_tx
+                            util.trigger_callback('adb_updated_tx', self, next_tx, spender_tx)
```
SomberNight, line 375: note: so this only triggers `adb_updated_tx` when we learn the first txi being is_mine for `next_tx`, but not for subsequent ones. E.g. maybe `next_tx` has two inputs that ultimately would both be is_mine, and we separately learn for each input that they are is_mine. `adb_updated_tx` would as-is only fire when we learn the first input to be is_mine but not when we learn the second one. Is this ok? Is this intended?

For the onchain chan backup use-case this does not matter though, as `maybe_add_backup_from_tx` only requires `any` input to be is_mine.

electrum/lnworker.py
```diff
@@ -3993,7 +3993,14 @@ async def _request_fclose(addresses):
         if not success:
             raise Exception('failed to connect')
 
-    def maybe_add_backup_from_tx(self, tx):
+    def maybe_add_backup_from_tx(self, tx: Transaction):
+        assert self.wallet.adb.db.is_in_verified_tx(tx.txid())
+        if not any(self.wallet.is_mine(self.wallet.adb.get_txin_address(txin)) for txin in tx.inputs()):
+            # only allow funding tx with inputs of our wallet to prevent replay of the channel backup.
+            # note: is_mine can be false if the parent tx was not yet received from the server,
+            #       so this also has to be called when its parent arrived (adb_updated_tx)
```
SomberNight, line 4001: The race is not about the ordering of receiving the parent vs the child tx first in general, but about whether the gap limit has already been rolled forward enough to discover the tx inputs.

- Consider a parent tx1:
  - input0 at m/1/1000
  - output0 at m/0/1001
  - output1 at m/1/0
- and a Child tx2:
  - input0 at m/0/1001  (this spends `tx1:0`)
  - output0 at m/1/1

Now regardless of whether we add tx1 or tx2 to adb first, it will be only much later in the sync process when we derive the `m/0/1001` address. Once that address is derived and we download its history using the Synchronizer, `adb.receive_history_callback` will call `adb.add_transaction` again for both tx1 and tx2:
https://github.com/spesmilo/electrum/blob/bef804218f8b15ea198378551a694afce1e01d0b/electrum/address_synchronizer.py#L470

Following comment is talking about this:
https://github.com/spesmilo/electrum/blob/bef804218f8b15ea198378551a694afce1e01d0b/electrum/address_synchronizer.py#L297-L300

-----

So something like this might be clearer:

```
# note: is_mine can be false during initial wallet synchronization: we might learn
#       of more-and-more inputs of being is_mine, as we roll the gap_limit forward.
#       Hence maybe_add_backup_from_tx also needs to be called on adb_updated_tx.
```

electrum/lnworker.py
SomberNight, line 1: even though looks like we will not change the OCB format for now, we should add a comment for our future selves, e.g.:
```diff
diff --git a/electrum/lnworker.py b/electrum/lnworker.py
index 566ed75004..2fd4f91f32 100644
--- a/electrum/lnworker.py
+++ b/electrum/lnworker.py
@@ -1839,6 +1839,9 @@ class LNWallet(Logger):
     def encrypt_cb_data(self, data: bytes, funding_address: str) -> bytes:
         funding_scripthash = bytes.fromhex(address_to_scripthash(funding_address))
         nonce = funding_scripthash[0:12]
+        # note: would have been nice, if besides the funding_script, we also committed to
+        #       the funding_amount (sats), and maybe all the inputs of the funding_tx. Without that,
+        #       the OP_RETURN can be replayed.
         # note: we are only using chacha20 instead of chacha20+poly1305 to save onchain space
         #       (not have the 16 byte MAC). Otherwise, the latter would be preferable.
         return chacha20_encrypt(key=self.backup_key, data=data, nonce=nonce)

```

electrum/lnworker.py
```diff
@@ -3993,34 +3996,41 @@ async def _request_fclose(addresses):
...
+            # note: if the channel was funded with wallet-external UTXOs we won't detect the backup (we don't do this).
+            return
+        funding_txid = tx.txid()
+        if any(funding_txid == c.funding_outpoint.txid for c in self.get_channel_objects().values()):
+            return
         funding_address = None
         node_id_prefix = None
         for i, o in enumerate(tx.outputs()):
```
SomberNight, line 4012: maybe add a comment:
```
# note: loop is quadratic but that's ok as we require at least 1 tx input to be is_mine
```

SomberNight (commented):
Some comments. Reviewed at 78ef07e3e2cbadddbc5a853fecea0b5fa6254004.

electrum/lnwatcher.py
```diff
@@ -110,16 +111,28 @@ async def on_event_blockchain_updated(self, *args):
...
+        if adb.db.is_in_verified_tx(tx_hash):
+            self.lnworker.maybe_add_backup_from_tx(tx)
         await self.trigger_callbacks()
 
     @event_listener
-    async def on_event_adb_added_verified_tx(self, adb, tx_hash):
+    async def on_event_adb_updated_tx(self, adb: AddressSynchronizer, tx_hash: str, tx: Transaction):
         if adb != self.adb:
             return
+        if adb.db.is_in_verified_tx(tx_hash):
```
SomberNight, line 126: just a note: pretty much everywhere else in the codebase, this check would have been written as `adb.get_tx_height(tx_hash).height() > 0`
I guess it's mostly the same though. Well, `adb.get_tx_height` would also check that the raw tx is available in `db.transactions` but you have your own checks for that in these usages.

electrum/lnworker.py
f321x: Added comments in github.com/spesmilo/electrum/pull/10984/commits/e52a24d1a4acacdb5a5fe5d899461f9577b194ef

electrum/address_synchronizer.py
f321x: > Is this intended?

yes, I added this condition so the callback fires less often, considering that it is fine for the (only) use case of detecting if a transaction is touching our wallet. 
But I see what you mean, it could be misleading for future consumers of `adb_update_tx`, for simplicity i just trigger the update unconditionally now in https://github.com/spesmilo/electrum/pull/10984/changes/1508f99534ca429db9a48e0b338d8710e3f4b2f9

electrum/lnworker.py
f321x: thanks for the clearer explanation, I updated the comment: https://github.com/spesmilo/electrum/pull/10984/changes/1508f99534ca429db9a48e0b338d8710e3f4b2f9

electrum/lnworker.py
f321x: Added in https://github.com/spesmilo/electrum/pull/10984/commits/e0156a9ec5fac7d574d6cd595ae4a8a4faec6a21

electrum/lnworker.py
f321x: Added in https://github.com/spesmilo/electrum/pull/10984/commits/b07b42c3cfc31dcb6194a06bfcaf5281b5a032a4

electrum/lnwatcher.py
f321x: For this use case `is_in_verified_tx` seems clearer, the name says why it's being used unlike the opaque height check, and it's a single fast dict lookup. Should be fine as tx is already passed to the callback?

electrum/address_synchronizer.py
SomberNight: How about:
```diff
diff --git a/electrum/address_synchronizer.py b/electrum/address_synchronizer.py
index 06051fa4a1..7ae302d6d4 100644
--- a/electrum/address_synchronizer.py
+++ b/electrum/address_synchronizer.py
@@ -366,11 +366,12 @@ class AddressSynchronizer(Logger, EventListener):
                     # give v to txi that spends me
                     next_tx = self.db.get_spent_outpoint(tx_hash, n)
                     if next_tx is not None:
-                        self.db.add_txi_addr(next_tx, addr, ser, v)
+                        is_new_txi = self.db.add_txi_addr(next_tx, addr, ser, v)
                         self._add_tx_to_local_history(next_tx)
-                        spender_tx = self.db.get_transaction(next_tx)
-                        assert spender_tx
-                        util.trigger_callback('adb_updated_tx', self, next_tx, spender_tx)
+                        if is_new_txi:
+                            spender_tx = self.db.get_transaction(next_tx)
+                            assert spender_tx
+                            util.trigger_callback('adb_updated_tx', self, next_tx, spender_tx)
             # add to local history
             self._add_tx_to_local_history(tx_hash)
             # save
diff --git a/electrum/wallet_db.py b/electrum/wallet_db.py
index 7aa8e73c74..e3f289b31d 100644
--- a/electrum/wallet_db.py
+++ b/electrum/wallet_db.py
@@ -1702,7 +1702,8 @@ class WalletDB(JsonDB):
         return {int(n): (v, cb) for (n, (v, cb)) in d.items()}
 
     @modifier
-    def add_txi_addr(self, tx_hash: str, addr: str, ser: str, v: int) -> None:
+    def add_txi_addr(self, tx_hash: str, addr: str, ser: str, v: int) -> bool:
+        """Returns True if the item was newly added to the DB, or False if it was already there."""
         assert isinstance(tx_hash, str)
         assert isinstance(addr, str)
         assert isinstance(ser, str)
@@ -1712,7 +1713,10 @@ class WalletDB(JsonDB):
         d = self.txi[tx_hash]
         if addr not in d:
             d[addr] = {}
+        if d[addr].get(ser) == v:
+            return False
         d[addr][ser] = v
+        return True
 
     @modifier
     def add_txo_addr(self, tx_hash: str, addr: str, n: Union[int, str], v: int, is_coinbase: bool) -> None:

```

f321x: Seems like a good compromise, added this and a similar guard to the other `adb_updated_tx` trigger in https://github.com/spesmilo/electrum/compare/04b295782bb01c7cc174f9516e66b6d59f6c0fe2..a2271a573f0d945e76ce182dac56a609f243bfac.

